PR #220, opened 2017-07-12
Can generate blocks through RPC

NicolasDorier:

PR #221, opened 2017-07-12
Make RPC methods easier to create with binders, improve testing

NicolasDorier:
Based on top of https://github.com/stratisproject/StratisBitcoinFullNode/pull/220#discussion_r126923128

* add support for net462 build, this make debug experience better, and we can use profilers
* stratis node runner use a preloaded wallet, this will make the node faster to bootstrap for tests
* creation of model binders for the rpc controller
* block generation only wait the block repository to be coherent at the end

If you have problem debugging the project, please make sure to uncheck `Use managed Compatibility Mode`

image

### DIFF
--- a/Stratis.Bitcoin/Miner/PowMining.cs
+++ b/Stratis.Bitcoin/Miner/PowMining.cs
@@ -17,6 +17,14 @@ namespace Stratis.Bitcoin.Miner
 {
 	public class ReserveScript
 	{
+		public ReserveScript()
+		{
+
+		}
+		public ReserveScript(Script reserveSfullNodecript)
+		{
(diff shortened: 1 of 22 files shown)


Stratis.Bitcoin/Miner/PowMining.cs
```diff
@@ -167,6 +167,7 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
 					// ensure the block is written to disk
 					ulong retry = 0;
 					while (++retry < maxTries &&
+                           nHeight != nHeight &&
```
dangershony: Super duper

NicolasDorier: oops lol

PR #222, opened 2017-07-12
Feature/xmldoc stratis.bitcoin.builder

Aprogiena:
Adds XML documentation comments to all files in Stratis.Bitcoin\Builder and also fixes several violations of the code style (mostly use of TABs instead of spaces).

This PR should not change ANY functionality of the code at all, it is all about comments and code style.

### DIFF
--- a/Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
+++ b/Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
@@ -29,9 +29,9 @@ namespace Stratis.Bitcoin.Builder.Feature
         void BuildFeature(IServiceCollection serviceCollection);
 
         /// <summary>
-        /// Initializes the list of delagates to configure DI services of the feature registration.
+        /// Initializes the list of delegates to configure DI services of the feature registration.
         /// </summary>
-        /// <param name="configureServices">List of deleages to configure DI services of the feature registration.</param>
+        /// <param name="configureServices">List of delegates to configure DI services of the feature registration.</param>
         /// <returns>This interface to allow fluent code.</returns>
         IFeatureRegistration FeatureServices(Action<IServiceCollection> configureServices);
 
@@ -90,9 +90,9 @@ namespace Stratis.Bitcoin.Builder.Feature
         }
 
         /// <summary>
-        /// Initializes the list of delagates to configure DI services of the feature registration.
+        /// Initializes the list of delegates to configure DI services of the feature registration.
         /// </summary>
-        /// <param name="configureServices">List of deleages to configure DI services of the feature registration.</param>
+        /// <param name="configureServices">List of delegates to configure DI services of the feature registration.</param>
         /// <returns>This interface to allow fluent code.</returns>
         public IFeatureRegistration FeatureServices(Action<IServiceCollection> configureServices)
         {
--- a/Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
+++ b/Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
@@ -7,72 +7,126 @@ using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Builder.Feature
 {
-	public interface IFeatureRegistration
-	{
-		Type FeatureStartupType { get; }
-		Type FeatureType { get; }
+    /// <summary>
+    /// Defines methods for a representation of registered features of the FullNode.
+    /// </summary>
+    public interface IFeatureRegistration
+    {
+        /// <summary>
+        /// Type of the feature startup class. If it implements ConfigureServices method, 
+        /// it is invoked to configure the feature's services.
+        /// </summary>
+        Type FeatureStartupType { get; }
 
-		void BuildFeature(IServiceCollection serviceCollection);
-		IFeatureRegistration FeatureServices(Action<IServiceCollection> configureServices);
-		IFeatureRegistration UseStartup<TStartup>();
-	}
+        /// <summary>Type of the feature class.</summary>
+        Type FeatureType { get; }
 
-	public class FeatureRegistration<TImplementation> : IFeatureRegistration where TImplementation : class, IFullNodeFeature
-	{
-		public readonly List<Action<IServiceCollection>> ConfigureServicesDelegates;
+        /// <summary>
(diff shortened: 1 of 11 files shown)


bokobza:
@dangershony I'm happy to do this one

Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
```diff
@@ -7,72 +7,126 @@
...
+        /// <param name="serviceCollection">Collection of feature registration's DI services.</param>
+        void BuildFeature(IServiceCollection serviceCollection);
 
-		public FeatureRegistration()
-		{		
-			this.ConfigureServicesDelegates = new List<Action<IServiceCollection>>();
-			this.FeatureType = typeof(TImplementation);
-		}	
+        /// <summary>
+        /// Initializes the list of delagates to configure DI services of the feature registration.
```
bokobza: typo in "delegates"

Stratis.Bitcoin/Builder/Feature/FeatureRegistration.cs
```diff
@@ -7,72 +7,126 @@
...
+        /// </summary>
+        /// <param name="configureServices">List of deleages to configure DI services of the feature registration.</param>
```
bokobza: typo in "delegates"

Stratis.Bitcoin/Builder/FullNodeBuilder.cs
```diff
@@ -9,175 +9,227 @@
...
+        /// <summary>List of delegates that configure the service providers.</summary>
+        private readonly List<Action<IServiceProvider>> configureDelegates;
+
+        /// <summary>List of delegates that add services to the builder.</summary>
+        private readonly List<Action<IServiceCollection>> configureServicesDelegates;
+
+        /// <summary>List of delegates that add features to the collection.</summary>
+        private readonly List<Action<IFeatureCollection>> featuresRegistrationDelegates;
+
+        /// <summary>true if the Build method has been called already (whether it succeeded or not), false otherwise.</summary>
```
bokobza: capital T at 'true'.

Stratis.Bitcoin/Builder/FullNodeBuilder.cs
```diff
@@ -9,175 +9,227 @@
...
+        private bool fullNodeBuilt;
+
+        /// <summary>Collection of features available to and/or used by the node.</summary>
+        public IFeatureCollection Features { get; }
+
+        /// <summary>User defined node settings.</summary>
+        public NodeSettings NodeSettings { get; set; }
+
+        /// <summary>Specification of the network the node runs on - regtest/testnet/default.</summary>
```
bokobza: regtest/testnet/**mainnet**

bokobza:
Overall, great PR, thank you for doing this! There are only a few typos I've noticed.

The only suggestion I have is about comments in inherited classes.
If you look at FeatureCollection.cs for example, the  AddFeature<>() method has the same comment in the interface where it is defined and in the class implementing the interface (FeatureCollection). Obviously if we change the comment in one place, we need to change it in the other and this might lead to differences.
the solution is to comment on the interface and to use  /// <inheritdoc /> on the implementation methods.
Like this
```
  /// <summary>
  /// Defines methods for collection of features of the FullNode.
  /// </summary>
    public interface IFeatureCollection
    {        
        /// <summary>Adds a new feature to the collection provided that the feature of the same type has not been added already.</summary>
        /// <typeparam name="TImplementation">Type of the feature to be added to the collection.</typeparam>
        /// <returns>Representation of the registered feature.</returns>
        IFeatureRegistration AddFeature<TImplementation>() where TImplementation : class, IFullNodeFeature;
    }

    /// <summary>
    /// Collection of features available to and/or used by the FullNode.
    /// </summary>
    public class FeatureCollection : IFeatureCollection
    {       
        /// <inheritdoc />
        public IFeatureRegistration AddFeature<TImplementation>() where TImplementation : class, IFullNodeFeature
        {
            if (this.featureRegistrations.Any(f => f.FeatureType == typeof(TImplementation)))
                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, "Feature of type {0} has already been registered.", typeof(TImplementation).FullName));

            var featureRegistration = new FeatureRegistration<TImplementation>();
            this.featureRegistrations.Add(featureRegistration);

            return featureRegistration;
        }
    }

```

This way, you don't need to replicate the comments and it has the added benefit of you knowing at a glance which method are from an interface.
Have a look at how we did it in WalletManager.cs.

Aprogiena:
tx @bokobza, I will fix the reported typos 

as for inheritdoc - The problem is that MSVS does not work with it. On one hand I do understand DRY and how having the same comment twice makes updates difficult. On the other hand, those XML comments are very handy when one writes the code as you get help from IDE and inheritdoc seems not to give you anything. If you are working with the interface, it obviously works, because the real comment is with the interface, but if you are working with the default class implementing that interface, then you have nothing. And as I can see some code indeed does work with the default class, not interface. So the question here is - what is the primary purpose for us to have XML doc? I can see two major purposes, one is to be able to understand the code quicker - for which inheritdoc is good enough and the second one is to have the help from IDE in writing the code, for which inheritdoc does not work. If you honor DRY principle over the later purpose, then OK, I'll replace the duplicity with inheritdoc. Please let me know.

bokobza:
@Aprogiena I very much like your approach of "purpose over principle".
The purpose of the Xmldoc is to help the developer in understanding the code.
When you have inheritdoc, although you don't see the comment by looking at the inherited member itself, you know that the comment will be on the interface. And when you're using the member somewhere else, you can (as you said) see the doc for this member.
So for me, the inheritdoc does what it's supposed to do: 
- give some info about a method when you need to use it
- tells you that the present method is inherited and that the comment is in the interface method
- prevent comments from begin out of sync
- avoid repeating yourself

Also, there is no principle I'm basing myself on, it's just that from experience, repeating comments on inherited members is unnecessary and leads to discrepancies over time :-)

dangershony:
I don't think we should spend too much time on comment structure.

I say just use  /// <inheritdoc /> on implementations, normally you don't need external comments on implementations if you do it means you don't use the the interface properly.

I would say:
- proper comments on Interfaces 
- targeted comments where needed in implementations 
- readable code, prefer readability over complexity when writing the code

Aprogiena:
all should be fixed now except for true->True, which was intended

PR #223, opened 2017-07-12
Feature/remove get service alias

Aprogiena:
Removing Service extension method, which was just an alias to GetService method because there is no reason not to use GetService method directly. Moreover, the Service alias method was put into extension class that it did not belong to.

Note that this PR requires https://github.com/stratisproject/StratisBitcoinFullNode/pull/222 to be merged first.


bokobza:
@dangershony I can take this one as well

bokobza:
Great, thank you

PR #227, opened 2017-07-13
Introduce WalletAccountReference type

NicolasDorier:
The pair accountname and walletname will often travel together, this make sense to create a value object to bundle them together instead of passing them as individual string.

### DIFF
--- a/Stratis.Bitcoin/RPC/Controllers/WalletRPCController.cs
+++ b/Stratis.Bitcoin/RPC/Controllers/WalletRPCController.cs
@@ -12,18 +12,6 @@ namespace Stratis.Bitcoin.RPC.Controllers
 {
     public class WalletRPCController : BaseRPCController
     {
-        public class UsedWallet
-        {
-            public string WalletName
-            {
-                get; set;
-            }
-            public HdAccount Account
-            {
-                get;
-                set;
-            }
-        }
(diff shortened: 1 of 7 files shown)


NicolasDorier:
Do not push now, not building

Stratis.Bitcoin/RPC/Controllers/WalletRPCController.cs
```diff
@@ -41,39 +29,28 @@ public IWalletManager WalletManager
...
+            var account = GetAccount();
+            var address = this.WalletManager.GetUnusedAddress(account);
             return mining.GenerateBlocks(new ReserveScript(address.Pubkey), (ulong)nBlock, int.MaxValue);
         }
 
-        private UsedWallet GetWallet()
+
+        private WalletAccountReference GetAccount()
         {
+            //TODO: Support multi wallet like core by mapping passed RPC credentials to a wallet/account
```
dangershony: An option is pass to the RPC a wallet and accoumt names and if null default to first()

NicolasDorier: The goal of RPC controller is to be compatible with what has been developped for core. Else I would use `WalletController` 
Core is using RPC creds for mapping to right wallet.

PR #234, opened 2017-07-13
Base refactor

dangershony:
A base refactor to get some more structure with all the various components 

Base feature - come with the minimal components to run a node, such as connect to peers and track the longest chain of headers.

All feature now moved under a Features folder, in the future features will move to their own project and nuget package.


bokobza:
looks good

PR #235, opened 2017-07-13
First batch of wallet manager tests

Neurosploit:
First part of tests for the walletmanager. Continuing with the next batch. When the walletmanager class is fully tested I will refactor the tests and optimize to speed up wallet creation so they run faster.

### DIFF
--- a/Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
+++ b/Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
@@ -37,11 +37,10 @@ namespace Stratis.Bitcoin.Features.Wallet
 
         /// <summary>
         /// Loads a wallet from a file.
-        /// </summary>
-        /// <param name="password">The user's password.</param>
+        /// </summary>        
         /// <param name="name">The name of the wallet.</param>
         /// <returns>The wallet.</returns>
-        Wallet LoadWallet(string password, string name);
+        Wallet LoadWallet(string name);
 
         /// <summary>
         /// Recovers a wallet.
--- a/Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
+++ b/Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
@@ -38,9 +38,10 @@ namespace Stratis.Bitcoin.Features.Wallet
         /// <summary>
         /// Loads a wallet from a file.
         /// </summary>        
+        /// <param name="password">The user's password.</param>
         /// <param name="name">The name of the wallet.</param>
         /// <returns>The wallet.</returns>
-        Wallet LoadWallet(string name);
+        Wallet LoadWallet(string password, string name);
 
         /// <summary>
         /// Recovers a wallet.
--- a/Stratis.Bitcoin/Features/Wallet/WalletManager.cs
+++ b/Stratis.Bitcoin/Features/Wallet/WalletManager.cs
@@ -17,18 +17,18 @@ using Transaction = NBitcoin.Transaction;
 
 namespace Stratis.Bitcoin.Features.Wallet
 {
-    /// <summary>
-    /// A manager providing operations on wallets.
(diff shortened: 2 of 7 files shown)


Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
```diff
@@ -37,11 +37,10 @@ public interface IWalletManager : IDisposable
 
         /// <summary>
         /// Loads a wallet from a file.
-        /// </summary>
-        /// <param name="password">The user's password.</param>
+        /// </summary>        
         /// <param name="name">The name of the wallet.</param>
         /// <returns>The wallet.</returns>
-        Wallet LoadWallet(string password, string name);
+        Wallet LoadWallet(string name);
```
bokobza: At the moment, the password is not used, but that's because we're not encrypting the wallets yet.
That's because we want to have quick access to the file so that it's easier to debug.
So the password was there as a placeholder really, because we know we'll need it in the future.

Stratis.Bitcoin/Features/Wallet/WalletManager.cs
```diff
@@ -413,7 +421,8 @@ public int LastBlockHeight()
 				return this.chain.Tip.Height;
 			}
 
-			return this.Wallets.Min(w => w.AccountsRoot.Single(a => a.CoinType == this.coinType).LastBlockSyncedHeight) ?? 0;
+            // todo: use max instead of min?
```
bokobza: I see where the confusion comes from.
It's true that it's called "LastBlockHeight" but in fact this doesn't describe the last block height that was indeed received but rather the height of the block from which we want to sync.
So imagine that you have 2 wallets, one synced up to block 500 and the other one up to block 800. Although the last block received is 800, we're going to behave as if the last block received is 500 because we need to sync from there.
Does that make sense?

bokobza:
Really good work. thank you so much for doing these extensive tests.
Can you please check why some tests are failing in the build?
The Appveyor build is set to not fail if there are errors (because we had a few failures before) but I'll change that now.

bokobza:
Leave the failing tests to me, I'll have a look.

Stratis.Bitcoin/Features/Wallet/IWalletManager.cs
Neurosploit: Ok, I'll put it back in. I think there was another place where I removed a parameter I have to check.

Stratis.Bitcoin/Features/Wallet/WalletManager.cs
Neurosploit: I need to remove that todo. I found out after testing the next method as well. It makes sense but I would rename the method to GetSyncStartBlockHeight etc.

Neurosploit:
The failing tests are the same as I reported here: https://github.com/stratisproject/StratisBitcoinFullNode/issues/214. I'm not that familiar with travis yet. Could be good to have a guide on how to resolve these for failing IC on platforms other than your own.

Stratis.Bitcoin/Features/Wallet/WalletManager.cs
```diff
@@ -366,7 +367,7 @@ private List<string> CreateAddressesInAccount(HdAccount account, int addressesQu
 			for (int i = firstNewAddressIndex; i < firstNewAddressIndex + addressesQuantity; i++)
 			{
 				// generate new receiving address
-				var pubkey = this.GenerateAddress(account.ExtendedPubKey, i, isChange);
+				var pubkey = this.GenerateAddress(account.ExtendedPubKey, i, isChange, this.network);
```
bokobza: here you can actually remove the Network.
I'll bring this change in anyway.

Neurosploit: Ok, that will be part of the next change

PR #238, opened 2017-07-15
Bugfix/issue 233

### LINKED ISSUE #233: Blockpuller DistributeDownload bug
aprogiena [3:43 PM] 
ok then

[3:44] 
so BlockPuller.cs in Stratis.Bitcoin.BlockPulling

[3:44] 
DistributeDownload method

[3:44] 
int minHeight parameter

[3:44] 
it is called from AskBlocks as 
DistributeDownload(vectors, nodes, downloadRequests.Min(d => d.Height));

[3:45] 
in downloadRequests i believe we have list of blocks that we want to download and we do have d.Height for each of them - i.e. their height in blockchain

[3:46] 
so we have a bunch of blocks we want to download and we have a bunch of nodes we can download from and this DistributeDownload method is about to distribute these download tasks among those nodes

[3:46] 
now we have a (already updated) comment in DistributeDownload:
// Be careful not to ask the block from a node that does not have it

[3:46] 
enforced as 
if (behavior.ChainBehavior?.PendingTip?.Height >= minHeight)
                    selectnodes.Add(behavior);

[3:47] 
i.e. for each of the node we check whether we know how long their vision of the blockchain is and only if it is long enough (>= minHeight) we add the node among candidates to receive download task (selectnodes)

[3:48] 
if so far so good, here comes my confusion and i hope i did not mess that up otherwise i'll look like an idiot :slightly_smiling_face:

dangershony [3:49 PM] 
This is all correct yes (hard to look like an idiot this is not so strait forward) (edited)

aprogiena [3:49 PM] 
consider we have peer nodes A, B, C, D with their vision of the reality (length of the chains they have) as follows:
A = 15
B = 16
C = 16
D = 11

[3:50] 
now assume we have last block no. 9 and so we want blocks 10, 11, 12, 13, 14, 15, 16 (edited)

[3:51] 
this would mean that 
DistributeDownload(vectors, nodes, downloadRequests.Min(d => d.Height));
would translate as 
DistributeDownload(vectors, nodes, 10);

[3:51] 
that would make the filtering  at the start of DistributeDownload not filter out any of the nodes

[3:52] 
so even node D would remain in the selection

[3:52] 
what i'm confused about is that after that filtering, there is no minHeight or d.Height used at all

[3:52] 
which looks like D could be asked for blocks 12, 13, 14, 15, 16

[3:52] 
even if it does not have them

[3:53] 
which contradicts the comment // Be careful not to ask the block from a node that does not have it

dangershony [3:53 PM] 
Well spotted bug.....!!!

[3:53] 
It might explain why download stalls sometimes

aprogiena [3:53 PM] 
i.e. i believe there should be Max() not Min()

[3:54] 
i.e. DistributeDownload(vectors, nodes, 16);

[3:54] 
or completely done in different manner

dangershony [3:54 PM] 
It should be both I guess

zeptin [3:54 PM] 
hmm, i'm not familiar with that part of the code, but in that case D would be a candidate for at least downloading 10 and 11, right? you might be unnecessarily restricting the pool of available nodes by being too pessimistic

aprogiena [3:54 PM] 
or checking for node height even after that filtering (i.e. in index selection)

[3:54] 
or whatever

[3:55] 
@zeptin exactly, optimally you would like not to disqualify D from giving you 10 and 11

[3:55] 
but if we just naively replaced Min with Max, that would do it

[3:55] 
so i don't thinkg just Min->Max is the right solution

[3:55] 
although it is the easiest one

zeptin [3:56 PM] 
i assume that A through D are nodes that have already been contacted that we know the definite height for

aprogiena [3:56 PM] 
actually they don't need to be (maybe) because there is a check 
if (behavior.ChainBehavior?.PendingTip?.Height >= minHeight)

dangershony [3:56 PM] 
Consider this
A = 15
B = 16
C = 16
D = 11
E = 4
F = 9 (us)

aprogiena [3:56 PM] 
so i assume those guys with ? can be null if you don't know their state of reality (edited)

dangershony [3:57 PM] 
So we need to filter out E and D so you need min and max height

[3:58] 
This is your to fix @aprogiena

zeptin [3:58 PM] 
i assume kicking slow nodes happens later once you've actually connected to them and started downloading

[3:58] 
because it is not necessarily true that the nodes with the longest chain are the fastest peers

aprogiena [3:58 PM] 
@dangershony yeah, you would like to filter out those that can't really help you at all, and among others you want to download assign tasks with respect to their chain length

[3:59] 
@dangershony well :confused: tx :smile:

[3:59] 
@zeptin there is quality score that is respected when distributing those tasks; this is kind of experience with the node

[4:00] 
so nodes with greater quality score are asked more than others

dangershony [4:00 PM] 
You found it you fix it :joy: and you also get 2 honour points :pray:

Aprogiena:
Separation of download distribution strategy from block puller to its own class + test of the class on the case that was previously buggy.

This needs very careful review as this touches the code in block puller for which there seems to be no test.

Fix #233


Aprogiena:
Note especially that BlockPuller.AskBlocks newly has a requirement that downloadRequests are unique - i.e. you can not have requests for the same block within one call of AskBlocks. For what I checked, this should trivially hold (but still should be reviewed - esp. the call from LookaheadBlockPuller). Similarly, the same was introduced for AssignPendingVectors and there ConcurrentBag pendingInventoryVectors is used, so it is possible to violate the requirement there unless pendingInventoryVectors is guaranteed to hold this requirement as well. It seems that it might be the case, but it is non-trivial to see.

In general, it makes a good sense to only try to download one block once at one try, the question is whether we should check for (and ignore if there is duplicity) it or whether it is implicit (thus crashing in case of violation would not matter).

dangershony:
Awesome i will review when i can

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// Implements a strategy for a block puller that needs to download 
+    /// a set of blocks and from its connected peers.
+    /// </summary>
+    public static class DownloadAssignmentStrategy
+    {
+        /// <summary>Information about a connected peer.</summary>
+        public class PeerInformation
+        {
+            /// <summary>Evaluation of the node's past experience with the peer.</summary>
+            public int QualityScore;
```
bokobza: any reason fields are used instead of properties? 
If not, please always use properties rather than public fields.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -383,70 +392,59 @@ protected void OnStalling(ChainedBlock chainedBlock)
...
+                        PeerId = behavior,
+                        ChainHeight = behavior.ChainHeadersBehavior.PendingTip.Height
+                    };
+                    peerInformation.Add(peerInfo);
+                }
             }
-            innernodes = selectnodes.ToArray();
 
-            if (innernodes.Length == 0)
+            Dictionary<DownloadAssignmentStrategy.PeerInformation, List<int>> assignBlocksToPeers = DownloadAssignmentStrategy.AssignBlocksToPeers(requestedBlockHeights, peerInformation);
```
bokobza: probably better if this is called blocksAssignedToPeers. The following foreach is then clearer to read.

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+            Dictionary<PeerInformation, List<int>> res = new Dictionary<PeerInformation, List<int>>();
+            foreach (PeerInformation peerInformation in availablePeersInformation)
+                res.Add(peerInformation, new List<int>());
+
+            foreach (int blockHeight in requestedBlockHeights)
+            {
+                // Only consider peers that have the chain long enough to be able to provide block at blockHeight height.
+                List<PeerInformation> filteredPeers = availablePeersInformation.Where(p => p.ChainHeight >= blockHeight).ToList();
+
+                int[] scores = filteredPeers.Select(n => n.QualityScore == BlockPuller.MaxQualityScore ? BlockPuller.MaxQualityScore * 2 : n.QualityScore).ToArray();
```
bokobza: I know this was in the previous version of the code but I'm not clear why the score doubles if it's already at max.

Aprogiena: I don't know that either, so I preserved it. I was thinking about it and I think it does make a good sense, because such a peer is considered "perfect" and "perfect" is just much better than anything else, which means there is extra bonus for it. So the score of 50 is 2x worse than the score of 100, but it is not 3x worse than the perfect 150, because we translate 150 to 300, so 50 is 6x worse than 150. 

This is probably just randomly picked up alg. but no problem with it. It would be nice to know the intention of the author, so we could verify that the code matches that, but we do not know that, so I preserved it until someone makes an objection with why this is not good.

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        /// <summary>
+        /// Choose random index proportional to the score.
+        /// </summary>
+        /// <param name="scores">Array of scores.</param>
+        /// <param name="totalScore">Sum of the values in <paramref name="scores"/>.</param>
+        /// <returns>Random index to <paramref name="scores"/> array - i.e. a number from 0 to scores.Length - 1.</returns>
+        private static int GetNodeIndex(int[] scores, int totalScore)
+        {
+            int pickScore = Rand.Next(totalScore);
```
Aprogiena: np with that, will change

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
Aprogiena: ok, np, will fix

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -383,70 +392,59 @@ protected void OnStalling(ChainedBlock chainedBlock)
...
+
+            // There are no available peers with long enough chains.
+            if (peerInformation.Count == 0)
```
Aprogiena: absolutely, this is mistake, will fix

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: yes, will fix

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
Aprogiena: changed to selectedScore

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
bokobza: fair enough :-)

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
bokobza: 👍 it wasn't broken in the first place :-)

bokobza:
As far as I'm concerned it's good to go.
@dangershony can you please cast your expert eyes on this when you return?

bokobza:
nice fix :-)

Stratis.Bitcoin.Tests/BlockPulling/DownloadAssignmentStrategyTest.cs
```diff
@@ -0,0 +1,106 @@
...
+                    case "A":
+                        // Peer A should not get any work.
+                        Assert.Equal(0, assignedBlockHeights.Count);
+                        break;
+
+                    case "B":
+                    case "C":
+                    case "D":
+                        // Peers B and C should only get tasks to download blocks up to its chain height.
+                        // Peer D can get be assigned anything.
```
dangershony: remove the "get"

Stratis.Bitcoin.Tests/BlockPulling/DownloadAssignmentStrategyTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        [Fact]
+        public void AssignBlocksToPeersWithNodesWithDifferentChainsCorrectlyDistributesDownloadTasks()
+        {
+            // Create list of numbers 6 to 40 and shuffle it.
+            Random rnd = new Random();
+            List<int> requiredBlockHeights = new List<int>();
+            for (int i = 6; i <= 40; i++)
+                requiredBlockHeights.Add(i);
+
+            requiredBlockHeights = requiredBlockHeights.OrderBy(a => rnd.Next()).ToList();
```
dangershony: My only comment here would be why use random, for testing we want to explicitly control every part of the flow.

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Concurrent;
+using NBitcoin.Protocol.Behaviors;
+using System.Threading;
+using Stratis.Bitcoin.Connection;
+
+namespace Stratis.Bitcoin.BlockPulling
+{
+    /// <summary>
+    /// Implements a strategy for a block puller that needs to download 
+    /// a set of blocks and from its connected peers.
```
dangershony: remove the "and"

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    public static class DownloadAssignmentStrategy
```
dangershony: I am not sure about naming the pattern in to the class, maybe a more suitable option is to include Puller like PullerDownloadAssignments or PullerDistributionAssignments.

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+            return res;
+        }
+
+        /// <summary>
+        /// Choose random index proportional to the score.
+        /// </summary>
+        /// <param name="scores">Array of scores.</param>
+        /// <param name="totalScore">Sum of the values in <paramref name="scores"/>.</param>
+        /// <returns>Random index to <paramref name="scores"/> array - i.e. a number from 0 to scores.Length - 1.</returns>
+        private static int GetNodeIndex(int[] scores, int totalScore)
```
dangershony: its not completely clear to me the benefits of this method. (discussion for later)

dangershony:
Some minor comments but generally looks good.
You could consider making the strategy an instance of the puller (instead of using static methods).

Stratis.Bitcoin.Tests/BlockPulling/DownloadAssignmentStrategyTest.cs
Aprogiena: yes, will fix

Stratis.Bitcoin.Tests/BlockPulling/DownloadAssignmentStrategyTest.cs
Aprogiena: The idea was to use randomization to avoid testing a single case, which could possibly be designed poorly and thus could lead to a false positive test. 

On the other hand, one does want to prevent situation in which randomization causes some runs of the test to provide different results than others. This is why I went for (hopefully) large enough sample so that chance of passing the test because of "good" order would be minimal. 

If this seems insufficient, we can certainly increase the number of blocks to virtually any reasonable number (i.e. going 100x current counts would do no harm).

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
Aprogiena: yes, will fix

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
Aprogiena: OK, I'll change it to PullerDownloadAssignments. 

In my view it is a strategy by its nature, regardless of whether we support other strategies to be implemented or not (in long term, everything can be changed to support another strategy).

Stratis.Bitcoin/BlockPulling/DownloadAssignmentStrategy.cs
Aprogiena: This is just preserving the original logic used before. There is some quality valuation of each peer and this method prefers giving more work to 

Maybe you ask about the implementation of this method - that I did not touch - I just checked that it does what it should and I don't quickly see a simpler way to do the same.

Aprogiena:
I will provide the requested changes soon. As for not using static methods - my goal here was to fix a bug. Should we think it is beneficial to support different strategies in the future, I would separate that task from this. Or is there any other reason why would prefer not to use a static class?

dangershony:
Yeah indeed a separate task but only if we find its necessary to break away form statics

PR #239, opened 2017-07-16
Feature/xmldoc stratis.bitcoin.blockpulling

Aprogiena:
XML documentation for Stratis.Bitcoin.BlockPulling folder sources.

This includes code style fixes and also minor code polishing (typos in names, improving clarity of the code ...).


bokobza:
Hi mate, thank yo so much for doing that, I guess I'll learn how it works from reading your comments!
Are there any code changes at all?

Aprogiena:
There are minor changes - e.g. minHight -> minHeight, or 75 -> MaxQualityScore / 2 - stuff like that. There should be no changes that actually do something differently than the original code. That comes in the other PR - https://github.com/stratisproject/StratisBitcoinFullNode/pull/238.

However, please do not consider these comments as "good enough". In this case, the discussion with the author of the code made it possible to make the comments quite good, especially the overall picture comment. However, other comments are just result of my reverse engineering of the code and might not reflect the reality. So it is important here that you suggest a change if you spot any comment that does not really fit, and even more importantly - if you know something about some part of the code, to expand the (short) comment, so that more information is provided. For example, writing more is needed especially in case of AskBlocks - see TODO there. But even comments without TODO could be improved - expanded.

And of course - if you write a new code, please try to jump on the comment train. I'm aiming for "100 % comment everything policy".

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -10,299 +10,400 @@
...
+    /// Each '-' represents a block and letters 'A' to 'E' represent blocks with important positions in the chain from 
+    /// the puller's perspective. The puller can be understood as a producer of the blocks (by requesting them from the peers 
+    /// and downloading them) for the component that uses the puller that consumes the blocks (e.g. validating them).
+    /// </para>
+    /// <para>
+    /// A is a position of a block that we call location. Blocks in front of A were already downloaded and consumed.
+    /// </para>
+    /// <para>
+    /// B is a position of a block A + MinimumLookahead, and E is a position of block A + MaximumLookahead. 
+    /// Blocks between B and E are the blocks that the puller is currently interested in. Blocks after E are currently
```
NicolasDorier: B and E are here only to make sure that ActualLookAhead has a sane value.

[1,  1 + ActualLookahead] is the currently downloading blocks

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -10,299 +10,400 @@
...
+    /// <para>
+    /// D is a position of a block A + ActualLookahead. ActualLookahead is a number of blocks that the puller wants 
+    /// to be downloading simultaneously. If there is a gap between C and D it means that the puller wants to start 
+    /// downloading these blocks.
+    /// </para>
+    /// <para>
+    /// Blocks between D and E are currently those that the puller does not want to be downloading right now, 
+    /// but should the ActualLookahead be adjusted, they can be requested in the near future.
+    /// </para>
+    /// </remarks>
```
NicolasDorier: I think we should simplify the explanation.

```
-------------A---------B--------

A is the location of the BlockPuller.
B is the lookaheadLocation.

Interval [A, B] is what is currently downloaded.

ActualLookahead is a number of block count that we wish to download at the same time, it varies between Minimum and MaximumLookahead depending on the consumer's speed and node download speed.

Calling AskBlocks() increases lookaheadLocation by ActualLookahead.

The block puller make sure that if  

lookaheadLocation - A < ActualLookahead then AskBlocks() get called.
```

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -10,299 +10,400 @@
...
+        {
+            this.downloadedCounts.Add(this.DownloadedBlocks.Count);
+            if (this.lookaheadLocation == null)
+            {
+                // Calling twice is intentional here.
+                // lookaheadLocation is null only during initialization
+                // or when reorganisation happens. Calling this twice will 
+                // make sure the initial work of the puller is away from 
+                // the lower boundary.
+                AskBlocks();
```
NicolasDorier: LookaheadBlockPuller makes sure that:

```
if B - A <= ActualLookahead then AskBlocks() get called.
```

At initialization B == A so we need to call AskBlocks() two times.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: It may seem that my explanation is complex, but it comes with a lot of additional details that your version is missing. I think those can be actually useful to anyone coming to the code later. 

I would thus suggest to modify the explanation in a way to expand it even more - to use what you wrote here, but not to lose the additional information. I will make another commit that will try to do that.

But in general, I would like to see more rather then less.

Aprogiena:
@NicolasDorier I have merged your additional comments to the code. I did not make it shorter, rather the opposite to give the consumer of the comment as much info as possible.

bokobza:
@Aprogiena  I'll have a look today

PR #248, opened 2017-07-24
Bugfix/issue 102

### LINKED ISSUE #102: BlockPuller bug: Node_MessageReceived
Node_MessageReceived threw this exception:

image

Locals:
image

Console:
image

Please let me know if there's any more information I can provide to help troubleshoot this.

Aprogiena:
This is a huge update of block puller code, which is aimed to fix #102 as well as prepare grounds for fixing other related issues.

This also adds loggers to block pulling code, which were not there before. So newly it is possible to add logs to this code.

It was uneasy to make somehow incremental changes, because the fundamental change was done - separating BlockPullerBehavior to separate file as non-data inner classes tend to produce poor design especially in multithreading environment. This PR does not preserve the original logic of the block puller in many places, but the overall approach was not changed.

I've introduced a lock protecting the integrity of the data, which should prevent the problems of #102. However, at the same time I've identified problem with the QualityScore calculate, which could lead (and in my tests it leads) to a problem that many or all peers end with very low score, which leads to work being reassigned very often, which leads to blocks being delivered by peers that no longer are assigned to download them (the task was assigned to other peer). This leads to great waste of resources and slowing down the download process. 

This is a big change and should be tested very carefully and many times again and again as there are no unit tests for this AFAIK.


Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -18,6 +19,43 @@ namespace Stratis.Bitcoin.BlockPulling
...
+    /// </para>
+    /// </para>
+    /// <para>
+    /// For a block to be considered as currently (or soon to be) being downloaded, its hash has to be 
+    /// either in <see cref="map"/> or <see cref="pendingInventoryVectors"/>.
+    /// </para>
+    /// <para>
+    /// When the puller is about to request blocks from the peers, it selects which of its peers will 
+    /// be asked to provide which blocks. These assignments of block downloading tasks is kept inside 
+    /// <see cref="map"/>. Unsatified requests go to <see cref="pendingInventoryVectors"/>, which happens
```
dangershony: "Unsatisfied"

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -26,14 +64,41 @@ public abstract class BlockPuller : IBlockPuller
...
+        protected readonly ILogger logger;
+
+        /// <summary>Lock protecting access to <see cref="map"/>, <see cref="pendingInventoryVectors"/>, <see cref="downloadedBlocks"/>, and <see cref="pendingDownloads"/></summary>
+        private readonly object lockObject = new object();
+
+        /// <summary>
+        /// List of relations to peer nodes mapped by block header hashes that the peers are requested to provide.
+        /// </summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly Dictionary<uint256, BlockPullerBehavior> map;
```
dangershony: I want a better name to "map" (maybe blockToPeerMappings or blocksToPeers)
Also I would suggest this text
"A list of mappings between a peer and block headers, the header represents the block that peer is asked to download"

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -319,36 +208,46 @@ private void AssignPendingVectors()
...
+            if (this.pendingInventoryVectors.Count == 0)
+                return;
 
+            int minHeight = int.MaxValue;
             Dictionary<int, InventoryVector> vectors = new Dictionary<int, InventoryVector>();
-            var minHeight = int.MaxValue;
-            foreach (InventoryVector vector in inventoryVectors.ToArray())
+
+            uint256 blockHash;
+            while (true)
```
dangershony: Hmm this is interesting, you can avoid the loop on a lock in this case, as the result will be almost the same if another threads comes in.  

In this code two (or more) threads can deplete the queue. and call the distributor twice.
 
You can pull the lock out of the loop and populate a local list to deplete the queue , if another thread will come it only waits for the deplete code and returns (as the queue will be empty)

Then you don't have to lock on every iteration.

dangershony:
```
                var items = new List<uint256>();
                lock (this.lockObject)
                {
                    if (this.pendingInventoryVectors.Count == 0)
                        break;

                    blockHash = this.pendingInventoryVectors.Dequeue();
                    while (blockHash != null)
                    {
                        items.Add(blockHash);
                        blockHash = this.pendingInventoryVectors.Dequeue();
                    }
                }

                foreach (var item in items)
                {
                    // bla
                }
```

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -439,7 +347,7 @@ private void DistributeDownload(Dictionary<int, InventoryVector> vectors, BlockP
                 foreach (int blockHeight in blockHeightsToDownload)
                 {
                     InventoryVector inventoryVector = vectors[blockHeight];
-                    if (this.map.TryAdd(inventoryVector.Hash, peerBehavior))
+                    if (AssignDownloadTaskToPeer(peerBehavior, inventoryVector.Hash))
```
dangershony: this.AssignDownloadTaskToPeer

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: tx, will fix

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: yes, i used map as it was in the original code, but you are right it should be renamed, and i'll change the comment as well

i would avoid "mappings" suffix and i tend to call this "download tasks" in the comments, so i would go for "blockDownloadTasks" or "assignedBlockTasks", maybe the later is better (for me), 

the comment could be 

"Hashes of blocks to be downloaded mapped by the peers that the download tasks are assigned to."

what do you think?

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: of course, i should have done it like that, i think i just did one level of refactoring and then forgot to think about next level, tx for that! it needs more memory, but it is well justified use and the number of pending vectors will never be that large for another copy to matter in memory

and i think i don't even need to dequeue, i should just .ToArray() and .Clear that

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: erm, not sure about this one, i thought we use this. on fields and properties of the current instance, but this is method and MSVS gives me the hint that this. is missing (according to the rules we have on the project), but not on methods

i think calling methods should be without this.
right?

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -26,14 +64,41 @@ public abstract class BlockPuller : IBlockPuller
...
+                lock (this.lockObject)
+                {
+                    return this.downloadedBlocks.Count;
+                }
+            }
+        }
+
+        /// <summary>Hash set holding set of block header hashes that are being downloaded, mapped by connected peer's behavior.</summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly Dictionary<BlockPullerBehavior, HashSet<uint256>> pendingDownloads = new Dictionary<BlockPullerBehavior, HashSet<uint256>>();
```
dangershony: Also this can be renamed to a more suitable purpose peersToBlocks

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -448,5 +356,279 @@ private void DistributeDownload(Dictionary<int, InventoryVector> vectors, BlockP
...
+        }
+
+        /// <summary>
+        /// Assigns a download task to a specific peer.
+        /// </summary>
+        /// <param name="peer">Peer to be assigned the new task.</param>
+        /// <param name="blockHash">Hash of the block to download from <paramref name="peer"/>.</param>
+        /// <returns><c>true</c> if the block was assigned to the peer, <c>false</c> in case the block has already been assigned to someone.</returns>
+        /// <remarks>The caller of this method is responsible for holding <see cref="lockObject"/>.</remarks>
+        private bool AssignDownloadTaskToPeerLocked(BlockPullerBehavior peer, uint256 blockHash)
```
dangershony: Push this to its parent caller, way more simple then to have a method inside a lock. (and avoid the Locked word)

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: depends on what name we end up for the "map"
if it is "assignedBlockTasks" as i suggest, it would be "peersTasks" here
but if we go with yours "blocksToPeers" then yes, it would be "peersToBlocks" here, i just don't know how i feel about that "X-To-Y" naming for dictionaries

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: oh yeah, i know what you mean, this is the result of refactoring that i did have another method calling the locked version and i removed it later (similarly for ReleaseDownloadTaskAssignmentLocked, but in that case there really are two methods to call it, so that one needs to stay)

so yes, this one i fix, it is not needed to be separated, tx

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -448,5 +356,279 @@ private void DistributeDownload(Dictionary<int, InventoryVector> vectors, BlockP
...
+        {
+            bool res = false;
+
+            if (this.map.TryAdd(blockHash, peer))
+            {
+                HashSet<uint256> peerPendingDownloads;
```
dangershony: this code is repetitive and can go in an extension or a separate method
```
                    HashSet<uint256> peerPendingDownloads;
                    if (!this.pendingDownloads.TryGetValue(peer, out peerPendingDownloads))
                    {
                        peerPendingDownloads = new HashSet<uint256>();
                        this.pendingDownloads.Add(peer, peerPendingDownloads);
                    }

                    peerPendingDownloads.Add(blockHash);

```

Aprogiena: yes, will fix

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        /// <summary>
+        /// Connects the puller to the node and the chain so that the puller can start its work.
+        /// </summary>
+        protected override void AttachCore()
+        {
+            this.AttachedNode.MessageReceived += Node_MessageReceived;
+            this.ChainHeadersBehavior = this.AttachedNode.Behaviors.Find<ChainHeadersBehavior>();
+            AssignPendingVector();
```
dangershony: this.AssignPendingVector()

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Disconnects the puller from the node and cancels pending operations and download tasks.
+        /// </summary>
+        /// <remarks>
+        /// TODO: https://github.com/stratisproject/StratisBitcoinFullNode/issues/246
+        /// </remarks>
+        protected override void DetachCore()
+        {
+            this.cancellationToken.Cancel();
+            this.AttachedNode.MessageReceived -= Node_MessageReceived;
+            ReleaseAll();
```
dangershony: this. ReleaseAll();

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -403,14 +407,21 @@ private Block NextBlockCore(CancellationToken cancellationToken)
                     }
                     else
                     {
+                        // TODO: Race condition here - as soon as IsDownloading returns false here, 
+                        // another thread can change that (consider BlockPullerBehavior.Node_MessageReceived, 
+                        // just before the newly downloaded block is pushed, the above check for DownloadedBlocks 
+                        // will fail - block not pushed yet - but then IsDownloading here returns false), which 
+                        // will cause this thread to call AskBlocks for a block that was already downloaded. 
+                        // https://github.com/stratisproject/StratisBitcoinFullNode/issues/244
```
dangershony: IsDownloading is only called form one thread however AskBlocks I am not sure, but if we ask a block twice the map dictionary will throw.

dangershony:
I get the rational of adding a lockobject, synchronizing access to 3 separate dictionaries is complicated and just asking for concurrency bugs.

Good job.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -448,5 +356,279 @@ private void DistributeDownload(Dictionary<int, InventoryVector> vectors, BlockP
...
+        /// <exception cref="InvalidOperationException">Thrown in case of data inconsistency between synchronized structures, which should never happen.</exception>
+        internal void ReleaseAllPeerDownloadTaskAssignments(BlockPullerBehavior peer)
+        {
+            lock (this.lockObject)
+            {
+                HashSet<uint256> peerPendingDownloads;
+                if (this.pendingDownloads.TryGetValue(peer, out peerPendingDownloads))
+                {
+                    // Make a fresh copy of items in peerPendingDownloads to avoid modification of the collection.
+                    foreach (uint256 blockHash in peerPendingDownloads.ToList())
```
Aprogiena: I don't understand here. ReleaseAllPeerDownloadTaskAssignments needs its own copy because ReleaseDownloadTaskAssignmentLocked removes from peerPendingDownloads.

And what is Confines?

What I could do is not to use ReleaseDownloadTaskAssignmentLocked and reimplement it without removing from pending downloads, but I decided for this instead as the logic of the code was better from my point of view. Or I could use parameter that would determine whether to remove or not, but again, it would complicated the code. So at the end I asked myself what is the cost of having it like this and I concluded that the cost is OK (not trivial or negligible, but given the frequency of this method is called, very much OK0.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: same as above - I can't find this in the code I've seen so far - e.g. LookaheadBlockPuller has NextBlock which calls NextBlockCore without this. and the code style speaks about arguments. So I think to be consistent, these should be without this. Or am I missing something?

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: ditto

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: I've separated this to its own issue, so it should not be considered in this PR, but the case is that one thread executes TryRemoveDownloadedBlock and that returns false (block not ready) and then another thread delivers the block (block ready and no longer is downloading), then the first one executes IsDownloading and it returns false and then we ask for the block that we already have (but is not in the map anymore).

The second part - asking for block which is in map won't throw as this is currently job of AssignDownloadTaskToPeer which uses TryAdd and ignores the request if the block is there in the map.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: Sure your suggestion is fine

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: assignedBlockTasks fine with me

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: Ah i meant .Contains()

Yeah it removes from a new fresh copy right? just to satisfy the condition of .Remove() or do you actually need to remove from the list in which case the ToList is wrong as it creates a new copy.

Aprogiena:
There is one more bug I am rarely getting, but can't see it, but will fix before submitting latest changes.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
dangershony: Generally anything that is part of the instance and not local to the method uses this.
This is my approach at least.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
dangershony: Ah right yeah so maybe we need more defensive approach.

Seperate tasks yes

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: No, it removes from the original, the copy is there just to have access to all those guys that I want to remove. I have a shared list there, right? Now I want to remove all its items, I could do .Clear() on this peerPendingDownloads, but I also want all those items to be removed from assignedBlockTasks. So what I could do is to have foreach on original (without copy) and just assignedBlockTasks.Remove(item) and then peerPendingDownloads.Clear(). So I would have a second implementation of ReleaseDownloadTaskAssignmentLocked for this case, which is what I wanted to avoid, also because ReleaseDownloadTaskAssignmentLocked checks for me whether data are in sync (this is kind of replication of the original exception raise that "should never occur" that was reported in #102).

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: OK, that is new to me, I've been doing that XML doc PRs where I also fix code style everywhere, but this was not part of the code style (and as I say, MSVS does mark only those I mentioned). 

So this has to be a separate issue because it basically means we need to update code-style and then enforce it everywhere. Not that I would be happy about this particular one though, but OK. Still, I would consider that a separate issue and we should redesign the code-style policy heavily.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: yeah indeed ToArray and Clear yes

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: Ah yeah I see my mistake peerPendingDownloads is passed untouched and ReleaseDownloadTaskAssignmentLocked will remove from it.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
dangershony: Ah ok I see yeah the code style does specify arguments, this is something I wish to fix.
So leave it as is for now.
I think also methods should use this. the point is to distinguish static to instance methods as well and once you get used to that it makes reading the code much easier.

dangershony:
One more thing I would suggest is writing unit tests on the puller, we would then be more confident with changes, but that may go as a separate PR.

Aprogiena:
Yes, tests for the puller would be nice, that was actually where I started more than week ago, but I was not able to make the setup because some classes for that don't support testing yet, or maybe my experience with the code base is not on that level yet.

Anyway, I did analyze the bug and I don't quite understand what happened there, but I added a sanity check that will prevent the crash and won't damage anything, so the bug should be fixed by now.

I've made all requested changes except for this. prefixing for methods as explained.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: OK, but separate issue pls :-)

Stratis.Bitcoin.Tests/BlockPulling/PullerDownloadAssignmentsTest.cs
```diff
@@ -107,5 +107,67 @@ public void AssignBlocksToPeersWithNodesWithDifferentChainsCorrectlyDistributesD
...
+                    maxPeerChainLength = Math.Max(maxPeerChainLength, peerInfo.ChainHeight);
+                }
+
+                // Use the assignment strategy to assign tasks to peers.
+                Dictionary<PullerDownloadAssignments.PeerInformation, List<int>> assignments = PullerDownloadAssignments.AssignBlocksToPeers(requiredBlockHeights, availablePeersInformation);
+
+                // Check the assignment is valid per our requirements.
+                int tasksAssigned = 0;
+                Assert.Equal(availablePeersInformation.Count, assignments.Count);
+                foreach (KeyValuePair<PullerDownloadAssignments.PeerInformation, List<int>> kvp in assignments)
```
dangershony: I think its way more readable to have here 
```foreach  (var assignment in assignments)```

Aprogiena: I would be OK with that (although I prefer the full form), but using the var here would violate the current code style policy (actually that part of the current code style I like very much, because I don't consider var as being a good friend almost in any case), so please remember that for code style policy redesign if you want this. Maybe we should open a wide discussion somewhere (issue in github) where everyone could express their preference and after some time we could converge to a new code style policy.

PR #249, opened 2017-07-24
add rpc method getblockheader

mikedennis:
Add the rpc method "getblockheader" with supporting integration unit test from NBitcoin.  Test CanGetBlockFromRPC() assertion is currently failing but not sure what I'm doing wrong - need another set of eyes to see what I missing.

### DIFF
--- a/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
+++ b/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
@@ -92,7 +92,7 @@ namespace Stratis.Bitcoin.IntegrationTests
                 Assert.Equal(expectedHeader.HashPrevBlock, actualHeader.HashPrevBlock);
                 Assert.Equal(expectedHeader.HashMerkleRoot, actualHeader.HashMerkleRoot);
                 Assert.Equal(expectedHeader.Time, actualHeader.Time);
-                Assert.Equal(expectedHeader.Bits.ToCompact(), actualHeader.Bits.ToCompact());
+                Assert.Equal(expectedHeader.Bits, actualHeader.Bits);
                 Assert.Equal(expectedHeader.Nonce, actualHeader.Nonce);
 
                 // Assert header hash matches genesis hash
--- a/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
+++ b/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
@@ -55,7 +55,7 @@ namespace Stratis.Bitcoin.IntegrationTests
         }
 
         /// <summary>
-        /// Tests RPC getbestblockhash
+        /// Tests RPC getbestblockhash.
         /// </summary>
         [Fact]
         public void CanGetGenesisFromRPC()
@@ -72,7 +72,7 @@ namespace Stratis.Bitcoin.IntegrationTests
         }
 
         /// <summary>
-        /// Tests RPC getblockheader
+        /// Tests RPC getblockheader.
         /// </summary>
         [Fact]
         public void CanGetBlockHeaderFromRPC()
--- a/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
+++ b/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
@@ -54,5 +54,32 @@ namespace Stratis.Bitcoin.IntegrationTests
             }
         }
 
+        [Fact]
+        public void CanGetGenesisFromRPC()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create())
+            {
+                RPCClient rpc = builder.CreateStratisNode().CreateRPCClient();
+                builder.StartAll();
+                RPCResponse response = rpc.SendCommand(RPCOperations.getblockhash, 0);
+                string actualGenesis = (string)response.Result;
+                Assert.Equal(Network.RegTest.GetGenesis().GetHash().ToString(), actualGenesis);
+                Assert.Equal(Network.RegTest.GetGenesis().GetHash(), rpc.GetBestBlockHash());
+            }
+        }
+
+        [Fact]
+        public void CanGetBlockFromRPC()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create())
+            {
+                RPCClient rpc = builder.CreateStratisNode().CreateRPCClient();
+                builder.StartAll();
+                BlockHeader response = rpc.GetBlockHeader(0);
+
+                // TODO: This assertion is currently failing
+                Assert.Equal(Network.RegTest.GenesisHash, response.GetHash());
+            }
+        }
     }
 }
--- a/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
+++ b/Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
@@ -54,6 +54,9 @@ namespace Stratis.Bitcoin.IntegrationTests
             }
         }
 
+        /// <summary>
+        /// Tests RPC getbestblockhash
+        /// </summary>
         [Fact]
         public void CanGetGenesisFromRPC()
         {
@@ -68,17 +71,32 @@ namespace Stratis.Bitcoin.IntegrationTests
             }
         }
 
+        /// <summary>
+        /// Tests RPC getblockheader
+        /// </summary>
         [Fact]
-        public void CanGetBlockFromRPC()
+        public void CanGetBlockHeaderFromRPC()
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                RPCClient rpc = builder.CreateStratisNode().CreateRPCClient();
+                CoreNode node = builder.CreateStratisNode();
+                RPCClient rpc = node.CreateRPCClient();
                 builder.StartAll();
-                BlockHeader response = rpc.GetBlockHeader(0);
 
-                // TODO: This assertion is currently failing
-                Assert.Equal(Network.RegTest.GenesisHash, response.GetHash());
+                uint256 hash = rpc.GetBlockHash(0);
+                BlockHeader expectedHeader = node.FullNode.Chain?.GetBlock(hash)?.Header;
+                BlockHeader actualHeader = rpc.GetBlockHeader(0);
+
+                // Assert block header fields match
+                Assert.Equal(expectedHeader.Version, actualHeader.Version);
+                Assert.Equal(expectedHeader.HashPrevBlock, actualHeader.HashPrevBlock);
+                Assert.Equal(expectedHeader.HashMerkleRoot, actualHeader.HashMerkleRoot);
+                Assert.Equal(expectedHeader.Time, actualHeader.Time);
+                Assert.Equal(expectedHeader.Bits.ToCompact(), actualHeader.Bits.ToCompact());
+                Assert.Equal(expectedHeader.Nonce, actualHeader.Nonce);
+
+                // Assert header hash matches genesis hash
+                Assert.Equal(Network.RegTest.GenesisHash, actualHeader.GetHash());
             }
         }
     }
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Check bitcoin code directly, this is more reliable than the doc.

https://github.com/bitcoin/bitcoin/blob/041dad94b047313a17edb0324742cf80cfd550f1/src/rpc/blockchain.cpp#L629

dangershony:
From the core reference just use the time property is fine
https://github.com/bitcoin/bitcoin/blob/041dad94b047313a17edb0324742cf80cfd550f1/src/rpc/blockchain.cpp#L91

dangershony:
@mikedennis whats the progress on this?

bokobza:
Hi @mikedennis, is this ready for review? apart from the median time, there are a few items in TODO.

mikedennis:
@dangershony @bokobza Still needs some work and clean up but I still haven't figured out why the unit test is not passing. I get a different hash back than the genesis. If anyone knows what I'm doing wrong and can point me in the right direction, I'll finish this baby off!

bokobza:
You're not getting the hash you're expecting because the hash is calculated from the BlockHeader's "value" itself.
If you look at Block.cs line 177 you'll see
`h = Hashes.Hash256(this.ToBytes());`
`this` being a BlockHeader.
What I suggest is for you to change this test to Assert all the fields in the BlockHeader, add what's missing in the BlockHeaderModel and leave the hash for last.
Does that make sense?

mikedennis:
@bokobza Ah yes that makes sense. It's because I haven't added all the fields yet. Sounds good! Thanks for the help.

bokobza:
Nice one, I'll have a look soon.

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
+﻿using NBitcoin;
+using NBitcoin.DataEncoders;
+
+namespace Stratis.Bitcoin.Features.RPC.Models
+{
+#pragma warning disable IDE1006 // Naming Styles (ignore lowercase)
```
bokobza: please remove this and look at my comment below on how to deal with these.

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+                this.merkleroot = blockHeader.HashMerkleRoot.ToString();
+                this.time = blockHeader.Time;
+                byte[] bytes = this.GetBytes(blockHeader.Bits.ToCompact());
+                string encodedBytes = Encoders.Hex.EncodeData(bytes);
+                this.bits = encodedBytes;
+                this.nonce = (int)blockHeader.Nonce;
+            }
+        }
+
+        /// <summary>
```
bokobza: I know you're basing yourself on what was done previously but here's a better way:
please call the property **V**ersion (Pascal casing) and use the following for it to be serialized/deserialized with a lower-case 
 [JsonProperty(PropertyName = "version")]

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public uint version { get; set; }
+
+        /// <summary>
+        /// The merkle root for this block encoded as hex in RPC byte order
+        /// </summary>
+        public string merkleroot { get; set; }
+
+        /// <summary>
+        /// The nonce thich was successful at turning this particular block
```
bokobza: which*

Stratis.Bitcoin/Features/RPC/Controllers/FullNodeController.cs
```diff
@@ -110,6 +117,28 @@ public GetInfoModel GetInfo()
             return model;
         }
 
+        /// <summary>
+        /// Implements getblockheader RPC call
+        /// </summary>
+        /// <param name="hash">hash of block</param>
```
bokobza: Please use an upper case at the beginning of XML documentation phrases, and a dot in the end.

Stratis.Bitcoin/Features/RPC/Controllers/FullNodeController.cs
```diff
@@ -110,6 +117,28 @@ public GetInfoModel GetInfo()
...
+        /// <param name="isJsonFormat">indicates whether to provide data in Json or binary format</param>
+        /// <returns>The block header rpc format</returns>
+        [ActionName("getblockheader")]
+        public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
+        {
+            Guard.NotNull(this.Chain, nameof(this.Chain));
```
bokobza: The Guard method is only really used to check parameters.
So here you could have Guard.NotEmpty(hash, nameof(hash));
To check that the chain is not null, a simple if(this.chain == null) is sufficient (or != null depending on how you want to do this).

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <see cref="https://bitcoin.org/en/developer-reference#getblockheader"/>
+    /// </summary>
+    public class BlockHeaderModel
+    {
+        /// <summary>
+        /// Constructs a RPC BlockHeaderModel from a block header object
+        /// </summary>
+        /// <param name="blockHeader">the block header</param>
+        public BlockHeaderModel(BlockHeader blockHeader)
+        {
```
bokobza: what happens if the block header is null here?

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+        public string merkleroot { get; set; }
+
+        /// <summary>
+        /// The nonce thich was successful at turning this particular block
+        /// into one that could be added to the best block chain
+        /// </summary>
+        public int nonce { get; set; }
+
+        /// <summary>
+        /// The target threshhold this blocks header had to pass
```
bokobza: - 1 'h' in threshold
- block's

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <param name="blockHeader">the block header</param>
+        public BlockHeaderModel(BlockHeader blockHeader)
+        {           
+            if (blockHeader != null)
+            {
+                this.version = (uint)blockHeader.Version;
+                this.previousblockhash = blockHeader.HashPrevBlock.ToString();
+                this.merkleroot = blockHeader.HashMerkleRoot.ToString();
+                this.time = blockHeader.Time;
+                byte[] bytes = this.GetBytes(blockHeader.Bits.ToCompact());
```
bokobza: Here it seems you're changing Bits to uint and then back to bytes. Can't you just do Bits.ToBytes()?

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,84 @@
...
+                string encodedBytes = Encoders.Hex.EncodeData(bytes);
```
bokobza: Are you sure this is according to the bitcoin spec? I'm not sure myself, not a tricky question :-)
https://github.com/bitcoin/bitcoin/blob/041dad94b047313a17edb0324742cf80cfd550f1/src/rpc/blockchain.cpp#L650
https://github.com/bitcoin/bitcoin/blob/041dad94b047313a17edb0324742cf80cfd550f1/src/rpc/blockchain.cpp#L94

mikedennis:
@bokobza Thanks for the feedback. I think I've got the issues sorted out in my latest commit. 

Regarding the questions around BlockHeader.Bits serialization. Another alternative would to make this https://github.com/MetacoSA/NBitcoin/blob/9843ab333f93394d70287bec3f3754610abe7897/NBitcoin/Target.cs#L28 NBitcoin method for GetBytes() public and then I could use it directly.  It was a very simple method so I just copied it into the Model. There is definitely some funkiness in the way Bitcoin is serializing this but I believe I am now doing it correctly as the hashes do match up now. Let me know if you see a better way or want me to modify access to the NBitcoin method  instead.

dangershony:
Can to you not just do a normal set
this.Bits = blockHeader.Bits

Stratis.Bitcoin.IntegrationTests/RPCTEsts.cs
```diff
@@ -54,5 +54,50 @@ public void CheckRPCFailures()
...
+                uint256 hash = rpc.GetBlockHash(0);
+                BlockHeader expectedHeader = node.FullNode.Chain?.GetBlock(hash)?.Header;
+                BlockHeader actualHeader = rpc.GetBlockHeader(0);
+
+                // Assert block header fields match
+                Assert.Equal(expectedHeader.Version, actualHeader.Version);
+                Assert.Equal(expectedHeader.HashPrevBlock, actualHeader.HashPrevBlock);
+                Assert.Equal(expectedHeader.HashMerkleRoot, actualHeader.HashMerkleRoot);
+                Assert.Equal(expectedHeader.Time, actualHeader.Time);
+                Assert.Equal(expectedHeader.Bits.ToCompact(), actualHeader.Bits.ToCompact());
```
bokobza: Are expectedHeader.Bits != actualHeader.Bits? I mean do you need to do .ToCompact to verify equality?

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        /// <summary>
+        /// Constructs a RPC BlockHeaderModel from a block header object.
+        /// </summary>
+        /// <param name="blockHeader">The block header.</param>
+        public BlockHeaderModel(BlockHeader blockHeader)
+        {
+            Guard.NotNull(blockHeader, nameof(blockHeader));
+
+            if (blockHeader != null)
```
bokobza: you can remove this check now that you have the Guard :-)

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,91 @@
...
+                this.Bits = encodedBytes;
+                this.Nonce = (int)blockHeader.Nonce;
+            }
+        }
+
+        /// <summary>
+        /// The blocks version number.
+        /// </summary>
+        [JsonProperty(PropertyName = "version")]
+        public uint Version { get; private set; }
```
bokobza: Lovely, thank you.

Stratis.Bitcoin/Features/RPC/Models/BlockHeaderModel.cs
```diff
@@ -0,0 +1,91 @@
...
+        public uint Time { get; private set; }
+
+        /// <summary>
+        /// Convert compact of miner challenge to byte format,
+        /// serialized for transmission via RPC.
+        /// </summary>
+        /// <param name="compact">Compact representation of challenge.</param>
+        /// <returns>Byte representation of challenge.</returns>
+        /// <seealso cref="Target"/>
+        private byte[] GetBytes(uint compact)
```
bokobza: Ok let's use this method 👍

mikedennis:
@dangershony I wish it was that simple! The Bits.ToString() gives you a uint256 encoded string but the Target constructor needs a "compact" encoded string. See here https://github.com/MetacoSA/NBitcoin/blob/9843ab333f93394d70287bec3f3754610abe7897/NBitcoin/Target.cs#L43

NicolasDorier:
Do I need to change something in NBitcoin to make it more easy?

bokobza:
Lovely, thank you @mikedennis.
Thanks @NicolasDorier I think we're OK for now :-)

PR #251, opened 2017-07-26
Feature/commenting policy

Aprogiena:
First version of code commenting policy that it would be nice to have for full node project. If it turns out to be good, we can then expand to other Stratis projects.

The main focus is on XML documentation comments. Inline comments are almost unrestricted - we only prescribe double slash comments to be used there and forbid trivial comments without any added value.

This is very much open for discussion and tweaking.


bokobza:
This is great, thanks @Aprogiena.
The only comment I have about the **form** is that I think the file should be called comments-policy (even if maybe the intention was to have future policies like 'policy-code-style' or a 'policy-unit-testing' stack up nicely in alphabetical order).
I'll get into the form (grammar, punctuation etc..) when the **content** is locked-in.

So about the content:
Overall, I agree with the intent as well as with the content but I'm fearful of contributors seeing this and deciding not to join because of it. Comments are important and I try to always add comments to class, methods and properties and my code is full of comments where it makes sense. 

**Motivation:**
I think detailing to readers the reasons behind this policy is beneficiary, although I think it should be said in fewer words if we want people to actually read it.

**Rules:**
For me, the only real hard rule is to "use comments on every single top-level entity using the construct you get when you type /// in Visual Studio."
However, whilst I agree that a bit more detail goes a long way, fewer rules will help avoiding people being discouraged. If you can, try to extract some of the soft rules as "general guidelines".

- USE and TRIVIAL seems to be the same rule. Can these be merged into USE: use //  to explain complicated or counter-intuitive segments of the code.
- INTENT and ADD-VALUE: I don't think these should be a rule to be enforced. I see this more as a general guideline to follow.
- ENUMS: comment all values unless super trivial (Monday, Tuesday for example).
- SENTENCES: don't "try", jfdi. And what about explanatory comments on pieces of code? Capital letter there as well? 
- Another rule would be to leave the code better commented that you found it, if possible.

Apart from that I think "Calculates N-th element of the Fibonacci sequence using iteration method." actually leaks implementation details ;-)

I'm going to add stylecop, which is a tool that can scan your code for violations of rules that you set. It won't enforce anything, but if we set some rules on it, users can easily spot where the code needs improvement.

Aprogiena:
Tx for review, as this needs wider consensus, I'll wait a little more to hear from others before changing anything based on your comment. 

enums - yes
sentences - yes and yes

Otherwise - you mention "general guidelines", that is exactly what I (tried to) wrote there - that soft rules are guidelines (i.e. be aware of it, but use your brain), hard rules are something that we insist on. So I'm now not sure if you don't like soft rules in general (and you'd rather see separation of them to a section called guidelines) or if you missed that the intention was just that. The reason why I went for this form - hard and soft rules - is for each rule to have its name. Then reviewer can just add "violates CMTP ADD-VALUE" and it is clear, whereas if it was just a paragraph with unmarked sentences as "general guidelines", the reviewer would need to say "violates CMTP guidelines as it does not add any value, bla bla bla". So the idea was to have an easy point of reference.

intent - is definitely soft - i.e. guideline - so yes
add-value - for me this is almost hard, but just because some uninteresting cases, i have it as soft - so again yes, it is just a guideline

motivation - i can cut it to half, let's see what others say

"fearful of contributors seeing this and deciding not to join because of it" - No :) this I can't agree with. For me this is on the same level as a fear of contributors writing in PHP no joining us because we write in C#. I mean, the policy is intended to be reasonable, if it is not, we should fix it, but once it is reasonable, anyone not willing to comply is a potential troublemaker. Same with the code style - it is reasonable to expect everyone to comply and if there is someone who does not want to then what? Will we let them write random style code to create more mess? Also - if we are open about discussion of the policies even when they are in place, then whoever comes and does not like them can suggest improvements and we will listen and argue.

bokobza:
> Otherwise - you mention "general guidelines", that is exactly what I (tried to) wrote there - that soft rules are guidelines (i.e. be aware of it, but use your brain), hard rules are something that we insist on. So I'm now not sure if you don't like soft rules in general (and you'd rather see separation of them to a section called guidelines) or if you missed that the intention was just that. The reason why I went for this form - hard and soft rules - is for each rule to have its name. Then reviewer can just add "violates CMTP ADD-VALUE" and it is clear, whereas if it was just a paragraph with unmarked sentences as "general guidelines", the reviewer would need to say "violates CMTP guidelines as it does not add any value, bla bla bla". So the idea was to have an easy point of reference.

I'm trying to have fewer rules so that it is/looks less stringent. It's true that being able to reference rules by their name makes it easier. 

> "fearful of contributors seeing this and deciding not to join because of it" - No :) this I can't agree with. For me this is on the same level as a fear of contributors writing in PHP no joining us because we write in C#.
Don't agree. The more rules we have , the less easy it'll be for contributors to join us. It's on a completely different level from having to stick to C# syntactic rules when you write C#.

Now whether we're being **super stringent** or **not caring**, and whether we want to have **all the contributors in the world** or having just **the ones that share our vision** of how the code should look,  there's a middle ground. I think the sweet spot will be found when the list of rules looks a bit less intimidating.

Aprogiena:
Maybe some assurance that we are open for improving the policies might work as well. 

What I meant is that someone could come and say - "hey guys, you are doing this in C#, but NodeJS is cooler, let's do it in that ..." which is somewhat similar to me as if someone comes with "hey guys, you are having this comment policy, but i don't like comments, can i still work with you?".

At some point you might want to draw a limit. To be perfectly honest here, years ago I found this book from about programming blockchain in C#, I've read it in whole, it was so cool. Then I looked at NBitcoin and there were no comments - this stop me from exploring further as I knew "hey something is very wrong here". 

So the argument you are presenting does actually work both ways - some people may be scared of being forced to certain policy, but others may be scared that no such policy exists. Now which of those devs you want to prefer and which you are OK to be scared off?

But yes, no problem to spend more time trying to make fewer rules (i.e. join some of them to one etc).

bokobza:
I agree, a statement inviting contributions to the rule-set and about these rules being a work in progress is a good idea.
> So the argument you are presenting does actually work both ways - some people may be scared of being forced to certain policy, but others may be scared that no such policy exists. Now which of those devs you want to prefer and which you are OK to be scared off?
Definitely want the rules, and I wouldn't want devs to leave because our comments are rubbish or because we don't have a policy. :weary:

Let's see what other people think.

Documentation/policy-comments.md
```diff
@@ -0,0 +1,169 @@
...
+them. Soft rules are more like guidelines - try to use them where possible. The reviewer can ask you to update your code 
+if it violates a soft rule and it can be fixed easily.
+
+ * **XMLDOC** - Use triple slash (XML documentation) comments for classes, interfaces, methods, parameters, fields, properties, structures, enums, etc. These comments are required - even for trivial code. *(HARD)*
+   * **SUMMARY-1** - First paragraph in `<summary>` must be short, no more than 3 lines. If you find you need more than 2 lines, maybe you are not writing the summary anymore and you are already explaining details. *(HARD)*
+   * **SUMMARY-2** - `<summary>` should contain information useful for someone who is trying to understand the code. Implementation details belong to `<remarks>`. *(SOFT)*
+   * **METHODS** - Method comments must contain `<summary>`. If the method has input and/or output parameters `<param>` must be used for each of them. If the method returns anything else than `void` or `Task` (but not `Task<T>`),
+it must contain `<returns>`. The order of the tags is `<summary>`, `<param>`, `<returns>`, `<exception>`, `<remarks>`, `<example>`, `<seealso>`. *(HARD)*
```
bokobza: The thing is, if we let the user decide whether something is trivial or not, we're gonna get into endless discussions about what's trivial and what's not.
When we say: we document all parameters, there is no argument, everyone should just do it.
Apart from the time gained in not thinking whether or not to comment a param, the code would look more homogenous.

Aprogiena: `<param>` should be always documented. Yes, some times this is prone to triviality, but it again helps you thinking about what the param is for, are there any catches. Also helps with IntelliSense. So I would be strict with `<param>`.

Documentation/policy-comments.md
```diff
@@ -0,0 +1,169 @@
...
+   * **FORMAT** - Use `<para>`, `<see>`, `<seealso>`, `<c>`, `<code>`, `<list>` to make the comment structure better. *(SOFT)*
+   * **INHERITDOC** - Use `<inheritdoc />` if the comment already exists on lower level (base class, interface, virtual/abstract method). But feel free to add additional comments if the base comment is not sufficient. *(HARD)*
+ * **INLINE** - Use double slash (inline) comments for function body code, variables, and inline comments. *(HARD)*
+   * **USE** - Double slash comments are optional and should be used to explain complicated or counter-intuitive segments of the code. *(SOFT)*
+   * **TRIVIAL** - Do not use double slash comments for trivial constructs. *(HARD)*
```
Aprogiena: Double slash trivial is something like this:
```
public int CalculateThat(int a, int b)
{
  // Calculates the sum of a and b.
  int x = a + b;
  // Returns x raised to the power of 2.
  return x * x;
}
```

Those comments strictly describe what the code does. They are pure translation of C# to English language. Thus they are trivial. Those comments are forbidden.

Documentation/policy-comments.md
dangershony: Ok you had me at IntelliSense

Documentation/policy-comments.md
```diff
@@ -0,0 +1,169 @@
...
+ * **INTENT** - Try to describe the intent or the big picture rather than what the code does. *(SOFT)*
+ * **INTERACTIONS** - Explain the interactions of the code with other parts of the system. This includes contracts and limitations that has to be respected - e.g. code protected by locks. *(SOFT)*
```
Aprogiena: Basically any rule that is described in a way that allows multiple interpretations must be soft. Also any rule that makes sense only for some code and not for other code must be soft. And this rule is both, so I think it must be soft.

However, you are still allowed to return the PR during the review for SOFT reasons. 
But you are required to return the PR due to violation of HARD rules. So there still is a difference between SOFT and HARD.

Documentation/policy-comments.md
```diff
@@ -0,0 +1,169 @@
...
+ * **ENUMS** - When a parameter or a return value is enum or if there are any special values, including lower and upper bounds, comment each such special value.  *(HARD)*
+ * **UPDATES** - Keep comments up to date when editing commented code. *(HARD)*
```
Aprogiena: That is a good idea and different issue :)

Aprogiena:
So I made some changes, but not too many actually.

@bokobza I did merge just 2 rules into 1. In case of INLINE.USE and INLINE.TRIVIAL, I could not really merge it as one is SOFT and one is HARD. But I don't think we do have too many rules. It is all basically about - specification of where to use XML doc and where to use inline comments and then a little about what to write there and that's it.

I did cut off a little of the motivation, to make it shorter, but again, not too much because @dangershony  liked it. 

So I think it what we have now is quite good.

Any more comments?

dangershony:
Ok I think I am read to check that in, only the basic rules are hard and that's fine (apart for TRIVIAL which I am not sure about)

Aprogiena:
@dangershony could you write more about the concerns regarding TRIVIAL?

dangershony:
Yeah TRIVIAL means dont just write anything inline but be conscious about your description inside methods, while I definitely get it, this can be restrictive as a HARD rule, there maybe devs that find it useful to have (or are used to) making in line trivial comments.

Aprogiena:
I think it is always possible to explain yourself and convince the reviewer with arguments of why it is useful. I don't see a problem. The purpose of this rule is to avoid obviously trivial inline comments, which are not helping anyone. But both the developer and the reviewer are likely to have understanding and be able to communicate together.

dangershony:
Ok that means you want the reviewer to focus on the inline comments, fair enough.

Aprogiena:
Yes, basically on everything :)

Aprogiena:
I've added new XMLDOC.REFERENCE rule to strictly require using `<see>`

RobertMCarr:
This is a very comprehensive document, however if I'm honest I think it is over complicated and I would expect that potential contributors would find it daunting.  I see a couple of you have talked about balance and I agree that we need to find that balance.  Here are some changes I am going to recommend and see how everyone feels about it:

1. Remove the "motivation" section.  It's a large section that detracts from the actual "rules" and fits more into a beginners programming tutorial than a products code base.  You don't have to justify why you want adequate code commenting.  It's generally understood that the progenitors of a project get to set the rules.
2. Renaming "Hard" to "Required" and "Soft" to "Nice to have" can be better understood by those whose first language isn't English.
3. Remove SUMMARY-2 - It's overkill, if the dev wants to include a second summary they can do but there's no need to define a rule about it, even a soft one.
4. Don't enforce method documentation unless they need it.  At most I would say that the reviewer can reserve the right to request a method is documented before authorising the pull request.  Methods names should be descriptive enough, combine this with the top level class documentation and it should be obvious what it does.  However describing the format for a comment (as has been done) is good to keep.  I would make it a "Nice to have".
5. Don't make any rules around inline comments.  This "Use double slash (inline) comments for function body code, variables, and inline comments" makes it sound like you have to comment every body and variable.  Inline comments are there for additional assistance to the reader and are not picked up by autodoc'ing tools (should we use one) and so there is no harm in them.  Even trivial comments shouldn't invalidate a review.
6. BIG-PICTURE - This part is a good thing to include, however I would move it to a guidelines section on writing style rather than include it in the rules. 
7. SENTENCES - again I would move this to a guidelines sections.
8. ADD-VALUE - I think this one is a bit arbitrary, no one writes something if they don't think they are adding value.  Maybe move it to a writing style guidelines section if you were to keep it.
9. Contracts - This section appears to be about naming conventions and not commenting.  I would remove this section entirely.

Hard/Required rules should be convertible to a stylecop (or similar) rule.  If they can't be then they are either not well enough defined (clear cut) or shouldn't be Hard/Required.

There is plenty I haven't commented on and so you should assume I agree with what is there and therefore have no recommendations.  Remember we can always add more rules later if we are finding the quality is slipping.

Aprogiena:
@RobertMCarr I have to say you did not make me happy with that comment :-) But thinking about what lead you to write it, I think I did not make you happy either with the proposal. 

To be honest I have considered the proposal as finished and it was not merged just because of lack of time. But it sounds like we need another round of discussion. 

OK, so later I'll come back here and comment on what you just wrote in detail.

Aprogiena:
OK, let's see:

> is over complicated and I would expect that potential contributors would find it daunting.

Consider potential contributors that would be scared of uncommented and undocumented code. My opinion? Rather scare those producing messy code than those producing code that is easy to maintain.

> Remove the "motivation" section

I considered that and realized that at least some motivation is necessary as otherwise people could see policies as policies for policies. But what we want is people to understand the point, which is often miss. Without any motivation, lot of people would be annoyed regardless of what the policy says. 

> You don't have to justify why you want adequate code commenting.

Considering your whole comment, it even maybe proves that we need much larger motivation section than we have. The point was missed. And the smaller the motivation section is, the more the point will be missed.

> Renaming "Hard" to "Required" and "Soft" to "Nice to have" can be better understood by those whose first language isn't English.

Reader is expected to be a programmer with great knowledge of what hardfork and softfork is, thus I find it hard to accept an argument that the reader could not grasp "hard" and "soft". Moreover, I even define what hard and soft means in this document.

> Remove SUMMARY-2 - It's overkill, if the dev wants to include a second summary they can do but there's no need to define a rule about it, even a soft one.

SUMMARY-2 rule is not about second summary, it is a soft requirement to avoid putting things into summary that belong to remarks.

> Don't enforce method documentation unless they need it.

Exactly. And since they are ALWAYS needed, we enforce them. The motivation explains why, but maybe it is too short (many more reasons can be added). IntelliSense is another reason.

>  combine this with the top level class documentation and it should be obvious what it does

Unless you make a mistake and create bug. Then what? Does it what was intended or not? Who can tell?

> Don't make any rules around inline comments. This "Use double slash (inline) comments for function body code, variables, and inline comments" makes it sound like you have to comment every body and variable.

Maybe it needs to be rephrased as your comment suggest it is possible to misinterpret it. The idea is not to comment every body and variable. If it is not clear, then OK, we need to rephrase it. But not remove it, because that is not the point.

> BIG-PICTURE - This part is a good thing to include, however I would move it to a guidelines section on writing style rather than include it in the rules.

Soft rules ARE guidelines. It is even written in the document. This is missing the point again.

> SENTENCES - again I would move this to a guidelines sections.

Same here.

> ADD-VALUE - I think this one is a bit arbitrary, no one writes something if they don't think they are adding value.

Many things are obvious to many people yet there are people for who things are not obvious. This should be easy to understand rule. You'd be surprised how often this rule is violated in real code. Sometimes it is not easy to find a proper comment, and people are lazy and just put trivial useless information there. Sometimes it is fine, but in many cases, it could be better.

> Contracts - This section appears to be about naming conventions and not commenting. I would remove this section entirely.

Missing the point. The text clearly states:

> As you can see we also use "Locked" suffix to the method name here, **but that is a matter of the code style. From the comments point of view**, the remarks section is very important for everyone who wants to call the method. 

It should be clear from the text that the Locked suffix is matter of naming conventions, but this is not what is interesting in this example. 

> Hard/Required rules should be convertible to a stylecop (or similar) rule.

Some do, some don't. StyleCop is not AI. How do you check with StyleCop that UPDATES rule is being followed? You can't.

>  Remember we can always add more rules later if we are finding the quality is slipping.

Exactly, that is what we do with this policy, because the quality is slipping. This is the response to the slipping quality.

dangershony:
Correct I was planning to commit this already, I want to give this restrictive commenting policy a try, I used to not be a big fan of such high details of comments but I see the benefits myself.

When I look at one of the classes @Aprogiena documented I get the whole logic very clear just from looking at the comments. it also makes me think more about my code when documenting myself.

If we find it too restrictive we can loose the policies.

bokobza:
@Aprogiena are all the changes included in your PR?
I'm keen to commit this and then modify the stylecop to enforce some of those rules.

Aprogiena:
I would like to make one smaller update to address one of the comment of Robert. This is minor change, I'll do it immediately.

Aprogiena:
Should be OK now.

PR #255, opened 2017-07-26
Remaining walletmanager tests

Neurosploit:


bokobza:
I'll take this one @dangershony

Neurosploit:
Just a comment. After refactoring these transactionmodel tests they still fail every now and then. 
So we are left with 2 possible causes: 1. Loadpooltest being async. 2. There is a bug in Transaction.Parse related to multi threading, the parsing of the varstring type or the use of the varstring type in parsing leading to a length that is outside the byte array we are loading.

bokobza:
maybe this has to do with these BitcoinStream errors: https://github.com/stratisproject/NStratis/commit/5bf64a0ccc0a2df6479da0527e09d834499d0745

I'll get on with your review and then I'll look into that.

Stratis.Bitcoin.Tests/Wallet/WalletManagerTest.cs
```diff
@@ -1330,24 +3185,62 @@ private Features.Wallet.Wallet GenerateBlankWallet(string name, string password)
                 AccountsRoot = new List<AccountRoot> { new AccountRoot { Accounts = new List<HdAccount>(), CoinType = (CoinType)Network.Main.Consensus.CoinType } },
             };
 
-            return walletFile;
+            return (walletFile, extendedKey);
         }
 
-        private ChainedBlock AppendBlock(ChainedBlock previous, params ConcurrentChain[] chains)
+        private Block AppendTransactionInNewBlockToChain(ConcurrentChain chain, Transaction transaction)
```
bokobza: I feel as though all of these methods can be made static and exported to a class that anyone can use for testing purposes. What do you think?

Stratis.Bitcoin.Tests/Wallet/WalletManagerTest.cs
```diff
@@ -1410,6 +3303,40 @@ private List<HdAddress> GenerateAddresses(int count)
...
+
+        private static (PubKey PubKey, BitcoinPubKeyAddress Address) GenerateAddressKeys(Features.Wallet.Wallet wallet, string accountExtendedPubKey, string keyPath)
+        {
+            var addressPubKey = ExtPubKey.Parse(accountExtendedPubKey).Derive(new KeyPath(keyPath)).PubKey;
+            var address = addressPubKey.GetAddress(wallet.Network);
+
+            return (addressPubKey, address);
+        }
+
+        private static ConcurrentChain GenerateChainWithHeight(int blockAmount)
```
bokobza: Not strictly necessary but adding a Network as a parameter would make this more widely usable.

bokobza:
Wow, again @Neurosploit, you never cease to amaze me.

I thought about something that could help us reduce the amount of code necessary to set up the tests.
What if we had something a bit simpler to create and populate a wallet?
We could either have something like this:
```CreateWallet(name, password, cointype).WithAccounts(3).ContaingAddresses(20, 20)```
or even 
```new WalletCreator 
	{
		Name = "my {i} wallet"
		WalletsCount = 3,
		AccountsCount = 2,
		ExternalAddressesCount = 20,
		InternalAddressesCount = 20,
		etc...
	}
```
What do you think?

Stratis.Bitcoin/Features/Wallet/WalletManager.cs
```diff
@@ -292,221 +292,234 @@ public HdAddress GetUnusedAddress(WalletAccountReference accountReference)
...
-			ExtKey addressExtKey = seedExtKey.Derive(new KeyPath(address.HdPath));
-			BitcoinExtKey addressPrivateKey = addressExtKey.GetWif(walletTree.Network);
-			return addressPrivateKey;
-		}
+            Guard.Assert(lastBlockSyncedHash != null);
+            return lastBlockSyncedHash;
+        }
+
+        /// <inheritdoc />
+        public List<UnspentInfo> GetSpendableTransactions(int confirmations = 0)
```
Neurosploit: I have already talked about trying it with dan however I wanted to do that in a different PR.

Neurosploit:
Yes I was thinking of something like a testwalletbuilder but I need to have a look at what I need to be able to get out of that builder like extkeys and stuff like that.

Stratis.Bitcoin.Tests/Wallet/WalletManagerTest.cs
Neurosploit: yes I would like to create some classes to construct a wallet and a chain.

Stratis.Bitcoin/Features/Wallet/WalletManager.cs
bokobza: thanks

PR #262, opened 2017-07-27
Added empty StyleCop settings file

bokobza:
This adds an empty stylecop settings file.
I'm starting with all the rules disabled so that on subsequent commits it's easy to see which have been enabled.
@dangershony


bokobza:
Closing this as I just realized there is a newer version of stylecop.

bokobza:
Actually, I think we're better off **starting** with the old version of StyleCop.
The reasons for this decision is as follows:
the new StyleCop (StyleCopAnalyzers) needs to be imported as a nuget package and is always run as part of the build. We would have so many warnings right now that some warnings needing immediate attention might be missed. When the code is in a better shape, we can switch to StyleCopAnalyzers.

PR #267, opened 2017-07-28
Bugfix/issue 245

Aprogiena:
This is a quick fix for 245 simply using a lock. The performance hit should be very minimal considering the frequency of ConsensusFeature.Log, which is 5 seconds. Thus I don't think this needs any further optimization or time being spent on this to make it more efficient.

### DIFF
--- a/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
+++ b/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
@@ -150,7 +150,11 @@ namespace Stratis.Bitcoin.BlockPulling
         /// <summary>Current number of bytes that unconsumed blocks are occupying.</summary>
         private long currentSize;
 
+        /// <summary>Lock object to protect access to <see cref="downloadedCounts"/>.</summary>
+        private object downloadedCountsLock = new object();
+
         /// <summary>Maintains the statistics of number of downloaded blocks. This is used for calculating new actualLookahead value.</summary>
+        /// <remarks>All access to this object has to be protected by <see cref="downloadedCountsLock"/>.</remarks>
         private List<int> downloadedCounts = new List<int>();
 
         /// <summary>Points to a block that was consumed last time. The next block returned by the puller to the consumer will be at location + 1.</summary>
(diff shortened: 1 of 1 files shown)


Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -150,7 +150,11 @@ public int MaxBufferedSize
         /// <summary>Current number of bytes that unconsumed blocks are occupying.</summary>
         private long currentSize;
 
+        /// <summary>Lock object to protect access to <see cref="downloadedCounts"/>.</summary>
+        private object downloadedCountsLock = new object();
+
         /// <summary>Maintains the statistics of number of downloaded blocks. This is used for calculating new actualLookahead value.</summary>
+        /// <remarks>All access to this object has to be protected by <see cref="downloadedCountsLock"/>.</remarks>
```
dangershony: You only have a single list why not use the ConcurrentBag?

Aprogiena: Mostly because of this:

```
        public decimal MedianDownloadCount
        {
            get
            {
                lock (this.downloadedCountsLock)
                {
                    if (this.downloadedCounts.Count == 0)
                        return decimal.One;
                    return (decimal)GetMedian(this.downloadedCounts);
                }
            }
        }
```

If I had a ConcurrentBag there, the race would not be prevented. I could still get .Count != 0 in the first IF line, then having the collection changed by another thread and then calling GetMedian with empty collection and get that exception again. Unless I changed the logic of course, but why would I do that. 

In general, using Concurrent collections is not saving your from anything except for trivial single operations. Once you have conditional operation like this, Concurrent stuff is not helping. This is why I try to avoid Concurrent collections completely as they tend to misguide you - giving you a false sense of security.

dangershony: You could ToList() on the dictionary and work with the copy of the list in that point in time.
But that's fine its actually way more clear what is happening when using a lock() then a concurrent collection, though I would try to avoid calling i to methods from inside a lock (ah well its a static method in this case so fine).

dangershony:
This is a very minor race condition though

PR #276, opened 2017-08-01
Feature/nlog

Aprogiena:
Migration to NLog from Serilog.

Devs should read using-nlog.md on how to use it.

### DIFF
--- a/Documentation/using-nlog.md
+++ b/Documentation/using-nlog.md
@@ -1,0 +1,133 @@
+# Using NLog
+
+We've migrated to NLog, but how do you use it now to create logs in the code? There are two things that you need to understand - configuration of the logging and logging itself. Starting with the latter, that is the simple part.
+
+## Adding Logging Code
+
+You will use a class logger or an instance logger (or even both in the same class). A class logger is used to log in static methods. An instance logger is more common and that is used when you log inside non-static methods. 
+
+
+### Creating Instance Logger
+
+You always need to pass `ILoggerFactory` to the class constructor and then you can create an instance logger as follows:
+
+```
+/// <summary>Instance logger.</summary>
+private readonly ILogger logger;
+...
+this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+```
+
+And that's it! Note that you don't want to pass ILogger, always pass ILoggerFactory. It is NLog's job to filter which logs do you want to see and which you don't. 
+For more information on that, see [Logging Configuration](#logging-configuration)
+
+
+### Creating Class Logger
+
+Class logger is little more difficult as you may not have access to any factory. So you will have to create and initialize a new one:
+
+```
+ILoggerFactory loggerFactory = new LoggerFactory();
+loggerFactory.AddConsole();  // Only if you want to include logging to the console.
+loggerFactory.AddNLog()();   // Only if you want to include logging to the disk. This requires using NLog.Extensions.Logging.
+
+```
+
+Then you can create a class logger as follows:
+
+```
+/// <summary>Class logger.</summary>
+private static readonly ILogger logger;
+...
+logger = loggerFactory.CreateLogger(typeof(YourClassName).FullName);
+```
+
+
+### Using Loggers
+
+Both types of loggers are used in a same way and we are using `Microsoft.Extensions.Logging` and `ILogger` interface, so if you knew how to log before migration to NLog, 
+there is no change for you - use `LogTrace`, `LogDebug`, `LogInformation`, `LogWarning`, `LogError`, and `LogCritical` methods on your logger:
+
+```
+// Instance logger on trace level:
+this.logger.LogTrace("Message: {0}", message);
+// Class logger on warning level:
+logger.LogWarning("Message: {0}", message);
+```
+
+
+## Logging Configuration
+
+
+### Default logging
+
+NLog defines the minimum log level to be logged for every class. If you run the node with default settings, information level is the minimum log level for all classes 
+and all matching logs go to both console and the main log file. The main log file is `node.txt` under `$DATADIR/Logs` directory. This log is rotating, which means 
+that every day there is a new log file and the old one is moved to `node-YYYY-MM-DD.N.txt` using the current time stamp. This means that the log file created on 31st July 2017 
+will go to `node-2017-08-01.1.txt` (not to `node-2017-07-31.1.txt`). Logs from last 7 days are archived, which means that if there are more than 7 old log files, the oldest 
+will be deleted.
+
+
+### Debug Option
+
+You can specify `-debug` command line option, which works similarly to how it works in Bitcoin Core. This enables logging of selected or all classes on trace level and this 
+setting is only for the log files, not console. To enable trace level for all classes, you use `-debug=1`. If you want to enable trace level just for some classes, you use `-debug=pattern1,pattern2,...`. 
+There are several shortcuts preset for most common components. For example `rpc` is a shortcut pattern for `Stratis.Bitcoin.Features.RPC.*`. This means setting `-debug=rpc` is 
+the same as setting `-debug=Stratis.Bitcoin.Features.RPC.*`, which means that all classes under `Stratis.Bitcoin.Features.RPC` namespaces will log on trace level. Similarly, 
+you can use `consensus` as a shortcut for `Stratis.Bitcoin.Features.Consensus.*`. The full list of shortcuts is available in `Stratis.Bitcoin.Logging.LogsExtension.cs`. 
+If a shortcut does not exist, you can simply use the full name of a class or a namespace. You have to use `.*` suffix for namespaces to include all its classes.
+
+
+### NLog.config
+
+If you don't want to use the debug option, or you need a more complex set of rules, you can use `NLog.config` file. You need to create `NLog.config` file in the folder where 
+you have your DLL files - e.g. `Stratis.Bitcoin.dll`. The path to this folder is usually something like `StratisBitcoinFullNode/Stratis.BitcoinD/bin/Debug/netcoreapp1.1`.
+
+Here is the very basic `NLog.config` configuration file that you should start with:
+
+```
+<?xml version="1.0" encoding="utf-8" ?>
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true">
+  <targets>
+    <target xsi:type="File" name="debugFile" fileName="debug.txt" layout="[${longdate:universalTime=true} ${threadid}${mdlc:item=id}] ${level:uppercase=true}: ${callsite} ${message}" encoding="utf-8" /> 
+    <target xsi:type="null" name="null" formatMessage="false" /> 
+  </targets>
+
+  <rules>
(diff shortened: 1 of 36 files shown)


Documentation/using-nlog.md
```diff
@@ -0,0 +1,134 @@
+# Using NLog
```
bokobza: This should be called Logging guidelines or whatever. 
You can also ignore the fact that we 'moved' to Nlog, maybe just say that we use NLog to help us with the logging.

Documentation/using-nlog.md
```diff
@@ -0,0 +1,134 @@
...
+/// <summary>Class logger.</summary>
+private static readonly ILogger logger;
+...
+logger = loggerFactory.CreateLogger(typeof(YourClassName).FullName);
+```
+
+
+### Using Loggers
+
+Both types of loggers are used in a same way and we are using `Microsoft.Extensions.Logging` and `ILogger` interface, so if you knew how to log before migration to NLog,
```
bokobza: No need to mention the migration, just show us how it's done :-)

Stratis.Bitcoin/Stratis.Bitcoin.csproj
```diff
@@ -21,7 +21,8 @@
   <ItemGroup>
     <PackageReference Include="DBreeze" Version="1.85.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="1.1.2" />
-    <PackageReference Include="Serilog.Extensions.Logging.File" Version="1.0.1" />
+    <PackageReference Include="NLog" Version="5.0.0-beta09" />
```
bokobza: as a rule of thumb we don't use preview versions of external libraries

Stratis.Bitcoin/Logging/LogsExtension.cs
```diff
@@ -7,99 +7,181 @@
...
+
+            // Configure main file target rules based on node settings.
+            // TODO: Preload enough args for -conf= or -datadir= to get debug args from there. We currently forbid logging before the logging is initialized.
+            // TODO: Currently only takes -debug arg.
             var keyToCategory = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
             {
                 //{ "addrman", "" },
                 //{ "alert", "" },
-                { "bench", "Stratis.Bitcoin.FullNode.ConsensusStats" },
+                { "bench", $"{nameof(Stratis)}.{nameof(Stratis.Bitcoin)}.{nameof(Stratis.Bitcoin.Features)}.{nameof(Stratis.Bitcoin.Features.Consensus)}.{nameof(Stratis.Bitcoin.Features.Consensus.ConsensusStats)}" },
```
bokobza: nice!

Documentation/using-nlog.md
Aprogiena: ok, will fix

Documentation/using-nlog.md
Aprogiena: ok, will fix

Stratis.Bitcoin/Stratis.Bitcoin.csproj
Aprogiena: Sadly, not possible in this case. NLog is waiting with stable release for .NET Core 2.0. NLog 4 is not .NET Core compatible. Can't fix this without getting rid of NLog ...

bokobza: ok fair enough

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: well ... :) looks super ugly to me, but ...

bokobza: what about
typeof(Stratis.Bitcoin.FullNode.ConsensusStats).FullName?

bokobza: not sure if that works lol

Aprogiena: it works only for final classes - but except for this ConsensusStats and FullNode, everything else is just namespaces, so yes, we can have this for those two, but that would be inconsistent with others ...

bokobza: ah yes indeed

bokobza:
I'm fine with the change. @dangershony you want to have a look?

Stratis.Bitcoin/Logging/LogsExtension.cs
```diff
@@ -7,99 +7,181 @@
...
+    /// Integration of NLog with Microsoft.Extensions.Logging interfaces.
+    /// </summary>
     public static class LogsExtension
     {
-        public static void AddFilters(this ILoggerFactory loggerFactory, LogSettings settings)
+        /// <summary>Width of a column for pretty console/log outputs.</summary>
+        public const int ColumnLength = 16;
+
+        /// <summary>Currently used node's log settings.</summary>
+        private static LogSettings logSettings;
```
dangershony: I think we should avoid static fields for logging, now that we have a DI all logs should be inject-able.
Is there any place this is used?

Stratis.Bitcoin/Logging/LogsExtension.cs
```diff
@@ -7,99 +7,181 @@
...
+            if (LogManager.Configuration == null) LogManager.Configuration = new LoggingConfiguration();
+
+            // Installs handler to be called when NLog's configuration file is changed on disk.
+            LogManager.ConfigurationReloaded += NLogConfigurationReloaded;
+        }
+
+        /// <summary>
+        /// Event handler to be called when logging <see cref="NLog.LogManager.Configuration"/> gets reloaded.
+        /// </summary>
+        /// <param name="sender">Not used.</param>
```
dangershony: An example for not adding a param ;)
However I can see how a caller would immediately see on the tolltip this are not used.

Stratis.Bitcoin/Logging/LogsExtension.cs
```diff
@@ -7,99 +7,181 @@
...
+        /// <summary>
+        /// Initializes application logging.
+        /// </summary>
+        static LogsExtension()
+        {
+            // If there is no NLog.config file, we need to initialize the configuration ourselves.
+            if (LogManager.Configuration == null) LogManager.Configuration = new LoggingConfiguration();
+
+            // Installs handler to be called when NLog's configuration file is changed on disk.
+            LogManager.ConfigurationReloaded += NLogConfigurationReloaded;
```
dangershony: What are the cases that the configuration will be reloaded?
And how is that triggered (file changed I guess)?

Stratis.Bitcoin/Logging/LogsExtension.cs
```diff
@@ -7,99 +7,181 @@
...
+            mainTarget.Name = "main";
+            mainTarget.FileName = Path.Combine(folder.LogPath, "node.txt");
+            mainTarget.ArchiveFileName = Path.Combine(folder.LogPath, "node-${date:universalTime=true:format=yyyy-MM-dd}.txt");
+            mainTarget.ArchiveNumbering = ArchiveNumberingMode.Sequence;
+            mainTarget.ArchiveEvery = FileArchivePeriod.Day;
+            mainTarget.MaxArchiveFiles = 7;
+            mainTarget.Layout = "[${longdate:universalTime=true} ${threadid}${mdlc:item=id}] ${level:uppercase=true}: ${callsite} ${message}";
+            mainTarget.Encoding = Encoding.UTF8;
+
+            LogManager.Configuration.AddTarget(mainTarget);
```
dangershony: Perhaps the instance of LogManager can live inside LogSettings

Aprogiena: LogManager is living in NLog, you don't create your own instance.

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: Yes and yes :)

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: Yes, most of the things you don't like here are solely for the adhoc reloading. This is for developers only. It does not happen very often, but you are running node and it is running very long time and has some very interesting context and then you find out that you are missing  some logs (or you think you are seeing too many logs, or you want some logs to put into separate file, etc.), so what you can do is to open the config file (it is actually described in the using-logging.md) and change it without stopping your node. This will cause this handler to be called. 

Now this handler is currently static method. As such, it needs those static fields. 

So an option is not to have this handler as a static method, but then we would need somehow to get access to NodeSettings. So What we could do is to put this handler to NodeSettings, which would make the dataFolder and LogsSettings accessible. And we would have to install the handler when we have NodeSettings and not before.

Do you think it is worth and better than having those static things?

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: See below.

Stratis.Bitcoin/Logging/LogsExtension.cs
dangershony: LogManager is a staic class then

Stratis.Bitcoin/Logging/LogsExtension.cs
dangershony: Ah ok I missed this part
```LogManager.ReconfigExistingLoggers();```
It will change the already configured logger factory I guess none of this static components are used during normal operation only at load/reload this is fine then.

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: yes - https://github.com/NLog/NLog/blob/master/src/NLog/LogManager.cs

Stratis.Bitcoin/Logging/LogsExtension.cs
Aprogiena: Yes, the most of the code complexity here is because of reloading functionality.

The level of interconnection with loggerFactory you are looking for is simply not there. Whole NLog configuration is a static thing, lives in NLog. The only thing we do here is that we need to initialize it correctly (including when it reloads) using what we have in NodeSettings.

dangershony: Ok I look bit further in to NLog, what I don't like is they keep a reference to a static ILoggerFactory.
But that's internal to NLog what you did here is fine actually.

dangershony:
Only comment I have now is to move the folder Logging inside Configuration and GTG

Aprogiena:
Done, but needed to modify many files for that. But done.

bokobza:
I apologize for not seeing this sooner, but can you change LogsExtension to LogsExtensions, by convention we use the plural form of "extension" for extension methods.
In fact, I think the name would better be either ILoggerFactoryExtensions or LoggingConfiguration (or something similar).

bokobza:
I'll push it in when you're done.
Looking forward to using it!

Aprogiena:
There you go

bokobza:
Love it

PR #279, opened 2017-08-02
Bugfix/issue 244

### LINKED ISSUE #244: LookaheadBlockPuller NextBlockCore race condition
Race condition (R1) and potential race condition (R2) exist in LookaheadBlockPuller: 
https://github.com/stratisproject/StratisBitcoinFullNode/blob/fd06b73f8251421c7084cddc44324f9166fedece/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs#L406

R1: As soon as IsDownloading returns false here, another thread can change the state (consider BlockPullerBehavior.Node_MessageReceived, just before the newly downloaded block is pushed, the above check for DownloadedBlocks will fail - block not pushed yet - but then IsDownloading here returns false), which will cause this thread to call AskBlocks for a block that was already downloaded. 

R2: Another potential race condition here is unclear: Is there also a possibility that this happens when the block is being downloaded? I.e. is this the only thread that can exist in the node and lead to execution of DistributeDownload? If there can be another thread executing DistributeDownload, the map can be updated after IsDownloading returns false, which would lead to a single block being downloaded twice. 

R2 is not clear because NextBlockCore is called from NextBlock, which is called from BlockNotification.Notify, which is run in its own async loop, but the NextBlock is also called from ConsensusLoop.ExecuteNextBlock, called from ConsensusLoop.Execute, called from ConsensusFeature.RunLoop, which runs in its own thread. So it seems like there are two different threads that can end with calling NextBlock, but it is unclear whether there is not some mechanism to prevent the concurrency.

The problem in both cases is the same - IsDownloading allows you to get state of shared resource, but once IsDownloading returns, the state is no longer guaranteed and you are working with potentially outdated information. It would be possible to use such an information, if later the code guaranteed that the unwanted behavior (downloading block that has been downloaded already, or is being downloaded) could not occur. This is not the case, however.

Aprogiena:
Fix of one of the race conditions by reworking NextBlockCore and PushBlock methods of the puller and lookahead puller. 

DownloadTaskFinished now contains adding block to the list of downloaded blocks, so the whole access to the shared structures is not atomic.

New method BlockPushed is called when a new block is downloaded and the puller can override this to react on this event.

New method InjectBlock is now used to inject blocks manually to the puller - this is only good for testing.

Fix #244


Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
```diff
@@ -729,9 +730,10 @@ public Block[] GenerateStratis(int blockCount, List<Transaction> passedTransacti
 				blocks.Add(block);
 				if (broadcast)
 				{
-					var newChain = new ChainedBlock(block.Header, block.GetHash(), fullNode.Chain.Tip);
+                    uint256 blockHash = block.GetHash();
```
dangershony: Weird spaces

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -149,21 +149,26 @@ protected BlockPuller(ConcurrentChain chain, IReadOnlyNodesCollection nodes, Pro
...
         {
-            uint256 hash = block.Header.GetHash();
-
-            DownloadedBlock downloadedBlock = new DownloadedBlock()
-            {
-                Block = block,
-                Length = length,
-            };
+            this.logger.LogTrace($"(blockHash:'{blockHash}',downloadedBlock.Length:{downloadedBlock.Length})");
+            this.logger.LogTrace("(-)");
```
dangershony: Why the second log entry?

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -235,14 +240,13 @@ private void AssignPendingVectors()
         }
 
         /// <inheritdoc />
-        public bool IsDownloading(uint256 hash)
+        public void CheckBlockStatus(uint256 hash, out bool IsDownloading, out bool IsReady)
```
dangershony: I suggest instead of using out param to return the new anonymous tuple result
```
 public (bool IsDownloading, bool IsReady) CheckBlockStatus(uint256 hash)
```

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -504,7 +513,7 @@ internal bool DownloadTaskFinished(BlockPullerBehavior peer, uint256 blockHash)
                             if (this.assignedBlockTasks.Remove(blockHash) && peerPendingDownloads.Remove(blockHash))
                             {
                                 // Task was assigned to this peer and was removed.
-                                res = true;
+                                res = this.downloadedBlocks.TryAdd(blockHash, downloadedBlock);
```
dangershony: Super duper

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -397,7 +375,13 @@ private Block NextBlockCore(CancellationToken cancellationToken)
                 cancellationToken.ThrowIfCancellationRequested();
                 ChainedBlock header = this.Chain.GetBlock(this.location.Height + 1);
                 DownloadedBlock block;
-                if (header != null && TryRemoveDownloadedBlock(header.HashBlock, out block))
+
+                bool isDownloading = false;
+                bool isReady = false;
+                if (header != null) CheckBlockStatus(header.HashBlock, out isDownloading, out isReady);
```
dangershony: This logic will make my tuple suggestion hard to implement so using "out" looks fine.

Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
Aprogiena: I know I did not fix code style in those files that needed just minor update to compile. Should I fix the code style as well? If so, hopefully I am not required to add comments there ...

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: That's very important thing of my logging style, but hard to explain here. Basically you want all entries and exists from the method to be in the log. This should be more evident when I fully log these as a part of quality score issue. Please ignore now.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: Is there any actual benefit to it? As you know my style of writing things is to avoid using these new things as they are not readable for many coders. Yes, you can read them, that is fine, but there can be many people that are not happy to read them. This is why I would never use it myself. It is just a syntactical sugar that is not very hard to understand, but does not bring much and is just too new to be used.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: Cool :)

Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
dangershony: yeah ok no need.

PR #280, opened 2017-08-02
replace submodule NStratis with NuGet

DanGould:
Also passes more tests. "External" still shows up in vs2017 with `NBitcoin (unavailable)` and `HashLib (unavailable)` but everything builds


dangershony:
Ah no that was not he intention to change the full node

PR #283, opened 2017-08-03
Breeze api move

bokobza:


bokobza:
@dev0tion @dangershony @Aprogiena 
guys, this is to move the Breeze.Api, Breeze.Api.Tests and Breeze.Wallet projects from the Breeze repo to the full node repo.
These projects are very linked so it was always a bit complicated to work across different repos.
With the TumbleBit repos added to the mix, it's a nightmare.
So this is the first step, the next step will be to update the Breeze repo to still use the Full Node as a submodule but to reference the projects that were there in situ with the ones brought from the FullNode submodule.

Aprogiena:
@bokobza Could you provide some details about the problem? I am all OK with the change. What I want to say is that I don't understand the problem. Here and there I hear that the problem is related to a project having dependencies that are in separated repos. Is that right?

If so, what is the problem? I did manage a project with several such dependencies and there was no problem, but maybe I was just used to do something that you find annoying, or maybe we did something differently. 

So, could you describe what the problem was and why it was nightmare to deal with it and how this change makes it less nightmare? Just for me to understand what is different to what I have experience with. Thanks!

bokobza:
Sure.
The problem here was that the Breeze projects, despite being very close to the Stratis node in terms of functionality and delivery schedule, was living in another repo on Github.
So every time the full  node was updated, we had to first push it into GitHub and then pull it on the Breeze repo. This created build problems as you can imagine and didn't allow us to move fast as it was quite a process to keep everything in sync.
Now that Tumblebit will also be referencing the full node and the light wallet stuff, it's easier if all the projects are in the same repo.

Aprogiena:
Chmm, I guess this is because of those submodules someone mentioned (I don't know what that is).

What we had was several repos depending on each other but only in C# project. Git(Hub) had no information about that dependency. So we had project A in repo RA and project B in repo RB and B depended on A.

To MSVS it looked like the 2 solutions SA and SB, but SA only contained A and SB contained both B and A.

Folders were like:

`$root/SA/A/structure`
`$root/SB/B/structure`

so solution SB had project with directory like `../SA/A/A.csproj` as well as its own `B/B.csproj`

This worked very flawlessly and conveniently for us without any problems and it allowed having 2 separated repos.

bokobza:
Yeah that's exactly what we had, before we moved to using submodules.
In any case, the Stratis.Bitcoin.Api and the LightWallet belong to the same repo as the full node, hence the change.

Aprogiena:
So would you say that submodules are worse than that alternative setup?

PR #286, opened 2017-08-03
RPC getpeerinfo

mikedennis:
Initial implementation of RPC method getpeerinfo. Some RPC properties are not populated. In trello backlog https://trello.com/c/4SRyzKjI/59-connection-rpc-getpeerinfo 

@bokobza

### DIFF
--- a/Stratis.Bitcoin/Features/RPC/Controllers/ConnectionManagerController.cs
+++ b/Stratis.Bitcoin/Features/RPC/Controllers/ConnectionManagerController.cs
@@ -1,9 +1,13 @@
-﻿using System;
-using System.Net;
-using Microsoft.AspNetCore.Mvc;
+﻿using Microsoft.AspNetCore.Mvc;
+using NBitcoin.Protocol;
+using Stratis.Bitcoin.Base;
 using Stratis.Bitcoin.Configuration;
 using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Features.RPC.Models;
 using Stratis.Bitcoin.Utilities;
+using System;
+using System.Collections.Generic;
+using System.Net;
 
 namespace Stratis.Bitcoin.Features.RPC.Controllers
 {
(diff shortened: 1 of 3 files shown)


bokobza:
Oh yeah much clearer :-)
Still mine!

bokobza:
I'll probably do it tomorrow morning @mikedennis.

Stratis.Bitcoin/Features/RPC/Models/PeerNodeModel.cs
```diff
@@ -0,0 +1,132 @@
+﻿using Newtonsoft.Json;
+
+namespace Stratis.Bitcoin.Features.RPC.Models
+{
+    /// <summary>
+    /// Data structure for RPC peer info nodes.
+    /// </summary>
+    public class PeerNodeModel
```
bokobza: I can see that a few properties are missing, like relaytxes or minping (there may be others). Can you please add them?

Stratis.Bitcoin/Features/RPC/Models/PeerNodeModel.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// <summary>
+        /// The Unix epoch time of the last send from this node.
+        /// Currently not populatated.
+        /// </summary>
+        [JsonProperty(PropertyName = "lastsend")]
+        public int LastSend { get; internal set; }
+
+        /// <summary>
+        /// The Unix epoch time when we last received data from this node.
+        /// Currently not populatated.
```
bokobza: extra 'ta' in populated.
Can you please add these as a 
```
/// <remarks>
///  Currently not populated.
/// </remarks>
```

Stratis.Bitcoin/Features/RPC/Models/PeerNodeModel.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// The heights of blocks we're currently asking from this peer.
+        /// Currently not populated.
+        /// </summary>
+        [JsonProperty(PropertyName = "inflight")]
+        public uint[] Inflight { get; internal set; }
+
+        /// <summary>
+        /// Number of blocks on peer.
+        /// </summary>
+        [JsonProperty(PropertyName = "blocks")]
```
bokobza: I can't see blocks in the cpp code 🤔 
Is the block height what is actually supposed to be in "inflight"?

Stratis.Bitcoin/Features/RPC/Controllers/ConnectionManagerController.cs
```diff
@@ -35,5 +39,60 @@ public bool AddNode(string endpointStr, string command)
             }
             return true;
         }
+
+        /// <summary>
+        /// RPC implementation of getpeerinfo
+        /// </summary>
+        /// <returns>List of RPC peer nodes</returns>
+        [ActionName("getpeerinfo")]
+        public List<PeerNodeModel> GetPeerInfo()
```
bokobza: Could you please add a list to the cpp code on github?
This way it'll be easier for people to see the source.

https://github.com/bitcoin/bitcoin/blob/0.14/src/rpc/net.cpp

Stratis.Bitcoin/Features/RPC/Controllers/ConnectionManagerController.cs
```diff
@@ -35,5 +39,60 @@ public bool AddNode(string endpointStr, string command)
...
+        {
+            List<PeerNodeModel> peerList = new List<PeerNodeModel>();
+            int index = 0;
+            foreach (Node node in this.ConnectionManager.ConnectedNodes)
```
dangershony: I wonder if it would be a good idea to create an iteration copy of ConnectedNodes.

mikedennis:
@bokobza @dangershony Thanks for the feedback. Will fix it up shortly.

Regarding "blocks" property. I found the same thing myself. The RPCClient does however pull it back and I ran the BitcoinD client and tested and it is populated with data so I'm a bit confused on where it comes from as well. I'm not sure if I'm putting the right data into it either, as I'm populating it as follows:

``` C#
                    ChainHeadersBehavior chainHeadersBehavior = node.Behavior<ChainHeadersBehavior>();
                    if (chainHeadersBehavior != null)
                    {
                        rpcNode.Blocks = chainHeadersBehavior.PendingTip.Height;
                    }
```

bokobza:
Concerning "blocks", you can remove the property entirely as I think it comes from an old version of Core.

bokobza:
When you're done with that, I'll merge.

bokobza:
thanks @mikedennis !

PR #289, opened 2017-08-04
Feature/xmldoc stratis.bitcoin.fullnode

Aprogiena:
XMLDOC for Stratis.Bitcoin FullNode/IFullNode


Stratis.Bitcoin/FullNode.cs
```diff
@@ -1,286 +1,336 @@
...
+using System.Threading.Tasks;
 
 namespace Stratis.Bitcoin
 {
+    /// <summary>
+    /// Node providing all supported features of the blockchain and its network.
+    /// </summary>
+    public class FullNode : IFullNode
+    {
+        /// <summary>Instance logger.</summary>
```
bokobza: Logger instance would be better :-)

bokobza:
@Aprogiena only one comment and then a little conflict. :-)

Stratis.Bitcoin/FullNode.cs
Aprogiena: I know what you mean, but I use this as a technical term. Please check [Using Logging](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md) - I define there Instance logger and Class logger as two different things. This is Instance logger, so in fact this is "Instance logger's instance". 

If you are happy with the terms, the comment should be OK, if you think the comment should be fixed, possibly the terms themselves are not OK.

bokobza: ah ah ok it means the logger for this instance, not an instance of  a Logger.
What about "this instance's logger"?
If not then leave it as it is, it's fine.

Aprogiena: If you don't really mind, I would say not (it is like this elsewhere already).

The meaning is supposed to be "logger for an instance of this object" vs "logger for this class".

bokobza: yeah I understood the second part.

Issue #285, opened 2017-08-03
NodeBuilder: AddRPC() requires UseWallet()


mikedennis:
If you don't include UseWallet() then the Stratis node crashes on startup. Crash is due to a dependency on WalletManager (which isn't there if you don't include UseWallet())

mikedennis:
Is this a dependency we require or should we load WalletRPCController conditionally depending upon whether UseWallet() was included?

dangershony:
For now a quick fix is to allow wallet to be null however a better fix is to separate the controllers in to their features (like with api).

mikedennis:
This is the actual exception being thrown:
```
System.InvalidOperationException occurred
  HResult=0x80131509
  Message=Unable to resolve service for type 'Stratis.Bitcoin.Features.Wallet.IWalletManager' while attempting to activate 'Stratis.Bitcoin.Features.Miner.PosMinting'.
  Source=<Cannot evaluate the exception source>
  StackTrace:
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.Service.PopulateCallSites(ServiceProvider provider, ISet`1 callSiteChain, ParameterInfo[] parameters, Boolean throwIfCallSiteNotFound)
```
The exception is thrown when copying the node services to the RPC services in this line
https://github.com/stratisproject/StratisBitcoinFullNode/blob/9aec20b1a922fc73f2aff755c35c44881bdbd40a/Stratis.Bitcoin/Features/RPC/RPCFeature.cs#L48

dangershony:
Unfortunately I cant see to reproduce this bug.
If I comment out .UseWallet() in NodeBuilder this tests fails only on the RPC call 'SendCommand' (not the node init)
```
        [Fact]
        public void CanMineBlocks()
        {
            using(NodeBuilder builder = NodeBuilder.Create())
            {
                var stratisNodeSync = builder.CreateStratisNode();
                builder.StartAll();
                var rpc = stratisNodeSync.CreateRPCClient();
                rpc.SendCommand(NBitcoin.RPC.RPCOperations.generate, 10);
                Assert.Equal(10, rpc.GetBlockCount());
            }
        }
```

mikedennis:
If you startup the node like this you will encounter the error:

``` C#
 var node = new FullNodeBuilder()
                .UseNodeSettings(nodeSettings)
                .UseStratisConsensus()
                .UseBlockStore()
                .UseMempool()
                .AddPowPosMining()
[...]

PR #291, opened 2017-08-06
Fix for crash on RPC Start when no wallet in node.

mikedennis:
If node service is missing dependencies then skip adding that service to RPC.

Fix for #285


dangershony:
See my comment in the issue #285

dangershony:
@mikedennis can we close this?

mikedennis:
Working on a better solution for this.

PR #293, opened 2017-08-06
Feature/logging async wrapper

Aprogiena:
This is previously unannounced feature, but it is so trivial and opt-in that I hope there will be no objection for accepting it.

All it does is that it allows developers to use async wrapper around the debugTarget in NLog.config. This needs just little extra handling in `LoggingConfiguration.cs` and the explanation of how to use it (and when it is a good idea) has been added to the documentation.

As a developer, you decide what goes to NLog.config, so you can ignore this completely, but if you want to use async wrapper, with this you can (previously you couldn't).


dangershony:
What are the benefits in an AsyncLogger? I assume its a fire and forget method so the caller doesnt wait for the the log method to finish. (else you would need to change all code to use async/away right?

Aprogiena:
Async wrapper means the operation is done asynchronously, so think about of a queue that you can very quickly insert to and then a separate thread from time to time takes all items in the queue and flushes them to the disk. That is what happens with async wrapper. The problem is that if your app crashes, you lose the queue, which can contain a lot of important logs.

Without async wrapper, if you make a log (calling logger.LogTrace for example), it is flushed to the disk immediately, so if the app crashes just after that, you have all logs. So for debugging (which is what we use NLog.config for), the default is usually better to be not-async. But sometimes you are not expecting crashes and then async wrapper will speed up things by not waiting for disk.

dangershony:
Can we perhaps make use the RELEASE flag to set to use AsyncWrapper when compiling for release?
I can imagine writing to disk on every log slows down the node quite a lot, or perhaps by default we should use the async wrapper.

Aprogiena:
There is a misunderstanding here. So let me try to explain this one step a time. Maybe try to reread the [using logging instructions](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md). There is a lot about that there.


1) By default we log to console and to node.txt file (with log rotation feature) on INFO+ levels. This means very few logs being actually created in the production mode ("release" mode). Those logs to node.txt are controlled by NodeSettings - i.e. command line argument `-debug`. Using `-debug` will lower the logging level to TRACE+ for all selected (or all) components if the user wants to. Most users don't want to and won't touch it. 

So by default, very little logs are created and async wrapper does not make much sense, but can be used (it is not now, but it is easy to add). This is controlled by the code in `LoggingConfiguration.cs` - find the code below commend line `// Configure main file target`

2) For developers, there is much more powerful option - to use NLog.config file. Users probably won't ever touch this, so you don't need to concern about speed because no production server will run with NLog.config file. The recommended defaults there log to another file - debug.txt.

But for developers, this is golden opportunity how to log very specifically what they need. And again in using logging documentation you have description on how to use this - and by default it is recommended not to use async wrapper because it is expected that if you dig into something, you want to avoid losing important logs if the app crashes. But in some cases, you might be interesting in the speedup that async wrapper offers, so this PR just allows you to do that with the default "log all on trace level debugFile target". 

You are still free not to use this target at all, you are still free to have as many different targets as you want, this is just convenient default thing for devs to have there. 

The documentation is being updated as a part of this PR to describe how to use that async wrapper here. But none of this is relevant for production (release) mode.

----------------
So to summarize - if you want using async wrapper for production logging (i.e. to node.txt) we can change `LoggingConfiguration.cs` how it configures the default main target. If we do that we will need to end flush and shutdown calls to NLog.LogManager at the end of the program (i.e. BitcoinD and StratisD or in the FullNode). The only difference in terms of performance would be in case the user runs with `-debug` flag, but arguably this is the case when the user (or the developer asking the user to do so) is interested in debugging some weird crash, in which case using async will be bad thing to do. However, if this is the case, the developer can advice the user to use NLog.config file to avoid that default async.

So I am not sure it is worth to have the main target for production using async. But we can implement it.

dangershony:
Ok I missed this part 
>This file and its rules are completely separated from logging to console and node.txt

OK lets add this functionality adding more capabilities to log definitely is good.

PR #298, opened 2017-08-08
getblock RPC command

netsfx:
Only these commits are relevant, please cherry-pick!

@netsfx	netsfx	getblock RPC command			0f0f589
 @netsfx	netsfx	Use block height + 1 to get the next block.			73f6670


dangershony:
I am not sure how to cherry pick (and it will get messy) so perhaps better is to just copy the code manually in a new PR?

PR #299, opened 2017-08-09
Bugfix/issue 247

### LINKED ISSUE #247: LookaheadBlockPuller - QualityScore function is too penalizing
Added a bunch of logs in my local source code, it revealed that the block puller wanted to get a block, the work was assigned to a peer that had 75 quality score. The node provided the block in about 2 seconds, but during this time it lost 40 % of its quality score because of "stalling". Then by finishing the task, it gained +1, so it went 75 -> 45 -> 46. The peer ended with much worse score than it started, while it did a pretty good job. 

Basically what happens with the current score function is that any peer that is assigned work that is not currently needed can gain score, while to be assigned with the task to download currently needed block (say it was assigned to a peer that disconnected and then reassigned to other peer) always ends with dramatic lowering of the score. In case the peer is assigned the current and couple of following blocks, it will almost surely be thrown away (stalling -> score down -> score zero -> all tasks released from the node).

It seems like this downloading strategy does not really reflect how good nodes are.

Aprogiena:
This is a complete rework of quality score assignment as per discussion in issue #247. It separates the logic of quality score calculation to its own class and adds couple of tests related to this new behavior.

I did make quite a lot of testing, but due to the complexity of the whole puller thing, it is **highly recommended for reviewers to run more tests for themselves** (just running the node over and over again and see what happens) as I don't feel that any number of tests here is sufficient.

Also please note that this code is the first one to take the advantage of NLog integration, so the puller is now heavily logged. If you use NLog.config, you could see extreme amounts of logs being created, slowing down the puller quite a lot. I've tested that in production, without NLog.config, those logs are not being used and the performance is excellent (I was able to sync ~75000 blocks in less than 4 minutes). 

Fix #247

### DIFF
--- a/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
+++ b/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
@@ -59,16 +59,26 @@ namespace Stratis.Bitcoin.BlockPulling
     /// </remarks>
     public abstract class BlockPuller : IBlockPuller
     {
-        /// <summary>Maximal quality score of a peer node based on the node's past experience with the peer node.</summary>
-        public const int MaxQualityScore = 150;
+        /// <summary>Description of a block together with its size.</summary>
+        public class DownloadedBlock
+        {
+            /// <summary>Size of the serialized block in bytes.</summary>
+            public int Length;
 
-        /// <summary>Minimal quality score of a peer node based on the node's past experience with the peer node.</summary>
-        public const int MinQualityScore = 1;
+            /// <summary>Description of a block.</summary>
+            public Block Block;
+        }
+
+        /// <summary>Number of historic samples we keep to calculate quality score stats from.</summary>
+        private const int QualityScoreHistoryLength = 100;
 
         /// <summary>Instance logger.</summary>
-        protected readonly ILogger logger;
+        private readonly ILogger logger;
 
-        /// <summary>Lock protecting access to <see cref="assignedBlockTasks"/>, <see cref="pendingInventoryVectors"/>, <see cref="downloadedBlocks"/>, and <see cref="peersPendingDownloads"/></summary>
+        /// <summary>
+        /// Lock protecting access to <see cref="assignedBlockTasks"/>, <see cref="pendingInventoryVectors"/>, <see cref="downloadedBlocks"/>, 
+        /// <see cref="peersPendingDownloads"/>, and <see cref="peerQuality"/>.
+        /// </summary>
         private readonly object lockObject = new object();
 
         /// <summary>
@@ -85,6 +95,10 @@ namespace Stratis.Bitcoin.BlockPulling
         /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
         private readonly Dictionary<uint256, DownloadedBlock> downloadedBlocks;
 
+        /// <summary>Statistics of the recent history of network peers qualities.</summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly QualityScore peerQuality;
+
         /// <summary>Number of items in <see cref="downloadedBlocks"/>. This is for statistical purposes only.</summary>
         public int DownloadedBlocksCount
         {
@@ -97,9 +111,9 @@ namespace Stratis.Bitcoin.BlockPulling
             }
         }
 
-        /// <summary>Sets of block header hashes that are being downloaded mapped by peers they are assigned to.</summary>
+        /// <summary>Sets of download tasks representing blocks that are being downloaded mapped by peers they are assigned to.</summary>
         /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
-        private readonly Dictionary<BlockPullerBehavior, HashSet<uint256>> peersPendingDownloads = new Dictionary<BlockPullerBehavior, HashSet<uint256>>();
+        private readonly Dictionary<BlockPullerBehavior, Dictionary<uint256, DownloadTask>> peersPendingDownloads = new Dictionary<BlockPullerBehavior, Dictionary<uint256, DownloadTask>>();
 
         /// <summary>Collection of available network peers.</summary>
         protected readonly IReadOnlyNodesCollection Nodes;
@@ -114,16 +128,6 @@ namespace Stratis.Bitcoin.BlockPulling
         private readonly NodeRequirement requirements;
         /// <summary>Specification of requirements the puller has on its peer nodes to consider asking them to provide blocks.</summary>
         public virtual NodeRequirement Requirements => this.requirements;
-
-        /// <summary>Description of a block together with its size.</summary>
-        public class DownloadedBlock
-        {
-            /// <summary>Size of the serialized block in bytes.</summary>
-            public int Length;
-
-            /// <summary>Description of a block.</summary>
-            public Block Block;
-        }
 
         /// <summary>
         /// Initializes a new instance of the object having a chain of block headers and a list of available nodes. 
@@ -140,6 +144,7 @@ namespace Stratis.Bitcoin.BlockPulling
             this.downloadedBlocks = new Dictionary<uint256, DownloadedBlock>();
             this.pendingInventoryVectors = new Queue<uint256>();
             this.assignedBlockTasks = new Dictionary<uint256, BlockPullerBehavior>();
+            this.peerQuality = new QualityScore(QualityScoreHistoryLength, loggerFactory);
 
             // set the default requirements
             this.requirements = new NodeRequirement
@@ -160,14 +165,14 @@ namespace Stratis.Bitcoin.BlockPulling
         /// <param name="cancellationToken">Cancellation token to be used by derived classes that allows the caller to cancel the execution of the operation.</param>
         public virtual void BlockPushed(uint256 blockHash, DownloadedBlock downloadedBlock, CancellationToken cancellationToken)
         {
-            this.logger.LogTrace($"({nameof(blockHash)}:'{blockHash}',{nameof(downloadedBlock.Length)}:{downloadedBlock.Length})");
+            this.logger.LogTrace($"({nameof(blockHash)}:'{blockHash}',{nameof(downloadedBlock)}.{nameof(downloadedBlock.Length)}:{downloadedBlock.Length})");
             this.logger.LogTrace("(-)");
         }
 
         /// <inheritdoc />
         public void InjectBlock(uint256 blockHash, DownloadedBlock downloadedBlock, CancellationToken cancellationToken)
         {
-            this.logger.LogTrace($"({nameof(blockHash)}:'{blockHash}',{nameof(downloadedBlock.Length)}:{downloadedBlock.Length})");
+            this.logger.LogTrace($"({nameof(blockHash)}:'{blockHash}',{nameof(downloadedBlock)}.{nameof(downloadedBlock.Length)}:{downloadedBlock.Length})");
 
             if (this.AddDownloadedBlock(blockHash, downloadedBlock))
               this.BlockPushed(blockHash, downloadedBlock, cancellationToken);
@@ -178,7 +183,7 @@ namespace Stratis.Bitcoin.BlockPulling
         /// <inheritdoc />
         public virtual void AskBlocks(ChainedBlock[] downloadRequests)
         {
-            this.logger.LogTrace($"({nameof(downloadRequests.Length)}:{downloadRequests.Length})");
+            this.logger.LogTrace($"({nameof(downloadRequests)}:{string.Join(",", downloadRequests.Select(r => r.Height))})");
 
             BlockPullerBehavior[] nodes = GetNodeBehaviors();
             if ((nodes.Length > 0) && (downloadRequests.Length > 0))
@@ -279,7 +284,7 @@ namespace Stratis.Bitcoin.BlockPulling
         /// <param name="chainedBlock">Block the node wanted to download, but something went wrong during the process.</param>
         protected void OnStalling(ChainedBlock chainedBlock)
         {
-            this.logger.LogTrace($"({nameof(chainedBlock.HashBlock)}:'{chainedBlock.HashBlock}')");
+            this.logger.LogTrace($"({nameof(chainedBlock)}.{nameof(chainedBlock.HashBlock)}:'{chainedBlock.HashBlock}')");
             BlockPullerBehavior behavior = null;
 
             lock (this.lockObject)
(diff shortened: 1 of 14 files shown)


Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -385,38 +428,46 @@ private Block NextBlockCore(CancellationToken cancellationToken)
...
                     Interlocked.Add(ref this.currentSize, -block.Length);
                     this.consumed.Set();
-                    return block.Block;
+
+                    res = block.Block;
                 }
                 else
                 {
-                    // Otherwise we either have reorg.
+                    // Otherwise we either have reorg, or we don't know the next block.
```
dangershony: I would say here
"Otherwise we either have reorg, or we reached the best chain tip.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -385,38 +428,46 @@ private Block NextBlockCore(CancellationToken cancellationToken)
...
                         // Or the block is still being downloaded or we need to ask for this block to be downloaded.
                         if (!isDownloading) AskBlocks(new ChainedBlock[] { header });
 
                         OnStalling(header);
                         this.IsStalling = true;
                     }
-                    WaitHandle.WaitAny(new[] { this.pushed, cancellationToken.WaitHandle }, waitTime[i]);
-                    i = i == waitTime.Length - 1 ? 0 : i + 1;
+
+                    WaitHandle.WaitAny(new[] { this.pushed, cancellationToken.WaitHandle }, WaitNextBlockRoundTimeMs);
```
dangershony: I would suggest we have two types of delays one during IBD and the other if we are at the tip 
(this can be know based on the else of this code ```if (!this.Chain.Contains(this.location.HashBlock))```.

IBD we keep it at 100 ms but if the chain is at the TIP we can potentially wait longer 500 ms (now blocks will arrive in the range a 1-10 min)

Only issue with this might be block propagation, it means a block will be delayed up to 500 ms longer before we ask it validate it and broadcast it.

Stratis.Bitcoin/BlockPulling/PullerDownloadAssignments.cs
```diff
@@ -20,19 +18,42 @@ public static class PullerDownloadAssignments
...
 
+        /// <summary>Class logger.</summary>
+        private static readonly ILogger logger;
+
+        /// <summary>Number of blocks from the currently last block that are protected from being assigned to poor peers.</summary>
+        /// <seealso cref="AssignBlocksToPeers"/>
+        private const int CriticalLookahead = 10;
+
+        /// <summary>Peer is considered to be assigned a large amount of work if it has more than this amount of download tasks assigned.</summary>
+        private const int HighWorkAmountThreshold = 50;
```
dangershony: Wowoo nice idea

Stratis.Bitcoin/BlockPulling/PullerDownloadAssignments.cs
```diff
@@ -58,17 +91,43 @@ public class PeerInformation
...
+                }
+
+                // Modify scores based on amount of work peers already have.
+                // The peer may get up to 50 % penalty to its score if it has too much work on itself.
+                int totalWork = filteredPeers.Sum(p => p.TasksAssignedCount);
+                for (int i = 0; i < filteredPeers.Count; i++)
+                {
+                    if (filteredPeers[i].TasksAssignedCount > HighWorkAmountThreshold)
+                    {
+                        double penaltyCoef = 1 + ((filteredPeers[i].TasksAssignedCount * filteredPeers[i].TasksAssignedCount) / (totalWork * totalWork));
```
dangershony:
```
   var peerTotalTasks = filteredPeers[i].TasksAssignedCount
   if (peerTotalTasks  > HighWorkAmountThreshold)
   {
      double penaltyCoef = 1 + ((peerTotalTasks * peerTotalTasks ) / (totalWork * totalWork));
```

Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                    if (this.peerReferenceCounter[oldSample.peer] == 0)
+                        this.peerReferenceCounter.Remove(oldSample.peer);
+                }
+                else this.samplesCount++;
+
+                // Add new sample to the mix.
+                this.samples[this.samplesIndex].timePerKb = timePerKb;
+                this.samples[this.samplesIndex].peer = peer;
+                this.samplesIndex = (this.samplesIndex + 1) % this.samples.Length;
```
dangershony: Ah this will reset back to zero so you start from the beginning of the samples array.
I wonder why not use a Queue and just pop the last entry, you will save a lot of logic?

Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                // If we reached the maximum number of samples, we need to remove oldest sample.
+                if (this.samplesCount == this.samples.Length)
+                {
+                    PeerSample oldSample = this.samples[this.samplesIndex];
+                    this.samplesSum -= oldSample.timePerKb;
+                    this.peerReferenceCounter[oldSample.peer]--;
+
+                    if (this.peerReferenceCounter[oldSample.peer] == 0)
+                        this.peerReferenceCounter.Remove(oldSample.peer);
```
dangershony: I think you forgot this.samplesCount = 0

dangershony:
As we are not in production yet, I would like to have this checked in first thing, the more it is in the wild the more feedback we get.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -97,9 +107,9 @@ public int DownloadedBlocksCount
             }
         }
 
-        /// <summary>Sets of block header hashes that are being downloaded mapped by peers they are assigned to.</summary>
+        /// <summary>Sets of download tasks representing blocks that are being downloaded mapped by peers they are assigned to.</summary>
         /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
-        private readonly Dictionary<BlockPullerBehavior, HashSet<uint256>> peersPendingDownloads = new Dictionary<BlockPullerBehavior, HashSet<uint256>>();
+        private readonly Dictionary<BlockPullerBehavior, Dictionary<uint256, DownloadTask>> peersPendingDownloads = new Dictionary<BlockPullerBehavior, Dictionary<uint256, DownloadTask>>();
```
Aprogiena: Hmm that's a pity, I think it should not affect our ability to use this word. But if you are strong on this, I'll change it to DownloadRequest or DownloadAssignment. It is not DownloadMessage though.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -160,29 +161,40 @@ protected BlockPuller(ConcurrentChain chain, IReadOnlyNodesCollection nodes, Pro
...
+            this.logger.LogTrace($"({nameof(downloadRequests)}:{string.Join(",", downloadRequests.Select(r => r.Height))})");
 
-            Dictionary<int, InventoryVector> vectors = new Dictionary<int, InventoryVector>();
-            foreach (ChainedBlock request in downloadRequests)
+            BlockPullerBehavior[] nodes = this.GetNodeBehaviors();
+            if ((nodes.Length > 0) && (downloadRequests.Length > 0))
             {
-                InventoryVector vector = new InventoryVector(InventoryType.MSG_BLOCK, request.HashBlock);
-                vectors.Add(request.Height, vector);
+                Dictionary<int, InventoryVector> vectors = new Dictionary<int, InventoryVector>();
```
Aprogiena: yes, that is my faul on code style rule no. 10  (although strictly speaking not really, but yeah, i'll fix it)

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -266,20 +290,23 @@ protected void OnStalling(ChainedBlock chainedBlock)
 
             if (behavior != null)
             {
-                behavior.QualityScore = Math.Max(MinQualityScore, behavior.QualityScore - 1);
-                if (behavior.QualityScore == MinQualityScore)
+                double penalty = this.peerQuality.CalculateNextBlockTimeoutQualityPenalty();
```
Aprogiena: yes I think this is good think in general - the original design had this and I think it was good decision exactly for the reason you've mentioned - the very next block is somehow special and it is easy to imagine an edge case when we could be completely stuck without it

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -92,27 +102,40 @@ public override object Clone()
...
-                    tx.CacheHashes();
-
+                // There are two pullers for each peer connection and each is having its own puller behavior.
+                // Both these behaviors get notification from the node when it receives a message,
+                // even if the origin of the message was from the other puller behavior.
+                // Therefore we first make a quick check whether this puller behavior was the one 
+                // who should deal with this block.
                 uint256 blockHash = block.Object.Header.GetHash();
-                DownloadedBlock downloadedBlock = new DownloadedBlock()
+                if (this.puller.CheckBlockTaskAssignment(this, blockHash))
```
Aprogiena: I am actually strongly convinced here that this is better. The situation you are describing takes about 0.1 milliseconds. We have usually 8 peers max, but even if we had 100 peers, the chance of running to each other within that 0.1 millisecond is nothing more than plain zero :-)

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: I also had DownloadAssignment in mind yeah, not too strong on it up to you.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -121,13 +144,20 @@ private void Node_MessageReceived(Node node, IncomingMessage message)
         /// </summary>
         internal void AssignPendingVector()
         {
+            this.logger.LogTrace($"[{this.GetHashCode():x}] ()");
+
             Node attachedNode = this.AttachedNode;
             if (attachedNode == null || attachedNode.State != NodeState.HandShaked || !this.puller.Requirements.Check(attachedNode.PeerVersion))
+            {
+                this.logger.LogTrace($"[{this.GetHashCode():x}] (-)[ATTACHED_NODE]");
```
Aprogiena: it is just to distinguish different exit from the method, `(-)` says we are existing the method and `[XYZ]` identifies which exit this is (standard exit is without any id), so reading the log seeing this line, you will know that the execution flow exited here because of this if

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: yup, will fix

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: I understand why you are asking for this - to avoid useless work ("are we there yet polling") to be done every so often, but as I look in the code, those checks are all O(1), i.e. we are talking about sub millisecond work again, which should be perfectly fine doing every 100ms, especially when the chain is synced already (i.e. not much work is actually in progress elsewhere)

this is why I would stay with the current simpler design - i don't think the change is worth it

Stratis.Bitcoin/BlockPulling/PullerDownloadAssignments.cs
Aprogiena: ok, no problem with that, will fix

Stratis.Bitcoin/BlockPulling/QualityScore.cs
Aprogiena: I would save a lot of logic in this code, but the underlying logic would be even worse. This is very simple circular array - once allocated it never changes and does no more allocations. This should be super efficient (although we are talking pennies again)

Aprogiena:
Will fix very soon

Stratis.Bitcoin/BlockPulling/QualityScore.cs
Aprogiena: I hope I did not - take a look couple of lines in front - it is 
```
                // If we reached the maximum number of samples, we need to remove oldest sample.
                if (this.samplesCount == this.samples.Length)
```

so here we are sure that we have all slots occupied

or did you mean something else?

Aprogiena:
I've fixes those things where we agreed, if there is something more to be done, please let me know

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
dangershony: Ah fair enough the chance is quite low.

You still mix locking logic between the puller and the behavior (which anyway happens but we can try to minimize it).

In any case I just suggested a second approach no need to make any changes.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
dangershony: Well we still consider chain as the tip, or at least the tip we know about, in reality if we are in IBD it means we didn't connect to a peer with a longer tip (yet).

As the chain never stops you may be right that we don't know the next block but that's because it wasn't mined yet.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
dangershony: Ok yeah true once not in IBD the node puller wont be so busy (but mempool might be hammered very hard)

But I don't see any complexity, just a simple flag on the if statement above.

```
WaitHandle.WaitAny(new[] { this.pushed, cancellationToken.WaitHandle }, waitIBD ?
 WaitNextBlockRoundIBDTimeMs : WaitNextBlockRoundTimeMs );
```

I am not going to insist thought because I am not sure myself if it will effect propagation.

Stratis.Bitcoin/BlockPulling/QualityScore.cs
dangershony: Ah no I didn't realize this will happen only once when the array is full.

Stratis.Bitcoin/BlockPulling/QualityScore.cs
dangershony: Yeah indeed a queue would be constantly allocating new memory.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: ok, if we consider that to be the tip then it makes sense, i'll fix that

Aprogiena:
OK last fix is there, hopefully we now agree on everything as is

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: Well, I agree I hate the fact that we can get the notification here in the first place for the behavior that has nothing to do with this block. So I'm not thrilled about the design, but to me it seems like the best solution for what we have. 

So I leave this one as is.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
Aprogiena: Yeah, I can't see the benefit. Leaving as is.

Stratis.Bitcoin/BlockPulling/QualityScore.cs
dangershony: I had some more thoughts on this and i think you are violating your own rule here of keeping code simple, think of it like this an avarage dev should know how the Queue implementation works, reading your code rquires to spend some more time to understand it, if you put a queue its immediately clear. 

You can even use concurrent queue as you dont have logic that requiers many steps inside a lock.

Aprogiena: I can add more comments, but I'm not considering inline circular array as an advanced construct. If it seems complex to you I can implement circular array as a separated class.

dangershony: once I know what the code does its not complicated at all, its just that I had to spend time on it where a queue queue would be immediate.

yes I think that a good idea to implement a circular array class (is ther enot one out of the box in the framework?) then a read will have it very easy to read the high level code.

Aprogiena: yes, i am happy to make it separate class
for me it was inside the boundaries of what could get inline, for you it is outside, so we put it separately and that's fine

PR #303, opened 2017-08-10
Validate feature dependencies on feature startup

mikedennis:
Add a hook for a dependency check when running node feature executor. Throw exception if a feature's dependency is missing.

Fix for issue #285

### DIFF
--- a/Stratis.Bitcoin/Builder/Feature/FeaturesExtensions.cs
+++ b/Stratis.Bitcoin/Builder/Feature/FeaturesExtensions.cs
@@ -9,15 +9,15 @@ namespace Stratis.Bitcoin.Builder.Feature
     public static class FeaturesExtensions
     {
         /// <summary>
-        /// Ensures a dependency feature type is present in the feature registration.
+        /// Ensures a dependency feature type is present in the feature list.
         /// </summary>
         /// <typeparam name="T">The dependency feature type.</typeparam>
-        /// <param name="features">List of feature registrations.</param>
-        /// <returns>List of feature registrations.</returns>
+        /// <param name="features">List of features.</param>
+        /// <returns>List of features.</returns>
         /// <exception cref="MissingDependencyException">Thrown if feature type is missing.</exception>
         public static IEnumerable<IFullNodeFeature> EnsureFeature<T>(this IEnumerable<IFullNodeFeature> features)
         {
-            if (!features.Any(i => i.GetType() == typeof(T)))
+            if (!features.OfType<T>().Any())
             {
                 throw new MissingDependencyException($"Dependency feature {typeof(T)} cannot be found.");
             }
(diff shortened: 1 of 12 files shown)


Stratis.Bitcoin/Builder/Feature/FeaturesExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// Ensures a dependency feature type is present in the feature registration.
+        /// </summary>
+        /// <typeparam name="T">The dependency feature type.</typeparam>
+        /// <param name="features">List of feature registrations.</param>
+        /// <returns>List of feature registrations.</returns>
+        /// <exception cref="MissingDependencyException">Thrown if feature type is missing.</exception>
+        public static IEnumerable<IFullNodeFeature> EnsureFeature<T>(this IEnumerable<IFullNodeFeature> features)
+        {
+            if (!features.Any(i => i.GetType() == typeof(T)))
```
dangershony: Can you not do?
```
!features.OfType<T>().Any()
```

Stratis.Bitcoin/Features/Miner/MiningFeature.cs
```diff
@@ -1,56 +1,66 @@
...
+        ///<inheritdoc />
+        public override void Stop()
+        {
+        }
 
-		public override void Stop()
-		{
-		}
-	}
+        ///<inheritdoc />
```
dangershony: Excellent this looks very nice.

Stratis.StratisD/Program.cs
```diff
@@ -32,6 +33,7 @@ public static void Main(string[] args)
                 .UseStratisConsensus()
                 .UseBlockStore()
                 .UseMempool()
+                .UseWallet()
```
dangershony: Hmm, I am not sure we need wallet in the daemon.

mikedennis: The daemon currently is built with .AddPowPosMining() 
which has the dependency on the wallet manager. Should we remove PowPosMining then? Or have another extension method that adds the wallet dependencies that mining needs?

Stratis.Bitcoin/Builder/Feature/FeaturesExtensions.cs
mikedennis: Sure - will give that a try!

Stratis.StratisD/Program.cs
dangershony: ah right yes so keep it

mikedennis:
Ok, I've fixed up the type checking code. I like that much better, thanks for the suggestion!

I see that there is a Stratis.Bitcoin.Features.LightWallet feature now that Jeremy has been working on.  I wonder whether it has enough of the wallet capability for the Miner feature? I see it does include the WalletManager. Potentially we could use that instead of the full blown Wallet? Let me know if you think this is worthwhile investigating and I can create another issue for it and investigate it.

PR #305, opened 2017-08-10
Bugfix/issue 246

### LINKED ISSUE #246: Potential race condition in BlockPuller AskBlocks->GetNodeBehaviors->DistributeDownload
I'm not sure about this one, but will add it anyway. So if you can argue why this can not happen, please do.

BlockPuller.AskBlocks calls BlockPuller.GetNodeBehaviors, which selects candidate nodes to assign work to. It seems to me that it can happen that another thread will call Node.Cleanup when the peer node disconnects. In that case the Nodes collection that BlockPuller.GetNodeBehaviors works with might be unupdated until BlockPullerBehavior.DetachCore is called, so BlockPuller.AskBlocks can get a list of nodes including the one that just disconnected.

Then let's say BlockPuller.AskBlocks calls BlockPuller.DistributeDownload and just before it modifies BlockPuller.map, the BlockPullerBehavior.DetachCore can be called in the other thread, which would release all the work assigned to this freshly disconnected node. Then BlockPuller.DistributeDownload would continue and modify the map. 

This would end with a work assigned to a no longer existing node, which would lead to puller stalling once the assigned block would be requested.

Aprogiena:
Introducing Disconnected flag for puller behavior, which prevents the puller to get more work once the flag is set. This is checked and set in locked environment, so the race is mitigated. 

Also minor refactoring of DistributeDownload to make the window of the race narrower. The race is prevented by lock, but narrowing the window helps reducing the number of occurrences  in which the lock is needed.

This is likely to have conflicts with my other PR, so let's merge one first and fix the other accordingly.

Fix #246


Aprogiena:
OK, it was VERY bad idea to work on this separately of issue-246 fix, as the conflict was much worse than expected. This created a huge mess and I just hope I've somehow resolved it :/

PR #308, opened 2017-08-12
Feature/prefix logger with console support

### LINKED ISSUE #292: [Logging] Implement prefix logger
We have migrated to NLog, but there is more we can do about making logging even more useful. I call this feature "prefix logger". All it does is that it allows you to do something like this:

`this.logger = loggerFactory.CreateLogger(this.GetType().FullName, PREFIX);`

and the effect is that all logs you create with it are prefixed with PREFIX. So let's say we have BlockPullerBehavior and for logging purposes, we identify each behavior with behavior.GetHashCode(). 

Currently I need to write something like this:

`this.logger.LogTrace($"[{this.GetHashCode():x}] ()");`

and what I would like instead is this:

```
this.logger = loggerFactory.CreateLogger(this.GetType().FullName, "[{this.GetHashCode():x}] ");
...
this.logger.LogTrace("()");
```

This is useful for logging events of instances of objects where you have multiple instances and you wnat to know which instance invoked the event. It saves you from repeating the identifier in every log call.

Aprogiena:
Implementation of prefix logger with console support, which allows you to create logger that puts a prefix in front of every log.

As a part of this issue we improve logging in BlockPullerBehavior which now uses the prefix logger.

**IMPORTANT**: Note that I've made a competitive pull request https://github.com/stratisproject/StratisBitcoinFullNode/pull/307 which implements prefix logger WITHOUT console support. The reason for that is that the console support is not very elegant and I'm not sure how to make it better at this moment. Only one of these pull requests can be accepted, the other one should be discarded. I'm still pretty much convinced that console output should be separated from logging completely. The other version without console support would thus be better for that. This version is better if we want to continue mixing logging and console output.

NOTE: The reason why this prefix logger is not implemented "trivially" (i.e. using ILogger) is that such an approach violates `${callsite}` in NLog. NLog needs to be informed about the wrapper so that it can skip it on the call stack. This can't be done with just ILogger and hence we have to create NLog specific logger for this.

Fix #292


dangershony:
I don't think prefix on console is very useful, prefix is easier for complex logging scenarios

Aprogiena:
I might not explained that well. In the component, you will either use a normal logger or a prefix logger. Now if there is a component that will use prefix logger then without this PR merged, you will be completely unable to output to console from this component - i.e. even if you do LogInformation it will not go to the console. 

It is true that probably no such component exists now - a component that would use prefix logger and wanted to have console output.

dangershony:
Ah I see, why this limitation? so is having a prefix really beneficial?
I am getting feeling the NLog framework is getting too messy, I would opt to reverting back to adding a prefix at each log output.

Aprogiena:
It is described in the first message of this PR, it is only messy(ier than it could) because you are using ILogger interface, which is a very simple interface, not very powerful. NLog allows you to have something called CALLSITE (https://github.com/NLog/NLog/wiki/Callsite-layout-renderer), which is handy because instead of writing
```
logger.LogTrace("MethodName()");
```
you can just write 
```
logger.LogTrace("()");
```
and by using CALLSITE in the logging target layout, you've got it for free. And so you have the prefix with prefix logger. The only problem (the messiness) is when you want to use ILogger. Then you need to implement support for that. If you don't you will lose what ILogger does - having console logger and file logger on the same interface. 

So this PR does exactly that - it not only implement prefix logger for NLog, but it also implements support for ILogger, so when you use it, it behaves exactly as you would expect (logging on INFO+ level goes to console). 

The other PR only supports NLog, so there you lose this logging to console.

The other PR is better if you want to have clean design - i.e. separated logging from console output (this should be separated, but for some reason you guys prefer to have it mixed). 

This PR is better if the mixing of console and logging sounds good to you, that's why I implemented it (to avoid further pushing for console output and logging separation).

So the decision should be like this:

Do we want the old way of doing this (i.e. we like mixing console output and logging)?
YES -> use this PR
NO -> use the other PR

dangershony:
AS we discussed can we close this?

Aprogiena:
I will close it with a pull request.

Aprogiena:
Yes, this will be void and I will make a new one.

PR #315, opened 2017-08-14
Refactor/stratis.bitcoin.signal issue 282

### LINKED ISSUE #282: Rework how Stratis.Bitcoin.Signals are used
Consider removing ISignals.Blocks and ISignals.Transactions from the interface and make them private in Signals. This will require Subscribe method to be implemented.

Reason: All code is currently using something like this:
`this.signals.Blocks.Broadcast(block.Block);`

instead of 

`this.signals.Signal(block.Block);`

and both `[Blocks|Transactions].Broadcast` implement guards and then `.Signal()` methods also implement guards. This is probably not used as intended.

Aprogiena:
Unification of how Signals are used. This required changing code in a lot of places, but also lead to removal of couple of classes in Wallet. Therefore I'd like @bokobza on the review.

Fix #282


dangershony:
Looks good can you resolve conflicts and we push it.

Aprogiena:
Easier said than done though, you've done some heavylifting with the wallet reorg test fix.

Aprogiena:
OK should be good to go, tests passed +-

kalani1234:
Join US in our new private Telegram channel which will show you how to make real money from the Crypto Markets!
We are not a Pump and Dump channel and we do not support them as most members in there lose money. We are limited to 5000 members and after this the invite link will expire.
All details will be shared inside the channel 
Join us here: https://t.me/joinchat/AAAAAEWfPN6QN9-wVIukIA

PR #325, opened 2017-08-15
Add xmldoc for some memory pool classes

mikedennis:
No functionality change just added comments and positioning of methods inside of classes. More to come in a future PR.

Not sure why my commits from Aug 14 show up here. They were in a separate PR that was merged into the repository earlier. Maybe something to do with the way they were merged in or some bad git  branching on my part?


Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -107,9 +233,15 @@ protected override void DetachCore()
...
-            
         }
 
+        /// <summary>
+        /// Asynchronous processing of node message.
+        /// Handles the following message payloads: TxPayload, MempoolPayload, GetDataPayload, InvPayload.
+        /// </summary>
+        /// <param name="node">Node sending the message.</param>
+        /// <param name="message">Incoming message.</param>
+        /// <returns></returns>
```
Aprogiena: You can omit `<returns>` if the function returns `void` or `Task` (but not `Task<T>`)

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -131,6 +263,73 @@ private Task AttachedNode_MessageReceivedAsync(Node node, IncomingMessage messag
...
+                }
+                return ret;
+            });
+
+            await this.SendAsTxInventory(node, sends.Select(s => s.Trx.GetHash()));
+            this.LastMempoolReq = this.manager.DateTimeProvider.GetTime();
+        }
+
+        /// <summary>
+        /// Asynchronous processing of inventory payload from the node.
```
Aprogiena: The word "Asynchronous" can be omitted here as the method is `async` and its name ends with `Async`.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -176,6 +375,13 @@ private Task AttachedNode_MessageReceivedAsync(Node node, IncomingMessage messag
                 await node.SendMessageAsync(send).ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Asynchronous processing of the get data payload message from node.
+        /// Sends the memory pool transaction info via TxPayload to the attached node.
+        /// </summary>
+        /// <param name="node">Node sending the message</param>
```
Aprogiena: missing `.`

Stratis.Bitcoin/Features/MemoryPool/MempoolFeature.cs
```diff
@@ -8,15 +8,57 @@
 
 namespace Stratis.Bitcoin.Features.MemoryPool
 {
+    /// <summary>
+    /// Transaction memory pool feature for the Full Node.
+    /// <see cref="https://github.com/bitcoin/bitcoin/blob/6dbcc74a0e0a7d45d20b03bb4eb41a027397a21d/src/txmempool.cpp"/>
```
Aprogiena: This should be `<seealso>` outside summary. `<see>` is inline construct and is used slightly differently.

Stratis.Bitcoin/Features/MemoryPool/MempoolFeature.cs
```diff
@@ -59,13 +106,20 @@ public override void Stop()
...
     /// <summary>
     /// A class providing extension methods for <see cref="IFullNodeBuilder"/>.
     /// </summary>
     public static partial class IFullNodeBuilderExtensions
     {
+        /// <summary>
+        /// Include the memory pool feature and related services in the full node.
+        /// </summary>
+        /// <param name="fullNodeBuilder">full node builder</param>
+        /// <returns>full node builder</returns>
```
Aprogiena: SENTENCES rule.

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
```diff
@@ -9,23 +9,49 @@
 
 namespace Stratis.Bitcoin.Features.MemoryPool
 {
+    /// <summary>
+    /// Memory pool scheduler.
+    /// </summary>
     public class MempoolScheduler : AsyncLock
```
Aprogiena: This is unfortunate naming, this `MempoolScheduler` is in fact just a way to provide exclusivity for operation (something like lock) and only internally it uses a scheduler. But from the memory pool point of view it is not a scheduler. This should probably be renamed to `MempoolLock` or something like that.

This PR is about the comment, so the comment here is somehow misleading because it should rather say that it is a lock not scheduler.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -13,30 +13,86 @@
...
+        /// <summary>
+        /// Maximum number of inventory items to send per transmission.
+        /// Limits the impact of low-fee transaction floods.
+        /// </summary>
         const int INVENTORY_BROADCAST_MAX = 7*INVENTORY_BROADCAST_INTERVAL;
 
+        /// <summary>
+        /// Memory pool validator injected dependency.
+        /// </summary>
         private readonly IMempoolValidator validator;
```
Aprogiena: This is just my opinion, no requirement. I tend to write these one line comments of properties as one liners including the summary tag:

```
<summary>Memory pool validator injected dependency.</summary>
```

I only expand to separated lines for the tags if the comment inside is multiline. But I use it only for fields and properties.

So this is just an idea to consider. No need for change if you prefer this multiline format.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -13,30 +13,86 @@
...
     {
-        // Average delay between trickled inventory transmissions in seconds.
-        //  Blocks and whitelisted receivers bypass this, outbound peers get half this delay. 
+        #region Fields
+
+        /// <summary>
+        /// Average delay between trickled inventory transmissions in seconds.
+        /// Blocks and whitelisted receivers bypass this, outbound peers get half this delay. 
+        /// </summary>
         const int INVENTORY_BROADCAST_INTERVAL = 5;
```
Aprogiena: Feel free to fix code style here (and below) as a part of this PR. Access modifier is missing here and PascalCase should be used.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -131,6 +263,73 @@ private Task AttachedNode_MessageReceivedAsync(Node node, IncomingMessage messag
...
+            //}
+
+            //if (connman.OutboundTargetReached(false) && !pfrom->fWhitelisted)
+            //{
+            //	LogPrint("net", "mempool request with bandwidth limit reached, disconnect peer=%d\n", pfrom->GetId());
+            //	pfrom->fDisconnect = true;
+            //	return true;
+            //}
+
+            var vtxinfo = await this.manager.InfoAllAsync();
```
Aprogiena: feel free to fix code style as well here - i.e. remove var

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
```diff
@@ -9,23 +9,49 @@
...
     { }
 
+    /// <summary>
+    /// Memory pool manager.
+    /// </summary>
```
Aprogiena: I would appreciate long winded description about the component here. If I'm not familiar with memory pool and I read this file I have no idea what is "memory pool manager" - i.e. what is its responsibility, what it manages, what is outside of its scope. Some kind of bigger picture description would be nice about how this component fits into the architecture.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -13,30 +13,86 @@
...
 
+        /// <summary>
+        /// Memory pool validator injected dependency.
+        /// </summary>
         private readonly IMempoolValidator validator;
+
+        /// <summary>
+        /// Memory pool manager injected dependency.
+        /// </summary>
         private readonly MempoolManager manager;
```
Aprogiena: So, I was thinking about this "XYZ injected dependency" comment and I found out that I don't like it, sorry :-) 
It is true that it is XYZ injected dependency, but that is not very useful information. That tells me how the property was set (and I somehow do have that information from `readonly` keyword already). But mostly not really interested to learn how the property was set. Then it tells me that it is memory pool manager, which I +- get from the type.

I know this is hard in many cases, but it seems like you have a pattern here for all such fields even for those where it is not so hard to produce a better comment. So I would say that if a better comment is very hard to invent then OK, use this as a workaround, but otherwise, I would prefer a better comment.

What is better comment? I would be looking for an information telling me what is this in context of the class. So why do I need it here? What kind of feature it provides to me? 

I think this kind of comment is not good especially in combination with the trivial comment on `MempoolManager` class - so OK, let's say it is hard here to produce a better comment (I think it is not, but just to make the point), so I would expect that if I go to `MempoolManager` class I would see what it is, but there I read just "Memory pool manager". And at this point I'm lost and I'm forced to analyze the code.

Stratis.Bitcoin/Features/MemoryPool/MempoolFeature.cs
dangershony: Ahh good to know to ```<see>``` is for inside summery?

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
dangershony: Indeed MempoolAsyncLock would be more appropriate.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
dangershony: I agree injected dependency is not so helpful, and hopefully everything will be injected. 
More explanation  is probably better.
Example
>The mempool manager contains high level methods that can be used from outside of the mempool.

Stratis.Bitcoin/Features/MemoryPool/MempoolFeature.cs
Aprogiena: `<see>` is used anywhere as inline reference and the tag is replaced by the referenced name in the text.

Example:

```
<summary>
This method compares <see cref="Apples" /> and <see cref="Oranges" />
</summary>
```

Intellisense will give you "This method compares Apples and Oranges." 
And in MSVS you are able to hit F12 on "Apples" (and "Oranges" as well) to go to those classes.

`<see>` can also be used as follows:

```
Bla bla bla bla - see `<see cref="..." /> for more information.
```

whereas `<seealso>` creates "see also" section as can be seen e.g. here https://msdn.microsoft.com/en-us/library/hh229816.aspx
which is a separate section at the end of the documentation page (if we generated such pages, which we can)

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
mikedennis: I will rename it. What do you think about pulling it out into a separate .cs file? At some places I've worked they had a rule about only one class per source file. Personally I think it's best to be in a separate source file if it's consumed anywhere outside of that source file.

mikedennis:
Thanks for the feedback guys - will look at addressing these issues.

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
Aprogiena: I don't mind such a smallish class to be included in the file with something else. I don't mind if you separate such class for whatever reason as well. 

As for interfaces, again I don't mind either. Lot of code has interface in the file with implementation. This makes sense to me if the default implementation is the only implementation. But again, I won't mind separation either.

However, such separation changes are definitely not part of XMLDOC task. So if you want to do this, pls create new tasks.

Stratis.Bitcoin/Features/MemoryPool/MempoolFeature.cs
dangershony: Ah yes, we should generate documentation pages for sure (at some point).

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
dangershony: I guess sits up to the dev, I like to have smaller connected classes in the same file, but I suspect with time people will pull every thing to its own class.

mikedennis:
Ok I think I got the suggestion mentioned previously fixed. That's the good news, the bad news is I think I should have done the MempoolScheduler renaming and/or property renaming in separate PRs. The renames ended up effecting a few other modules so this PR has now grown in size. In future I will create a separate task for renaming related items.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
```diff
@@ -70,16 +124,70 @@ public class MempoolBehavior : NodeBehavior
...
+
+        #endregion
+
+        #region Message Handlers
+
+        /// <summary>
+        /// Handler for processing incoming message from node.
+        /// </summary>
+        /// <param name="node">Node sending the message.</param>
+        /// <param name="message">Incoming message.</param>
```
dangershony: Can you add a  
```
   <remarks>
     TODO: Fix the exception handling of the async event.
   </remarks>
```

dangershony: It can be in a later commit doesnt have to be now.

dangershony:
Ok this looks very good, ofcourse the mempool has lots more to refactor and change but this is a good start.

Stratis.Bitcoin/Features/MemoryPool/MempoolBehavior.cs
mikedennis: For sure. Just added it, it will be in my next xml doc PR.

PR #326, opened 2017-08-15
WalletManager refactoring part 1

bokobza:
* Moved the IWalletManager.CreateNewAccount method to the Wallet object and renamed it to AddNewAccount to better reflect this method's functionalities. 
I started with doing this but then instead of moving further along I worked on extracting HD-related methods to a separate class.
* Extracted all the methods relating to HD wallets' addresses and accounts to a separate class.
* Updated NStratis to the latest submodule

### DIFF
--- a/Stratis.Bitcoin/Features/Wallet/HdOperations.cs
+++ b/Stratis.Bitcoin/Features/Wallet/HdOperations.cs
@@ -1,0 +1,211 @@
+﻿using System;
+using NBitcoin;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Wallet
+{
+    /// <summary>
+    /// Class providing helper methods for working with Hierarchical Deterministic (HD) wallets.
+    /// </summary>
+    /// <seealso cref="https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki" />
+    public class HdOperations
+    {
+        /// <summary>
+        /// Generates an HD public key derived from an extended public key.
+        /// </summary>
+        /// <param name="accountExtPubKey">The extended public key used to generate child keys.</param>
+        /// <param name="index">The index of the child key to generate.</param>
+        /// <param name="isChange">A value indicating whether the public key to generate corresponds to a change address.</param>
+        /// <returns>
+        /// An HD public key derived from an extended public key.
+        /// </returns>
+        public static PubKey GeneratePublicKey(string accountExtPubKey, int index, bool isChange)
+        {
+            Guard.NotEmpty(accountExtPubKey, nameof(accountExtPubKey));
+
+            int change = isChange ? 1 : 0;
+            KeyPath keyPath = new KeyPath($"{change}/{index}");
+            ExtPubKey extPubKey = ExtPubKey.Parse(accountExtPubKey).Derive(keyPath);
+            return extPubKey.PubKey;
+        }
+
+        /// <summary>
+        /// Gets the extended private key for an account.
+        /// </summary>
+        /// <param name="privateKey">The private key from which to generate the extended private key.</param>
+        /// <param name="chainCode">The chain code used in creating the extended private key.</param>
+        /// <param name="hdPath">The HD path of the account for which to get the extended private key.</param>
+        /// <param name="network">The network for which to generate this extended private key.</param>
+        /// <returns></returns>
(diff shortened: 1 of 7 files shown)


Stratis.Bitcoin/Features/Wallet/Wallet.cs
```diff
@@ -107,6 +107,23 @@ public IEnumerable<Script> GetAllPubKeysByCoinType(CoinType coinType)
                 yield return address.ScriptPubKey;
             }
         }
+
+        /// <summary>
+        /// Adds an account to the current wallet.
+        /// </summary>
+        /// <remarks>The name given to the account is of the form "account (i)" by default, where (i) is an incremental index starting at 0.
```
dangershony: <remarks> in new line is easier no?

Stratis.Bitcoin/Features/Wallet/Wallet.cs
```diff
@@ -107,6 +107,23 @@ public IEnumerable<Script> GetAllPubKeysByCoinType(CoinType coinType)
...
+        /// According to BIP44, an account at index (i) can only be created when the account at index (i - 1) contains at least one transaction.
```
dangershony: Maybe add to the comment that this method assumes that the BIP44 requirement of account i-1 has transactions

Stratis.Bitcoin/Features/Wallet/Wallet.cs
```diff
@@ -180,6 +197,57 @@ public HdAccount GetAccountByName(string accountName)
...
+        /// <param name="chainCode">The chain code for this wallet.</param>
+        /// <param name="network">The network for which this account will be created.</param>
+        /// <returns>A new hd account.</returns>
+        public HdAccount AddNewAccount(string password, string encryptedSeed, byte[] chainCode, Network network)
+        {
+            Guard.NotEmpty(password, nameof(password));
+            Guard.NotEmpty(encryptedSeed, nameof(encryptedSeed));
+            Guard.NotNull(chainCode, nameof(chainCode));
+
+            // Get the accounts for this type of coin.
```
dangershony: Is the comment correct?

Stratis.Bitcoin/Features/Wallet/HdOperations.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            Guard.NotNull(privateKey, nameof(privateKey));
+            Guard.NotNull(chainCode, nameof(chainCode));
+
+            var accountHdPath = GetAccountHdPath(coinType, accountIndex);
+            return GetExtendedPublicKey(privateKey, chainCode, accountHdPath);
+        }
+
+        /// <summary>
+        /// Gets the extended public key for an account.
```
dangershony: >Gets the extended public key for an HD path 

Is more accurate no?

Stratis.Bitcoin/Features/Wallet/Wallet.cs
bokobza: you mean like this:
<remarks>
The name given to the account is of the form "account (i)" by default, where (i) is an incremental index starting at 0.
</remarks>

Stratis.Bitcoin/Features/Wallet/Wallet.cs
bokobza: it doesn't assume it actually, the method called from this method checks whether the account has transactions or not.

Stratis.Bitcoin/Features/Wallet/Wallet.cs
bokobza: nope, good spot.

Stratis.Bitcoin/Features/Wallet/HdOperations.cs
bokobza: yeah but it's the hd path of an account.
Maybe Gets the extended public key for an account corresponding to an HD path.

Stratis.Bitcoin/Features/Wallet/Wallet.cs
dangershony: Ah ok cool so this is correct actually.

Stratis.Bitcoin/Features/Wallet/HdOperations.cs
dangershony: But any HD path can go in there unless there is a condition limiting the path to have an account

PR #333, opened 2017-08-16
Bugfix/issue 277 blocking push block

### LINKED ISSUE #277: BlockPullerBehavior.Node_MessageReceived is calling blocking PushBlock
This is probably not intentional as it seems that we could still ask for more blocks to arrive and they will but we will block in PushBlock until the consumer consumes them because there are limits on how many bytes we want to use by the puller. However, we will still use more and more RAM as new blocks arrive, they will just not go to the puller. So the limits in the puller won't achieve their original goal - limit RAM consumption.

If we really want to limit amount of RAM consumed, we would need to stop asking, rather than blocking processing (of already received) blocks.

So this issue should be resolved by not asking for more if we are full and probably completely removing the blocking logic in PushBlock.

### LINKED ISSUE #327: Peer.Disconnected incorrectly set to true
@Aprogiena 
The property is set to true in this method ReleaseAllPeerDownloadTaskAssignments()
However this method is not only called when the peers is disconnected but also when a peer IsStalling()
This causes a peer that may still be useful to not be used for downloading. this is especially problematic when the node is connected to only one peer.

I think this caused some of the tests to fail randmoly

Aprogiena:
Inversion of logic of enforcing `LBP.MaxBufferedSize` - blocks are not asked anymore if it is expected that they would violate the limit. 

Fix #277 
Fix #327


Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -504,15 +504,16 @@ private bool ReleaseDownloadTaskAssignmentLocked(Dictionary<uint256, DownloadAss
         /// Releases all pending block download tasks assigned to a peer.
         /// </summary>
         /// <param name="peer">Peer to have all its pending download task released.</param>
+        /// <param name="peerDisconnected">If set to <c>true</c> the peer is considered as disconnected and should be prevented from being assigned additional work.</param>
         /// <exception cref="InvalidOperationException">Thrown in case of data inconsistency between synchronized structures, which should never happen.</exception>
-        internal void ReleaseAllPeerDownloadTaskAssignments(BlockPullerBehavior peer)
+        internal void ReleaseAllPeerDownloadTaskAssignments(BlockPullerBehavior peer, bool peerDisconnected)
```
dangershony: I would suggest to name it disconnectPeer.
>If set to <c>true</c> the peer needs to be is set to a disconnected state and should be prevented from being assigned additional work.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -190,7 +196,8 @@ public LookaheadBlockPuller(ConcurrentChain chain, IConnectionManager connection
         public void SetLocation(ChainedBlock tip)
         {
             Guard.NotNull(tip, nameof(tip));
-            this.location = tip;
+            lock (this.locationLock)
```
Aprogiena: And I agree with you, I was just testing the grounds. I don't like it either. Will fix.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -393,13 +388,84 @@ private void AskBlocks()
...
+                        bool bufferFull = this.currentBufferedSize + (requestsToAsk.Count + 1) * avgBlockSize > this.MaxBufferedSize;
+
+                        if (isNextBlock || !bufferFull)
+                        {
+                            requestsToAsk.Add(request);
+                            this.askBlockQueue.Dequeue();
+                        }
+                        else
+                        {
+                            // The buffer is either full, so we do not want to ask for more blocks.
```
Aprogiena: We do have issues with reorgs in all code, we do have an issue for a memory leak somewhere but I believe there will be more issues. The good thing here is that it seems that reorg will just cause requested blocks from different chain to be downloaded. So they will arrive earlier then they would otherwise (if we cleaned the list). So there seems to be not even a memory leak in this queue in case of reorg. So maybe there is no problem to solve here. But again, I think there are problems with reorgs and many yet to be uncovered.

Stratis.Bitcoin/BlockPulling/BlockPuller.cs
Aprogiena: ok

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -457,11 +529,27 @@ private Block NextBlockCore(CancellationToken cancellationToken)
                         // Or the block is still being downloaded or we need to ask for this block to be downloaded.
                         if (!isDownloading) AskBlocks(new ChainedBlock[] { header });
 
-                        OnStalling(header);
-                        this.IsStalling = true;
+                        this.OnStalling(header);
                     }
 
-                    WaitHandle.WaitAny(new[] { this.pushed, cancellationToken.WaitHandle }, WaitNextBlockRoundTimeMs);
+                    while (!cancellationToken.IsCancellationRequested)
```
Aprogiena: The weirdest part is that in the same method - NBC - you set this.consumed and you also wait for this.consumed AND you expect only one thread to exist in the node that executes NBC. That is the weird part. But due to the logic of NC that it actually calls `AskBlocks` and because `AskBlocks` then calls `ProcessQueue` it actually works. But is very nontrivial to grasp. The original idea with the thread would separate the logic and the design would be very clean, but yes, it would cost the thread :)

Aprogiena:
OK, should be ready to go.

Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
dangershony: Ah yeah this list will just clear itself true so no issue here.
There are issue with reorgs true but if we keep proper separation then the issues will be contained to each feature, so we just solve reorg per feature.

In the puller case we still need to clean the download collections on reorg.

Aprogiena: Let's create discussion issue for reorgs elsewhere - https://github.com/stratisproject/StratisBitcoinFullNode/issues/337

Issue #258, opened 2017-07-27
Code style policy changes


Aprogiena:
The purpose if this issue is to open discussion about changing coding style - https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/coding-style.md

Currently we only ask contributors to follow the coding style, yet we do not have any rules on whether or not a code can be accepted if it violates the coding style. 

The current code base uses different code styles and a lot of code violates many rules of the current style.

I think we should define both - the rules we want (as the current rule set is not sufficient) as well as the policy about how this is or is not enforced - i.e. which rules are required and which are only recommended.

This issue is for StratisBitcoinFullNode project only at this moment, and once we converge to something here, other projects are welcome to follow the outcome as well.

Besides the rules and the policies on their enforcement, we should also define the decision process of which rules will be applied should contributors disagree on some (i.e. who is the boss if any, are there any vetos, or is this just majority voting?).

**PROPOSING RULE (NEW OR CHANGE)**: If you want to propose a new rule, or change the text of the current rule, please start your comment starting with **RULE PROPOSAL** and use three sections: _DEFINITION_ - in this first part just write the rule as you would like to see it; _RATIONALE_ - try to explain why you think this is a good idea to have; _POLICY_ - explain what kind of policy you would like to have on this rule (i.e. would you like it to have only as a general guideline/recommendation, or strictly enforced - i.e. not accepting new code violating it?).

dangershony:
Indeed the policy is if you are working on some part of the code that they styling is different then bring them in line.

Code should not be accepted if it violates the code-style but it wasn't an iron rule (for various reasons).

My take on adding optional and strict on some of the rules seem fine, some contributors are used to other styling from other projects and we can make it easy to contribute by not staying too strict

Aprogiena:
Do you think someone would actually have a problem if in the PR review the reviewer would say "here and here, please fix the style according to code style"? I've seen you already do that in the review process, and I think it is the right think to do - you have a code style for a reason. For example, I have 2 space indentation by default in MSVS, but I changed that to 4 to comply with the project code style. I can't imagine it would be acceptable if I started submitting code with 2 spaces. It would just create messy code.

Aprogiena:
**RULE PROPOSAL**
Properties should be written on a single line unless having a non-trivial getter or setter.

Examples:
`public int X { get; set; }`
`public int X { get; private set; }`
`public BlockPuller Puller { get { return this.puller; } }`
[...]

PR #348, opened 2017-08-18
XML Docs for Mempool classes

mikedennis:


Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -9,35 +9,103 @@
...
             this.memPool = memPool;
-            this.mempoolScheduler = mempoolScheduler;
+            this.mempoolLock = mempoolLock;
             this.mempoolValidator = mempoolValidator;
             this.Set = new UnspentOutputSet();
         }
 
+        #endregion
+
+        #region Properties
```
dangershony: Ho I am not that big fan of so many regions

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -48,11 +116,21 @@ public MempoolCoinView(CoinView inner, TxMempool memPool, AsyncLock mempoolSched
...
+        /// </summary>
+        /// <param name="txid">Transaction identifier.</param>
+        /// <returns>The unspent outputs.</returns>
         public UnspentOutputs GetCoins(uint256 txid)
         {
             return this.Set.AccessCoins(txid);
         }
 
+        /// <summary>
+        /// Whether coins exist for a given transaction id.
```
dangershony: >Check whether a transaction id exists in the <see ref= Mempool or in the <see ref=MempoolCoinView.

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
Aprogiena: Me neither, never found them very useful, but if someone likes them, I don't mind.

mikedennis: Ok I'll get rid of them if no one else uses them.

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -9,35 +9,103 @@
...
     public class MempoolCoinView : CoinView , IBackedCoinView
     {
+        #region Fields
+
+        /// <summary>Transaction memory pool for managing transactions in the memory pool.</summary>
         private readonly TxMempool memPool;
-        private readonly AsyncLock mempoolScheduler;
+
+        /// <summary>A lock for managing asynchronous access to memory pool.</summary>
+        private readonly AsyncLock mempoolLock;
```
Aprogiena: Here I would very much like what I suggest in here - https://github.com/Aprogiena/StratisBitcoinFullNode/blob/d2e959613b87221ad7e5021f6a410bf39069181e/Documentation/commenting-policy.md#contracts

I.e. what you have here is a contract between mempoolLock and memPool and you should describe this contract. I.e. you would like to make sure that everyone who is using memPool is using the lock because it has to be protected.

You can check it here https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin/BlockPulling/BlockPuller.cs#L84 how I use it - check the comment for the lockObject I have there and then e.g. for assignedBlockTasks. What I am looking for here is the remarks section as you can see in assignedBlockTasks - that tells the developers that they have to use the lock for all access. Somewhere else I have some other lock and some other object and I require the lock to be used for write access only, so then I just say that for write access you need to use the lock. So it is important for me to have the contract described properly.

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -9,35 +9,103 @@
...
+
+        #region Operations
+
+        /// <summary>
+        /// Load the coin view for a memory pool transaction.
+        /// </summary>
+        /// <param name="trx">Memory pool transaction.</param>
         public async Task LoadView(Transaction trx)
         {
             // lookup all ids (duplicate ids are ignored in case a trx spends outputs from the same parent)
```
Aprogiena: Feel free to fix this to look like a sentence (i.e. SENTENCES rule).

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -9,35 +9,103 @@
 
 namespace Stratis.Bitcoin.Features.MemoryPool
 {
+    /// <summary>
+    /// Memory pool coin view.
+    /// Provides coin view representation of memory pool transactions via a backed coin view.
+    /// </summary>
```
Aprogiena: When I first learned about this code base I did not know what coin view was. In Bitcoin, I think people understand what UTXOs are, but I never heard about "coin view", so it might be a good idea to use that Bitcoin terminology when commenting. However, I admit that I am not familiar with Bitcoin Core code base, so maybe "coin view" is term from there and I am just ignorant of there, in which case feel free to ignore me :)

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
```diff
@@ -9,35 +9,103 @@
...
         }
 
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// Gets the unspent transaction output set.
+        /// </summary>
+        public UnspentOutputSet Set { get; private set; }
```
Aprogiena: I think we want to have fields as well as properties at the top of the type (i.e. before constructor). We only have this as rule 14 in code style speaking about fields, but I think we should treat properties like that as well for very same reason.

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
Aprogiena: other inline comments as well

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
Aprogiena: +1

Stratis.Bitcoin/Features/MemoryPool/MempoolErrors.cs
```diff
@@ -3,92 +3,214 @@
...
 namespace Stratis.Bitcoin.Features.MemoryPool
 {
+    /// <summary>
+    /// Exception thrown when memory pool exception occurs.
+    /// </summary>
     public class MempoolErrorException : Exception
     {
+        /// <summary>
+        /// Constructs a memory pool exception object.
+        /// Exception message is set from <see cref="MempoolValidationState.ErrorMessage"/>
```
Aprogiena: missing dot at the end

Stratis.Bitcoin/Features/MemoryPool/MempoolErrors.cs
```diff
@@ -3,92 +3,214 @@
...
         {
             this.ValidationState = state;
         }
 
+        /// <summary>Gets the validation state of the memory pool.</summary>
         public MempoolValidationState ValidationState
         {
             get;
             private set;
         }
```
Aprogiena: as per new code style, we use one liners for this

`public MempoolValidationState ValidationState { get; private set; }`

this one already is accepted, but see https://github.com/stratisproject/StratisBitcoinFullNode/issues/258

Stratis.Bitcoin/Features/MemoryPool/MempoolErrors.cs
```diff
@@ -3,92 +3,214 @@
...
 
+        /// <summary>Gets or sets the string representation of the error code.</summary>
         public string Code { get; set; }
+
+        /// <summary>Gets or sets the numeric representation of the error code.</summary>
         public int RejectCode { get; set; }
 
+        /// <summary>Gets or sets the error from consensus feature.</summary>
         public ConsensusError ConsensusError { get; set; }
```
Aprogiena: again, can we move these guys in front of constructors please?

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
mikedennis: I try to follow the stylecop rule for ordering http://stylecop.soyuz5.com/SA1201.html but if we want to vary from it I'm ok with that as well. Honestly I find it more natural to have them right after fields but in stylecop they put them after the constructor.

mikedennis: Looks like MS seems to put properties after constructors as well. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.ViewFeatures/RemoteAttribute.cs

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
dangershony: Yes indeed bitcoin core uses the term CoinView

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
dangershony: Yeah I no preference either before or after the constructor currently

Aprogiena: I am not sure we use that rule from SC, because i certainly do put fields and properties together in front of constructors and it does not complain.

@mikedennis are you and me using the same set of SC rules? I have those from @bokobza that are currently in the Stratis repo.

Stratis.Bitcoin/Features/MemoryPool/MemPoolCoinView.cs
Aprogiena: I see, tx

PR #351, opened 2017-08-18
Feature/circular array class

Aprogiena:
Separation of QualityScore and its circular array. 
Tests included.


bokobza:
I'll do this one today.

dangershony:
FYI I just noticed that a Queue is also implemented using circular array techniques
https://msdn.microsoft.com/en-us/library/system.collections.queue(v=vs.110).aspx

Aprogiena:
Interesting :) Did not know that. The question is - can you avoid the allocations with Queue?

dangershony:
I don't think, in the docs I see nothing that limits new allocation of memory, and in the second link it is suggested there is no such existing implementation.

Aprogiena:
Well anyway, we have our thing now :)

Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -104,28 +93,27 @@ public void AddSample(IBlockPullerBehavior peer, long blockDownloadTimeMs, int b
...
                     if (this.peerReferenceCounter[oldSample.peer] == 0)
                         this.peerReferenceCounter.Remove(oldSample.peer);
                 }
-                else this.samplesCount++;
 
                 // Add new sample to the mix.
-                this.samples[this.samplesIndex].timePerKb = timePerKb;
-                this.samples[this.samplesIndex].peer = peer;
-                this.samplesIndex = (this.samplesIndex + 1) % this.samples.Length;
+                int index = this.samples.AddNoSet();
```
bokobza: Why not use Add(T item) here?

Stratis.Bitcoin/Utilities/CircularArray.cs
```diff
@@ -0,0 +1,110 @@
...
+            int index = this.AddNoSet();
+            this.items[index] = item;
+        }
+
+        /// <summary>
+        /// Add a new item to the circular array without setting its value.
+        /// </summary>
+        /// <returns>Index to the array where the new item should be placed.</returns>
+        /// <remarks>This method only moves the array indexes as if a new item was added, but the caller is responsible for actually setting new values of the item.</remarks>
+        public int AddNoSet()
```
bokobza: This should probably be made private. The consumer of the class doesn't need to know at which index the next item be inserted. Also, the name AddNoSet doesn't sound right. Is it really adding an item to the array?

Stratis.Bitcoin/BlockPulling/QualityScore.cs
Aprogiena: That was the whole purpose of using it - to avoid allocation. When it was inline, it was super simple, but @dangershony did not like the inline implementation. So I separated it only to find out that it is hard to avoid the allocation with the "standard" separated implementation (I've basically ended up with simpler version of Queue), so to gain back the full power of why I used it in the first place, I had to use AddNoSet.

Stratis.Bitcoin/Utilities/CircularArray.cs
Aprogiena: See the above comment. I agree on the name - I was just unable to find a better name, feel free to suggest one. It does not really add an item, it rather allocates a space for it, but then I was thinking - I'm doing this to avoid allocation, so calling it AllocateSpace or something like that is even worse :-) 

To be honest, for me the original inline implementation was best, but now I separated it, have tests, the code using it is slightly shorter, no performance penalty when using AddNoSet ... so OK.

Aprogiena:
@bokobza I was thinking about returning removed object, but then I rejected it. I was also considering adding more features but then I've rejected that as well as per rule "don't implemented until you need it".

But to be honest, all your comments are perfectly valid and I'm not very excited about the outcome. I'm happy to discuss further improvements, but I'm definitely looking for preserving the original goal - to have a single allocation at the start and then no allocations while using it. Happy to take suggestions.

bokobza:
I see what you're saying mate.
Personally, I really think the overhead of allocation is so negligible it's not even worth going through the trouble. I mean we're not developing a game here or a real-time system.
What about making PeerSample a struct rather than a class, would that work?

Aprogiena:
Struct - Not sure if that would help, probably not. Can't make my mind about this right now though :-) Having said that, it should be struct because it just is struct by nature (I'm just always lazy to use structs instead of classes - the benefits are mostly negligible).

Well, it can be considered as a little bit of over optimization. Although the number of allocations saved is 1 per block and you have 400k blocks. Still quite negligible yes. And basically we've saved like 8 lines of code having this as a separated class. 

So when I made the original code (with inline circular array), it felt natural to put it there to save allocations (i.e. it costed me almost nothing and it was trivial for me). But with the separation, the benefit of using it vanishes quickly. Maybe the original unseparated implementation was the best, but then @dangershony objected about the complexity of the code. So then using Queue might be the best, but then if we compare this separated our own implementation of circular array, what is the cost of having it? Nothing much really, maybe except that use of NoSet version of Add.

In any case, whichever version (1 - this CA, 2 - original inline, 3 - Queue; A - class, B - struct) we use, it is not critical because as you mentioned this is not bottleneck.

Aprogiena:
@bokobza maybe check the original inline implementation of that circular array that @dangershony was questioning. How does that look to you?

bokobza:
Well with a struct there is no allocation on the stack because the whole struct is kept on it. so your array doesn't constitute of 100 pointers to PeerSample class objects, but rather it is made up of 100 PeerSample struct objects, as a value type would be.
Therefore, by using struct, we're removing the need for allocation and GC optimization.

Aprogiena:
@bokobza yes, but does that really helps us somehow? i.e. if we switch to struct here and use Queue instead of our own thing, will it work? I am very unsure about it, is it possible to somehow prevent Queue from expanding?

so what you say would probably help us only with that we would not need AddNoSet, right?

bokobza:
Yeah it'll help us with avoiding multiple allocations (if I understand that's what we want to avoid) and therefore we could make AddNoSet private and also have a AddOrReplace(PeerSample) method or something. Then I think the benefit of this class will be more apparent :-)

Aprogiena:
@bokobza not so fast :) there is nothing in CA that would justify doing so, CA is independent structure that should care about itself. The fact that we use it for storing PeerSample is separated thing that should not affect CA's implementation. Thus CA to be generic, should have this AddNoSet public thing. We still make PeerSample struct though.

bokobza:
I didn't mean for CA to actually know anything about PeerSample.
By AddOrReplace(PeerSample) I meant a method AddOrReplace that accepts an object, in our case it's PeerSample.
The signature of the method I had in mind would be `T AddOrReplace(T item)` (`where T : struct`) - basically the same as Add(T item) that's already there. 
My only additional suggestion is to return the replaced object. 
I think if the method is called AddOrReplace, it is perfectly valid to return the replaced object.

Aprogiena:
I don't get it. If we take down AddNoSet, we suddenly would prevent using CA for classes in efficient way. I'm also failing to see what is AddOrReplace good for. In CA, every add is in fact add or replace.

Aprogiena:
OK, so we discussed this and there we go, next attempt :)

Stratis.Bitcoin/Utilities/CircularArray.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <item><see cref="AddNoSet"/> - O(1),</item>
+    /// <item><see cref="this[int]"/> - O(1).</item>
+    /// </list>
+    /// </remarks>
+    public class CircularArray<T> : IEnumerable<T> where T : new()
+    {
+        /// <summary>Maximal number of items that can be stored in <see cref="items"/> array.</summary>
+        public int Capacity { get; private set; }
+
+        /// <summary>Number of valid items in <see cref="items"/> array.</summary>
```
bokobza: What does "valid" mean in this context?

Stratis.Bitcoin/Utilities/CircularArray.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <summary>
+    /// Generic circular array is a fixed length array which stores collection 
+    /// of items. Once the array is full, adding a new item removes the oldest entry. 
+    /// </summary>
+    /// <typeparam name="T">Type of the items stored in the array.</typeparam>
+    /// <remarks>
+    /// Complexity of supported operations:
+    /// <list type="bullet">
+    /// <item><see cref="Add"/> - O(1),</item>
+    /// <item><see cref="AddNoSet"/> - O(1),</item>
```
bokobza: This can be removed.

Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -103,29 +92,28 @@ public void AddSample(IBlockPullerBehavior peer, long blockDownloadTimeMs, int b
...
+                // Add new sample to the mix.
+                PeerSample newSample = new PeerSample();
+                newSample.timePerKb = timePerKb;
+                newSample.peer = peer;
+
+                if (this.peerReferenceCounter.ContainsKey(peer)) this.peerReferenceCounter[peer]++;
+                else this.peerReferenceCounter.Add(peer, 1);
+
+                PeerSample oldSample;
+                if (this.samples.Add(newSample, out oldSample))
```
bokobza: you can even do `if (this.samples.Add(newSample, out PeerSample oldSample))` now, how cool is that?

Aprogiena: i only use that construct if i don't use the out variable
if i want to use it, i want to be explicitly visible - i.e. have separated line

bokobza: cool :-)

Aprogiena:
tx, try now

Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -103,29 +92,28 @@ public void AddSample(IBlockPullerBehavior peer, long blockDownloadTimeMs, int b
 
             lock (this.lockObject)
             {
-                // If we reached the maximum number of samples, we need to remove oldest sample.
-                if (this.samplesCount == this.samples.Length)
-                {
-                    PeerSample oldSample = this.samples[this.samplesIndex];
+                // Add new sample to the mix.
+                PeerSample newSample = new PeerSample();
```
dangershony: I don't understand this part, if we call this method 1 million times wont we create 1mil new objects? 
The idea was to prevent allocation of new memory.

Aprogiena: if it is struct, which it is and this code is responsible for that, then allocation here means just extra 2 items on stack, i.e. this is safely in align with what was the original purpose

Issue #339, opened 2017-08-17
[Improvement] Refactor BlockStorePuller/BlockStoreLoop so that BlockStoreInnerStepFindBlocks can ask for multiple blocks


fassadlr:
Set a threshold of 10 blocks to be asked for.

fassadlr:
![](https://github.trello.services/images/mini-trello-icon.png) [[BlockStore] 2 Improve how the DownloadBlocks step asks for blocks](https://trello.com/c/sBfhAgTA/167-blockstore-2-improve-how-the-downloadblocks-step-asks-for-blocks)

fassadlr:
@dangershony I'm just looking at this and I see that we check the BatchDownloadSize before we stop asking for new blocks in BlockStoreInnerStepFindBlocks:

```
context.BlockStoreLoop.BlockPuller.AskForSingleBlock(context.NextChainedBlock);
context.DownloadStack.Enqueue(context.NextChainedBlock);

if (context.DownloadStack.Count == context.BlockStoreLoop.BatchDownloadSize)
    context.StopFindingBlocks();
```

Can't we change that BatchDownloadSize to 10 (it is currently hard coded to 1000) and then at the same time call the new method which asks for multiple blocks? Like so:

```
context.DownloadStack.Enqueue(context.NextChainedBlock);

if (context.DownloadStack.Count == context.BlockStoreLoop.BatchDownloadSize)
{
    context.StopFindingBlocks();
    context.BlockStoreLoop.BlockPuller.AskForMultipleBlocks(context.DownloadStack.ToArray());
}
```

Then in BlockStoreInnerStepDownloadBlocks we'll need to be able to potentially have BlockPuller.TryGetBlock return mulitple blocks? Somethign like this:

```
BlockPuller.DownloadedBlock[] downloadedBlocks;

if (context.BlockStoreLoop.BlockPuller.TryGetBlocks(context.DownloadStack, out downloadedBlocks))
{
```

Thoughts?

fassadlr:
I think I might misunderstood where we want to ask for multiple blocks... Should this be done here?
```

 internal sealed class DownloadBlockStep : BlockStoreLoopStep
{
    internal DownloadBlockStep(BlockStoreLoop blockStoreLoop)
            : base(blockStoreLoop)
    {
    }

    /// <inheritdoc/>
[...]

PR #361, opened 2017-08-23
[Improvement] Refactor BlockStorePuller/BlockStoreLoop so that BlockStoreInnerStepFindBlocks can ask for multiple blocks #339

fassadlr:


Stratis.Bitcoin/BlockPulling/StoreBlockPuller.cs
```diff
@@ -25,14 +24,16 @@ public StoreBlockPuller(ConcurrentChain chain, Connection.IConnectionManager nod
...
         /// </summary>
-        /// <param name="downloadRequest">Description of a block to download.</param>
-        public void AskBlock(ChainedBlock downloadRequest)
+        public void AskForMultipleBlocks(ChainedBlock[] downloadRequests)
         {
-            this.logger.LogTrace("({0}:'{1}/{2}')", nameof(downloadRequest), downloadRequest.HashBlock, downloadRequest.Height);
+            foreach (var downloadRequest in downloadRequests)
+            {
+                this.logger.LogTrace("({0}:'{1}/{2}')", nameof(downloadRequest), downloadRequest.HashBlock, downloadRequest.Height);
+            }
```
Aprogiena: `()` should appear only once in a log for each method call. So does `(-)`
This is better to be replaced with 
```
 this.logger.LogTrace("({0}.{1}:{2}')", nameof(downloadRequest), nameof(downloadRequest.Count), downloadRequest.Count);
```

Note that you are calling `BlockPuller.AskBlocks` which does enumerate block heights, so the count here is enough.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepContext.cs
```diff
@@ -40,34 +40,27 @@ public sealed class BlockStoreInnerStepContext
...
-        public BlockStoreInnerStepContext(CancellationToken cancellationToken, BlockStoreLoop blockStoreLoop, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider)
+        public BlockStoreInnerStepContext(CancellationToken cancellationToken, BlockStoreLoop blockStoreLoop, ChainedBlock nextChainedBlock, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider)
         {
             Guard.NotNull(blockStoreLoop, nameof(blockStoreLoop));
+            Guard.NotNull(nextChainedBlock, nameof(nextChainedBlock));
 
             this.loggerFactory = loggerFactory;
             this.logger = loggerFactory.CreateLogger(GetType().FullName);
-            this.DateTimeProvider = dateTimeProvider;
+            this.logger.LogTrace("({0}:'{1}/{2}')", nameof(nextChainedBlock), nextChainedBlock?.HashBlock, nextChainedBlock?.Height);
```
Aprogiena: We do have a guarded guarantee of nextChainedBlock not being `null` here. So you can remove question marks.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,66 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+                context.GetNextBlock();
             }
 
+            context.StopFindingBlocks();
+
+            if (context.DownloadStack.Any())
+                context.BlockStoreLoop.BlockPuller.AskForMultipleBlocks(context.DownloadStack.ToArray());
+
             this.logger.LogTrace("(-):{0}", InnerStepResult.Next);
+
```
Aprogiena: I don't like this space between `(-)` and return :-)

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,66 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
             this.logger.LogTrace("()");
 
-            bool res = (context.NextChainedBlock == null)
-                || (context.NextChainedBlock.Header.HashPrevBlock != context.InputChainedBlock.HashBlock)
-                || (context.NextChainedBlock.Height > context.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
-                || (context.BlockStoreLoop.PendingStorage.ContainsKey(context.NextChainedBlock.HashBlock))
-                || (await context.BlockStoreLoop.BlockRepository.ExistAsync(context.NextChainedBlock.HashBlock));
+            if (context.NextChainedBlock == null)
+            {
+                this.logger.LogTrace("(-): {0} was null, return true", nameof(context.NextChainedBlock));
```
Aprogiena: If we could use `[]` for context and `(-):{0}` for return value:

```
               this.logger.LogTrace("(-)[NO_NEXT_CHAINED_BLOCK]:true"));
```

you are not very much interested in the description of the logic in the log, you only need a reference point (given by the context `[NO_NEXT_CHAINED_BLOCK]`), so you are able to find it in the code if you were interested.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,66 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+                return true;
+            }
+
+            if (context.InputChainedBlock != null && (context.NextChainedBlock.Header.HashPrevBlock != context.InputChainedBlock.HashBlock))
+            {
+                this.logger.LogTrace("(-): {0} previous hash does not match {1} hash, return true.", nameof(context.NextChainedBlock), context.InputChainedBlock);
```
Aprogiena: similarly in other cases here, this could be 
```
               this.logger.LogTrace("(-)[HASH_MISMATCH]:true"));
```

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,66 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+                return true;
+            }
+
+            if (context.NextChainedBlock.Height > context.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
+            {
+                this.logger.LogTrace("(-): {0} height was higher than {1}, return true.", nameof(context.NextChainedBlock), nameof(context.BlockStoreLoop.ChainState));
```
Aprogiena: this is reorg situation right?
if so, then the easiest is
```
               this.logger.LogTrace("(-)[REORG]:true"));
```

if it is not reorg, then 
```
               this.logger.LogTrace("(-)[HEIGHT_MISMATCH]:true"));
```

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,66 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+                return true;
+            }
+
+            if (await context.BlockStoreLoop.BlockRepository.ExistAsync(context.NextChainedBlock.HashBlock))
+            {
+                this.logger.LogTrace("(-): {0} contains {1}, return true", nameof(context.BlockStoreLoop.BlockRepository), nameof(context.NextChainedBlock));
+                return true;
+            }
+
+            if (context.DownloadStack.Count >= 10)
```
Aprogiena: could we remove the hardcoded value of `10` here and replace it with commented constant?

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepReadBlocks.cs
```diff
@@ -3,13 +3,13 @@
...
 using System.Collections.Generic;
-using System.Linq; 
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
 {
     /// <summary>
-    /// Downloads blocks from the BlockPuller removes block from the DownloadStack.
+    /// Reads blocks from the BlockPuller and removes block from the DownloadStack.
```
Aprogiena: `<see>` for DownloadStack and BlockPuller 

OR

Reads blocks from the puller and removes block from the download stack.

both variants are OK towards renaming of those classes, but the original is not

Aprogiena: when you see such lack of `<see>` elsewhere, pls apply this too

Stratis.Bitcoin/Features/BlockStore/LoopSteps/DownloadBlockStep.cs
```diff
@@ -13,27 +13,28 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
...
         private readonly IDateTimeProvider dateTimeProvider;
 
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
         internal DownloadBlockStep(BlockStoreLoop blockStoreLoop, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider)
             : base(blockStoreLoop, loggerFactory)
         {
-            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.logger = loggerFactory.CreateLogger(GetType().FullName);
```
Aprogiena: this is violation of `this.` code style requirement

Aprogiena: ctrl+shift+f reveals that only BS code is using GetType() without `this.` all other code follows the code style rule for `this.`, let's change all these to follow the code style

Stratis.Bitcoin/BlockPulling/StoreBlockPuller.cs
fassadlr: Have fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepContext.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: @Aprogiena I honestly cannot see the benefit of adding logs to every IF statement here. If we want to know why a particular stop condition was reached for a particular case then we can use the debugger to step into the code and see what's going on. I'm going to take them out and just log the result.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: See above

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: See above

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepReadBlocks.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/DownloadBlockStep.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
Aprogiena: logging the result is fine - as mentioned, you are the one who should decide which extra information is useful to have in logs

so first, you have some base line (log input parameters and output results) and then you decide if you want to add something, so if you don't see a benefit in having that context information, it is fine, although in this specific case you would end up with the code you have replaced (which is just logging the result without that context - and yes i understand that in debugger you could see more easier which case it was with your separated ifs, but still even in that joined code you could see that quite easily (just hover your mouse over the expression)) so if you will keep the ifs separated, having the context does not add any lines, just small little context identifier

but change it as you think and let me check then

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,50 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
 
-            bool res = (context.NextChainedBlock == null)
-                || (context.NextChainedBlock.Header.HashPrevBlock != context.InputChainedBlock.HashBlock)
-                || (context.NextChainedBlock.Height > context.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
-                || (context.BlockStoreLoop.PendingStorage.ContainsKey(context.NextChainedBlock.HashBlock))
-                || (await context.BlockStoreLoop.BlockRepository.ExistAsync(context.NextChainedBlock.HashBlock));
+            bool result = false;
+
+            if (context.NextChainedBlock == null)
+                result = true;
```
Aprogiena: here we need `else`s because it would do extra work compared to original code and it would also die on exception when NCB is `null`

fassadlr: Oh yes, sorry! But @Aprogiena I'm actually going to remove logging here entirely. We know the while stopped because of some condition. It doesn't benefit us, like you said, to know that the obvious happened. If the need arises in the future to have more logging here we can add it in :)

fassadlr:
[[BlockStore] 2 Improve how the DownloadBlocks step asks for blocks](https://trello.com/c/sBfhAgTA/167-blockstore-2-improve-how-the-downloadblocks-step-asks-for-blocks)

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
Aprogiena: Well, I'm not really happy about your decision here, in fact I think the code that you just refactored to is actually worse than what you started with (that one with single return and logging). It seems the very nature of the whole mechanism will have a log gap and I don't understand why (to save lines of code or to have a bit easier debugging?). I'd be absolutely happy with that separated logging with multiple returns giving the context of why it returned true. But we seem to disagree there.

I'll leave it as you think it is better. So I will just run the tests to check the functionality is not somehow broken and approve it.

Aprogiena:
Not being able to run integration tests, so I'm approving changes, but not merging.

Stratis.Bitcoin/Features/BlockStore/BlockStoreLoop.cs
```diff
@@ -53,16 +53,11 @@ public class BlockStoreLoop
         /// <summary>Blocks that in PendingStorage will be processed first before new blocks are downloaded.</summary>
         public ConcurrentDictionary<uint256, BlockPair> PendingStorage { get; }
 
-        /// <summary>The highest stored block in the repository</summary>
-        internal ChainedBlock StoreTip { get; private set; }
-
-        /// <summary>TODO: Should be configurable?</summary>
+        /// <summary>The amount blocks that can be stored in pending storage before we stop processing them.</summary>
```
Aprogiena: Number of blocks that can ...

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -7,19 +7,13 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
     /// <summary>
     /// Find blocks to download by asking the BlockPuller.
     /// <para>
-    /// If a stop condition is found <see cref="ShouldStopFindingBlocks"/> and
-    /// there aren't blocks to download anymore, return a Break() result causing the 
-    /// BlockStoreLoop to break execution and start again.
+    /// Find blocks until <see cref="BlockStoreInnerStepContext.DownloadStack"/> contains 10 blocks.
```
Aprogiena: avoid `10`, replace with `<see>` on the constant

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepReadBlocks.cs
```diff
@@ -3,27 +3,27 @@
...
-    /// Downloads blocks from the BlockPuller removes block from the DownloadStack.
+    /// Reads blocks from the <see cref="BlockPuller"/> and removes block from the <see cref="BlockStoreInnerStepContext.DownloadStack"/>.
     /// <para>
     /// If the block exists in the puller add the the downloaded block to the store to
     /// push to the repository. If the <see cref="BlockStoreLoop.MaxInsertBlockSize"/> has been reached
     /// push the blocks in the context's Store to the repository.
     /// </para> 
     /// <para>
-    /// When the download stack is empty return a Break() result causing the BlockStoreLoop to
+    /// When the download stack is empty return a Break() result causing the <see cref="BlockStoreLoop"/> to
```
Aprogiena: Break is not a method, I event can't see any Break in the code, do you mean InnerStepResult.Stop?

Stratis.Bitcoin/Features/BlockStore/LoopSteps/DownloadBlockStep.cs
```diff
@@ -13,23 +13,24 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
...
     /// </list>
     /// </para> 
     /// <para>
-    /// After a "Stop" condition is found the FindBlocksTask will be removed from 
-    /// the <see cref="BlockStoreInnerStepContext.InnerSteps"/> and only the 
-    /// <see cref="BlockStoreInnerStepDownloadBlocks"/> will continue to execute until the <see cref="BlockStoreInnerStepContext.DownloadStack"/> is empty.
+    /// After a "Stop" condition is found the AskBlocksTask will be removed from
+    /// <see cref="BlockStoreInnerStepContext.InnerSteps"/> and only the 
+    /// <see cref="BlockStoreInnerStepReadBlocks"/> task will continue to execute 
+    /// until the DownloadStack is empty.
```
Aprogiena: `<see>` here for DownloadStack, and for AskBlockTask above, but again, it seems that AskBlocksTask does not exist in the code. Was it renamed?

Stratis.Bitcoin/BlockPulling/StoreBlockPuller.cs
```diff
@@ -25,14 +25,13 @@ public StoreBlockPuller(ConcurrentChain chain, Connection.IConnectionManager nod
...
         /// <summary>
-        /// Prepares and invokes download task for a single block.
+        /// Prepares and invokes a download task for multiple blocks.
         /// </summary>
-        /// <param name="downloadRequest">Description of a block to download.</param>
-        public void AskBlock(ChainedBlock downloadRequest)
+        public void AskForMultipleBlocks(ChainedBlock[] downloadRequests)
         {
-            this.logger.LogTrace("({0}:'{1}/{2}')", nameof(downloadRequest), downloadRequest.HashBlock, downloadRequest.Height);
+            this.logger.LogTrace("({0}:'{1} to ask.')", nameof(downloadRequests), downloadRequests.Count());
```
fassadlr: `this.logger.LogTrace("({0}.{1}:{2})", nameof(downloadRequests), nameof(downloadRequests.Count), downloadRequests.Count());`

Isn't valid, give compiler error ;)

Stratis.Bitcoin/Features/BlockStore/BlockStoreLoop.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepReadBlocks.cs
fassadlr: Fixed

Stratis.Bitcoin/Features/BlockStore/LoopSteps/DownloadBlockStep.cs
fassadlr: Fixed

Stratis.Bitcoin/BlockPulling/StoreBlockPuller.cs
fassadlr: Suggestions?

Aprogiena: sorry, yes, it is array -> .Length

Aprogiena:
```
            this.logger.LogTrace("({0}.{1}:{2})", nameof(downloadRequests), nameof(downloadRequests.Length), downloadRequests.Length); 
```

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -39,43 +33,44 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+            if (await context.BlockStoreLoop.BlockRepository.ExistAsync(context.NextChainedBlock.HashBlock))
+                return true;
 
-            bool res = (context.NextChainedBlock == null)
-                || (context.NextChainedBlock.Header.HashPrevBlock != context.InputChainedBlock.HashBlock)
-                || (context.NextChainedBlock.Height > context.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
-                || (context.BlockStoreLoop.PendingStorage.ContainsKey(context.NextChainedBlock.HashBlock))
-                || (await context.BlockStoreLoop.BlockRepository.ExistAsync(context.NextChainedBlock.HashBlock));
+            if (context.DownloadStack.Count >= BlockStoreInnerStepContext.DownloadStackThreshold)
+                return true;
```
Aprogiena: i think it would be beneficial to put this condition first or second because its evaluation costs nothing compared to possibly other checks, especially BS.ExistAsync could probably be heavy (heavier)

Stratis.Bitcoin/BlockPulling/StoreBlockPuller.cs
fassadlr: Fixed :)

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: Well spotted! Fixed 👍

PR #366, opened 2017-08-23
Stubs for mempool validator tests

mikedennis:
- Moved mempool unit tests to appropriate folder location/namespace inline with testing standards (caused some changes to other files do to usings being slightly different)
- Created test stubs for ~45 tests for testing different mempool validator scenarios. Added some comments about how to potentially test.

### DIFF
--- a/Stratis.Bitcoin.Tests/BlockStore/BlockRepositoryInMemory.cs
+++ b/Stratis.Bitcoin.Tests/BlockStore/BlockRepositoryInMemory.cs
@@ -7,7 +7,7 @@ using System.Threading.Tasks;
 
 namespace Stratis.Bitcoin.Tests.BlockStore
 {
-    public class BlockRepositoryInMemory : Features.BlockStore.IBlockRepository
+    public class BlockRepositoryInMemory : Bitcoin.Features.BlockStore.IBlockRepository
     {
         private ConcurrentDictionary<uint256, Block> store;
         public uint256 BlockHash { get; private set; }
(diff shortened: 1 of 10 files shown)


bokobza:
I'll take this one.

bokobza:
Good work on the stubs.
Do you need to add the "Bitcoin" to these objects' types? Does it not work without?
Like in  "internal static **Bitcoin**.Features.Wallet.Wallet CreateWallet(string name)"

mikedennis:
Problem is there is a parallel "Features" namespace in the test project now so dropping Bitcoin is ambiguous. I think probably could fix it by getting rid of the fully qualified name, and have a using instead. I could give that a try if that's preferred?

bokobza:
ah ok. :-)
either using the full namespace (quite long), or what you suggested.
I mean it's personal preference and I was surprised to see something starting with Bitcoin.
No big deal anyway, so if it's not too much trouble, try to see what's more pleasing and I'll take that.

bokobza:
also there is a file conflict that needs addressing.

mikedennis:
Sure thing. I prefer using short names with using (instead of fully qualified) and only qualifying them if there is a conflict, so I will give that a try. Maybe there was a reason for that in the first place I don't know about but will give it a try. Also will sort out that conflict. Thanks!

mikedennis:
Ok I see now why the namespace is fully qualified. There is a conflict with that same interface, IBlockRepository, in NBitcoin that's why it was fully qualified. I think it's probably best to leave it as is as a lot less references to it than NBitcoin objects.

bokobza:
ah yes indeed. 
Cool I'm happy with that. Are you done with this change?

mikedennis:
Yes I am. Thanks Jeremy!

PR #369, opened 2017-08-25
Blockstore download to have internal loops, this is to ask batches of …

dangershony:
…work to the puller

### DIFF
--- a/Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
+++ b/Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
@@ -1,6 +1,9 @@
-﻿using Microsoft.Extensions.Logging;
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
 using System.Linq;
 using System.Threading.Tasks;
+using NBitcoin;
 
 namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
 {
@@ -35,16 +38,23 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
         {
             this.logger.LogTrace("()");
 
-            while (await ShouldStopFindingBlocks(context) == false)
+            var batchSize = BlockStoreInnerStepContext.DownloadStackThreshold - context.DownloadStack.Count;
+            var batchList = new List<ChainedBlock>(batchSize);
+			while (batchList.Count < batchSize)
             {
+	            if (await ShouldStopFindingBlocks(context))
+	            {
+					context.StopFindingBlocks();
+	                break;
+	            }
+
+                batchList.Add(context.NextChainedBlock);
                 context.DownloadStack.Enqueue(context.NextChainedBlock);
                 context.GetNextBlock();
             }
 
-            context.StopFindingBlocks();
-
-            if (context.DownloadStack.Any())
-                context.BlockStoreLoop.BlockPuller.AskForMultipleBlocks(context.DownloadStack.ToArray());
+            if (batchList.Any())
+                context.BlockStoreLoop.BlockPuller.AskForMultipleBlocks(batchList.ToArray());
 
             this.logger.LogTrace("(-):{0}", InnerStepResult.Next);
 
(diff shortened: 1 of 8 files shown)


Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -35,16 +38,23 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
         {
             this.logger.LogTrace("()");
 
-            while (await ShouldStopFindingBlocks(context) == false)
+            var batchSize = BlockStoreInnerStepContext.DownloadStackThreshold - context.DownloadStack.Count;
+            var batchList = new List<ChainedBlock>(batchSize);
```
fassadlr: I don't understand why we can't work with the DownloadStack directly?

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -35,16 +38,23 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
         {
             this.logger.LogTrace("()");
 
-            while (await ShouldStopFindingBlocks(context) == false)
+            var batchSize = BlockStoreInnerStepContext.DownloadStackThreshold - context.DownloadStack.Count;
```
fassadlr: In ShouldStopFindingBlocks we check to see if the DownloadStack's count has exceeded the DownloadStackThreshold so why do we need a batchSize to check? I think it comes to the same thing?

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
```diff
@@ -35,16 +38,23 @@ public BlockStoreInnerStepFindBlocks(ILoggerFactory loggerFactory)
...
+					context.StopFindingBlocks();
+	                break;
+	            }
+
+                batchList.Add(context.NextChainedBlock);
                 context.DownloadStack.Enqueue(context.NextChainedBlock);
                 context.GetNextBlock();
             }
 
-            context.StopFindingBlocks();
```
fassadlr: Should'nt we always STOP to ask for blocks until a stop condition has been found or the download stack threshold reached?

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepReadBlocks.cs
```diff
@@ -41,65 +42,69 @@ public BlockStoreInnerStepReadBlocks(ILoggerFactory loggerFactory)
...
             }
+            
 
-            BlockPuller.DownloadedBlock downloadedBlock;
-
-            ChainedBlock nextBlock = context.DownloadStack.Peek();
+            var loopCount = BlockStoreInnerStepContext.DownloadStackThreshold;
 
-            if (context.BlockStoreLoop.BlockPuller.TryGetBlock(nextBlock, out downloadedBlock))
+            while (--loopCount > 0)
```
fassadlr: I'm not sure that this is the right thing to do as by the time we get here the outer loop would only be running this InnerStep and as such the loop seems a bit unnecessary? Basically, we'll continue to call TryGetBlock until the DownloadStack is empty, then we'll stop and the outer BS loop will start again.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
dangershony: In your implementation the code will ask for x blocks and then move on to the read blocks and wait for it to be drained.

That introduces a bottle neck, ideally we want to continuously ask for blocks and store blocks.

Currently the code loops over the two steps together so the result was
Loop over:
- Ask one block
- Take one block (and add to a pending commit list if the list is above a threshold then push)
> this gives a continuation loop that asks and pushes

After you previous change
- ask for 10 blocks
Loop:
- Take one block (and add to a pending commit list if the list is above a threshold then push)

The issue is we wait for the 10 blocks to download before we ask for the next 10 blocks.
The desire is:
- Ask 10 blocks
- 5 blocks arrived 
- Ask for another 5 blocks

dangershony:
I am happy to close this or let you continue in this branch.

Stratis.Bitcoin/Features/BlockStore/LoopSteps/BlockStoreInnerStepFindBlocks.cs
fassadlr: OK thanks for clarifying @dangershony! I understand now what the issue was with my implementation.

dangershony:
@fassadlr I made some more changes please take over this branch and finish it up.

dangershony:
This is continued on another PR #369

PR #372, opened 2017-08-25
Added the Hex of receiving and spending transactions.

bokobza:
@dangershony @zeptin


dangershony:
Great we just need to get @zeptin to take this across now.

PR #375, opened 2017-08-25
adding improvements on GetAll method in TextFileConfiguration

alexandre-spieser:
This relates to this Trello story
https://trello.com/c/tnh91zAa/123-nodesettings-textfileconfiguration-get-all-method

### DIFF
--- a/Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
+++ b/Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
@@ -5,68 +5,80 @@ namespace Stratis.Bitcoin.Tests.NodeConfiguration
 {
     public class TextFileConfigurationTests
     {
-
+        /// <summary>
+        /// Assert that command line arguments with no value assigned default to "1"
+        /// </summary>
         [Fact]
         public void GetAllWithArrayArgs()
         {
             // Arrange
-            var SUT = new TextFileConfiguration(new[] { "test" });
+            var textFileConfiguration = new TextFileConfiguration(new[] { "test" });
             // Act
-            var parsed = SUT.GetAll("test");
+            string[] result = textFileConfiguration.GetAll("test");
             // Assert
-            Assert.Equal("1", parsed[0]);
+            Assert.Equal("1", result[0]);
         }
 
+        /// <summary>
+        /// Assert that we can get all the default values of command line arguments with or without a dash prefixing the key
+        /// </summary>
         [Fact]
-        public void GetAllMultipleKeysWithArrayArgsNoAssignment()
+        public void GetAllWithArrayArgsNoAssignment()
         {
             // Arrange
-            var SUT = new TextFileConfiguration(new[] { "test", "-test" });
+            var textFileConfiguration = new TextFileConfiguration(new[] { "test", "-test" });
             // Act
-            var parsed = SUT.GetAll("test");
+            string[] result = textFileConfiguration.GetAll("test");
             // Assert
-            Assert.Equal(2, parsed.Length);
-            Assert.Equal("1", parsed[0]);
-            Assert.Equal("1", parsed[1]);
+            Assert.Equal(2, result.Length);
+            Assert.Equal("1", result[0]);
+            Assert.Equal("1", result[1]);
         }
 
+        /// <summary>
+        /// Assert that the parsing of command line arguments does not support spaces on each side of the = sign
+        /// </summary>
         [Fact]
-        public void GetAllMultipleKeysWithArrayArgsAssignment_Spaces()
+        public void FailsToGetAllKeysWithArrayArgsAssignmentWithSpaces()
         {
             // Arrange
-            var SUT = new TextFileConfiguration(new[] { "test = testValue1", "-test = testValue2" });
+            var textFileConfiguration = new TextFileConfiguration(new[] { "test = testValue1", "-test = testValue2" });
             // Act
-            var parsed = SUT.GetAll("test");
+            string[] result = textFileConfiguration.GetAll("test");
             // Assert
-            Assert.Equal(2, parsed.Length);
-            Assert.Equal("testValue1", parsed[0]);
-            Assert.Equal("testValue2", parsed[1]);
+            Assert.Equal(0, result.Length);
         }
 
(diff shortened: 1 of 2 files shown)


Stratis.Bitcoin/Configuration/TextFileConfiguration.cs
```diff
@@ -44,10 +40,11 @@ public TextFileConfiguration(string[] args)
             foreach (var arg in args)
             {
                 string[] splitted = arg.Split('=');
+                // we must also trim in case the CL commands has spaces: test = testValue1
```
Aprogiena: This is quite unlikely to happen. It would require someone to write command line as

`program "arg = value"`

instead of 

`program arg=value`

but also instead of 

`program arg = value`

So you kind of assume that the user is familiar with how to properly escape with quotes `"` but you assume the user to be unaware of the fact that there should be no spaces. 

I'm entirely unsure this is something that should be supported. 

On the other hand, however, it probably does no harm, I just can't understand the logic why would you want to support it.

Also see [Comment Policy](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/commenting-policy.md).SENTENCES rule

Stratis.Bitcoin/Configuration/TextFileConfiguration.cs
```diff
@@ -151,10 +148,13 @@ public bool Contains(string key)
         public string[] GetAll(string key)
```
Aprogiena: See [Comment Policy](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/commenting-policy.md).UPDATES rule - here we have 

```
        /// <param name="key">Name of the argument. If it is not found, an alternative argument name with additional '-' prefix will be checked.</param>
```

And this does not really reflect the reality after this PR update.

Stratis.Bitcoin/Configuration/TextFileConfiguration.cs
```diff
@@ -151,10 +148,13 @@ public bool Contains(string key)
...
         {
             List<string> values;
-            if (!this.args.TryGetValue(key, out values))
-                if (!this.args.TryGetValue($"-{key}", out values))
-                    return new string[0];
-            return values.ToArray();
+            List<string> dashValues;
+            // get the values without the - prefix
```
Aprogiena: SENTENCES rule here again and below as well.

Stratis.Bitcoin/Configuration/TextFileConfiguration.cs
```diff
@@ -151,10 +148,13 @@ public bool Contains(string key)
...
+            this.args.TryGetValue(key, out values);
+            // get the values with the - prefix
+            this.args.TryGetValue($"-{key}", out dashValues);
+            // concat those
+            return (values ?? new List<string>()).Concat(dashValues ?? new List<string>()).ToArray();
```
Aprogiena: This construction is unnecessary. It requires too much brainpower to process.

Consider this instead:
```
// Get the values from argument without '-' prefix.
if (!this.args.TryGetValue(key, out values))
    values = new List<string>();

// Get the values from argument with '-' prefix.
if (!this.args.TryGetValue($"-{key}", out dashValues))
    dashValues = new List<string>();

return values.Concat(dashValues).ToArray();
```

I also don't need the trivial "concat those" comment because the return line is readable and it is obvious.

Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Stratis.Bitcoin.Configuration;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.NodeConfiguration
+{
+    public class TextFileConfigurationTests
+    {
+
```
Aprogiena: remove this space pls

Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Stratis.Bitcoin.Configuration;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.NodeConfiguration
+{
+    public class TextFileConfigurationTests
```
Aprogiena: I know the current commenting policy prescribes comments only in main code, not tests, but I would like you to consider commenting the tests as well. Won't hold PR for that though, just for consideration.

Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+
+        [Fact]
+        public void GetAllWithArrayArgs()
+        {
+            // Arrange
+            var SUT = new TextFileConfiguration(new[] { "test" });
+            // Act
+            var parsed = SUT.GetAll("test");
```
Aprogiena: [Coding style](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/coding-style.md).#10

multiple instances in this file

Stratis.Bitcoin.Tests/NodeConfiguration/TextFileConfigurationTests.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Stratis.Bitcoin.Tests.NodeConfiguration
+{
+    public class TextFileConfigurationTests
+    {
+
+        [Fact]
+        public void GetAllWithArrayArgs()
+        {
+            // Arrange
+            var SUT = new TextFileConfiguration(new[] { "test" });
```
Aprogiena: What is SUT?

Stratis.Bitcoin/Configuration/TextFileConfiguration.cs
```diff
@@ -44,10 +40,11 @@ public TextFileConfiguration(string[] args)
             foreach (var arg in args)
```
Aprogiena: We could fix the violation of code style #10 here as a part of this PR. I.e. we are looking for 

```
foreach (string arg in args)
``` 

here.

PR #391, opened 2017-08-30
Feature/gttxout

justinpatten:


Stratis.Bitcoin/Features/Consensus/ConsensusManager.cs
```diff
@@ -57,5 +58,12 @@ public Target GetNetworkDifficulty()
             else
                 return null;
         }
+
+        /// <inheritdoc />
+        public async Task<UnspentOutputs> GetUnspentTransactionAsync(uint256 trxid)
+        {
+            var outputs =  await this.ConsensusLoop.UTXOSet?.FetchCoinsAsync(new[] { trxid });
+            return outputs == null || outputs.UnspentOutputs == null || outputs.UnspentOutputs.Length == 0 ? null : outputs.UnspentOutputs[0];
```
dangershony: Shortcuts are nice but less readable, but if you already go down that path then this would work no?
```
return outputs?.UnspentOutputs?.SingleOrDefault()
```

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
```diff
@@ -245,5 +253,18 @@ public Task RemoveForBlock(Block block, int blockHeight)
                 this.Validator.PerformanceCounter.SetMempoolDynamicSize(this.memPool.DynamicMemoryUsage());
             });
         }
+
+        /// <inheritdoc />
+        public async Task<UnspentOutputs> GetUnspentTransactionAsync(uint256 trxid)
+        {
+            var txInfo = await InfoAsync(trxid);
```
dangershony: Please read our code styling methods (available in the documentation folder)
we use this. for instance methods

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
```diff
@@ -245,5 +253,18 @@ public Task RemoveForBlock(Block block, int blockHeight)
...
+            if(txInfo == null)
+            {
+                return null;
+            }
+            var memPoolCoinView = new MempoolCoinView(this.coinView, this.memPool, this.MempoolLock, this.Validator);
```
dangershony: I think if you use MempoolCoinView you dont need the call before that InfoAsync()

Stratis.Bitcoin/Features/RPC/Controllers/FullNodeController.cs
```diff
@@ -75,6 +76,38 @@ public Task Stop()
...
+        /// <param name="includeMemPool">Whether to include the mempool</param>
+        /// <returns>The GetTxOut rpc format</returns>
+        [ActionName("gettxout")]
+        public async Task<GetTxOutModel> GetTxOut(string txid, uint vout, bool includeMemPool = true)
+        {
+            uint256 trxid;
+            if (!uint256.TryParse(txid, out trxid))
+                throw new ArgumentException(nameof(txid));
+
+            UnspentOutputs uo = null;
```
dangershony: for readability use the full param name 
```
UnspentOutputs unspentOutputs  = null;
```

Stratis.Bitcoin/Features/MemoryPool/MempoolManager.cs
justinpatten: I was doing that to get a reference to the transaction to feed in to the load method. Is there a way to load the unspent coins with just using the hash?

dangershony: ah yeah good point, come to think about it if a trx is in the mempool its outputs are probably not spent (unless its parent is in the mempool) and that will be detected by the MempoolCoinView.
So I guess you are right we want to see is the trx parents are not in the memppol so this is the best approach (else we would just return the given trx).

justinpatten: MempoolCoinView exposes a "HasCoins" method, but that doesn't grab the transaction from memory. It simply checks if it exists. The alternative is to pull some of the logic out of MempoolCoinView.LoadView and just use that in the manager. Otherwise, I think I have to leave this as is.

dangershony: yes leave it as is

Issue #401, opened 2017-09-02
[POS] 3 or more peers are required for mining, but it is not enforced


Aprogiena:
The logic in `PosMinting.cs` in `GenerateBlocks` is weird:
```
if (tryToSync)
                {
                    tryToSync = false;
                    // TODO: This condition to have at least 3 peers is disqualifying us on testnet quite often.
                    // Yet the 60 secs delay does not prevent us to mine because tryToSync will be false next time we are here
                    // unless we are completely disconnected. So this is weird logic.
                    bool fewPeers = this.connection.ConnectedNodes.Count() < 3;
                    bool lastBlockTooOld = this.chain.Tip.Header.Time < (this.dateTimeProvider.GetTime() - 10 * 60);
                    if (fewPeers || lastBlockTooOld)
                    {
                        if (fewPeers) this.logger.LogTrace("Node is connected to few peers.");
                        if (lastBlockTooOld) this.logger.LogTrace("Last block is too old, timestamp {0}.", this.chain.Tip.Header.Time);

                        Task.Delay(TimeSpan.FromMilliseconds(60000), this.nodeLifetime.ApplicationStopping).GetAwaiter().GetResult();
                        continue;
                    }
                }
```

it seems like we want at least 3 peers, but we only wait 1 minute and don't check again (previously where was not even the wait). This is also a problem on testnet, it is not unusual to be connected to just one peer there. 

What to do with it? Do we want this condition at all? It kind of makes sense (for mainnet), but current implementation does not make sense (just wait 1 minute and don't check again?).

dangershony:
This is probably related to the syncing of time with peers, maybe its not crucial thats why we wait a 1 and continue. If the node is not in the right time eventually when it will sync with peers the out of time mined blocks will be rejected.

Aprogiena:
I think the code wants to prevent us mining if we are not well connected. I'm talking solely about `feePeers` condition here. So please consider the issue only in context of that condition.

dangershony:
This is probably related to the syncing of time with peers, maybe its not crucial thats why we wait a 1 and continue. If the node is not in the right time eventually when it will sync with peers the out of time mined blocks will be rejected.

Aprogiena:
I don't understand. How is this related to time syncing? The code just looks for how many nodes are we connected to. There is no time taken in account.

Aprogiena:
[...]

PR #408, opened 2017-09-04
Bugfix/issue 389 390 and other improvements

### LINKED ISSUE #389: [POS] PosMinting.SignBlock uses wrong fields
lastCoinStakeSearchTime and LastCoinStakeSearchTime are two different fields, but it should be one value.

### LINKED ISSUE #390: [POS] Exception inside PosMining.Mine causes miner to stop
`PosMinting.Mine` creates AsyncLoop which stops when exception occurs in its code. There should be try/catch handler around, especially when the exception can be remotely enforced by a peer pretending not accepting the newly minted block from the node for 100 seconds.

Aprogiena:
**NOTE:** This builds up on https://github.com/stratisproject/StratisBitcoinFullNode/pull/398, so that PR has to be accepted first.

This issue fixes some important bugs in POS mining code, adds some TODOs for further improvements of this code and also makes some improvements to that code.

Fix #389 
Fix #390


Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -208,6 +208,9 @@ public void GenerateBlocks(WalletSecret walletSecret)
                 if (tryToSync)
                 {
                     tryToSync = false;
+                    // TODO: This condition to have at least 3 peers is disqualifying us on testnet quite often.
```
dangershony: I think the logic is to give the node more time to connect to peers, and if none found just continue with the ones we have.

There should also be a limitation where staking is not allowed if not connected to any peer (not sure where it is)

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -497,12 +497,18 @@ private bool SignBlock(List<StakeTx> stakeTxes, Block block, ChainedBlock pindex
...
+                long minimumAllowedTime = pindexBest.Header.Time + 1;
 
-                for (uint n = 0; n < Math.Min(nSearchInterval, maxStakeSearchInterval) && !fKernelFound; n++)
+                for (uint n = 0; (n < nSearchInterval) && !fKernelFound; n++)
                 {
                     uint txTime = txNew.Time - n;
+
+                    // Once we reach previous block time + 1, we can't go any lower
+                    // because it is required that the block time is greater than the previous block time.
+                    if (txTime < minimumAllowedTime)
```
dangershony: This is a change in functionality, its interesting.
Implications a search will last much longer then 60 iterations.
That means the the current output is probably used for longer periods, while before it would move on to the next output to find a stake after 60 attempts.

Note  this line above (was not taken from stratisx) higher value outputs are selected first.
>setCoins = setCoins.OrderByDescending(o => o.TxOut.Value).ToList();

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -149,10 +149,17 @@ public Task Mine(WalletSecret walletSecret)
                 {
                     // Application stopping, nothing to do as the loop will be stopped.
                 }
+                catch (MinerException me)
+                {
```
dangershony: This is good

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -495,20 +495,28 @@ private bool SignBlock(List<StakeTx> stakeTxes, Block block, ChainedBlock pindex
             // it sorts the coins from heighest weight.
             setCoins = setCoins.OrderByDescending(o => o.TxOut.Value).ToList();
 
-            this.logger.LogTrace("Trying to find staking solution among {0} transactions.", setCoins.Count);
+            long minimalAllowedTime = pindexBest.Header.Time + 1;
```
dangershony: Very nice this can potentially even go in the top of this method (it can even go in the call before setCoins is created)

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -215,7 +215,7 @@ public void GenerateBlocks(WalletSecret walletSecret)
                     // unless we are completely disconnected. So this is weird logic.
                     bool fewPeers = this.connection.ConnectedNodes.Count() < 3;
                     bool lastBlockTooOld = this.chain.Tip.Header.Time < (this.dateTimeProvider.GetTime() - 10 * 60);
-                    if (fewPeers || lastBlockTooOld)
+                    if ((fewPeers && !this.network.IsTest()) || lastBlockTooOld)
```
dangershony: I dont think we need this. take it away now and we can discuss later.

Stratis.Bitcoin/Features/Consensus/ConsensusFeature.cs
```diff
@@ -180,7 +180,9 @@ private void RunLoop()
 
                 // TODO Need to revisit unhandled exceptions in a way that any process can signal an exception has been
                 // thrown so that the node and all the disposables can stop gracefully.
+                this.logger.LogDebug("Exception occurred in consensus loop: {0}", ex.ToString());
                 this.logger.LogCritical(new EventId(0), ex, "Consensus loop unhandled exception (Tip:" + this.consensusLoop.Tip?.Height + ")");
+                NLog.LogManager.Flush();
```
dangershony: This is not the right place to do that.

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
```diff
@@ -10,6 +10,23 @@
 
 namespace Stratis.Bitcoin.Features.Consensus
 {
+    /// <summary>
```
dangershony: I am happy to document this here for now.
though my plan in the future is to break all the rules to dedicated classes.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -500,16 +500,16 @@ private bool SignBlock(List<StakeTx> stakeTxes, Block block, ChainedBlock pindex
 
                 for (uint n = 0; n < Math.Min(nSearchInterval, maxStakeSearchInterval) && !fKernelFound; n++)
                 {
+                    uint txTime = txNew.Time - n;
+                    if ((txTime & PosConsensusValidator.StakeTimestampMask) != 0)
```
dangershony: you forgot (~)
```
 if ((txTime & ~PosConsensusValidator.StakeTimestampMask) != 0)
```

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: This has its own issue now - https://github.com/stratisproject/StratisBitcoinFullNode/issues/401
as explained there, the problem is that this delay is happening after every new block, not only once during startup

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -145,7 +145,7 @@ public Task Mine(WalletSecret walletSecret)
                 {
                     this.GenerateBlocks(walletSecret);
                 }
-                catch (OperationCanceledException)
+                catch (TaskCanceledException)
```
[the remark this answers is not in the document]
Aprogiena: TaskCanceledException is the correct one [according to MSDN](https://msdn.microsoft.com/en-us/library/hh194845%28v=vs.110%29.aspx)

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: Yes this is done in followup PR

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: This causes problems when developing on testnet. And I'm not sure how to make a partial change in file that I would avoid being committed to the repository. Will discuss on slack

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: No :) 

The mask is there to only allow tiestamps divisible by 16. 
I'm here skipping such a timestamp that is not divisible by 16.
You would skip timestamps that are greater than 15.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -141,7 +141,27 @@ public Task Mine(WalletSecret walletSecret)
...
+                    // Application stopping, nothing to do as the loop will be stopped.
+                }
+                catch (ConsensusErrorException cee)
+                {
+                    // All consensus exceptions should be ignored. It means that the miner 
+                    // possibly mined a block that was not accepted or is even invalid,
+                    // but it should not halted the mining operation.
+                    this.logger.LogDebug("Consensus error exception occurred in miner loop: {0}", cee.ToString());
+                }
+                catch (Exception e)
```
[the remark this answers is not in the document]
Aprogiena: Yes, this can be simplified, I'll do it in followup PR

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
Aprogiena: Still the documentation should be in one place. I would like to avoid having to go to 20 files to be able to find out all rules that apply.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: Ignorance :), will put it back

Stratis.Bitcoin/Features/Consensus/ConsensusFeature.cs
Aprogiena: this leads to app crash, if i don't do it here i don't do it ever
this is not elegant, but since we are using exceptions, there is little we can do about it 

this is the highest level in the thread running this code, so there is nowhere else this could be done

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: More change in functionality coming in following PR as well.
Yes, previously the backward search was limited to 60 seconds, now I'm allowing unlimited, but this is again limited by the timestamp of the latest block. But according to my current understanding of the consensus rules, there is no 60 seconds limit for backward search on consensus level, so this should be valid.

Yes, I know about the sort, it is minor optimization preventing doing extra work. Very tiny though.

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
dangershony: yes thats true I would even pull it out to its own readme.

Stratis.Bitcoin/Features/Consensus/ConsensusFeature.cs
dangershony: hmm I didn't understand can you not just do it in the ConsensusFeatire.Stop()?

Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: ok yeah I think there is no limitation in consensus about backward search just make a mental point this is different in case we come up with issues.

Maybe we can keep a list of logic that is different from X in just in case.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: ok lets discuss on slack

Stratis.Bitcoin/Features/Consensus/ConsensusFeature.cs
Aprogiena: no because exception here crashes the node, there is no normal shutdown, this is hard crash

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I don't really understand here. If not for optimization then for what? The only reason could be optimization for that sort. Yes with sorted list , you have greater chance to pick the right UTXO earlier instead of later. This saves some work, not much though, but still good.

The code will diverge from X very quickly, esp. with my next PR, so I don't know, I think it is important to document all consensus requirements, not really the implementation. This should not be like Bitcoin.

Issue #413, opened 2017-09-05
[POS] CheckSync is not implemented in C#


Aprogiena:
`Check against synchronized checkpoint` is in stratisX which prevents reorgs longer than 500 blocks

dangershony:
Ah ok yeah we need to add that.

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[POS] CheckSync is not implemented in C#](https://trello.com/c/mzfQRdAl/269-pos-checksync-is-not-implemented-in-c)

dangershony:
I would suggest we put this limit in to the `ChainHeadersBehavior`
Just before this part
```
	if (this._PendingTip.ChainWork > this.Chain.Tip.ChainWork)
	{
		this.Chain.SetTip(this._PendingTip);
	}
```

Aprogiena:
Yes, but I'd still like to have the test for this working before I finish this.

dangershony:
Ah the test is working do you have issues with the test?

Aprogiena:
Yes, from day 1 the test does not work for me.

dangershony:
Ah ok I realise that's because we need to pull a fix from NStratis.

dangershony:
This test now works fine.

Aprogiena:
Yes, it is now better, except that it does not use ConsensusOptions.MaxReorgLength, which it actually should use. 

Another thing is that the test don't pass even if I set MaxReorgLength manually to 10.

Aprogiena:
So, could you please check the latest version of #492?

Aprogiena:
I will probably introduce logs to ChainHeadersBehavior to see what's happening.

Aprogiena:
I mean, we certainly need that test to use ConsensusOptions - so I'd like to ask you to update the test code. And my task is to see why it does not work. I will analyze that and let you know.

Aprogiena:
OK, so I've advanced enough with the implementation (PR #492), so I believe that PR implements long reorg protection well now. However, the test `Given__NodesAreSynced__When__ABigReorgHappens__Then__TheReorgIsIgnored` does not currently work with this implementation. The test needs to be fixed:

1) There are two mining nodes and coincidentally as their mining tasks are run in parallel, they actually mine the same blocks. This prevents the two mined chains to be different enough (above the max reorg length). Therefore the reorg is accepted.

This seems to have an easy fix:
```
[...]

PR #415, opened 2017-09-05
Bugfix/issue 393

### LINKED ISSUE #393: [POS] CreateCoinStake loop looking for staking solution could be improved
The loop searches backwards for solutions with older times and the timestamp mask check is wrongly placed.

Aprogiena:
Addresses series of issues in POS alg implementation. It does diverge from StratisX implementation heavily now. Also cleansup the code a little bit and adds some comments.

Fix #393


Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -626,97 +629,72 @@ private bool SignBlock(List<StakeTx> stakeTxes, Block block, ChainedBlock pindex
...
+                    }
                 }
             }
 
             // Calculate coin age reward.
-            // TODO: This is not used  anywhere, delete it?
-            ulong nCoinAge;
-            if (!this.posConsensusValidator.StakeValidator.GetCoinAge(this.chain, this.coinView, txNew, pindexPrev, out nCoinAge))
-                return false; //error("CreateCoinStake : failed to calculate coin age");
+            if (!this.posConsensusValidator.StakeValidator.GetCoinAge(this.chain, this.coinView, coinstakeTx, chainTip, out ulong coinAge))
```
dangershony: FYI for your knowledge coin age is not use in this version of POS anymore, so I am not sure if this condition is required (I think its not checked in consensus)

For reference:
https://blackcoin.co/blackcoin-pos-protocol-v2-whitepaper.pdf

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -820,97 +798,73 @@ public Money GetBalance(List<StakeTx> stakeTxes)
...
+        /// <para>
+        /// Such a coin has to be confirmed with enough confirmations - i.e. has suitable depth,
+        /// and it also has to be mature and meet requirement for minimal value.
+        /// </para>
+        /// </summary>
+        /// <param name="stakeTxes">List of coins that are candidates for being used for staking.</param>
+        /// <param name="spendTime">Timestamp of the coinstake transaction.</param>
+        /// <param name="maxValue">Maximal amount of coins that can be used for staking.</param>
+        /// <returns>List of coins that meet the requirements.</returns>
+        private List<StakeTx> SelectSuitableCoinsForStaking(List<StakeTx> stakeTxes, uint spendTime, long maxValue)
```
dangershony: I would actually rename this to FindCoinsForStaking

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -820,97 +798,73 @@ public Money GetBalance(List<StakeTx> stakeTxes)
...
-                if (this.GetBlocksToMaturity(pcoin) > 0)
+                if (this.GetBlocksToMaturity(stakeTx) > 0)
                 {
-                    this.logger.LogTrace("UTXO '{0}/{1}' can't be added because it is not mature.", pcoin.OutPoint.Hash, pcoin.OutPoint.N);
+                    this.logger.LogTrace("UTXO '{0}/{1}' can't be added because it is not mature.", stakeTx.OutPoint.Hash, stakeTx.OutPoint.N);
                     continue;
                 }
 
-                if (pcoin.TxOut.Value >= this.minimumInputValue)
+                if (stakeTx.TxOut.Value < this.minimumInputValue)
```
dangershony: this was ">=" any reason its changed?

dangershony: Ah ok you swapped the else sorry didn't see that. [ignore]

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
```diff
@@ -14,17 +14,30 @@ namespace Stratis.Bitcoin.Features.Consensus
...
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
-    /// <item>TODO</item>
+    /// <item>New block's timestamp ('BlockTime') MUST be strictly greater than previous block's timestamp.</item>
+    /// <item>Coinbase transaction's (first transaction in the block with no inputs) timestamp MUST be inside interval ['BlockTime' - 15; 'BlockTime'].</item>
```
dangershony: I would avoid putting hard limits, those can be configured in the protocol options, you can specify that as an example (i.e in stratis block time mast be ['BlockTime' - 15; 'BlockTime'])

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: Yes I agree with you but did not actually put any serious time into exploring if i can completely remove that. I'll create issue

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
```diff
@@ -14,17 +14,30 @@ namespace Stratis.Bitcoin.Features.Consensus
...
+    /// <item>Coinstake transaction's (second transaction in the block with at least one input and at least 2 outputs and first output being empty) timestamp
+    /// MUST be equal to 'BlockTime' and it MUST have lower 4 bits set to 0 (i.e. be divisible by 16) - see <see cref="StakeTimestampMask"/>.</item>
+    /// <item>Block's header 'nBits' field MUST be set to the correct POS target value.</item>
```
dangershony: is it not more correc tot say "the 'nBits' must be bellow the target"

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: OK, will do. For me this is rather selection than finding, but no problemo.

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
```diff
@@ -14,17 +14,30 @@ namespace Stratis.Bitcoin.Features.Consensus
...
+    /// i.e. it MUST have that many confirmation at least.</item>
+    /// <item>Coinstake transaction's kernel must meet the staking target using this formula:
+    /// <code>hash(nStakeModifier + txPrev.block.nTime + txPrev.nTime + txPrev.vout.hash + txPrev.vout.n + nTime) < bnTarget * nWeight</code>
+    /// <para>
+    /// where 'txPrev' is the coinstake's kernel transaction, 'txPrev.vout' is the kernel's output in that transaction, 
+    /// 'txPrev.vout.hash' is the hash of that transaction; 'nTime' is coinstake's transaction time; 'bnTarget' is the target as 
+    /// in 'nBits' block header; 'nWeight' is the value of the kernel's input. 
+    /// </para>
+    /// </item>
+    /// <item>Block's height MUST NOT be more than 500 blocks back - i.e. reorganizations longer than 500 are not allowed.</item>
```
dangershony: >Block's height MUST NOT be more than 500 blocks back

Is that enforced in the rules?

Aprogiena: It is not (in C#) - https://github.com/stratisproject/StratisBitcoinFullNode/issues/413

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
Aprogiena: hmm, I think nBits defines the target and the coinstake hash must be below that. So I would say it would be incorrect. but am I right?

PR #433, opened 2017-09-07
Feature/peer time sync issue 382

### LINKED ISSUE #382: [POS] Time syncing with peers
StratisX implements time syncing with peers, this is then used for POS block validation.
C# node does not do that at the moment, it needs to be ported.

### LINKED ISSUE #383: [POS] POS Block time validation
C# node is currently using different (old) mechanism of validating POS block than StratisX. The relevant function is around `FutureDrift`.

Aprogiena:
Time sync with peers + tests. Also new FutureDrift to behave like StratisX.

Be very careful as this changes consensus code! I've spend more than 16 hours testing this, but it needs more. 

Fix #382 
Fix #383


Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -257,7 +257,7 @@ public void GenerateBlocks(WalletSecret walletSecret)
                 // TODO: This has to be changed once 
                 // https://github.com/stratisproject/StratisBitcoinFullNode/issues/382
                 // is implemented.
-                uint coinstakeTimestamp = Utils.DateTimeToUnixTime(DateTime.UtcNow) & ~PosConsensusValidator.StakeTimestampMask;
+                uint coinstakeTimestamp = (uint)this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp() & ~PosConsensusValidator.StakeTimestampMask;
```
dangershony: you can remove the todo above this line

Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -117,7 +134,10 @@ public override void Start()
             connectionParameters.IsRelay = !this.nodeSettings.ConfigReader.GetOrDefault("blocksonly", false);
             connectionParameters.TemplateBehaviors.Add(new ChainHeadersBehavior(this.chain, this.chainState));
             connectionParameters.TemplateBehaviors.Add(new AddressManagerBehavior(this.addressManager));
+            var timeSyncBehaviorState = new TimeSyncBehaviorState(this.dateTimeProvider, this.nodeLifetime, this.asyncLoopFactory, this.loggerFactory);
```
dangershony: Just a suggestion you can let TimeSyncBehaviorState be injected in the constructor of the feature.

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
```diff
@@ -322,10 +330,14 @@ private static long FutureDriftV1(long nTime)
             return nTime + 10 * 60;
         }
 
-        // TODO: https://github.com/stratisproject/StratisBitcoinFullNode/issues/383
+        private static bool IsDriftReduced(long nTime)
```
dangershony: This was introduced as a fix to address the specific issue of the stratis chain (future drift bug)
Bear in mind as for other chain it wont be necessarily to make the check.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: of course :) tx

Stratis.Bitcoin/Base/BaseFeature.cs
Aprogiena: Do you mean to treat the TSBState just like chainRepository in this code? I.e. call the constructor inside `UseBaseFeature`?

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
Aprogiena: Sorry, I don't understand what you mean. Are you talking about BitcoinD node? If so, this code is only called from PosConsensusValidator, which is not part of Bitcoin chain validation. Or are you talking about something else?

dangershony: No I mean this condition is only correct for the stratis chain, on stratis testnet for example it is not needed, as we dont have blocks with the wrong future drift.

Stratis.Bitcoin/Base/BaseFeature.cs
dangershony: Almost, you dont need to call the constructor just register the type tot he DI and it will be done for you.

Stratis.Bitcoin/Features/Consensus/TimeSyncBehavior.cs
```diff
@@ -44,10 +44,10 @@ public struct TimestampOffsetSample
         }
 
         /// <summary>Maximal number of samples to keep inside <see cref="inboundTimestampOffsets"/>.</summary>
-        private const int MaxInboundSamples = 200;
+        public const int MaxInboundSamples = 200;
```
dangershony: This is fine, are you are aware of the InternalVisibleTo attribute right?

Stratis.Bitcoin/Features/Consensus/TimeSyncBehavior.cs
```diff
@@ -0,0 +1,458 @@
...
+            this.asyncLoopFactory = asyncLoopFactory;
+
+            this.inboundTimestampOffsets = new CircularArray<TimestampOffsetSample>(MaxInboundSamples);
+            this.outboundTimestampOffsets = new CircularArray<TimestampOffsetSample>(MaxOutboundSamples);
+            this.inboundSampleSources = new HashSet<IPAddress>();
+            this.outboundSampleSources = new HashSet<IPAddress>();
+
+            this.timeOffset = TimeSpan.Zero;
+            this.WarningLoopStarted = false;
+            this.SwitchedOff = false;
```
dangershony: What about adding this as a configuration value? -synctime

dangershony:
@Aprogiena I am going to commit this (I want people to strat using it)
The other comments you can do as a separate PR.

Stratis.Bitcoin/Features/Consensus/PosConsensusValidator.cs
Aprogiena: Yes, I've checked it against testnet, it works correctly as is because our testnet is younger than the condition timestamp.

Stratis.Bitcoin/Features/Consensus/TimeSyncBehavior.cs
Aprogiena: yes, found it ugly and never used :-)
but maybe i should reconsider in the future

Stratis.Bitcoin/Features/Consensus/TimeSyncBehavior.cs
Aprogiena: yes, this is exactly prepared for that further development
i did not have enough time here to implement -synctime as a part of this one
but it should now be easy, the logic is basically there

Stratis.Bitcoin/Features/Consensus/TimeSyncBehavior.cs
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        /// <inheritdoc />
+        public bool AddTimeData(IPAddress peerAddress, TimeSpan offsetSample, bool isInboundConnection)
+        {
+            this.logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(peerAddress), peerAddress, nameof(offsetSample), offsetSample.TotalSeconds, nameof(isInboundConnection), isInboundConnection);
+
+            bool res = false;
+
+            bool startWarningLoopNow = false;
```
dangershony: This is a classic double lock scenario
 ```
if (!this.SwitchedOff)
{
      lock (this.lockObject)
      {
             if (!this.SwitchedOff)
             {
```

Aprogiena: tx, https://github.com/stratisproject/StratisBitcoinFullNode/issues/461

PR #448, opened 2017-09-12
Cli daemon 

dangershony:


bokobza:
looking at this now

Stratis.Cli/Program.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Linq;
+using NBitcoin;
+using NBitcoin.RPC;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Features.RPC;
+
+namespace Stratis.Cli
```
bokobza: what do you think of calling this Stratis.Bitcoin.Cli or maybe Stratis.Bitcoin.RPC.Cli?
I find Stratis.Cli a bit too generic.

bokobza: You said this is just a start so I'm not going to review the code here.

PR #454, opened 2017-09-13
[WatchOnlyWallet] Adding id and Merkle proof data in transactions

bokobza:
@zeptin here is your change
@dangershony


dangershony:
got it

PR #464, opened 2017-09-18
Pending Storage Performance Improvements

fassadlr:
https://trello.com/c/8Y2OuIfh/221-blockstore-2-process-pending-storage-performance


Stratis.Bitcoin.Features.BlockStore/LoopSteps/CheckNextChainedBlockExistStep.cs
```diff
@@ -31,10 +31,10 @@ internal CheckNextChainedBlockExistStep(BlockStoreLoop blockStoreLoop, ILoggerFa
         /// <inheritdoc/>
         internal override async Task<StepResult> ExecuteAsync(ChainedBlock nextChainedBlock, CancellationToken cancellationToken, bool disposeMode)
         {
-            this.logger.LogTrace("({0}:'{1}/{2}',{3}:{4})", nameof(nextChainedBlock), nextChainedBlock?.HashBlock, nextChainedBlock?.Height, nameof(disposeMode), disposeMode);
-
             if (await this.BlockStoreLoop.BlockRepository.ExistAsync(nextChainedBlock.HashBlock))
             {
+                this.logger.LogTrace("({0}:'{1}/{2}',{3}:{4})", nameof(nextChainedBlock), nextChainedBlock?.HashBlock, nextChainedBlock?.Height, nameof(disposeMode), disposeMode);
+
```
Aprogiena: Why this change? First, the log should be on the method entry, not after the first if. Second, there should be log on each method exit.

Stratis.Bitcoin.Features.BlockStore.Tests/LoopTests/BlockStoreLoopStepBaseTest.cs
```diff
@@ -60,11 +60,13 @@ internal List<Block> CreateBlocks(int amount)
             return blocks;
         }
 
-        internal Block CreateBlock(int blockNumber)
+        internal Block CreateBlock(int blockNumber, bool bigBlocks = false)
         {
             var block = new Block();
 
-            for (int j = 0; j < 10; j++)
+            var transactions = bigBlocks ? 1000 : 10;
```
Aprogiena: `int transactionCount` would be better. Also why to replace int with var?

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,149 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
         /// <inheritdoc/>
         internal override async Task<StepResult> ExecuteAsync(ChainedBlock nextChainedBlock, CancellationToken cancellationToken, bool disposeMode)
         {
-            this.logger.LogTrace("({0}:'{1}/{2}',{3}:{4})", nameof(nextChainedBlock), nextChainedBlock.HashBlock, nextChainedBlock.Height, nameof(disposeMode), disposeMode);
```
Aprogiena: Not having logs in ExecuteAsync methods of steps seems weird. It is exactly where you want to have logs if you want to be able to track which steps were executed.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,149 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
-                return true;
-            }
+            this.BlockStoreLoop.SetStoreTip(context.PendingBlockPairsToStore.First().ChainedBlock);
 
-            return false;
+            this.logger.LogTrace("({0}:{1} / {2}.{3}:{4} / {5}:{6} / {7}:{8})",
+                nameof(context.BlockStoreLoop.ChainState.IsInitialBlockDownload), context.BlockStoreLoop.ChainState.IsInitialBlockDownload,
+                nameof(context.PendingBlockPairsToStore), nameof(context.PendingBlockPairsToStore.Count),
+                context.PendingBlockPairsToStore?.Count, nameof(context.PendingStorageBatchSize), context.PendingStorageBatchSize,
+                nameof(context.BlockStoreLoop.StoreTip), context.BlockStoreLoop.StoreTip);
```
Aprogiena: This is wrong use of logging. If you want to log some particular state, not just method inputs, don't use method entry log - i.e. the one with `()`, instead use a descriptive log written in English. Also when using method entry logs, pls follow the same log style (i.e. not using ` / ` as delimiters).

Also consider overriding ToString of ProcessPendingStorageContext

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,149 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
 
-            if ((context.InputChainedBlock != null && (context.NextChainedBlock.Header.HashPrevBlock != context.InputChainedBlock.HashBlock)))
-            {
-                this.logger.LogTrace("{0}:'{1}/{2}' [INVALID_PREVIOUS]", nameof(context.NextChainedBlock.Header.HashPrevBlock), context.NextChainedBlock.Header.HashPrevBlock, context.InputChainedBlock.HashBlock);
-                return true;
-            }
+        /// <summary>
+        /// Store missing blocks and remove them from pending blocks and set the Store's tip to <see cref="ProcessPendingStorageContext.NextChainedBlock"/>
+        /// </summary>
+        private async Task PushBlocksToRepository(ProcessPendingStorageContext context)
```
Aprogiena: argument comment missing

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -189,12 +205,27 @@ internal ProcessPendingStorageContext(BlockStoreLoop blockStoreLoop, ChainedBloc
...
         /// <summary>
-        /// Gets the next block to process.
+        /// Break execution if:
+        /// <list>
+        ///     <item>1: At the tip</item>
+        ///     <item>2: Block is already in store or pending insertion</item>
+        /// </list>
         /// </summary>
-        internal void GetNextBlock()
+        internal bool CanProcessNextBlock()
```
Aprogiena: return value comment missing and SENTENCES

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ReorganiseBlockRepositoryStep.cs
```diff
@@ -41,14 +41,12 @@ internal ReorganiseBlockRepositoryStep(BlockStoreLoop blockStoreLoop, ILoggerFac
         /// <inheritdoc/>
         internal override async Task<StepResult> ExecuteAsync(ChainedBlock nextChainedBlock, CancellationToken cancellationToken, bool disposeMode)
         {
-            this.logger.LogTrace("({0}:'{1}/{2}',{3}:{4})", nameof(nextChainedBlock), nextChainedBlock.HashBlock, nextChainedBlock.Height, nameof(disposeMode), disposeMode);
-
```
Aprogiena: same here as above

Stratis.Bitcoin.Features.BlockStore.Tests/LoopTests/BlockStoreLoopStepBaseTest.cs
fassadlr: Fixed

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: Yes, this needs to be corrected. I moved the log output to each method in ProcessPending (ProcessWhenDisposing, ProcessWhenInInIBD and ProcessWhenNotInIBD) so that I can which one executed. It is more meaningful now. I also removed the method entry syntax. I'm only concerned with which method executed.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/CheckNextChainedBlockExistStep.cs
fassadlr: I really only want the log to output something if the logic contained in it became applicable. I'll remove the method entry format from the log. I ran the node with the log and it output thousands of useless method entry and exit logs which wasn't useful at all.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: ToString for ProcessPendingStorageContext is a great idea. I'll implement that. Thanks!

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: Will fix

Stratis.Bitcoin.Features.BlockStore/LoopSteps/CheckNextChainedBlockExistStep.cs
Aprogiena: Then consider DEBUG level, but don't remove or change TRACE level. I understand that for lot of tasks, you are not interested in seeing entry/exit logs, but for other tasks, other devs might, so if you want to have something cherrypicked, use DEBUG level and then modify your NLog.config in a way that you only see DEBUG level from that module. This will perfectly fit what you need while you won't harm the general trace logging.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: Will fix

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
Aprogiena: See above comment about DEBUG level rather than TRACE level for this.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ReorganiseBlockRepositoryStep.cs
fassadlr: Same comment as for CheckExistsInPending step...

Stratis.Bitcoin.Features.BlockStore/LoopSteps/CheckNextChainedBlockExistStep.cs
fassadlr: Ok yes, that makes sense! Ok Ive added them back and changed the other ones to Debug.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
+            if (disposeMode)
+                return await ProcessWhenDisposing(context);
+
+            if (this.BlockStoreLoop.ChainState.IsInitialBlockDownload)
+                return await ProcessWhenInIBD(context);
+
+            if (!this.BlockStoreLoop.ChainState.IsInitialBlockDownload)
+                return await ProcessWhenNotInIBD(context);
+
+            return StepResult.Stop;
```
dangershony: Seems we don't need this return 'StepResult.Stop;'

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -189,12 +214,39 @@ internal ProcessPendingStorageContext(BlockStoreLoop blockStoreLoop, ChainedBloc
...
             this.InputChainedBlock = this.NextChainedBlock;
             this.NextChainedBlock = this.BlockStoreLoop.Chain.GetBlock(this.NextChainedBlock.Height + 1);
+
+            if (this.NextChainedBlock == null)
+                return false;
+            if (this.NextChainedBlock.Header.HashPrevBlock != this.InputChainedBlock.HashBlock)
+                return false;
+            if (this.NextChainedBlock.Height > this.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
+                return false;
+            if (this.BlockStoreLoop.StoreTip.Height >= this.BlockStoreLoop.ChainState.HighestValidatedPoW?.Height)
```
dangershony: Do we need this check?
We know that InputChainedBlock is equal to StoreTip right?

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
+                PrepareNextBlockFromPendingStorage(context);
+
+                await PushBlocksToRepository(context);
+
+                var canProcessNextBlock = context.CanProcessNextBlock();
+                if (canProcessNextBlock == false)
+                    break;
 
-            this.logger.LogTrace("({0}:{1} / {2}.{3}:{4} / {5}:{6}')", nameof(context.BlockStoreLoop.ChainState.IsInitialBlockDownload), context.BlockStoreLoop.ChainState.IsInitialBlockDownload, nameof(context.PendingBlockPairsToStore), nameof(context.PendingBlockPairsToStore.Count), context.PendingBlockPairsToStore?.Count, nameof(context.PendingStorageBatchSize), context.PendingStorageBatchSize);
+            } while (context.CancellationToken.IsCancellationRequested == false);
```
dangershony: No logical reason to use do;while as we are only checking IsCancellationRequested so better check it before rather then after.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
-                        return StepResult.Next;
+                PrepareNextBlockFromPendingStorage(context);
 
-                    if (context.PendingBlockPairsToStore.Any())
-                    {
-                        this.logger.LogTrace("[FLUSH_NEXTBLOCKNOTINPENDING]", nameof(context.PendingStorageBatchSize), context.PendingStorageBatchSize);
-                        await this.PushBlocksToRepository(context);
+                var canProcessNextBlock = context.CanProcessNextBlock();
+                if (canProcessNextBlock == false)
+                    if (!context.PendingBlockPairsToStore.Any())
```
dangershony: I dont understand this condition 
```
if (canProcessNextBlock == false)
    if (!context.PendingBlockPairsToStore.Any())
        break;
```

if `PendingBlockPairsToStore` has items in it we will not exit the loop and continue to drain `PendingStorage` however in that case all the blocks will return canProcessNextBlock = false right? so there is no need to continue the loop. (please confirm my logic)

I would say: 
```
if(canProcessNextBlock == false)
  break;
```

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
+        /// </summary>
+        private async Task<StepResult> ProcessWhenInIBD(ProcessPendingStorageContext context)
+        {
+            this.logger.LogDebug("{0}:'{1}/{2}'", nameof(context.NextChainedBlock), context.NextChainedBlock.HashBlock, context.NextChainedBlock.Height);
+
+            if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold)
+                return StepResult.Continue;
 
-                context.GetNextBlock();
+            do
```
dangershony: no need for do;while see other comment

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
                 {
                     if (context.PendingBlockPairsToStore.Any())
-                        await this.PushBlocksToRepository(context);
+                        await PushBlocksToRepository(context);
 
                     break;
                 }
-            }
+
+                if (context.PendingStorageBatchSize > BlockStoreLoop.MaxPendingInsertBlockSize)
```
dangershony: Excellent

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -43,132 +42,158 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
         {
-            await this.BlockStoreLoop.BlockRepository.PutAsync(context.PendingBlockPairsToStore.First().ChainedBlock.HashBlock, context.PendingBlockPairsToStore.Select(b => b.Block).ToList());
-            this.BlockStoreLoop.SetStoreTip(context.PendingBlockPairsToStore.First().ChainedBlock);
+            this.logger.LogDebug("{0}:'{1}/{2}'", nameof(context.NextChainedBlock), context.NextChainedBlock.HashBlock, context.NextChainedBlock.Height);
+
+            do
+            {
+                PrepareNextBlockFromPendingStorage(context);
+
+                await PushBlocksToRepository(context);
```
dangershony: Danger of pushing a block out of sync 
I think we have to put the `context.CanProcessNextBlock()` here as well before we push.

We could also take the same functionality from the IBD loop they are very similar (apart for the threshold condition)
IBD is time based  so  sometimes there could be over 100 blocks while not in IBD.

Basically you could have one method for both (reuse the logic from the IBD method) and when in IBD call it only if the threshold is correct for example
```
ProcessWhenInIBD()
{
  if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold)
                return StepResult.Continue;
  
        this.ProcessWhenNotInIBD()
}
```

What do you think?
(you could then rename ProcessWhenNotInIBD to ProcessesBatch())

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: Have fixed this with en "else"

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: Yes I agree with you, that extra check is not necessary, we need to exit the loop if we cant process the next block.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: You are correct, I have removed this check.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: OK, I had a look through the code you wrote first, way back when, and there we did first check the cancellation token before processing or pushing anything to the repo, I've swopped this around.

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: As per your other comment I have changed this to a `while { }`

Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
fassadlr: OK yes, I refactored a bit.. and now see that the code is the same, apart from the max size threshold. I've amended this now.. will commit soon :D

fassadlr: @dangershony I've rethought your suggestion above and actually still feel that the logic for InIBD and NotInIBD should be in separate methods. Its easier to test and read that way cause it very clearly encapsulates the logic in one method for each scenario. Its just that otherwise, you need to start introducing more "If" statements again for IBD and disposing and thresholds etc. I like to code in a way that keeps the logic simple and easily readable  (Single responsibility).

I do however agree with you that there is some slight logic duplication and have done some work to fix it, however, I feel for sake of readability and keeping logic separate, its not that bad.

dangershony: Sure there is no rule for code against code duplication when it intentional, my point was just that the code in both cases needs to be similar.

fassadlr:
[[BlockStore] 2 Process Pending Storage Performance](https://trello.com/c/8Y2OuIfh/221-blockstore-2-process-pending-storage-performance)

PR #466, opened 2017-09-18
Load the miner settings

dangershony:


Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -49,6 +52,8 @@ public override void Start()
 
                 if (!string.IsNullOrEmpty(minto))
                 {
+                    this.logger.LogInformation($"Mining enabled.");
```
Aprogiena: i quite dislike using interpolated strings where it is actually not used at all inside that particular string

also, do not use interpolated string in logging

Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -58,6 +63,8 @@ public override void Start()
                 if (!string.IsNullOrEmpty(this.minerSettings.WalletName)
                     && !string.IsNullOrEmpty(this.minerSettings.WalletPassword))
                 {
+                    this.logger.LogInformation($"Staking enabled on wallet {this.minerSettings.WalletName}.");
```
Aprogiena: no interpolated strings in logging

Stratis.Bitcoin.Features.Miner/MiningFeature.cs
dangershony: Ah but this is not perf critical (but ok)

PR #469, opened 2017-09-19
Optimize feerate query for bad internet

nopara73:


nopara73:
There should not be any conflict, I created the branch from the latest master. What's going on?

nopara73:
Context: Before, when the user's internet was down, it waited 3 to 10 minutes randomly to update the feerates the wallet is using.  
This is fine, but if it doesn't get the feerates the first time, then when the wallet needs it, it will not have it.  
Modified so the feerate queries must happen every 3 seconds (instead of 3 to 10min) until the feerate is acquired the first time. This happened with Carlton when he was using unreliable internet 4 days ago and we fixed it in a similar way (inside the tb-alpha branch's Alpha folder), so it's tested.

Stratis.Bitcoin.Features.LightWallet/LightWalletFeePolicy.cs
```diff
@@ -15,10 +15,11 @@ namespace Stratis.Bitcoin.Features.LightWallet
     public class LightWalletFeePolicy : IWalletFeePolicy
     {
         private readonly Money maxTxFee;
-        private readonly HttpClient httpClient;
+        private static readonly HttpClient HttpClient = new HttpClient();
```
bokobza: We try as much as possible to avoid initializing fields outside of the constructor.
It's a coding preference, but it prevents the debugger to jump up and down during debugging.
Look at item 17 of this doc: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/coding-style.md

Stratis.Bitcoin.Features.LightWallet/LightWalletFeePolicy.cs
```diff
@@ -29,29 +30,63 @@ public LightWalletFeePolicy(IAsyncLoopFactory asyncLoopFactory, INodeLifetime no
             this.highTxFeePerKb = null;
```
bokobza: I think these are null by default, no need to initialize them here (although the compiler might disagree).

Stratis.Bitcoin.Features.LightWallet/LightWalletFeePolicy.cs
```diff
@@ -29,29 +30,63 @@ public LightWalletFeePolicy(IAsyncLoopFactory asyncLoopFactory, INodeLifetime no
...
+                catch (Exception ex) when (ex is HttpRequestException || ex is TimeoutException)
+                {
+                    response = null;
+                }
+                if (response == null || !response.IsSuccessStatusCode)
+                {
+                    // If it's already been initialized once just keep using the feerate we already have
+                    if (this.initializedOnce)
+                    {
+                        await Task.Delay(TimeSpan.FromMinutes(waitMinutes), token).ContinueWith(t => { }).ConfigureAwait(false);
```
bokobza: can't this be just 
`await Task.Delay(TimeSpan.FromMinutes(waitMinutes), token).ConfigureAwait(false);`

bokobza:
There is a merge conflict as well, can you please check it out?

Stratis.Bitcoin.Features.LightWallet/LightWalletFeePolicy.cs
nopara73: To swallow exception

PR #471, opened 2017-09-19
Take puller one block back when syncing

bokobza:
Fix for a bug that happened when syncing from a height or a hash.
When asked to sync, the block puller was starting to sync from the next block.
Now when it's asked to sync from height h, we move the puller to height h-1 so that the next block is the one we seek to start the syncing from.


bokobza:
@dangershony

PR #472, opened 2017-09-20
Better Logging for ProcessPendingStorage and clear PendingBlockPairsToStore after push to repo

fassadlr:
Added better/extra logging to the ProcessPendingStorageStep
Clear context.PendingBlockPairsToStore after we have pushed to repository.

Relates to:
https://trello.com/c/8Y2OuIfh/221-blockstore-2-process-pending-storage-performance


Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -161,24 +156,25 @@ private StepResult PrepareNextBlockFromPendingStorage(ProcessPendingStorageConte
         /// <param name="context"><see cref="ProcessPendingStorageContext"/></param>
         private async Task PushBlocksToRepository(ProcessPendingStorageContext context)
         {
-            await this.BlockStoreLoop.BlockRepository.PutAsync(context.PendingBlockPairsToStore.First().ChainedBlock.HashBlock, context.PendingBlockPairsToStore.Select(b => b.Block).ToList());
+            this.logger.LogDebug(context.ToString());
 
+            await this.BlockStoreLoop.BlockRepository.PutAsync(context.PendingBlockPairsToStore.First().ChainedBlock.HashBlock, context.PendingBlockPairsToStore.Select(b => b.Block).ToList());
```
dangershony: Using `context.PendingBlockPairsToStore.First()` is not so intuitive right?
we could use  `context.PendingBlockPairToStore` instead.
But maybe a good idea is to assert that?

fassadlr: Yes! I like it, will amend 👍

PR #478, opened 2017-09-21
Add stratis fee policy to light wallet

mikedennis:
Stratis fee is based on a fixed fee of 10,000/KB as per StratisX.
Unit tests to come in future commit.


Stratis.Bitcoin.Features.LightWallet/LightWalletFixedFeePolicy.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Constructor for the light wallet fixed fee policy.
+        /// </summary>
+        /// <param name="loggerFactory">The factory for building logger instances.</param>
+        public LightWalletFixedFeePolicy(ILoggerFactory loggerFactory)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            // TODO: For now default is hardcoded, should probably be configurable 
+            this.TxFeeRate = new FeeRate(10000);
```
dangershony: FYI the TODO good point, maybe this should go in to the Network class and be set when building the network as it will apply to every network a min fee policy and always be mandatory.

Also one more idea, what do you think when we take code from X we can post the link to the place in X that we copied from?

Stratis.Bitcoin.Features.LightWallet/LightWalletFeature.cs
```diff
@@ -134,7 +137,10 @@ public static IFullNodeBuilder UseLightWallet(this IFullNodeBuilder fullNodeBuil
                         services.AddSingleton<IWalletSyncManager, LightWalletSyncManager>();
                         services.AddSingleton<IWalletTransactionHandler, WalletTransactionHandler>();
                         services.AddSingleton<IWalletManager, WalletManager>();
-                        services.AddSingleton<IWalletFeePolicy, LightWalletFeePolicy>();
+                        if (isStratis)
+                            services.AddSingleton<IWalletFeePolicy, LightWalletFixedFeePolicy>();
+                        else
+                            services.AddSingleton<IWalletFeePolicy, LightWalletFeePolicy>();
```
dangershony: Currently LightWalletFeePolicy is actually a bitcoin only fee policy and is collecting the fee from an external source.
I propose we change the name to LightWalletBitcoinFeePolicy or LightWalletBitcoinExternalFeePolicy

Stratis.Bitcoin.Features.LightWallet/LightWalletFixedFeePolicy.cs
mikedennis: RE: minfee policy For fullnode wallet looks like min tx fee is 1000 for bitcoin and 10000 for strat.  What do you think about putting fallbackfee into network the same way?

Stratis.Bitcoin.Features.LightWallet/LightWalletFeature.cs
mikedennis: I changed it to LightWalletBitcoinExternalFeePolicy

bokobza:
Having a look as well.

Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
```diff
@@ -46,42 +54,42 @@ public class WalletFeePolicy : IWalletFeePolicy
...
-            // network level or the consensus options
-            if (network.Name.ToLower().Contains("stratis"))
-            {
-                this.minTxFee = new FeeRate(10000);
-                this.fallbackFee = new FeeRate(40000);
-                this.payTxFee = new FeeRate(0);
-                this.maxTxFee = new Money(0.1M, MoneyUnit.BTC);
-                this.minRelayTxFee = new FeeRate(10000);
-            }
+            this.minRelayTxFee = network.IsBitcoin() ? MempoolValidator.MinRelayTxFee : this.minTxFee;
```
bokobza: is the minRelayTxFee supposed to be equal to the minTxFee or do they just happen to be the same value?

mikedennis: This was the existing logic. I'm not sure why it isn't using the mempool validators value the same as bitcoin. I will look into mempool validator and see if I can figure out why.

mikedennis: Looks like Mempool Validator is definitely hardcoded to bitcoin a value, which explains the reason for the conditional logic in the walletfeepolicy

https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs#L144

I'm not sure if this is causing other problems in mempool validator on a stratis network? Stratis mintxfee is 10000 not 1000. 

For example this check: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs#L692

Stratis.Bitcoin.Features.LightWallet/LightWalletFixedFeePolicy.cs
dangershony: No NodeSettings is not distinguishable by network we would have to put an if condition (I see you already added the if condition)
I think its not a big deal to add it in NBitcoin.
So you propose to add MinFee and FallBackFee in to network in NBitcoin? sounds goo to me.
I would suggest we do that and then your PR will change a bit.
I am happy to commit this as it is now and then make the other changes:
- Commit as is
- Add MinFee  and FallBackFee  to Network in NBitcoin
- Bring changes to the fullnode and change in the correct places

Sounds good?

Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
dangershony: Hmm should we add MinRelayTxFee to network as well?

Stratis.Bitcoin.Features.LightWallet/LightWalletFixedFeePolicy.cs
mikedennis: Sounds good. Right now I created a network extension method for the default fees which has the nasty "if" in it.  I agree it would be cleaner if the defaults were stored directly in the NBitcoin networks.

Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
mikedennis: Could do and then have Mempool validator use that instead of hardcoded value?

dangershony:
So we agree this can go in?
Then a follow up change will come with the NBitocin changes.

bokobza:
Yeah that's fine for me :-) Thanks @mikedennis

PR #479, opened 2017-09-21
Fix logging in with wrong password bug

nopara73:
Resolves: https://github.com/stratisproject/Breeze/issues/226


dangershony:
Looks good to me I will push @bokobza can you retro check?

PR #484, opened 2017-09-22
Organize Extensions and Interfaces

nopara73:


bokobza:
> Separate IWalletFeePolicy into its own file

 - yes I'm up for doing that.

> Separate extensions into their own folder

 - up for splitting them up in their own file, but the namespaces should be a stratis namespace, not System or System.Linq. Also, the Extensions folder could be under the Utilities folder.

> Add ThreadingExtensions.cs

 - not really sure of that.

> Separate interfaces of Features.Wallet project into their own folder

 - yes.

> Add extension methods to GetAwaiter().GetResult()

 - don't see the utility of that. It doesn't really extend the functionality but provide a syntactic shortcut. Or not: AwaiterWait() is longer than Wait(). Also, both extensions methods do the same thing, and the replacement in the classes is inconsistent.

nopara73:
Replaced by https://github.com/stratisproject/StratisBitcoinFullNode/pull/539

PR #470, opened 2017-09-19
[WIP] Implement proper transaction propagation

nopara73:

PR #486, opened 2017-09-22
[WIP] Implement transaction broadcasting and propagation 3 ways

nopara73:
Superseeds: https://github.com/stratisproject/StratisBitcoinFullNode/pull/470

ToDo:  
@dev0tion We need to resolve this issue before it gets merged: https://github.com/stratisproject/Breeze/issues/207  
@bokobza Please sort out the tests. Because from now on we don't get false positive results testing without real network is impossible. In `WalletControllerTests` `SendTransactionFailedReturnsBadRequest` and `SendTransactionWithExceptionReturnsBadRequest` are failing.  
Furthermore please remove the WalletManager's sendtransaction, because the wallet is not the one who broadcasts and propagates the transaction. At least not anymore. I don't remove it because, many more tests would fail.


dangershony:
Ah jesus this has a dependency on the awaiter extension, it will be very hard to review.

nopara73:
@dangershony I reviewed it, no worry:)  
Jokes aside, we talked about this before and you said it'll be fine like this. Not sure why you're surprised.

dangershony:
Generally we need to separate such PR so one does not depend on the other.

dangershony:
Is there anyway this can not be a dependency of #484

PR #487, opened 2017-09-23
Feature/pos calculate stake information

### LINKED ISSUE #387: DBreezeCoinView.Rewind() - reuse this.GetRewindIndex()

Aprogiena:
Calculation of network weight and time estimation for a new block with current node stake.

Also fixes a couple of newly revealed bugs - note that there is a difference between `consensusLoop.Tip` and `chain.Tip` (where chain is `ConcurrentChain`). We can not use `chain.Tip` if we then ask `StakeChainStore` to obtain `BlockStake` information. This is because it might not be available and there is assert in `GetAsync` method in `StakeChainStore`. So couple of methods were fixed to rather use `consensusLoop.Tip` instead of `chain.Tip`.

We also 
Fix #387


Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
```diff
@@ -53,7 +53,7 @@ protected override void UpdateHeaders()
             base.UpdateHeaders();
 
             var stake = new BlockStake(this.pblock);
-            this.pblock.Header.Bits = StakeValidator.GetNextTargetRequired(this.stakeChain, this.chain.Tip, this.network.Consensus, this.options.IsProofOfStake);
+            this.pblock.Header.Bits = StakeValidator.GetNextTargetRequired(this.stakeChain, this.consensusLoop.Tip, this.network.Consensus, this.options.IsProofOfStake);
```
dangershony: I see why you did this be in reality we should never be staking if TIP are not the same.
In fact this might be buggy as if tips are not the same we are staking on the wrong height and will create a split.
The solution is to throw and a miner exception and let the processes start again.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -323,22 +324,25 @@ private void CheckStake(ContextInformation context, ChainedBlock pindexPrev)
...
+            // The following is wrong, this should be REORG not SOLUTION FOUND
+            // but this would just narrow the race condition in case of reorg, not mitigate it.
+            // --------
             // Found a solution.
-            if (block.Header.HashPrevBlock != this.chain.Tip.HashBlock)
-            {
-                this.logger.LogTrace("(-)[SOLUTION_FOUND]");
-                return;
-            }
+            // if (block.Header.HashPrevBlock != chainTip.HashBlock)
```
dangershony: Its not reorg but rather the chain has advanced, I think it might still be needed.

Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
Aprogiena: just run into a bug which crashed my miner - see #500 
using of chain.Tip is to be very careful, consensus is better, but still not good really enough
all these tips should come from above (parameter)

chain.Tip can change at any time, if we at any point rely on it, it is an open vector for a bug, but this should be solved/discussed in #500

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: yes, the problem here is described in #488 
the check itself is useless - it only narrows the race condition window, which is dangerous thing, it is better to have wide race condition windows until the race is fixed

if the chain advanced after this IF was executed, then #488 crash happens, now it happens if the chain advances sooner, but if it advances at another point, we've got #500 crash

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -889,5 +898,109 @@ private int GetDepthInMainChain(StakeTx stakeTx)
...
+        /// <returns>Estimated number of coins that are used by all stakers on the network.</returns>
+        /// <remarks>
+        /// The idea behind estimating the network staking weight is very similar to estimating 
+        /// the total hash power of PoW network. The difficulty retarget algorithm tries to make 
+        /// sure of certain distribution of the blocks over a period of time. Base on real distribution
+        /// and using the actual difficulty targets, one is able to compute how much stake was 
+        /// presented on the network to generate each block.
+        /// <para>
+        /// The method was ported from 
+        /// https://github.com/stratisproject/stratisX/blob/47851b7337f528f52ec20e86dca7dcead8191cf5/src/rpcblockchain.cpp#L74 .
```
dangershony: Excellent thanks for including the link to the port

Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
dangershony: Also consensus tip can change in flight I agree we need to create a param.
Maybe better take a param and compare tips else throw?

Aprogiena: I think argument suggestion as in #500 will be just good enough. You don't need to check it really. You check once when you start the round, then you use the value for that round, that should be enough. So either we don't mine it, then next round we check again, or we mine it, in which case, if we fix #488, it will at that point recognize that we were mining something else (and there we have to decide whether we want to continue with our own or jump on remote train, better strategy is probably to continue with our own).

Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: The #500 is probably wrong to be a crash it should just be ignored especially when staking.
But yeah we should probably not use Tip any where but rather pass a ChainedBlock parameter instead (refactor).

PR #492, opened 2017-09-24
Feature/checksync issue 413

### LINKED ISSUE #413: [POS] CheckSync is not implemented in C#

Aprogiena:
This implements similar approach as StratisX in their `Checkpoints::CheckSync`. The logic in our implementation is in `ChainHeadersBehavior`.

When a peer's new headers are received, we find a fork (last common chained header) in its header chain and in our consensus chain. If the difference in heights of our consensus tip and the fork is greater than allowed maximal reorg length, we mark the headers as invalid. This prevents further syncing from the node.

This PR also improves code style in `ChainHeadersBehavior` and introduces logs to it.

**BELOW IS OLD DESCRIPTION, NO LONGER VALID:**
This is an **ATTEMPT** to implement what StratisX knows as "synchronized checkpoints". It is a mechanism that should prevent reorgs longer than 500 blocks. 

However, I'm completely unsure that I've got it right - esp. the check is in the right place and checking the right thing. I believe we construct our chain differently than StratisX, so that is why I'm unsure.

I assume the ConcurrentChain's Tip is just the headers we've seen on the network, but not necessarily a valid chain. And I consider ConsensusLoop's Tip as the block that is at the end of our fully validated longest chain. 

So, in `ContextualCheckBlockHeader` of PosConsensusValidator, I implement the mechanism ported from StratisX in their `Checkpoints::CheckSync`. I've used different naming as I don't really consider this as a checkpoint thing. 

Anyway, what I'm unsure if such a block that would not pass the check would enter our `AcceptBlock` method in ConsensusLoop. If so, then it should be alright.

I'm bit afraid of the rewinding inside ConsensusLoop's `ExecuteNextBlock`. I don't really understands when it does take place.

**So I would like the reviewer to check whether the placement of this code is actually correct and does what is intended - i.e. that if we are fully synced and then suddenly we connect to a new peer who claims a longer chain, for which we would need to reorganize more than 500 blocks then it would be rejected by this.** 

Fix #413


Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -312,6 +312,17 @@ public override void ContextualCheckBlockHeader(ContextInformation context)
                 ConsensusErrors.BlockTimestampTooEarly.Throw();
             }
 
+            int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
```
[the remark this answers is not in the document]
dangershony: Ah and you can fairly easily write a test to check that.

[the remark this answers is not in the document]
Aprogiena: I've already created an issue for such a test - https://github.com/stratisproject/StratisBitcoinFullNode/issues/493 - and frankly, once this test exists, I should be able to find a place where to but the check. I was expecting what you just confirmed, so I need to rewrite this. If you could help with the test, that would help a lot.

I'm not sure that this is attacker only scenario though. A poorly connected big staker (say 40 % of the network) will have some chance to achieve this by mistake. Well it is a pretty big mistake, but I would argue that it is possible without any malicious code being present on that particular node. I would the normal consensus temporary ban as it is usual is the right reaction. But since we don't implement these bans yet, this is to be decided later anyway.

OK, so to progress here - if you could help with the test, this will really help me in getting that extra bit of knowledge that is needed to find the spot. And certainly I would then start with those places you've mentioned.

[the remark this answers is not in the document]
dangershony: Unfortunately we don't have many POS tests yet until I remove the static flags from NBitcoin, but such a test does not depend on POS I can simulate this using POW.

Yes sure I will create a test with 510 size reorg and submit a PR.

Aprogiena:
I think we are done here, we just need to finish the test to reflect that this works well. More about the test in #413

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -850,7 +848,6 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
                             context.Result.KernelCoin = coin;
 
                             context.Logger.LogTrace("Kernel accepted, coinstake input is '{0}/{1}', stopping work.", prevoutStake.Hash, prevoutStake.N);
-                            kernelFound = true;
```
dangershony: You rely on this to break out of the loop?
` if (context.Result.KernelFoundIndex != CoinstakeWorkerResult.KernelNotFound)`

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,326 +7,390 @@
...
-			if (this.AttachedNode.State == NodeState.Connected)
-			{
-				var highPoW = this.SharedState.HighestValidatedPoW;
+        }
+
+        protected override void AttachCore()
+        {
+            this.logger.LogTrace("()");
+
+            this.refreshTimer = new Timer(o =>
```
dangershony: I would be happy t remove this timer all together (not not this PR)

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: ` if (stopWork) break;` this one

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,326 +7,390 @@
...
+                if ((maxReorgLength != 0) && (network != null) && (consensusTip != null))
+                {
+                    ChainedBlock fork = this.pendingTip.FindFork(consensusTip);
+                    if ((fork != null) && (fork != consensusTip))
+                    {
+                        int reorgLength = consensusTip.Height - fork.Height;
+                        if (reorgLength > maxReorgLength)
+                        {
+                            this.logger.LogTrace("Reorganization of length {0} prevented, maximal reorganization length is {1}, consensus tip is '{2}'.", reorgLength, maxReorgLength, consensusTip);
+                            this.invalidHeaderReceived = true;
```
dangershony: I think we should trigger a disconnect from this node.
But perhaps we will do that when working on node banning (as we will also want to remember a banned node).

Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "commandLineArgs": "-testnet"
```
dangershony: This should go in the next line `"commandLineArgs": "-testnet"`

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: yes this is for node banning

Stratis.StratisD/Properties/launchSettings.json
Aprogiena: this file should not be changed, but the PR, probably committed by mistake, will remove it

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,326 +7,390 @@
...
+            // This payload notifies peers of our current best validated height. 
+            // Use the ChainState.ConsensusTip property (not Chain.Tip)
+            // if the peer is behind/equal to our best height an empty array is sent back.
+
+            // Ignoring "getheaders" from peers because node is in initial block download.
+            var getheaders = message.Message.Payload as GetHeadersPayload;
+            if ((getheaders != null) 
+                && this.CanRespondToGetHeaders
+                // If not in IBD whitelisted won't be checked.
+                && (!this.chainState.IsInitialBlockDownload || this.AttachedNode.Behavior<ConnectionManagerBehavior>().Whitelisted))
```
dangershony: A suggested improvement, I think this will never change so we could set a property with this behaviour 
`this.AttachedNode.Behavior<ConnectionManagerBehavior>().Whitelisted))`

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: i think you've mentioned that already somewhere (maybe in 413 discussion?) but i was not sure if this can be removed safely; wouldn't removing this stop us from doing some kind of recovery if something somewhere goes wrong? i don't know, you probably understands this code better, so i would leave that for you, if you want to have this removed, just create an issue for it ...

dangershony:
@Aprogiena @bokobza we want to take another version of Breeze release so perhaps better we wait with this PR til we take that release.
As this introduce new functionality that I would like to have tested more.

Aprogiena:
We mostly need that test to be finished, it then tests the functionality.

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: possibly true, but offtopic ;)

Aprogiena:
So I've fixed the conflicts, ...

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,319 +7,354 @@
...
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Utilities;
+using Microsoft.Extensions.Logging;
 
 namespace Stratis.Bitcoin.Base
 {
+    /// <summary>
+    /// The Chain Behavior is responsible for keeping a ConcurrentChain up to date with the peer, it also responds to getheaders messages.
+    /// </summary>
+    public partial class ChainHeadersBehavior : NodeBehavior
```
dangershony: Partial can go away

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,319 +7,354 @@
...
+        public bool CanSync { get; set; }
+
+        /// <summary><c>true</c> to sync the chain as headers come from the network, <c>false</c> not to sync automatically.</summary>
+        public bool AutoSync { get; set; }
+
+        /// <para>
+        /// The announced tip is accepted if it seems to be valid. Validation is only done on headers 
+        /// and so the announced tip may refer to invalid block.
+        /// </para>
+        /// </summary>
```
dangershony: Invalid ` </summary>`

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -7,319 +7,354 @@
...
+
+                if (pendingTipBefore != this.pendingTip)
+                    this.logger.LogTrace("Pending tip changed to '{0}'.", this.pendingTip);
+
+                // Long reorganization protection on POS networks.
+                bool reorgPrevented = false;
+                uint maxReorgLength = this.chainState.MaxReorgLength;
+                Network network = this.AttachedNode?.Network;
+                ChainedBlock consensusTip = this.chainState.ConsensusTip;
+                if ((maxReorgLength != 0) && (network != null) && (consensusTip != null))
```
dangershony: Logically this looks fine, semantically I would suggest to put the whole reorg bloc inside its own `if` statement and avoid the allocations of network and consensusTip
It make the code more readable. 

```
 bool reorgPrevented = false;
  if(this.preventReorgs)
  {
                uint maxReorgLength = this.chainState.MaxReorgLength;
                Network network = this.AttachedNode?.Network;
                ChainedBlock consensusTip = this.chainState.ConsensusTip;
                if ((network != null) && (consensusTip != null))
                {
                         bla..
                }
  }
``` 
You could even have a flag at the constructor that specifies if prevention is enabled
```
this.preventReorgs = this.chainState.MaxReorgLength != 0
```

dangershony:
If you can merge and I will check the test is fine.

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: well, ok, little off topic though :)

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: damn it seems i've deleted first part of the comment, will try to find it, but will definitely fix it somehow

Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: I see, ok, will fix

Aprogiena: I can't do the constructor thing, MaxReorgLength is not initialized at that point, but except for that, i'll do as you suggest

Aprogiena:
All done, we now need the test.

Aprogiena:
Also can't merge until approved, but once approved, feel free to merge yourself.

bokobza:
@Aprogiena @dangershony is this ok to go in?

Aprogiena:
I would be happier if the flow was like this:

1. the test from 413 was finished
2. it was checked that the test FAILS against current master
3. then it was checked that the test PASSES against this PR
4. only then it was approved

PR #496, opened 2017-09-25
Fix build

bokobza:


bokobza:
ok the build is ok but tests are failing, which is somehow expected but needs looking into.

PR #502, opened 2017-09-27
Feature/pos rpc getstakinginfo tests

Aprogiena:
Tests realted to `getstakinginfo` RPC:

- RPC model tests - JSON serialization and deseralization

- Controller test of `getstakinginfo` RPC command. Note that this is very trivial test that only checks that the staking node run with correct set of arguments enabled staking and the check is done via RPC command. A much better test would be to have the node actually start staking, but for that we would need a wallet with some coins and also we would need the node to think it is not in IBD.

**Note: This PR requires https://github.com/stratisproject/StratisBitcoinFullNode/pull/499 to be merged.**


dangershony:
Conflicts,
Regarding testing we could actually do such a thing, mine 100 blocks using POW and then try to stake using the premine coins.

Aprogiena:
Conflict solved. If you think the test like that is possible, we should create issue for it. I don't know how to create it and even less how to create it right.

dangershony:
I will eventually but it is doable, in the integration tests I mean.

Issue #426, opened 2017-09-06
[StakeChainStore] Crash


Aprogiena:
```
      WalletSyncManager initialized. wallet at block 100373.
info: Stratis.Bitcoin.Api.ApiFeature[0]
      Api starting on url http://localhost:5000/
fail: Stratis.Bitcoin.Builder.FullNodeFeatureExecutor[0]
      An error occurred starting the application

Unhandled Exception: System.AggregateException: One or more errors occurred. (Missing stake information, delete the data folder and re-download the chain) ---> Stratis.Bitcoin.Configuration.ConfigurationException: Missing stake information, delete the data folder and re-download the chain
   at Stratis.Bitcoin.Features.Consensus.StakeChainStore.<Load>d__9.MoveNext() in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Features\Consensus\StakeChainStore.cs:line 81
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Stratis.Bitcoin.Features.Consensus.ConsensusFeature.Start() in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Features\Consensus\ConsensusFeature.cs:line 97
   at Stratis.Bitcoin.Builder.FullNodeFeatureExecutor.<>c.<Start>b__3_1(IFullNodeFeature service) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Builder\FullNodeFeatureExecutor.cs:line 57
   at Stratis.Bitcoin.Builder.FullNodeFeatureExecutor.Execute(Action`1 callback, Boolean reverseOrder) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Builder\FullNodeFeatureExecutor.cs:line 102
   --- End of inner exception stack trace ---
   at Stratis.Bitcoin.Builder.FullNodeFeatureExecutor.Execute(Action`1 callback, Boolean reverseOrder) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Builder\FullNodeFeatureExecutor.cs:line 120
[...]

PR #503, opened 2017-09-27
[Stake Chain Store] Refactor StakeItem to ensure validity when querying and persisting BlockStake items in the StakeChainStore

fassadlr:
This is further refactoring which relates to #426.

This PR is to better how we work with StakeItem in the StakeChainStore when querying and persisting BlockStake objects.

We will either create a StakeItem for querying via the static Query() method or for persisting via the Create() method. This way none of the internal properties inside Stakeitem can ever be changed for any other reason than the above.

I've also added better logging.


Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -300,38 +301,69 @@ public Task PutStake(IEnumerable<StakeItem> stakeEntries)
...
+        /// <param name="stakeItem">Partially initialized <see cref="StakeItem"/> that is populated by <see cref="BlockStake"/>.</param>
+        public Task GetStakeItem(StakeItem stakeItem)
+        {
+            return this.session.Execute(() =>
+            {
+                this.logger.LogTrace("({0}:{1})", nameof(stakeItem.BlockHash), stakeItem.BlockHash);
+
+                var blockStake = this.session.Transaction.Select<byte[], BlockStake>("Stake", stakeItem.BlockHash.ToBytes(false));
+                if (!blockStake.Exists)
+                    this.logger.LogTrace("{0}:{1} [NOT IN DATABASE]", nameof(stakeItem.BlockHash), stakeItem.BlockHash);
```
Aprogiena: I know I've still did not create the how to write logging document, so it is not really your fault, but this is not what I'm looking for in logging. So I'll try to explain.

There are two special kinds of logs - entry and exit logs in methods. Almost every method should have them except for methods that are called too often. And by too often I rather mean relative count. I.e. we can have a block with 1000 transactions and it is perfectly OK if a complex method that works with transaction logs on entry and exit for each transaction. Too often is when for a single object we have a for loop which executes trivial methods.

Entry logs come in pattern `"(arg1:value1,arg2:value2,...)"`, exit logs come in pattern `"(-):resultValue,outarg1=outval1,outarg2=outval2,..."`. And we are looking for every entry log to have its matching exit log. Recently, I've had a crash and there was not full stack info, so I did not know where exactly the crash happened. But I did have exit logs for every entry logs except for the final path to the crashing line -> easy find.

Now if a method has a multiple ways of exit (multiple returns, exceptions) then we want to distinguish in log, which return it was, in that case the exit log pattern extends to: `"(-)[UNIQUE_CONTEXT]:resultValue,outarg1=outval1,outarg2=outval2,..."`

Now with both entry and exit logs we use nameof() to put names of arguments in the log.
And finally, entry and exit logs are made on trace level.

OK, so those are entry and exit logs, but we also log interesting information inside the methods. We want those to look differently (visually) compared to entry/exit. 
Their pattern is `"Not very complex, but unique, English sentence."`. We do not use '[CONTEXT]' as the context is given by the sentence itself. We also do not use nameof() because we rather describe 
the event in English. We also don't use `"name:value"` notation or `"name=value"`.

These logs can be on trace level, debug level or info level, or even warning or error/fatal level.

So, instead of 

```
this.logger.LogTrace("{0}:{1} [NOT IN DATABASE]", nameof(stakeItem.BlockHash), stakeItem.BlockHash); 
```

We are looking for something like this:
``` 
this.logger.LogTrace("Item hash '{0}' not found in the database.", stakeItem.BlockHash); 
```

Whether or not to enclose a value in `'{0}'` depends on its type. All small integer types, booleans etc. go without apostrophes, strings, block hashes or hash values, or more complex values do, except for arrays that are `'[val1,val2,...]'`.

Similarly
```
this.logger.LogTrace("{0} [NULL]", nameof(stakeItem.BlockStake));
```

->

```
this.logger.LogTrace("Stake information is mossing for item hash '{0}'.", stakeItem.BlockHash);
```

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+
+        /// <summary>The block stake.</summary>
+        public BlockStake BlockStake { get; private set; }
+
+        /// <summary>The height of the block.</summary>
+        public long? Height { get; private set; }
+
+        /// <summary>
+        /// Returns <c>true</c> if the item exists in the store. It is used to determine
+        /// if the <see cref="StakeChainStore"/> should persist the item or not. />
```
Aprogiena: excessive `/>`

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        public uint256 BlockHash { get; private set; }
+
+        /// <summary>The block stake.</summary>
+        public BlockStake BlockStake { get; private set; }
+
+        /// <summary>The height of the block.</summary>
+        public long? Height { get; private set; }
+
+        /// <summary>
+        /// Returns <c>true</c> if the item exists in the store. It is used to determine
```
Aprogiena: I wouldn't use `Returns` here as this is rather a property, yes its getter does return, but I'd speak about it as a value, so just `<c>true</c> if the item exists in the store. ...`

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        private StakeItem() { }
+
+        /// <summary>The hash of the block.</summary>
+        public uint256 BlockHash { get; private set; }
+
+        /// <summary>The block stake.</summary>
+        public BlockStake BlockStake { get; private set; }
+
+        /// <summary>The height of the block.</summary>
+        public long? Height { get; private set; }
```
Aprogiena: When you use such nullable thingy, it is always beneficial to comment the null case. But is it really needed here? Wouldn't a simple long be enough?

Also, afaik everywhere Height is just integer, so why long?

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
-        public bool InStore;
-        public long Height;
-    }
-}
+        private StakeItem() { }
+
+        /// <summary>The hash of the block.</summary>
+        public uint256 BlockHash { get; private set; }
+
+        /// <summary>The block stake.</summary>
```
Aprogiena: I think this is rather `staking information`, "block stake" is a name of the class or variable.

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        /// <summary>
+        /// Returns <c>true</c> if the item exists in the store. It is used to determine
+        /// if the <see cref="StakeChainStore"/> should persist the item or not. />
+        /// </summary>
+        public bool ExistsInStore { get; private set; }
+
+        /// <summary>
+        /// Sets the item's <see cref="ExistsInStore"/> value to <c>true</c>.
+        /// </summary>
+        internal void InStore()
```
Aprogiena: Wouldn't `MakeInStore` be better name? Or any other action verb with it

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        {
+            Guard.NotNull(blockStake, nameof(blockStake));
 
+            this.BlockStake = blockStake;
+
+            InStore();
+        }
+
+        /// <summary>
+        /// Static method to create a valid <see cref="StakeItem"/> object.
```
Aprogiena: I don't think it is beneficial to repeat here that it is static method. I'd just use `Creates a valid <see cref="StakeItem"/> object.`

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        /// </summary>
+        /// <param name="blockHash">The hash of the stake block.</param>
+        /// <param name="blockStake">The block stake.</param>
```
Aprogiena: in case you consider change above, don't forget here.

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        /// <param name="height">The height of the stake block.</param>
+        /// <returns>Fully initialized <see cref="StakeItem"/></returns>
```
Aprogiena: missing `.`

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+            return stakeItem;
+        }
+
+        /// <summary>
+        /// Static method to partially initialize <see cref="StakeItem"/> to query the stake chain store
+        /// database with.
+        /// </summary>
+        /// <param name="blockHash">The hash of the stake block to query.</param>
+        /// <param name="height">The height of the stake block to query.</param>
+        /// <returns>Partially initialized <see cref="StakeItem"/>StakeItem</returns>
```
Aprogiena: wrong comment, delete second StakeItem, add dot.

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+
+            var stakeItem = new StakeItem();
+            stakeItem.BlockHash = blockHash;
+            stakeItem.BlockStake = blockStake;
+            stakeItem.Height = height;
+            return stakeItem;
+        }
+
+        /// <summary>
+        /// Static method to partially initialize <see cref="StakeItem"/> to query the stake chain store
```
Aprogiena: static method--

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        /// database with.
+        /// </summary>
+        /// <param name="blockHash">The hash of the stake block to query.</param>
+        /// <param name="height">The height of the stake block to query.</param>
+        /// <returns>Partially initialized <see cref="StakeItem"/>StakeItem</returns>
+        internal static StakeItem Query(uint256 blockHash, int? height = null)
```
Aprogiena: you are using int here, long above, int is correct, but again, if nullable is justified, it needs to be commented as a special value. Not sure if nullable is needed though here.

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
```diff
@@ -1,13 +1,84 @@
...
+        }
+
+        /// <summary>
+        /// Static method to create a valid <see cref="StakeItem"/> object.
+        /// </summary>
+        /// <param name="blockHash">The hash of the stake block.</param>
+        /// <param name="blockStake">The block stake.</param>
+        /// <param name="height">The height of the stake block.</param>
+        /// <returns>Fully initialized <see cref="StakeItem"/></returns>
+        internal static StakeItem Create(uint256 blockHash, BlockStake blockStake, long height)
```
Aprogiena: long or int again

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
dangershony: Height will never be null

Stratis.Bitcoin.Features.Consensus/StakeItem.cs
dangershony: I dont like so much the method Query if anything it should be Create when using a static method to build an instance.

Stratis.Bitcoin.Features.Consensus/StakeChainStore.cs
```diff
@@ -51,66 +49,58 @@ public StakeChainStore(Network network, ConcurrentChain chain, DBreezeCoinView d
...
-            if (load.Any(l => l.BlockStake == null))
-            {
-                this.logger.LogTrace("(-)[STAKE_INFO_MISSING]");
-                throw new ConfigurationException("Missing stake information, delete the data folder and re-download the chain");
-            }
-
-            foreach (StakeItem stakeItem in load)
-                this.items.TryAdd(stakeItem.BlockId, stakeItem);
+            foreach (StakeItem stakeItem in stakeItems)
+                this.items.TryAdd(stakeItem.BlockHash, stakeItem);
```
dangershony: You removed the check that all items are not null why?
```
  if (load.Any(l => l.BlockStake == null))
```
If we did not have this check til now finding the bug in the first place would be much harder.

Stratis.Bitcoin.Features.Consensus/StakeChainStore.cs
```diff
@@ -51,66 +49,58 @@ public StakeChainStore(Network network, ConcurrentChain chain, DBreezeCoinView d
...
+            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);
 
-            var stakeItem = new StakeItem { BlockId = blockid };
-            await this.dBreezeCoinView.GetStake(new[] { stakeItem }).ConfigureAwait(false);
+            var stakeItem = StakeItem.Query(blockHash);
+            await this.dBreezeCoinView.GetStakeItem(stakeItem);
 
-            if (stakeItem.BlockStake != null) this.logger.LogTrace("(-):*.{0}='{1}'", nameof(stakeItem.BlockStake.HashProof), stakeItem.BlockStake.HashProof);
-            else this.logger.LogTrace("(-):null");
+            this.logger.LogTrace("(-)");
```
dangershony: You removed the guard check is that wise?

dangershony:
Generally I think we found the null bug so this refactor is very low priority.

dangershony:
I am closing this for now, keep the branch we might add it later.

PR #509, opened 2017-09-28
Bugfix/issue 495

### LINKED ISSUE #495: [Consensus] Couple of omissions
1) in PowMining when block is found, there is a sequence AcceptBlock, Puller.SetLocation, consensusLoop.FlushAsync, blockRepository.PutAsync and signals.SignalBlock
while in PosMining, i can only see AcceptBlock, Puller.SetLocation, blockRepository.PutAsync and signals.SignalBlock, i am missing consensusLoop.FlushAsync there, i guess these two sequences should be similar

2) `this.UTXOSet.SaveChangesAsync(context.Set.GetCoins(this.UTXOSet), null, this.Tip.HashBlock, context.BlockResult.ChainedBlock.HashBlock);` in AcceptBlock in ConsensusLoop should use await or GetResult to wait for it, because otherwise we rely on the fact that the cached coin view that is used here is synchronous

Aprogiena:
Trivial changes

Fix #495


Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -111,35 +112,35 @@ public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
...
         }
 
         public void AcceptBlock(ContextInformation context)
         {
             using (this.watch.Start(o => this.Validator.PerformanceCounter.AddBlockProcessingTime(o)))
             {
-                // check that the current block has not been reorged
-                // catching a reorg at this point will not require a rewind
+                // Check that the current block has not been reorged.
+                // Catching a reorg at this point will not require a rewind.
```
dangershony: This is a condition I have added, I don't think core or X do such a thing. but also we use consensus a bit different.

PR #510, opened 2017-09-28
Block store to push batches when not in IBD

dangershony:


Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -93,18 +93,16 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
-                }
 
                 if (context.PendingStorageBatchSize > BlockStoreLoop.MaxPendingInsertBlockSize)
-                    await PushBlocksToRepository(context);
-            };
+                    await this.PushBlocksToRepository(context);
+            }
+
+            if (context.PendingBlockPairsToStore.Any())
+                await this.PushBlocksToRepository(context);
```
fassadlr: I'm not sure moving the push out of the while makes any difference here Dan? The issue with ProcessWhenInIBD was that it started pushing was pushing single blocks when I got out of IBD. We said that we wanted to batch the blocks up until about 5 blocks from Consensus Tip?

Above you'll see that the Stop condition is reached and then with your change, it'll break out of the loop and then immediately push the blocks in pending to the repo (even if there's one block).

fassadlr: Also, can you maybe a unit test for this scenario so that we can that it behaves as expected?

dangershony: No I moved it out of the loop for the dispose issue.
If cancellation is triggered and we dont push whats in PendingBlockPairsToStore then when we call the code with dispose the blocks will not be in the pending storage list anymore, that will not have the effect we wanted.

Issue #493, opened 2017-09-24
[POS] Create integration test for long reorg protection


Aprogiena:
We need a test that check the following scenario (after https://github.com/stratisproject/StratisBitcoinFullNode/pull/492 is merged):

- the node is fully synced 
- suddenly it connects to a new peer who claims a longer chain, for which we would need to reorganize more than 500 blocks 
- this chain is rejected because the reorg is too long (the chain is otherwise valid and is longer than our chain)

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[POS] Create integration test for long reorg protection](https://trello.com/c/bsgxLyfq/246-pos-create-integration-test-for-long-reorg-protection)

Aprogiena:
This is almost finished, just couple of things needs to be fixed in the test. It is discussed in #413.

PR #511, opened 2017-09-28
Unit test that simulate a 500 block reorg

dangershony:
This will also fix the failing tests we had til now

### DIFF
--- a/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
+++ b/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
@@ -138,6 +138,7 @@ namespace Stratis.Bitcoin.Features.Consensus
         {
             if (network.Equals(Network.Main)) this.checkpoints = bitcoinMainnetCheckpoints;
             else if (network.Equals(Network.TestNet)) this.checkpoints = bitcoinTestnetCheckpoints;
+            else if (network.Equals(Network.RegTest)) this.checkpoints = new Dictionary<int, CheckpointInfo>();
             else if (network.Equals(Network.StratisMain)) this.checkpoints = stratisMainnetCheckpoints;
             else if (network.Equals(Network.StratisTest)) this.checkpoints = stratisTestnetCheckpoints;
             else this.checkpoints = new Dictionary<int, CheckpointInfo>();
--- a/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -208,7 +208,7 @@ namespace Stratis.Bitcoin.Features.Consensus
                 // TODO Need to revisit unhandled exceptions in a way that any process can signal an exception has been
                 // thrown so that the node and all the disposables can stop gracefully.
                 this.logger.LogDebug("Exception occurred in consensus loop: {0}", ex.ToString());
-                this.logger.LogCritical(new EventId(0), ex, "Consensus loop unhandled exception (Tip:" + this.consensusLoop.Tip?.Height + ")");
+                this.logger.LogCritical(new EventId(0), ex, "Consensus loop at Tip:{0} unhandled exception {1}", this.consensusLoop.Tip?.Height, ex.ToString());
                 NLog.LogManager.Flush();
                 throw;
             }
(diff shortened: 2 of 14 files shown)


Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -138,6 +138,7 @@ public Checkpoints(Network network)
         {
             if (network.Equals(Network.Main)) this.checkpoints = bitcoinMainnetCheckpoints;
             else if (network.Equals(Network.TestNet)) this.checkpoints = bitcoinTestnetCheckpoints;
+            else if (network.Equals(Network.RegTest)) this.checkpoints = new Dictionary<int, CheckpointInfo>();
```
Aprogiena: I don't get it. I'm reading this over and over again and I don't get it :(
What am I missing? Why is this needed?

Aprogiena:
Tried to run the test, but no luck for about 5 minutes (still running). How long is it supposed to run?

Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
dangershony: This is only for the truly knowledgeable in blockchain ;)

dangershony:
Test can run for a while  as it needs to mine 500 blocks, maybe we can make the mining more efficient

Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
dangershony: Seriously, this is related to an issue in NBitcoin where POS sets static flags when initiating the stratis network.
I am working on fixing this.

Aprogiena:
OK, it finished after 7 minutes, which is little long, but maybe OK. However, the test passed, which I don't like :-) I would like the test to fail actually until the CheckSync feature is implemented correctly.

Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
Aprogiena: I still don't understand this. Do you say that network.Equals gives TRUE for some of the IFs we've had there before in case the network was RegTest and not to go with the final else?

dangershony:
We can maybe make it pass faster by making the reorg threshold configurable, we then set it to a lower number and mining less blocks to prevent the reorg

Aprogiena:
Yes that is a good idea, we can test with 50 instead of 500.

Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
Aprogiena: OK, you've explained this to me. Weird!

Aprogiena:
OK, so I've read the whole test now and I think I understand it +-. So yes, it actually tests the behavior that we would like to prevent - at the end this test ends up with all 3 guys synced on the same chain, but we would like two different forks to be there at the end and the nodes diverged forever exactly because the long reorg would be prevented. But fair enough, it gives me enough clues to start with on fixing PR 492.

So in general I'd like to ask you to modify the test accordingly:
- use `PosConsensusOptions.MaxReorgLength` and set it to some lower value, say 25 blocks to have this test finished in much reasonable time (within 30 seconds)
- generate those two forks as you do with length of something more than `PosConsensusOptions.MaxReorgLength`
- actually test that the tested node does not accept the longer chain from reorg node and that it considers itself as fully synced

Aprogiena:
So, I've been testing this a little bit and the test does not use Stratis at all. It rather uses Bitcoin. `PosConsensusOptions` don't seem to be available and when the blocks are accepted, `PosConsensusValidator` is not used at all.

The long reorg prevention should only be implemented on Stratis only, so this test should reflect that. Currently, the test is not very useful, as I hoped, for the development of that feature because those nodes are not running on Stratis network.

Aprogiena:
Also the PR's description should contain `Fix #493` because that is what it aims to do.

dangershony:
Ah right the reorg prevention is the POS validator hmm..
We can simulate this, with the pow part of the pos validator, however it will be hard to run a stratis node,
Let me revisit this test and try it to use the POS network.

PR #514, opened 2017-09-30
Fix or better handle all async loops in the solution [Part 1 - BlockNotification]

fassadlr:
Refactor BlockNotification so that it can be better unit tested.
Move useful create block and append to chain logic to TestBase.
Updated the unit tests and commented on their purpose.
Added a Location property to LookAheadBLockPuller for unit testing purposes.


Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
```diff
@@ -75,58 +72,70 @@ public virtual void SyncFrom(uint256 startHash)
...
+        /// We don't need to wait for this asyncloop to complete as it doesn't rely or depend on
+        /// any other features finishing before shutting down. Stopping the node will stop the async 
+        /// loop task immediately.
+        /// </para>
         /// </summary>
-        public virtual IAsyncLoop Notify()
+        public void NotifyLoop()
         {
-            return this.asyncLoopFactory.Run("block notifier", token =>
+            this.asyncLoopFactory.Run("Notify", async token =>
```
dangershony: Don't you need the IAsyncLoop  to call dispose on shutdown?

fassadlr: Dan, please see the comments on top of the method. We don't need to wait for the async loop to complete as with the ones as they don't depend on any other features completing first ;)

fassadlr: @dangershony on looking at this again, I'm going to implement the dispose properly. We need to follow the pattern everywhere ;) Submitting soon.

Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
```diff
@@ -76,57 +78,73 @@ public virtual void SyncFrom(uint256 startHash)
...
 
-                this.reSync = false;
+                // In reorg we reset the puller to the fork.
+                // When a reorg happens the puller is pushed back and continues from the current fork.
+                // Find the location of the fork.
+                while (this.Chain.GetBlock(this.tip.HashBlock) == null)
+                    this.tip = this.tip.Previous;
+
+                // Set the puller to the fork location.
+                this.Puller.SetLocation(this.tip);
```
dangershony: The  first return `Task.CompletedTask` will break out of the loop right?
This is different form the previous logic, perhaps just remove the `  return Task.CompletedTask;`

Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
```diff
@@ -76,57 +78,73 @@ public virtual void SyncFrom(uint256 startHash)
...
-                {
-                    return Task.CompletedTask;
-                }
+            this.asyncLoop = this.asyncLoopFactory.Run("Notify", async token =>
+           {
+               await Notify(this.nodeLifetime.ApplicationStopping);
+           },
+            this.nodeLifetime.ApplicationStopping,
+            TimeSpans.Ms100);
+        }
```
fassadlr: @dangershony Can you perhaps confirm that the "repeat every" time frame is correct here?

Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
fassadlr: Ah yes, the "While" will execute until the SyncFrom method sets the reSync to true.

Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
dangershony: No it previously was 1000 ms. (default)

PR #515, opened 2017-09-30
Feature/allow multiple debug files issue 430

### LINKED ISSUE #430: [Logging] Allow multiple debug files to be created inside correct data folder
Currently only `debugFile` target is handled in a way that it finds the correct data folder. We could greatly benefit if all targets starting with `debug` prefix would find their correct data folder.

Aprogiena:
A tiny improvement in logging setup, which allows convenient use of multiple log files that are logged into the correct data directory. Note that names of these targets have to start with `debug` prefix. 

This change has been documented in `using-logging.md` documentation file and some new examples of configuration files were also added.

Fix #430


bokobza:
lovely, i needed that :-)

PR #520, opened 2017-10-01
Bugfix/stopwatch issue 362

### LINKED ISSUE #362: Make StopWatch use StopWatch instead of DateTimeOffset
StopwatchDisposable is using DateTimeOffset instead of StopWatch and the reason is commented as 
`//Somehow, the watch is imprecise when time accumulate (maybe due to performance impact of calling underlying high precision API)`

this does not sound right. DateTime should not be used to for serious measurements with low granularity.

Aprogiena:
Replacing implementation of disposable stopwatch that was based on DateTimeOffset with the one that is based on actual stopwatch.

The reason why it was previously switched to DTO was probably misunderstanding that units of `DateTime.Ticks` are different from `Stopwatch.ElapsedTicks`. It then lead to invalid measurements. Should be fixed now as we now use `Stopwatch.Elapsed.Ticks`, which is what the all use cases in the code base expect.

Fix #362


Stratis.Bitcoin/Utilities/Stopwatch.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// Replacement for <see cref="System.Diagnostics.Stopwatch"/> class that allows the caller
```
bokobza: Replacement for **the**  Stopwatch class that allows the caller to use **a** convenient way of calling the watch with **the**  <c>using</c> **statement**...

Stratis.Bitcoin/Utilities/Stopwatch.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// Replacement for <see cref="System.Diagnostics.Stopwatch"/> class that allows the caller 
+    /// to use convenient way of calling the watch with <c>using</c> keyword due to 
+    /// the implementation of <see cref="IDisposable"/> interface.
+    /// </summary>
+    /// <remarks>
+    /// Note that we are using <see cref="DateTime.Ticks"/> as a basic unit of measurement,
```
bokobza: good spot!

Aprogiena:
Fixed that comment. 

> basically a class that merges the two?

Depends. I would say yes. However, it seems to me from the code that the original intention of the author was to avoid creating a new stopwatch every time, which the code that uses this class actually satisfies everywhere (there are just coupe of uses for this). However, in general, such a use is not safe and one has to be very careful. I think it works mostly by accident that no one used it improperly. I've added a warning to reflect that.

The goal of this issue was to replace the time measurement engine inside this class, not to change how this class is being used. I agree that further improvement towards much safer design is probably a good idea, despite the fact that it will require creating new object every time.

There are three possible designs:
- current design, which creates one instance per the class that uses this stopwatch - see ConsensusLoop for example. This is very unsafe, unless one is sure that there are no two parallel executions at the time. Very tricky, but current code satisfies that.
- use one stopwatch per performance counter (actually put SW into PC, not e.g. ConsensusLoop) - this is better than current design and it also avoids new SW allocation with each measurement, but still it needs some carefulness so that one performance counter does not have two parallel executions at the same time.
- use new stopwatch with each measurement - instead of `using (this.watch.Start(..)) ...` you would have `using (new StopwatchDisposable(..)) ...` and this would be perfectly thread-safe + you could have only one class instead of two classes, this would be at the price of creating new `System.Diagnostic.Stopwatch` instance with every measurement.

bokobza:
I think the third option is best, no problem in creating a new stopwatch every time, that's why there is a method called StartNew in there. 
Just called is StopWatch, no need to point out that it's disposable in the name.
Also, if you're gonna fix the original code, just fix it all the way if you can and have time.

Aprogiena:
ok, so now we have the third design with single class and new instance at every use

Stratis.Bitcoin/Utilities/Stopwatch.cs
```diff
@@ -10,78 +10,47 @@ namespace Stratis.Bitcoin.Utilities
...
-        /// <summary>Initializes a new instance of the object.</summary>
-        public Stopwatch()
+        /// <summary>
+        /// Creates a new disposable object and starts the time measurement.
+        /// </summary>
+        /// <param name="action">Action to execute when the measurement is done.</param>
+        public StopwatchDisposable(Action<long> action)
         {
+            this.action = action;
             this.watch = new System.Diagnostics.Stopwatch();
```
bokobza: Can even replace these 2 lines with `this.watch.StartNew();`

Stratis.Bitcoin/Utilities/Stopwatch.cs
```diff
@@ -10,78 +10,47 @@ namespace Stratis.Bitcoin.Utilities
...
+        /// Stops the time measurement and calls the action with the measured elapsed ticks.
         /// </summary>
-        /// <param name="action">Action to execute when the measurement is done. See <see cref="StopwatchDisposable.action"/>.</param>
-        /// <returns>Disposable interface for convenient calling with <c>using</c> keyword.</returns>
-        public IDisposable Start(Action<long> action)
+        public void Dispose()
         {
-            return new StopwatchDisposable(this.watch, action);
+            this.watch.Stop();
+            this.action(this.watch.Elapsed.Ticks);
```
bokobza: should probably check action is not null here, and/or put a Guard on the constructor.
I would say add a null check here so that this class can be used without an action.

Aprogiena: using this class without an action does not make sense to me - i can't see use case, so i'm happy with the guard

Stratis.Bitcoin/Utilities/Stopwatch.cs
Aprogiena: yup

bokobza:
so you think we shouldn't call this class without an action?

bokobza:
ah sorry didn't see your comments.

Aprogiena:
I can't figure out a use case when you would use this instead of normal Stopwatch and not provided action.

Stratis.Bitcoin/Utilities/Stopwatch.cs
bokobza: if for example someone wants to do:
```
StopwatchDisposable sw;
using(sw = new StopwatchDisposable())
{
// some codez
}
var ticks = sw.ElapsedTicks;

**Comment by bokobza** [comment]
maybe just to use the using.

**Comment by Aprogiena** [comment]
but how? for what? i mean show me the code of such use case - how would you obtain the values from the watch then?

**Comment by bokobza** [comment]
I replied in your comments above.

File: Stratis.Bitcoin/Utilities/Stopwatch.cs (hunk 4 above)
Comment (Aprogiena) [reply Stratis.Bitcoin/Utilities/Stopwatch.cs]: OK, so this construction is 5 line construction, which you preferred over
```
Stopwatch sw = Stopwatch.StartNew();
// some code
sw.Stop;
tick = sw.ElapsedTicks;
```
Why on earth would you chose that using here :D 
The only reason why that disposable thing makes sense is that it is 3 line construction (if action is present) and 2 of those lines are brackets, which is possibly better than this 3 line construction I'm having here. But that 5 line thing seems useless to me. Any rationale for that?

bokobza: You're right, but the using also has the advantage of wrapping the code for the user, so that they don't forget to stop the stopwatch. so it has some value, even without an action.

dangershony: I think we can make stop watch more flexible (an override with no action).

Generally I don't even think we need to do it like that, its just confusing I would not use IDisposable at all rather put a stop watch in the context and every measurement take the ElapsedTicks and dispose it at the end.

Aprogiena: Guys, we are getting off topic I think :-) The purpose of this PR was to get avoid of DTO as an internal measurement engine. Now we are talking about completely turning this upside down. 

I don't think that anyone can actually think of reasonable use case for no action call, therefore I would say - don't implement it until you need it. 

I don't agree that "don't forget to stop the stopwatch" is relevant, because sw.ElapsedTicks would work even without stopping the watch. So you don't actually need to stop it. There is no resource leak if you don't stop it.

bokobza: ok mate, it's fine, I'll check it in.

PR #523, opened 2017-10-01
Transaction Fee Refactor

mikedennis:
- Default mintxfee and fallbackfee from network
- Make minrelayfee configurable and default to network value
- Remove mempool validator static field for minrelayfee

### DIFF
--- a/Stratis.Bitcoin/Utilities/NetworkExtensions.cs
+++ b/Stratis.Bitcoin/Utilities/NetworkExtensions.cs
@@ -10,18 +10,6 @@ namespace Stratis.Bitcoin.Utilities
         /// <summary>Fake height value used in Coins to signify they are only in the memory pool (since 0.8).</summary>
         public const int MempoolHeight = 0x7FFFFFFF;
 
-        /// <summary>Default minimum transaction fee for bitcoin network.</summary>
-        private const long DefaultBitcoinMinTxFee = 1000;
-
-        /// <summary>Default fallback fee for bitcoin network.</summary>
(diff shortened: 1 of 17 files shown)


Stratis.Bitcoin/Utilities/NetworkExtensions.cs
```diff
@@ -10,18 +10,6 @@ public static class NetworkExtensions
         /// <summary>Fake height value used in Coins to signify they are only in the memory pool (since 0.8).</summary>
         public const int MempoolHeight = 0x7FFFFFFF;
 
-        /// <summary>Default minimum transaction fee for bitcoin network.</summary>
```
bokobza: so where are all these actually set?

mikedennis: The defaults are now in NStratis in Network when the network is created. This avoids the ugly if (bitcoin) logic.

bokobza: ah is it in your latest change to NStratis?

mikedennis: Yes it is. It's already merged into fullnode. https://github.com/stratisproject/NStratis/commit/d897a44114b564e982fb03b3408457725cdb055d

bokobza: ah ok cool, I'm happy with this.
Can you please just reply to my other comments and I'll push it in.

mikedennis: There you go!

PR #524, opened 2017-10-01
Fix segwit mempool validator test

mikedennis:
@nopara73 if you could take a quick look at this and verify I'm building the segwit tx correctly? I believe it's what you sent me. The mempool validator does accept the p2sh over p2wpkh tx but I guess that is probably because it's p2sh. I will have to dig further into the validator logic and see what exactly it is doing with this tx.


Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
```diff
@@ -404,25 +404,24 @@ public void AcceptToMemoryPool_WithSegWitValidTxns_IsSuccessfull()
             // 50 Coins come from first tx on chain - send bob 42 and change back to miner
             ScriptCoin witnessCoin = new ScriptCoin(context.SrcTxs[0].GetHash(), 0, context.SrcTxs[0].TotalOut, miner.PubKey.ScriptPubKey.WitHash.ScriptPubKey.Hash.ScriptPubKey, miner.PubKey.ScriptPubKey);
             TransactionBuilder txBuilder = new TransactionBuilder();
-            Transaction p2wshTx = txBuilder
+            Transaction p2shOverp2wpkh = txBuilder
```
dangershony: Argh this can get so confusing `miner.PubKey.ScriptPubKey.WitHash.ScriptPubKey.Hash.ScriptPubKey`
It would be a good idea to put an example of how the script looks like for future devs that visit this code.

dangershony:
Ah sorry I assume this was ready to be committed @nopara73 ?

PR #528, opened 2017-10-02
Fix or better handle all async loops in the solution [Part 3 - Mempool]

fassadlr:
Refactor mempool features to properly dispose of async loops


Stratis.Bitcoin.Features.MemoryPool/MempoolSignaled.cs
```diff
@@ -88,5 +88,11 @@ private void RelayWorker()
             repeatEvery: TimeSpans.TenSeconds,
             startAfter: TimeSpans.TenSeconds);
         }
+
+        public void Stop()
+        {
+            if (this.asyncLoop != null)
```
dangershony: FYI could be cleaner to do 
``` this.asyncLoop?.Dispose();```

PR #530, opened 2017-10-03
Update console if no wallet file for full node and light wallets.

mikedennis:


Stratis.Bitcoin.Features.Wallet/IWalletManager.cs
```diff
@@ -210,6 +210,12 @@ public interface IWalletManager : IDisposable
         /// <returns>The wallet filenames, along with the folder in which they're contained.</returns>
         (string folderPath, IEnumerable<string>) GetWalletsFiles();
 
+        /// <summary>
```
bokobza: I would rather have AreWallets be called ContainsWallets or something like that.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -402,6 +402,15 @@ public int LastBlockHeight()
             return this.Wallets.Min(w => w.AccountsRoot.SingleOrDefault(a => a.CoinType == this.coinType)?.LastBlockSyncedHeight) ?? 0;
         }
 
+        /// <inheritdoc />
+        public bool AreWallets
```
bokobza: Can we have this on one line please?

Stratis.Bitcoin.Features.Wallet/IWalletManager.cs
mikedennis: Yes I like that better. I changed it to ContainsWallets.

bokobza:
Thanks @mikedennis , are you going to do the same for the Consensus line?

Issue #522, opened 2017-10-01
[Checkpoints] Investigate the delay during IBD when reaching the latest checkpoint


Aprogiena:
There is a significant delay in IBD when the node's consensus reaches the height of the last checkpoint. Eventually, the node advances, but something weird is going on at that very moment and it is certainly worth exploring. However, it is likely that fixing other issues that are open already might fix this as well, so this should be examined not too early.

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[Checkpoints] Investigate the delay during IBD when reaching the latest checkpoint](https://trello.com/c/B80yYdXl/271-checkpoints-investigate-the-delay-during-ibd-when-reaching-the-latest-checkpoint)

Aprogiena:
Examination done on the top of PR https://github.com/stratisproject/StratisBitcoinFullNode/pull/534 and it produced no such result. It may be the case that the problem was mitigated. Testing once more on the top of this PR and then will do with the current master.

Aprogiena:
Second run on top of PR 534 - no such result again and it seems like the previous idea of connection to #525 was not correct. Running once more to confirm and then one last time on current master (without PR 534)

Aprogiena:
Third run, no such result.

Aprogiena:
Fourth run on master revealed the problem. The delay is most likely present due to stake modifier v1 calculations that require loading block stake information from the disk, which takes a lot of time. Block stake database is not used for checkpointed blocks and it is thus needed to reload a lot of data from disk to count the next block stake modifier v1. PR 534 removes calculation and thus seems to solve the issue in full.

dangershony:
Ha wowowo nice spot away with you V1

Aprogiena:
Reading the logs I can confirm this was the issue. With last checkpoint set at 50k, it was loading 46k block stake entries from the disk, one by one.

Aprogiena:
This issue was fixed by #534

Issue #477, opened 2017-09-20
[Speeding up] Improvements of checkpoints and beyond 


Aprogiena:
We've implemented hardcoded checkpoints to save some processing time on old blocks.

However, there is still some space for further improvements. Currently, we are aware of these:

1. POS StakeModifier (v1) is not used anymore for V3 blocks (which is place for quite a long time already, so latest checkpoints are V3). However, we still spend a lot of CPU and I/O time on calculating this v1 StakeModifier. We think we can remove this completely, which will speed up validation time on blocks after last checkpoint.

2. `AcceptBlock` method in `ConsensusLoop` contains this line:
```
                context.Flags = this.NodeDeployments.GetFlags(context.BlockResult.ChainedBlock);
```
It seems that it takes quite a lot of time to execute this line and the question is if that time could be saved for historic blocks - i.e. before last checkpoint. If so, this could save significant amount of time during IBD with checkpoints enabled.

3. Transactions and blocks are still validated even for historic blocks. This consumes a lot of CPU especially on mainnets. If we could skip validations, we could gain extremely fast syncing times.

Aprogiena:
@dangershony wants to remove V2 modifier from checkpoint and compute it always as it is negligible once we remove V1 completely

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[Speeding up] Improvements of checkpoints and beyond](https://trello.com/c/4Y4a0VrA/247-speeding-up-improvements-of-checkpoints-and-beyond)

Aprogiena:
1. will be done as a part of #494

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[Speeding up] 3 Improvements of checkpoints and beyond](https://trello.com/c/IgyooTGn/296-speeding-up-3-improvements-of-checkpoints-and-beyond)

PR #534, opened 2017-10-04
Refactor/remove pre v3 consensus issue 494

### LINKED ISSUE #494: [POS] Get rid of V1 POS code
We want to get rid of all POS v1 related code from the code base as we don't need it on Stratis chain. These versions of POS consensus are from old coins project that Stratis evolved from.

Aprogiena:
This is quite a huge PR that removes V1 and V2 consensus from Stratis network. It also includes lots of new comments, lots of code style fixes and code polishing, some minor simplifications of logic here and there, but overall it makes whole code related to Stratis network consensus much cleaner.

There is a significant speedup in syncing included - this is due to complete removal of stake modifier v1. This also solves first part of issue #477.

This also seems to solve #522 - see rationale there.

Fix #494 #522


Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -104,7 +95,7 @@ public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
                     }
                 }
 
-                this.AcceptBlock(new ContextInformation(result, this.Validator.ConsensusParams));
+                this.AcceptBlock(new ContextInformation(result, this.Tip, this.Validator.ConsensusParams));
```
dangershony: I don't think you need to pass in the tip here, tip should only ever be set from this thread right?

Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -290,58 +278,44 @@ public override void ContextualCheckBlockHeader(ContextInformation context)
...
             if (context.Stake.BlockStake.IsProofOfStake()
-                && !PosConsensusValidator.CheckCoinStakeTimestamp(chainedBlock.Height, chainedBlock.Header.Time, context.BlockResult.Block.Transactions[1].Time))
+                && !this.CheckCoinStakeTimestamp(chainedBlock.Height, chainedBlock.Header.Time, context.BlockResult.Block.Transactions[1].Time))
             {
                 this.logger.LogTrace("(-)[BAD_TIME]");
                 ConsensusErrors.StakeTimeViolation.Throw();
             }
 
             // Check timestamp against prev.
-            if ((chainedBlock.Header.Time <= StakeValidator.GetPastTimeLimit(chainedBlock.Previous))
```
dangershony: What happened to this check?
```
FutureDrift(chainedBlock.Header.Time, chainedBlock.Height) < chainedBlock.Previous.Header.Time))		
```

Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -476,21 +440,23 @@ public void CheckAndComputeStake(ContextInformation context)
             // Record proof hash value.
             blockStake.HashProof = context.Stake.HashProofOfStake;
 
-            if (pindex.Height > lastCheckpointHeight)
+            if (chainedBlock.Height > lastCheckpointHeight)
```
dangershony: Did we not decide to calculate stake modifier in any case as the calculation is not a heavy operation.

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -641,19 +261,10 @@ public void CheckKernel(ContextInformation context, ChainedBlock pindexPrev, uin
...
-                    this.logger.LogTrace("(-)[LOW_COIN_AGE]");
-                    ConsensusErrors.InvalidStakeDepth.Throw();
-                }
-            }
-            else
+            if (this.IsConfirmedInNPrevBlocks(prevUtxo, pindexPrev, this.consensusOptions.StakeMinConfirmations - 1))
             {
-                uint nTimeBlockFrom = prevBlock.Header.Time;
-                if (nTimeBlockFrom + this.consensusOptions.StakeMinAge > nTime)
-                    ConsensusErrors.MinAgeViolation.Throw();
```
dangershony: This rule can be deleted? `MinAgeViolation`

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: you may be right, i've added this while doing 413, but maybe it is no longer needed, i'll check that

Aprogiena: yes, i think you are right, but only after PR #549, which solves the issue completely, so i'll remove this as a part of PR #549

Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
Aprogiena: It should be described in the commit that removed that. That check is void, refer to that commit description pls.

Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
Aprogiena: We kind of did that, although I forgot the reason why. But it is not relevant to this PR. This PR was about taking away V1 and V2 code and checkpoints are different issue. So if we still want to do that, which is possible, but I forget the reason why it would be good to do (still consumes some CPU, although not much), then we can do it, just not as a part of this PR.

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
Aprogiena: Ah yes, you mean the error itself, yes, this can be removed and I'll do that as part of PR 549.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: fixed there

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
Aprogiena: fixed there

Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
dangershony: ah ok I need to find it its not visible when viewing the entire set

dangershony: Ah ok yeah once you strip away the GetPastTimeLimit its more clear the second condition has no effect.

PR #535, opened 2017-10-04
Unit tests for WalletSyncManager

Neurosploit:
Commits of particular interest:  
4f99028 - Refactoring done to make testing a lot easier to do on this class. Otherwise the walletmanager cast would have made it very difficult. Let me know what you think of it. Doing the same refactor to lightsyncwalletmanager makes it easy to test that as well after this walletmanager refactor from this pull request.
b2a5593 - Ran into and fixed an issue in proceessblock. Already shortly notified dan of the bug that was in it. It has to do with being able to recover using the blockstorecache after the discovery was made earlier that we are not on the best chain anymore.

### DIFF
--- a/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
+++ b/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
@@ -11,7 +11,7 @@ namespace Stratis.Bitcoin.Features.Wallet
 {
     public class WalletSyncManager : IWalletSyncManager
     {
-        protected readonly WalletManager walletManager;
+        protected readonly IWalletManager walletManager;
         protected readonly ConcurrentChain chain;
         protected readonly CoinType coinType;
         protected readonly ILogger logger;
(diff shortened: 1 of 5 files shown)


Neurosploit:
I see that one test is still failing on the build. I'll have a look at it.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -85,22 +98,25 @@ public virtual void ProcessBlock(Block block)
...
+                    var fork = walletTip;
 
                     // We walk back the chained block object to find the fork.
                     while (this.chain.GetBlock(fork.HashBlock) == null)
                         fork = fork.Previous;
 
                     this.walletManager.RemoveBlocks(fork);
+
+                    // update temporary wallet tip to be able to catch up again.
+                    walletTip = this.chain.GetBlock(fork.HashBlock);
```
dangershony: Ah I see why I missed this, previously the second if was an else so I didn't need to remember the fork.

Good catch.

Neurosploit:
Fixed the branch my tests are all passing again. The current failing tests have to do with fee estimation outside the scope of this pull request.

Issue #500, opened 2017-09-26
[Mining] PowBlockAssembler race condition


Aprogiena:
POS Miner crashed with
```
 PosMining.Mine threw an unhandled exception: System.ArgumentException: The previous block has not the expected hash
   at Stratis.Bitcoin.Features.Miner.PosMinting.<>c__DisplayClass35_0.<Mine>b__0(CancellationToken token) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Miner\PosMinting.cs:line 206
   at Stratis.Bitcoin.Utilities.AsyncLoop.<>c__DisplayClass12_0.<<StartAsync>b__0>d.MoveNext() in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Utilities\AsyncLoop.cs:line 129
```

The origin of this exception is:
`PosMinting.GenerateBlocks` -> `PosBlockAssembler.CreateNewBlock` -> `PowBlockAssembler.CreateNewBlock` ->  `PosBlockAssembler.UpdateHeaders` -> `PowBlockAssembler.UpdateHeaders`

here we call 
```
            this.pblock.Header.HashPrevBlock = this.pindexPrev.HashBlock;
            this.pblock.Header.UpdateTime(this.dateTimeProvider.GetTimeOffset(), this.network, this.chain.Tip);
            this.pblock.Header.Bits = this.pblock.Header.GetWorkRequired(this.network, this.chain.Tip);
```

which already by itself seems wrong because those two lines potentially use different chain Tip. The problem is then raised from `GetWorkRequired`, which leads to

```
return new ChainedBlock(this, null, prev).GetWorkRequired(consensus);
```
and the constructor has this check
```
				if(previous.HashBlock != header.HashPrevBlock)
					throw new ArgumentException("The previous block has not the expected hash");
```

Solution is probably to use `pindexPrev` instead of `this.chain.Tip` in both lines.

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[Mining] PowBlockAssembler race condition](https://trello.com/c/ozmGxAny/262-mining-powblockassembler-race-condition)

Aprogiena:
However, a better fix would probably be to pass chain tip as an argument to `CreateNewBlock` or maybe even the constructor of `PowBlockAssembler`.

Any thoughts @dangershony ?

Aprogiena:
[...]

PR #538, opened 2017-10-05
Don't wait for peers when minting new POS block.

### LINKED ISSUE #403: [POS] Block rejected by peers
If at least one peer does not accept our new block in 100 seconds, we lost 100 seconds of mining, then we raise exception and start over. This should be optimized.

mikedennis:
Fixes #403

### DIFF
--- a/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -13,7 +13,6 @@ using Stratis.Bitcoin.Utilities;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using IBlockRepository = Stratis.Bitcoin.Features.BlockStore.IBlockRepository;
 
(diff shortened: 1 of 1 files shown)


dangershony:
Hi @mikedennis looks good did you try to mint?

mikedennis:
@dangershony I got some advice from apro and going to get setup to do some POS minting to verify this works. Hold off merging until I can confirm it doesn't break anything.

dangershony:
@mikedennis are you happy with this change now?

mikedennis:
Not sure, had this error. It doesn't seem related #554

Aprogiena:
This one is #500, fixed by my next PR. #554 does not seem related, but it seems quite new.

Aprogiena:
OK, 554 is also not related.

mikedennis:
I can no longer make #554 happen but had #500 (previous block unexpected hash) happen several times.

dangershony:
ok i think I will merge this, it is indeed not related to #554

dangershony:
FYI I think this line can now be removed
```
 this.blockRepository.PutAsync(context.BlockResult.ChainedBlock.HashBlock, new List<Block> { block }).GetAwaiter().GetResult();
```
This signaler will not relay to peers until a block is persisted to store.
I will fix that when I do the consensus fix

PR #549, opened 2017-10-06
Bugfix/pow block asm race issue 500

### LINKED ISSUE #500: [Mining] PowBlockAssembler race condition

Aprogiena:
Depends on PR https://github.com/stratisproject/StratisBitcoinFullNode/pull/534

chainTip now comes as an argument to the block assembler, so the whole assembling instance is working with a fixed chain tip value and this prevents problems when the concurrent chain tip or consensus tip is changed during the process

Fix #500


dangershony:
Up to now looks good but can you resolve the conflicts?

Aprogiena:
These are expected when I continue on the previously unmerged branch, so of course, I'm going to fix the conflicts very soon.

PR #551, opened 2017-10-06
fix feetests

nopara73:


dangershony:
This should be fixed if you update the submodule

PR #559, opened 2017-10-09
feature/logs to wallet and consensus loop

Aprogiena:
Adds logs and fixes comments and code style in consensus loop and wallet manager and wallet sync manager.

### DIFF
--- a/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
+++ b/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
@@ -31,7 +31,6 @@ namespace Stratis.Bitcoin.Features.Consensus
         public ChainedBlock Tip { get; private set; }
         public NodeDeployments NodeDeployments { get; private set; }
 
-
         public ConsensusLoop(PowConsensusValidator validator, ConcurrentChain chain, CoinView utxoSet, LookaheadBlockPuller puller, NodeDeployments nodeDeployments, ILoggerFactory loggerFactory, StakeChain stakeChain = null)
         {
             Guard.NotNull(validator, nameof(validator));
--- a/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
+++ b/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
@@ -1,4 +1,5 @@
-﻿using NBitcoin;
+﻿using Microsoft.Extensions.Logging;
+using NBitcoin;
 using Stratis.Bitcoin.Base.Deployments;
 using Stratis.Bitcoin.BlockPulling;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
@@ -19,13 +20,27 @@ namespace Stratis.Bitcoin.Features.Consensus
     
     public class ConsensusLoop
     {
-        public ConsensusLoop(PowConsensusValidator validator, ConcurrentChain chain, CoinView utxoSet, LookaheadBlockPuller puller, NodeDeployments nodeDeployments, StakeChain stakeChain = null)
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        public StakeChain StakeChain { get; }
+        public LookaheadBlockPuller Puller { get; }
+        public ConcurrentChain Chain { get; }
+        public CoinView UTXOSet { get; }
+        public PowConsensusValidator Validator { get; }
+        public ChainedBlock Tip { get; private set; }
+        public NodeDeployments NodeDeployments { get; private set; }
+
+
+        public ConsensusLoop(PowConsensusValidator validator, ConcurrentChain chain, CoinView utxoSet, LookaheadBlockPuller puller, NodeDeployments nodeDeployments, ILoggerFactory loggerFactory, StakeChain stakeChain = null)
         {
             Guard.NotNull(validator, nameof(validator));
             Guard.NotNull(chain, nameof(chain));
             Guard.NotNull(utxoSet, nameof(utxoSet));
             Guard.NotNull(puller, nameof(puller));
             Guard.NotNull(nodeDeployments, nameof(nodeDeployments));
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
 
             this.Validator = validator;
             this.Chain = chain;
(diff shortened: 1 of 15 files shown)


bokobza:
I'll do the wallet bits @dangershony @Aprogiena

bokobza:
Ok it looks good for the wallet-related files.
However, @Aprogiena if you could tell us how you create these logging it would be great, I would have no idea how to follow the methodology you're using.
For example, a line like this
`this.logger.LogTrace("(-):*.{0}={1}", nameof(res.Count), res.Count);`
I have no idea what it means :grimacing:

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -78,41 +90,56 @@ public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
...
+                            if (rewinded == null)
                             {
-                                var hash = this.UTXOSet.Rewind().GetAwaiter().GetResult();
-                                var rewinded = this.Chain.GetBlock(hash);
-                                if (rewinded == null)
-                                    continue;
-                                this.Tip = rewinded;
-                                this.Puller.SetLocation(rewinded);
-                                break;
+                                this.logger.LogTrace("Rewound to '{0}', which is still not a part of the current best chain, rewinding further.", hash);
```
dangershony: `Rewound` no?

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -78,41 +90,56 @@ public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
...
         {
+            this.logger.LogTrace("()");
+
             using (new StopwatchDisposable(o => this.Validator.PerformanceCounter.AddBlockProcessingTime(o)))
             {
                 // Check that the current block has not been reorged.
                 // Catching a reorg at this point will not require a rewind.
                 if (context.BlockResult.Block.Header.HashPrevBlock != this.Tip.HashBlock)
+                {
+                    this.logger.LogTrace("Reorganization detected.");
```
dangershony: I am not sure this is a reorg, its rather a bug the wrong block was sent in to the consensus engine.
Its a requirement that the next block must always be pushed.

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -14,19 +14,24 @@ namespace Stratis.Bitcoin.Features.Consensus
 {
     public class StakeValidator
     {
-        // Ratio of group interval length between the last group and the first group.
-        private const int ModifierIntervalRatio = 3;
+        /// <summary>Expected (or target) block time in seconds.</summary>
+        public const int TargetSpacingSeconds = 64;
```
dangershony: ModifierIntervalRatio  has been deleted?

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -300,12 +305,37 @@ public static ChainedBlock GetLastBlockIndex(StakeChain stakeChain, ChainedBlock
...
+        /// Over <see cref="RetargetIntervalMinutes"/> minutes there are certain number (say <c>N</c>) of blocks expected to be mined if the target block time 
+        /// of <see cref="TargetSpacingSeconds"/> was reached every time. Then the next target is calculated as follows:</para>
+        /// <code>
+        /// NewTarget = PrevTarget * ((N - 1) * TargetSpacingSeconds + 2 * LastBlockTime) / ((N + 1) * TargetSpacingSeconds)
+        /// </code>
+        /// <para>
+        /// Which basically says that the block time of the last block is counted twice instead of two optimal block times.
+        /// And the <c>N</c> determines how strongly will the deviation of the last block time affect the difficulty.
+        /// </para>
+        /// </remarks>
```
dangershony: Very good explanation thanks.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -46,50 +50,59 @@ public class WalletSyncManager : IWalletSyncManager
...
                 this.walletTip = fork;
             }
+
+            this.logger.LogTrace("(-)");
         }
 
         /// <inheritdoc />
         public void Stop()
         {
+            this.logger.LogTrace("()");
```
dangershony: double log?

dangershony:
Well done.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: as for what i found, rewound should be past tense of rewind, so this should be correct, but no native speaker here :)

Aprogiena:
@bokobza that is actually very very good comment and i plan to write it down for several weeks already, i've got a basic draft and my intention was to finish this sprint and write it down, so ... soon

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: Sometimes I tend to follow the inline comment when I insert logs to code I am not that familiar with. So if I read that we are catching a reorg here, I just write reorg to the log. 

However, when this actually happened in real life to my node, it was during reorg, so I believe it is possibly correct here. If there is another possibility in case of bug, that is not very interesting as the log should reflect the correct and expected behavior.

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
Aprogiena: well ... it was and the code compiles ...

Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -300,12 +305,37 @@ public static ChainedBlock GetLastBlockIndex(StakeChain stakeChain, ChainedBlock
             return index;
         }
 
-        public static Target GetNextTargetRequired(StakeChain stakeChain, ChainedBlock indexLast, NBitcoin.Consensus consensus, bool proofOfStake)
+        /// <summary>
```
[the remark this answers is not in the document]
Aprogiena: it is supposed to be explained by the next sentence "so even a big change in the mining power on the network will be fixed by retargeting relatively quickly." 

What it means that the function of block time (spacing) is converging to the expected target spacing very quickly. So imagine your target spacing is 64 seconds and imagine your block time was 2000 seconds. If the function was slower, which could be done by trying to modify the target difficulty just by a bit to reach next target block time 10 seconds earlier, the function would go like this: 2000 -> 1990 -> 1980 -> 1970 ... and eventually reached 70 -> 64. That would be very slow. This happens when you add/subtract constant. What we do instead is that we multiply, which creates geometric decay - https://en.wikipedia.org/wiki/Exponential_growth and let's say we would only use constant factor of 10 % difference - i.e. 1.10x multiplier. It would go like this: 2000 -> 2000 / 1.1 = 1818 -> 1818 / 1.1 = 1652 -> 1652 / 1.1 = 1502 ... So it goes down much much faster. And this is almost what we do here except that the multiplier is not constant, but the overall exponential decay is preserved.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: yes, intended

Issue #556, opened 2017-10-08
[WalletSyncManager] ProcessBlock race condition


Aprogiena:
On this line https://github.com/stratisproject/StratisBitcoinFullNode/blob/d35ecdd20bf3585f224d510ff1766b4887bbfdf2/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs#L111 the `walletTip` can be assigned `null` if the concurrent chain changes just before it to earlier reorg.

Aprogiena:
Why it is not `walletTip = fork` ?

Aprogiena:
And just below that 
```
ChainedBlock incomingBlock = this.chain.GetBlock(block.GetHash());
```
again, this can produce `null` in `incomingBlock` for the very same reason.

Aprogiena:
And once more 
```
next = this.chain.GetBlock(next.Height + 1);
```
Again, this can produce `null` in `next`.

dangershony:
Yes very good points, we sould indeed use this ` walletTip = fork`.
As for the other issues I guess we do want to call chain but if we get back null then the current block received in the wallet is not valid and the wallet should return execution (let the next block sort the reorg issue). 

Any fix here should also be ported tot he light wallet

Aprogiena:
There are two options here, either fix every line of code that works with the chain like that, or make a big change in a way that ConcurrentChain is not used like this anywhere. I believe that a better design to work with would be each instance working with its own copy of chain headers (now there is a shared copy called ConcurrentChain). And you could always be sure your own copy won't change over time and thus the code is safe. And the copy should provide a mechanism for you to check if the local copy is the best chain the node knows.

dangershony:
ConcurrentChain is really just holding the pointer to the best TIP (we don't need every feature to hold an entire copy of it just the TIP)

So I agree we should not use ConcurrentChain (almost anywhere but only at the point where the feature determines if its on the right chain).

All other checks must be done against the features current TIP.
That way if a reorg happened the internals of a feature wont know about it and will not be subject to the race condition, then the feature can control the reorg time and place.

Aprogiena:
Not every feature, but the Tip is usually not the only thing they use. Many features use GetBlock as a way to check if they are on the right chain, that usage is prone to problems. And basically any walking through the headers should be done on a local copy. With a good implementation, this is not heavy at all. Note that your local copy only needs to be an actual copy for the headers that are different.

dangershony:
[...]

PR #561, opened 2017-10-09
Wallet bug 556 fixing the ConcurrentChain race condition 

dangershony:
I came up with what can be a fix for the ConcurrentChain race.
Comments on #556 make note that using ConcurrentChain randomly is a bad idea.
This PR tries to fix that issue, for the wallet only.

In general we need to try to limit the usage of ConcurrentChain on features and try to use more the TIP of a feature.

Please help review that properly @bokobza @Aprogiena @Neurosploit

### DIFF
--- a/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
+++ b/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
@@ -110,6 +110,16 @@ namespace Stratis.Bitcoin.Features.LightWallet
         /// <inheritdoc />
         public void ProcessBlock(Block block)
         {
+            Guard.NotNull(block, nameof(block));
+
+            // If the new block previous hash is the same as the 
+            // wallet hash then just pass the block to the manager. 
+            var newTip = this.chain.GetBlock(block.GetHash());
+
+            // new block is not in the main chain nothing more to do right now. 
+            if (newTip == null)
+                return; // reorg
+
             // if the new block previous hash is the same as the 
             // wallet hash then just pass the block to the manager 
             if (block.Header.HashPrevBlock != this.walletTip.HashBlock)
@@ -130,22 +140,30 @@ namespace Stratis.Bitcoin.Features.LightWallet
 
                     Guard.Assert(fork.HashBlock == block.Header.HashPrevBlock);
                     this.walletManager.RemoveBlocks(fork);
+                    this.walletTip = fork;
                     this.logger.LogWarning($"Reorg detected, wallet tip reverted back to Height = {fork.Height} hash = {fork.HashBlock}.");
+                }
+
+                if (newTip.Height > this.walletTip.Height)
+                {
+                    // check that the new tip is in the chain of the wallet tip
+                    var findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
+                    Guard.Assert(findTip == this.walletTip); // this should never happen
+
+                    // the wallet is falling behind we need to catch up
+                    this.logger.LogWarning($"block received with height: {newTip.Height} and hash: {block.Header.GetHash()} is too far in advance. put the puller back.");
+                    this.blockNotification.SyncFrom(this.walletTip.HashBlock);
+                    return;
                 }
                 else
                 {
-                    ChainedBlock incomingBlock = this.chain.GetBlock(block.GetHash());
-                    if (incomingBlock.Height > this.walletTip.Height)
-                    {
-                        // the wallet is falling behind we need to catch up
-                        this.logger.LogWarning($"block received with height: {inBestChain.Height} and hash: {block.Header.GetHash()} is too far in advance. put the puller back.");
-                        this.blockNotification.SyncFrom(this.walletTip.HashBlock);
-                        return;
-                    }
+                    // check that the new tip is in the chain of the wallet tip
+                    var findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
+                    Guard.Assert(findTip == newTip); // this should never happen
                 }
             }
 
-            this.walletTip = this.chain.GetBlock(block.GetHash());
+            this.walletTip = newTip;
             this.walletManager.ProcessBlock(block, this.walletTip);
         }
 
--- a/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
+++ b/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
@@ -113,7 +113,7 @@ namespace Stratis.Bitcoin.Features.LightWallet
             Guard.NotNull(block, nameof(block));
             this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
 
-            var newTip = this.chain.GetBlock(block.GetHash());
+            ChainedBlock newTip = this.chain.GetBlock(block.GetHash());
             if (newTip == null)
             {
                 this.logger.LogTrace("(-)[NEW_TIP_REORG]");
@@ -131,7 +131,6 @@ namespace Stratis.Bitcoin.Features.LightWallet
                 {
                     // The current wallet hash was not found on the main chain.
                     // A reorg happened so bring the wallet back top the last known fork.
-
                     ChainedBlock fork = this.walletTip;
 
                     // We walk back the chained block object to find the fork.
@@ -143,10 +142,10 @@ namespace Stratis.Bitcoin.Features.LightWallet
                     this.walletManager.RemoveBlocks(fork);
                     this.walletTip = fork;
 
-                    this.logger.LogTrace("Wallet tip set to '{0}'.", this.walletTip);
+                    this.logger.LogTrace("Wallet tip set to '{0}/{1}'.", this.walletTip.HashBlock, this.walletTip.Height);
                 }
 
-                // The new tip can be ahead or behind the wallet
+                // The new tip can be ahead or behind the wallet.
                 // If the new tip is ahead we try to bring the wallet up to the new tip.
                 // If the new tip is behind we just check the wallet and the tip are in the same chain.
 
@@ -161,7 +160,7 @@ namespace Stratis.Bitcoin.Features.LightWallet
 
                     this.logger.LogTrace("Wallet tip '{0}/{1}' is behind the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
 
-                    // the wallet is falling behind we need to catch up
+                    // The wallet is falling behind we need to catch up.
                     this.logger.LogWarning("New tip '{0}/{1}' is too far in advance, put the puller back.", newTip.HashBlock, newTip.HashBlock);
                     this.blockNotification.SyncFrom(this.walletTip.HashBlock);
                     return;
@@ -174,6 +173,7 @@ namespace Stratis.Bitcoin.Features.LightWallet
                         this.logger.LogTrace("(-)[NEW_TIP_BEHIND_NOT_IN_WALLET]");
                         return;
                     }
+
                     this.logger.LogTrace("Wallet tip '{0}/{1}' is ahead or equal to the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
                 }
             }
--- a/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
+++ b/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
@@ -111,60 +111,78 @@ namespace Stratis.Bitcoin.Features.LightWallet
         public void ProcessBlock(Block block)
         {
             Guard.NotNull(block, nameof(block));
+            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
 
-            // If the new block previous hash is the same as the 
-            // wallet hash then just pass the block to the manager. 
             var newTip = this.chain.GetBlock(block.GetHash());
-
-            // new block is not in the main chain nothing more to do right now. 
             if (newTip == null)
-                return; // reorg
+            {
+                this.logger.LogTrace("(-)[NEW_TIP_REORG]");
+                return;
+            }
 
-            // if the new block previous hash is the same as the 
-            // wallet hash then just pass the block to the manager 
+            // If the new block's previous hash is the same as the 
+            // wallet hash then just pass the block to the manager. 
             if (block.Header.HashPrevBlock != this.walletTip.HashBlock)
             {
-                // if previous block does not match there might have 
-                // been a reorg, check if the wallet is still on the main chain
+                // If previous block does not match there might have 
+                // been a reorg, check if the wallet is still on the main chain.
                 ChainedBlock inBestChain = this.chain.GetBlock(this.walletTip.HashBlock);
                 if (inBestChain == null)
                 {
-                    // the current wallet hash was not found on the main chain
-                    // a reorg happenend so bring the wallet back top the last known fork
+                    // The current wallet hash was not found on the main chain.
+                    // A reorg happened so bring the wallet back top the last known fork.
 
                     ChainedBlock fork = this.walletTip;
 
-                    // we walk back the chained block object to find the fork
+                    // We walk back the chained block object to find the fork.
                     while (this.chain.GetBlock(fork.HashBlock) == null)
                         fork = fork.Previous;
 
-                    Guard.Assert(fork.HashBlock == block.Header.HashPrevBlock);
+                    this.logger.LogInformation("Reorg detected, going back from '{0}/{1}' to '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, fork.HashBlock, fork.Height);
+
                     this.walletManager.RemoveBlocks(fork);
                     this.walletTip = fork;
-                    this.logger.LogWarning($"Reorg detected, wallet tip reverted back to Height = {fork.Height} hash = {fork.HashBlock}.");
+
+                    this.logger.LogTrace("Wallet tip set to '{0}'.", this.walletTip);
                 }
+
+                // The new tip can be ahead or behind the wallet
+                // If the new tip is ahead we try to bring the wallet up to the new tip.
+                // If the new tip is behind we just check the wallet and the tip are in the same chain.
 
                 if (newTip.Height > this.walletTip.Height)
                 {
-                    // check that the new tip is in the chain of the wallet tip
-                    var findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
-                    Guard.Assert(findTip == this.walletTip); // this should never happen
+                    ChainedBlock findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
+                    if (findTip == null)
+                    {
+                        this.logger.LogTrace("(-)[NEW_TIP_AHEAD_NOT_IN_WALLET]");
+                        return;
+                    }
+
+                    this.logger.LogTrace("Wallet tip '{0}/{1}' is behind the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
 
                     // the wallet is falling behind we need to catch up
-                    this.logger.LogWarning($"block received with height: {newTip.Height} and hash: {block.Header.GetHash()} is too far in advance. put the puller back.");
+                    this.logger.LogWarning("New tip '{0}/{1}' is too far in advance, put the puller back.", newTip.HashBlock, newTip.HashBlock);
                     this.blockNotification.SyncFrom(this.walletTip.HashBlock);
                     return;
                 }
                 else
                 {
-                    // check that the new tip is in the chain of the wallet tip
-                    var findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
-                    Guard.Assert(findTip == newTip); // this should never happen
+                    ChainedBlock findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
+                    if (findTip == null)
+                    {
+                        this.logger.LogTrace("(-)[NEW_TIP_BEHIND_NOT_IN_WALLET]");
+                        return;
+                    }
+                    this.logger.LogTrace("Wallet tip '{0}/{1}' is ahead or equal to the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
                 }
             }
+            else this.logger.LogTrace("New block follows the previously known block '{0}/{1}'.", this.walletTip.HashBlock, this.walletTip.Height);
 
             this.walletTip = newTip;
-            this.walletManager.ProcessBlock(block, this.walletTip);
+            this.walletManager.ProcessBlock(block, newTip);
+
+            this.logger.LogTrace("(-)");
         }
 
         /// <inheritdoc />
--- a/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -8,7 +8,6 @@ using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Features.MemoryPool;
 using Stratis.Bitcoin.Features.Wallet.Interfaces;
 using Stratis.Bitcoin.Features.Wallet.JsonConverters;
-using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.Tests.Logging;
 using Stratis.Bitcoin.Tests.Utilities;
 using Stratis.Bitcoin.Utilities;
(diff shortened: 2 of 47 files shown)


Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -86,39 +86,49 @@ public virtual void ProcessBlock(Block block)
...
                     // The current wallet hash was not found on the main chain.
-                    // A reorg happenend so bring the wallet back top the last known fork.
+                    // A reorg happened so bring the wallet back top the last known fork.
 
-                    var fork = walletTip;
+                    var fork = this.walletTip;
 
                     // We walk back the chained block object to find the fork.
                     while (this.chain.GetBlock(fork.HashBlock) == null)
                         fork = fork.Previous;
```
Aprogiena: In general, I find it very difficult to get my brain grasping this loop. Why? Consider how much `this.chain` can change in this method and in this loop. At any point in time in this loop, `this.chain` can be different chain. So we might start with chain 1:
```
A-B-C-D-E-F-G
```

and we just jumped onto different chain 2:

```
A-B-C-D-E2-F2-G2-H2
```
So, the wallet tip when we entered the method was G, so we are not executing this loop to find the fork - we are hoping to find D. So we go from G to F, but then the chain switches to chain 3:

```
A-B3-C3-D3-E3-F3-G3-H3-I3
```

we know nothing, so we continue down the chain to E, then to D, then to C and now the chain switches again to 

```
A-B-C-D-E2-F2-G2-H2-I2-J2
```

so we go to B instead of D. Not sure if that matters for further execution of the method and the logic behind it, I wrote it just to demonstrate how difficult is to think about this loop. So at the end we could arrive anywhere really and once we finish this loop, we don't have any guarantees that `fork` is actually in our chain. 

So I don't say there is a bug in this code, I just say that I'm very close to be unable to think about it clearly as there are too many ways how this can play out. This does not seem like the correct design, it is just too complex to think about (for me).

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -86,39 +86,49 @@ public virtual void ProcessBlock(Block block)
...
+                // The new tip can be ahead or behind the wallet
+                // If the new tip is ahead we try to bring the wallet up to the new tip.
+                // If the new tip is behind we just check the wallet and the tip ad in the same chain.
+            
+                if (newTip.Height > this.walletTip.Height)
                 {
+                    // check that the new tip is in the chain of the wallet tip
+                    var findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
+                    Guard.Assert(findTip == this.walletTip); // this should never happen
+
```
Aprogiena: As per my analysis above, this is wrong for two reasons:
1) FindAncestorOrSelf should probably never be used in such circumstances as it can take a lot of CPU looping through the whole chain down to genesis.
2) It can certainly be the case that `findTip != walletTip` exactly because `newTip` and `fork` can be on completely different chains. More specifically, consider that `newTip` is not on best chain and `walletTip` is. In that case, this triggers the assert.

Aprogiena: So to further explain, on line https://github.com/dangershony/StratisBitcoinFullNode/blob/64df43a3641428632552610dcfb68134b653511e/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs#L94 the chain changes to new chain -> `newTip` is not on the chain and the rest should be obvious.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: You have valid points, and the solution to this is making the wallet stand alone, what does that mean? the wallet not depending on consensus (like with store, consensus will feed blocks to the wallet but if the blocks are out of band wallet will handle that itself without stopping consensus by either asking from store or downloading form peers).

This will be the best approach.

> So at the end we could arrive anywhere really and once we finish this loop, we don't have any guarantees that fork is actually in our chain.
This is true for any given moment right? as the chain could in theory reorg at any point, best we can do is finish (fully/partially) our current reorg operation and next round try to fix it.

I think the most important is that the wallet can recover from any scenario, if we end up on the wrong chain after all this work, the next block will fix that and the wallet will recover without losing fund.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: 1. This will only happen on reorg so it should not happen that much.

>2. It can be the case that findTip != walletTip exactly because newTip and fork can be on different chains. 

Hmm good point (probably really rare but still valid).
I made the assumption that if we reorged then wallet and new tip are on the same chain.

Perhaps in this case the solution is to do nothing instead of Assert.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: Yes, at no given point you have any guarantees about the value in local variable representing something on the best chain.

I will try to present a concurrent chain replacement idea "soon".

But yes, you just need to be able to recover. This includes not crashing the node or loop you are in and possibly not executing too much complex logic with invalid values, which can be tricky.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: I have very hard time understanding why `FindAncestorOrSelf` is implemented in a way it is. This just feels wrong from theoretical perspective. It is O(N) function and it should be O(1) using dictionary. As such, FAOS should never be used in case there is a chance that the hash you are looking for is not on the same chain. It should also not be used if the hash could be far from the block you are starting with. By far I mean anything above 50 items. 

So assert is wrong, yes, but just by removing it, the code is still unwell.

Aprogiena: Let's say I'm not sure we want to fix this issue right now, or using the design like this that is very complex to think about.

dangershony: There is an O(1) method but its in ConcurrentChain and we can not use ConcurrentChain.
You would want to use `FindAncestorOrSelf ` (or a variation of it) so not to depend on concurrent chain.
You can think of it in the same way as your initial session suggestion where a reorg will not happen in that used session, The selected ChainedBlock is representing the chain at that point in time and its guaranteed to not reorg.

Perhaps we could improve the look up on `FindAncestorOrSelf ` .

dangershony: If `findTip` is not found we can replace the assert with doing nothing giving executing back and letting the next round try to recover the wallet, this is a very very rare case of ace condition.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: I will remind you that this will be solved much better if the wallet was a stand alone feature.
It would run in the background and only make one check once in a while that it's on the best chain (and if its not reorg and rebuild the wallet without depending on he consensus feeding blocks, either by downloading blocks or read from store)

Aprogiena: Yes, that is excellent point. As mentioned, maybe it is not worth fixing it now because of this upcoming change that will dramatically change the whole thing.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: Ah I see, this is to avoid using the chain ... In general, I would say FAOS is OK to use if you are absolutely sure that those two chained blocks are on the same chain (and not so far away, which is always true for minor reorgs). So can we possibly make that guarantee?

dangershony: We can't guarantee no, but I would say this would be a very rare case that 'findTip' is not in 'this.walltTip` for two reasons:
- Big reorgs are not so common.
- Earlier we checked that `findTip` is on the chain and we have passed the reorg code.

So both blocks by now should be on the chain, this check is to ensure that.
There would have multiple reorgs in the right moment for this to happen.
Also this is not happy path 99% of blocks will never go down this path.

Aprogiena: It is correct that this is rare and thus the inefficiency is not critical. I will try to reread the code as if the assert was replaced with IF and see if it makes any sense to me.

Aprogiena: OK, so I've read the code and this else block
```
                    // check that the new tip is in the chain of the wallet tip
                    var findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
                    Guard.Assert(findTip == newTip); // this should never happen
```
should probably be also treated similarly - no guard and don't process block 
btw `var`, vars everywhere in this code :-)

Then 
```
 next = newTip.GetAncestor(next.Height + 1);
```
This will be mostly OK, but looks ugly from theoretical point of view. I'd replace it with better approach. Do you know where I am going here?

Then yes, the big while trying to somehow find and wait for a block if it is not there that's crazy, but we know that it is subject for replacement so OK.

Overall yes, it "should" work with guards replaced and having in the mind all the previous comments.

dangershony: OK awesome I will fix the `var` stuff and remove the asserts and we can review again.

Aprogiena: yup

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -86,39 +86,49 @@ public virtual void ProcessBlock(Block block)
...
+                    // A reorg happened so bring the wallet back top the last known fork.
 
-                    var fork = walletTip;
+                    var fork = this.walletTip;
 
                     // We walk back the chained block object to find the fork.
                     while (this.chain.GetBlock(fork.HashBlock) == null)
                         fork = fork.Previous;
 
                     this.walletManager.RemoveBlocks(fork);
```
Neurosploit: What if anything goes wrong in the removeblocks method? We could have removed some uxto's in the walletmanager and failed halfway due to a concurrency issue. I still have to think of a scenario here but let's say it happened would a next block be able to repair it?

Neurosploit: Also would it make sense to create some tests for the mentioned scenario's discussed before my comment?

Neurosploit:
you can probably close my PR by the looks of it when we're done with 561 as I think that should take the failing test into account as well and make it pass because it's still a valid test case.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: If removeblocks fails that should throw an exception, remove blocks has a lock internally so it will not have race issues, the race condition is in regard to the chain being reorged (this method should (and will be synchronized once we push the fix to make consensus synchronised). 

Yes definitely need to write tests to cover each scenario.
Can you specify which is the failing test? 
I will follow your tests to create a test for each path.

Neurosploit: The failing test is mentioned in #570.

Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
```diff
@@ -1242,7 +1242,7 @@ public void GetSpendableTransactionsWithChainOfHeightZeroReturnsNoTransactions()
 
             walletManager.Wallets.Add(wallet);
 
-            var result = walletManager.GetSpendableTransactionsInWallet("myWallet", confirmations: 0);
+            var result = walletManager.GetSpendableTransactionsInWallet("myWallet", confirmations: 1);
```
Aprogiena: no appreciation of all these vars everywhere ... but i've understood that code and comment policy is not for tests (why oh why), so whatever :)

Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -614,16 +614,16 @@ public List<UnspentOutputReference> GetSpendableTransactions(int currentChainHei
                 // When calculating the confirmations the tip must be advanced by one.
 
                 var countFrom = currentChainHeight + 1;
-                var unspentTransactions = address.UnspentTransactions()
-                    .Where(a => countFrom - (a.BlockHeight ?? countFrom) >= confirmations).ToList();
+                var unspentTransactions = address.UnspentTransactions().Select(s =>
```
Aprogiena: we could incrementally improve the code style when we are touching old code ...
could we un-var this please?

Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -614,16 +614,16 @@ public List<UnspentOutputReference> GetSpendableTransactions(int currentChainHei
...
+                    new {Transaction = s, Confirmations = s.BlockHeight == null ? 0 : countFrom < s.BlockHeight ? 0 : countFrom - s.BlockHeight }).ToList();
```
Aprogiena: sorry, but no :)
this `a ? b : c ? d : e` 
i just can't, i don't really understand the change from before here, but it seems that LINQ is used here just to make it unreadable, how about we forget all this LINQ for a moment and go with good old foeach (which we have there anyway) even without those ternary operators, which, if nested, make my brain hurt

```
int countFrom = currentChainHeight + 1;
foreach (TransactionData transactionData in address.UnspentTransactions())
{
    int confirmationCount = 0;
    if (transactionData.BlockHeight != null) 
        confirmationCount = countFrom >= transactionData.BlockHeight ? transactionData.BlockHeight - countFrom : 0;

    if (confirmationCount >= confirmations)
    {
        unspentOutputs.Add(new UnspentOutputReference
        {
            Account = this,
            Address = address,
            Transaction = transactionData
        });
    }
}              
```

unless i've read it wrong, this should be equivalent and i think it is much more readable than LINQ-double-ternary mix

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -90,63 +90,75 @@ public void Stop()
             this.logger.LogTrace("(-)");
         }
 
+        /// <inheritdoc />
         public virtual void ProcessBlock(Block block)
         {
             Guard.NotNull(block, nameof(block));
             this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
 
+            var newTip = this.chain.GetBlock(block.GetHash());
```
Aprogiena: same as above, so i won't repeat the comments i made in light version ...

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -176,16 +191,26 @@ public virtual void ProcessBlock(Block block)
                         this.walletManager.ProcessBlock(nextblock, next);
                     }
                 }
-                else this.logger.LogTrace("New block's height {0} is not above wallet's tip height {1}.", incomingBlock.Height, this.walletTip.Height);
+                else
+                {
+                    ChainedBlock findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
```
Aprogiena: as in light version, not sure what to think about this else branch

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -90,63 +90,75 @@ public void Stop()
...
                         // TODO: This code should be replaced with a different approach
-                        // similar to BlockStore. The wallet should be standalone and not depend on consensus.
+                        // Similar to BlockStore the wallet should be standalone and not depend on consensus.
                         // The block should be put in a queue and pushed to the wallet in an async way.
                         // If the wallet is behind it will just read blocks from store (or download in case of a pruned node).
 
-                        next = this.chain.GetBlock(next.Height + 1);
+                        token.ThrowIfCancellationRequested();
+
+                        next = newTip.GetAncestor(next.Height + 1);
```
Aprogiena: as mentioned somewhere, i would prefer here to have a list of such `next` blocks instead of looking for one again and again

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -176,16 +191,26 @@ public virtual void ProcessBlock(Block block)
                         this.walletManager.ProcessBlock(nextblock, next);
                     }
                 }
```
Aprogiena: shouldn't we return here or not to process newTip block ? seems like it is processed twice then as we reach the end of the method, i think this is incorrect ...

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -110,43 +110,79 @@ public void Stop()
...
                         fork = fork.Previous;
 
-                    Guard.Assert(fork.HashBlock == block.Header.HashPrevBlock);
+                    this.logger.LogInformation("Reorg detected, going back from '{0}/{1}' to '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, fork.HashBlock, fork.Height);
+
                     this.walletManager.RemoveBlocks(fork);
-                    this.logger.LogWarning($"Reorg detected, wallet tip reverted back to Height = {fork.Height} hash = {fork.HashBlock}.");
+                    this.walletTip = fork;
+
+                    this.logger.LogTrace("Wallet tip set to '{0}'.", this.walletTip);
```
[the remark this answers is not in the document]
dangershony: Perhaps we should just change the ToString() on ChainedBlock to use `/` instead of `-`.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -110,43 +110,79 @@ public void Stop()
...
+                }
+
+                // The new tip can be ahead or behind the wallet
+                // If the new tip is ahead we try to bring the wallet up to the new tip.
+                // If the new tip is behind we just check the wallet and the tip are in the same chain.
+
```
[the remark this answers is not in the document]
dangershony: For me chunks of text describing the high level of the following code is more readable if there is an empty line at the end.
If its describing the immediate code yes I would not put an empty line.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: oh yes, ignore my comment, that should be enough

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: okie

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: maybe i just do a very quick PR changing that and making all such logs easier, what do you think?

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -110,43 +110,79 @@ public void Stop()
...
+                if (newTip.Height > this.walletTip.Height)
+                {
+                    ChainedBlock findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
```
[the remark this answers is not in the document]
dangershony: That's a great idea actually, this can also stop at the last checkpoint (we will need to put the checkpoint on the network class for that.

I will add an issue to do that.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -110,43 +110,79 @@ public void Stop()
...
+                    {
+                        this.logger.LogTrace("(-)[NEW_TIP_AHEAD_NOT_IN_WALLET]");
+                        return;
+                    }
+
+                    this.logger.LogTrace("Wallet tip '{0}/{1}' is behind the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
+
+                    // the wallet is falling behind we need to catch up
+                    this.logger.LogWarning("New tip '{0}/{1}' is too far in advance, put the puller back.", newTip.HashBlock, newTip.HashBlock);
+                    this.blockNotification.SyncFrom(this.walletTip.HashBlock);
```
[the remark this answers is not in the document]
dangershony: Off topic indeed we can look at it as part of the bigger task of finding the ConcurrentChain race conditions.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -110,43 +110,79 @@ public void Stop()
...
+                    ChainedBlock findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
+                    if (findTip == null)
                     {
-                        // the wallet is falling behind we need to catch up
-                        this.logger.LogWarning($"block received with height: {inBestChain.Height} and hash: {block.Header.GetHash()} is too far in advance. put the puller back.");
-                        this.blockNotification.SyncFrom(this.walletTip.HashBlock);
+                        this.logger.LogTrace("(-)[NEW_TIP_BEHIND_NOT_IN_WALLET]");
                         return;
                     }
+                    this.logger.LogTrace("Wallet tip '{0}/{1}' is ahead or equal to the new tip '{2}/{3}'.", this.walletTip.HashBlock, this.walletTip.Height, newTip.HashBlock, newTip.HashBlock);
```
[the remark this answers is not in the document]
dangershony: This can happen rarely (if the node crashed and consensus did not save its tip) it can also happen in the light wallet if you set the puller back in time to pull blocks again.

The ProcessesBlock is idempotent so no problem to processes a block again.

Stratis.Bitcoin.Features.Wallet/Wallet.cs
dangershony: I don't think its that hard to understand, though I agree too much linq can be tedious.
But as you went to the trouble of rewriting the method (and it is more readable) I will take your snippet.

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: It wont matter much as the store can anyway just take one block at the time.
Also this whole code might change later so I wont bother with this refactor right now.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
dangershony: Sure yeah sounds good

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
dangershony: Actually we want to processes again, in case the puller was put back.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: I can take your word on it, I am not familiar with PB. It sounds dangerous on the first look, but if you know it is (or should be) safe, then OK, although I'd be slightly nervous about it.

Stratis.Bitcoin.Features.Wallet/Wallet.cs
Aprogiena: there were not even brackets, my brain was bleeding ;-)

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: it won't, it just feels bad from theoretical point; you won't have here very long differences in height between newTip and next, so O(N^2) complexity should be OK, but from theoretical point, this is just ugly

but exactly as you say - no need to do anything about it since this will be changed (soon)

so we are good with this, just a general note

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: cool, will do #591

Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: i can again take your word on that, but are you sure in this case? like double-sure ? :-)

i mean, i was quite sure this is a bug, then you argued that double processing should be handled well, so i would change my classification from bug to non-optimal code, but you are saying here that you actually want to process it again, my mind is blown 

can you explain that? why would we want/need processing the same block again? i can't see it, tx

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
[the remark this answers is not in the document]
Aprogiena: oh, actually that is ChainedBlock in NBitcoin ... so i can do the log part in the node

PR #564, opened 2017-10-10
Implement transaction input and output shuffling for increased privacy

nopara73:
https://github.com/nopara73/ZeroLink/#transaction-output-indexing

#### Transaction Output Indexing
|Basic Post-Mix Wallet Requirement|Post-Mix Wallet Uniformity Requirement|
|---------------------------------|--------------------------------------|
||Post-mix wallet SHOULD index its built transaction outputs randomly.|

A post-mix wallet, due to its design, will only have one input and a maximum of two outputs at all times. Uniform indexing of outputs is necessary in order for multiple post-mix wallet implementations to look the same. A post-mix wallet SHOULD use random indexing of outputs.

Random indexing is not exclusively beneficial for post-mix wallet uniformity, conversely it has another privacy benefit. When a wallet software always generates the change output on the second index, observers always know which output is the change.

It must be mentioned [BIP69](https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki), Lexicographical Indexing of Outputs was created for the same purpose, however random indexing is slightly more private. If a blockchain observer wants to know if a transaction is in a wallet, using the BIP is a track, because it uses a deterministic algorithm, while random indexing leaves no tracks.

### DIFF
--- a/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
+++ b/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
@@ -495,11 +495,12 @@ namespace Stratis.Bitcoin.Features.Wallet.Controllers
             {
                 var context = new TransactionBuildContext(
                     new WalletAccountReference(request.WalletName, request.AccountName),
-                    new[] {new Recipient {Amount = request.Amount, ScriptPubKey = destination}}.ToList(),
+                    new[] { new Recipient { Amount = request.Amount, ScriptPubKey = destination } }.ToList(),
                     request.Password)
                 {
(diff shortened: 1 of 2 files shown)


bokobza:
@dangershony @nopara73 
I think we shouldn't have the option to shuffle, it should always shuffle.
So in my opinion, you can remove the boolean Shuffle from the context and just do
`context.TransactionBuilder.Shuffle();`.

Also, this is not what I would call critical to Breeze.
What if the Shuffle method in Nbitcoin doesn't work properly? Maybe it does, but now is not the time to introduce some uncertainties.

Issue #565, opened 2017-10-10
Consider refactoring based on "Async/Await - Best Practices in Asynchronous Programming"


nopara73:
https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

dangershony:
That's a good idea we should try to start removing the usages of GetAwaiter().Getresult()
It should be done in small commits, all our loops are already async supported so adding this should be fairly easy, However it might mean changing method signatures which will make this task a bit longer.

nopara73:
Do you think it is a better practice to use [async unit tests](http://bradwilson.typepad.com/blog/2012/01/xunit19.html)?

dangershony:
It depends on the unit test, some of our unit tests are async already

Aprogiena:
Yes, this is known issue. I am very much towards changing it, my first attempt should be #553

PR #568, opened 2017-10-11
Replace .Wait(), .Result, .GetAwaiter().GetResult() with await .Confi…

nopara73:
…gureAwait(false) in Test projects

As a starting point of resolving the issue: https://github.com/stratisproject/StratisBitcoinFullNode/issues/565

### DIFF
--- a/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
+++ b/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
@@ -5,15 +5,16 @@
     using Stratis.Bitcoin.Tests;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Threading.Tasks;
     using Xunit;
 
     public class BlockRepositoryTest : TestBase
     {
         [Fact]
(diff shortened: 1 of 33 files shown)


dangershony:
@Neurosploit can you help review this as well?

dangershony:
@nopara73 we normally try to work with smaller commits so its easier to review.

Neurosploit:
Can you first clarify why configureawait(false) is needed at all? See http://blog.stephencleary.com/2017/03/aspnetcore-synchronization-context.html

dangershony:
Ha they dropped the `AspNetSynchronizationContext` wowoo so the aspnet core context is now multi threaded by default.

We can remove that in tests but in libraries (the fullnode) we probably keep it.

nopara73:
@Neurosploit What do you propose?  
ConfigureAwait(false) in Stratis.Bitcoin, NStratis and Features projects  
and not ConfigureAwait(false) in Tests, Api, Cli and Daemon projects?  

@dangershony  What do you think?

Neurosploit:
Sounds good @nopara73.

PR #570, opened 2017-10-11
Reapply walletsyncmanager catchup fix

Neurosploit:
If inbestchain == null in the if statement before we need to check with the wallettip that was set back to block n instead of the current walletsyncmanager wallet tip. Otherwise we won't retrieve the missing blocks and end up with a gap.

The related test ProcessBlock_NewBlock_PreviousHashNotSameAsWalletTip_WalletTipNotOnBestChain_RemovesBlocksAfterForkOnOldChainFromWalletManager_CatchUpWalletManagerUsingBlockStoreCache proves this:
1. wallettip set to block nr 4 of current chain
2. process block nr 5 on other chain
3. rewind to block 2 where fork was caused because inBestChain == null
4. verify block 3,4,5 of the other chain are then processed by the underlying walletmanager


dangershony:
Did we not apply that already? hmm.. odd

dangershony:
But this will become redundant once #561 goes in

Neurosploit:
Yes I agree with that. This can then be closed.

PR #571, opened 2017-10-11
Wallet balance correctly calculated 

dangershony:


dangershony:
@Aprogiena this I hope will fix the incorrect balance.

Stratis.Bitcoin.Features.Wallet/WalletFeature.cs
```diff
@@ -74,6 +74,24 @@ public void AddNodeStats(StringBuilder benchLogs)
...
+        public void AddFeatureStats(StringBuilder benchLog)
+        {
+            var walletNames = this.walletManager.GetWalletsNames();
+
+            if (walletNames.Any())
+            {
+                benchLog.AppendLine();
+                benchLog.AppendLine("======Wallets======");
+
+                foreach (var walletName in walletNames)
```
Aprogiena: you are still in war with var, here and down there and up there, everywhere ;)

Aprogiena:
I went through the code and I'd just try to improve the code style and comment style everywhere (incrementally) as well as not to introduce new violations of the policies. 

But this will not fix the problem. I am quite certain about it because I've already locally checked for all transactions with 0 confs and the coins are not there in the wallet. But maybe I'm missing some internal connection and it will work, so I'd say let's test it. So if you could just little improve on the code style and otherwise it is fine and I'll test it then. It seems that as I removed all those UTXOs from my wallet, I should be able again to reproduce these kinds of things.

Stratis.Bitcoin.Features.Wallet/WalletFeature.cs
dangershony: This is just a string so I would say var is fine.

Aprogiena: How do you know it is a string if you see `var`?

PR #576, opened 2017-10-12
Refactor/issue 553 async pos mining

Aprogiena:
Improvements
- network weight is only calculated once, not twice
- removed GetAwaiters and converted whole async loop to be really async
- added parallelism to kernel finding - multiple tasks are created, each processing 25 UTXOs, default scheduler will use multiple threads to make processing parallel

### DIFF
--- a/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -46,6 +46,76 @@ namespace Stratis.Bitcoin.Features.Miner
             public string WalletName;
         }
 
+        /// <summary>
+        /// Information needed by the coinstake worker for finding the kernel.
+        /// </summary>
+        public class CoinstakeWorkerContext
+        {
+            /// <summary>Worker's ID / index number.</summary>
+            public int Index { get; set; }
+
+            /// <summary>Logger with worker's prefix.</summary>
+            public ILogger Logger { get; set; }
+
+            /// <summary>List of UTXO descriptions that the worker should check.</summary>
+            public List<StakeTx> Coins { get; set; }
+
+            /// <summary>Information related to coinstake transaction.</summary>
+            public CoinstakeContext CoinstakeContext { get; set; }
+
+            /// <summary>Result shared by all workers. A structure that determines the kernel founder and the kernel UTXO that satisfies the target difficulty.</summary>
+            public CoinstakeWorkerResult Result { get; set; }
+        }
+
+        /// <summary>
+        /// Result of a task of coinstake worker that looks for kernel.
+        /// </summary>
+        public class CoinstakeWorkerResult
+        {
+            /// <summary>Invalid worker index as a sign that kernel was not found.</summary>
+            public const int KernelNotFound = -1;
+
+            /// <summary>Index of the worker that found the index, or <see cref="KernelNotFound"/> if no one found the kernel (yet).</summary>
+            private int kernelFoundIndex;
+
+            /// <summary>Index of the worker that found the index, or <see cref="KernelNotFound"/> if no one found the kernel (yet).</summary>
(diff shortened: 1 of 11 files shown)


Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -190,13 +275,13 @@ public IAsyncLoop Mine(WalletSecret walletSecret)
...
 
-            this.mining = this.asyncLoopFactory.Run("PosMining.Mine", token =>
+            this.mining = this.asyncLoopFactory.Run("PosMining.Mine", async token =>
             {
                 this.logger.LogTrace("()");
 
                 try
                 {
-                    this.GenerateBlocks(walletSecret);
+                    await this.GenerateBlocks(walletSecret);
```
nopara73: We decided on to ConfigureAwait(false) in libraries (Stratis.Bitcoin, NStratis and Features) and not ConfigureAwait(false) in in Tests, Api, Cli and Daemon projects.  
https://github.com/stratisproject/StratisBitcoinFullNode/pull/568#issuecomment-336212110

Aprogiena: k, tx, done

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -549,67 +633,185 @@ public bool CreateCoinStake(List<StakeTx> stakeTxes, Block block, ChainedBlock c
...
+        /// This allows the kernel finding task to be processed on multiple processors in parallel.
+        /// </para>
+        /// </summary>
+        /// <param name="context">Context information with worker task description. Results of the worker's attempt are also stored in this context.</param>
+        /// <param name="chainTip">Tip of the best chain. Used only to stop working as soon as the chain advances.</param>
+        /// <param name="block">Template of the block that we are trying to mine.</param>
+        /// <param name="minimalAllowedTime">Minimal valid timestamp for new coinstake transaction.</param>
+        /// <param name="searchInterval">Length of an unexplored block time space in seconds. It only makes sense to look for a solution within this interval.</param>
+        /// <returns></returns>
+        private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainTip, Block block, long minimalAllowedTime, long searchInterval)
```
dangershony: This method is a work of art very nice!

dangershony:
I will wait til @dev0tion updates the submodule then we can commit

dangershony:
Actually I will commit as this does not effect any of the Breeze light wallet code.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -70,8 +141,15 @@ public class WalletSecret
         /// multiplied by this value. See <see cref="GetSplitStake(int)"/>.</summary>
         public const int CoinstakeSplitLimitMultiplier = 3;
 
+        /// <summary>Number of UTXOs that a single worker's task will process.</summary>
+        /// <remarks>To achieve a good level of parallelism, this should be low enough so that CPU threads are used,
+        /// but high enough to compensate for tasks' overhead.</remarks>
+        private const int CoinsPerCoinstakeWorker = 25;
```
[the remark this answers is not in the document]
Aprogiena: 3x coin age limit gives us max of 150, but note that a user of a wallet can get much more UTXO by just using the wallet normally. So in fact, there is no limit and one can have 1000 UTXOs "easy". That is where this parallelism would really help. But even with 3x 50 = 150 UTXOs just from coinstake splitting, there will be 150/25 = 6 workers

Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: Cheers, deeply appreciated.

PR #602, opened 2017-10-16
Cleanup redundant code

nopara73:


bokobza:
Good stuff. I'll wait for after the release to push this in, as DBreeze is used even in the Breeze Wallet for the chain.

PR #603, opened 2017-10-16
Unit tests for LightWalletSyncManager

Neurosploit:


bokobza:
This is really good stuff @Neurosploit, thanks for doing that.
I'll have a look to see if we want to include this now so that it'll be in the release next week. 
Is there anything critical in there? I don't want to introduce new code that's not fixing something critical.

bokobza:
Actually let me have a look, there doesn't seem to be any functionality changes.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -220,10 +264,15 @@ public void SyncFromDate(DateTime date)
         /// <inheritdoc />
         public void SyncFromHeight(int height)
         {
+            if (height < 0)
+            {
+                throw new WalletException("Invalid block height");
```
bokobza: Can we have some logging here to tell us what happened?
Also please add the height in the exception.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -197,7 +241,7 @@ public void SyncFromDate(DateTime date)
             // before we start syncing we need to make sure that the chain is at a certain level.
             // if the chain is behind the date from which we want to sync, we wait for it to catch up, and then we start syncing.
             // if the chain is already past the date we want to sync from, we don't wait, even though the chain might not be fully downloaded.
-            if (this.chain.Tip.Header.BlockTime.LocalDateTime < date)
+            if (this.chain.Tip != null && this.chain.Tip.Header.BlockTime.LocalDateTime < date)
```
bokobza: Can the tip be null? what shall we do if that's the case? @dangershony

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -220,10 +264,15 @@ public void SyncFromDate(DateTime date)
...
+            if (height < 0)
+            {
+                throw new WalletException("Invalid block height");
+            }
+
             // before we start syncing we need to make sure that the chain is at a certain level.
             // if the chain is behind the height from which we want to sync, we wait for it to catch up, and then we start syncing.
             // if the chain is already past the height we want to sync from, we don't wait, even though the chain might  not be fully downloaded.
-            if (this.chain.Tip.Height < height)
+            if (this.chain.Tip != null && this.chain.Tip.Height < height)
```
bokobza: Same comment as above.

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -87,24 +102,53 @@ public void Start()
...
                 // we start by looking at the heights of the wallets and we start syncing from the oldest one (the smallest height).
                 // if for some reason we can't find a height, we look at the creation date of the wallets and we start syncing from the earliest date.
-                int? earliestWalletHeight = this.walletManager.Wallets.Min(w => w.AccountsRoot.Single(a => a.CoinType == this.coinType).LastBlockSyncedHeight);
+                int? earliestWalletHeight = this.walletManager.GetEarliestWalletHeight();
                 if (earliestWalletHeight == null)
                 {
-                    DateTimeOffset oldestWalletDate = this.walletManager.Wallets.Min(w => w.CreationTime);
+                    DateTimeOffset oldestWalletDate = this.walletManager.GetOldestWalletCreationTime();
+
+                    if (fork != null && oldestWalletDate > fork.Header.BlockTime)
```
bokobza: I need @dangershony or @Aprogiena to have a look at this and the change just below.

dangershony: I am just wondering why do we need a new parameter called `fork`, using the `this.walletTip` should yield the same result right?

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
dangershony: The tip can not be null, it will default to genies (height zero)

Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
Aprogiena: I would leave this to @dangershony as I'm not that familiar with this part of code base. It would take me quite a long time to learn it. But if you really need me, ping me again.

bokobza:
This is excellent, thank you.

PR #604, opened 2017-10-17
Feature/issue 401

### LINKED ISSUE #401: [POS] 3 or more peers are required for mining, but it is not enforced

Aprogiena:
Removed code that delayed staking. Rationale for that provided here: https://github.com/stratisproject/StratisBitcoinFullNode/issues/401#issuecomment-337182651

In short - there is no penalty for staker and so the incentive is to start to stake as soon as possible.

Fix #401

### DIFF
--- a/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -335,20 +335,19 @@ namespace Stratis.Bitcoin.Features.Miner
 
             while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
             {
-                ChainedBlock chainTip = this.chain.Tip;
-                if (chainTip != this.consensusLoop.Tip)
-                {
-                    this.logger.LogTrace("(-)[SYNC_OR_REORG]");
-                    return;
-                }
-
                 while (!this.connection.ConnectedNodes.Any() || this.chainState.IsInitialBlockDownload)
                 {
                     if (!this.connection.ConnectedNodes.Any()) this.logger.LogTrace("Waiting to be connected with at least one network peer...");
                     else this.logger.LogTrace("Waiting for IBD to complete...");
 
                     await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
-                    continue;
+                }
+
+                ChainedBlock chainTip = this.chain.Tip;
+                if (chainTip != this.consensusLoop.Tip)
+                {
+                    this.logger.LogTrace("(-)[SYNC_OR_REORG]");
+                    return;
                 }
 
                 if (this.lastCoinStakeSearchPrevBlockHash != chainTip.HashBlock)
--- a/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -356,6 +356,9 @@ namespace Stratis.Bitcoin.Features.Miner
                     this.lastCoinStakeSearchPrevBlockHash = chainTip.HashBlock;
                     this.lastCoinStakeSearchTime = chainTip.Header.Time;
                     this.logger.LogTrace("New block '{0}' detected, setting last search time to its timestamp {1}.", chainTip, chainTip.Header.Time);
+
+                    // Reset the template as the chain advanced.
+                    blockTemplate = null;
                 }
 
                 uint coinstakeTimestamp = (uint)this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp() & ~PosConsensusValidator.StakeTimestampMask;
--- a/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -366,11 +366,6 @@ namespace Stratis.Bitcoin.Features.Miner
                     return;
                 }
 
-                if (blockTemplate == null)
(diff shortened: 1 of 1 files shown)


Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -426,6 +397,11 @@ public IAsyncLoop Mine(WalletSecret walletSecret)
 
                 this.logger.LogTrace("Wallet contains {0} coins.", new Money(totalBalance));
 
+                if (blockTemplate == null)
```
bokobza: Moving code closer to where it is needed. I like it.

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -348,31 +347,8 @@ public IAsyncLoop Mine(WalletSecret walletSecret)
...
-                    bool lastBlockTooOld = chainTip.Header.Time < (this.dateTimeProvider.GetTime() - 10 * 60);
-                    if ((fewPeers && !this.network.IsTest()) || lastBlockTooOld)
-                    {
-                        if (fewPeers) this.logger.LogTrace("Node is connected to few peers.");
-                        if (lastBlockTooOld) this.logger.LogTrace("Last block is too old, timestamp {0}.", chainTip.Header.Time);
-
-                        await Task.Delay(TimeSpan.FromMilliseconds(60000), this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
-                        continue;
-                    }
+                    continue;
```
bokobza: What's the purpose of this continue here? Don't you want to loop until we're connected and not in IBD?

Aprogiena: Sorry, this is wrong. Will fix that.

Aprogiena: So, now it should be better - the continue obviously has no purpose there, but what I intended to do is to continue with the external loop, not internal. I've fixed that by reordering the code. So now when we exit the loop (after we waited for IBD to complete) we check if consensus is aligned with the tip.

bokobza:
I'm ok with the change, it all makes sense to me.
However, @dangershony do you want to say what you think about the removal of the tryToSync piece?

Aprogiena:
Thanks, yes, it is good idea to check with @dangershony as he originally suggested different approach. But I hope my latest explanation will satisfy him :-)

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -332,54 +332,32 @@ public IAsyncLoop Mine(WalletSecret walletSecret)
...
-                    tryToSync = true;
                     await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
                 }
 
-                // TODO: What is the purpose of this conditional block?
-                // It seems that if it ends with continue, the next round of the loop will just 
-                // not execute the above while loop and it won't even enter this block again,
-                // so it seems like no operation.
-                // In StratisX there is the wait uncommented. Then it makes sense. Why we have it commented?
-                if (tryToSync)
```
dangershony: Why remove this code?
This will wait for at least 3 peers  before staking starts (I guess it so we have a better time sync sample)

dangershony: OK added a comment, I don't see much value in waiting for 3 nodes (apart for the time sync) but then again there is no downside (apart for waiting a bit longer to stake) in connecting to some more nodes.

So I am unsure about this.

Aprogiena: Any code we have in the code base should be doing something useful. We can wait longer, but what is the benefit of it? If we found a block instead and sent it to the network, we could earn some money. Instead we are waiting. The reason we are waiting is probably to get a better chance for our block to be accepted. But at the same time, if we are not looking for it, our chance of getting reward 0.

Time syncing is another topic I think and just by connected to 3 nodes it won't trigger. The threshold is little higher than that (and we also intend to have an option to have it disabled). However, the worst case scenario is that we have wrong time and we mine a new block with wrong time stamp. In worst case it would be rejected. 

To summarize, I can't see enough justification for this code, so I removed it.

bokobza: I'm happy with this.

dangershony: hmf... ok

Issue #229, opened 2017-07-13
Longest chain timeout


dangershony:
When a peer that connected with a longer chain got disconnected before our node downloaded the blocks we need to disconnect and reset the longest chain to one of the connected peers

dangershony:
![](https://github.trello.services/images/mini-trello-icon.png) [[Consensus] Longest chain timeout](https://trello.com/c/AvvS5n6l/117-consensus-longest-chain-timeout)

Aprogiena:
This issue applies to both - Bitcoin and Stratis,

but in case of Stratis, there seems to be an attack scenario for POS where one is able to assemble a chain of block headers (but not blocks!) and present them as a longer chain, the node will probably try to get that chain, which will result in either the peer providing invalid block (he does not have those blocks), or he will refuse to provide them (won't reply). What we need to do here is that we make sure such node is disconnected (immediately [and banned] if provided invalid block; or after some timeout if it does not respond) and then we need to resolve as @dangershony mentioned above - we need to reset our longest chain to the best chain of currently connected peers (and make sure to handle a situation where we don't have any connected peers when that happens, so switch to the best chain of any new peer that connects later).

bokobza:
![](https://github.trello.services/images/mini-trello-icon.png) [[Chain] Longest chain timeout](https://trello.com/c/Sj2MTv69/316-chain-longest-chain-timeout)

Aprogiena:
So what was the approach we wanted to take here? Have we agreed on this going to ChainHeadersBehavior at the end or not?

Aprogiena:
I've forgotten the resolution, but I can remember now that we've mentioned that we need to access all nodes, which is not easy to do in CHB. I think I've suggested putting list of nodes into ChainState. But don't know if we made further progress on the discussion and resolved it somehow. 

So what is the design here?

dangershony:
We need to first identify if this is a consensus problem or node problem, and also how to approach the fix:
1. on disconnect, default to the next longest chain form one of the peers.
2. when asking for a block and the block does not arrive or its invalid.   
There is much more complexity with option 2.
 
This is probably a node problem (and not only consensus) while its easier to determine false headers when we have a consensus mechanism in place, this issue can also effect nodes that are not running the consensus verification code. (a light wallet that was given a fake longer chain will not swap back to the correct chain until its longer, or until the blocks are asked).

In my view the approach to fix this would be on a disconnect of a peer the following should happen.
[...]

PR #609, opened 2017-10-17
Making consensus sequential

### LINKED ISSUE #488: [WalletManager] block too far in the future has arrived to the wallet
Mined on testnet with StratisD, when suddenly Consensus height stopped to increase and Headers went far in front of it. Wallet height was stopped 2 blocks behind:

```
Headers.Height:        122618   Headers.Hash:         d517459ca0a7d8eed6adff1e684aadf844e589cf51861b8b95c53e7b2b3f648d
Consensus.Height:      122586   Consensus.Hash:       4836bc89a3f6be0248f07d5531f811363eb0684748f4c8d3273ffc1f4624d647
BlockStore.Height:     122586   BlockStore.Hash:      4836bc89a3f6be0248f07d5531f811363eb0684748f4c8d3273ffc1f4624d647
Wallet.Height:         122584   Wallet.Hash:          d6efc1f9a8b0c06dbba117fe8025f3ca26a5197c0840cd2071f7862726305c17
```

And in logs I can see the reason:

```
[2017-09-23 11:19:29.4024 12] TRACE: Stratis.Bitcoin.Features.Wallet.WalletManager.ProcessBlock block notification - height: 122586, hash: 4836bc89a3f6be0248f07d5531f811363eb0684748f4c8d3273ffc1f4624d647, coin: Stratis
[2017-09-23 11:19:29.4189 12] DEBUG: Stratis.Bitcoin.Features.Consensus.ConsensusFeature.RunLoop Exception occurred in consensus loop: Stratis.Bitcoin.Features.Wallet.WalletException: block too far in the future has arrived to the wallet
   at Stratis.Bitcoin.Features.Wallet.WalletManager.ProcessBlock(Block block, ChainedBlock chainedBlock) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\WalletManager.cs:line 508
   at Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\WalletSyncManager.cs:line 143
   at Stratis.Bitcoin.Features.Wallet.Notifications.BlockObserver.OnNextCore(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\Notifications\BlockObserver.cs:line 24
   at System.Reactive.SafeObserver`1.OnNext(TSource value)
   at System.Reactive.Observer`1.OnNext(T value)
   at System.Reactive.SynchronizedObserver`1.OnNextCore(T value)
   at Stratis.Bitcoin.Signals.Signaler`1.Broadcast(T item) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Signals\Signaler.cs:line 70
   at Stratis.Bitcoin.Signals.Signals.SignalBlock(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Signals\Signals.cs:line 74
   at Stratis.Bitcoin.Features.Consensus.ConsensusFeature.RunLoop(CancellationToken cancellationToken) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Consensus\ConsensusFeature.cs:line 179
[2017-09-23 11:19:29.4189 12] FATAL: Stratis.Bitcoin.Features.Consensus.ConsensusFeature.RunLoop Consensus loop unhandled exception (Tip:122586)
[2017-09-23 11:19:29.4354 12] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0.<StartAsync>b__0 Consensus Loop stopping
[2017-09-23 11:19:29.4354 12] FATAL: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0.<StartAsync>b__0 Consensus Loop threw an unhandled exception
[2017-09-23 11:19:29.4354 12] DEBUG: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0.<StartAsync>b__0 Consensus Loop threw an unhandled exception: Stratis.Bitcoin.Features.Wallet.WalletException: block too far in the future has arrived to the wallet
   at Stratis.Bitcoin.Features.Wallet.WalletManager.ProcessBlock(Block block, ChainedBlock chainedBlock) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\WalletManager.cs:line 508
   at Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\WalletSyncManager.cs:line 143
   at Stratis.Bitcoin.Features.Wallet.Notifications.BlockObserver.OnNextCore(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Wallet\Notifications\BlockObserver.cs:line 24
   at System.Reactive.SafeObserver`1.OnNext(TSource value)
   at System.Reactive.Observer`1.OnNext(T value)
   at System.Reactive.SynchronizedObserver`1.OnNextCore(T value)
   at Stratis.Bitcoin.Signals.Signaler`1.Broadcast(T item) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Signals\Signaler.cs:line 70
   at Stratis.Bitcoin.Signals.Signals.SignalBlock(Block block) in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Signals\Signals.cs:line 74
   at Stratis.Bitcoin.Features.Consensus.ConsensusFeature.RunLoop(CancellationToken cancellationToken) in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Consensus\ConsensusFeature.cs:line 200
   at Stratis.Bitcoin.Features.Consensus.ConsensusFeature.<<Start>b__22_0>d.MoveNext() in C:\StratisBitcoinFullNode\Stratis.Bitcoin.Features.Consensus\ConsensusFeature.cs:line 114
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Stratis.Bitcoin.Utilities.AsyncLoop.<>c__DisplayClass12_0.<<StartAsync>b__0>d.MoveNext() in C:\StratisBitcoinFullNode\Stratis.Bitcoin\Utilities\AsyncLoop.cs:line 122
[2017-09-23 11:19:29.7310 9] TRACE: Stratis.Bitcoin.Features.BlockStore.BlockStoreSignaled.<RelayWorker>b__14_0 ()
[2017-09-23 11:19:29.7310 9] TRACE: Stratis.Bitcoin.Features.BlockStore.BlockRepository.ExistAsync (hash:'4836bc89a3f6be0248f07d5531f811363eb0684748f4c8d3273ffc1f4624d647')
[2017-09-23 11:19:29.7310 9] TRACE: Stratis.Bitcoin.Features.BlockStore.BlockRepository.ExistAsync (-)
```

Exception in consensus loop simply stopped consensus to go futher, so no new block were accepted.

dangershony:
This is a version of the code to fix #488 where access to the consensus validation rules is synchronised.

This code uses a plain lock, however a lock will not work with an async approach.
The two option are to sue a semaphore async lock or break the consensus logic to two different async thread loops one for the puller and a second for the rules validation (and use a BlockingCollection) to synchronise access.

I will be exploring the two options.


bokobza:
Would you mind resolving the conflicts first?

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -226,26 +238,42 @@ public void AcceptBlock(BlockItem item)
...
                 if (this.Chain.Tip.HashBlock == item.ChainedBlock?.HashBlock)
                     this.FlushAsync().GetAwaiter().GetResult();
 
+                if (this.Tip.ChainWork > this.Chain.Tip.ChainWork)
+                {
+                    // This is a newly mined block.
+                    this.Puller.SetLocation(this.Tip);
+                    this.Chain.SetTip(this.Tip);
+
+                    this.logger.LogTrace("Block extends consensus tip to ");
```
bokobza: Comment seems incomplete.

Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -1,78 +1,81 @@
...
 
+        /// <summary>
+        /// This method will check if the node is in a state of IBD (Initial Block Download)
+        /// </summary>
         public bool IsInitialBlockDownload
         {
             get
             {
-                if (this.lastUpdate < this.fullNode.DateTimeProvider.GetUtcNow().Ticks)
+                if (this.ibdlastUpdate < this.dateTimeProvider?.GetUtcNow().Ticks)
```
bokobza: change to ibdLastUpdate and ibdLastResult

bokobza:
I don't have much to say on this, just that I'd rather wait to push it in before Breeze is released as some of the classes might affect its execution.
Also this will give time for @Aprogiena to have a look.

Aprogiena:
This is huge. Please allow a lot of time, but I certainly want to review this.

Aprogiena:
What is the rationale behind removing `CacheSettings.cs`?

Aprogiena:
Changes in `ChainHeadersBehavior.cs` will cause conflicts with my PR #492. I believe I have untabified the file there as well. I might only not reordered the using section.

dangershony:
>Changes in ChainHeadersBehavior.cs will cause conflicts with my PR #492.
Ok I can undo my untabifying no problem.

`CacheSettings` was never really used, probably left over from early days (and your change even obsoletes this further).

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
```diff
@@ -100,5 +92,12 @@ public void Log()
             benchLogs.AppendLine(this.connectionManager.GetStats());
             this.logger.LogInformation(benchLogs.ToString());
         }
+
+        protected override void OnNextCore(Block value)
```
Aprogiena: This is interesting. Not sure how I feel about it (btw inheritdoc comment missing i guess). So if i understand this correctly, instead of having there a loop, i.e. something like a timer that would just put these stats on the console, you are now using this signaller so that you are executing within the context of the consensus loop thread. Is it good for anything? 

I mean, I understand that this prevents concurrency problems if there were any (well there were, but with chain, not sure if with consensus loop), but the nature of this thing sounds to me more like the timer (or async loop). 

So now this methods when it gets executed delays consensus from doing its job. Is that a good thing?  I know there is a parallel with my coinview flushing approach, but in that code there really was a concurrency problem with consensus, so although I was not really happy about that I did like this there. 

Doing it here. Well, the performance hit is very very small, so I would OK. Just not sure I like it from the design point of view. 

Not a big deal anyway, just wondered what is the rationale here, if I'm missing something or not.

Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -155,25 +155,28 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
...
+                    {
+                        if (blockResult.Error == ConsensusErrors.InvalidPrevTip)
+                            continue;
+
+                        this.logger.LogTrace("(-)[ACCEPT_BLOCK_ERROR]");
                         return blocks;
+                    }
 
                     // Push the block to disk, so it is available when peers ask for it.
                     this.blockRepository.PutAsync(blockResult.ChainedBlock.HashBlock, new List<Block> { pblock }).GetAwaiter().GetResult();
```
Aprogiena: chmm, in both cases - POS and POW, I can see you have removed a lot of that stuff that cares about having a block being inserted into the system and you only left AcceptBlock - which is doing all that stuff; but I can see you have also left this PutAsync of blockRepository. Is this something that we don't want to be part of AcceptBlock?

Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -1,78 +1,81 @@
...
 
 namespace Stratis.Bitcoin.Base
 {
     public class ChainState
     {
-        private readonly FullNode fullNode;
+        /// <summary>The fullnode interface.</summary>
+        private readonly IFullNode fullNode;
+
+        /// <summary>The last time the <see cref="ibdlastResult"/> was updated.</summary>
```
Aprogiena: ibdLastResult and ibdLastUpdate with big L

Stratis.Bitcoin.Features.Miner/PowMining.cs
dangershony: Yeah PutAsync is part of the store feature (we actually don't have access to it in consensus) the Signal will do that anyway but just to be sure I left the code at the miner, though this can be remove as well.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -23,25 +27,84 @@ public class ConsensusLoop
...
+        /// <summary>Connection manager of all the currently connected peers.</summary>
+        private readonly IConnectionManager connectionManager;
+
+        /// <summary>A signaler that used to signal messages between features.</summary>
+        private readonly Signals.Signals signals;
+
+        /// <summary>A lock object that synchronizes access to the consensus rules.</summary>
+        private readonly object consensusLock;
+
+        public ConsensusLoop(
```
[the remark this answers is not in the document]
dangershony: Ah good point will do that

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
dangershony: Actually this does not behave differently then before, the consensus loop would have also block to update the stats.

Perhaps we should make it a background loop in a (later PR) or as part of the node bench loop

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
```diff
@@ -103,7 +94,9 @@ public void Log()
 
         protected override void OnNextCore(Block value)
         {
-            throw new NotImplementedException();
+            if (DateTimeOffset.UtcNow - this.lastSnapshot.Taken > TimeSpan.FromSeconds(5.0))
+                if (this.chainState.IsInitialBlockDownload)
```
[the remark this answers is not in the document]
dangershony: Easier to read I guess.

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
Aprogiena: Oh, i can see it now, i was expecting the original was something else. You are right

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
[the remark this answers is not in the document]
Aprogiena: I would probably also use two lines, but one if :D 
This is detail, each of us have slightly different preferences. All fine

Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -432,14 +431,12 @@ public IAsyncLoop Mine(WalletSecret walletSecret)
         /// One a new block is staked, this method is used to verify that it
         /// is a valid block and if so, it will add it to the chain.
         /// </summary>
-        /// <param name="context">Information about the new block.</param>
+        /// <param name="block">The new block.</param>
         /// <param name="chainTip">Block that was considered as a chain tip when the block staking started.</param>
-        private async Task CheckStakeAsync(ContextInformation context, ChainedBlock chainTip)
+        private async Task CheckStakeAsync(Block block, ChainedBlock chainTip)
```
Aprogiena: got compiler warning `warning CS1998: This async method lacks 'await' operators and will run synchronously.`

Aprogiena: also i've noticed i've introduced a warning at https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs#L66 - `this.` is missing there, could you maybe just add it together with fixing your warning?

Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -1,78 +1,81 @@
...
+        /// <summary>The last time the <see cref="ibdLastResult"/> was updated.</summary>
+        private long ibdLastUpdate;
+        
+        /// <summary>A cached result of the IBD method.</summary>
+        private bool ibdLastResult;
 
-        private long lastUpdate;
-        private bool lastResult;
+        /// <summary>A collection of blocks that have been found to be invalid.</summary>
+        internal ConcurrentDictionary<uint256, uint256> InvalidBlocks;
```
Aprogiena: I think this should not be internal, but private. The only place we use it is in `CheckAnnouncedBlocks` and that can easily be rewritten in a way that this can be private. All we need to do is to expose number of elements here and then to avoid accessing ContainsKey directly and just use IsMarkedInvalid which we already have.

Aprogiena: OK, I correct myself. Just delete CheckAnnouncedBlocks completely, it is not used anywhere and it is flawed - it uses EnumerateToGenesis, which seems crazy. But not used anywhere, so just delete it and make this private.

Aprogiena: erm, now I realized you are using concurrent dictionary instead of HashSet ... the correct thing you would like to use here is ConcurrentHashSet which does not exist, so you can either implement it, or just use a lock

however, if you still insist on using concurrent dictionary instead (which is wrong ;) ) then please, at least make it CD<uint256, byte> not to occupy memory with unused 32 bytes of a thing

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+        /// Dispose components in <see cref="ConsensusLoop"/>.
+        /// </summary>
+        public void Stop()
         {
-            while (true)
+            this.asyncLoop?.Dispose();
+        }
+
+        /// <summary>
+        /// A puller method that will continuously loop and ask for the next block  in the chain from peers.
```
Aprogiena: double space typo

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+        /// The block will then be passed to the consensus validation. 
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token that will stop the loop.</param>
+        /// <returns>The task the loop will be running under.</returns>
+        private Task PullerLoop(CancellationToken cancellationToken)
```
Aprogiena: why is this Task? why not void? there is no awaiting and you just return completed task, it does not make much sense to me

Aprogiena: the cancellation can be thrown up the async loop would handle it well i think, so i don't see any reason for task here

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+                    BlockItem item = new BlockItem();
+
+                    using (new StopwatchDisposable(o => this.Validator.PerformanceCounter.AddBlockFetchingTime(o)))
+                    {
+                        item.Block = this.Puller.NextBlock(cancellationToken);
+                    }
+
+                    this.logger.LogTrace("Block received from puller.");
+                    this.AcceptBlock(item);
+
```
Aprogiena: delete this empty line

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+            return Task.CompletedTask;
         }
 
-        public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
+        /// <summary>
+        /// A method that will accept a new block to the node.
+        /// The block will be validated and the <see cref="CoinView"/> db will be updated.
+        /// If its a new block that extends the chain then the <see cref="ConcurrentChain.Tip"/> will be set.
+        /// </summary>
+        /// <param name="item">The block to validate</param>
```
Aprogiena: missing dot

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
             }
+
+            return Task.CompletedTask;
         }
 
-        public BlockResult ExecuteNextBlock(CancellationToken cancellationToken)
+        /// <summary>
+        /// A method that will accept a new block to the node.
+        /// The block will be validated and the <see cref="CoinView"/> db will be updated.
+        /// If its a new block that extends the chain then the <see cref="ConcurrentChain.Tip"/> will be set.
```
Aprogiena: it's

Aprogiena: this comment confused me, only after a while i've realized that such a block can extend the chain if we mined it (pos/pow) and it should not happen if we downloaded it from peer, i would explain this little bit more, so this is clear

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
-            try
+            lock (this.consensusLock)
             {
-                using (new StopwatchDisposable(o => this.Validator.PerformanceCounter.AddBlockFetchingTime(o)))
+                if (item.Block == null)
                 {
+                    this.logger.LogTrace("No block received from puller due to reorganization, rewinding.");
+                    ChainedBlock lastTip = this.Tip;
+
+                    var token = this.nodeLifetime.ApplicationStopping;
```
Aprogiena: unvar

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+        public void AcceptBlock(BlockItem item)
         {
             this.logger.LogTrace("()");
 
-            BlockResult result = new BlockResult();
-            try
+            lock (this.consensusLock)
             {
-                using (new StopwatchDisposable(o => this.Validator.PerformanceCounter.AddBlockFetchingTime(o)))
+                if (item.Block == null)
```
Aprogiena: are you sure this belongs here? if so, please make sure the method's param comment for `item` specifically mentions this specific configuration

however, i'm not sure why this is not in the puller loop, it could as well use this lock, you would just need to change its comment, which currently says that it is only used in accept block, instead it should say which shared structures it protects

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+                    if (item.Error == ConsensusErrors.BadWitnessNonceSize)
+                    {
+                        this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
+                        this.connectionManager.AddDiscoveredNodesRequirement(NodeServices.NODE_WITNESS);
+                        this.Puller.RequestOptions(TransactionOptions.Witness);
+                        return;
+                    }
+
+                    // Set the chain back to ConsensusLoop.Tip.
+                    this.Chain.SetTip(this.Tip);
```
Aprogiena: this seems dangerous, but it is probably necessary
is this the improvement that should solve the situation when a peer claims long chain but then we find out that the chain is invalid (previously this would leave the tip at the high value and only after the real chain would advance enough, it would catch up) ?

the little problem with this is that when you are here, you could already see a valid chain that is longer than this.Tip, so you are throwing something valid away, but maybe that is fine to do - my concern would be that your good peers won't resend you those headers and so it seems to me that you can actually be stuck if this happens

i'm not sure if i'm explaining this race condition i have in the mind well, please let me know if you can see it

Aprogiena: the problem is little bit worse, because let's say we have one bad peer and two good peers, bad peer gives us headers going up to block 200, but real chain is just 150 and we are (consensus) at 100

so this resets the consensus to 100, but those two guys won't send you headers again, so you are still at 100 and i don't think you would even try to get more blocks because you think the chain is at 100

what you could do, however, is to go over all your good peers and check what's their position and set headers there

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+
+                    // Since ChainHeadersBehavior check PoW, MarkBlockInvalid can't be spammed.
+                    this.logger.LogError("Marking block as invalid.");
+                    this.chainState.MarkBlockInvalid(item.Block?.GetHash());
+                }
+                else
+                {
+                    this.logger.LogTrace("Block accepted '{0}' ", this.Tip);
+
+                    this.chainState.HighestValidatedPoW = this.Tip;
```
Aprogiena: in the error branch above, how is that we don't set this HVPoW?

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: Now thinking about it again, this seems to me that it is actually an improvement over the current state. It is probably true that good peers won't send you headers, so you would be stuck in block 100, but when 1 new block arrives, you will get that information and you will sync to 151. Previously you would need to wait until block 201 is available. So I think this is an improvement and maybe it is good enough and maybe we don't need to go over the list of good peers and set tip to their best.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+        /// If its a new block that extends the chain then the <see cref="ConcurrentChain.Tip"/> will be set.
+        /// </summary>
+        /// <param name="item">The block to validate</param>
+        public void AcceptBlock(BlockItem item)
         {
             this.logger.LogTrace("()");
 
-            BlockResult result = new BlockResult();
-            try
+            lock (this.consensusLock)
```
Aprogiena: I have three problems here. First, the critical section is enormous. It can take very long time. Second, you can't await inside lock(). Third, as the section is very long, another thread can be stuck here without cancellation. 

All these problems could be solved by not using lock(), but using something else. And since it seems we will want to await, the newly merged AsyncLock with cancellation token is probably the way to go.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+
+                    // We really want to flush if we are at the top of the chain.
+                    // Otherwise, we just allow the flush to happen if it is needed.
+                    bool forceFlush = this.Chain.Tip.HashBlock == item.ChainedBlock?.HashBlock;
+                    this.FlushAsync(forceFlush).GetAwaiter().GetResult();
+
+                    if (this.Tip.ChainWork > this.Chain.Tip.ChainWork)
+                    {
+                        // This is a newly mined block.
+                        this.Chain.SetTip(this.Tip);
```
Aprogiena: at this point the above condition about chainwork might be invalidated. does it matter? it would mean that we are aware of longer chain, but we set the tip back to our chain

again, when the very next block arrives from that peer who sent us the longer chain, we will get that information, so we will recover and hence it seems it is fine again

so probably OK, just trying to brainstorm over possibilities

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+                        return;
+                    }
+
+                    // Set the chain back to ConsensusLoop.Tip.
+                    this.Chain.SetTip(this.Tip);
+                    this.logger.LogTrace("Chain reverted back to block '{0}' ", this.Tip);
+
+                    // Since ChainHeadersBehavior check PoW, MarkBlockInvalid can't be spammed.
+                    this.logger.LogError("Marking block as invalid.");
+                    this.chainState.MarkBlockInvalid(item.Block?.GetHash());
```
Aprogiena: the ? here is probably never used because item.Block == null is handled above

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
+                    this.FlushAsync(forceFlush).GetAwaiter().GetResult();
+
+                    if (this.Tip.ChainWork > this.Chain.Tip.ChainWork)
+                    {
+                        // This is a newly mined block.
+                        this.Chain.SetTip(this.Tip);
+                        this.Puller.SetLocation(this.Tip);
+
+                        this.logger.LogDebug("Block extends consensus tip to '{0}'", this.Tip);
+                    }
```
Aprogiena: this branch does not really prefer the blocks that we mined as we discussed it earlier, right?

however, if we really think about what will happen exactly, it does not sound right:
- first we receive next block from the peer
- block goes to this method, is being validated
- then we mine a new block, block goes to this method, waits for lock to enter
- peer block is finished and leaves the method
- mined block goes in critical sections
- item.Block is not null, so first part is skipped
- then we go to ValidateBlock where `context.BlockItem.Block.Header.HashPrevBlock != this.Tip.HashBlock` this condition will be true, so we should get `ConsensusErrors.InvalidPrevTip.Throw`
- we go to `Block rejected` branch in AcceptBlock

so we would actually treat a perfectly valid block as invalid block

in previous discussion i've suggested that such a block should instead replace the current tip and we should try to prolong that chain as it seems more beneficial to us than other strategy. maybe that is more complicated and not really necessary to do, so i'm not saying we need to rewrite it as such

but having it jump to Block rejected seems like a mistake. i don't think anything bad happen, that block will be marked as invalid, which just occupies some memory, but except for that no harm will be done. conceptually it feels wrong however - maybe a simple check on whether that InvalidPrevTip was received and handle it gracefully would feel better? i.e. just do nothing instead of going to Block rejected branch

Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -155,24 +155,24 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
...
+                    {
+                        this.logger.LogTrace("(-)[REORG-2]");
                         return blocks;
+                    }
 
-                    // Push the block to disk, so it is available when peers ask for it.
-                    this.blockRepository.PutAsync(blockResult.ChainedBlock.HashBlock, new List<Block> { pblock }).GetAwaiter().GetResult();
+                    if (blockResult.Error != null)
+                    {
+                        if (blockResult.Error == ConsensusErrors.InvalidPrevTip)
```
Aprogiena: this can happen , but just below, we have catch clause that catches this very consensus error and imho that can't happen, so you can maybe completely remove this try/catch block

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: ok, so the above scenario is just weird, but INVERSE scenario is actually dangerous. this is serious bug and MUST be fixed

Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: OK yeah I will fix both cases

Stratis.Bitcoin/Base/ChainState.cs
dangershony: But the value in the dictionary doe not get cloned so its just a reference (no extra memory is allocated)
There is a ConcurrenctHashset nuget package heavily used we did not include it the project (yet).

Yeah that method `CheckAnnouncedBlocks` is anyway not used but I keep it there as it may be used when we implement banning.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: OK, so to just explain that - in the first post here I've analyzed the situation when you first get block from peer and immediately after we mine a block. That's not optimal, but it is handled +- OK without much harm. But the inverse situation is very bad - if we mine a block and just after that we receive block from peer, it seems there is a disaster. That block from peer will be marked as invalid and it is very good chance that the real chain will be built on the top of it (and not our block). Then our node is out of the chain and can't go back.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,188 @@ public void Initialize()
...
-            return result;
+            this.logger.LogTrace("(-):*.{0}='{1}',*.{2}='{3}'", nameof(item.ChainedBlock), item.ChainedBlock, nameof(item.Error), item.Error?.Message);
         }
 
-        public void AcceptBlock(ContextInformation context)
+        /// <summary>
+        /// Validate a block using the consensus rules.
+        /// </summary>
+        /// <remarks>
+        /// WARNING: This method can only be called from <see cref="ConsensusLoop.AcceptBlock"/>,
```
Aprogiena: despite this comment there is PowBlockAssembler.TestBlockValidity which calls this method directly and it seems that it is actually used regularly ...

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Hmmm thats a good point, this reorg can go in the puller loop, perhaps its even better there as this is reorg logic that is very much connected to the puller (block = null is a puller logic)

Ok yeah I will move it.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -215,6 +427,11 @@ public void AcceptBlock(ContextInformation context)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// This method try to load from cash the items of the next block, in a background task.
```
Aprogiena: cach`e`

Aprogiena: and maybe not items, but UTXOs ? or what do you mean by items?

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -215,6 +427,11 @@ public void AcceptBlock(ContextInformation context)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// This method try to load from cash the items of the next block, in a background task. 
+        /// </summary>
+        /// <param name="flags">A flag to provide information about activated features.</param>
```
Aprogiena: I would just say `Information about activated features.`

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -232,6 +449,12 @@ private Task TryPrefetchAsync(DeploymentFlags flags)
             return prefetching;
         }
 
+        /// <summary>
+        /// Get transaction identifiers to try to pref-etch them from cache.
```
Aprogiena: prefetch or pre-fetch

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -232,6 +449,12 @@ private Task TryPrefetchAsync(DeploymentFlags flags)
...
+        /// </summary>
+        /// <param name="block">The block containing transactions to fetch.</param>
+        /// <param name="enforceBIP30">The BIP30 flag.</param>
```
Aprogiena: I'd say `<c>true</c> to enforce BIP30.`

Stratis.Bitcoin/Base/ChainState.cs
Aprogiena: hmm you are correct with the reference, but still are we not talking 1 reference amount of memory vs 1 byte? anyway, I would just go for hashset with lock, that's what you want to use, you don't want dictionary ...

that CAB function is flawed entirely, we should remove it for sure and make this private and only when we need something then to implement it correctly

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: This functionality is same as before but I see your point, it is necessary though.
It will be possible to detect block 150 yeah if we disconnect from the 200 (malicious peer) and assume the rest are honest.

This is should definitely not happen in the consensus code but we may be able to trigger such an operation.

There are 3 ways we can fix this (as part of another PR)
- Just disconnect the peer and don't change Chain.Tip and let issue #229 rewind the header chain to the most recent best height. this will never set chain to 100 but have it directly set to 150 slightly delays.
- Create a new behavior that will do the same as the previous point but triggered from consensus (not from node disconnect).
- Set Chain tip to consensus and trigger out nodes to sync the chain (send an GetHeaders to all peers) this will set Cahin back to 100 and sync it up to 150

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Yes good point fixed

Stratis.Bitcoin/Base/ChainState.cs
dangershony: Ok fair enough but I will leave this of another PR as I am already doing too much out of this task.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: I removed it (this will just come back when we make this async).

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Yes, I am still not sure how we will make this async I have two ideas (either AsyncLock or a dedicated loop) but in any case only 2 (or 3 threads) will be ever accessing this code (puller and both the miners) so not a big deal (right now).

For that reason the fact the critical section is big is not such a big deal.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Yeah it matters if we are in IBD then chain will be behind, but if we are beyond IBD then this could be assumed to always be correct

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Ah yes this is a big but however the mark as invalid code is not used (yet).
But you are right we should probably verify the reason of failure if the reason is only less work then to not mark the block as invalid, but I guess this should be done in the banning PR.

Stratis.Bitcoin.Features.Miner/PowMining.cs
dangershony: Good point

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Yeah the next line explains that.
>it the requirement is to validate a block without affecting the consensus db then <see cref="ContextInformation.OnlyCheck"/> must be set to true.

Perhaps breaking this method in two is not a bad idea (one part validates without changing any state) the other updates state (the db, cache)

This will be in another PR but I will add a TODO

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: fixed

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: I think for this PR this is good enough. Further improvements should be out of scope here as you say. So I'm OK here.

Stratis.Bitcoin/Base/ChainState.cs
Aprogiena: agree

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: yes, i think it is better to have incremental progress and not to have half-baked things that "wait for later use". when we need it, the change is trivial, but should we forget about it or should we later come with different evolution of the code, we would leave a technical debt and we should avoid that

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: that is fine, the problem is that lock() is not very good thing when your critical section is so large
that asynclock is based on semaphore and that's very suitable here
i'm talking about the (in)efficiency of entering the lock 

so yes, i agree it is not a big deal, but the fix is absolutely trivial and it would look much better
instead of 
`lock (x)` we use `using (x.LockAsync(cancellationtokenhere))` for some `private AsyncLock x = new AsyncLock();`

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: actually I think this is used already in CHB,
and even if it was not, this is a bug problem and i would say it must be fixed in this PR
the fix is rather trivial, so no need not to do it
that situation will be very often (esp. on testnet)

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: I see what you mean now, but the comment is really misleading. It should say
`WARNING: This method can only be called from <see cref="ConsensusLoop.AcceptBlock"/>, otherwise <see cref="ContextInformation.OnlyCheck"/> must be set to <c>true</c>, which causes that the method only validates the block without affecting the consensus DB.`

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,197 @@ public void Initialize()
...
+                    this.Chain.SetTip(this.Tip);
+                    this.logger.LogTrace("Chain reverted back to block '{0}' ", this.Tip);
+
+                    // Since ChainHeadersBehavior check PoW, MarkBlockInvalid can't be spammed.
+                    this.logger.LogError("Marking block as invalid.");
+                    this.chainState.MarkBlockInvalid(item.Block.GetHash());
+                }
+                else
+                {
+                    this.logger.LogTrace("Block accepted '{0}' ", this.Tip);
```
Aprogiena: I would prefer `Block '{0}' accepted.`, otherwise `Block accepted: '{0}'` with colon and no space at the end

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -62,78 +164,197 @@ public void Initialize()
...
+                    {
+                        this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
+                        this.connectionManager.AddDiscoveredNodesRequirement(NodeServices.NODE_WITNESS);
+                        this.Puller.RequestOptions(TransactionOptions.Witness);
+                        return;
+                    }
+
+                    // Set the chain back to ConsensusLoop.Tip.
+                    this.Chain.SetTip(this.Tip);
+                    this.logger.LogTrace("Chain reverted back to block '{0}' ", this.Tip);
```
Aprogiena: space -> dot 
at the end

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -23,25 +27,74 @@ public class ConsensusLoop
...
+
+        /// <summary>The current tip of the cahin that has been validated.</summary>
         public ChainedBlock Tip { get; private set; }
+
+        /// <summary>Contain information about deployment and activation of features in the chain.</summary>
         public NodeDeployments NodeDeployments { get; private set; }
 
-        public ConsensusLoop(PowConsensusValidator validator, ConcurrentChain chain, CoinView utxoSet, LookaheadBlockPuller puller, NodeDeployments nodeDeployments, ILoggerFactory loggerFactory, StakeChain stakeChain = null)
+        /// <summary>Factory for creating and also possibly starting application defined tasks inside async loop.</summary>
+        private readonly IAsyncLoopFactory asyncLoopFactory;
```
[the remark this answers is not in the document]
Aprogiena: this is done

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: this is done

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: this is done

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: this is done

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
Aprogiena: this is done

Stratis.Bitcoin/Base/ChainState.cs
Aprogiena: this is done

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Where is it used? CHB does not act on invalid blocks.
But ok as per your request I'll remove it.

Aprogiena: i can see it in headerspayload processing: 
`validated &= !this.SharedState.IsMarkedInvalid(tip.HashBlock);`

Aprogiena:
I've created issue #624 for things that were discussed in this PR, but won't be included, but should be done relatively soon.

Besides the list of things in this new issue, another separated thing is making this code fully async.

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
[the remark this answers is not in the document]
dangershony: This anyway gets optimized by the CLR

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -1,47 +1,146 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Stratis.Bitcoin.Features.Consensus
 {
-    public class BlockResult
+    /// <summary>
+    /// A block object that is passed around when downloading a block from peers and has information about the validation state of the block.
+    /// </summary>
+    public class BlockItem
```
dangershony: I would be up for another name for this class, perhaps ConsensusBlock or ValidationBlock

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Actually I did not move this yet to the puller loop as i am not sure how the async code will work.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: OK yeah I see, the fix in this particular case will be to not mark as invalid if the reason is only work.
So if the error is InvalidPreviousBlock then will not mark the block as invalid, how ever when banning is implemented each error will have a severity then the severity will need to be used..

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: The issue I have here is the lock, and that depends on how we implement async.
I might decide to add a second loop for consensus that will be async and not require a lock at all then the rewind will have to go back to the AccpetBlock method.
However if we use the AsyncLock it might be find to move the code tot he puller loop.

Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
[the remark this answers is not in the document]
Aprogiena: of course, we were just talking our subjective preference for a beautiful code :)

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: sure, no problem with that, just make sure you are aware of all code that actually uses it, so that the name fits all

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: yes, just handle InvalidPrevTip differently
it is not about work, it is about whether the block that arrived prolongs what we consider the best chain or not and as we shown it can be the case that valid block that does not prolong the best chain is processed here

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: the last version i was testing did have this in the puller loop already
all that is needed there is just to replace the lock with AsyncLock and it should be fine

Aprogiena:
The latest commit have the if branch too low. The if should be even before the puller is set - so just after logging that block was rejected.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: You normally prolong the chain by providing proof you made enough work to prolong it (or in this case stake)

dangershony:
>this should go in front of SetTip just like you do with BadWitnessNonceSize  

Yes I agree.

Aprogiena:
during testing, a reorg happened and it went like this, is that ok?

```
[2017-10-22 10:31:26.2261 123] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__70_0 ======Node stats====== 10/22/2017 10:31:26 agent StratisBitcoin:1.0.2
Headers.Height:        158880   Headers.Hash:         995df73e6db7f1bae734becb597215175f0898a1061863b8fd9c1721d55b188a
Consensus.Height:      158880   Consensus.Hash:       995df73e6db7f1bae734becb597215175f0898a1061863b8fd9c1721d55b188a
BlockStore.Height:     158880   BlockStore.Hash:      995df73e6db7f1bae734becb597215175f0898a1061863b8fd9c1721d55b188a
Wallet.Height:         158880   Wallet.Hash:          995df73e6db7f1bae734becb597215175f0898a1061863b8fd9c1721d55b188a

...

[2017-10-22 10:31:30.0196 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.1289 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.2309 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.3379 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.4473 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.5567 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.6587 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:30.7625 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.

...

[2017-10-22 10:31:31.2316 123] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__70_0 ======Node stats====== 10/22/2017 10:31:31 agent StratisBitcoin:1.0.2
Headers.Height:        158881   Headers.Hash:         88d6197442a499118c1360beb6a68abb8ce8fb55cd5f0d69aedf8bd42e37036a
Consensus.Height:      158881   Consensus.Hash:       88d6197442a499118c1360beb6a68abb8ce8fb55cd5f0d69aedf8bd42e37036a
BlockStore.Height:     158880   BlockStore.Hash:      995df73e6db7f1bae734becb597215175f0898a1061863b8fd9c1721d55b188a
Wallet.Height:         158879   Wallet.Hash:          0d7440f4dab2a817dfed522a2ac4f40890230d67d8ac4a214150d96afe418ff4

...

[2017-10-22 10:31:32.2627 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:32.3720 13] WARN: Stratis.Bitcoin.Features.Wallet.WalletSyncManager.ProcessBlock Wallet is behind the best chain and the next block is not found in store.
[2017-10-22 10:31:36.2350 79] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__70_0 ======Node stats====== 10/22/2017 10:31:36 agent StratisBitcoin:1.0.2
Headers.Height:        158881   Headers.Hash:         88d6197442a499118c1360beb6a68abb8ce8fb55cd5f0d69aedf8bd42e37036a
Consensus.Height:      158881   Consensus.Hash:       88d6197442a499118c1360beb6a68abb8ce8fb55cd5f0d69aedf8bd42e37036a
BlockStore.Height:     158881   BlockStore.Hash:      88d6197442a499118c1360beb6a68abb8ce8fb55cd5f0d69aedf8bd42e37036a
Wallet.Height:         158879   Wallet.Hash:          0d7440f4dab2a817dfed522a2ac4f40890230d67d8ac4a214150d96afe418ff4

...

[2017-10-22 10:32:51.4278 79] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__70_0 ======Node stats====== 10/22/2017 10:32:51 agent StratisBitcoin:1.0.2
Headers.Height:        158882   Headers.Hash:         34906c8e12ddc36ae4a079c587f2ec87aee8520071233d71434861682cdcab5b
Consensus.Height:      158882   Consensus.Hash:       34906c8e12ddc36ae4a079c587f2ec87aee8520071233d71434861682cdcab5b
BlockStore.Height:     158882   BlockStore.Hash:      34906c8e12ddc36ae4a079c587f2ec87aee8520071233d71434861682cdcab5b
Wallet.Height:         158882   Wallet.Hash:          34906c8e12ddc36ae4a079c587f2ec87aee8520071233d71434861682cdcab5b
```


actually another reorg looking very similarly to this one happened before that as well and it seems that wallet is often reacting like this, but then everything seems ok, so probably all good

D3m0nKingx:
I've seen this too before @Aprogiena and it just continued as normally. but when I saw this it was appearing prior to showing the heights.

Aprogiena:
Oh yes i remember now, this should be solved by wallet having its own puller. so seems unrelated

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -16,23 +16,18 @@
 namespace Stratis.Bitcoin.Features.Consensus
 {
     /// <summary>
-    /// A block object that is passed around when downloading a block from peers and has information about the validation state of the block.
+    /// Information about a block that is required for its validation. 
+    /// It is used when a new block is downloaded or mined.
     /// </summary>
-    public class BlockItem
+    public class BlockValidationContext
```
dangershony: This is a good name yes, come to think about it I would actually swap this with the validation context so  `ContextInformation` to `ValidationContext` and this field to be higher level perhaps `ConsensusContext`.

Leave it as is for now.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -199,64 +199,47 @@ private void PullerLoop(CancellationToken cancellationToken)
...
-                    if (ex is OperationCanceledException)
-                    {
-                        if (this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
-                            return;
-                    }
 
-                    // TODO Need to revisit unhandled exceptions in a way that any process can signal an exception has been
-                    // thrown so that the node and all the disposables can stop gracefully.
-                    this.logger.LogCritical(new EventId(0), ex, "Consensus loop at Tip:{0} unhandled exception {1}", this.Tip, ex.ToString());
-                    NLog.LogManager.Flush();
```
dangershony: do we need to put this in in the AsyncLoop exception handler?

dangershony:
Yeah this is fine, the problem is a store delay, wallet reorged and it needs to get the next block ( I would expect it to be the block that was received) but if not the wallet will look for the block in store, however store has its own internal loop so it sometimes has a delay before the block is pushed to store (waiting for batches or something) hence this message, the wallet will abandon the block and try to fetch it again when the next block is received.  

Yeah it will be solved when wallet has its own puller.

dangershony:
This looks very good now I will do some more mining and we push it.

Aprogiena:
i've done about 12 hours of mining, i'm approving this

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: we don't because asyncloop crash just crashes the loop itself, not the whole app, flush is needed only when whole app crashes

PR #610, opened 2017-10-17
Implement best IEnumerable practices (better performance)

nopara73:


Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -502,20 +502,16 @@ public IEnumerable<FlatHistory> GetHistory(Wallet wallet)
             lock (this.lockObject)
             {
                 // Get transactions contained in the wallet.
-                List<FlatHistory> items = this.GetHistoryInternal(wallet).SelectMany(s => s.Transactions.Select(t => new FlatHistory { Address = s, Transaction = t })).ToList();
+                IEnumerable<FlatHistory> items = this.GetHistoryInternal(wallet).SelectMany(s => s.Transactions.Select(t => new FlatHistory { Address = s, Transaction = t }));
```
dangershony: This will be enumerated outside of the lock right?

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -650,7 +646,7 @@ public List<UnspentOutputReference> GetSpendableTransactionsInAccount(WalletAcco
                 res = account.GetSpendableTransactions(this.chain.Tip.Height, confirmations);
             }
 
-            this.logger.LogTrace("(-):*.{0}={1}", nameof(res.Count), res.Count);
+            this.logger.LogTrace("(-):*.Count={1}", res.Count());
```
bokobza: `*.Count={1}"` should be `*.Count={0}"`.
Also, this is a tracing log so we're not too worried about performance here and we're taken the general approach to use `nameof` as it leads to fewer mistakes.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -612,30 +608,30 @@ public uint256 LastReceivedBlockHash()
...
             Wallet wallet = this.GetWalletByName(walletName);
-            List<UnspentOutputReference> res;
+            IEnumerable<UnspentOutputReference> res;
             lock (this.lockObject)
             {
                 res = wallet.GetAllSpendableTransactions(this.coinType, this.chain.Tip.Height, confirmations);
             }
 
-            this.logger.LogTrace("(-):*.{0}={1}", nameof(res.Count), res.Count);
+            this.logger.LogTrace("(-):*.Count()={1}", res.Count());
```
bokobza: same comment as below.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: I would think the same as Dan.
With a lock we protect this piece of code, not the wallet object itself, but what we really want to protect is iterating on the wallet outside of a lock.
It'll be interesting to know how this plays out.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
nopara73: You cannot do nameof on a function.

bokobza: ah it's a function now, fair enough.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -604,38 +597,37 @@ public uint256 LastReceivedBlockHash()
...
         /// <inheritdoc />
-        public List<UnspentOutputReference> GetSpendableTransactionsInWallet(string walletName, int confirmations = 0)
+        public IEnumerable<UnspentOutputReference> GetSpendableTransactionsInWallet(string walletName, int confirmations = 0)
         {
             Guard.NotEmpty(walletName, nameof(walletName));
             this.logger.LogTrace("({0}:'{1}',{2}:{3})", nameof(walletName), walletName, nameof(confirmations), confirmations);
 
             Wallet wallet = this.GetWalletByName(walletName);
-            List<UnspentOutputReference> res;
+            IEnumerable<UnspentOutputReference> res;
```
bokobza: you can put this inline with the initialization of `res` I guess.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -384,10 +384,9 @@ public string GetExtPubKey(WalletAccountReference accountReference)
                 HdAccount account = wallet.GetAccountByCoinType(accountReference.AccountName, this.coinType);
                 res = account.ExtendedPubKey;
 
+                this.logger.LogTrace("(-):'{0}'", res);
```
dangershony: I suggest putting the trace out of the lock.
Inside a lock we have to do the minimal possible operations that only effect the component the lock is protecting

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -502,20 +501,16 @@ public IEnumerable<FlatHistory> GetHistory(Wallet wallet)
             lock (this.lockObject)
             {
                 // Get transactions contained in the wallet.
-                List<FlatHistory> items = this.GetHistoryInternal(wallet).SelectMany(s => s.Transactions.Select(t => new FlatHistory { Address = s, Transaction = t })).ToList();
+                FlatHistory[] items = this.GetHistoryInternal(wallet).SelectMany(s => s.Transactions.Select(t => new FlatHistory { Address = s, Transaction = t })).ToArray();
 
-                this.logger.LogTrace("(-):*.{0}={1}", nameof(items.Count), items.Count);
+                this.logger.LogTrace("(-):*.Count={0}", items.Count());
```
dangershony: Same as previous comment

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
nopara73: You are reading the resource with LogTrace.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
nopara73: Same as previous comment.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
nopara73: Thanks. Fixed.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
dangershony: The `resource` you mean the parameter `res`?
That's fine it can be out of the lock, you are not manipulating it..

nopara73: > That's fine it can be out of the lock, you are not manipulating it.  

I'm reading it. Consider this: 1. lock ends, 2. other thread writes it, 3. count() calculated (wrongly!)

Aprogiena: I don't really mind having a log inside lock. Only look at performance when logging is at default configuration (i.e. trace logs are disabled). Then the performance hit is absolutely negligible.

dangershony: >I'm reading it. Consider this: 1. lock ends, 2. other thread writes it, 3. count() calculated (wrongly!)

If you want to trace such a count then do the count inside the lock and trace it outside, but its negligible, besides if count can be different outside of the lock then you have synchronization issues regardless of the trace, 

@Aprogiena I consider it bad practice to trace inside a lock  (unless absolutely required lets avoid it)

nopara73: I consider it bad practice to access something access something outside the lock when the lock is meant to access that something.  

> I consider it bad practice to trace inside a lock  

Indeed.  

> If you want to trace such a count then do the count inside the lock and trace it outside,  

I'd do this. @Aprogiena What do you think?

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -611,10 +611,9 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactionsInWallet(stri
             this.logger.LogTrace("({0}:'{1}',{2}:{3})", nameof(walletName), walletName, nameof(confirmations), confirmations);
 
             Wallet wallet = this.GetWalletByName(walletName);
-            IEnumerable<UnspentOutputReference> res;
             lock (this.lockObject)
             {
-                res = wallet.GetAllSpendableTransactions(this.coinType, this.chain.Tip.Height, confirmations).ToArray();
+                UnspentOutputReference[] res = wallet.GetAllSpendableTransactions(this.coinType, this.chain.Tip.Height, confirmations).ToArray();
```
dangershony: In this case `res` is already enumerated using the count outside of the loop will is completely safe.
outside of the lock even if the wallet changes res will not be effected.

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
Aprogiena: there are different situations

1) if you put log into the critical section, the value logged will guarantee the correct state of the object and if you put it outside, it won't (i.e. it might be different) - in most cases this is not a problem to put the log outside as the logs are mostly informative, but rarely it can happen that you have a bug and you would really appreciate to have all information in logs correct, otherwise you can waste quite a lot of time wondering what just happened

the con of having the log inside the critical section is having the CS is little wider. however, in case of trace log, it is only wider of one call and one if, that's practically absolute zero performance hit. imho this performance hit is not enough reason to put the log outside

i would certainly **not** call this a bad practice

2) if you put log into the critical section, nothing changes (because the log logs something that is orthogonal to the CS), in this case naturally you would put the log outside **as a good practice** - there is no benefit in this case having the log inside CS, therefore it should not be there because CSs should be as short as possible

-----------

so which one is it here, imho this one is obviously 2) because it could be written as 

```
           string res = null;
           lock (this.lockObject)
            {
                // Get the account.
                HdAccount account = wallet.GetAccountByCoinType(accountReference.AccountName, this.coinType);
                res = account.ExtendedPubKey;
            }
            this.logger.LogTrace("(-):'{0}'", res);
            return res;
```

the `res` value can't change outside the lock because it is local variable, therefore there is no benefit to have the log inside CS, therefore it is 2) and the original suggestion of @dangershony is correct

Aprogiena: in this particular case, i'm not sure if even this:
```
            lock (this.lockObject)
            {
                // Get the account.
                HdAccount account = wallet.GetAccountByCoinType(accountReference.AccountName, this.coinType);
            }

            string res = account.ExtendedPubKey;
            this.logger.LogTrace("(-):'{0}'", res);
            return res;
```

would not work. the lock is said to be needed for collections over wallet, so it should have no effect on account and its ExtendedPubKey. but i'm not familiar with this code, this is just how i read the contract of lockObject

little offtopic here, but i would definitely add a comment to GetAccountByCoinType that its caller is responsible to hold this lock and i would change its name to GetAccountByCoinTypeLocked

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
Aprogiena: yes, here again it seems like GetAllSpendableTransactions should be named GetAllSpendableTransactionsLocked and there should be extra comment that the lock must be held

and again, it seems like this could be written as 

```
            UnspentOutputReference[] res;
            lock (this.lockObject)
            {
                res = wallet.GetAllSpendableTransactions(this.coinType, this.chain.Tip.Height, confirmations).ToArray();
            }

            this.logger.LogTrace("(-):*.{0}={1}", nameof(res.Length), res.Length);
            return res;
```

so the CS would be shorter

nopara73: Will do so. Btw: this is the result of our failure of backpedalling from deferred execution :)

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
nopara73: @dangershony Make your decision and let's wrap this small issue up.

dangershony: Yes @Aprogiena is right on the last case `ExtendedPubKey` does not need to be protected by the lock (only the lookup for account) so we just need to lock the the method (and enumeration).

Stratis.Bitcoin.Features.Wallet/WalletManager.cs
Aprogiena: could you pls explain what you mean here? i mean what do you suggest in general and what kind of failure this is and how it would be prevented

dangershony:
This looks good I will let Jeremy commit

PR #613, opened 2017-10-18
Add AsyncLock

nopara73:


nopara73:
Ugly commit history (I'll make it in two commits)

PR #618, opened 2017-10-19
Feature/coinview flushing issue 525 take two

### LINKED ISSUE #426: [StakeChainStore] Crash

### LINKED ISSUE #525: [CoinView] Investigate no flush during IBD
It seems to me that CV has no flushing during IBD, which makes the memory to grow extremely during IBD.

Note that if we implemented flush, we have to add lock and be careful not to introduce #426 again due to race condition between Ctrl+C handler which executes the flush and this new flush.

Aprogiena:
Before, during IBD, coinview did not flush. Flushing was implemented to be done in consensus loop in case our consensus chain processed a block that is at the tip of the best chain (i.e. not in IBD). Second flushing was implemented in `FetchCoinsAsync` method if the number of items in the cache reached certain (very high limit). Third flushing execution path was during shutdown. 

Before, we fixed #426, which was a race condition between consensus loop accepting of a new block and the shutdown flushing thread. We solved it by first stopping the consensus loop and only then doing the flush. Naturally this problem is not present in the first flush way because it is done from consensus loop, but we did not fix the second way from `FetchCoinsAsync`. 

So we have now removed flushing possibility from FCA and we kept the already fixed shutdown flushing and we changed flushing from consensus loop thread in a way it would flush more often (not just when we are at the tip). Now it is normal that this consensus loop flush does the flush in the middle of IBD and it does so if the last flush is more than 1 minute old.

This prevents the negative effect of #525, which was that due to no flushing for very long time, there were even hundreds of thousands blocks of data unflushed and when user invoked shutdown of the node, the waiting time for the flush to finish was even in hours. This should not happen anymore since we flush every minute during IBD.

Fix #525

### DIFF
--- a/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
+++ b/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
@@ -241,12 +241,12 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// WARNING: This method can only be run from <see cref="ConsensusLoop.Execute(System.Threading.CancellationToken)"/> thread context
         /// or when consensus loop is stopped. Otherwise, there is a risk of race condition when the consensus loop accepts new block.
         /// </remarks>
-        public async Task FlushAsync(bool force)
+        public async Task FlushAsync(bool force = true)
         {
             this.logger.LogTrace("({0}:{1})", nameof(force), force);
 
             DateTime now = this.dateTimeProvider.GetUtcNow();
-            if (!force && (now - this.lastCacheFlushTime).TotalSeconds < CacheFlushTimeIntervalSeconds)
+            if (!force && ((now - this.lastCacheFlushTime).TotalSeconds < CacheFlushTimeIntervalSeconds))
             {
                 this.logger.LogTrace("(-)[NOT_NOW]");
                 return;
--- a/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
+++ b/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using NBitcoin;
 using Stratis.Bitcoin.Utilities;
 using Microsoft.Extensions.Logging;
+using Stratis.Bitcoin.Base;
 
 namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 {
@@ -57,6 +58,10 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// <summary>Default maximum number of transactions in the cache.</summary>
         public const int CacheMaxItemsDefault = 100000;
 
+        /// <summary>Length of the coinview cache flushing interval in seconds.</summary>
+        /// <seealso cref="lastCacheFlushTime"/>
+        public const int CacheFlushTimeIntervalSeconds = 3 * 60;
+
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
@@ -98,14 +103,21 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// <summary>Task that handles rewinding of the the underlaying coinview. Used for synchronization.</summary>
         private Task rewindingTask = Task.CompletedTask;
 
+        /// <summary>Provider of time functions.</summary>
+        private readonly IDateTimeProvider dateTimeProvider;
+
+        /// <summary>Time of the last cache flush.</summary>
+        private DateTime lastCacheFlushTime;
+
         /// <summary>
         /// Initializes instance of the object based on DBreeze based coinview.
         /// </summary>
         /// <param name="inner">Underlaying coinview with database storage.</param>
+        /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="loggerFactory">Factory to be used to create logger for the puller.</param>
         /// <param name="stakeChainStore">Storage of POS block information.</param>
-        public CachedCoinView(DBreezeCoinView inner, ILoggerFactory loggerFactory, StakeChainStore stakeChainStore = null) :
-            this(loggerFactory, stakeChainStore)
+        public CachedCoinView(DBreezeCoinView inner, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory, StakeChainStore stakeChainStore = null) :
+            this(dateTimeProvider, loggerFactory, stakeChainStore)
         {
             Guard.NotNull(inner, nameof(inner));
             this.inner = inner;
@@ -115,14 +127,15 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// Initializes instance of the object based on memory based coinview.
         /// </summary>
         /// <param name="inner">Underlaying coinview with memory based storage.</param>
+        /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="loggerFactory">Factory to be used to create logger for the puller.</param>
         /// <param name="stakeChainStore">Storage of POS block information.</param>
         /// <remarks>
         /// This is used for testing the coinview.
         /// It allows a coin view that only has in-memory entries.
         /// </remarks>
-        public CachedCoinView(InMemoryCoinView inner, ILoggerFactory loggerFactory, StakeChainStore stakeChainStore = null) :
-            this(loggerFactory, stakeChainStore)
+        public CachedCoinView(InMemoryCoinView inner, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory, StakeChainStore stakeChainStore = null) :
+            this(dateTimeProvider, loggerFactory, stakeChainStore)
         {
             Guard.NotNull(inner, nameof(inner));
             this.inner = inner;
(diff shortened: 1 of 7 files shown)


dangershony:
Ah this is a much more simple solution then I though you are going to do.
Let me review closer and we commit.

Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -201,8 +201,11 @@ private Task RunLoop(CancellationToken cancellationToken)
                     if (block.Error == null)
                     {
                         this.chainState.HighestValidatedPoW = this.consensusLoop.Tip;
-                        if (this.chain.Tip.HashBlock == block.ChainedBlock?.HashBlock)
-                            this.consensusLoop.FlushAsync().GetAwaiter().GetResult();
+
+                        // We really want to flush if we are at the top of the chain.
+                        // Otherwise, we just allow the flush to happen if it is needed.
+                        bool mustFlush = this.chain.Tip.HashBlock == block.ChainedBlock?.HashBlock;
```
dangershony: `mustFlush` -> `forceFlush`

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -228,9 +236,23 @@ private CachedCoinView(ILoggerFactory loggerFactory, StakeChainStore stakeChainS
...
+        /// WARNING: This method can only be run from <see cref="ConsensusLoop.Execute(System.Threading.CancellationToken)"/> thread context
+        /// or when consensus loop is stopped. Otherwise, there is a risk of race condition when the consensus loop accepts new block.
+        /// </remarks>
+        public async Task FlushAsync(bool force = true)
         {
-            this.logger.LogTrace("()");
+            this.logger.LogTrace("({0}:{1})", nameof(force), force);
+
+            DateTime now = this.dateTimeProvider.GetUtcNow();
+            if (!force && ((now - this.lastCacheFlushTime).TotalSeconds < CacheFlushTimeIntervalSeconds))
```
dangershony: The only issue here is that `CacheMaxItemsDefault` is not applied, so the cache might grow beyond that, probably if we flash every 60 sec it wont.

Perhaps we add that as a condition as well or delete it all together.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -228,9 +236,23 @@ private CachedCoinView(ILoggerFactory loggerFactory, StakeChainStore stakeChainS
...
+            {
+                this.logger.LogTrace("(-)[NOT_NOW]");
+                return;
+            }
+
+            this.lastCacheFlushTime = now;
```
bokobza: I'd move this to the bottom of the method.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
bokobza: I think it's ok as we moved from a size-based threshold to a time-based one.
Checking both seems superfluous to me as like you said, flushing every 60 sec should be enough.

Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
Aprogiena: Yes, will fix

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: Actually, this is something else in my understanding. We use that max items limit for calling Evict inside that FCA method. That actually removes some items from the cache, making the memory consumption of the structure lower. 

However, flushing is not about making the memory consumption lower - note that we don't call Evict in this case. Flushing is only to persist unsaved data to disk, that is all. It won't make the structure smaller. It is true, that flushing allows the Evict to get rid of more items, but those are two separated things. The max items limit is still used for eviction, but this is now separated from flushing.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: It should not matter. At the end of the method, I would tend to use GetUtcNow again, that is maybe what you actually suggest, isn't it?

Aprogiena:
Yes that is correct, initially I was looking for little more complex solution, but due to #426 problem, it was not easily possible. I would have to start introducing async locks (that don't yet exist in the code base) and the whole thing would be little bit messy. So I've decided to solve the problem with nonflushing during IBD with this simple approach and I think it is good enough.

Aprogiena:
Originally, I wanted to introduce two things - time limit and item limit. However, after quick analysis, the item limit would require maintenance of the counter and that would just increase the complexity without any great benefit. So at the end I ended up with just time limit.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: OK, I will move it. The problem is that I'm unsure about the second return there is for null inner tip. I am not familiar with  the code enough to decide when this actually happens and if we should reset lastCFT in that branch as well. Hopefully not, so I will just set it at the end of the method.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
dangershony: Ah right yes correct flush and evict two different things, all good then ignore my comment.

dangershony:
Looks good now.

Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
bokobza: yeah just so that you get the latest time.

PR #623, opened 2017-10-20
Corrected Node Building Instructions

D3m0nKingx:
Corrected missing step for building node, changing `dotnet build` to `dotnet restore` (Restores the dependencies and tools of a project.). dotnet build is redundant as it is already incorporated into `dotnet run`, however it does not display logging, but dotnet run will report any build issues anyway.

### DIFF
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -23,8 +23,7 @@ To install and run the node, you need
 ```
 git clone https://github.com/stratisproject/StratisBitcoinFullNode.git  
 cd StratisBitcoinFullNode
-git checkout master
-git submodule updated --init --recursive
+git submodule update --init --recursive
 ```
 
 ### Build and run the code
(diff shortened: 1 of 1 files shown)


bokobza:
As of dotnet core 2.0, `dotnet restore` is not necessary anymore 😛 
dotnet run will be enough.

Documentation/getting-started.md
```diff
@@ -24,7 +24,7 @@ To install and run the node, you need
 git clone https://github.com/stratisproject/StratisBitcoinFullNode.git  
 cd StratisBitcoinFullNode
 git checkout master
-git submodule init --init --recursive
+git submodule updated --init --recursive
```
bokobza: you're right!
should be `update` though

bokobza: git checkout master is unnecessary indeed.

bokobza:
Thanks @D3m0nKingx !

PR #626, opened 2017-10-20
Apply Async suffixes

nopara73:


Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -60,7 +60,7 @@ public BlockStoreBehavior(ConcurrentChain chain, IBlockRepository blockRepositor
             Guard.NotNull(chain, nameof(chain));
             Guard.NotNull(blockRepository, nameof(blockRepository));
             Guard.NotNull(blockStoreCache, nameof(blockStoreCache));
-            Guard.NotNull(blockStoreCache, nameof(logger));
+            Guard.NotNull(blockStoreCache, nameof(this.logger));
```
Aprogiena: this line was wrong and still is wrong, it checks blockStoreCache, but it most likely was supposed to check the loggerFactory argument

Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -93,7 +93,9 @@ protected override void DetachCore()
             this.logger.LogTrace("(-)");
         }
 
+#pragma warning disable IDE1006 // Naming Styles
         private async void AttachedNode_MessageReceived(Node node, IncomingMessage message)
+#pragma warning restore IDE1006 // Naming Styles
```
Aprogiena: why can't we just append `Async` or `_Async` here instead of using pragmas?

Stratis.Bitcoin.Features.Notifications/TransactionReceiver.cs
```diff
@@ -32,20 +32,20 @@ public TransactionReceiver(TransactionNotification transactionNotification, Tran
...
         }
 
         protected override void DetachCore()
         {
-            this.AttachedNode.MessageReceived -= this.AttachedNode_MessageReceived;
+            this.AttachedNode.MessageReceived -= this.AttachedNode_MessageReceivedAsync;
         }
 
-        private async void AttachedNode_MessageReceived(Node node, IncomingMessage message)
+        private async void AttachedNode_MessageReceivedAsync(Node node, IncomingMessage message)
```
Aprogiena: here i'm confused, above we've had 2 instances of exactly this and we used pragmas
here we use suffix
so if there is nothing that prevents us from using that suffix, use it in all instances and do not introduce those pragmas
or am i missing something?

Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
nopara73: I just didn't think it through.

Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
nopara73: Fixed.

Stratis.Bitcoin.Features.Notifications/TransactionReceiver.cs
nopara73: Actually when I fixed the other ones I got a bunch of errors. I thought it's because of inheritance, so I just pragmad them in all other cases. However in this case I noticed the problem wasn't inheritance, but just a naming conflict. Anyway, fixed.

Aprogiena: i see, tx for explanation, looks good

PR #627, opened 2017-10-21
Feature/asynclock improvement

Aprogiena:
added comments to AsyncLock, minor code polishing included as well

@nopara73 my attempt of making this class commented, I think it can help devs that come to project and e.g. never worked with locks with using to easier understand how this is implemented. Not that it would be difficult to get from the code itself, but sometimes you can appreciate that what you think you see in the code is confirmed in the comment. This assures you that there is nothing you've missed.

### DIFF
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -59,7 +59,7 @@ namespace Stratis.Bitcoin.Utilities
             /// <inheritdoc />
             public void Dispose()
             {
-                this.toRelease?.semaphore.Release();
+                this.toRelease.semaphore.Release();
             }
         }
 
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -28,17 +28,10 @@ namespace Stratis.Bitcoin.Utilities
             Task wait = this.semaphore.WaitAsync(cancel);
 
             if (wait.IsCompleted)
-            {
                 return this.releaser;
-            }
-            else
-            {
-                return wait.ContinueWith((_, state) =>
-                    (IDisposable)state,
-                    this.releaser.Result, cancel,
-                    TaskContinuationOptions.ExecuteSynchronously,
-                    TaskScheduler.Default);
-            }
+
+            return wait.ContinueWith((_, state) => (IDisposable)state, this.releaser.Result, 
+                cancel, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
         }
 
         private sealed class Releaser : IDisposable
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -79,7 +79,7 @@ namespace Stratis.Bitcoin.Utilities
         public AsyncLock()
         {
             this.semaphore = new SemaphoreSlim(1, 1);
-            this.releaser = Task.FromResult((IDisposable)new Releaser(this));
+            this.releaser = Task.FromResult<IDisposable>(new Releaser(this));
         }
 
         /// <summary>
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -36,7 +36,7 @@ namespace Stratis.Bitcoin.Utilities
     /// </code>
     /// </example>
     /// <remarks>Based on https://www.hanselman.com/blog/ComparingTwoTechniquesInNETAsynchronousCoordinationPrimitives.aspx .</remarks>
-    public sealed class AsyncLock
+    public sealed class AsyncLock : IDisposable
     {
         /// <summary>
         /// Disposable mechanism that is attached to the parent lock and releases it when it is disposed.
@@ -126,6 +126,12 @@ namespace Stratis.Bitcoin.Utilities
 
             // We are holding the lock here, so we will want unlocking.
             return this.releaser.Result;
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            this.semaphore.Dispose();
         }
     }
 }
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -15,7 +15,7 @@ namespace Stratis.Bitcoin.Utilities
     /// </summary>
     /// <example>
     /// <code>
-    /// private asyncLock = new AsyncLock();
+    /// private AsyncLock asyncLock = new AsyncLock();
     /// ...
     /// using (await asyncLock.LockAsync(cancellationToken))
     /// {
--- a/Stratis.Bitcoin/Utilities/AsyncLock.cs
+++ b/Stratis.Bitcoin/Utilities/AsyncLock.cs
@@ -7,13 +7,61 @@ using System.Threading.Tasks;
 namespace Stratis.Bitcoin.Utilities
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


nopara73:
Just another idea to drop here: how about obsolating the problem instead of solving it?  

I noticed Steaphan Cleary's asynclock (AsyncEx nuget library) actually works both blocking and not blocking way, so you can combine them depending on the context, while using Toub's AsyncLock only works in a non-blocking way.

Aprogiena:
That should probably be discussed outside this PR, i.e. if you'd like to suggest using AsyncEx library, pls create an issue for that.

I don't have any experience with that library, so I would be looking for some reason why we think we would need that library. To answering such questions, it would be great to show some existing code in our code base that could be improved easily with AsyncEx, but not so easily without it.

Stratis.Bitcoin/Utilities/AsyncLock.cs
```diff
@@ -7,53 +7,81 @@
...
-            internal Releaser(AsyncLock toRelease)
-            {
-                this.toRelease = toRelease;
-            }
-
-            public void Dispose()
-            {
-                this.toRelease.semaphore.Release();
-            }
+            return wait.ContinueWith((_, state) => (IDisposable)state, this.releaser.Result,
```
dangershony: Can you replace this variable `_`  from this line  `(_, state)` with a better name?

dangershony:
@nopara73 your a natural at writing documentation good job.

Regarding AsyncEx what do you mean by 
>works both blocking and not blocking 

In what context are you refereing to?

nopara73:
@dangershony You can do myLock.LockAsync() or myLock.Lock()  
Like the SemaphoreSlim. You can do WaitAsync() or Wait().

Stratis.Bitcoin/Utilities/AsyncLock.cs
Aprogiena: i was quite OK with having _ for unused thing, but OK, no problem, will fix

Aprogiena:
@nopara73 I've added `Lock()` variant (i.e. non async) to our AsyncLock. Could you pls check that?

Stratis.Bitcoin/Utilities/AsyncLock.cs
```diff
@@ -7,53 +7,105 @@
...
-            {
-                this.toRelease = toRelease;
-            }
-
-            public void Dispose()
-            {
-                this.toRelease.semaphore.Release();
-            }
+            this.semaphore.Wait(cancel);
+            return this.releaser.Result;
```
nopara73: `this.releaser.GetAwaiter().GetResult()` to avoid aggregate exception wrapper.

nopara73:
@Aprogiena I think it's good. Not 100% sure though.

Aprogiena:
Thanks, I have got some ideas why this might be wrong. So pls do not merge yet. I need to test this more carefully.

Aprogiena:
Currently it seems to me that the introduction of the cancellation token broke the lock even in its original version. Will test further tomorrow.

Stratis.Bitcoin/Utilities/AsyncLock.cs
dangershony: '.Wait(cancel)' will throw so why is this an issue?

Aprogiena: i don't think GetAwaiter().GetResult() is useful here, the releaser task is finished (it has always been)
i'll test it thoroughly to be sure

Aprogiena:
introduction of cancellation token make it more complex. we need to throw operation cancelled exception in case the cancellation token triggers, otherwise we would enter the critical section without having the lock and then we would release the lock that we don't own

Stratis.Bitcoin/Utilities/AsyncLock.cs
```diff
@@ -1,59 +1,131 @@
...
-                    this.releaser.Result, cancel,
-                    TaskContinuationOptions.ExecuteSynchronously,
-                    TaskScheduler.Default);
-            }
+                // We only hold the lock if the task was completed successfully.
+                // If the task was cancelled, we don't hold the lock and we need to throw.
+                if (task.IsCanceled) throw new OperationCanceledException();
+
+                return (IDisposable)state;
+            }, this.releaser.Result, cancel, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
```
dangershony: Is there not an overload without `TaskScheduler.Default`

Stratis.Bitcoin/Utilities/AsyncLock.cs
```diff
@@ -1,59 +1,131 @@
...
+        /// <summary>
+        /// Helper object that allows implementation of disposable lock for convenient use with <c>using</c> statement.
+        /// <para>This releaser is used when the lock has been acquired and disposing it will release the lock.</para>
+        /// </summary>
+        /// <remarks>We use the disposable interfaced in a task here to avoid allocations on acquiring the lock when it is free.</remarks>
         private readonly Task<IDisposable> releaser;
 
+        /// <summary>
+        /// Initializes an instance of the object.
+        /// </summary>
```
dangershony: This would be nicer to write
` this.releaser = Task.FromResult<IDisposable>(new Releaser(this));`

Stratis.Bitcoin/Utilities/AsyncLock.cs
```diff
@@ -1,59 +1,131 @@
...
             {
-                return wait.ContinueWith((_, state) =>
-                    (IDisposable)state,
-                    this.releaser.Result, cancel,
-                    TaskContinuationOptions.ExecuteSynchronously,
-                    TaskScheduler.Default);
-            }
+                // We only hold the lock if the task was completed successfully.
+                // If the task was cancelled, we don't hold the lock and we need to throw.
+                if (task.IsCanceled) throw new OperationCanceledException();
```
dangershony: If we reach this point then a lock is already held right? (the `wait` task will not continue until the semaphore acquires a lock).

Stratis.Bitcoin/Utilities/AsyncLock.cs
Aprogiena: if it delete it, it does not work :) i don't know, this works

Stratis.Bitcoin/Utilities/AsyncLock.cs
Aprogiena: ok ok

Stratis.Bitcoin/Utilities/AsyncLock.cs
Aprogiena: no, ContinueWith is executed regardless, so if the task completes successfully, or if task throws, or if cancellation token triggers, in all cases, CW is executed, which means we need to handle this just as the previous case; the task is only guaranteed to be complete here, but we don't know if successfully

Stratis.Bitcoin/Utilities/AsyncLock.cs
Aprogiena: since the original - https://www.hanselman.com/blog/ComparingTwoTechniquesInNETAsynchronousCoordinationPrimitives.aspx also have it, i doubt that

Stratis.Bitcoin/Utilities/AsyncLock.cs
dangershony: Ah right so `if (task.IsCanceled)` then this guarantees the lock is not acquired.

Stratis.Bitcoin/Utilities/AsyncLock.cs
dangershony: Fair enough yeah sometimes the original just tries to do more then needed.

dangershony:
It looks correct to me and we have tests to proof it so green light.

Aprogiena:
one more thing, the semaphore should be disposed

Aprogiena:
now it should be perfect

PR #629, opened 2017-10-21
Feature/use datetimeprovider issue 519

### LINKED ISSUE #519: [DateTime] Find and fix all instances of DateTime/DateTimeOffset being used instead of IDateTimeProvider
For example ` Stratis.Bitcoin.Features.Consensus.CoinViews.BackendPerformanceCounter` is one such class. There are many others.

Aprogiena:
Thought it was low hanging fruit. And it was nightmare :D 

Introduced DateTimeProvider almost everywhere. The remaining cases are couple of static classes and are +- legitimate, or not replaceable now. Also I tried to use DateTime with DateTime and DateTimeOffset with DateTimeOffset, not to mix them, although I believe DateTimeOffset should be completely avoided. I also replaced all `Now`s with `UtcNow`s as I believe non-UTC time should only be used on UI side, not server side or backend side. I haven't touched NBitcoin, however!

Minor polishing included as usual.

Fix #519


dangershony:
You will have to merge it again please

Aprogiena:
Yes, of course. Done that, but it needs a lot of testing and careful review. This was maybe too big task to make it in one go.

dangershony:
yeah and you will continue to accumulate conflicts as this touches so many files, do you think you can easily break it to smaller PRs?

Aprogiena:
not really, we should rather focus on cleaning the PRs we have before advancing further, things are waiting for ages and this will happen not only to this PR but to others, so i think the best solution is to just clean the queue

this PR is so big because a simple change in one component affects so many files, so it is not really easy to make this multiple small PRs

I will fix the conflicts here once PR 492 is merged

Aprogiena:
ok, this should be ready again

dangershony:
Ah crap more conflicts, please fix the conflicts it and I will now review this PR next before anything else.

Aprogiena:
No worries, I've expected this to happen. This is nightmarish PR. So thanks.

Stratis.Bitcoin.Features.Consensus/ConsensusPerformanceCounter.cs
```diff
@@ -189,9 +190,17 @@ public DateTime Taken
...
+
+        /// <summary>
+        /// Initializes an instance of the object.
+        /// </summary>
+        /// <param name="dateTimeProvider">Provider of date time functionality.</param>
+        public ConsensusPerformanceCounter(IDateTimeProvider dateTimeProvider)
 		{
-            this._Start = DateTime.UtcNow;
+            this.dateTimeProvider = dateTimeProvider;
+            this._Start = this.dateTimeProvider.GetUtcNow();
```
dangershony: `this._Start` -> `this.Start`

Aprogiena: these are off topics, we could make many new PRs fixing code style, tabs and comments, but if the old code was bad, it is OK not to fix; it is not OK to introduce new bad

Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -227,8 +228,16 @@ public IEnumerable<UnspentOutputReference> GetAllSpendableTransactions(CoinType
...
     {
-        public AccountRoot()
+        /// <summary>Provider of date time functionality.</summary>
+        private readonly IDateTimeProvider dateTimeProvider;
+
+        /// <summary>
+        /// Initializes a new instance of the wallet.
+        /// </summary>
+        /// <param name="dateTimeProvider">Provider of date time functionality.</param>
+        public AccountRoot(IDateTimeProvider dateTimeProvider)
```
dangershony: In this case I would prefer to not have `IDateTimeProvider` inside `AccountRoot`.
Rather pass the creation date to the constructor.

Stratis.Bitcoin.Features.Wallet/WalletExtensions.cs
```diff
@@ -11,7 +11,7 @@ public static class WalletExtensions
         /// <param name="chain">The chain.</param>        
         public static bool IsDownloaded(this ConcurrentChain chain)
         {
-            return chain.Tip.Header.BlockTime.ToUnixTimeSeconds() > (DateTimeOffset.Now.ToUnixTimeSeconds() - TimeSpan.FromHours(1).TotalSeconds);
+            return chain.Tip.Header.BlockTime.ToUnixTimeSeconds() > (DateTimeOffset.UtcNow.ToUnixTimeSeconds() - TimeSpan.FromHours(1).TotalSeconds);
```
dangershony: Hmm not the best but I guess its an exception

Stratis.Bitcoin.Features.Wallet/Wallet.cs
Aprogiena: yes i agree, i've made such decision elsewhere, but not here, but i don't like it either :)

Stratis.Bitcoin.Features.Wallet/WalletExtensions.cs
Aprogiena: one of 2 or 3 exceptions only, no problem to have it though

Stratis.Bitcoin/Base/TimeSyncBehavior.cs
```diff
@@ -450,7 +450,7 @@ private void AttachedNode_MessageReceived(Node node, IncomingMessage message)
                     VersionPayload version = node.PeerVersion;
                     if (version != null)
                     {
-                        TimeSpan timeOffset = version.Timestamp - this.dateTimeProvider.GetUtcNow();
+                        TimeSpan timeOffset = version.Timestamp - this.dateTimeProvider.GetTimeOffset();
```
dangershony: Was that a bug?

Aprogiena: not really, but i was quite strict here in this PR, to really use the same types everywhere
but in general, i would rather not use DateTimeOffset anywhere, but well ... it is there, so let's be consistent

Stratis.Bitcoin.Features.Consensus/ConsensusPerformanceCounter.cs
dangershony: ok yeah ofcourse

PR #632, opened 2017-10-21
Add witness transaction validation to mempool validator

mikedennis:
This is basically a C# port of this bitcoin logic sections in the mempool validator: https://github.com/bitcoin/bitcoin/blob/ea729d55b4dbd17a53ced474a8457d4759cfb5a5/src/validation.cpp#L584-#L587
https://github.com/bitcoin/bitcoin/blob/aa624b61c928295c27ffbb4d27be582f5aa31b56/src/policy/policy.cpp#L196

As well as some clean up of some async naming issues with some of the mempool tests.

### DIFF
--- a/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -465,7 +465,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
         private async Task SendAsTxInventory(Node node, IEnumerable<uint256> trxList)
         {
             this.logger.LogTrace("({0}:'{1}',{2}.{3}:{4})", nameof(node), node.RemoteSocketEndpoint, nameof(trxList), "trxList.Count", trxList?.Count());
-            Queue<InventoryVector> queue = new Queue<InventoryVector>(trxList.Select(s => new InventoryVector(InventoryType.MSG_TX, s)));
+            Queue<InventoryVector> queue = new Queue<InventoryVector>(trxList.Select(s => new InventoryVector(node.AddSupportedOptions(InventoryType.MSG_TX), s)));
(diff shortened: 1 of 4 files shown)


Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -465,7 +465,7 @@ private Task AttachedNode_MessageReceivedAsync(Node node, IncomingMessage messag
         private async Task SendAsTxInventory(Node node, IEnumerable<uint256> trxList)
         {
             this.logger.LogTrace("({0}:'{1}',{2}.{3}:{4})", nameof(node), node.RemoteSocketEndpoint, nameof(trxList), "trxList.Count", trxList?.Count());
-            Queue<InventoryVector> queue = new Queue<InventoryVector>(trxList.Select(s => new InventoryVector(InventoryType.MSG_TX, s)));
+            Queue<InventoryVector> queue = new Queue<InventoryVector>(trxList.Select(s => new InventoryVector(node.AddSupportedOptions(InventoryType.MSG_TX), s)));
```
dangershony: At this point we could make the Deployment class available to check if segwit is activated, for next PR I guess.

dangershony:
This looks good to me, do you still get rejections based on witness errors?

mikedennis:
Yes I'm still getting those WitnessProgramMissmatch errors and haven't been able to get IsWitnessStandard() except from a unit test. So think there is still something more I need to do.

dangershony:
Shall we push this in any case?

mikedennis:
You can push it - it doesn't seem to have any negative effect on the node as my code is currently ignored :-) Need to get the segwit messaging sorted out with the peers before it will get executed.

dangershony:
Ok fair enough I will push then you can sort out messaging in a new PR

PR #634, opened 2017-10-23
Reduce the amount of outputs when UTXO set is higher then a threshold.

dangershony:


Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -261,10 +262,18 @@ private void AddCoins(TransactionBuildContext context)
                 }
             }
 
+            Money sum = 0;
+            int index = 0;
+            int threshold = 500;
```
Aprogiena: too much magic here

Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -261,10 +262,18 @@ private void AddCoins(TransactionBuildContext context)
...
             var coins = new List<Coin>();
-            foreach (var item in context.UnspentOutputs)
+            foreach (var item in context.UnspentOutputs.OrderByDescending(a => a.Transaction.Amount))
```
bokobza: What this means:
- the coins are ordered by size, the biggest ones are selected first
- there is a max limit of 500 transactions selected

What happens if the 500 biggest transactions don't add up to the amount required?

If anything I would start by using the oldest transactions rather than the biggest one, and stop when we reach the amount, no limit on the number of transactions.
The coin selection algorithm here needs to be carefully decided and well-documented.

bokobza: Sorry I read that wrong, I thought `if (index > threshold && sum > totalToSend)` was an OR condition, not an AND!
`sum > totalToSend` can probably be `sum >= totalToSend`.
My other comments still apply.

dangershony: There is proper coin selection in the TransactionBuilder, this code is just to put a stop limit so not to provide too many UTXO to the trx builder (and slow it down).
I purposely didn't put too much logic in to this code.

Most wallets will not have so many UTXO so normally all the wallets UTXO will go to the coin selection algo.
Yeah biggest first just in case this is a wallet with many UTXO we won't go much beyond the 500 threshold.
500 was randomly selected though, I feel most wallets wont have so many UTXO and the builder can easily cope with that number but we could revisit this number.

Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
dangershony: Do you suggest different names?

Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
bokobza: In this case, let's start with the smallest transactions first. This way the more transactions you send the fewer transactions you have to select.
Also make the threshold a constant, and I would suggest calling it UnspentOutputsSelectionThreshold or something a bit more explicit.
And comments on the different scenarios, so that it feels less magical.

dangershony: I am not sure its right to start from smallest first, this will bloat the tx fee, but this happens in the coins selection algo which is done later , in this case the first 500 UTXO will go to the coin selection algo.

If we do find the wallet bloated with UTXO then I think we want to first send the biggest possible amounts in to the coin selection, as the fee will now be quite big.

bokobza: ah yes I didn't think about the fee. Fair point.

Issue #635, opened 2017-10-24
Exception thrown on closing node


bokobza:
Pressing ctrl+C on a Stratis node running on testnet.
Received this:
```
Application is shutting down...
info: Stratis.Bitcoin.Base.BaseFeature[0]
      FlushAddressManager stopping
info: Stratis.Bitcoin.Base.BaseFeature[0]
      FlushChain stopping
info: Stratis.Bitcoin.FullNode[0]
      BlockStore.DownloadAndStoreBlocks stopping
info: Stratis.Bitcoin.FullNode[0]
      Closing node pending...
info: Stratis.Bitcoin.FullNode[0]
      wallet persist job stopping
info: Stratis.Bitcoin.FullNode[0]
      BlockStore.RelayWorker stopping
info: Stratis.Bitcoin.FullNode[0]
      PeriodicLog stopping
info: Stratis.Bitcoin.FullNode[0]
      MemoryPool.RelayWorker stopping
PS G:\GitHub\StratisBitcoinFullNode\Stratis.StratisD> info: Stratis.Bitcoin.Connection.ConnectionManager[0]
      Node [::ffff:84.192.155.226]:26178 offline
info: Stratis.Bitcoin.Connection.ConnectionManager[0]
      Node [::ffff:51.140.104.135]:26178 offline
info: Stratis.Bitcoin.Connection.ConnectionManager[0]
      Node [::ffff:13.64.76.48]:26178 offline
info: Stratis.Bitcoin.FullNode[0]
      Waiting for wallet persist job to finish.
fail: Stratis.Bitcoin.Features.MemoryPool.MempoolBehavior[0]
      Exception occurred: System.Exception: Assertion failed
         at Stratis.Bitcoin.Utilities.Guard.Assert(Boolean condition) in G:\GitHub\StratisBitcoinFullNode\Stratis.Bitcoin\Utilities\Guard.cs:line 22
         at Stratis.Bitcoin.Features.MemoryPool.MempoolBehavior.<ProcessInvAsync>d__29.MoveNext() in G:\GitHub\StratisBitcoinFullNode\Stratis.Bitcoin.Features.MemoryPool\MempoolBehavior.cs:line 324
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #638, opened 2017-10-25
Replace node guard checks with logging in mempool peer messaging

mikedennis:
#635

### DIFF
--- a/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -262,9 +262,10 @@ namespace Stratis.Bitcoin.Features.MemoryPool
         /// <param name="message">The message payload.</param>
         private async Task SendMempoolPayloadAsync(Node node, MempoolPayload message)
         {
+            Guard.NotNull(node, nameof(node));
             if (node != this.AttachedNode)
             {
-                this.logger.LogError("Node Mismatch ({0}:'{1}', {2}:'{3}')", nameof(node), node?.RemoteSocketEndpoint, nameof(this.AttachedNode), this.AttachedNode?.RemoteSocketEndpoint);
+                this.logger.LogDebug("Node Mismatch ({0}:'{1}', {2}:'{3}')", nameof(node), node.RemoteSocketEndpoint, nameof(this.AttachedNode), this.AttachedNode?.RemoteSocketEndpoint);
                 this.logger.LogTrace("(-)[NODE_MISMATCH]");
                 return;
             }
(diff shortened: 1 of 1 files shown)


bokobza:
Looking at this.

Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -262,7 +262,13 @@ private Task ProcessMessageAsync(Node node, IncomingMessage message)
         /// <param name="message">The message payload.</param>
         private async Task SendMempoolPayloadAsync(Node node, MempoolPayload message)
         {
-            Guard.Assert(node == this.AttachedNode); // just in case
+            if (node != this.AttachedNode)
+            {
+                this.logger.LogError("Node Mismatch ({0}:'{1}', {2}:'{3}')", nameof(node), node?.RemoteSocketEndpoint, nameof(this.AttachedNode), this.AttachedNode?.RemoteSocketEndpoint);
```
bokobza: We should probably have a `Guard.NotNull(node, nameof(node));` in front of all these methods. We can then remove the conditonal access (`node?`).

mikedennis:
Added the guards on node and downgraded the log from Error to Debug which is probably more appropriate.

Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -262,7 +262,14 @@ private Task ProcessMessageAsync(Node node, IncomingMessage message)
...
-            Guard.Assert(node == this.AttachedNode); // just in case
+            Guard.NotNull(node, nameof(node));
+            if (node != this.AttachedNode)
+            {
+                this.logger.LogDebug("Node Mismatch ({0}:'{1}', {2}:'{3}')", nameof(node), node.RemoteSocketEndpoint, nameof(this.AttachedNode), this.AttachedNode?.RemoteSocketEndpoint);
+                this.logger.LogTrace("(-)[NODE_MISMATCH]");
+                return;
+            }
+
             this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(node), node.RemoteSocketEndpoint, nameof(message), message.Command);
```
Aprogiena: the entry log should be preserved just after guard

then the if is OK, but should not be followed by entry log
it should be followed by log line such as
```
this.logger.LogDebug("Attached node '{0}' does not match the originating node '{1}'.", node.RemoteSocketEndpoint, this.AttachedNode?.RemoteSocketEndpoint);
```

same for the code below

mikedennis:
Ok I think this should be good to go now.

PR #643, opened 2017-10-25
Bugfix/issue 636 invalidprevtip

### LINKED ISSUE #636: [ConsensusLoop] PR 609 followup 2 - Sequential consensus logic still not perfect on concurrent events
**Motivation:**

when this if triggered in ConsensusLoop
```
                    if (blockValidationContext.Error == ConsensusErrors.InvalidPrevTip)
                    {
                        this.logger.LogTrace("(-)[INVALID_PREV_TIP]");
                        return;
                    }
```
the chain stopped advancing

**Previous Work:**
Discussion took place here - https://github.com/stratisproject/StratisBitcoinFullNode/pull/609#discussion_r145900436 - the original implementation was resetting consensus tip and chain tip when InvalidPrevTip happened and we argued that this is not the right thing to do, so we changed it not to reset anything. 

That seems to be the right thing to do for the block collision at the tip, but then for next block the problem is that consensus tip is not reset anywhere.

**Scenarios:**
 We have 3 sources of new blocks to consensus
- puller loop
- pow mining
- pos mining

We could possibly reduce this to 2 sources if we disregard pow mining in sidechains, which is possible. Neither in Stratis testnet or mainnet pow mining is no longer possible. But if we find solution that works for 3 sources, why not to use it ...

Notation: 
- block chain is formed with upper cases blocks A -> B -> C-> D -> E, last in the sequence is the tip
- each block has an index `p`, `w` or `s` depending on how we obtained it, so we can have `Ap -> Bp -> Cs -> Dp -> Ew`
- if there are two blocks from different sources at the same height, we add index number: `B1p` vs `B2s` and which one we processed first in AcceptBlock depends on which one more to the left:

```
Ap -> Bp -> C1p        -> Dp
                -> C2s
```
this means that block C from puller came first and then block C from PoS mining arrived (before block D arrived and block D extended C1p)

_Scenario 1a_

```
Ap -> Bp -> C1p        -> Dp
                -> C2s
           
```
here we called AcceptBlock(C1p) and then AcceptBlock(C2s), block C1p was processed, block C2s returned InvalidPrevTip
this situation is currently handled well because InvalidPrevTip does not change puller tip nor consensus tip
therefore the following AcceptBlock(Dp) is processed without problem as Dp's prev hash is our tip C1p (C2s was just discarded)

_Scenario 1b_

```
Ap -> Bp -> C1s        -> Dp
                -> C2p
```
here we called AcceptBlock(C1s) and then AcceptBlock(C2p), block C1s was processed, block C2p returned InvalidPrevTip
this situation is currently handled well because InvalidPrevTip does not change puller tip nor consensus tip
therefore the following AcceptBlock(Dp) is processed without problem as Dp's prev hash is our tip C1s (C2p was just discarded)

_Scenario 2_

```
Ap -> Bp -> C1s
                -> C2p -> Dp
           
```
here we called AcceptBlock(C1s) and then AcceptBlock(C2p) and then AcceptBlock(Dp), block C1s was processed, block C2p returned InvalidPrevTip
and was just discarded, but then Dp arrived, which is now longer chain, but its previous tip is C2p, which is not our tip and hence 
InvalidPrevTip was returned and the block was discarded
this is currently not handled well

_Scenario 3_

```
Ap -> Bp -> C1s        -> D1s 
                -> C2p        -> D2p -> Ep
           
```
Here we mined a block C1s and then received C2p, which was discarded on InvalidPrevTip, then we mined another block on the top of C1s, 
but then D2p arrived form puller and was discarded on InvalidPrevTip again, then Ep came and was discarded
this is currently not handled well

_Scenario 4_

Puller can return null in case it detected reorganization, this happens when 

```
this.Chain.GetBlock(this.location.Height + 1);
```

this returns null and subsequently this

```
if (!this.Chain.Contains(this.location.HashBlock))
```

returns true

that can happen when a new peer provides its headers or when existing peer provides different chain than before

let's call this operation `pR`
the problem is that between the detection of the reorg and the reorg itself in `PullerLoop`, new AcceptBlock can be called from PoS or PoW, thus we can have:

```
Ap -> Bp -> pR_detected        -> pR_executed
                        -> C1s
```

this will result in rewind being made on consensus even if it shouldn't be done and the final state is going to be 

```
Ap -> Bp 
```

this situation is not handled optimally as we lost the block we just mined even if there is no longer chain available (consider the reorg was for `Ap -> B'p -> C'p`)
 

-----

We need to find a solution that at least fixes scenarios 2 and 3 and if possible, fixes 4 as well, although that is not critical. 
And we do not want to disrupt scenarios 1a and 1b (nothing should happen in that case and that is correct). However, it is possible to have alternative resolution for scenario 1a, which makes `C2s` the tip 
(as a preference for our own blocks over remote block. But then the scenario 1b must be handled without any changes of the tip.

Feel free to add more scenarios using the given notation.

Aprogiena:
Everything about this is discussed in #636. 
+ I've created test that previously failed and now passes after the fix.

Fix #636


Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -203,35 +203,26 @@ private void PullerLoop(CancellationToken cancellationToken)
...
                         {
-                            this.logger.LogTrace("No block received from puller due to reorganization, rewinding.");
-                            ChainedBlock lastTip = this.Tip;
-
-                            CancellationToken token = this.nodeLifetime.ApplicationStopping;
+                            this.logger.LogTrace("No block received from puller due to reorganization.");
 
-                            ChainedBlock rewinded = null;
-                            while (rewinded == null)
+                            if (!consensusTip.Equals(this.Tip))
```
dangershony: This condition is to avoid rewinding in case a block was mined in the mean while?

If we assume that `consensusTip` will always be in sync with puller.Tip then this is fine, in case the chain advanced the reorg will just be triggered in the next block.

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -272,6 +296,14 @@ public void AcceptBlock(BlockValidationContext blockValidationContext)
                     // Check if the error is a consensus failure.
                     if (blockValidationContext.Error == ConsensusErrors.InvalidPrevTip)
                     {
+                        if (!this.Chain.Contains(this.Tip.HashBlock))
+                        {
+                            // Our consensus tip is not on the best chain, which means that the current block
+                            // we are processing might be rejected only because of that. The consensus is on wrong chain
+                            // and need to be reset.
+                            RewindCoinViewLocked();
```
dangershony: this.

Aprogiena: of course!

Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: Yes, to avoid rewinding in case a block was delivered from our miner. I.e. I guess puller could return null as for reorg, but the reason was not that was intended in the puller - peer changed the chain, but that our miner added a block.

So I'm just saving the state of consensus before i go to puller, then i check if that changed or not. If not, I know that the reorg is from outside. If yes, then I know that reorg is from inside.

Aprogiena:
Tx, feel free to do the merges yourself after approval.

PR #645, opened 2017-10-26
Add new endpoint in api for estimating transaction fee

mikedennis:
This endpoint is so breeze can do a fee estimate without requiring the user to enter a password. Still need to add unit tests for this.


bokobza:
Great change, nothing to say.

PR #647, opened 2017-10-26
Feature/issue 477 beyond checkpoints

### LINKED ISSUE #477: [Speeding up] Improvements of checkpoints and beyond

Aprogiena:
Optimizes deployment flags, which saves us 1-70ms on each block.
And partially skips block validation, which saves a lot as well.

On review, **please check that nothing important is skipped.**

Fix #477


Stratis.Bitcoin/Base/Deployments/DeploymentFlags.cs
```diff
@@ -22,80 +29,59 @@ public DeploymentFlags(ChainedBlock nextBlock, ThresholdState[] prevBlockStates,
...
+                || (nextBlock.Height == 91880 && nextBlock.HashBlock == new uint256("00000000000743f190a18c5577a3c2d2a1f610ae9601ac046a38084ccb7cd721")));
 
             // Once BIP34 activated it was not possible to create new duplicate coinbases and thus other than starting
             // with the 2 existing duplicate coinbase pairs, not possible to create overwriting txs.  But by the
             // time BIP34 activated, in each of the existing pairs the duplicate coinbase had overwritten the first
             // before the first had been spent.  Since those coinbases are sufficiently buried its no longer possible to create further
             // duplicate transactions descending from the known pairs either.
             // If we're on the known chain at height greater than where BIP34 activated, we can save the db accesses needed for the BIP30 check.
-            var pindexBIP34height = nextBlock.GetAncestor(chainparams.BuriedDeployments[BuriedDeployments.BIP34]);
+            ChainedBlock bip34HeightChainedBlock = chain.GetBlock(chainparams.BuriedDeployments[BuriedDeployments.BIP34]);
```
dangershony: Remember that chain might return null, if this is the case its a reorg and perhaps we need to throw a new `error` that will not trigger invalidation of the block (similar to InvalidPrevHeader)

dangershony:
There are consensus changes here I will have to download the code and check it better, I think you might be skipping too many checks.

Aprogiena:
Agree, this needs very careful testing. I've fully sinced stratis testnet and mainnet from scratch, now testing bitcoin.

Stratis.Bitcoin/Base/Deployments/DeploymentFlags.cs
Aprogiena: hmm, are you sure? I mean it is only used on the next line in the check:
```
            this.EnforceBIP30 = this.EnforceBIP30 && ((bip34HeightChainedBlock == null) || !(bip34HeightChainedBlock.HashBlock == chainparams.BIP34Hash));
```
so the possibility of null is somehow handled. Now this bip34 height is some 227k for bitcoin, something for bitcoin testnet it is 21k and 0 for both stratis networks. So for stratis, this always returns genesis, i.e. the hash is never equal. So only bitcoin network and its testnet are relevant. 

What is happening there? those blocks are well buried, so you would need to be connected to a peer that is on different chain, but not just recent, but like 2 year different. Such node would be banned by everyone on the network. So this is only possible if someone specifically attacks you and owns your connections to the network. Then if you have checkpoints enabled, such a chain would be rejected soon (we've got CP on 250k). So this is concerning only if you don't have checkpoints enabled and you have no chance of connecting to the real network. But in bitcoin you always need to have at least one good connection, once you have that, you have the longest chain, and you have no problems here.

I don't think there is any problem. If there is, please explain.

Aprogiena:
On stratis testnet I was able to mine blocks after the sync.

Stratis.Bitcoin/Base/Deployments/DeploymentFlags.cs
dangershony: Ho I missed the next check for null so this is ok.
But generally try to not relay on specific networks for such checks, that's my point here.
Ideally this would not even be checked in a POS network, at some point I hope to introduce a consensus rules engine where BIP30 rules will only be added to the Bitcoin network rules.

Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -167,20 +166,28 @@ public virtual void ExecuteBlock(ContextInformation context, TaskScheduler taskS
             DeploymentFlags flags = context.Flags;
             UnspentOutputSet view = context.Set;
 
+            int lastCheckpointHeight = this.Checkpoints.GetLastCheckpointHeight();
```
dangershony: I had in mind that this info will go in the Context and so be accessible every where.
But I guess this is fine like this.

PR #654, opened 2017-10-27
Reduce the peers discovery to a smaller number

dangershony:


D3m0nKingx:
Probably not significant but correct spelling should be **Advertise**

Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -149,7 +149,7 @@ public override void Start()
             NodeConnectionParameters connectionParameters = this.connectionManager.Parameters;
             connectionParameters.IsRelay = !this.nodeSettings.ConfigReader.GetOrDefault("blocksonly", false);
             connectionParameters.TemplateBehaviors.Add(new ChainHeadersBehavior(this.chain, this.chainState));
-            connectionParameters.TemplateBehaviors.Add(new AddressManagerBehavior(this.addressManager));
+            connectionParameters.TemplateBehaviors.Add(new AddressManagerBehavior(this.addressManager) { PeersToDiscover = 10 });
```
bokobza: yes but when you do that you're reducing it to all the networks, not just StratisTestnet.
Is it intentional?

Aprogiena:
what is the reason behind this? what does it solve? is any of the configuration switches affected by this or related to this?

Stratis.Bitcoin/Base/BaseFeature.cs
dangershony: yeah I think that's ok for now, there is a task to refactor the discovery away from being dependent on the connection loop (as its now)

dangershony:
@Aprogiena on first time connect addr list is empty so we try to discover peers by connecting tot the seed node, this blocks the connection loop until 1000 potential peers are found (which we don't have) so no connections are made.
This effects only the first time the node starts.

PR #657, opened 2017-10-28
Adding the Breeze agent to the node version

dangershony:
I am considering we should change the default agent from `StratisBitcoin` to be `StratisNode`.  

Alternatively we can also enforce a certain structure in the agent name to `stratis:[node-description]:version`.
Then for the fullnode it will be `stratis:fullnode:1.0.2` and for breeze it will be `stratis:breeze:1.0.2`

We could also put macros to for example for the version, Breeze has its own version this will allow to override version.

### DIFF
--- a/Stratis.Bitcoin/Connection/ConnectionManager.cs
+++ b/Stratis.Bitcoin/Connection/ConnectionManager.cs
@@ -75,8 +75,8 @@ namespace Stratis.Bitcoin.Connection
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
         }
         public void Start()
-		{
-			this.parameters.UserAgent = "StratisBitcoin:" + GetVersion();
+        {
+            this.parameters.UserAgent = $"{this.NodeSettings.Agent}:{this.GetVersion()}";
 			this.parameters.Version = this.NodeSettings.ProtocolVersion;
 			if (this.connectionManagerSettings.Connect.Count == 0)
 			{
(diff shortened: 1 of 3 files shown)


Stratis.BreezeD/Program.cs
```diff
@@ -26,6 +26,7 @@ public static void Main(string[] args)
             var apiUri = args.GetValueOf("apiuri");
             var isTestNet = args.Contains("-testnet");
             var isStratis = args.Contains("stratis");
+            var agent = "Breeze(spv)";
```
bokobza: I think the agent should be "Breeze", no need to specify spv.

Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -75,8 +75,8 @@ public ConnectionManager(Network network, NodeConnectionParameters parameters, N
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
         }
         public void Start()
-		{
-			this.parameters.UserAgent = "StratisBitcoin:" + GetVersion();
+        {
+            this.parameters.UserAgent = $"{this.NodeSettings.Agent}:{this.GetVersion()}";
```
bokobza: please review formatting here.

Stratis.BreezeD/Program.cs
dangershony: I added spv so observers in the network that dont know what breeze is will get an idea of the nodes purpose. (It applies mostly for the bitcoin netwrok).

dangershony: Ok I will change it for now and remove the `spv`

bokobza:
totah

PR #660, opened 2017-10-29
Fix V3041 warnings from PVS-Studio Static Analyzer

AdmAlexus:
Hello. I'm a member of the Pinguem.ru competition on finding errors in
open source projects. A bug, found using PVS-Studio.

Warnings with high priority:

V3041 The expression was implicitly cast from 'int' type to 'double'
type. Consider utilizing an explicit type cast to avoid the loss of a
fractional part. An example: double A = (double)(X) / Y;.
PullerDownloadAssignments.cs 115

### DIFF
--- a/Stratis.Bitcoin/BlockPulling/PullerDownloadAssignments.cs
+++ b/Stratis.Bitcoin/BlockPulling/PullerDownloadAssignments.cs
@@ -112,7 +112,7 @@ namespace Stratis.Bitcoin.BlockPulling
                     int peerTaskCount = filteredPeers[i].TasksAssignedCount;
                     if (peerTaskCount > HighWorkAmountThreshold)
                     {
-                        double penaltyCoef = 1 + ((peerTaskCount * peerTaskCount) / (totalWork * totalWork));
+                        double penaltyCoef = 1 + ((double)(peerTaskCount * peerTaskCount) / (totalWork * totalWork));
                         scores[i] /= penaltyCoef;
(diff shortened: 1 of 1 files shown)


dangershony:
Looks like a cool project thank you very much for contributing.
What is the "5 pieces" that you can win per fix? (I used google translate so I might have not gotten the correct translation)

AdmAlexus:
@dangershony 
Under the terms of the competition - the winner receives PlaStation 4 Pro.
To enter the final of the contest, you need at least 5 correct pool requests.
But I try, in principle, not so much for the PS, but for the sake of improving the code :)

Sry, i'm using google translate too)))

Aprogiena:
I agree, this line should be improved as suggested. I even like to have the explicit casts when they are not needed, just for the code to be more readable. Will fix this.

bokobza:
@Aprogiena can you please approve this PR and then build on top of it? so that @AdmAlexus gets his points? 😋

Aprogiena:
oh omg, sorry :-) I've mistaken this for an issue

AdmAlexus:
Oh, thanks.
If you want, I can, as time will, make a full analysis of the project and throw you a list of errors (which will generate PVS). I'm afraid, without a detailed immersion in the project, to fix something else. And you, perhaps, this list will be useful.

Aprogiena:
Sure, if you have more, just shoot :-) I'm closing this issue now

PR #666, opened 2017-10-30
Bugfix/issue 639 lbp after ibd

### LINKED ISSUE #639: [LookaheadBlockPuller] Behaves oddly after IBD is completed
It seems like after IBD is completed, LBP asks many blocks to be downloaded twice or more and that is even after they have been delivered, which is rather odd. This seems like an efficiency issue, but possibly, there could be memory leaks as well as those blocks that were downloaded and consumed and redownloaded again are no longer consumed and are just waiting to be consumed forever. This last part I have not checked.

Aprogiena:
The puller behaved oddly after IBD, it downloaded blocks twice - see the issue for more description.

The problem was that in `NextBlockCore` this line:
```
                        if (!isDownloading) this.AskBlocks(new ChainedBlock[] { header });
```
was using different way than the usual way of asking new blocks. Normally, `LBP.AskBlocks()` is called and it calls `BP.AskBlocks` and it always adjusts `lookaheadLocation`. But when this shortcut was used, `lookaheadLocation` was not properly modified, which then caused many blocks to be downloaded twice when it was not needed. So what we do is that we make an extra checks in `LBP.AskBlocks()` that fixes `lookaheadLocation` if it is find to be in front of `location`. Alternatively, we could adjust `lookaheadLocation` in the shortcut execution path, but the `lookaheadLocation` fixing logic is already placed in `LBP.AskBlocks` and the logic is not trivial. It thus would be a more complex question on how to adjust `lookaheadLocation` correctly in the shortcut execution path. Therefore we chose this very simple fix by extending the fixing logic.

Fix #639

### DIFF
--- a/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
+++ b/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
@@ -372,13 +372,16 @@ namespace Stratis.Bitcoin.BlockPulling
             if (this.location == null)
                 throw new InvalidOperationException("SetLocation should have been called");
 
-            if (this.lookaheadLocation == null && !this.Chain.Contains(this.location))
+            if ((this.lookaheadLocation != null) && (this.location != null) && (this.lookaheadLocation.Height < this.location.Height))
+                this.lookaheadLocation = this.location;
+
+            if ((this.lookaheadLocation == null) && !this.Chain.Contains(this.location.HashBlock))
             {
                 this.logger.LogTrace("(-)[REORG]");
                 return;
             }
 
-            if (this.lookaheadLocation != null && !this.Chain.Contains(this.lookaheadLocation))
+            if ((this.lookaheadLocation != null) && !this.Chain.Contains(this.lookaheadLocation.HashBlock))
                 this.lookaheadLocation = null;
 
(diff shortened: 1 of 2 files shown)


Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -372,13 +372,16 @@ private void AskBlocks()
             if (this.location == null)
                 throw new InvalidOperationException("SetLocation should have been called");
 
-            if (this.lookaheadLocation == null && !this.Chain.Contains(this.location))
+            if ((this.lookaheadLocation != null) && (this.location != null) && (this.lookaheadLocation.Height < this.location.Height))
```
dangershony: Won't it be more readable to do 
```
if (this.lookaheadLocation?.Height < this.location?.Height)
        this.lookaheadLocation = this.location;
```

dangershony: Just commit if this comment does not apply.

Aprogiena: I'm still undecided about those question marks and double question marks. I think I tend to use them in cases like `loop?.Dispose()`, but in comparison it seems too much for me at the moment. Merged.

PR #670, opened 2017-10-31
 Fix create or restore during chain headers download bug.

bokobza:
If the chain is still downloading when the user creates a wallet, we now wait until it is downloaded in order to set the lastSyncedBlockHeight. Otherwise, the height of the wallet will be the height of the chain at that moment.

### DIFF
--- a/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
+++ b/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
@@ -234,9 +234,18 @@ namespace Stratis.Bitcoin.Features.Wallet
                 account.CreateAddresses(this.network, UnusedAddressesBuffer, true);
             }
 
-            // Update the height of the we start syncing from.
-            this.UpdateLastBlockSyncedHeight(wallet, this.chain.Tip);
-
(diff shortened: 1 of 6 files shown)


Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -234,9 +228,18 @@ public Mnemonic CreateWallet(string password, string name, string passphrase = n
...
+            // If the chain is downloaded, we set the height of the newly created wallet to it.
+            // However, if the chain is still downloading when the user creates a wallet, 
+            // we wait until it is downloaded in order to set it. Otherwise, the height of the wallet will be the height of the chain at that moment.
+            if (this.chain.IsDownloaded())
+            {
+                this.UpdateLastBlockSyncedHeight(wallet, this.chain.Tip);
+            }
+            else
+            {
+                this.UpdateWhenChainDownloaded(wallet, DateTime.Now);
```
dangershony: My only comment would be to use the `IDateTimeProvider` instead of `DateTime.Now`

bokobza:
Thanks. I'll keep a note of the IDateTimeProvider, there are a few places where it can be used.

PR #681, opened 2017-11-02
Refactor/issue 656 dbreeze part 4

### LINKED ISSUE #656: [DBreeze] Refactor all code touching DBreeze database to use it correctly
We are not using DBreeze correctly. There was a general assumption that DBreeze needs a single threaded context. This assumption is wrong. The correct restriction is

> one transaction can be run only in one .NET managed thread and can not be delegated to other threads.

What we currently have is DBreezeSingleThreadSession, which uses a single threaded scheduler and a single DBreeze transaction that is open for the entire app life. 

The correct usage is to create a transaction everytime we want to access the database and to dispose it when our work is done. This allows multithread access to DBreeze database, which we may or may not use.

The restriction mentioned above is enforced by the DBreeze engine itself. It is not necessary to do anything about it if the correct pattern is used, which is:

```
using (Transaction transaction = engine.GetTransaction())
{
    // do something here with the database
}
```

If we want to prevent parallel access, DBreeze has its own mechanism of locking that should be used instead of having our own thing.

Aprogiena:
This is last part of DBreeze refactoring. It removes DBreezeSingleThreadSession and CustomThreadPoolTaskScheduler completely from the codebase. 

Depends on #680 

Fix #656


Stratis.Bitcoin/Utilities/CustomThreadPoolScheduler.cs
```diff
@@ -1,164 +0,0 @@
...
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace Stratis.Bitcoin.Utilities
-{
-    /// <summary>
-    /// Task scheduler that spawns an exact number of threads for execution of its tasks.
-    /// </summary>
-    public class CustomThreadPoolTaskScheduler : TaskScheduler, IDisposable
```
dangershony: I guess we don't need this class anymore!

Aprogiena: hope so :-)

PR #688, opened 2017-11-02
apply suggestions from MSVS

Aprogiena:
This is what MSVS suggest to do in compiler messages (i.e. not errors, not warnings, but lowest level).

### DIFF
--- a/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
+++ b/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
@@ -286,9 +286,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(block), block?.HashBlock);
 
-            var blockRepository = this.BlockRepository as BlockRepository;
-
-            if (blockRepository != null)
+            if (this.BlockRepository is BlockRepository blockRepository)
                 blockRepository.HighestPersistedBlock = block;
 
             this.logger.LogTrace("(-)");
--- a/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
+++ b/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
@@ -46,12 +46,12 @@ namespace Stratis.Bitcoin.BlockPulling
         /// Token that allows cancellation of async tasks.
         /// It is used during component shutdown.
         /// </summary>
-        public CancellationTokenSource CancellationTokenSource => this.cancellationToken;
+        public CancellationTokenSource CancellationTokenSource { get { return this.cancellationToken; } }
 
         /// <summary>Reference to the parent block puller.</summary>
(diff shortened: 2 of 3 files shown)


dangershony:
You couldn't allow 1 night of 0 PRs?

Aprogiena:
i know, sorry :|

Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
```diff
@@ -286,9 +286,7 @@ protected virtual void SetHighestPersistedBlock(ChainedBlock block)
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(block), block?.HashBlock);
 
-            var blockRepository = this.BlockRepository as BlockRepository;
-
-            if (blockRepository != null)
+            if (this.BlockRepository is BlockRepository blockRepository)
```
dangershony: This is a really cool new feature of the framework (I assume its new)

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -46,12 +46,12 @@ public class BlockPullerBehavior : NodeBehavior, IBlockPullerBehavior
         /// Token that allows cancellation of async tasks.
         /// It is used during component shutdown.
         /// </summary>
-        public CancellationTokenSource CancellationTokenSource => this.cancellationToken;
+        public CancellationTokenSource CancellationTokenSource { get { return this.cancellationToken; } }
```
dangershony: This is odd, we normally allow expression properties

Aprogiena: MSVS suggest to do it like this. I mean if we are OK with the `=>` on properties, maybe there is something we can do about the MSVS suggestions so they don't appear.

bokobza:
I'll have a look at this too.

Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
bokobza: I also prefer the `=>` in this case.

Aprogiena: i don't really have a preference except that i like the compiler not to report anything to me, be it error, warning or message

Aprogiena:
so i mean it is perfectly fine for me if you reject this PR
but we should then make MSVS not to make those reports for our projects

bokobza:
yes I'd rather disable the warning in VS.
I'm not rejecting the PR. You can just make this small change and then we can disable these warnings from MSVS.

Aprogiena:
ok, i'll try to find why MSVS suggest those first

bokobza:
I'm gonna push this PR, we can make a rule about which way we want to use, and then either disable the warning or change the code where relevant.

Aprogiena:
can't find any rationale for it, it is called IDE0025

Issue #679, opened 2017-11-01
[ConnectionManager] Application stopped but CM still alive


Aprogiena:
This shouldn't happen:

```
info: Stratis.Bitcoin.Base.BaseFeature[0]
      Flushing headers chain
info: Stratis.Bitcoin.FullNode[0]
      Waiting for FlushChain to finish.

Application stopped.

info: Stratis.Bitcoin.Connection.ConnectionManager[0]
      Node [::ffff:51.141.28.47]:26178 offline
info: Stratis.Bitcoin.Connection.ConnectionManager[0]
      Reason: Unexpected exception while connecting to socket
```

Aprogiena:
I think this issue is related to the crash that I've just experienced on shutdown:
 
```
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext PeriodicLog stopping
[2017-11-03 08:05:30.5915 13] INFO: Stratis.Bitcoin.FullNode.Stop Closing node pending...
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext BlockStore.DownloadAndStoreBlocks stopping
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext PosMining.Mine stopping
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext FlushAddressManager stopping
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext MemoryPool.RelayWorker stopping
[2017-11-03 08:05:30.5915 7] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass12_0+<<StartAsync>b__0>d.MoveNext FlushChain stopping
[...]

PR #692, opened 2017-11-03
Src folder

dangershony:
Moving all the source code in to the src folder.


Aprogiena:
You forgot about appveyor ... but let's say we can fix that later

bokobza:
@dangershony you rolled back NStratis to the 13th October with this change.

Aprogiena:
hmm, that is not necessarily a bad thing after all. it seems like the changes done after that introduced some problems in the Node that some people on slack reported, including for example #695

Aprogiena:
The above seems not entirely correct, at least #679 exists even with the current master. But I can see one race condition being reverted in the code by this.

PR #694, opened 2017-11-04
making consensus loop fully async

### LINKED ISSUE #633: [ConsensusLoop] Make it async
We want to replace lock() with AsyncLock and get rid of GetAwaiter stuff and make the whole thing nice and clean as it should be.

Aprogiena:
Fix #633

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
@@ -86,7 +86,7 @@ namespace Stratis.Bitcoin.Features.Consensus
         /// <summary>A signaler that used to signal messages between features.</summary>
         private readonly Signals.Signals signals;
 
-        /// <summary>A lock object that synchronizes access to the <see cref="ConsensusLoop.AcceptBlock"/> and the reorg part of <see cref="ConsensusLoop.PullerLoop"/> methods.</summary>
+        /// <summary>A lock object that synchronizes access to the <see cref="ConsensusLoop.AcceptBlockAsync"/> and the reorg part of <see cref="ConsensusLoop.PullerLoopAsync"/> methods.</summary>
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -481,22 +480,18 @@ internal void ValidateAndExecuteBlock(ContextInformation context)
...
             this.logger.LogTrace("({0}:{1})", nameof(flags), flags);
 
-            Task prefetching = Task.FromResult<bool>(true);
-
             if (this.UTXOSet is CachedCoinView)
             {
                 Block nextBlock = this.Puller.TryGetLookahead(0);
                 if (nextBlock != null)
-                    prefetching = this.UTXOSet.FetchCoinsAsync(GetIdsToFetch(nextBlock, flags.EnforceBIP30));
+                    await this.UTXOSet.FetchCoinsAsync(GetIdsToFetch(nextBlock, flags.EnforceBIP30)).ConfigureAwait(false);
```
dangershony: We don't want to await here, this should fire and forget.

Aprogiena: the logic you are looking for is there, just not here

the only one who calls `TryPrefetchAsync` is `ValidateAndExecuteBlock`:

```
            // Attempt to load into the cache the next set of UTXO to be validated.
            // The task is not awaited so will not stall main validation process.
            this.TryPrefetchAsync(context.Flags);
```

in this code we do not await and hence `TryPrefetchAsync` is `async void`
but inside `TryPrefetchAsync` we want await as that is no longer the point of entry to the new async task

so it works as you want it to and this is correct i believe

dangershony: ah right that's fine.

PR #697, opened 2017-11-05
Throw exception if prune mode is attempted with mining or staking.

### LINKED ISSUE #622: Staking and mining should depend on the store feature
Currently if the node is in pruned mode staking can still happen.
This is problematic as peers will not be able to ask for the found block.

Staking and mining must depend on the store feature, and make sure the feature is active.

mikedennis:
Fixes #622

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -134,7 +134,7 @@ namespace Stratis.Bitcoin.Features.Miner
             if (this.minerSettings.Mine || this.minerSettings.Stake)
             {
                 services.Features.EnsureFeature<BlockStoreFeature>();
-                var blockSettings = services.ServiceProvider.GetService<StoreSettings>();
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -128,6 +129,15 @@ public override void ValidateDependencies(IFullNodeServiceProvider services)
             {
                 services.Features.EnsureFeature<WalletFeature>();
             }
+
+            // Mining and staking require block store feature.
+            if (this.minerSettings.Mine || this.minerSettings.Stake)
+            {
+                services.Features.EnsureFeature<BlockStoreFeature>();
+                var blockSettings = services.ServiceProvider.GetService<StoreSettings>();
```
Aprogiena: not var here i guess

dangershony:
Excellent, I suspect in core prune mode allows to store the last `Tip - x` blocks and deletes anything before x, this would allow mining in prune mode, however I think we should encourage mining in fullnode mode only.

PR #700, opened 2017-11-05
implements a new list of temporary invalid blocks

### LINKED ISSUE #644: [PosConsensusValidator] Enforcing network split using timing attack
The following only applies for C# node, StratisX is safe.

There are these rules for time stamps in POS consensus validation (and executed in this order during validation flow):
- TimeTooNew - block header timestamp must not be greater than coinbase tx timestamp + 15 
- StakeTimeViolation - block header timestamp must be equal to coinstake tx timestamp & it must be divisible by 16 (represented as unix timestamp)
- BlockTimestampTooEarly - block header timestamp must be at least previous block header timestamp +1
- BlockTimestampTooFar - block header timestamp must not be greater than node's adjusted time +15
- BlockTimeBeforeTrx - each transaction (including coinbase/coinstake) timestamp must be lower than or equal to block header timestamp

If any of those errors occur, AcceptBlock in consensus loop uses `this.chainState.MarkBlockInvalid(rejectedBlockHash);` to mark such a block as invalid and will never sync with the chain based on this block again. 

This can be used by malicious attacker, who can find a new block with timestamp in the future - say 20 seconds from now. The malicious attacker will then propagate this block to 1 big miner (node that is known to have a big stake on the network and create lots of blocks). 

That node will reject the block and mark it as invalid. 

The malicious attacker then wait 6 seconds and propagates the block to the rest of the network, which accepts it (unless another block was found in the meantime). The network will then be split as the big miner (the target of this attack) will not be on the longest chain. Unless the big miner is super big, it is unlikely for it to mine longer chain on its own than the rest of the network. 

The attacker eliminated 1 competitor and can continue to attack another big miner. After a while, it should succeed to eliminate as many big miners as needed for itself to earn more coins.

-------

Solution is quite simple, BlockTimestampTooFar must be treated specially in AcceptBlock and must not lead to marking the block as invalid. The block should be rejected, but the rejection should be just for now and future chains built on the top of this block should be accepted.

Aprogiena:
Based on the last comment in the issue - https://github.com/stratisproject/StratisBitcoinFullNode/issues/644#issuecomment-341978606

We implement a new list of temporary invalid blocks and we use this when the block is rejected with error BlockTimestampTooFar, which means that the block timestamp is too far in the future. So the block can be valid, but not right now. Such a block is set as invalid until its timestamp, but after that it can be considered again and if provided, it will be revalided.

Fix #644


dangershony:
Hmm I am not even sure this is the right approach, the peer that produced such blocks is acting malicious.  
Besides the block is now rejected and most likely the same block hash will never be circulated againcwe should ban the peer for a time period if anything.

Aprogiena:
There is one minor update i need to do here. But, banning the block for just until it can be valid is a right approach. And the peer should be banned as well. The same block will be circulated again as soon as it belongs to the longest chain. That's why we can't ban it forever.

The peer does not need to be acting malicious provided that its clock is just shifted 20 seconds to the future from ours and that can very naturally happen.

Aprogiena:
Maybe i should explain one more thing - the protection here DOES REQUIRE the peer to be disconnected and it could be even temporarily banned. The temp ban on the block is just helping with the whole process - well why we do a list of banned block anyway - to save time on validation when we see them again ... so this is only to help the whole process, but actual banning (to be implemented later) is absolutely necessary.

Aprogiena:
So the last commit is not considering future drift as to calculate when the block could potentially be considered valid. This is more accurate then taking just its block time.

dangershony:
I started working on peer banning so let me go further on that and revisit this PR.

Aprogiena:
To me this PR is still valid. It resolves the attack vector. There is another issue I've described here - https://github.com/stratisproject/StratisBitcoinFullNode/pull/710#issuecomment-343866080 but that should be address on subsequent PR based on this one.

Aprogiena:
ok resolved conflicts

dangershony:
This looks fine to me, it allows to reject based with expiration two point I have.
1. We could one dictionary and only use expiration dictionary, for none expire blocks we set a value fat in the future.
2. Blocks are only removed when asked, so a potential memory leak (it was there before I know).

I am fine committing this as is but perhaps this code should be considered to move in the PeerBanning implementation, we could employ there a background thread that will do all sort of cleaning.

Aprogiena:
i'll prepare this one for merge and i'll rework this little bit as a part of following work, which is #735 

as we don't persist this information on the disk, we can safely do what you suggested and only use one collection with future date (+1 year), i'll do it as a part of #735

Aprogiena:
alright, i've fixed the conflicts, now i need someone to approve, @dangershony ?

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -135,6 +135,8 @@ public override void CheckBlock(ContextInformation context)
             // Check timestamp.
             if (block.Header.Time > this.FutureDrift(this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp()))
             {
+                // The block can be valid only after its time minus the future drift.
+                context.BlockValidationContext.RejectUntil = Utils.UnixTimeToDateTime(block.Header.Time - this.FutureDrift(0)).UtcDateTime;
```
dangershony: `block.Header.Time` is already in UTC no? do we need `.UtcDateTime`

dangershony: Ah `Utils.UnixTimeToDateTime` is DateTimeOffset ignore my comment.

dangershony: I am not sure if this code should be here or in the rejected if condition, putting it here requires to add `RejectUntil` to the context this can be avoided.

Aprogiena: elsewhere we don't have FutureDrift, that's why it is here

dangershony: Ah good point, very well lets push it.

Issue #476, opened 2017-09-20
[Checkpoints] Implement -checkpoints switch that enables/disables checkpoints


Aprogiena:
Currently checkpoints are always enabled. `-checkpoints` config switch could be 1 by default, but it would allow the user not to use checkpoints if they set it to 0.

Aprogiena:
![](https://github.trello.services/images/mini-trello-icon.png) [[Checkpoints] Implement -checkpoints switch that enables/disables checkpoints](https://trello.com/c/UCLmdiAs/248-checkpoints-implement-checkpoints-switch-that-enables-disables-checkpoints)

PR #701, opened 2017-11-05
Add a node setting to control use of checkpoints. 

mikedennis:
Should address issue #476

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
@@ -1,4 +1,5 @@
 ﻿using NBitcoin;
+using Stratis.Bitcoin.Configuration;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -139,9 +140,11 @@ public class Checkpoints : ICheckpoints
         /// Initializes a new instance of the object.
         /// </summary>
         /// <param name="network">Specification of the network the node runs on - regtest/testnet/mainnet/stratis test/main.</param>
-        public Checkpoints(Network network)
+        /// <param name="settings">Settings for full node - used to see if checkpoints have been disabled or not.</param>
+        public Checkpoints(Network network, NodeSettings settings)
         {
-            if (network.Equals(Network.Main)) this.checkpoints = bitcoinMainnetCheckpoints;
+            if (settings != null && !settings.UseCheckpoints) this.checkpoints = new Dictionary<int, CheckpointInfo>();
```
Aprogiena: if you want to make me happy, pls use parentheses like this `if ((settings != null) && !settings.UseCheckpoints)` - i.e. use them even if not needed, so that one does not need to know the operator priority

Issue #257, opened 2017-07-27
Running node with invalid parameters/settings cause crash


Aprogiena:
`dotnet run -datadir=xyz`

if xyz does not exist, the application crashes. This does not sound very nice towards the user. 

In general I think the user should never see exception and app crash even if it is their fault. Similar thing happens when you use -port and the port is already used by another application. The app crashes, and the user sees a lot of information they are probably not interested in, instead of "this port is in use" or "this directory does not exist" etc.

dangershony:
Sounds like a task for the board

bokobza:
Agreed, we should just stop execution and show a helpful message.
https://trello.com/c/82aV0442/192-nodesettings-running-node-with-invalid-parameters-settings-cause-crash

bokobza:
![](https://github.trello.services/images/mini-trello-icon.png) [[NodeSettings] Running node with invalid parameters/settings cause crash](https://trello.com/c/82aV0442/192-nodesettings-running-node-with-invalid-parameters-settings-cause-crash)

Aprogiena:
#707 solves part of this, we still have exceptions when there is port collision (could probably be any kind of port like main port, api port, rpc port)

Aprogiena:
also we could improve on telling the user, which setting was wrong:
`There was a problem in the arguments passed. Details: 'Value was either too large or too small for an Int32.'`

this just tells us that some value was wrong ...

bokobza:
@Aprogiena good points mate. I'm looking at the config folder at the moment, trying to make the settings stuff more robust and more readable.

bokobza:
This whole thing is being refactored by @quantumagi at the moment so I'm closing this.

PR #707, opened 2017-11-06
Avoid app crash on bad command line arguments

bokobza:


bokobza:
https://github.com/stratisproject/StratisBitcoinFullNode/issues/257

PR #710, opened 2017-11-06
Adding PeerLifetime

dangershony:
A first approach on peer banning,
This might not be the final version but rather submitted to discuss the approach.
This code still require unit tests.


src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+        private readonly ConcurrentDictionary<IPAddress, DateTime> banned = new ConcurrentDictionary<IPAddress, DateTime>();
+
+        public void BanPeer(IPEndPoint endPoint, DateTime banUntil)
+        {
+            this.banned.AddOrUpdate(endPoint.Address, banUntil, (point, time) => banUntil);
+        }
+
+        public DateTime? TryGetBanedPeer(IPEndPoint endPoint)
+        {
+            if (this.banned.TryGetValue(endPoint.Address, out DateTime banUntil))
```
bokobza: Can yo not just do: 
```
this.banned.TryGetValue(endPoint.Address, out DateTime banUntil);
return banUntil;
```

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+        public PeerLifetime(IConnectionManager connectionManager, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider)//, IBanStore banStore)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.connectionManager = connectionManager;
+            this.banStore =  new MemoryBanStore();// banStore;
+            this.dateTimeProvider = dateTimeProvider;
+        }
+
+        /// <inheritdoc />
+        public void BanPeer(IPEndPoint endPoint, string reason = null)
```
bokobza: `endPoint` should be `endpoint`.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            Guard.NotNull(endPoint, nameof(endPoint));
+
+            this.logger.LogTrace("()");
+
+            this.banStore.BanPeer(endPoint, this.dateTimeProvider.GetUtcNow().AddDays(7));
+
+            var peer = this.connectionManager.ConnectedNodes.FindByEndpoint(endPoint);
+            if (peer != null)
```
bokobza: I don't really understand the logic here.
So this peer has been banned, but then we check if it's in the whitelist to disconnect. Should you not always disconnect when you ban a peer?

src/Stratis.Bitcoin/Connection/ConnectionManagerBehavior.cs
```diff
@@ -83,12 +77,7 @@ private void AttachedNode_StateChanged(Node node, NodeState oldState)
 
             if ((node.State == NodeState.Failed) || (node.State == NodeState.Offline))
             {
-                this.infoLogger.LogInformation("Node {0} offline.", node.RemoteSocketEndpoint);
-
-                NodeDisconnectReason reason = node.DisconnectReason;
-                if ((reason != null) && !string.IsNullOrEmpty(reason.Reason))
-                    this.infoLogger.LogInformation("Reason: {0}", reason.Reason);
-
+                this.infoLogger.LogInformation("Node {0} offline reason {1}.", node.RemoteSocketEndpoint, node.DisconnectReason?.Reason);
```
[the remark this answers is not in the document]
dangershony: Done.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /// <summary>
+    /// A class that will manage the lifetime of connected peers and be in charge of banning peers that are malicious.
+    /// </summary>
+    public interface IBanStore
+    {
+        void BanPeer(IPEndPoint endPoint, DateTime banUntil);
+
+        DateTime? TryGetBanedPeer(IPEndPoint endPoint);
```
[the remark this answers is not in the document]
dangershony: The peer endpoint? not need for that but it could return a tuple of the two entries.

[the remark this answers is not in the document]
dangershony: Actually no need to return a tuple as the peer is already passe as a parameter.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
dangershony: Yes correct

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        /// <summary>A store of banned peers.</summary>
+        private readonly IBanStore banStore;
+
+        /// <summary>Functionality of date and time.</summary>
+        private readonly IDateTimeProvider dateTimeProvider;
+
+        public PeerLifetime(IConnectionManager connectionManager, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider)//, IBanStore banStore)
```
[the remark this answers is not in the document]
dangershony: Temporary until AddressPeerManager can take over and delete this class.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
dangershony: Done

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
dangershony: Whitelisted are peers we configured to connect to `addnode` they are normally trusted peers.
But you are right I changed it so we don't ban whitelisted peers either.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public bool IsBanned(IPEndPoint endPoint)
+        {
+            Guard.NotNull(endPoint, nameof(endPoint));
+
+            var peerBanedUntil = this.banStore.TryGetBanedPeer(endPoint);
```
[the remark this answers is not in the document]
dangershony: Done

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,109 @@
...
+            if (peerBanedUntil == null)
```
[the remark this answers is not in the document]
dangershony: Done

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            this.logger.LogTrace("()");
+
+            var node = this.AttachedNode;
+            if (node.State == NodeState.Connected)
+            {
+                if (this.peerLifetime.IsBanned(node.RemoteSocketEndpoint))
+                {
+                    this.logger.LogDebug("Node '{0}' was previously banned", node.RemoteSocketEndpoint);
+                    node.DisconnectAsync("A banned node tried to connect");
```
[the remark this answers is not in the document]
dangershony: I think this is anyway traced out.

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        private void AttachedNode_MessageReceived(Node node, IncomingMessage message)
+        {
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(node), node.RemoteSocketEndpoint, nameof(message), message.Message.Command);
+
+            if (this.chainHeadersBehavior.InvalidHeaderReceived && !this.connectionManagerBehavior.Whitelisted)
+            {
+                this.peerLifetime.BanPeer(node.RemoteSocketEndpoint);
+                node.DisconnectAsync("Invalid block received");
```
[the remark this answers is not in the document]
dangershony: I think this will anyway trace the peer endpoint.
But I will add trace.

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            this.logger.LogTrace("(-)");
+        }
+
+        private void AttachedNode_StateChanged(Node node, NodeState oldState)
+        {
+            this.logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(node), node.RemoteSocketEndpoint, nameof(oldState), oldState, nameof(node.State), node.State);
```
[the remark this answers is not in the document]
dangershony: yeah node should always have an instance, its `this.Attachednode` that might be null.

Aprogiena:
conflicts pls

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -160,6 +165,7 @@ public override void Start()
             connectionParameters.IsRelay = !this.nodeSettings.ConfigReader.GetOrDefault("blocksonly", false);
             connectionParameters.TemplateBehaviors.Add(new ChainHeadersBehavior(this.chain, this.chainState, this.loggerFactory));
             connectionParameters.TemplateBehaviors.Add(new AddressManagerBehavior(this.addressManager) { PeersToDiscover = 10 });
+            connectionParameters.TemplateBehaviors.Add(new PeerLifetimeBehavior(this.connectionManager, this.loggerFactory, this.peerLifetime));
```
Aprogiena: does the order of behaviors matter? if so, is CHB, AMB, PLB the right order?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -278,11 +278,11 @@ private void Intercept(IncomingMessage message, Action act)
                     this.pendingTip = chainedPendingTip; 
                 }
 
-                if ((newHeaders.Headers.Count != 0) && (pendingTipBefore.HashBlock != this.GetPendingTipOrChainTip().HashBlock))
+                if ((this.invalidHeaderReceived == false) && (newHeaders.Headers.Count != 0) && (pendingTipBefore.HashBlock != this.GetPendingTipOrChainTip().HashBlock))
```
Aprogiena: elsewhere we use !bool instead of bool == false

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -296,6 +296,8 @@ public NodeSettings()
                 nodeSettings.ConnectionManager.ExternalEndpoint = new IPEndPoint(IPAddress.Loopback, nodeSettings.Network.DefaultPort);
             }
 
+            nodeSettings.ConnectionManager.BanTime = config.GetOrDefault<int>("bantime", ConnectionManagerSettings.DefaultMisbehavingBantime);
```
Aprogiena: if this is a new option we support, it should go to PrintHelp as well

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -8,36 +8,32 @@ namespace Stratis.Bitcoin.Configuration.Settings
...
         /// <summary>External (or public) IP address of the node.</summary>
-        public IPEndPoint ExternalEndpoint
-        {
-            get;
-            internal set;
-        }
+        public IPEndPoint ExternalEndpoint { get; internal set; }
+
+        /// <summary>Number of seconds to keep misbehaving peers from reconnecting.</summary>
+        public int BanTime { get; internal set; }
```
Aprogiena: i tend to have all these time values to reflect their units in the suffix - BanTimeSeconds, DefaultMisbehavingBantimeSeconds etc. what do you think?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -8,36 +8,32 @@ namespace Stratis.Bitcoin.Configuration.Settings
...
+            this.AddNode = new List<IPEndPoint>();
+            this.Listen = new List<NodeServerEndpoint>();
         }
 
         /// <summary>List of exclusive end points that the node should be connected to.</summary>
-        public List<IPEndPoint> Connect
-        {
-            get; set;
-        } = new List<IPEndPoint>();
+        public List<IPEndPoint> Connect { get; set; }
```
Aprogiena: good, can we agree on moving these above the constructor
some kind of order like
- constants 
- static fields
- fields
- auto-properties
- properties
- abstract methods
- static constructor
- constructor
- methods

src/Stratis.Bitcoin/Connection/ConnectionManagerBehavior.cs
```diff
@@ -51,44 +50,26 @@ protected override void AttachCore()
...
-
         private void AttachedNode_StateChanged(Node node, NodeState oldState)
         {
             this.logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(node), node.RemoteSocketEndpoint, nameof(oldState), oldState, nameof(node.State), node.State);
 
             if (node.State == NodeState.HandShaked)
             {
                 this.ConnectionManager.AddConnectedNode(node);
-                this.infoLogger.LogInformation("Node {0} connected ({1}), agent {2}, height {3}", node.RemoteSocketEndpoint, this.Inbound ? "inbound" : "outbound", node.PeerVersion.UserAgent, node.PeerVersion.StartHeight);
+                this.infoLogger.LogInformation("Node '{0}' connected ('{1}'), agent '{2}', height '{3}'", node.RemoteSocketEndpoint, this.Inbound ? "inbound" : "outbound", node.PeerVersion.UserAgent, node.PeerVersion.StartHeight);
```
Aprogiena: this won't be as pretty on the console, esp. `connected ('inbound')` part. also height is a number, does not really belong between apostrophes.

src/Stratis.Bitcoin/Connection/ConnectionManagerBehavior.cs
```diff
@@ -51,44 +50,26 @@ protected override void AttachCore()
...
 
             if ((node.State == NodeState.Failed) || (node.State == NodeState.Offline))
             {
-                this.infoLogger.LogInformation("Node {0} offline.", node.RemoteSocketEndpoint);
-
-                NodeDisconnectReason reason = node.DisconnectReason;
-                if ((reason != null) && !string.IsNullOrEmpty(reason.Reason))
-                    this.infoLogger.LogInformation("Reason: {0}", reason.Reason);
-
+                this.infoLogger.LogInformation("Node '{0}' offline. reason '{1}'.", node.RemoteSocketEndpoint, node.DisconnectReason?.Reason);
```
Aprogiena: offline`,` 

comma instead of dot

Aprogiena: also consider if reason is empty, the console output will contain `reason ''.` - is that what you wanted to have there?

dangershony:
Thanks @Aprogiena good comments I get to this in a bit.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+            this.dateTimeProvider = dateTimeProvider;
+            this.connectionManagerSettings = nodeSettings.ConnectionManager;
+        }
+
+        /// <inheritdoc />
+        public void BanPeer(IPEndPoint endpoint, string reason = null)
+        {
+            Guard.NotNull(endpoint, nameof(endpoint));
+
+            this.logger.LogTrace("()");
```
Aprogiena: `this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(endpoint), endpoint, nameof(reason), reason);`

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+            if (peer != null)
+            {
+                ConnectionManagerBehavior peerBehavior = peer.Behavior<ConnectionManagerBehavior>();
+                if (!peerBehavior.Whitelisted)
+                {
+                    banPeer = false;
+                    this.logger.LogDebug("Peer '{0}' banned for reason: '{1}'", endpoint, reason);
+                    peer.DisconnectAsync($"The peer was banned, reason: {reason}");
+                }
+                else this.logger.LogTrace("Peer '{0}' whitelisted, reason: '{1}', not banned!", reason, endpoint);
```
Aprogiena: `whitelisted for reason` to be consistent with above `banned for reason`

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,97 @@
+using Microsoft.Extensions.Logging;
+using NBitcoin.Protocol;
+using NBitcoin.Protocol.Behaviors;
+using Stratis.Bitcoin.Base;
+
+namespace Stratis.Bitcoin.Connection
+{
+    /// <summary>
+    /// A behaviour that will manage the lifetime of peers.
```
Aprogiena: just note that if we change naming and comments above (i think we should), we need to change this as well

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>Instance of the <see cref="ConnectionManagerBehavior"/> that belongs to the same peer as this behaviour.</summary>
+        private ConnectionManagerBehavior connectionManagerBehavior;
+
+        public PeerLifetimeBehavior(IConnectionManager connectionManager, ILoggerFactory loggerFactory, IPeerLifetime peerLifetime)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.loggerFactory = loggerFactory;
+            this.peerLifetime = peerLifetime;
+            this.ConnectionManager = connectionManager as ConnectionManager;
```
Aprogiena: this is offtopic question, but what is the purpose of this? don't we want to work with the interface?

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <param name="message">The message payload.</param>
+        private void AttachedNode_MessageReceived(Node node, IncomingMessage message)
+        {
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(node), node.RemoteSocketEndpoint, nameof(message), message.Message.Command);
+
+            if (this.chainHeadersBehavior.InvalidHeaderReceived && !this.connectionManagerBehavior.Whitelisted)
+            {
+                this.peerLifetime.BanPeer(node.RemoteSocketEndpoint);
+                node.DisconnectAsync("Invalid block received");
+                this.logger.LogTrace("Invalid block received from peer '{0}'", node.RemoteSocketEndpoint);
```
Aprogiena: dot

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
bokobza: I meant the type returned is not what you'd expect when you look at the name of the method.
So either the method should return a peer object with a Datetime inside or the method should be called TryGetBannedPeerDateTime.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+using System.Collections.Concurrent;
+using System.Net;
+
+namespace Stratis.Bitcoin.Connection
+{
+    /// <summary>
+    /// A class that will manage the lifetime of connected peers and be in charge of banning malicious peers.
+    /// </summary>
+    /// <remarks>
+    /// By default peers are banned for 24 hours, this value can change using configuration (-bantime=[seconds]).
```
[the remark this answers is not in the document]
dangershony: Initially I did have 7 days, but then decided to do like core (which has it at 24h).
Interesting idea of keeping history but is that really necessarily, I would increase the ban time rather then keep history.

Rate limiting in what way? a node that was previously marked is now allowed only a certain amount of blocks?

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        /// <param name="endpoint">The endpoint to set that it was banned.</param>
+        /// <param name="reason">An optional reason for the ban, the 'reason' is only use for tracing.</param>
+        void BanPeer(IPEndPoint endpoint, string reason = null);
+
+        /// <summary>
+        /// Check if a peer is banned.
+        /// </summary>
+        /// <param name="endpoint">The endpoint to check if it was banned.</param>
+        /// <returns><c>True</c>If the peer was banned.</returns>
```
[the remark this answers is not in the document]
dangershony: 🤕

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// A class that will manage the lifetime of connected peers and be in charge of banning peers that are malicious.
+    /// </summary>
+    public interface IBanStore
+    {
+        /// <summary>
+        /// Set a peer endpoint to a banned state.
+        /// </summary>
+        /// <param name="endpoint">The endpoint to set that it was banned.</param>
+        /// <param name="banUntil">The date when the ban will expire.</param>
+        void BanPeer(IPEndPoint endpoint, DateTime banUntil);
```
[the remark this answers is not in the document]
dangershony: As this will be stored in the address manger we will probably never delete addresses, just keep them around.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <returns>The <see cref="DateTime"/> of the end of the ban or null if the peer was not found.</returns>
+        DateTime? TryGetBannedPeer(IPEndPoint endpoint);
+    }
+
+    /// <summary>
+    /// A temporary memory store for banned peers.
+    /// </summary>
+    public class MemoryBanStore : IBanStore
+    {
+        private readonly ConcurrentDictionary<IPAddress, DateTime> banned = new ConcurrentDictionary<IPAddress, DateTime>();
```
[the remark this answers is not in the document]
dangershony: I can but this is temporary code to be replaced by address manager.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+        private readonly IDateTimeProvider dateTimeProvider;
+
+        /// <summary>Functionality of date and time.</summary>
+        private readonly ConnectionManagerSettings connectionManagerSettings;
+
+        public PeerLifetime(IConnectionManager connectionManager, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider, NodeSettings nodeSettings)//, IBanStore banStore)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.connectionManager = connectionManager;
+            this.banStore =  new MemoryBanStore();// banStore;
```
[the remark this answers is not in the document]
dangershony: Yes we will use what @fassadlr is working on now the address store, so that code will just need to implement the interface `IBanStore`

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
Aprogiena: i don't know if it is necessary, but having a one bantime to fit them all seems weird to me. but if we make the design little more flexible in this, we will be able to tweak the protection per every use case. the trade off is little more time invested at the beginning. risk is that we then find it to be an overkill. but in general, if there is a possibility of a node being somehow faulty, but not malicious, we should be able to ban it for short time instead of long time. if we go with 1 week for everything, it might be little too much. tough decision, as always :) i don't mind little more complexity for better flexibility

as for rate limiting - there are many different message types on p2p network and each has different meaning and usage and for each of those types you could have different number of allowed frequency. for example, if someone sends you ping every second, i'd say that is not OK, rather ban that guy after certain threshold. same for every message type. however, i don't know if something like this is not there already (but probably not as we don't have any banning implemented).

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
Aprogiena: to store something forever that by its nature is temporary is always red flag for me 
if the list of peers we met in the history is never deleted, that's imho bad design already in address manager. why would you keep information about nodes you have not have any contact with for more than 1 month?

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
Aprogiena: hmm, i thought it was like memory coin view, i.e. useful for tests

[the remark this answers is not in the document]
Aprogiena: i.e. that it will stay in the code

src/Stratis.Bitcoin/Base/BaseFeature.cs
dangershony: No order does not matter (and should not in my opinion)
In the case of CHA the logic is inside a filter so that gets called before all the other behaviours.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
dangershony: done

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
dangershony: yes done

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
dangershony: Hmm this are public properties, normally above the constructor you have only private and const no?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
dangershony: Yes good idea

src/Stratis.Bitcoin/Connection/ConnectionManagerBehavior.cs
dangershony: Ah I did think this is following your pattern where you add`'{0}'` to all trace outputs.
Sure will fix

Aprogiena: actually not to all of them. this is INFO level, so not tracing at all. optimally (imho) we would not have bound logging with console output, so in that case i would use the notation everywhere, but as we have INFO+ on console, one should rather prefer nice console output instead of following the notation

however, the notation is described here - https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/logging-style.md#value-log-format - this says where to use apostrophes and where not

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
Aprogiena: i don't know what is "normally", we don't have a policy for that :-) so i'm just sharing my personal preference (and also i'm building and refactoring my code as such)

if you want to do it differently, but don't want to have a policy for that, then ignore me

if you want to have a policy for that, let's create issue to discuss that

src/Stratis.Bitcoin/Connection/ConnectionManagerBehavior.cs
dangershony: I will put there `unknown`

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
dangershony: I would say yes if we find a need to ban for for different reasons then we can add that functionality at that point.

As for rate limiting yes I like the idea, this can be a new PR on the PeerLifetimeBehaviour that will listen to all messages and increment a counter, I like that idea see #722

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
dangershony: Well for various reasons, if we are a node that only connects outbound so we only attempt to connect to 8 nodes, how ever peers might still ask as for a collection of address.
But in any case that is for the design of the AddressManager

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
dangershony: hmm didnt think about it, ok yes.

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
```diff
@@ -0,0 +1,145 @@
...
+        public DateTime? TryGetBannedPeer(IPEndPoint endpoint)
+        {
+            this.banned.TryGetValue(endpoint.Address, out DateTime banUntil);
+
+            return banUntil;
+        }
+    }
+
+    /// <summary>
+    /// A class that manages the lifetime of a peer.
```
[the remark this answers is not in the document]
dangershony: fixed

src/Stratis.Bitcoin/Connection/PeerLifetime.cs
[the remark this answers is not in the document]
Aprogiena: for permanent stuff there imho always should be threshold and ordering criterion

so let's say we are talking about these peers we've met. so threshold could be 100 and if we store more than 100 then we should, from time to time, check if there are some peers that we can get rid of - and i would say they should be ordered by time we have actively seen those peers alive 

if you don't like 100, pick any number you like, but there should limit

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly ILoggerFactory loggerFactory;
+
+        /// <summary>Handle the lifetime of a peer.</summary>
+        private readonly IPeerLifetime peerLifetime;
+
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        /// <summary>Instance of the <see cref="ConnectionManager"/>.</summary>
+        public ConnectionManager ConnectionManager { get; private set; }
```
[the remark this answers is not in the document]
dangershony: Ah its not, was used before but not needed anymore.

src/Stratis.Bitcoin/Connection/PeerLifetimeBehavior.cs
dangershony: This is now removed. but yes you are right, however its a different PR I would say.

dangershony:
After this is committed we can apply it to ConsensusLoop and add severity on consensus errors (and mempool errors)

Aprogiena:
Couple of things.

1) BanPeer is working with IPEndPoint, which is imho a bad idea. It should work only with IPAddress. You have it like that in the very first implementation of BanStore there, but I don't think it would make sense anywhere else. If we ban a node on IP X then I think we are very much interested to ban it regardless on which port it exists. This is not only much safer, but it is also necessary for inbound connections.

However, if we are about to use addrman store then I'm not sure if addrman does not use the port information. But then we can create a new persistent store, so no problem.

So I suggest we replace IPEndPoint in all interfaces with just IPAddress.

2) I'm now thinking about how we could use this. You say we can apply it to the ConsensusLoop, but in order to call BanPeer, I need to know the endpoint (hopefully IP address in the future). But ConsensusLoop does not have such information. So we need to extend the BlockValidationContext. However, the puller loop does not pass the context to the puller and the puller does not report which node it received the block from. So we would need to extend the puller to provide this information in result of `NextBlock()`. This would require extending DownloadedBlock, which is OK I guess.

We would also need mining/staking threads to include information about the source to BlockValidationContext, which is also not a problem I think. Just put there loopback or whatever and make an exception that such a thing is not banned even if wrong block is provided.

So that is how we could make this work in ConsensusLoop.

3) We don't only need that in ConsensusLoop, but also in ChainHeadersBehavior and I think we would like to replace invalidHeaderReceived with direct calls to BanPeer. Do you agree? 

What I mean here is that invalidHeaderReceived would be only private field (and thus only used for logic inside CHB) and there would be no public InvalidHeaderReceived. This would lead to removing AttachedNode_MessageReceived handler from PeerBanningBehavior. And we would instead just disconnected the node directly from BanPeer. What do you think? Imho that would be cleaner design.

4) This PR should come with at least one integration test. 

Integration test - I'd say let's create two nodes, both are mining and syncing their chains, then we ban one node from another (only unidirectionally) and we continue to mine on both and we verify that they no longer sync. 

I was also thinking about a unit test for expiration of the banstore entry, but i can see memory ban store does not really implement this. So maybe this should be added when a real implementation of ban store is there.

Aprogiena:
One more thing - with what we have now with list of invalid blocks we've seen, there is a problem that such information does not expire. So it is a kind of memory leak. Once the banning will be in production, an attacker would be able to add one entry to that list of invalid blocks per day per IP address. That might sound trivial, and it indeed is for IPv4 where an attacker with a /24 subnet could cost you 3.5 MB per year in memory. 

However, for IPv6, this is a huge attack vector. I said above we should provide bans per IP address not per end point, but that is still too weak. I think we need to introduce banning per network.

Imagine on IPv6, the attack is assigned something as small as /96 network (and currently it is trivial to get much much more). Suddenly you have a huge problem with single IP banning approach. This can be trivially used to kill a node due to filling all its memory with rubbish block hashes, each from different IP address.

I think we would need some approach that first bans 1 IP address, then if a new ban is added we should check if any other IP address in its subnet of certain size is banned. If it is, we should issue ban on the subnet of certain size. Then if another ban is added and we would detect that a subnet is banned, we should issue ban on even greater subnet. 

This way we would not have a fixed subnet size, which is an approach that does not really work on IPv6, unless you are banning really huge ranges. It would naturally evolve from a single IP address to whatever the attacker has - e.g. /48 subnet on IPv6 and the cost of it would be very very low.

dangershony:
Point 1: yes work with IPEndPointm however the underline dictionary uses IPAddress which would get the desired effect.
Point 2: yes correct we will need to modify the puller, wont be too hard we just need to capture the IPEndPoint of the node that delivered the block an pass it back form `nextblock` alongside the block.
Regarding mining yes correct (didn't think about that one).
Point 3: Ideally yes we can then ban the peer directly form the CHB, we could do that fairly easy (when CHB is loaded we get a copy of PBB.
Removing `AttachedNode_MessageReceived` from PBB yes I didn't think about that good point indeed.

I would put a unit test when we implement the consensus part of banning (the PR that will come after this one).

Regarding the list of invalid blocks, I think a time expiration on such blocks is a good option, I would even actually use a queue of up to say 1000 blocks new banned blocks go to the top of the queue.

Aprogiena:
I would still want to see that integration test. Yes, unit tests to come with later things, but this one should be tested as well.

PR #714, opened 2017-11-07
Moved asynchronous code to functions with Async suffix

noescape00:
Following asynchronous functions naming convention moved async code from Main functions to newly created MainAsync


bokobza:
Thanks @noescape00 !

PR #717, opened 2017-11-08
refactor/nbitcoin-protocol-filters

Aprogiena:
code style cleanup in NBitcoin/Protocol/Filters


src/NBitcoin/Protocol/Filters/ActionFilter.cs
```diff
@@ -1,34 +1,26 @@
-﻿#if !NOSOCKET
+﻿using NBitcoin.Protocol;
```
bokobza: System namespace should go first.

PR #720, opened 2017-11-08
Refactor/nbitcoin protocol

Aprogiena:
code cleanup in NBitcoin.Protocol

### DIFF
--- a/src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
+++ b/src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
@@ -1,10 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-
-namespace NBitcoin.Protocol
+﻿namespace NBitcoin.Protocol
 {
 	public class BitcoinSerializablePayload<T> : Payload where T : IBitcoinSerializable, new()
 	{
--- a/src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
+++ b/src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
@@ -1,30 +1,22 @@
 ﻿namespace NBitcoin.Protocol
 {
-	public class BitcoinSerializablePayload<T> : Payload where T : IBitcoinSerializable, new()
-	{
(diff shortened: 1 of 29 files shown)


src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
```diff
@@ -1,30 +1,22 @@
 ﻿namespace NBitcoin.Protocol
 {
-	public class BitcoinSerializablePayload<T> : Payload where T : IBitcoinSerializable, new()
-	{
-		public BitcoinSerializablePayload()
-		{
+    public class BitcoinSerializablePayload<T> : Payload where T : IBitcoinSerializable, new()
+    {
+        private T @object = new T();
```
bokobza: Can we just have a property that's initialized in the constructor?
Also, please let's have a different name, like "o to the b to the j etc".

src/NBitcoin/Protocol/InventoryVector.cs
```diff
@@ -22,56 +16,48 @@ public enum InventoryType : uint
...
+            }
+            set
+            {
+                this.type = (uint)value;
+            }
+        }
 
-		public InventoryVector()
-		{
+        private uint256 hash = uint256.Zero;
```
bokobza: same here, can we have the property defined like we like (without the backing field)?

bokobza:
@Aprogiena is there 0 functionality change here?
My only comment would be to make sure you remove backing fields wherever possible.

src/NBitcoin/Protocol/BitcoinSerializablePayload.cs
Aprogiena: i don't like it either :-) but being inside these classes i was afraid of changing something too much, but it seems like this is safe, so i've changed it now, it did propagate to quite many files though :/

Aprogiena:
this should be 99.99% equivalent code. the most different stuff i've done so far is to replace Task.Factory.StartNew with Task.Run

src/NBitcoin/Protocol/InventoryVector.cs
dangershony: Yeah you can't change it, `ref` will work on the same copy that was passed to the method.
This allows to use the serializer as both read and write

PR #723, opened 2017-11-08
Refactor/chain related classes

Aprogiena:
code cleanup of NBitcoin classes that are closesly related to concurrent chain

### DIFF
--- a/src/NBitcoin/Block.cs
+++ b/src/NBitcoin/Block.cs
@@ -2,9 +2,7 @@
 using NBitcoin.Crypto;
 using NBitcoin.DataEncoders;
 using NBitcoin.RPC;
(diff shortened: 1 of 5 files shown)


src/NBitcoin/Block.cs
```diff
@@ -1,550 +1,493 @@
...
+        {
+            return GetHash().ToString();
+        }
+
+        /// <summary>
+        /// Set time to consensus acceptable value.
+        /// </summary>
+        /// <param name="network">Network.</param>
+        /// <param name="prev">Previous block.</param>
+        public void UpdateTime(Network network, ChainedBlock prev)
```
dangershony: I think this two methods are never used `UpdateTime` maybe delete them?

PR #724, opened 2017-11-09
Bugfix: dependency check for features

noescape00:
1) Changed BuilderExtension class names to follow same naming format (they were named in 2 different formats)
2) Removed partial from all BuilderExtension classes since almost all of them were not partial because they were in different assemblies and because of the change made in 1)
3) Added dependency check to all BuilderExtension that have dependencies- it now fixes the bug when node attempted to add feature that depends on another feature\features that were not added to node. 

More about 3) :
When feature is added feature's services initialize first, then actual feature initializes and after that ValidateDependencies gets called. The problem is that some of the feature's services initialization rely on other services that get added by other features. 

Example: if we disable consensus mempool will add MempoolValidator that need PowConsensusValidator for the creation and PowConsensusValidator is a part of consensus so this causes exceptions and cansellation of node building. 

So I added the check before we add and init the feature that can possibly have missing dependency.


dangershony:
Interesting, I will review this.
So you say that the current validate dependency method does not work?
If I'm not mistaken services will not get resolved until they are called.

src/Stratis.Bitcoin.Features.IndexStore/IndexStoreFeature.cs
```diff
@@ -28,10 +31,24 @@ public override BlockStoreBehavior BlockStoreBehaviorFactory()
         }
     }
 
-    public static class IndexStoreBuilderExtension
+    public static class FullNodeBuilderIndexStoreExtension
     {
         public static IFullNodeBuilder UseIndexStore(this IFullNodeBuilder fullNodeBuilder, Action<IndexSettings> setup = null)
         {
+            try
```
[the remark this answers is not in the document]
noescape00: >If I'm not mistaken services will not get resolved until they are called.

Correct. Problem is that they get called for the feature creation that should actually perform the dependency check after it is created. So in case we use mempool without consensus that's what happening: 

Node gets builded without consensus, then node.Start() => fullNodeFeatureExecutor.Start(); which calls feature constructor that need MempoolBehavior feature. And MempoolBehavior needs MempoolOrphans which needs Mempoolvalidator which needs PowConsensusValidator which is a part of consensus feature aaand it fails.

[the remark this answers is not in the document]
noescape00: >This is probably not the right way to check dependencies, if doing it inside the feature is not good its should probably be done some where inside he builder itself.

Ok, I'll remake the PR. 
Also I guess I need to remove old dependency checking (since it gets called too late) and move it's usages to a new dependency checking

dangershony:
The problem here is that the validate dependencies requires the feature to be initialised right?
We need a new approach, to be able to validate dependencies before the feature is initialised.  

I propose we allow a static method in the feature that the builder, using reflection, will look for. if found it will be invoked and take a the list of feature types (before the feature is initialised) then it will be properly handled.

So such a method on a feature
```
        public static void ValidateDependencies(IFullNodeServiceProvider services)
        {
            services.Features.EnsureFeature<ConsensusFeature>();
        }
```

Then `IFullNodeServiceProvider` will need to be able to return a list of `FeatureTypes` you have it available as a private list so just expose a public get I would say.
```
    public class FullNodeServiceProvider : IFullNodeServiceProvider
    {
        /// <summary>List of registered feature types.</summary>
        private readonly List<Type> featureTypes;
```

dangershony:
Another approach is something like this
```
                 features
                .AddFeature<MempoolFeature>()
                .DependOn<ConsensusFeaure>()
                .DependOn<StoreFeaure>().
                .FeatureServices(services =>
                    {
                        services.AddSingleton<MempoolSchedulerLock>();
                        services.AddSingleton<TxMempool>();
                        services.AddSingleton<BlockPolicyEstimator>();
```

bokobza:
Personally I prefer this option:
```
 features
                .AddFeature<MempoolFeature>()
                .DependOn<ConsensusFeaure>()
```

because this way you don't need to have an extra static method in every feature class and it's more 'flowy'.

dangershony:
Can you fix merge conflicts?

src/Stratis.Bitcoin/Builder/FullNodeBuilder.cs
```diff
@@ -201,6 +202,33 @@ private IServiceCollection BuildServices()
             foreach (var configureFeature in this.featuresRegistrationDelegates)
                 configureFeature(this.Features);
 
+            // unregister features that depend on not registered features
+            while (true)
```
dangershony: I don't think you need this  loop (especially not a while(true) ) just foreach the feature and call `EnsureDependencies` we should not continue if any of the dependencies is broken registering is not needed.

noescape00:
>just foreach the feature and call EnsureDependencies we should not continue if any of the dependencies is broken registering is not needed.
>Can you fix merge conflicts?

Conflicts fixed & replaced that loop

noescape00:
Though without the old loop there is a possibility that app will fire unhanded exception in case 2 or more features missing (feature A depends on disabled feature B so building A will be prevented by dependency checking but if there is feature C that depends on A it will not be caught until the node starts). Though, first red message will always be from dependency checker so it will be clear what is the source of the problem, so I guess ok, just foreach is enough.

src/Stratis.Bitcoin.Features.IndexStore/IndexStoreFeature.cs
```diff
@@ -0,0 +1,61 @@
...
+using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Features.BlockStore;
+using Stratis.Bitcoin.Features.MemoryPool;
+using Stratis.Bitcoin.Features.RPC;
+using Stratis.Bitcoin.Features.RPC.Controllers;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.IndexStore
+{
+    public class IndexStoreFeature : BlockStoreFeature
```
dangershony: `IndexStoreFeature` was deleted from this repo and moved to the BlockExplorer repo, I think you didn't merge correctly.

src/Stratis.Bitcoin.Features.IndexStore/IndexStoreFeature.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public static class FullNodeBuilderIndexStoreExtension
+    {
+        public static IFullNodeBuilder UseIndexStore(this IFullNodeBuilder fullNodeBuilder, Action<IndexSettings> setup = null)
+        {
+            fullNodeBuilder.ConfigureFeature(features =>
+            {
+                features
+                .AddFeature<IndexStoreFeature>()
```
dangershony: You will need to create a second PR in the BlockExplorer repo to add this code.

src/Stratis.Bitcoin.Features.RPC/RPCFeature.cs
```diff
@@ -89,6 +90,7 @@ public static IFullNodeBuilder AddRPC(this IFullNodeBuilder fullNodeBuilder, Act
             {
                 features
                 .AddFeature<RPCFeature>()
+                .DependOn<MempoolFeature>()
```
dangershony: Hmm its odd ther eis a dependency on mempool here

noescape00: Yeah, fixed it. 
RPC feature project had mempool project in dependency projects list so I added mempool. But since mempool depend on consensus there was no bug. 

I've removed mempool project dependency and added consensus + changed depend on to depend on consensus in the latest commit.

dangershony:
Are all the tests passing?

src/Stratis.Bitcoin.Features.RPC/Stratis.Bitcoin.Features.RPC.csproj
```diff
@@ -35,7 +35,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\NBitcoin\NBitcoin.csproj" />
-    <ProjectReference Include="..\Stratis.Bitcoin.Features.MemoryPool\Stratis.Bitcoin.Features.MemoryPool.csproj" />
+    <ProjectReference Include="..\Stratis.Bitcoin.Features.Consensus\Stratis.Bitcoin.Features.Consensus.csproj" />
```
dangershony: This is still wrong but leave it for now. we can remove the dependency on consensus it in another PR.

src/Stratis.Bitcoin/Builder/FullNodeBuilder.cs
```diff
@@ -203,7 +204,22 @@ private IServiceCollection BuildServices()
...
             // configure features startup
             foreach (var featureRegistration in this.Features.FeatureRegistrations)
+            {
+                try
+                {
+                    featureRegistration.EnsureDependencies(this.Features.FeatureRegistrations);
+                }
+                catch (MissingDependencyException e)
+                {
+                    var logger = this.NodeSettings.LoggerFactory.CreateLogger(typeof(FullNodeBuilder).FullName);
```
dangershony: Ah FYI settings has a logger already so you can just do 
`this.NodeSettings.Logger`

src/Stratis.Bitcoin/Builder/FullNodeBuilder.cs
```diff
@@ -203,7 +204,22 @@ private IServiceCollection BuildServices()
...
+                    logger.LogCritical("Feature {0} cannot be configured because it depends on other features that were not registered: {1}",
+                        featureRegistration.FeatureType.Name, e.Message);
+
+                    continue;
```
dangershony: `continue;` will just continue to execution of the code no? don't we want to throw at this point?

noescape00: Yes. The idea behind it was to run node without features that have broken dependencies. But I guess you right, so I'll change it to throw forward caught exception.

noescape00:
>You will need to create a second PR in the BlockExplorer repo to add this code.
Sure. I'll wait till nuget packages get updated to include this PR's code and then add PR in the BlockExplorer repo.

PR #727, opened 2017-11-09
Fix network

dangershony:


src/NBitcoin/Block.cs
```diff
@@ -1,12 +1,12 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using NBitcoin.BouncyCastle.Math;
+﻿using NBitcoin.BouncyCastle.Math;
```
bokobza: So we agreed Systems namespaces first, then the other namespaces.

src/NBitcoin/NetworkBuilder.cs
```diff
@@ -1,141 +1,143 @@
...
-		/// </summary>
-		/// <returns></returns>
-		public Network BuildAndRegister()
-		{
-			return Network.Register(this);
-		}
-	}
+    public class NetworkBuilder
+    {
+        internal string Name;
```
bokobza: According to rule 3 of our coding style doc, 

> We use camelCase for internal and private fields and use readonly where possible.

src/NBitcoin/NetworkBuilder.cs
```diff
@@ -1,141 +1,143 @@
...
+        public NetworkBuilder()
+        {
+            this.Base58Prefixes = new Dictionary<Base58Type, byte[]>();
+            this.Bech32Prefixes = new Dictionary<Bech32Type, Bech32Encoder>();
+            this.Aliases = new List<string>();
+            this.Seeds = new List<DNSSeedData>();
+            this.FixedSeeds = new List<NetworkAddress>();
+        }
+
+        public NetworkBuilder SetName(string name)
```
bokobza: very Java looking. Do we need this?
I see now that we use this construct so that we can chain the construction of NetworkBuilder.
It doesn't seem to give any advantage over a normal object construction.
Can we get rid of that and just use properties instead of internal fields?

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            {
+                System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+            }
+
+            return this.addresses;
+        }
+
+        public override string ToString()
+        {
+            return this.Name + " (" + this.Host + ")";
```
bokobza: Can even do a bit of string extrapolation here.

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+                return lastPOWBlock;
+            }
+            set
+            {
+                EnsureNotFrozen();
+                lastPOWBlock = value;
+            }
+        }
+
+        private bool frozen = false;
```
bokobza: Do we need to set this to false?

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+                }
+            }
+        }
+        public class BIP9DeploymentsArray
+        {
+            Consensus _Parent;
+            BIP9DeploymentsParameters[] _Parameters;
+            public BIP9DeploymentsArray(Consensus parent)
+            {
+                _Parent = parent;
```
bokobza: No underscores allowed in this town.

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            consensus.powTargetSpacing = this.powTargetSpacing;
+            consensus.powTargetTimespan = this.powTargetTimespan;
+            consensus.ruleChangeActivationThreshold = this.ruleChangeActivationThreshold;
+            consensus.subsidyHalvingInterval = this.subsidyHalvingInterval;
+            consensus.minimumChainWork = this.minimumChainWork;
+            consensus.GetPoWHash = this.GetPoWHash;
+            consensus.coinType = this.CoinType;
+            consensus.litecoinWorkCalculation = this.litecoinWorkCalculation;
+            consensus.LastPOWBlock = this.LastPOWBlock;
+            consensus.ProofOfStakeLimit =
```
bokobza: Move to ctor.

bokobza: Can we just make consensus a property instead of a field?

bokobza:
Ok that's just a quick review.
I think if you fix these, we can then push it and we can add another coat right after that, there is just too much to change otherwise.

src/NBitcoin/Network.cs
dangershony: No but I am actually tempted to remove that logic all together.

src/NBitcoin/Block.cs
dangershony: Ah right I need to configure the vs tools to do that

src/NBitcoin/Network.cs
dangershony: Perhaps yeah but maybe in a later PR?
Network will move to the FN at some point, and consensus might change as well.

src/NBitcoin/NetworkBuilder.cs
dangershony: Method chaining? I find it useful.

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            if (this.addresses != null)
+            {
+                return this.addresses;
+            }
+
+            try
+            {
+                this.addresses = Dns.GetHostAddressesAsync(this.Host).Result;
+            }
+            catch(AggregateException ex)
```
Aprogiena: when fixing code style, i always do delete last `}` in the file and then write it there again. MSVS will format the whole code of the file as per its rules. It should change this line to `catch (AggregateException ex)` - i.e. insert the space

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+        STEALTH_ADDRESS,
+        ASSET_ID,
+        COLORED_ADDRESS,
+        MAX_BASE58_TYPES,
+    };
+
+    public enum Bech32Type
+    {
+        WITNESS_PUBKEY_ADDRESS,
+        WITNESS_SCRIPT_ADDRESS
```
Aprogiena: there comes the question - do we want to modify these to comply with our style? XYZ_ABC -> XyzAbc? We probably should ...

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+    }
+
+    public partial class Network
+    {
+        internal byte[][] base58Prefixes = new byte[12][];
+        internal Bech32Encoder[] bech32Encoders = new Bech32Encoder[2];
+        public Bech32Encoder GetBech32Encoder(Bech32Type type, bool throws)
+        {
+            var encoder = bech32Encoders[(int)type];
```
Aprogiena: unvar

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            if(encoder == null && throws)
```
Aprogiena: again all these missing spaces would be fixed with automatic reformat

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            return prefix?.ToArray();
+        }
+
+        internal static string CreateBase58(Base58Type type, byte[] bytes, Network network)
+        {
+            if(network == null)
+                throw new ArgumentNullException("network");
+            if(bytes == null)
+                throw new ArgumentNullException("bytes");
+            var versionBytes = network.GetVersionBytes(type, true);
```
Aprogiena: unvar unvar everywhere :)

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            return Encoders.Base58Check.EncodeData(versionBytes.Concat(bytes));
+        }
+
+        internal static string CreateBech32(Bech32Type type, byte[] bytes, byte witnessVersion, Network network)
+        {
+            if(network == null)
+                throw new ArgumentNullException("network");
+            if(bytes == null)
+                throw new ArgumentNullException("bytes");
+            var encoder = network.GetBech32Encoder(type, true);
```
Aprogiena: ...

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+                    return _Heights[(int)index];
+                }
+                set
+                {
+                    _Parent.EnsureNotFrozen();
+                    _Heights[(int)index] = value;
+                }
+            }
+        }
+        public class BIP9DeploymentsArray
```
Aprogiena: add empty line here

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+        /// </summary>
+        public class ConsensusOptions
+        {
+        }
+
+        public ConsensusOptions Options { get; set; }
+
+        public class BuriedDeploymentsArray
+        {
+            Consensus _Parent;
```
Aprogiena: add access modifier and rename these _ stuffs

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+                return this.subsidyHalvingInterval;
+            }
+            set
+            {
+                EnsureNotFrozen();
+                this.subsidyHalvingInterval = value;
+            }
+        }
+
+        private Func<BlockHeader, uint256> getPoWHash = h => h.GetHash();
```
Aprogiena: can't we just move this implementation to the public GetPoWHash, we don't need this whatever it is

src/NBitcoin/Network.cs
```diff
@@ -10,1429 +12,1079 @@
...
+            }
+            set
+            {
+                EnsureNotFrozen();
+                this.getPoWHash = value;
+            }
+        }
+
+
+        int majorityEnforceBlockUpgrade;
```
Aprogiena: access

Aprogiena: everywhere ...

src/NBitcoin/Network.cs
Aprogiena: do it! suddenly you would get rid of the fields, would be much better

src/NBitcoin/Network.cs
Aprogiena: that should probably be just initialized in a constructor and be { get; private set; } without having a property and field separated

Aprogiena: we do these initializations in constructor
and then you can get rid of those separated public properties

Aprogiena: whole block of unvars

Aprogiena: everywhere unvar, spaces ...

src/NBitcoin/NetworkBuilder.cs
Aprogiena: ah, i may have violated that as well
the rule seems inconsistent towards `protected` modifier, as `internal` seems stronger than protected ...

src/NBitcoin/NetworkBuilder.cs
```diff
@@ -1,141 +1,143 @@
...
+    {
+        internal string Name;
+        internal Dictionary<Base58Type, byte[]> Base58Prefixes;
+        internal Dictionary<Bech32Type, Bech32Encoder> Bech32Prefixes;
+        internal List<string> Aliases;
+        internal int RPCPort;
+        internal int Port;
+        internal uint Magic;
+        internal Consensus Consensus;
+        internal List<DNSSeedData> Seeds ;
```
Aprogiena: no space

src/NBitcoin/Networks.cs
```diff
@@ -1,119 +1,262 @@
...
...33", "202.60.69.232:8333", "202.124.109.103:8333", "203.30.197.77:8333", "203.88.160.43:8333", "203.151.140.14:8333", "203.219.14.204:8333", "205.147.40.62:8333", "207.235.39.214:8333", "207.244.73.8:8333", "208.12.64.225:8333", "208.76.200.200:8333", "209.40.96.121:8333", "209.126.107.176:8333", "209.141.40.149:8333", "209.190.75.59:8333", "209.208.111.142:8333", "210.54.34.164:8333", "211.72.66.229:8333", "212.51.144.42:8333", "212.112.33.157:8333", "212.116.72.63:8333", "212.126.14.122:8333", "213.66.205.194:8333", "213.111.196.21:8333", "213.122.107.102:8333", "213.136.75.175:8333", "213.155.7.24:8333", "213.163.64.31:8333", "213.163.64.208:8333", "213.165.86.136:8333", "213.184.8.22:8333", "216.15.78.182:8333", "216.55.143.154:8333", "216.115.235.32:8333", "216.126.226.166:8333", "216.145.67.87:8333", "216.169.141.169:8333", "216.249.92.230:8333", "216.250.138.230:8333", "217.20.171.43:8333", "217.23.2.71:8333", "217.23.2.242:8333", "217.25.9.76:8333", "217.40.226.169:8333", "217.123.98.9:8333", "217.155.36.62:8333", "217.172.32.18:20993", "218.61.196.202:8333", "218.231.205.41:8333", "220.233.77.200:8333", "223.18.226.85:8333", "223.197.203.82:8333", "223.255.166.142:8333" };
```
Aprogiena: just a thought - when was the last time someone actually checked this list of end points

src/NBitcoin/Networks.cs
```diff
@@ -137,192 +280,219 @@ private static Network InitStratisMain()
...
+            consensus.HashGenesisBlock = genesis.GetHash();
 
-			// The message start string is designed to be unlikely to occur in normal data.
-			// The characters are rarely used upper ASCII, not valid as UTF-8, and produce
-			// a large 4-byte int at any alignment.
-			var pchMessageStart = new byte[4];
+            // The message start string is designed to be unlikely to occur in normal data.
+            // The characters are rarely used upper ASCII, not valid as UTF-8, and produce
+            // a large 4-byte int at any alignment.
+            var pchMessageStart = new byte[4];
```
Aprogiena: when possible, get rid of this pch prefix (or other hungarian whatever prefixes)

dangershony:
As this is still not moving to the FN (its still in the NBitcoin project) and there will be more refactoring here I don't mind its not fully to standard.
I will act on some of the reviews.

src/NBitcoin/Network.cs
bokobza: Did you see my review? You're saying a few things I've already said :-)

Aprogiena: oh sorry, didn't :)

bokobza:
@Aprogiena in your last comment I see we arrived at the same conclusion lol.

src/NBitcoin/Network.cs
dangershony: ok I will

src/NBitcoin/NetworkBuilder.cs
dangershony: Ah crap it is actually a public field internal to the project, Jeremy look away.

src/NBitcoin/Network.cs
dangershony: Not yet, making this change will effect many files.

src/NBitcoin/Network.cs
dangershony: 🤕

src/NBitcoin/Network.cs
dangershony: Yes but I am not sure yet how we will change that so preferred not to change functionality.

src/NBitcoin/NetworkBuilder.cs
dangershony: Fine for now when we add documentation we can add spaces.

src/NBitcoin/Networks.cs
dangershony: This is taken from core, it might be outdated we could also copy over the current list from core (if changed)

src/NBitcoin/Networks.cs
dangershony: ah yea done

PR #730, opened 2017-11-10
Node Settings Winter Remix part 1

bokobza:
The way settings are set is a bit hard to follow so I'm trying to to sort them out.
With the speed at which things are evolving, I'm pushing a first pull request and will carry on with the rest of the work after it's pushed.


bokobza:
Whoever takes this change, do it commit by commit as they're quite atomic, very little is done in each.

PR #733, opened 2017-11-11
Add assumevalid flag for faster IBD

### LINKED ISSUE #22: Feature: Add the assumevalid flag
https://github.com/bitcoin/bitcoin/commit/e440ac7ef3b6f3ad1cd8fc7027cece40413202d9

mikedennis:
Fixes #22


src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -231,6 +237,19 @@ public static NodeSettings Default(Network network = null, ProtocolVersion proto
...
             nodeSettings.Logger.LogDebug("Checkpoints are {0}.", nodeSettings.UseCheckpoints ? "enabled" : "disabled");
 
+            // TODO: Figure out where to put defaults so they are easy to configure.
+            uint256 defaultAssumeValid = null;
+            if (!nodeSettings.Network.IsBitcoin() && nodeSettings.Network.IsTest())
+            {
+                // Block Height 184096 https://chainz.cryptoid.info/strat-test/block.dws?184096.htm
+                defaultAssumeValid = new uint256("0x74427b2f85b5d9658ee81f7e73526441311122f2b23702b794be557ba43ca43e");
+            }
+            // TODO: Add defaults for bitcoin and mainnet networks.
```
mikedennis: These default settings should be placed somewhere that is easy to update, as you will will want to update them with each release. Network seems like logical place to me? Any better idea?

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -32,6 +33,9 @@ public class BlockValidationContext
 
         /// <summary>If the block validation failed this will be set with the reason of failure.</summary>
         public ConsensusError Error { get; set; }
+
+        /// <summary>Whether to skip block validation for this block due to either a checkpoint or assumevalid hash set.</summary>
+        public bool SkipValidation { get; set; }
```
dangershony: This should go in the other context (this context is wrongly named its more about downloading blocks)

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -110,6 +110,12 @@ public NodeSettings()
         /// <summary>Whether use of checkpoints is enabled or not.</summary>
         public bool UseCheckpoints { get; set; }
 
+        /// <summary>
+        /// If this block is in the chain assume that it and its ancestors are valid and skip their script verification. 
+        /// Null to not assume valid blocks and therefore validate all blocks.
+        /// </summary>
+        public uint256 BlockAssumedValid { get; set; }
```
dangershony: Is there not a ConsensusSettings?
I would expect `BlockAssumedValid` to go there.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -1,17 +1,18 @@
...
 using Stratis.Bitcoin.BlockPulling;
+using Stratis.Bitcoin.Configuration;
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
 using Stratis.Bitcoin.Utilities;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
```
Aprogiena: thought we've discussed this recently that System usings are first and then the rest is lexicographically

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: you sure? i would put it exactly here as well. i think this is the correct context, there is that error that you receive if validation failed ...

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -405,15 +416,34 @@ public void ValidateBlock(ContextInformation context)
...
+
+                if (!context.BlockValidationContext.SkipValidation)
                 {
                     this.Validator.ContextualCheckBlock(context);
 
                     // Check the block itself.
                     this.Validator.CheckBlock(context);
                 }
-                else this.logger.LogTrace("Block validation partially skipped because block height {0} is not greater than last checkpointed block height {1}.", context.BlockValidationContext.ChainedBlock.Height, lastCheckpointHeight);
+                else
```
Aprogiena: why? this is ugly ;)

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
mikedennis: No ConsensusSettings but I could create one. What else would be in it? There is a ConsensusOptions class but it doesn't seem to be initialized by config.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: Changed it back for you. I'm not that passionate about else statements :-)

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: Fixed this. Will continue this discussion in slack.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: If this is for only downloading blocks we need to change the comments too as they all refer to "downloaded or mined".

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
dangershony: yeah create one, basically anything that is runtime configurable should go in settings, consensus options is part of the consensus rules so its a different thing altogether.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: To me this context is for interacting with the consensus (you send  context and check the error), but stuff that are part of the consensus validation go in the other context.

Well let me think about it again, I will review the code better tomorrow.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: I also don't mind, but I like to keep @Aprogiena happy 😄

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
mikedennis: Sounds good. Working on it...

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -416,11 +414,11 @@ public void CheckAndComputeStake(ContextInformation context)
                     ConsensusErrors.PrevStakeNull.Throw();
 
                 // Only do proof of stake validation for blocks after last checkpoint.
```
Aprogiena: this comment needs to be updated

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -405,15 +416,33 @@ public void ValidateBlock(ContextInformation context)
...
+                    context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= this.checkpoints.GetLastCheckpointHeight();
+                    if (context.BlockValidationContext.SkipValidation)
+                        this.logger.LogTrace("Block validation will be partially skipped due to block height {0} is not greater than last checkpointed block height {1}.", context.BlockValidationContext.ChainedBlock.Height, lastCheckpointHeight);
+                }
+
+                // Check whether to use assumevalid switch to skip validation.
+                if (!context.BlockValidationContext.SkipValidation && (this.nodeSettings.BlockAssumedValid != null))
+                {
+                    ChainedBlock assumeValidBlock = this.Chain.GetBlock(this.nodeSettings.BlockAssumedValid);
+                    context.BlockValidationContext.SkipValidation = (assumeValidBlock != null) && (context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height);
```
dangershony: What I would prefer to see is this
```
context.SkipValidation = ...
```
Unless there are reasons to set 'SkipValidation' from the caller but I don't see that as being the case.
WDYT

dangershony:
This looks good I will download the branch and try it out as well.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: Yes I can do one big conditional. I thought it might be easier to understand like this but I'm not sure it is. It's still a bit ugly. :grin:

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -405,15 +416,33 @@ public void ValidateBlock(ContextInformation context)
...
                 context.Flags = this.NodeDeployments.GetFlags(context.BlockValidationContext.ChainedBlock);
 
-                int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
-                if (context.BlockValidationContext.ChainedBlock.Height > lastCheckpointHeight)
+                // Check whether to use checkpoint to skip block validation.
+                context.BlockValidationContext.SkipValidation = false; 
+                if (this.nodeSettings.UseCheckpoints)
+                {
+                    int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
+                    context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= this.checkpoints.GetLastCheckpointHeight();
```
dangershony: this should be 
` context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= lastCheckpointHeight;`

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: That's fine actually. 
Another thing `(assumeValidBlock != null)` should this not throw if null?

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
dangershony: hmm yeah the Network is probably the best candidate.
Alternatively we could put it tin the deamon.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: Yeah good question. So I guess the question is if a incorrect hash is specified do we want to throw an exception or ignore assume valid check? At a minimum I should at least log it.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: Yes - I will fix that up.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: if assumevalid hash is not on the chain, all is fine! that's perfectly normal scenario and you don't even need to log it imho, because you would just add extra log for each block on the chain ...

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -234,22 +229,10 @@ public static NodeSettings Default(Network network = null, ProtocolVersion proto
...
-                // Block Height 184096 https://chainz.cryptoid.info/strat-test/block.dws?184096.htm
-                defaultAssumeValid = new uint256("0x74427b2f85b5d9658ee81f7e73526441311122f2b23702b794be557ba43ca43e");
-            }
-            // TODO: Add defaults for bitcoin and mainnet networks.
-
-            nodeSettings.BlockAssumedValid = config.GetOrDefault<uint256>("assumevalid", defaultAssumeValid);
-            if (nodeSettings.BlockAssumedValid == 0) // 0 means validate all blocks
-                nodeSettings.BlockAssumedValid = null;
-
+            nodeSettings.Consensus.Load(config, nodeSettings.Network);
```
dangershony: Hi mike one more comment, ConsensusSettings can be standalone and not inside the Settings file (see other examples like StoreSettings) that way we don't bind consensus setting to the base feature.

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -234,22 +229,10 @@ public static NodeSettings Default(Network network = null, ProtocolVersion proto
...
+            if (nodeSettings.Logger.IsEnabled(LogLevel.Debug))
```
dangershony: This can be moved to the ConsensusFeature startup.

src/Stratis.Bitcoin/Configuration/Settings/ConsensusSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        /// <summary>
+        /// Gets the default setting for the hash for the block to assume is valid.
+        /// </summary>
+        /// <param name="network">Network to return hash for.</param>
+        /// <returns>The hash for the assume valid block.</returns>
+        private uint256 GetDefaultAssumeValidBlock(Network network)
+        {
+            // TODO: Figure out where to put defaults so they are easy to configure.
```
dangershony: ok yes lets keep this here for now.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -405,15 +416,37 @@ public void ValidateBlock(ContextInformation context)
...
+                if (!context.BlockValidationContext.SkipValidation && (this.nodeSettings.Consensus.BlockAssumedValid != null))
+                {
+                    ChainedBlock assumeValidBlock = this.Chain.GetBlock(this.nodeSettings.Consensus.BlockAssumedValid);
+                    if (assumeValidBlock != null)
+                    {
+                        context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height;
+                        if (context.BlockValidationContext.SkipValidation)
+                            this.logger.LogTrace("Block validation will be partially skipped due to block height {0} is not greater than assumed valid block height {1}.", context.BlockValidationContext.ChainedBlock.Height, assumeValidBlock.Height);
+                    }
+                    else this.logger.LogWarning("Cannot find block specified by config flag assumevalid: '{0}'. Validation will not be skipped.", this.nodeSettings.Consensus.BlockAssumedValid);
```
dangershony: I still think the else needs to throw.
If the assume valid is not in the main chain do we continue with validation enabled or stop?
If the assume valid hash was passed in configuation then the user wants to be on a certain chain.
If we do decide to throw the block should not be added to the InvalidBlocks received.

mikedennis: Core is just ignoring flag and doing validation if it can't find the hash. If we want to throw an exception I think we might want to validate it earlier (maybe after headers are downloaded) as with the  default settings this code won't even get executed until after the checkpoints. So you won't see it until near the end of the IBD.

mikedennis:
@dangershony  I will also move consensusSettings to standalone settings as per your suggestion. I'm not sure why it doesn't show up here anymore (I think the conflict erased it) but I got an email with your comment so know what to do.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: Good point about validating earlier ofcourse.
Right if that's what core are doing then we do the same.

TBH I think this code will replace checkpoints, this is duplicate effort of the same thing only AssumeValid is more flexible.

Aprogiena: @dangershony if the assume valid hash is not on your chain you need to do full validation and this is perfectly valid case - it just mean that the chain that you knew that was valid was reorged and that is perfectly fine thing to happen, this is the difference between this feature and checkpoints

PR #734, opened 2017-11-12
minor cleanup

Aprogiena:
this is just removing one compiler warning and removing two things from staking code that are unused, no harm should be caused by this


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -793,7 +793,6 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
                         {
                             context.Logger.LogTrace("Kernel found with solution hash '{0}'.", contextInformation.Stake.HashProofOfStake);
 
-                            BitcoinAddress outPubKey = scriptPubKeyKernel.GetDestinationAddress(this.network);
```
bokobza: Is that not used anywhere?

Aprogiena: well ... that's why i remove it :-)

bokobza: cool then ;-)

PR #738, opened 2017-11-13
Refactor/protocol node cleanup

Aprogiena:
cleanup of code, hopefully does not damage anything


Aprogiena:
please don't merge until appveyor checks that all is ok

fassadlr:
Some tests failed, are they related?

Aprogiena:
test failed - hopefully not, but i don't really know, it is really hard to tell if the test failure is related when it is normal for some test to fail - this is exactly why i would love to have a condition that PR is not merged if even 1 test fails. if we just did that, suddenly the tests would be useful 10x more than now, currently they only help us to see if a major problem happened, but otherwise, you can't really tell easily

if we could really have 100 % tests passed and then really not allow any new piece of code being merged without all tests passing, that would improve reliability of our development process extremely

PR #742, opened 2017-11-15
Updated tests to match changes done in previous commit

bokobza:
The offending commit: https://github.com/stratisproject/StratisBitcoinFullNode/commit/34275331375894d40ba4f2dfe4b7c9439d5a36fa

### DIFF
--- a/src/Stratis.Bitcoin.Features.LightWallet.Tests/LightWalletSyncManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.LightWallet.Tests/LightWalletSyncManagerTest.cs
@@ -311,7 +311,7 @@ namespace Stratis.Bitcoin.Features.LightWallet.Tests
             var asyncLoop = new Mock<IAsyncLoop>();
             this.asyncLoopFactory.Setup(
                 a => a.RunUntil(
-                    "WalletFeature.DownloadChain",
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.LightWallet.Tests/LightWalletSyncManagerTest.cs
```diff
@@ -311,7 +311,7 @@ public void Stop_DisposesDependencies()
             var asyncLoop = new Mock<IAsyncLoop>();
             this.asyncLoopFactory.Setup(
                 a => a.RunUntil(
-                    "WalletFeature.DownloadChain",
+                    "LightWalletSyncManager.SyncFromHeight",
                     It.IsAny<CancellationToken>(),
```
fassadlr: Actually cant we use:

```nameof(LightWalletSyncManager.SyncFromDate)```

?

bokobza: yeah we could, although I'm not sure it'll provide great value here.
Maybe something we can look at changing in all the async loops everywhere.

fassadlr: OK cool! Ready to merge then.

PR #743, opened 2017-11-15
Add assumevalid hash for other networks

mikedennis:
Also fixed an issue where if checkpoints are turned off and assumevalid is on, the assumevalid check was being turned off on a fresh IBD because the headers were not yet downloaded. Reverted that code back to what bitcoin core is doing - if can't find assumevalid block it will just proceed to validate it.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
@@ -429,17 +429,9 @@ namespace Stratis.Bitcoin.Features.Consensus
                 if (!context.BlockValidationContext.SkipValidation && (this.settings.BlockAssumedValid != null))
                 {
                     ChainedBlock assumeValidBlock = this.Chain.GetBlock(this.settings.BlockAssumedValid);
-                    if (assumeValidBlock != null)
-                    {
-                        context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height;
-                        if (context.BlockValidationContext.SkipValidation)
-                            this.logger.LogTrace("Block validation will be partially skipped due to block height {0} is not greater than assumed valid block height {1}.", context.BlockValidationContext.ChainedBlock.Height, assumeValidBlock.Height);
-                    }
-                    else
-                    {
-                        this.logger.LogWarning("Cannot find block specified by config flag assumevalid: '{0}'. Assume valid block feature will be disabled.", this.settings.BlockAssumedValid);
-                        this.settings.BlockAssumedValid = null;
-                    }
+                    context.BlockValidationContext.SkipValidation = (assumeValidBlock != null) && (context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height);
+                    if (context.BlockValidationContext.SkipValidation)
+                        this.logger.LogTrace("Block validation will be partially skipped due to block height {0} is not greater than assumed valid block height {1}.", context.BlockValidationContext.ChainedBlock.Height, assumeValidBlock.Height);
                 }
 
                 if (!context.BlockValidationContext.SkipValidation)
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -429,17 +429,9 @@ public void ValidateBlock(ContextInformation context)
...
-                        context.BlockValidationContext.SkipValidation = context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height;
-                        if (context.BlockValidationContext.SkipValidation)
-                            this.logger.LogTrace("Block validation will be partially skipped due to block height {0} is not greater than assumed valid block height {1}.", context.BlockValidationContext.ChainedBlock.Height, assumeValidBlock.Height);
-                    }
-                    else
-                    {
-                        this.logger.LogWarning("Cannot find block specified by config flag assumevalid: '{0}'. Assume valid block feature will be disabled.", this.settings.BlockAssumedValid);
-                        this.settings.BlockAssumedValid = null;
-                    }
+                    context.BlockValidationContext.SkipValidation = (assumeValidBlock != null) && (context.BlockValidationContext.ChainedBlock.Height <= assumeValidBlock.Height);
```
dangershony: Ah good point yes

src/Stratis.Bitcoin.Features.Consensus/ConsensusSettings.cs
```diff
@@ -58,20 +58,27 @@ private ConsensusSettings LoadFromConfig()
...
+        /// Gets the default setting for the block hash of the block to assume is valid.
         /// </summary>
         /// <param name="network">Network to return hash for.</param>
         /// <returns>The hash for the assume valid block.</returns>
         private uint256 GetDefaultAssumeValidBlock(Network network)
         {
-            // TODO: Figure out where to put defaults so they are easy to configure.
             uint256 defaultAssumeValidBlock = null;
-            if (!network.IsBitcoin() && network.IsTest())
+            if (network.IsBitcoin())
```
dangershony: This is ok for this PR but how about we put it in Network? (its now part of the project so should be easy)

src/Stratis.Bitcoin.Features.Consensus/ConsensusSettings.cs
```diff
@@ -58,20 +58,27 @@ private ConsensusSettings LoadFromConfig()
         }
 
         /// <summary>
-        /// Gets the default setting for the hash for the block to assume is valid.
+        /// Gets the default setting for the block hash of the block to assume is valid.
```
dangershony: Do we have a way to switch off AssumeValid?
I am not even sure we need to switch it off as if the block is not present it will default to validate it

mikedennis: assumevalid=0 will switch it off, it's the same as core.

src/Stratis.Bitcoin.Features.Consensus/ConsensusSettings.cs
mikedennis: Sure. I also realized that bitcoin core puts in what would be our ConsensusOptions. Do we have a different set of ConsensusOptions for each network?

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
mikedennis: If we want to validate that a good hash was entered for assumevalid we need to wait until we have all the headers first. So I'm not sure a real good way of doing that yet.

src/Stratis.Bitcoin.Features.Consensus/ConsensusSettings.cs
dangershony: Sort of yes we can set stuff in our ConsensusOptions when building the network

mikedennis: Do you think better in network or consensusOptions?

PR #745, opened 2017-11-15
Move default value for assumeValid hash into network settings.

mikedennis:


dangershony:
@mikedennis maybe while you are t it also put checkpoints in to Network

PR #752, opened 2017-11-18
Fix Build & Fix 2 miner tests with checkpoint logic

fassadlr:
Two miner integration tests were failing because consensus uses check points which causes the block validation to be skipped. This causes the test to fail as it expects a exception to be thrown.


bokobza:
wow, I have no idea how I missed these 2  `NodeSettings.FromArguments` 😕

PR #732, opened 2017-11-10
Minor code improvements based on static code analysis

noescape00:
Most of the changes are ensuring following 'Code Commenting Policy' and 'C# Coding Style' guidelines

PR #753, opened 2017-11-18
PoS minting big picture comment

noescape00:
Added PoS minting big picture comment, edited some PoS\PoW variables descriptions, renamed some variables and functions. 

Added 2 small fixes that were mentioned after my previews PRs got already merged.  (First is about adding Guard.NotNull, mentioned by bokobza here https://github.com/stratisproject/StratisBitcoinFullNode/pull/732, second is about accidentally deleting '<' mentioned by aprogiena in slack's pm)


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,6 +20,39 @@
 
 namespace Stratis.Bitcoin.Features.Miner
 {
+    /// <summary>
+    /// PosMinting is used in order to generate new blocks. It involves a sort of lottery, similar to proof-of-work,
+    /// but the chances of winning this lottery depend on how many coins you are staking, not on hashing power.
```
dangershony: but the chances of winning this lottery is `proportional` to how many coins you are staking

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,6 +20,39 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Staking is attempted only if the node is fully synchronized and connected to the network.
+    /// If not it will wait  till node is synced. Only transactions that were confirmed at least
+    /// PosConsensusOptions.StakeMinConfirmations blocks ago are eligible for staking.
```
dangershony: >PosConsensusOptions.StakeMinConfirmations
Use <seeRef="PosConsensusOptions.StakeMinConfirmations">

Aprogiena:
Let me check this one

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,6 +20,39 @@
...
+    /// Search interval is a length of an unexplored block time space in seconds.
+    /// Worker calculates the kernel's hash using the next formula:
+    /// hash(stakeModifierV2 + stakingCoins.Time + prevout.Hash + prevout.N + transactionTime).
+    /// Then it calculates staking target using the next formula:
+    /// (header of a block that we are trying to mine as BigInteger * amount of satoshi we are staking).
+    /// We compare kernel's hash (as BigInteger) against the staking target, if it's greater then we met the criteria
+    /// and kernel is found.
+    /// So the more coins we stake the higher the staking target => higher chances of meeting the criteria.
+    /// 4) In case kernel is found we add a coinbase transaction, staking transaction and prepare the block normally,
+    /// with transactions from network, sign the block and add it to the chain.
```
dangershony: At this point transactions where already added to the block

dangershony:
Looks good to me lets wait for @Aprogiena to review.

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -29,7 +29,7 @@ namespace Stratis.Bitcoin.Features.Consensus
     /// <item>Coinstake transaction's kernel (first) input MUST not be created within last <see cref="PosConsensusOptions.StakeMinConfirmations"/> blocks,
     /// i.e. it MUST have that many confirmation at least.</item>
     /// <item>Coinstake transaction's kernel must meet the staking target using this formula:
-    /// <code>hash(stakeModifierV2 + stakingCoins.Time + prevout.Hash + prevout.N + transactionTime) target * weight</code>
+    /// <code><![CDATA[hash(stakeModifierV2 + stakingCoins.Time + prevout.Hash + prevout.N + transactionTime) < target * weight]]></code>
```
Aprogiena: i'd be happier with `&lt;`

Aprogiena:
@dangershony rather than readme file, i'd suggest having an external knowledge base 

but i'd not be happy to duplicate the information from class comment into that or readme file, you will find the comment will change in one source and won't in other source, then you have a mess

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -58,7 +91,7 @@ public class CoinstakeWorkerContext
             public ILogger Logger { get; set; }
 
             /// <summary>List of UTXO descriptions that the worker should check.</summary>
-            public List<StakeTx> Coins { get; set; }
+            public List<StakeTx> Transactions { get; set; }
```
Aprogiena: as explained on slack, this change is not correct imho
Coin != Money, Coin is UTXO
i would stay with the terminology that is being used, so i would not make any changes here, but if i were not to use Coin, i'd go with something like "UtxoStakeInfo" or something like that, but then Coin is better again :-)
but also note that StakeTx is not a good name for that class

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -119,10 +152,15 @@ public class CoinstakeContext
         // Default for -blockmintxfee, which sets the minimum feerate for a transaction in blocks created by mining code
         public const int DefaultBlockMinTxFee = 1000;
 
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create
+        // Default for -blockmaxsize, which controls the maximum number of serialized bytes in a block the mining code will create.
         public const int DefaultBlockMaxSize = 750000;
```
Aprogiena: when you make any such modifications, feel free to improve the comment style to be xmldoc, i.e. `/// <summary> ...`

and about the change itself - i don't really get it, the constant is called DefaultBlockMaxSize, so it seems like it is a maximum size, so is it not (then we should rename it in the first place)?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,6 +20,39 @@
 
 namespace Stratis.Bitcoin.Features.Miner
 {
+    /// <summary>
+    /// PosMinting is used in order to generate new blocks. It involves a sort of lottery, similar to proof-of-work,
```
Aprogiena: always use `<see>` for everything that exists and you want to refer to it, i.e. all classes, constants etc. but not in formulas (like target formula below), but elsewhere, always use it, i won't repeat this comment, but it should be applied everywhere

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -119,10 +152,15 @@ public class CoinstakeContext
...
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create
+        // Default for -blockmaxsize, which controls the maximum number of serialized bytes in a block the mining code will create.
         public const int DefaultBlockMaxSize = 750000;
 
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create
+        /// <summary>
+        /// Default for -blockmaxweight, which controls the maximum of block weight the mining code will create.
+        /// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
+        /// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
+        /// <remarks>See also: http://learnmeabitcoin.com/faq/segregated-witness </remarks>
```
Aprogiena: instead of this, use `<seealso>`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -119,10 +152,15 @@ public class CoinstakeContext
         // Default for -blockmintxfee, which sets the minimum feerate for a transaction in blocks created by mining code
         public const int DefaultBlockMinTxFee = 1000;
 
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create
+        // Default for -blockmaxsize, which controls the maximum number of serialized bytes in a block the mining code will create.
         public const int DefaultBlockMaxSize = 750000;
 
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create
+        /// <summary>
+        /// Default for -blockmaxweight, which controls the maximum of block weight the mining code will create.
```
Aprogiena: I'd prefer having those switches inside "" or alike. maybe even `<c>`, but "" would be good enough imho

Aprogiena: also pls check if that can be changed elsewhere

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -144,7 +182,7 @@ public class CoinstakeContext
         /// <summary>Number of UTXOs that a single worker's task will process.</summary>
         /// <remarks>To achieve a good level of parallelism, this should be low enough so that CPU threads are used,
         /// but high enough to compensate for tasks' overhead.</remarks>
-        private const int CoinsPerCoinstakeWorker = 25;
+        private const int TransactionsPerCoinstakeWorker = 25;
```
Aprogiena: same here, it is not transactions

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -578,21 +616,21 @@ private void CheckStake(Block block, ChainedBlock chainTip)
             {
                 this.rpcGetStakingInfoModel.Staking = false;
 
-                this.logger.LogTrace("Total balance of available UTXOs is {0}, which is lower than reserve balance {1}.", balance, this.reserveBalance);
+                this.logger.LogTrace("Total balance of available UTXOs is {0}, which is equal or lower than reserve balance {1}.", balance, this.reserveBalance);
```
Aprogiena: `lower or equal` (think `<=` it is first `<` then `=`) and probably best is `less than or equal to`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -891,11 +929,11 @@ public Money GetMatureBalance(List<StakeTx> stakeTxes)
         /// and it also has to be mature and meet requirement for minimal value.
         /// </para>
         /// </summary>
-        /// <param name="stakeTxes">List of coins that are candidates for being used for staking.</param>
+        /// <param name="stakeTxes">List of UTXOs that are candidates for being used for staking.</param>
```
Aprogiena: ok, let's say "UTXO descriptions" so that we are consistent with the comments above

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -29,12 +29,18 @@ public class PowMining
         // Default for -blockmintxfee, which sets the minimum feerate for a transaction in blocks created by mining code.
         public const int DefaultBlockMinTxFee = 1000;
 
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create.
+        // Default for -blockmaxsize, which controls the maximum number of serialized bytes in a block the mining code will create.
```
Aprogiena: same as in Pos case above

Aprogiena: also i don't repeat similar comments from pos case

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -29,12 +29,18 @@ public class PowMining
...
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create.
+        /// <summary>
+        /// Default for -blockmaxweight, which controls the maximum of block weight the mining code will create.
+        /// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
+        /// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
+        /// <remarks>See also: http://learnmeabitcoin.com/faq/segregated-witness </remarks>
+        /// </summary>
         public const int DefaultBlockMaxWeight = 3000000;
 
+        /// <summary>InnerLoopCount = 65536.</summary>
```
Aprogiena: that's not a good description of the constant
also we are missing an access modifier here, can we add it pls?

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -98,6 +104,14 @@ public IAsyncLoop Mine(Script reserveScript)
             return this.mining;
         }
 
+        /// <summary>
+        /// Generates specified amount of blocks in specified maximum amount of tries.
+        /// It is possible that less than the required number of blocks will be mined.
+        /// </summary>
+        /// <param name="reserveScript"></param>
```
Aprogiena: empty comment is not a good one ;)

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -98,6 +104,14 @@ public IAsyncLoop Mine(Script reserveScript)
...
+        /// <param name="generate">Blocks count to generate</param>
```
Aprogiena: missing dot at the end of a sentence

Aprogiena: "Number of blocks" or "Block count", not "Blocks count"

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -98,6 +104,14 @@ public IAsyncLoop Mine(Script reserveScript)
...
+        /// <param name="maxTries">Maximum amount of times miner will calculate PoW hash in order to find suitable ones to generate specified amount of blocks</param>
```
Aprogiena: Maximum number of attempts the miner will calculate ...

and missing dot

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -125,7 +139,7 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
 
                 if (Block.BlockSignature)
                 {
-                    // POS: make sure the POS consensus rules are valid 
+                    // POS: make sure the POS consensus rules are valid
```
Aprogiena: missing dot and we can even make it better comment, maybe something like 

`Make sure the POS consensus rules are valid. This is required for generation of blocks inside tests, where it is possible to generate multiple blocks within one second.`

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -98,6 +104,14 @@ public IAsyncLoop Mine(Script reserveScript)
             return this.mining;
         }
 
+        /// <summary>
+        /// Generates specified amount of blocks in specified maximum amount of tries.
+        /// It is possible that less than the required number of blocks will be mined.
```
Aprogiena: too much of `specified` :-)

also not amount

->

Generates up to a specified number of blocks with a limited number of attempts.

and I would then move the second sentence into the comment of generate parameter

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: Why not? StakeTx immediately convey the purpose of the class. But I am very happy with UtxoStakeInfo as well.

noescape00:
Fixed everything but this
>empty comment is not a good one ;)
Because I'm not sure yet what this one actually does. Will add comment when I'll be doing comments for this class.

Also new names for the variables and functions might be considered a bit long, but at least they represent the purpose much better. 
So let me know what you think =)

Aprogiena:
ok, you have a conflict here, can you resolve?

i know you don't know what that one is, that was the point :) look at BlockAssembler class and how GenerateBlock is called, it might help:
```
                    this.powLoop = this.powMining.Mine(BitcoinAddress.Create(minto, this.network).ScriptPubKey);

->

                BlockTemplate pblockTemplate = this.blockAssemblerFactory.Create(chainTip).CreateNewBlock(reserveScript.reserveSfullNodecript);

->

        public abstract BlockTemplate CreateNewBlock(Script scriptPubKeyIn, bool fMineWitnessTx = true);

->

        protected virtual void CreateCoinbase()
        {
            // Create coinbase transaction.
            // Set the coin base with zero money.
            // Once we have the fee we can update the amount.
            this.coinbase = new Transaction();
            this.coinbase.AddInput(TxIn.CreateCoinbase(this.ChainTip.Height + 1));
            this.coinbase.AddOutput(new TxOut(Money.Zero, this.scriptPubKeyIn));

```

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: StakeTx is wrong name for the class and also Transactions is wrong name of this list
Coins or StakingCoins is probably the best for this list and we need to decide how to rename StakeTx because StakeTx is no transaction, it even contains wallet password ...

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
 
 namespace Stratis.Bitcoin.Features.Miner
 {
+    /// <summary>
+    /// <see cref="PosMinting"/> is used in order to generate new blocks. It involves a sort of lottery, similar to proof-of-work,
+    /// but the chances of winning this lottery is proportional to how many coins you are staking, not on hashing power.
+    /// </summary>
+    /// <remarks>
+    /// Staking is attempted only if the node is fully synchronized and connected to the network.
+    /// If not it will wait  till node is synced. Only transactions that were confirmed at least
```
Aprogiena: double space before `till`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// <see cref="PosConsensusOptions.StakeMinConfirmations"/> blocks ago are eligible for staking.
+    /// </remarks>
+    /// <remarks>
+    /// The overall process for "attempting" to mine a PoS block looks like this:
+    /// 1) Create new block with transactions from <see cref="TxMempool"/>
```
Aprogiena: use <list> - see e.g. PullerDownloadAssignmentsTest how this tag is used

+ missing dot here

and I tend to avoid mentioning concrete implementation of something when it is not needed. So you say here `from TxMempool`, but at this point I'm not sure the reader appreciates this information that it is actually TxMempool class, I think the reader would be happy with just `from mempool`.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// 2) Get UTXOs (Unspent Transaction Output) that can participate in staking (have suitable depth).
```
Aprogiena: we're bitcoin ppl, i think we know what UTXO stands for, if we would repeat it in every file ...

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// 3) Split those UTXOs in subsets so each subset contains max of <see cref="utxoStakeDescriptionsPerCoinstakeWorker"/> transactions
+    /// and run CoinstakeWorker for each subset. Splitting is done only to achieve a good level of parallelism.
```
Aprogiena: `<see>` for CoinstakeWorker 

OR (imho better) abstract the reader of these details and just say something like
`Split these UTXOs in subsets and create tasks processing each subset to allow for parallel processing.`

i think it is better because mentioning the actual constant we have there, that's a detail that maybe does not belong to the big picture, if one is interested in such a detail, he would go to the code and will find this constant and if the constant is properly described, the information will be there.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// 4) CoinstakeWorker will try different versions of a block we are trying to mine (the only difference is the
```
Aprogiena: again `<see>` for CW, but if we modify 3) as suggested, then it would be `Each task is run inside <see cref="CoinstakeWorker"/>, which will try ...`

however, not sure if we can speak of different versions of a block (that might be misleading due to what version means in bitcoin)

maybe i'd say `will try to find a solution for proof of stake target. This is done by creating a coinstake transaction with each of the available UTXOs combined with all valid timestamps that were not checked.`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// timestamp which is tried in a time interval from now to now - searchInterval seconds.
```
Aprogiena: any formula should be inside `<c>` or `<code>`

I'm missing information about the mask here. This is quite important thing imho, even at big picture level. The mask is 0xf and it basically means that only unix timestamps that are divisible by 16 are valid candidates. So if you have block with timestamp X, you can't have a block with timestamp X+1 or X+2. but the next block must have X+16, X+32, X+48, ... see StakeTimestampMask

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// Search interval is a length of an unexplored block time space in seconds.
+    /// Worker calculates the kernel's hash using the next formula:
+    /// hash(stakeModifierV2 + stakingCoins.Time + prevout.Hash + prevout.N + transactionTime).
```
Aprogiena: formula in `<code>` here

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// Then it calculates staking target using the next formula:
+    /// (header of a block that we are trying to mine as BigInteger * amount of satoshi we are staking).
```
Aprogiena: this does not make much sense, reading this I'd imagine `(BigInteger)header`

i would completely avoid mentioning BigInteger here, it is just a tool for number calculation, not interesting here

what is well known in the industry is "difficulty target", so basically each block has to meet certain difficulty target, PoW or PoS, does not matter, there is a difficulty target and here in PoS we have a coinstake transaction input (aka kernel) hash (unlike in PoW where we have a block header hash) that has to be lower than the target

also note that this is not `amount of satoshi we are staking`, it is not because
1) in stratis there are no satoshis :-) - say just `amount` and it is enough
2) it is not the amount we are staking, but value of that UTXO 

so the target is calculated as `block difficulty * UTXO value`

also why this is in `()` and also `<code>` for formulas

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// We compare kernel's hash (as BigInteger) against the staking target, if it's greater then we met the criteria
+    /// and kernel is found.
```
Aprogiena: here you are working with `kernel`, which i think deserves to be defined (kernel is the first input in the coinstake transaction)

avoid BigInteger

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// So the more coins we stake the higher the staking target => higher chances of meeting the criteria.
```
Aprogiena: avoid `=>`, use english, in non-formulas ... `and so the higher the chance is to meet the criteria.`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// 5) In case kernel is found we add a coinbase transaction and staking transaction, sign the block and add it to the chain.
```
Aprogiena: where do you see the coinbase being manipulated after the solution is found?

i see coinstake transaction being added to the block and the block is signed

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    ///
+    /// More about PoS kernel:
```
Aprogiena: i'd remove this part, it is also incorrect - see the mask comment above

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -20,15 +20,48 @@
...
+    /// <summary>
+    /// <see cref="PosMinting"/> is used in order to generate new blocks. It involves a sort of lottery, similar to proof-of-work,
+    /// but the chances of winning this lottery is proportional to how many coins you are staking, not on hashing power.
+    /// </summary>
+    /// <remarks>
+    /// Staking is attempted only if the node is fully synchronized and connected to the network.
+    /// If not it will wait  till node is synced. Only transactions that were confirmed at least
+    /// <see cref="PosConsensusOptions.StakeMinConfirmations"/> blocks ago are eligible for staking.
+    /// </remarks>
+    /// <remarks>
```
Aprogiena: there should be just single remarks section
use `<para>` to structure longer section (summary or remarks or alike) into different paragraphs

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -116,13 +149,18 @@ public class CoinstakeContext
...
         public const int DefaultBlockMaxSize = 750000;
 
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create
+        /// <summary>
+        /// Default for "-blockmaxweight", which controls the maximum of block weight the mining code will create.
+        /// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
+        /// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
+        /// <seealso>http://learnmeabitcoin.com/faq/segregated-witness </seealso>
+        /// </summary>
         public const int DefaultBlockMaxWeight = 3000000;
```
Aprogiena: how about we delete this constant, it is not used ...

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -141,10 +179,10 @@ public class CoinstakeContext
         /// multiplied by this value. See <see cref="GetSplitStake(int)"/>.</summary>
```
Aprogiena: missing opening summary tag and also just before that there are two consts with old comments, xmldoc them pls

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -141,10 +179,10 @@ public class CoinstakeContext
...
         public const int CoinstakeSplitLimitMultiplier = 3;
 
-        /// <summary>Number of UTXOs that a single worker's task will process.</summary>
+        /// <summary>Number of UTXO descriptions that a single worker's task will process.</summary>
         /// <remarks>To achieve a good level of parallelism, this should be low enough so that CPU threads are used,
         /// but high enough to compensate for tasks' overhead.</remarks>
-        private const int CoinsPerCoinstakeWorker = 25;
+        private const int utxoStakeDescriptionsPerCoinstakeWorker = 25;
```
Aprogiena: code style -> PascalCase for all constants

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -555,44 +593,44 @@ private void CheckStake(Block block, ChainedBlock chainTip)
...
                 this.logger.LogTrace("(-)[BELOW_RESERVE]:false");
                 return false;
             }
 
-            // Select coins with suitable depth.
-            List<StakeTx> setCoins = this.FindCoinsForStaking(stakeTxes, coinstakeContext.CoinstakeTx.Time, balance - this.reserveBalance);
-            if (!setCoins.Any())
+            // Select UTXOs with suitable depth.
+            List<UtxoStakeDescription> stackingUtxoDescriptions = this.GetUtxoStakeDescriptionsSuitableForStaking(utxoStakeDescriptions, coinstakeContext.CoinstakeTx.Time, balance - this.reserveBalance);
+            if (!stackingUtxoDescriptions.Any())
```
Aprogiena: stacking or staking?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -628,14 +666,14 @@ private void CheckStake(Block block, ChainedBlock chainTip)
...
-                    Coins = new List<StakeTx>(),
+                    utxoStakeDescriptions = new List<UtxoStakeDescription>(),
                     CoinstakeContext = coinstakeContext,
                     Result = workersResult
                 };
 
-                int coinCount = Math.Min(setCoins.Count - coinIndex, CoinsPerCoinstakeWorker);
-                cwc.Coins.AddRange(setCoins.GetRange(coinIndex, coinCount));
-                coinIndex += coinCount;
+                int txesCount = Math.Min(stackingUtxoDescriptions.Count - coinIndex, utxoStakeDescriptionsPerCoinstakeWorker);
```
Aprogiena: coinCount was better, but feel free to rename to stakingUtxoCount or whatever, just not transactions

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -726,19 +764,19 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
...
             context.Logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(chainTip), chainTip, nameof(minimalAllowedTime), minimalAllowedTime, nameof(searchInterval), searchInterval);
 
-            context.Logger.LogTrace("Going to process {0} UTXOs.", context.Coins.Count);
+            context.Logger.LogTrace("Going to process {0} UTXOs.", context.utxoStakeDescriptions.Count);
 
-            // Sort coins by amount, so that highest amounts are tried first
+            // Sort staking UTXOs by amount, so that highest amounts are tried first
             // because they have greater chance to succeed and thus saving some work.
-            List<StakeTx> orderedCoins = context.Coins.OrderByDescending(o => o.TxOut.Value).ToList();
+            List<UtxoStakeDescription> orderedutxoStakeDescriptions = context.utxoStakeDescriptions.OrderByDescending(o => o.TxOut.Value).ToList();
```
Aprogiena: orderedUtxoStakeDescriptions - upper case U

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -793,18 +831,19 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
                         {
                             context.Logger.LogTrace("Kernel found with solution hash '{0}'.", contextInformation.Stake.HashProofOfStake);
 
-                            Wallet.Wallet wallet = this.walletManager.GetWalletByName(coin.Secret.WalletName);
-                            context.CoinstakeContext.Key = wallet.GetExtendedPrivateKeyForAddress(coin.Secret.WalletPassword, coin.Address).PrivateKey;
+                            BitcoinAddress outPubKey = scriptPubKeyKernel.GetDestinationAddress(this.network);
```
Aprogiena: remove outPubKey, it is not used

Aprogiena: it has been removed before, you have introduced it again

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -726,19 +764,19 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
...
             bool stopWork = false;
-            foreach (StakeTx coin in orderedCoins)
+            foreach (UtxoStakeDescription utxoStakeInfo in orderedutxoStakeDescriptions)
             {
-                context.Logger.LogTrace("Trying UTXO from address '{0}', output amount {1}...", coin.Address.Address, coin.TxOut.Value);
+                context.Logger.LogTrace("Trying UTXO from address '{0}', output amount {1}...", utxoStakeInfo.Address.Address, utxoStakeInfo.TxOut.Value);
 
                 // Script of the first coinstake input.
-                Script scriptPubKeyKernel = coin.TxOut.ScriptPubKey;
+                Script scriptPubKeyKernel = scriptPubKeyKernel = utxoStakeInfo.TxOut.ScriptPubKey;
```
Aprogiena: you have introduced this again, this has been simplified - there is double assignment 
->
`Script scriptPubKeyKernel = utxoStakeInfo.TxOut.ScriptPubKey;`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -890,81 +929,81 @@ public Money GetMatureBalance(List<StakeTx> stakeTxes)
         /// and it also has to be mature and meet requirement for minimal value.
         /// </para>
         /// </summary>
-        /// <param name="stakeTxes">List of coins that are candidates for being used for staking.</param>
+        /// <param name="utxoStakeDescriptions">List of UTXO descriptions that are candidates for being used for staking.</param>
         /// <param name="spendTime">Timestamp of the coinstake transaction.</param>
         /// <param name="maxValue">Maximal amount of coins that can be used for staking.</param>
         /// <returns>List of coins that meet the requirements.</returns>
```
Aprogiena: if you change elsewhere from coin to UTXO description, make sure you change it everywhere 
i haven't checked all the comments, so try to go through them and see if we are somewhere still using `coin` for this UtxoStakeDescription

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -26,16 +26,21 @@ public ReserveScript(Script reserveSfullNodecript)
...
+        /// <summary>
+        /// Default for "-blockmaxweight", which controls the maximum of block weight the mining code will create.
+        /// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
+        /// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
+        /// <seealso>http://learnmeabitcoin.com/faq/segregated-witness </seealso>
+        /// </summary>
         public const int DefaultBlockMaxWeight = 3000000;
 
-        const int InnerLoopCount = 0x10000;
+        private const int innerLoopCount = 0x10000;
```
Aprogiena: code style - PascalCase for all consts

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -26,16 +26,21 @@ public ReserveScript(Script reserveSfullNodecript)
...
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create.
+        /// <summary>Default for "-blockmaxsize", which controls the maximum size of block the mining code will create.</summary>
         public const int DefaultBlockMaxSize = 750000;
 
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create.
+        /// <summary>
+        /// Default for "-blockmaxweight", which controls the maximum of block weight the mining code will create.
+        /// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
+        /// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
+        /// <seealso>http://learnmeabitcoin.com/faq/segregated-witness </seealso>
```
Aprogiena: syntax of seealso is 
`<seealso cref="XYZ"/>`

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -26,16 +26,21 @@ public ReserveScript(Script reserveSfullNodecript)
...
+        /// <summary>Default for "-blockmintxfee", which sets the minimum feerate for a transaction in blocks created by mining code.</summary>
         public const int DefaultBlockMinTxFee = 1000;
 
-        // Default for -blockmaxsize, which controls the maximum size of block the mining code will create.
+        /// <summary>Default for "-blockmaxsize", which controls the maximum size of block the mining code will create.</summary>
         public const int DefaultBlockMaxSize = 750000;
 
-        // Default for -blockmaxweight, which controls the range of block weights the mining code will create.
+        /// <summary>
+        /// Default for "-blockmaxweight", which controls the maximum of block weight the mining code will create.
```
Aprogiena: maximum block weight - without OF
and it is not "will create" but "can create"

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -98,6 +103,13 @@ public IAsyncLoop Mine(Script reserveScript)
             return this.mining;
         }
 
+        /// <summary>
+        /// Generates up to a specified number of blocks with a limited number of attempts.
+        /// </summary>
+        /// <param name="reserveScript"></param>
+        /// <param name="generate">Number of blocks to generate. It is possible that less than the required number of blocks will be mined.</param>
+        /// <param name="maxTries">MMaximum number of attempts the miner will calculate PoW hash in order to find suitable ones to generate specified amount of blocks.</param>
```
Aprogiena: MM -> M

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -125,7 +137,8 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
 
                 if (Block.BlockSignature)
                 {
-                    // POS: make sure the POS consensus rules are valid 
+                    //Make sure the POS consensus rules are valid. This is required for generation of blocks inside tests,
```
Aprogiena: space after `//`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: This has not been changed on purpose because of issue #754. When I get to that issues I'll check what was intended by the writer of that piece and if there is a problem with implementation of this functionality.

Aprogiena: are we talking about the same thing?
i was talking about here we previous had `List of coins that are candidates` for `stakeTxes` parameter, we changed that to `UTXO descriptions` 

but in returns tag, we still return `List of coins`, which is inconsistent

also i've just noticed in StakeAndSignBlockAsync we have 
` /// <param name="utxoStakeDescriptions">List of coins that are available in the wallet for staking.</param>`

which is also inconsistent - so i was talking about going through the all comments and make this consistent

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -30,10 +30,12 @@ namespace Stratis.Bitcoin.Features.Miner
     /// <see cref="PosConsensusOptions.StakeMinConfirmations"/> blocks ago are eligible for staking.
     /// <para>
     /// The overall process for "attempting" to mine a PoS block looks like this:
-    /// 1) Create new block with transactions from mempool.
-    /// 2) Get UTXOs that can participate in staking (have suitable depth).
-    /// 3) Split these UTXOs in subsets and create tasks processing each subset to allow for parallel processing.
-    /// 4) Each of the tasks mentioned above will try to find a solution for proof of stake target. This is done by creating a coinstake
+    /// <list>
```
Aprogiena: almost :) you do have 5) out of list and you should not need prefixes like 1), 2) in list if you use `<list type="number">`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: Hmm yeah, StakingCoins sounds right to me it does indeed represent more then just a UTXO.
So class should be `StakingCoin`  and the list `StakingCoins`

Aprogiena: ok, cool, now the thing is that @noescape00 went forward and renamed StakeTx to UtxoStakeDescription and also modified some comments (but not all of them!! :D ) as such

so the question is, are we happy with UtxoStakeDescription instead of StakingCoin? or do we want to tease @noescape00 little more and let him redo?

Aprogiena: i mean everyone has to have a limit when they break ;) can we find one for @noescape00 :D

dangershony: @noescape00 I will buy you a beer when you come to London if you survive @Aprogiena code reviews.

dangershony: `UtxoStakeDescription` is not a bad name either, so I will leave it to @noescape00 to decide.

Aprogiena: ok cool, if you don't mind, i don't mind, so i will also leave it to @noescape00

noescape00: >i mean everyone has to have a limit when they break ;) can we find one for @noescape00 :D

You can try =) 

>@noescape00 I will buy you a beer when you come to London if you survive @Aprogiena code reviews.

That's a deal! 

>now the thing is that @noescape00 went forward and renamed StakeTx to UtxoStakeDescription

The reason for that is that we have a lot of lists so naming them utxoStakeInfos would have looked weird.

Aprogiena: that's fine, let's leave it as it is UtxoStakeDescription

but the point was missed, so i just explain that (but we won't be doing any changes) - what we were discussing is that we don't really use UTXO in variables/classes names, we use Coin. When you see Coin, you know it is UTXO, but it is just everywhere else called Coin (CoinView for example). So we were suggesting having StakingCoin class instead of StakeTx class. We have UtxoStakeDescription now, which is fine, just very slightly inconsistent with the rest of the code base, but still very fine because the coin IS UTXO, so there is no harm. 

so this is just for future - if you see "coin" think about UTXO, that's all. so when you see CoinView, you think some kind of view of UTXOs, and in fact it is some perspective of the UTXO set

noescape00: >so this is just for future - if you see "coin" think about UTXO, that's all. so when you see CoinView, you think some kind of view of UTXOs, and in fact it is some perspective of the UTXO set.

Yep, got it. 

Though not sure why btc core guys came with this name in the first place, that was really confusing.

Aprogiena: It seems so, but then it is not :-) 
If you think about it, these cryptocurrencies have coins, so what exactly is the coin? The coin is something you can prove ownership to and you can send to someone else. In Bitcoin, that's basically UTXO. So after all, UTXO is the coin.

noescape00: That's a good way to put it, actually. Now I see the logic behind it.

PR #756, opened 2017-11-19
Refactor ChainedBlock to add comments and styling.

mikedennis:
Basic styling and commenting - no real structural changes other than I removed
```
public ChainedBlock FindAncestorOrSelf(int height)
```
It was not referenced by anything and does exactly the same thing as ```GetAncestor()``` which is actually used.

### DIFF
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -16,7 +16,7 @@ namespace NBitcoin
         /// <summary>Window length for calculating median time span.</summary>
         private const int MedianTimeSpan = 11;
 
-        /// <summary>The hash of the block, if any.</summary>
+        /// <summary>The hash of the block.</summary>
         public uint256 HashBlock { get; private set; }
 
         /// <summary>Predecessor of this block.</summary>
@@ -31,8 +31,8 @@ namespace NBitcoin
         /// <summary>Integer representation of the <see cref="ChainWork"/>.</summary>
         private BigInteger chainWork;
 
-        /// <summary>Total amount of work (expected number of hashes) in the chain up to and including this block.</summary>
-        public uint256 ChainWork { get {return Target.ToUInt256(this.chainWork);} }
+        /// <summary>Total amount of work in the chain up to and including this block.</summary>
+        public uint256 ChainWork { get { return Target.ToUInt256(this.chainWork); } }
 
         /// <summary>
         /// Constructs a chained block.
@@ -68,7 +68,7 @@ namespace NBitcoin
         /// Constructs a chained block.
         /// </summary>
         /// <param name="header">The header for the chained block.</param>
-        /// <param name="height">The height of the chained block</param>
+        /// <param name="height">The height of the chained block.</param>
         public ChainedBlock(BlockHeader header, int height)
         {
             this.Header = header ?? throw new ArgumentNullException("header");
@@ -79,14 +79,14 @@ namespace NBitcoin
 
 
         /// <summary>
-        /// Calculates the total amount of work (expected number of hashes) in the chain up to and including this block.
+        /// Calculates the total amount of work in the chain up to and including this block.
         /// </summary>
         private void CalculateChainWork()
         {
             this.chainWork = (this.Previous == null ? BigInteger.Zero : this.Previous.chainWork).Add(GetBlockProof());
         }
 
-        /// <summary>Amount of work for this entry in the chain.</summary>
+        /// <summary>Calculates the amount of work that this block contributes to the total chain work.</summary>
         /// <returns>Amount of work.</returns>
         private BigInteger GetBlockProof()
         {
@@ -94,10 +94,10 @@ namespace NBitcoin
             if ((target.CompareTo(BigInteger.Zero) <= 0) || (target.CompareTo(Pow256) >= 0))
                 return BigInteger.Zero;
 
-            // We need to compute 2**256 / (bnTarget+1), but we can't represent 2**256
-            // as it's too large for a arith_uint256. However, as 2**256 is at least as large
-            // as bnTarget+1, it is equal to ((2**256 - bnTarget - 1) / (bnTarget+1)) + 1,
-            // or ~bnTarget / (nTarget+1) + 1.
+            // We need to compute 2**256 / (target+1), but we can't represent 2**256
+            // as it's too large for a uint256. However, as 2**256 is at least as large
+            // as target+1, it is equal to ((2**256 - target - 1) / (target+1)) + 1,
+            // or ~target / (target+1) + 1.
             return ((Pow256.Subtract(target).Subtract(BigInteger.One)).Divide(target.Add(BigInteger.One))).Add(BigInteger.One);
         }
 
@@ -356,7 +356,7 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Check PoW and that the blocks connect correctly.
+        /// Check PoW/PoS and that the blocks connect correctly.
         /// </summary>
         /// <param name="network">The network being used.</param>
         /// <returns>True if PoW is correct.</returns>
@@ -373,7 +373,7 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Check PoW and that the blocks connect correctly.
+        /// Check PoW/PoS and that the blocks connect correctly.
         /// </summary>
         /// <param name="consensus">The consensus being used.</param>
         /// <returns>True if PoW is correct.</returns>
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -122,7 +122,7 @@ namespace NBitcoin
                     nStep *= 2;
             }
 
-            BlockLocator locators = new BlockLocator();
+            var locators = new BlockLocator();
             locators.Blocks = blockHashes;
             return locators;
         }
@@ -352,10 +352,10 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Check PoW/PoS on the network and that the blocks connect correctly.
+        /// Check that the header is a valid block header including the work done for PoW blocks.
         /// </summary>
         /// <param name="network">The network to verify against.</param>
-        /// <returns>True if PoW/Pos is valid.</returns>
+        /// <returns><c>true</c> if the header is a valid block header, <c>false</c> otherwise.</returns>
         public bool Validate(Network network)
         {
             if (network == null)
@@ -372,7 +372,7 @@ namespace NBitcoin
         /// Check PoW against consensus and that the blocks connect correctly.
         /// </summary>
         /// <param name="consensus">The consensus rules being used.</param>
-        /// <returns>True if PoW is valid.</returns>
+        /// <returns><c>true</c> if the header is a valid block header, <c>false</c> otherwise.</returns>
         public bool Validate(Consensus consensus)
         {
             if (consensus == null)
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -16,7 +16,7 @@ namespace NBitcoin
         /// <summary>Window length for calculating median time span.</summary>
         private const int MedianTimeSpan = 11;
 
-        /// <summary>The hash of the block.</summary>
+        /// <summary>The hash of the block which is also known as the block id.</summary>
         public uint256 HashBlock { get; private set; }
 
         /// <summary>Predecessor of this block.</summary>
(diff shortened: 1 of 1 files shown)


src/NBitcoin/ChainedBlock.cs
```diff
@@ -62,11 +64,30 @@ public ChainedBlock(BlockHeader header, uint256 headerHash, ChainedBlock previou
...
         }
 
+        /// <summary>
+        /// Constructs a chained block.
+        /// </summary>
+        /// <param name="header">The header for the chained block.</param>
+        /// <param name="height">The height of the chained block</param>
+        public ChainedBlock(BlockHeader header, int height)
+        {
+            this.Header = header ?? throw new ArgumentNullException("header");
```
fassadlr: Mike, changes this to:

`Guard.NotNull(header , nameof(header ));`

src/NBitcoin/ChainedBlock.cs
```diff
@@ -62,11 +64,30 @@ public ChainedBlock(BlockHeader header, uint256 headerHash, ChainedBlock previou
...
+            this.Height = height;
+            this.HashBlock = header.GetHash();
+            this.CalculateChainWork();
```
fassadlr: We might as well have the logic  of CalculateChainWork in the constructor. I've seen @dangershony request changes like this often ;)

src/NBitcoin/ChainedBlock.cs
```diff
@@ -267,10 +312,11 @@ public Target GetWorkRequired(Consensus consensus)
             }
 
             ChainedBlock firstChainedBlock = this.EnumerateToGenesis().FirstOrDefault(o => o.Height == pastHeight);
-            assert(firstChainedBlock);
+            if (firstChainedBlock == null)
+                throw new NotSupportedException("Can only calculate work of a full chain");
```
fassadlr: Also add a Guard.NotNull here

Aprogiena:
I will check this one as well

src/NBitcoin/ChainedBlock.cs
```diff
@@ -111,19 +126,22 @@ public BlockLocator GetLocator()
                     nStep *= 2;
             }
 
-            var locators = new BlockLocator();
+            BlockLocator locators = new BlockLocator();
```
Aprogiena: we actualy DO use var in this specific case

`X x = new X();` -> we use `var x = new X();`

src/NBitcoin/ChainedBlock.cs
```diff
@@ -305,12 +355,6 @@ public DateTimeOffset GetMedianTimePast()
...
         }
 
-        private static void assert(object obj)
-        {
-            if (obj == null)
-                throw new NotSupportedException("Can only calculate work of a full chain");
-        }
-
         /// <summary>
         /// Check PoW and that the blocks connect correctly.
```
Aprogiena: we need to update the comment of ` public bool Validate(Network network) ` as it is not correct, there is PoS validation as well

src/NBitcoin/ChainedBlock.cs
mikedennis: This is in NBitcoin - no reference to Guard in here.

src/NBitcoin/ChainedBlock.cs
mikedennis: There are two constructors that call this though. I think I might be able to make one constructor call the other with a bit of work though. I will look into it and see if I can figure out something.

src/NBitcoin/ChainedBlock.cs
```diff
@@ -10,43 +10,45 @@ namespace NBitcoin
...
+        /// <summary>Value of 2^256.</summary>
         private static BigInteger Pow256 = BigInteger.ValueOf(2).Pow(256);
 
+        /// <summary>Window length for calculating median time span.</summary>
         private const int MedianTimeSpan = 11;
 
-        // Pointer to the hash of the block, if any. memory is owned by this CBlockIndex.
-        private uint256 hashBlock;
-        public uint256 HashBlock { get { return this.hashBlock; } }
+        /// <summary>The hash of the block, if any.</summary>
```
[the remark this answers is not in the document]
mikedennis: I'm not sure. I can't see where it will be null. That comment was from C++ code. I removed the ",if any" bit for now as I can't see under what condition it's null.

src/NBitcoin/ChainedBlock.cs
```diff
@@ -10,43 +10,45 @@ namespace NBitcoin
...
         private BigInteger chainWork;
-        public uint256 ChainWork
-        {
-            get
-            {
-                return Target.ToUInt256(this.chainWork);
-            }
-        }
 
+        /// <summary>Total amount of work (expected number of hashes) in the chain up to and including this block.</summary>
```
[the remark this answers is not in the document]
mikedennis: That was from C++ code as well. I've removed the bit about the "expected number of hashes"

src/NBitcoin/ChainedBlock.cs
Aprogiena: good point :D

src/NBitcoin/ChainedBlock.cs
Aprogiena: i was thinking of that as well, but not sure how much work that is, if not too much, feel free to do it

mikedennis: Think it's best to leave as two constructors, the logic is a bit different. So going to leave CalculateChainWork in method rather than duplicating the code in each constructor.

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
Aprogiena: ok, never just copy and paste c++ comments :)

src/NBitcoin/ChainedBlock.cs
```diff
@@ -80,14 +101,8 @@ private BigInteger GetBlockProof()
             return ((Pow256.Subtract(target).Subtract(BigInteger.One)).Divide(target.Add(BigInteger.One))).Add(BigInteger.One);
```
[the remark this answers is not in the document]
mikedennis: Fixed it up a bit, see if it makes more sense to you now. Still not sure I understand what it is doing though :grin:

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
Aprogiena: let me check ... so if you go with either constructor, you will have HashBlock calculated for sure and since there is private setter, you can't change it to null ...

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
Aprogiena: let's assume that it does what it claims, so it is supposed to calculate `2**256 / (target+1)`
it claims that we can't represent 2**256 as uint256, which is correct, however, we are using BigInteger, so therefore we can represent it, therefore whole comment is bogus imho

what it claims then is that 2**256 is greater than target + 1, which is true, so in order to compute 2**256 / (target+1) they first subtract one target+1 from it and then they add 1 to the result of that division after that subtraction

so i think we can safely refactor this to really just do `2**256 / (target+1)` without any mumbo jumbo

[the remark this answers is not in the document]
Aprogiena: and once we do that we can remove the comment which is bogus for us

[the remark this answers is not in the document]
Aprogiena: hmm, thinking about it, this is not code from StratisX, wtf then?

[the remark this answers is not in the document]
Aprogiena: do change it and then just try to sync without checkpoints

they are using arith_uint256, which is not as powerful as BigInteger, that's why ...

src/NBitcoin/ChainedBlock.cs
```diff
@@ -356,7 +356,7 @@ public DateTimeOffset GetMedianTimePast()
         }
 
         /// <summary>
-        /// Check PoW and that the blocks connect correctly.
+        /// Check PoW/PoS and that the blocks connect correctly.
         /// </summary>
         /// <param name="network">The network being used.</param>
         /// <returns>True if PoW is correct.</returns>
```
Aprogiena: ee

src/NBitcoin/ChainedBlock.cs
```diff
@@ -373,7 +373,7 @@ public bool Validate(Network network)
         }
 
         /// <summary>
-        /// Check PoW and that the blocks connect correctly.
+        /// Check PoW/PoS and that the blocks connect correctly.
```
Aprogiena: actually this one only checks PoW :D

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
dangershony: >they are using arith_uint256, which is not as powerful as BigInteger

Ah interesting whats the difference?

@mikedennis compact (I assume you mean the target which is Bits) is a calculation that allows to store a smaller representation of a the big int number.
See https://bitcoin.stackexchange.com/questions/36744/why-is-the-target-stored-in-compact-form-in-the-block-header

src/NBitcoin/ChainedBlock.cs
dangershony: If I remember correctly POS validation happens in another place.

src/NBitcoin/ChainedBlock.cs
dangershony: WHAT? I don't believe my eyes...

src/NBitcoin/ChainedBlock.cs
```diff
@@ -10,43 +10,45 @@ namespace NBitcoin
...
+        /// <summary>Value of 2^256.</summary>
         private static BigInteger Pow256 = BigInteger.ValueOf(2).Pow(256);
 
+        /// <summary>Window length for calculating median time span.</summary>
         private const int MedianTimeSpan = 11;
 
-        // Pointer to the hash of the block, if any. memory is owned by this CBlockIndex.
-        private uint256 hashBlock;
-        public uint256 HashBlock { get { return this.hashBlock; } }
+        /// <summary>The hash of the block.</summary>
```
dangershony: The hash of the block, which is also know as the block id.

src/NBitcoin/ChainedBlock.cs
mikedennis: Typo or is another Apro-ism like chmmm?

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
mikedennis: No fear! So I changed and chain still seems to sync and tests all seem happy. So seems like it's ok? I guess I could maybe write a unit test that computed it both ways and compared the results if we worry if its not right but I think I'm the only one worrying. :worried:

src/NBitcoin/ChainedBlock.cs
fassadlr: Ah sorry of course!

src/NBitcoin/ChainedBlock.cs
Aprogiena: this is not full validation, just headers validation with BlockStake.Validate

src/NBitcoin/ChainedBlock.cs
Aprogiena: dude, all i'm doing is just telling what is in the policy, nothing else :D

src/NBitcoin/ChainedBlock.cs
Aprogiena: ee is the sound my baby does when she does not like it :)

what i was commenting about was that return value - in summary you have PoW/PoS but in returns tag you have PoW only (so the idea was to have returns to `<c>true</c> if PoW/PoS is correct.` to have it consistent with summary, but see below) 

anyway, this whole comment is little under-descriptive in the sense that what does it really mean to check PoW/PoS and the method does something else really

so i'd go with summary
```
Check that the header is a valid block header including the work done for PoW blocks.
```

and returns should be 
```
<c>true</c> if the header is a valid block header, <c>false</c> otherwise.
```

BTW, However, I've now looked at where exactly this method is being called and I found this:

Look at CHB, when we receive new headers, we call
```
                    bool validated = this.Chain.GetBlock(tip.HashBlock) != null || tip.Validate(this.AttachedNode.Network);
```

so either the tip (the new header that arrived) is on the best chain OR we call this Validate method

now we do 

```
            if (Block.BlockSignature)
                return BlockStake.Validate(network, this);
```

so i assume this is set when we are on stratis with POS blocks, so we go there to validate POS header

```
		public static bool Validate(Network network, ChainedBlock chainedBlock, StakeChain stakeChain = null)
```

and here it seems that stakeChain is always null, so the method can be simplified. But that is offtopic for this PR, so no worries about it.

src/NBitcoin/ChainedBlock.cs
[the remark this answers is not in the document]
Aprogiena: no fear, you can write a unit test that's fine, but most importantly if we peer review that, it should be ok

src/NBitcoin/ChainedBlock.cs
Aprogiena: sure no problem

src/NBitcoin/ChainedBlock.cs
Aprogiena: this will be for later when we have those guards, so non issue here

PR #759, opened 2017-11-20
[Tests] Fix node setting overriding file configuration

fassadlr:
As stated by @bokobza, argument injection should be preferred over file configuration but currently if we specifiy to use for example, RegTest in the file we will over write the boolean straight after if its NOT in the arguments.

This causes the integration tests to run on the bitcoin network.


bokobza:
Nice one!

Issue #747, opened 2017-11-16
[BaseFeature-ChainRepository] Possible node starting time speedup 


noescape00:
ChainRepository.LoadAsync takes 14 sec to execute (out of total 14.9 sec for node.Start()) because it calculates sha256 hash a few thousand times in order to set ConcurrentChain's Tip. 
Maybe consider cashing tip and starting from that point instead of recalculating all hashes every time app launches?

bokobza:
Is that what's happening? Looking to improve this method has been at the back of my mind for months.
Do yo want to have a go at it @noescape00 ?

noescape00:
@bokobza sure, assign this to me. I'll finish current one that aprogiena suggested (making global comments for a few PoS\PoW classes) and start this one.

bokobza:
![](https://github.trello.services/images/mini-trello-icon.png) [[BaseFeature-ChainRepository] Possible node starting time speedup](https://trello.com/c/NVKTh9CP/323-basefeature-chainrepository-possible-node-starting-time-speedup)

dangershony:
>Maybe consider cashing tip and starting from that point 
Can you explain more what you mean?

noescape00:
>Can you explain more what you mean?

I haven't thought through the solution for this very well, just bumped into it because it significantly slowed down debugging process. 
By a first glance it looks like when the chain gets loaded from dbreeze db every block's hash gets recalculated in order to verify that blocks are valid. So I guess we don't need to verify the same stuff over and over again and we can start from the same point where verification ended on the next launch.

noescape00:
I guess I'll also need to check how loading chain and setting tip was approached by the other projects. I am almost completely sure that they came up with a solution because current approach dont scale. On my machine it takes 14 sec for this process and stratis only has ~600k blocks. So the launching time will increase linearly with the blocks count

dangershony:
Ah I see you mean hashing the header to get the block id, hashing the header I believe has a very small perf impact but if it does not I guess we can defo work around that (by storing to disk the block hash, it will bloat the store a bit but its negligent).
You can easily test that by building a second chain (as headers are cashed so you will get an idea of the perf improvement).

PR #763, opened 2017-11-20
Optimization of ChainRepository.LoadAsync

noescape00:
Issue #747
Used to load for 14 sec (out of total 14.9 for node startup) on my PC, now 40% faster.

### DIFF
--- a/src/Stratis.Bitcoin/Base/ChainRepository.cs
+++ b/src/Stratis.Bitcoin/Base/ChainRepository.cs
@@ -46,8 +46,10 @@ namespace Stratis.Bitcoin.Base
                     transaction.ValuesLazyLoadingIsOn = false;
                     ChainedBlock tip = null;
                     Row<int, BlockHeader> firstRow = transaction.Select<int, BlockHeader>("Chain", 0);
+
                     if (!firstRow.Exists)
                         return;
+
                     BlockHeader previousHeader = firstRow.Value;
                     Guard.Assert(previousHeader.GetHash() == chain.Genesis.HashBlock); // can't swap networks
 
@@ -57,7 +59,6 @@ namespace Stratis.Bitcoin.Base
                             break;
 
                         tip = new ChainedBlock(previousHeader, row.Value.HashPrevBlock, tip);
-
                         previousHeader = row.Value;
                     }
 
--- a/src/Stratis.Bitcoin/Base/ChainRepository.cs
+++ b/src/Stratis.Bitcoin/Base/ChainRepository.cs
@@ -46,7 +46,7 @@ namespace Stratis.Bitcoin.Base
                     transaction.ValuesLazyLoadingIsOn = false;
                     ChainedBlock tip = null;
                     BlockHeader previousHeader = transaction.Count("Chain") != 0 ?
-                        transaction.Select<int, BlockHeader>("Chain", 0).Value : null;
+                        transaction.Select<int, BlockHeader>("Chain", 0)?.Value : null;
                     bool first = true;
 
                     foreach (Row<int, BlockHeader> row in transaction.SelectForwardSkip<int, BlockHeader>("Chain", 1))
--- a/src/Stratis.Bitcoin/Base/ChainRepository.cs
+++ b/src/Stratis.Bitcoin/Base/ChainRepository.cs
@@ -43,21 +43,33 @@ namespace Stratis.Bitcoin.Base
             {
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
+                    transaction.ValuesLazyLoadingIsOn = false;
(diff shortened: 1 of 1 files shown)


Aprogiena:
looks good, but haven't tested it myself

dangershony:
I would suggest to boost perf we try to store the chain not in dbreeze but as a binary file.

Aprogiena:
that's a weirdly interesting idea, but probably rather just weird
OTOH DBreeze itself should be able to do it much faster if we stored bigger blobs, not each header separately

Aprogiena:
let's say we'd have new table that would store block headers not one by one but by 10000 at once. and the current table would just store the blocks that can't fit in there to fill the whole batch

noescape00:
I can do a research about this and optimize this solution as a next task if you wish. Had some experience with bunch of solutions. 
Probably one of the most interesting is to use mix of 2 nosql databases- normal and in memory (like redis)

dangershony:
Its weird at all, if you look at ConcurrentChain there are methods there do do just that (convert to byte arrays) that's how its currently done in NBitcoin. 
Its also much faster, reading 50mb of data as a byte array takes 30ms on my laptop (SSD I know)

The chain does not need indexing capabilities its stored in memory so storing it as a single binary file will not change much, the only time we read write is in start/stop (and periodic task)

dangershony:
>DBreeze itself should be able to do it much faster if we stored bigger blobs
That's also true part of the issue is headers are not batched.

Aprogiena:
the weird part is to implement our own persisting mechanism when we have dbreeze

that's why i say to just use batches to avoid having too many rows
then it will work well imho

src/Stratis.Bitcoin/Base/ChainRepository.cs
```diff
@@ -43,22 +43,30 @@ public Task LoadAsync(ConcurrentChain chain)
             {
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
+                    transaction.ValuesLazyLoadingIsOn = false;
                     ChainedBlock tip = null;
+                    BlockHeader previousHeader = transaction.Count("Chain") != 0 ?
+                        transaction.Select<int, BlockHeader>("Chain", 0).Value : null;
```
Aprogiena: i think this is not a correct way of doing that
you should get Row and check if it exists
just because the table is not empty, it does not mean there is value stored with key 0

src/Stratis.Bitcoin/Base/ChainRepository.cs
```diff
@@ -43,22 +43,27 @@ public Task LoadAsync(ConcurrentChain chain)
...
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
+                    transaction.ValuesLazyLoadingIsOn = false;
                     ChainedBlock tip = null;
-                    bool first = true;
+                    Row<int, BlockHeader> firstRow = transaction.Select<int, BlockHeader>("Chain", 0);
+                    if (!firstRow.Exists)
+                        throw new InvalidOperationException("Block header with index 0 doesn't exist in the database");
+                    BlockHeader previousHeader = firstRow.Value;
+                    Guard.Assert(previousHeader.GetHash() == chain.Genesis.HashBlock); // can't swap networks
```
dangershony: @Aprogiena are you happy with this?

dangershony: What happens if this is a new fresh node will we still have raw zero?

Aprogiena: of course not, one does not just make me happy, you know that ...

but making it better is outside of the scope of this PR, that's like most of the changes we make now - we make incremental changes that improve the code somehow, but don't make it perfect, so it can still be improved later

so in absolute terms - no, i am not happy with this
in relative terms - yes, for this PR, this is good enough

there are two problems in this code
- using assert for user error instead of handling gracefully (e.g. via throw and catch at higher level)
- there is a penalty from rare user error case to 99.99 % of runs, which is wrong by principle - i'd rather place the check AFTER the chain is processed thus saving that one GetHash call, which would save that GetHash call for those 99.99 % of runs and make that 1 faulty run in 10000 wait for the whole chain loading; so the principle is that such a rare case should not affect the major code flow, which is not satisfied here

but again - for this PR, good enough

noescape00: >What happens if this is a new fresh node will we still have raw zero?

It would throw the exception "Block header with index 0 doesn't exist in the database". 
Looks like it wasn't a good idea, fixed in the latest commit.

Aprogiena:
I have some weird things going on while testing this, can't merge yet, there might be a bug.

PR #766, opened 2017-11-21
Feature/issue 735

### LINKED ISSUE #735: [Validation] Invalid block cache vulnerable to attacks from IPv6
The problem is described here - https://github.com/stratisproject/StratisBitcoinFullNode/pull/710#issuecomment-343866080

There are two solutions described there as well. This issue is about implementing the second simpler solution.

Aprogiena:
The collection of invalid block header hashes has been separated from ChainState into its own class. The store of hashes now has a capacity of 1000 entries and once this is reached, oldest entries are removed. This eliminates the attack vector described in #735.

Tests are included.

Fix #735

### DIFF
--- a/src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
+++ b/src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
@@ -1,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using NBitcoin;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Base
+{
+    /// <summary>
+    /// Contract of a store of block header hashes that are considered invalid.
+    /// </summary>
+    public interface IInvalidBlockHashStore
+    {
+        /// <summary>
+        /// Check if a block is marked as invalid.
+        /// </summary>
+        /// <param name="hashBlock">The block hash to check.</param>
+        /// <returns><c>true</c> if the block is marked as invalid, <c>false</c> otherwise.</returns>
+        bool IsInvalid(uint256 hashBlock);
+
+        /// <summary>
+        /// Marks a block as invalid.
+        /// </summary>
+        /// <param name="hashBlock">The block hash to mark as invalid.</param>
+        /// <param name="rejectedUntil">Time in UTC after which the block is no longer considered as invalid, or <c>null</c> if the block is to be considered invalid forever.</param>
+        void MarkInvalid(uint256 hashBlock, DateTime? rejectedUntil = null);
+    }
+
+    /// <summary>
+    /// In memory store of invalid block header hashes.
+    /// </summary>
+    /// <remarks>
+    /// The store has a limited capacity. When a new block header hash is marked as invalid 
+    /// once the capacity is reached, the oldest entry is removed and no longer considered invalid.
+    /// <para>
+    /// Entries with specified expiration time are either removed just like other entries - i.e. during 
+    /// an add operation when a capacity is reached, or they are removed when they are touched 
+    /// and it is detected that their expiration time is no longer in the future.
+    /// </para>
+    /// </remarks>
+    public class InvalidBlockHashStore : IInvalidBlockHashStore
+    {
+        /// <summary>Default value for the maximal number of hashes we can store.</summary>
+        public const int DefaultCapacity = 1000;
+
+        /// <summary>A provider of the date and time.</summary>
+        private readonly IDateTimeProvider dateTimeProvider;
+
+        /// <summary>Lock object to protect access to <see cref="invalidBlockHashes"/> and <see cref="orderedHashList"/>.</summary>
+        private readonly object lockObject = new object();
+
+        /// <summary>
+        /// Collection of blocks that are to be considered invalid. If the value of the entry is not <c>null</c>, 
+        /// the entry is considered invalid only for a certain amount of time.
+        /// </summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly Dictionary<uint256, DateTime?> invalidBlockHashes;
+
+        /// <summary>Circular array of block header hash entries to allow quick removal of the oldest entry once the capacity is reached.</summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly CircularArray<uint256> orderedHashList;
+
+        /// <summary>
+        /// Initializes the instance of the object.
+        /// </summary>
+        /// <param name="capacity">Maximal number of hashes we can store.</param>
+        public InvalidBlockHashStore(IDateTimeProvider dateTimeProvider, int capacity = DefaultCapacity)
+        {
+            this.dateTimeProvider = dateTimeProvider;
+
+            this.invalidBlockHashes = new Dictionary<uint256, DateTime?>(capacity);
+            this.orderedHashList = new CircularArray<uint256>(capacity);
+        }
+
+        /// <inheritdoc />
+        public bool IsInvalid(uint256 hashBlock)
+        {
+            bool res = false;
+
+            lock (this.lockObject)
+            {
+                // First check if the entry exists.
+                DateTime? expirationTime;
+                if (this.invalidBlockHashes.TryGetValue(hashBlock, out expirationTime))
+                {
+                    // The block is banned forever if the expiration date was not set,
+                    // or it is banned temporarily if it was set.
+                    if (expirationTime != null)
+                    {
+                        // The block is still invalid now if the expiration date is still in the future.
+                        res = expirationTime > this.dateTimeProvider.GetUtcNow();
+
+                        // If the expiration date is not in the future anymore, remove the record from the list.
+                        // Note that this will leave entry in the orderedHashList, but that is OK, 
+                        // that entry will be removed later.
+                        if (!res) this.invalidBlockHashes.Remove(hashBlock);
+                    }
+                    else res = true;
+                }
+            }
+
+            return res;
+        }
+
+        /// <inheritdoc />
+        public void MarkInvalid(uint256 hashBlock, DateTime? rejectedUntil = null)
+        {
+            lock (this.lockObject)
+            {
+                DateTime? expirationTime;
+                bool existsAlready = this.invalidBlockHashes.TryGetValue(hashBlock, out expirationTime);
+                if (existsAlready)
+                {
+                    // Entry is existing already, only replace its value if the ban is stronger.
+                    // This happens if the new ban is forever - i.e. rejectedUntil is null,
+                    // or if the existing ban is not forever and the rejectedUntil value is greater 
+                    // than existing expiration time.
+                    bool strongerBan = (rejectedUntil == null) || ((expirationTime != null) && (rejectedUntil.Value > expirationTime.Value));
+                    if (strongerBan) this.invalidBlockHashes[hashBlock] = rejectedUntil;
+                }
+                else
+                {
+                    // No previous entry found, so we add a new ban. This means we can reach 
+                    // the capacity, in which case we first remove the oldest entry.
+                    // Note that there can be entries in the circular array that are no longer in the dictionary.
+                    // We skip all such entries.
+
+                    // We start by adding the new entry to the queue, which will possibly remove the oldest entry if the capacity is reached.
+                    // If capacity has not been reached, there is nothing more to do with the queue.
+                    uint256 oldestEntry;
+                    if (this.orderedHashList.Add(hashBlock, out oldestEntry))
+                    {
+                        // Then we check the dictionary whether it contains the removed entry.
+                        // If not, we remove the next entry from the queue, if there is any,
+                        // and we check the dictionary again. We repeat this until we find 
+                        // an existing entry or until we removed everything from the queue except 
+                        // for the entry we just added.
+                        while (!this.invalidBlockHashes.Remove(oldestEntry))
+                        {
+                            if (this.orderedHashList.Count == 1)
+                                break;
+
+                            this.orderedHashList.RemoveFirst(out oldestEntry);
+                        }
+                    }
+
+                    this.invalidBlockHashes.Add(hashBlock, rejectedUntil);
+                }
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        /// <summary>Lock object to protect access to <see cref="invalidBlockHashes"/> and <see cref="orderedHashList"/>.</summary>
+        private readonly object lockObject = new object();
+
+        /// <summary>
+        /// Collection of blocks that are to be considered invalid. If the value of the entry is not <c>null</c>, 
+        /// the entry is considered invalid only for a certain amount of time.
+        /// </summary>
+        /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
+        private readonly Dictionary<uint256, DateTime?> invalidBlockHashes;
```
fassadlr: Isn't it better to make this a ConcurrentDictionary? Thus removing the need for a lock?

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
```diff
@@ -0,0 +1,154 @@
...
+            this.orderedHashList = new CircularArray<uint256>(capacity);
+        }
+
+        /// <inheritdoc />
+        public bool IsInvalid(uint256 hashBlock)
+        {
+            bool res = false;
+
+            lock (this.lockObject)
+            {
```
fassadlr: We can remove this lock if you use a ConcurrentDictionary.

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            return res;
+        }
+
+        /// <inheritdoc />
+        public void MarkInvalid(uint256 hashBlock, DateTime? rejectedUntil = null)
+        {
+            lock (this.lockObject)
+            {
```
fassadlr: The same here as with the other lock.

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <inheritdoc />
+        public bool IsInvalid(uint256 hashBlock)
+        {
+            bool res = false;
+
+            lock (this.lockObject)
+            {
+                // First check if the entry exists.
+                DateTime? expirationTime;
+                if (this.invalidBlockHashes.TryGetValue(hashBlock, out expirationTime))
```
fassadlr: I think that the name "invalidBlockHashes" is confusing when used in this context. I.e you are calling a TryGetValue on a field called "invalidBlockHashes" but returning a date? Can we maybe rename this to something more descriptive like invalidBlockHashesByExpiry or invalidBlockExpiryHashes ?

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
```diff
@@ -0,0 +1,154 @@
...
+            this.dateTimeProvider = dateTimeProvider;
+
+            this.invalidBlockHashes = new Dictionary<uint256, DateTime?>(capacity);
+            this.orderedHashList = new CircularArray<uint256>(capacity);
+        }
+
+        /// <inheritdoc />
+        public bool IsInvalid(uint256 hashBlock)
+        {
+            bool res = false;
```
fassadlr: This might be pedantic of me but can we name variables like this properly? I.e. "result", its just if we start shortening variables then whats stopping us from doing it everywhere? Lets keep naming consistent.

src/Stratis.Bitcoin/Utilities/CircularArray.cs
```diff
@@ -75,6 +75,26 @@ public bool Add(T item, out T oldItem)
...
         }
 
+        /// <summary>
+        /// Removes the first item from the circular array, which is the oldest entry in the array.
+        /// </summary>
+        /// <param name="firstItem">If the function returns <c>true</c>, this is filled with the oldest item that was removed.</param>
+        /// <returns><c>true</c> if the oldest item was removed, <c>false</c> if there were no items.</returns>
+        public bool RemoveFirst(out T firstItem)
+        {
+            bool res = false;
```
fassadlr: Same here with naming 👍

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: no, there are two structures that are synchronized, so we need a lock

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: we have a lot of `res` in other code base, thus seems fine
we also have a lot of "tx" and "txes" instead of "transaction(s)" 
i think this is consistent with what we have elsewhere

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: you are right, the current name is not perfect

but it is actually more like invalidBlockHashesExpirationTimesByBlockHashes or "expirations-by-hashes", rather than "hashes-by-expiration", so the other way around, so i am trying to find a sensible name for it ...

maybe invalidBlockHashesData or invalidBlockHashesInfo, what do you think?

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
fassadlr: Yes but only in the ```MarkInvalid``` method. The ```IsInvalid``` method does not require a lock if you use a Concurrent dictionary. Its more appropriate to use ConcurrentDictionary if you can do so 👍

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
fassadlr: That is exactly my point @Aprogiena ;) I don't think its good to continue that practice of shortening variables, so lets start fixing them going forward.

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
fassadlr: Hmmmm what about invalidBlockExpirationsByHash? Then we know we return an expiry when we TryGetValue?

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: ah i see, but i don't really agree :-) i quite like short names when it makes sense and does not change the meaning or is not confusing

so let me dodge the bullet by saying, i'd need further consensus to be happy with that

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: i disagree

CD has its own lock, so what you are saying is that you want 2 locks instead of 1

but most importantly, the logic in MarkInvalid is nontrivial and i can't see how you can achieve it easily with CD 

what i'm doing ther is that i'm updating an entry if the existing entry has some specific properties, this is something CD does not support thread-safely

i can't examine CD item, make a test on that value and then change it in one atomic step with just internal CD lock

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: personally i like class names to be little more explanatory, not that much for local variables

for example, i'm mostly perfectly happy with `i` in a (for) loop, only when clarity is better with something like `somethingIndex` i use it, but using `index` or `iterationIndex` every time i would not like

i there are many cases where short naming is perfectly fine and can't be confused - `tx` is always transaction and `res` is always result and `i` is always iteration index, etc. etc. i think there are many many cases where i'd rather save my wrists from typing more :-) and it won't hurt anything imho

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
fassadlr: OK yeah, leave as is, I understand now what you are trying to do @Aprogiena 👍

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: i am thinking about this, this is a classic example and i can see where the confusion is coming from

in general, this is classic situation when you have ID and RECORD, so we have Dictionary(ID, RECORD) and RECORD is 
```
public class RECORD
{
  public Type1 ID;
  public Type2 Data1;
  public Type3 Data2;
  public Type4 Data3;
  ...
}

and we would name such a things as "private Dict<ID, RECORD> RECORD records;"
```

and in this case ID is uint256 and is actually the most important thing in the record
and the only other data we have for it is that `DateTime?`

so instead of creating this class holding the whole record information, we take a shortcut and say that value in the dict is just `DateTime?`, but in fact what we want is the whole structure `{ uint256, DateTime? }` and we have it when we work with the dict

the hash here is the primary purpose, DateTime is secondary (but still very important), this is why i don't really like invalidBlockExpirationsByHash because the pattern is invalidRECORD-By-ID
and as explained, it is not the DateTime that is the whole record here

so this is why I am more for "invalidBlockHashesData" or "invalidBlockHashesInfo"

Aprogiena: and that's why the current name is not correct as you pointed out correctly, because the RECORD here is not just the hash, so i very much agree with renaming, just not sure to what

@dangershony sometimes have good ideas for names, could he help possibly?

my best candidates at this moment are  "invalidBlockHashesData" or "invalidBlockHashesInfo"

fassadlr: How about we introduce a small class that holds that data? Something called InvalidBlockHashExpiry which holds the hash as well as the expiry date?

```
public class InvalidBlockHashExpiry 
{
    public uint BlockHash {get;set;}
    public DateTime Expiry {get;set;}
}
```

Then you have a list instead of a dictionary?

Or will this be too much overhead?

Aprogiena: actually the whole trouble was to find the name for such a class, because that is immediately the name of the dictionary

dictionary is used for O(1) access, so that should stay, 

but `invalidBlockHashExpirations` is probably a good name for the dictionary, are you happy with that?

fassadlr: I'm happy with ```invalidBlockHashExpirations``` @Aprogiena 👍

Aprogiena:
i've just merged the latest master to this branch to see if i can get green build

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
dangershony: @fassadlr I tend to prefer the Concurrent ns merely for simplicity of the code, but apro is more of a `lock` guys both are fine.

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
dangershony: I tend to also prefer more explanatory names (even though our repo is full of `res`)
So to me its not a big deal.
I would whoever tend to prefer to use `index` over `i` come to think about it.

fassadlr: @dangershony and @Aprogiena can we then decide on what to use throughout? We can't have variable naming short in some areas and explanatory in other areas? I dont mind either way, just want consistency 👍

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
dangershony: wowoo I was fine with any name to be honest as this whole class is for expired blocks. 
My favourite was `invalidBlockHashesByExpiry` but no big deal.

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: i'd be happy with a rule such as `variable/class name must not be confusing, or too similar to some other variable/class that is unrelated (so singular and plural variables are OK like foreach (item in items), the name must be explanatory. and then if the variable/class name is as such, it should be as short as possible`

with such a rule, in almost all code (unless there are conflicts with something else) `res` is better than `result`, `tx` is better than `transaction`, `id` is better than `identifier`, `txid` is better than `transactionIdentifer`

fassadlr: @Aprogiena omg this discussion is getting out of hand. It would've been so much easier if you just renamed the variable. Then at least we have some consistency. Cause now, with your way, we need to define what variables can be shortened or not. (And I'm not referring to "for" or while variables etc). @dangershony

Aprogiena:
ok, i've confirmed that my issue had nothing to do with this PR and can be seen with master, so this is unreleated

src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs
Aprogiena: @fassadlr  i understand what you mean, but i just don't agree with you :-) so i'm trying to avoid such a change by thinking about it. and so far as i've thought about it i find that i'm ok with my original position, so i still don't want the change :-)

Aprogiena: there are many things that i just do by intuition and have never really supported those things with any rational arguments that would make sense to me. so it makes sense to me, if something is a source of controversy, to sort my thoughts about it and either change my opinion or not. here i still consider my naming as rational and better than the proposed alternative

PR #767, opened 2017-11-21
 Added convenience methods for tests wishing to create a folder used during execution

bokobza:
@Neurosploit Is that going to clash with something you're doing?


Neurosploit:
@bokobza Not at all I'm happy with it. I'll apply it to indexstore while I'm at it because I've about done the same.

PR #774, opened 2017-11-22
fixed compiler warnings by removing some stuff from nbitcoin

Aprogiena:
ever since we moved NBitcoin to FN codebase, we received extra 3 compiler warnings, this PR fixes those

Start and StartAsync were removed from node builder in nbitcoin tests
it was never used and it used uninitialized variables anyway ...

### DIFF
--- a/src/NBitcoin.Tests/NodeBuilderStratis.cs
+++ b/src/NBitcoin.Tests/NodeBuilderStratis.cs
@@ -109,8 +109,6 @@ namespace NBitcoin.Tests
 			//}
 			var node = new CoreNodeStratis(child, this);
 			Nodes.Add(node);
-			if(start)
-				node.Start();
 			return node;
 		}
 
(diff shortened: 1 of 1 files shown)


src/NBitcoin.Tests/NodeBuilderStratis.cs
```diff
@@ -109,8 +109,6 @@ public CoreNodeStratis CreateNode(bool start = false)
 			//}
 			var node = new CoreNodeStratis(child, this);
 			Nodes.Add(node);
-			if(start)
```
dangershony: is it never used?

src/NBitcoin.Tests/NodeBuilderStratis.cs
```diff
@@ -263,44 +247,7 @@ public Node CreateNodeClient(NodeConnectionParameters parameters)
 		}
 #endif
 
-		public async Task StartAsync()
```
dangershony: Same question.

src/NBitcoin.Tests/NodeBuilderStratis.cs
Aprogiena: the code compiles, the code runs, all calls to here were with start = false ... it is not used because if it was, those uninitialized variables i removed would probably caused troubles, + this is some nbitcoin test only ... we do have our own tests for running nodes and + we will redesign node class ...

PR #778, opened 2017-11-22
Consensus banning

dangershony:


bokobza:
@dangershony can you please resolve the conflicts first?

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -220,13 +244,12 @@ public void Stop()
         /// The block will then be passed to the consensus validation. 
         /// </summary>
         /// <remarks>
-        /// If the <see cref="Block"/> returned from the puller is null that means the puller is signalling a reorg was detected.
+        /// If the <see cref="Block"/> returned from the puller is null that means the puller is signaling a reorg was detected.
```
Aprogiena: are we using US not British spelling?

Aprogiena: `is <c>null</c>`

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -18,11 +18,12 @@ namespace Stratis.Bitcoin.Connection
     public interface IPeerBanning
     {
         /// <summary>
-        /// Set a peer as banned.
+        /// Set a peer as banned, if<see cref="banTimeSeconds"/> is <c>null</c> the default ban time is <see cref="ConnectionManagerSettings.DefaultMisbehavingBantimeSeconds"/>.
```
Aprogiena: missing space after if

Aprogiena: and this should be paramref not see

dangershony:
>i'm not sure about banTimeSeconds being nullable, i usually prefer avoiding nullable types when possible,
The idea was to avoid making too many changes to the code but having a default value represent an operation.
The plan is to use the default `ConnectionManagerSettings.BanTimeSeconds` and allow at the point of the error to override this either by a different time or by setting the ban to zero (disabled). 
What you suggest means I will have to set the default value of the ban time at creation of context, which perhaps is not a bad idea.

I could preset the ban duration to default in the method `PullerLoop`
```
      BlockValidationContext blockValidationContext = new BlockValidationContext {  BanDurationSeconds = this.connectionManagerSettings.BanTimeSeconds };
```
However that means ConnectionSettings will have to be injected in to consensus, which I guess is ok.

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
```diff
@@ -3,151 +3,177 @@
...
         public override int GetHashCode()
         {
             return this.Code.GetHashCode();
         }
 
+        /// <inheritdoc />
         public override string ToString()
         {
-            return this._Code + ": " + this._Message;
+            return $"{this.Code}: {this.Message}";
```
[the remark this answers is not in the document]
dangershony: This is really not called much only on errors, but good point I will change to string format

Aprogiena:
Are you sure? I mean if you introduce the Unban, then you can use value of zero as default and you just got rid of nullable for free, or not?

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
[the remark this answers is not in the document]
Aprogiena: i know, just a matter of general principle

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: >are we using US not British spelling

not sure either is fine I guess

dangershony:
@Aprogiena this last commit also acts on your comments

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            block.Header.HashPrevBlock = context.Chain.Tip.HashBlock;
+            await context.Consensus.AcceptBlockAsync(new BlockValidationContext { Block = block, Peer = peer });
+
+            Assert.False(context.PeerBanning.IsBanned(peer));
+        }
+
+        [Fact]
+        public async Task GivenANodeIsSyncedAndConnected_WhenAPeerSendsABadBlockAndErrorIsNotBanError_ThePeerIsNotBanned_Async()
+        {
+            string dataDir = Path.Combine("TestData", nameof(PeerBanningTest), nameof(this.GivenANodeIsSyncedAndConnected_WhenAPeerSendsABadBlockAndErrorIsNotBanError_ThePeerIsNotBanned_Async));
```
Aprogiena: this aint end good on disk 
those folders will be too long

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            node.Behaviors.Add(connectionManagerBehavior);
+            context.MockReadOnlyNodesCollection.Setup(s => s.FindByEndpoint(It.IsAny<IPEndPoint>())).Returns(node);
+
+            var blocks = await TestChainFactory.MineBlocksAsync(context, 2, new Key().ScriptPubKey);
+            // create a new block that breaks consensus.
+            var block = blocks.First();
+            block.Header.HashPrevBlock = context.Chain.Tip.HashBlock;
+            await context.Consensus.AcceptBlockAsync(new BlockValidationContext { Block = block, Peer = peer, BanDurationSeconds = BlockValidationContext.BanDurationNoBan });
+
+            Assert.False(context.PeerBanning.IsBanned(peer));
```
Aprogiena: how exactly is this test different from the previous test?

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            var blocks = await TestChainFactory.MineBlocksAsync(context, 2, new Key().ScriptPubKey);
+            // create a new block that breaks consensus.
+            var block = blocks.First();
+            block.Header.HashPrevBlock = context.Chain.Tip.HashBlock;
+            await context.Consensus.AcceptBlockAsync(new BlockValidationContext { Block = block, Peer = peer, BanDurationSeconds = 1}); // ban for 1 sec
+            
+            // wait 1 sec for ban to expire.
+            Thread.Sleep(1000);
+
+            Assert.False(context.PeerBanning.IsBanned(peer));
```
Aprogiena: should this test body not be in previous test? 
your test method names do not match what the test is doing actually, check last 3 tests here

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
```diff
@@ -3,181 +3,177 @@
...
         {
             this.ConsensusError = error;
         }
 
+        /// <summary>The error that triggered this exception. </summary>
         public ConsensusError ConsensusError { get; private set; }
     }
 
+    /// <summary>
+    /// A consensus error used to specify different types of reasons a block does not confirm to the consensus rules.
```
Aprogiena: IS used to specify?

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
```diff
@@ -3,181 +3,177 @@
...
 
+        /// <inheritdoc />
         public override bool Equals(object obj)
         {
             ConsensusError item = obj as ConsensusError;
-            if (item == null)
-                return false;
-            return this.Code.Equals(item.Code);
+
+            return (item != null) && (this.Code.Equals(item.Code));
```
Aprogiena: your `==` and `equals` implementation differs, is that intentional? i.e. is it intentional that you don't have `this.Code == item.Code` here?

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
```diff
@@ -3,181 +3,177 @@
...
         public override int GetHashCode()
         {
             return this.Code.GetHashCode();
         }
 
+        /// <inheritdoc />
         public override string ToString()
         {
-            return this.Code + ": " + this.Message;
+            return string.Format("{0} : {1}", this.Code, this.Message);
```
Aprogiena: `+` is faster than string.Format ... we've discussed this already :)

Aprogiena: and also before it was `{0}: {1}` now you introduced ugly space before `:`, use the old version here anyway

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -104,7 +127,13 @@ public class ConsensusLoop
         private readonly ICheckpoints checkpoints;
 
         /// <summary>Consensus settings for the full node.</summary>
-        private readonly ConsensusSettings settings;
+        private readonly ConsensusSettings consensusSettings;
+
+        /// <summary>Settings for the full node.</summary>
+        private readonly NodeSettings nodeSettings;
+
+        /// <summary>Handle the banning of peers.</summary>
```
Aprogiena: Handles

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -125,7 +154,9 @@ public class ConsensusLoop
         /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="signals">A signaler that used to signal messages between features.</param>
         /// <param name="checkpoints">Provider of block header hash checkpoints.</param>
-        /// <param name="settings">Consensus settings for the full node.</param>
+        /// <param name="consensusSettings">Consensus settings for the full node.</param>
+        /// <param name="nodeSettings">Settings for the full node.</param>
+        /// <param name="peerBanning">Handle the banning of peers.</param>
```
Aprogiena: s

src/Stratis.Bitcoin.Features.Notifications/BlockNotification.cs
```diff
@@ -125,12 +125,12 @@ internal Task Notify(CancellationToken token)
             {
                 token.ThrowIfCancellationRequested();
 
-                var block = this.Puller.NextBlock(token);
-                if (block != null)
+                var lookaheadResult = this.Puller.NextBlock(token);
```
Aprogiena: unvar

src/Stratis.Bitcoin/BlockPulling/LookaheadBlockPuller.cs
```diff
@@ -47,7 +48,25 @@ public interface ILookaheadBlockPuller
...
+        /// <summary>
+        /// The downloaded <see cref="Block"/> that was requested by the puller.
+        /// </summary>
+        public Block Block { get; set; }
+
+        /// <summary>
+        /// The peer this block came from.
+        /// </summary>
+        public IPEndPoint Peer { get; set; }
+
```
Aprogiena: remove empty line

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -19,11 +18,12 @@ namespace Stratis.Bitcoin.Connection
...
     {
         /// <summary>
-        /// Set a peer as banned.
+        /// Set a peer as banned, if <paramref name="banTimeSeconds"/> is <c>zero</c> the default ban time is <see cref="ConnectionManagerSettings.DefaultMisbehavingBantimeSeconds"/>.
         /// </summary>
         /// <param name="endpoint">The endpoint to set that it was banned.</param>
+        /// <param name="banTimeSeconds">The time in seconds this peer should be banned.</param>
         /// <param name="reason">An optional reason for the ban, the 'reason' is only use for tracing.</param>
-        void BanPeer(IPEndPoint endpoint, string reason = null);
+        void BanPeer(IPEndPoint endpoint, int banTimeSeconds = 0, string reason = null);
```
Aprogiena: shouldn't this 0 be that constant?

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -101,13 +101,15 @@ public PeerBanning(IConnectionManager connectionManager, ILoggerFactory loggerFa
...
             this.dateTimeProvider = dateTimeProvider;
             this.connectionManagerSettings = nodeSettings.ConnectionManager;
+
+            // TODO: MemoryBanStore should be replaced with the address manager store
+            this.banStore = new MemoryBanStore();
         }
 
         /// <inheritdoc />
-        public void BanPeer(IPEndPoint endpoint, string reason = null)
+        public void BanPeer(IPEndPoint endpoint, int banTimeSeconds = 0, string reason = null)
```
Aprogiena: shouldn't this 0 be that constant?

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -120,16 +122,19 @@ public void BanPeer(IPEndPoint endpoint, string reason = null)
...
+                {
+                    banPeer = false;
+                    this.logger.LogTrace("Peer '{0}' is whitelisted, for reason '{1}' it was not banned!", endpoint, reason ?? "unknown");
+                }
             }
 
             if (banPeer)
             {
-                this.banStore.BanPeer(endpoint, this.dateTimeProvider.GetUtcNow().AddSeconds(this.connectionManagerSettings.BanTimeSeconds));
+                this.banStore.BanPeer(endpoint, this.dateTimeProvider.GetUtcNow().AddSeconds(banTimeSeconds > 0 ? banTimeSeconds : this.connectionManagerSettings.BanTimeSeconds));
```
Aprogiena: hmm the logic is unclear here
in consensus loop you have explained in comments that those constants are special
but here you only default to 0, but you don't mention it is somehow special
so i'm confused, is it here -1 and 0 also special or not?

if yes then -1 here would actually default to BanTimeSeconds which is probably not intended
i think you should use the constants here as well not `= 0` and `> 0`

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
dangershony: `+` is faster? we did not discuss concatenation this only interpolation.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
dangershony: Ok this actually raises the questiosn should we put this logic here at all.
Maybe just check and if  `banTimeSeconds== 0` then don't ban at all

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
dangershony: yes `this.Code == item.Code` will trigger a recursive loop on on the `==` operand just bellow.

dangershony: This code can actually be removed all together we don't need operand overrides really.
But maybe as another PR.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
dangershony: FYI on the interface there is an explanation on the behaviour of the zero.

dangershony: Ok as suggested I will move the constants in to the PeerBan class that will work for all cases.

dangershony: Hmm I am actually not so happy about moving the constants in to the banning class as they represents actions in consensus, I would rather leave them there and not have unban logic in the PeerBan class.

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
Aprogiena: yes, + is faster, but logging is different case

in logging string.Format is faster because if the logging is not at that level the string is not formatted, whereas if we used + it would be processed

so more exactly ... + would be faster even in logging, but only if the logging actually happened. if logging did not happen, string.Format is faster, which is what we care about in logging

in ToString() methods, + is faster

src/Stratis.Bitcoin/Connection/PeerBanning.cs
Aprogiena: on interface i see this:
`        /// <param name="banTimeSeconds">The time in seconds this peer should be banned.</param>`

so i don't see any special values there

Aprogiena: ok new version, checking ..

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
Aprogiena: i can see the recursive loop here, is Object.ReferenceEquals calling Equals?

src/Stratis.Bitcoin.Features.Consensus/ConsensusErrors.cs
dangershony: ok fair enough but this is not a critical path so I am fine with keeping string.Format in this case.

Aprogiena: i know, it is just another best practice
so i'd also be happy for this to be sF, but why i did mention this and actually wanted the change is because you changed it. you just shouldn't touch it :) 
so i mean, if you touch it, improve it, not make it worse :-)

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -201,9 +238,9 @@ public class ConsensusLoop
 
             this.asyncLoop = this.asyncLoopFactory.Run($"Consensus Loop", async (token) =>
             {
-                await this.PullerLoopAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
-            },
-            this.nodeLifetime.ApplicationStopping,
+                await this.PullerLoopAsync().ConfigureAwait(false);
+            }, 
+            this.nodeLifetime.ApplicationStopping,
```
Aprogiena: everyone else is removing these end line spaces ...

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -356,6 +395,12 @@ public void Stop()
                     this.Chain.SetTip(this.Tip);
                     this.logger.LogTrace("Chain reverted back to block '{0}'.", this.Tip);
 
+                    if ((blockValidationContext.Peer != null) && (blockValidationContext.BanDurationSeconds > BlockValidationContext.BanDurationNoBan))
```
Aprogiena: should be `!=` not `>`

dangershony: Well its a safe guard really, in this case `>` will cover all cases.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
dangershony: You mean put it in one line?

Aprogiena:
ok, let's do this

PR #779, opened 2017-11-22
Comments and minor refactoring for StakeValidator and PowMining

noescape00:
Since I've reorganized functions in StakeValidator to be in the following order: 
1) public static
2) public
3) private static
4) private
its better to review this file's changes by downloading the pr instead of looking at the diff, it looks kind of weird.


src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -45,13 +48,137 @@ static StakeValidator()
...
+            if ((target.CompareTo(BigInteger.Zero) <= 0) || (target.CompareTo(targetLimit) >= 1))
+                target = targetLimit;
+
+            var finalTarget = new Target(target);
+            clogger.LogTrace("(-):'{0}'", finalTarget);
+            return finalTarget;
+        }
+
+        /// <summary>
+        /// Checks if provided <see cref="tx"/> is a valid coinstake transaction.
```
dangershony: Can you rename `tx` to transaction if you are already here?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,27 +50,35 @@ namespace Stratis.Bitcoin.Features.Miner
...
         }
 
+        /// <summary>
+        /// Information needed for getting staking UTXOs and signing blocks.
+        /// </summary>
         public class WalletSecret
         {
+            /// <summary> Wallet's password that is needed for getting wallet's private key which is used for signing generated blocks.</summary>
             public string WalletPassword;
+            /// <summary> Name of the wallet which UTXOs are used for staking.</summary>
```
dangershony: Spaces

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -19,20 +22,20 @@ public class StakeValidator
...
 
-        /// <summary>Specification of the network the node runs on - regtest/testnet/mainnet.</summary>
-        private readonly Network network;
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
 
         /// <summary>Database of stake related data for the current blockchain.</summary>
         private readonly StakeChain stakeChain;
 
+        /// <summary>Cheain of headers.</summary>
```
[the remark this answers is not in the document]
Aprogiena: Chain not cheain

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -45,13 +48,137 @@ static StakeValidator()
...
+        /// <param name="consensus">Consensus rules for the current network.</param>
+        /// <param name="proofOfStake"><c>true</c> for calculation of PoS difficulty target, <c>false</c> for calculation of PoW difficulty target.</param>
+        /// <returns>The difficulty target for the next block after <paramref name="chainedBlock"/>.</returns>
+        /// <remarks>
+        /// The calculation of the next target is based on the last target value and the block time (aka spacing) of <paramref name="chainedBlock"/>
+        /// (i.e. difference in time stamp of this block and its immediate predecessor). The target changes every block and it is adjusted
+        /// down (i.e. towards harder to reach, or more difficult) if the time to mine last block was lower than the target block time.
+        /// And it is adjusted up if it took longer than the target block time. The adjustments are done in a way the target is moving towards
+        /// the target spacing exponentially, so even a big change in the mining power on the network will be fixed by retargeting relatively quickly.
```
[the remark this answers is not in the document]
noescape00: Me neither- it's an old one, I just added xmldoc for it.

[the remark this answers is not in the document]
dangershony: I see so small fix, I think this is correct.
>The adjustments are done in a way where the target is moving towards the target-spacing exponentially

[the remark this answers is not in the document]
Aprogiena: target spacing is what normal ppl would call block time, or maybe rather the expected block time, or the target block time we want to achieve. if you want to write it as target-spacing, whatever

the idea is that once you are off your target, you change the difficulty, so that if no more and no less stake power is on the network next time, you would expect the next block time to be closer to your target - this forms a series which converges to the target and the speed of the converging is exponential

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
Aprogiena: yeah i like the descriptions to use `see` only if necessary. If you use English it is often more readable, so if you write `Checks if the provided transaction is a valid coinstake transaction.` that's imho better

however, it seems that you are missing `param` comments. which are mandatory

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -92,40 +219,78 @@ public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPre
...
-            this.logger.LogTrace("(-):{0}", res);
-            return res;
+            return stakeModifier;
         }
 
-        private bool VerifySignature(UnspentOutputs txFrom, Transaction txTo, int txToInN, ScriptVerify flagScriptVerify)
+        /// <summary>
+        /// Checks that staking UTXO is valid for staking and calls <see cref="CheckStakeKernelHash"/>.
+        /// </summary>
+        public void CheckKernel(ContextInformation context, ChainedBlock pindexPrev, uint nBits, long nTime, OutPoint prevout, ref long pBlockTime)
```
Aprogiena: again param comments here
also pls remove hungarian notation - nBits, nTime, pindexPrev, pBlockTime ...
-> headerBits, transactionTime,  previousHeader (or prevChainedBlock), blockTime

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +308,78 @@ private static uint256 ToUInt256(BigInteger input)
...
+            var checker = new TransactionChecker(txTo, txToInN, output.Value, txData);
+            var ctx = new ScriptEvaluationContext { ScriptVerify = flagScriptVerify };
+
+            bool res = ctx.VerifyScript(input.ScriptSig, output.ScriptPubKey, checker);
+            this.logger.LogTrace("(-):{0}", res);
+            return res;
+        }
+
+        /// <summary>
+        /// Stratis kernel protocol. Checks the stake kernel hash.
```
Aprogiena: i'd remove `Stratis kernel protocol.` it means nothing to me
and the actual summary could be little more descriptive - like `Checks that the stake kernel hash satisfies the target difficulty.`
also consider adding `<throws>` tag

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +308,78 @@ private static uint256 ToUInt256(BigInteger input)
...
+        /// <item>stakeModifierV2: Scrambles computation to make it very difficult to precompute future proof-of-stake.</item>
+        /// <item>stakingCoins.Time: Time of the coinstake UTXO. Slightly scrambles computation.</item>
+        /// <item>prevout.Hash: Hash of stakingCoins UTXO, to reduce the chance of nodes generating coinstake at the same time.</item>
+        /// <item>prevout.N: Output number of stakingCoins UTXO, to reduce the chance of nodes generating coinstake at the same time.</item>
+        /// <item>transactionTime: Timestamp of the coinstake transaction.</item>
+        /// </list>
+        /// Block or transaction tx hash should not be used here as they can be generated in vast
+        /// quantities so as to generate blocks faster, degrading the system back into a proof-of-work situation.
+        /// </para>
+        /// </remarks>
```
Aprogiena: in general, those old comments are old, they are not necessary good comments, so i'd prefer rewriting them. just styling them like this is better than nothing, but could be much better
also you are lacking lots of `<see>`s here

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,27 +50,35 @@ namespace Stratis.Bitcoin.Features.Miner
...
-        }
-
+        /// <summary>
+        /// Information related to UTXO that is required for staking.
+        /// </summary>
         public class UtxoStakeDescription
         {
+            /// <summary>Block's hash.</summary>
             public uint256 HashBlock;
+            /// <summary>UTXO.</summary>
```
Aprogiena: i'd very much like empty line before each new XMLDOC comment of unrelated properties/fields
the only exception where i tend NOT to have an empty line before the new XMLDOC is where i have a private field and then a property with getter that returns that field, like 

```
/// <summary>..something..</summary>
private int x;
/// <summary>..something..</summary>
public int X { get { return x; } };
```
in which case it feels that those two guys are so tightly coupled that they don't deserve an empty space

also all these guys should be properties according to our coding style -> add { get; set; } to all of them

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -164,45 +173,54 @@ public class CoinstakeContext
...
+        /// <summary>Mining loop.</summary>
         private IAsyncLoop mining;
+        /// <summary>Balance that shouldn't participate in taking.</summary>
         private Money reserveBalance;
+        /// <summary>UTXO's value threshold for being selected for staking.</summary>
         private readonly int minimumInputValue;
+        /// <summary>Time in milliseconds between attempts to generate PoS blocks.</summary>
         private readonly int minerSleep;
 
+        /// <summary>Used for asynchronously calling <see cref="mempool"/>.</summary>
```
Aprogiena: that does not make any sense actually
lock description should describe lock contract 
and the purpose of locks is not to async call something
the purpose of the lock is to have an exclusive access or in case of RW lock, to prevent readers and writers to operate at the same time

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -92,40 +219,78 @@ public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPre
             this.logger.LogTrace("(-)[OK]");
         }
 
-        private bool IsConfirmedInNPrevBlocks(UnspentOutputs utxoSet, ChainedBlock pindexFrom, long maxDepth)
+        /// <summary>
+        /// Stake Modifier (hash modifier of proof-of-stake):
```
Aprogiena: the purpose of the stake modifier should rather be in the big picture comment, not here
this method does not define the purpose, so its summary should describe what the method does

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -92,40 +219,78 @@ public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPre
...
+        /// The purpose of stake modifier is to prevent a txout (coin) owner from computing future proof-of-stake
+        /// generated by this txout at the time of transaction confirmation. To meet kernel protocol, the txout
+        /// must hash with a future stake modifier to generate the proof.
+        /// </summary>
+        /// <param name="prevChainedBlock">Previous chained block.</param>
+        /// <param name="blockStakePrev">Previous PoS block.</param>
+        /// <param name="kernel">The PoS kernel.</param>
+        public uint256 ComputeStakeModifierV2(ChainedBlock prevChainedBlock, BlockStake blockStakePrev, uint256 kernel)
```
Aprogiena: return value uncommented

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -19,20 +28,22 @@ public class StakeValidator
...
 
-        /// <summary>Specification of the network the node runs on - regtest/testnet/mainnet.</summary>
-        private readonly Network network;
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
 
         /// <summary>Database of stake related data for the current blockchain.</summary>
         private readonly StakeChain stakeChain;
 
+        /// <summary>Chain of headers.</summary>
```
Aprogiena: we don't have this one described well in the code base, so i'd probably be most happy about

`<summary>Thread safe access to the best chain of block headers from genesis.</summary>`

- it is important for me to know it is THE best chain the node is aware of
- it is supposed to be thread safe 
- goes down to the genesis

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -19,20 +28,22 @@ public class StakeValidator
...
         private readonly ConcurrentChain chain;
+
+        /// <summary>Used for getting UTXOs.</summary>
         private readonly CoinView coinView;
```
Aprogiena: same here, we do not have any good common on this one, let's go with
`<summary>Consensus' view of UTXO set.</summary>`

the thing itself is just UTXO set, but we add information that the consensus component defines what is understood as the current UTXO set - when you sync the node, you can see you have different progression of headers, consensus, block store and wallet

and the UTXO set is defined by consensus, so it is probably worth mentioning

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -45,25 +56,156 @@ static StakeValidator()
...
+        /// PoS if <see cref="proofOfStake" /> is <c>true</c> or PoW if <see cref="proofOfStake" /> is <c>false</c>.
+        /// </summary>
+        /// <param name="stakeChain">Chain that is used for retrieving blocks.</param>
+        /// <param name="index">Block that we start from. Only blocks before that one will be checked.</param>
+        /// <param name="proofOfStake">Specifies what kind of block we are looking for: PoS or PoW.</param>
+        /// <returns>Last block in the chain that satisfies provided requirements.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if index is null.</exception>
+        public static ChainedBlock GetLastBlockIndex(StakeChain stakeChain, ChainedBlock index, bool proofOfStake)
+        {
+            if (index == null)
```
Aprogiena: this should be guard

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -74,58 +216,102 @@ public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPre
...
+            }
 
-            this.logger.LogTrace("(-):{0}", res);
-            return res;
+            return stakeModifier;
         }
 
-        private bool VerifySignature(UnspentOutputs txFrom, Transaction txTo, int txToInN, ScriptVerify flagScriptVerify)
+        /// <summary>
+        /// Checks that staking UTXO is valid for staking and calls <see cref="CheckStakeKernelHash" />.
```
Aprogiena: can we replace `calls X` with something more English?

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -74,58 +216,102 @@ public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPre
...
+        /// </summary>
+        /// <param name="context">The context.</param>
```
Aprogiena: but which context?

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +329,88 @@ private static uint256 ToUInt256(BigInteger input)
...
-        //   stakeModifierV2: Scrambles computation to make it very difficult to precompute future proof-of-stake.
-        //   stakingCoins.Time: Time of the coinstake UTXO. Slightly scrambles computation.
-        //   prevout.Hash: Hash of stakingCoins UTXO, to reduce the chance of nodes generating coinstake at the same time.
-        //   prevout.N: Output number of stakingCoins UTXO, to reduce the chance of nodes generating coinstake at the same time.
-        //   transactionTime: Timestamp of the coinstake transaction.
-        //   Block or transaction tx hash should not be used here as they can be generated in vast
-        //   quantities so as to generate blocks faster, degrading the system back into a proof-of-work situation.
-        //
+        /// <summary>
+        /// Returns <c>true</c> if provided UTXOs were confirmed in less then <see cref="maxDepth" /> number of blocks.
```
Aprogiena: there is no maxDepth here, targetDepth?

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +329,88 @@ private static uint256 ToUInt256(BigInteger input)
...
+        /// </summary>
+        /// <param name="utxoSet">The utxo set.</param>
```
Aprogiena: UTXO in upper case in comments

also i think this parameter is wrongly named, this is not utxoSet we have here,
i mean it is little difficult because UnspentOutputs is nontrivial to understand

it represents some (selected) outputs inside a single transaction
but when we say "UTXO set" we are usually talking about a coin view - i.e. global set of unspent outputs

so i'd suggest we rename this parameter - elsewhere we use `coins` for such a thing, maybe that is good enough here as well

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +329,88 @@ private static uint256 ToUInt256(BigInteger input)
...
+        /// <param name="indexFrom">Chained block from which we are counting the depth.</param>
+        /// <param name="targetDepth">The target depth.</param>
+        /// <returns><c>true</c> if [is confirmed in n previous blocks] [the specified utxo set]; otherwise, <c>false</c>.</returns>
```
Aprogiena: why those `[` and `]` ? make normal English sentence
`<c>true</c> if the coins were spent within N blocks from <see cref="indexFrom"/>, <c>false</c> otherwise.`

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +329,88 @@ private static uint256 ToUInt256(BigInteger input)
...
+        private bool IsConfirmedInNPrevBlocks(UnspentOutputs utxoSet, ChainedBlock indexFrom, long targetDepth)
```
Aprogiena: note that this pindexFrom and any such parameter/variable named *index* of type ChainedBlock is most likely poorly named. this naming comes from bitcoin core, but they do not have ChainedBlock, they have CBlockIndex, therefore it makes sense for them to call things things indexes, for us, it makes less sense

it would make sense to me to name it something like `referenceChainedBlock` or `fromChainedBlock` something like that

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -143,26 +329,88 @@ private static uint256 ToUInt256(BigInteger input)
...
+            bool res = actualDepth < targetDepth;
+
+            this.logger.LogTrace("(-):{0}", res);
+            return res;
+        }
+
+        /// <summary>
+        /// Verifies transaction's signature.
+        /// </summary>
+        /// <param name="txFrom">UTXO that is spent in the transaction.</param>
```
Aprogiena: the comment is correct, but the name of the parameter seems incorrect, it is not transaction

Aprogiena:
conflict please

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,28 +48,36 @@ namespace Stratis.Bitcoin.Features.Miner
...
-            public uint256 HashBlock;
-            public TxOut TxOut;
-            public OutPoint OutPoint;
-            public int OutputIndex;
-            public HdAddress Address;
-            public UnspentOutputs UtxoSet;
-            public WalletSecret Secret;
-            public Key Key;
+            /// <summary>Block's hash.</summary>
+            public uint256 HashBlock { get; set; }
```
Aprogiena: i thought i've mentioned spaces between properties already here, but it was probably somewhere else

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,28 +48,36 @@ namespace Stratis.Bitcoin.Features.Miner
...
+            /// <summary>UTXO.</summary>
```
Aprogiena: can we be little more descriptive here?
i'm little bit confused here with this TxOut vs what is below as UtxoSet
as i described elsewhere (was in the other PR i think), i'm somewhat convinced that thing should not be called UtxoSet, 
i'm not sure if i'm sure about that :-/

what is the difference between TxOut and UnspentOutputs?
TxOut seems to to refer to a single UTXO inside UnspentOutputs, which represents selected outputs of a transactions

the question is whether we can somehow provide a comment here that would make it easier for the reader to understand that logic here

although i know this is difficult, those structures are not trivial to understand, what they actually represent

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,28 +48,36 @@ namespace Stratis.Bitcoin.Features.Miner
...
+            public TxOut TxOut { get; set; }
+            /// <summary>Information about transaction id and index.</summary>
+            public OutPoint OutPoint { get; set; }
+            /// <summary>Transaction's address.</summary>
+            public HdAddress Address { get; set; }
+            /// <summary>Utxo set.</summary>
+            public UnspentOutputs UtxoSet { get; set; }
+            /// <summary><see cref="WalletSecret"/></summary>
```
Aprogiena: this is not good summary

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,28 +48,36 @@ namespace Stratis.Bitcoin.Features.Miner
...
+            public WalletSecret Secret { get; set; }
+            /// <summary>Private key.</summary>
```
Aprogiena: but which private key?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -50,28 +48,36 @@ namespace Stratis.Bitcoin.Features.Miner
...
-            public UnspentOutputs UtxoSet;
-            public WalletSecret Secret;
-            public Key Key;
+            /// <summary>Block's hash.</summary>
+            public uint256 HashBlock { get; set; }
+            /// <summary>UTXO.</summary>
+            public TxOut TxOut { get; set; }
+            /// <summary>Information about transaction id and index.</summary>
+            public OutPoint OutPoint { get; set; }
+            /// <summary>Transaction's address.</summary>
```
Aprogiena: but which transaction?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: how about "Credentials to wallet that contains the private key for the staking UTXO."

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -164,27 +171,43 @@ public class CoinstakeContext
         /// but high enough to compensate for tasks' overhead.</remarks>
         private const int UtxoStakeDescriptionsPerCoinstakeWorker = 25;
 
+        /// <summary>Responsible for downloading and validating blocks.</summary>
```
Aprogiena: puller is responsible for downloading blocks

consensus loop is a component that consumes incoming blocks from puller or from miner/staker and then validates and executes (processes) such a block. processing means that it modifies the UTXO set according to the transactions in the block (among other things)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -164,27 +171,43 @@ public class CoinstakeContext
...
         private readonly ConsensusLoop consensusLoop;
+
+        /// <summary>Chain of headers from genesis.</summary>
```
Aprogiena: see the comment in the other PR

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -164,27 +171,43 @@ public class CoinstakeContext
...
         private readonly ConcurrentChain chain;
 
         /// <summary>Specification of the network the node runs on - regtest/testnet/mainnet.</summary>
         private readonly Network network;
+
+        /// <summary>Used to verify that node is connected to network before we start staking.</summary>
```
Aprogiena: implementation -> this could go to remarks
in summary, i'd prefer something like `Provider of information about the node's connection to its network peers.`

Aprogiena: as mentioned on slack, all those "Used to" and alike, i'm not a big fan of

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -193,18 +216,28 @@ public class CoinstakeContext
...
         private readonly ILogger logger;
 
+        /// <summary>Mining loop.</summary>
         private IAsyncLoop mining;
+
+        /// <summary>Balance that shouldn't participate in taking.</summary>
         private Money reserveBalance;
+
+        /// <summary>UTXO's value threshold for being selected for staking.</summary>
         private readonly int minimumInputValue;
```
Aprogiena: seems unused, always 0 ... should we remove it? @dangershony ?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -969,18 +993,28 @@ private List<UtxoStakeDescription> GetUtxoStakeDescriptionsSuitableForStaking(Li
             return res;
         }
 
-        private int GetBlocksToMaturity(UtxoStakeDescription utxoStakeDescription)
+        /// <summary>
+        /// Calculates blocks count till UTXO is considered mature for staking.
+        /// </summary>
+        /// <param name="utxoStakeDescription">The utxo stake description.</param>
```
Aprogiena: UTXO in upper case in comments

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -969,18 +993,28 @@ private List<UtxoStakeDescription> GetUtxoStakeDescriptionsSuitableForStaking(Li
...
         }
 
-        // Return depth of transaction in blockchain:
-        // -1  : not in blockchain, and not in memory pool (conflicted transaction),
-        //  0  : in memory pool, waiting to be included in a block,
-        // >=1 : this many blocks deep in the main chain.
+        /// <summary>
+        /// Gets depth of transaction in blockchain.
+        /// </summary>
+        /// <param name="utxoStakeDescription">The utxo stake description.</param>
```
Aprogiena: upper case

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -969,18 +993,28 @@ private List<UtxoStakeDescription> GetUtxoStakeDescriptionsSuitableForStaking(Li
...
+        /// <returns>
+        /// <c>-1</c> if not in blockchain, and not in memory pool (conflicted transaction).
+        /// <c>0</c> if in memory pool, waiting to be included in a block.
+        /// <c>>=1</c> if included in a block. Shows how many blocks deep in the main chain.
+        /// </returns>
```
Aprogiena: remember that `<` inside <code> that we replaced with &lt; ?
this `<=` should cause the same thing in your IDE

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -12,14 +12,9 @@
 namespace Stratis.Bitcoin.Features.Consensus
```
Aprogiena: you still have `using NLog.Extensions.Logging;` there which is no longer needed

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: For some reason if I use '&ge;' VS marks it as a mistake- undefined entity 'ge', but it's fine with >=. 
Though this is not fired with &lt; and &gt;
So are you ok with leaving it as it is?

Aprogiena: Actually we could use `Value greater than <c>0<c/> if included in a block.`
using `≥` character is fine for me, my VS does not mark it as invalid
but if you want, you can use `&gt;=`

Aprogiena: but TBH my VS does not complain about > or < or anything else, so you probably have something extra?

noescape00: Yeah, it's probably a warning from resharper or power tools.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -45,11 +45,16 @@ public class BlockValidationContext
...
-    /// and validating this blocks using either the <see cref="PowConsensusValidator"/> for POF networks or <see cref="PosConsensusValidator"/> for POS networks.
+    /// Consumes incoming blocks and then validates and executes such a block.
     /// </summary>
     /// <remarks>
+    /// <para>
+    /// Blocks are coming from <see cref="ILookaheadBlockPuller"/> or Miner/Staker and get validated by
+    /// either the <see cref="PowConsensusValidator"/> for PoW or the <see cref="PosConsensusValidator"/> for PoS.
+    /// </para>
+    /// <para>
     /// An internal loop will manage such background operations.
```
Aprogiena: remove this paragraph

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -45,11 +45,16 @@ public class BlockValidationContext
     }
 
     /// <summary>
-    /// A class that is responsible for downloading blocks from peers using the <see cref="ILookaheadBlockPuller"/>
-    /// and validating this blocks using either the <see cref="PowConsensusValidator"/> for POF networks or <see cref="PosConsensusValidator"/> for POS networks.
+    /// Consumes incoming blocks and then validates and executes such a block.
```
Aprogiena: Consumes incoming blocks, validates and executes them.

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -62,7 +63,7 @@ public class PosConsensusValidator : PowConsensusValidator
 
         private readonly PosConsensusOptions consensusOptions;
 
-        public PosConsensusValidator(StakeValidator stakeValidator, ICheckpoints checkpoints, Network network, StakeChain stakeChain, ConcurrentChain chain, CoinView coinView, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
+        public PosConsensusValidator(StakeValidator stakeValidator,ICheckpoints checkpoints,Network network,StakeChain stakeChain,ConcurrentChain chain,CoinView coinView,IDateTimeProvider dateTimeProvider,ILoggerFactory loggerFactory)
```
Aprogiena: erm? no! put them back

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -5,12 +5,15 @@
...
 using NBitcoin.Crypto;
-using NLog.Extensions.Logging;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
 using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.Consensus
 {
+    /// <summary>
+    /// Provides functionality for checking validity of PoS blocks.
+    /// See <see cref="PosMinting"/> for more information about PoS solutions.
```
Aprogiena: hmm, damn, this reference is not recognized by MSVS because consensus does not depend on miner, not sure what to do with it, but probably nothing :/ 
maybe we could at least put a fully qualified name here `Stratis.Bitcoin.Features.Miner.PosMinting`, although it does not really help, but ...

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
             this.stakeChain = stakeChain;
             this.chain = chain;
             this.coinView = coinView;
             this.consensusOptions = network.Consensus.Option<PosConsensusOptions>();
         }
 
-        public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPrev, BlockStake prevBlockStake, Transaction tx, uint headerBits)
+        /// <summary>
+        /// Gets the last block in the chain that was generated using
+        /// PoS if <see cref="proofOfStake" /> is <c>true</c> or PoW if <see cref="proofOfStake" /> is <c>false</c>.
```
Aprogiena: paramref instead of see

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
+        /// </summary>
+        /// <param name="stakeChain">Chain that is used for retrieving blocks.</param>
```
Aprogiena: this is not correct
StakeChain provides stake information related to a block (that information is called BlockStake), not blocks themselves. 

so, not a chain and not for retrieving blocks - look at the comment of this parameter in GetNextTargetRequired, copy it from there

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
+        /// <param name="index">Block that we start from. Only blocks before that one will be checked.</param>
```
Aprogiena: as mentioned elsewhere, this "Index" naming should probably be changed,
maybe we can do it as a part of this PR as well, the whole method name should be changed
this could be something like `startChainedBlock` and the method name could be GetLastPowPosChainedBlock

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
+        /// <param name="proofOfStake">Specifies what kind of block we are looking for: PoS or PoW.</param>
+        /// <returns>Last block in the chain that satisfies provided requirements.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if index is null.</exception>
```
Aprogiena: `<c>null</c>`

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
         }
 
-        public void CheckProofOfStake(ContextInformation context, ChainedBlock pindexPrev, BlockStake prevBlockStake, Transaction tx, uint headerBits)
+        /// <summary>
+        /// Gets the last block in the chain that was generated using
+        /// PoS if <see cref="proofOfStake" /> is <c>true</c> or PoW if <see cref="proofOfStake" /> is <c>false</c>.
+        /// </summary>
+        /// <param name="stakeChain">Chain that is used for retrieving blocks.</param>
+        /// <param name="index">Block that we start from. Only blocks before that one will be checked.</param>
+        /// <param name="proofOfStake">Specifies what kind of block we are looking for: PoS or PoW.</param>
```
Aprogiena: i'd be even more explicit:
`...looking for: <c>true</c> for PoS or <c>false</c> for PoW.`

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -22,213 +25,210 @@ public class StakeValidator
...
-        private void CheckStakeKernelHash(ContextInformation context, ChainedBlock prevChainedBlock, uint headerBits, uint prevBlockTime,
-            BlockStake prevBlockStake, UnspentOutputs stakingCoins, OutPoint prevout, uint transactionTime)
+        /// <summary>
+        /// Checks if provided transaction is a valid coinstake transaction.
+        /// </summary>
+        /// <param name="context">Staking context.</param>
+        /// <param name="prevChainedBlock">Previous chained block.</param>
+        /// <param name="prevBlockStake">Information about previous staked block.</param>
+        /// <param name="transaction">The transaction.</param>
+        /// <param name="headerBits">Chained block's header bits.</param>
```
Aprogiena: Chained block's header bits, which define the difficulty target.

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -266,139 +275,187 @@ public void CheckKernel(ContextInformation context, ChainedBlock pindexPrev, uin
...
+        /// </summary>
+        /// <param name="input"><see cref="uint256"/> input value.</param>
+        /// <returns><see cref="BigInteger"/> version of <paramref name="input"/>.</returns>
+        private BigInteger FromUInt256(uint256 input)
+        {
+            return BigInteger.Zero;
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> if provided coins were confirmed in less then <paramref name="targetDepth"/> number of blocks.
```
Aprogiena: less then -> less than

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -266,139 +275,187 @@ public void CheckKernel(ContextInformation context, ChainedBlock pindexPrev, uin
...
+        /// </summary>
+        /// <param name="coins">Set of outputs inside a single transaction.</param>
```
Aprogiena: although this is true, i'd go with `Coins to check confirmation depth for.` because it seems more useful for the caller

src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -266,139 +275,187 @@ public void CheckKernel(ContextInformation context, ChainedBlock pindexPrev, uin
...
+        /// <item><paramref name="transactionTime"/>: Timestamp of the coinstake transaction.</item>
+        /// </list>
+        /// Block or transaction tx hash should not be used here as they can be generated in vast
+        /// quantities so as to generate blocks faster, degrading the system back into a proof-of-work situation.
         /// </para>
         /// </remarks>
-        public static Target GetNextTargetRequired(StakeChain stakeChain, ChainedBlock chainedBlock, NBitcoin.Consensus consensus, bool proofOfStake)
+        /// <exception cref="ConsensusErrors.StakeTimeViolation">Thrown in case transaction time is lower than it's own UTXO timestamp.</exception>
+        /// <exception cref="ConsensusErrors.StakeHashInvalidTarget">Thrown in case PoS hash doesn't meet target protocol.</exception>
+        private void CheckStakeKernelHash(ContextStakeInformation context, ChainedBlock prevChainedBlock, uint headerBits, uint prevBlockTime,
```
Aprogiena: this method is missing `<param>` tags for its parameters

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -177,39 +187,47 @@ public class CoinstakeContext
...
-        private readonly ChainState chainState;
-
-        private readonly Signals.Signals signals;
-
         /// <summary>Global application life cycle control - triggers when application shuts down.</summary>
         private readonly INodeLifetime nodeLifetime;
 
-        private readonly NodeSettings settings;
-
+        /// <summary>Used to fetch UTXOs.</summary>
```
Aprogiena: here we will use the same comment as we have in stakevalidator

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -177,39 +187,47 @@ public class CoinstakeContext
...
         private readonly CoinView coinView;
 
         /// <summary>Database of stake related data for the current blockchain.</summary>
         private readonly StakeChain stakeChain;
 
+        /// <summary>Provides functionality for checking validity of PoS blocks.</summary>
+        private readonly StakeValidator stakeValidator;
+
+        /// <summary>Used for creating <see cref="mining"/> loop.</summary>
```
Aprogiena: copy this from BaseFeature

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -218,6 +236,7 @@ public class CoinstakeContext
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
+        /// <summary>Mining loop.</summary>
         private IAsyncLoop mining;
```
Aprogiena: it is actually staking loop
let's rename it to `stakingLoop` and modify comment accordingly -> `Loop in which the node attempts to generate new POS blocks by staking coins from its wallet.`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -226,12 +245,16 @@ public class CoinstakeContext
         /// </summary>
         private Money targetReserveBalance;
 
+        /// <summary>UTXO's value threshold for being selected for staking.</summary>
         private readonly int minimumInputValue;
```
Aprogiena: remove this field from code base entirely

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -226,12 +245,16 @@ public class CoinstakeContext
...
 
+        /// <summary>Time in milliseconds between attempts to generate PoS blocks.</summary>
         private readonly int minerSleep;
 
+        /// <summary>Used for managing asynchronous access to <see cref="mempool"/>.</summary>
```
Aprogiena: we did mention this already somewhere, didn't we?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -226,12 +245,16 @@ public class CoinstakeContext
...
         protected readonly MempoolSchedulerLock mempoolLock;
 
+        /// <summary>Used for populating <see cref="rpcGetStakingInfoModel"/> with pooled transactions.</summary>
```
Aprogiena: this as well

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -262,20 +285,37 @@ public class CoinstakeContext
...
+        /// <param name="chainState">Used to verify that node is not in a state of IBD (Initial Block Download).</param>
+        /// <param name="nodeLifetime">Global application life cycle control - triggers when application shuts down.</param>
+        /// <param name="coinView">Used to fetch UTXOs.</param>
+        /// <param name="stakeChain">Database of stake related data for the current blockchain.</param>
+        /// <param name="stakeValidator">Provides functionality for checking validity of PoS blocks.</param>
+        /// <param name="mempoolLock">Used for managing asynchronous access to <see cref="mempool"/>.</param>
+        /// <param name="mempool">Used for populating <see cref="rpcGetStakingInfoModel"/> with pooled transactions.</param>
+        /// <param name="wallet">A manager providing operations on wallets.</param>
+        /// <param name="asyncLoopFactory">Used for creating <see cref="mining"/> loop.</param>
+        /// <param name="loggerFactory">Factory for creating loggers.</param>
```
Aprogiena: if you change anything above, this constructor, don't forget to change the comments here as well

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -834,12 +870,11 @@ private void CoinstakeWorker(CoinstakeWorkerContext context, ChainedBlock chainT
...
                     try
                     {
-                        var prevoutStake = new OutPoint(utxoStakeInfo.UtxoSet.TransactionId, utxoStakeInfo.OutputIndex);
-                        long nBlockTime = 0;
+                        var prevoutStake = new OutPoint(utxoStakeInfo.UtxoSet.TransactionId, utxoStakeInfo.OutPoint.N);
 
                         var contextInformation = new ContextInformation(new BlockValidationContext { Block = block }, this.network.Consensus);
                         contextInformation.SetStake();
-                        this.posConsensusValidator.StakeValidator.CheckKernel(contextInformation, chainTip, block.Header.Bits, txTime, prevoutStake, ref nBlockTime);
+                        this.posConsensusValidator.StakeValidator.CheckKernel(contextInformation.Stake, chainTip, block.Header.Bits, txTime, prevoutStake, out _);
```
Aprogiena: pls don't use `_`, use `unused` instead or similar
but here, probably just remove the out parameter from the codebase entirely, it is not used anywhere

Issue #750, opened 2017-11-18
[ConnectionManager] SocketException while attempting to start node if stratis-qt wallet is launched


noescape00:
Throws at ConnectionManager:167 "SocketException: Only one usage of each socket address (protocol/network address/port) is normally permitted"

Bug appears if 16178 port is already taken by the stratis-qt wallet that was launched before the full node.

dangershony:
Is this a bug though?
If the port is taken we cannot start the node on the same port (you can specify a different port in config).

noescape00:
It's not. But I'm pretty sure there will be a lot of people who would run node (when it get released) in parallel with qt wallet and that's not UX they should get. 
So maybe the port should be changed automatically instead of shutting down the node in case default one is taken.

dangershony:
I would be reluctant to just randomly select a port, rather we should give a proper error message maybe something like
 >P2P port is used (is a StratisX QT running?) please use a different port using the -port config argument

noescape00:
Sounds like a good idea. I'll make a PR.

PR #787, opened 2017-11-23
Added handling for situation when default port is being used.

noescape00:
Fix: #750

### DIFF
--- a/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
+++ b/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
@@ -171,7 +171,7 @@ namespace Stratis.Bitcoin.Connection
                 }
                 catch (SocketException e)
                 {
-                    this.logger.LogCritical($"P2P port {listen.Endpoint.Port} is being used. Please use a different port by running the node with '-port=[differentPort]'.");
+                    this.logger.LogCritical("P2P port {0} is being used. Please use a different port by running the node with '-port=[differentPort]'.", listen.Endpoint.Port);
                     throw e;
                 }
 
--- a/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
+++ b/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
@@ -171,7 +171,7 @@ namespace Stratis.Bitcoin.Connection
                 }
                 catch (SocketException e)
                 {
-                    this.logger.LogCritical("P2P port {0} is being used. Please use a different port by running the node with '-port=[differentPort]'.", listen.Endpoint.Port);
+                    this.logger.LogCritical("Unable to listen on port {0} (you can change the port using '-port=[number]'). Error message: {1}", listen.Endpoint.Port, e.Message);
                     throw e;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -164,7 +165,15 @@ public void Start()
...
 
                 server.InboundNodeConnectionParameters = cloneParameters;
-                server.Listen();
+                try
+                {
+                    server.Listen();
+                }
+                catch (SocketException e)
+                {
+                    logger.LogCritical("P2P port is used (is a StratisX QT running?) please use a different port using the '-port config' argument for runing the node.");
```
bokobza: It can also be a bitcoin node that's running and causing trouble. Does the exception give us any interesting data we can include in the log?
Also, `runing` should be `running` and `please...` should be preceded by a full stop and start with an uppercase. 😸

noescape00: >Does the exception give us any interesting data we can include in the log?
Not really.
So I guess optimal solution would be to log the port that can't be listened. 

Sorry about grammar, silly mistakes, should have checked.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -164,7 +165,16 @@ public void Start()
...
 
                 server.InboundNodeConnectionParameters = cloneParameters;
-                server.Listen();
+                try
+                {
+                    server.Listen();
+                }
+                catch (SocketException e)
+                {
+                    logger.LogCritical("P2P port " + listen.Endpoint.Port + " is being used. Please use a different port by running the node with '-port=[differentPort]'.");
```
dangershony: Please use string interpolation here $"P2P port {listen.Endpoint.Port} is bla..." 

also `this.logger`

no need to call ` logs.Append("Can't start listening on port: " + listen.Endpoint.Port);` you are throwing

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -164,7 +165,15 @@ public void Start()
...
 
                 server.InboundNodeConnectionParameters = cloneParameters;
-                server.Listen();
+                try
+                {
+                    server.Listen();
+                }
+                catch (SocketException e)
+                {
+                    this.logger.LogCritical($"P2P port {listen.Endpoint.Port} is being used. Please use a different port by running the node with '-port=[differentPort]'.");
```
Aprogiena: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md#interpolated-strings

not critical here as it is one-time exceptional state, but please do for the sake of consistency of the logs

noescape00: Fixed.

Aprogiena: tx

dangershony: Yes correct I tend to use string interpolation when its not a method that will get hit a lot, but for consistency indeed use string format.

PR #789, opened 2017-11-23
Fix: ConnectionManager gets disposed 3 times.

### LINKED ISSUE #691: [BaseFeature-FullNode] ConnectionManager is being disposed 3 times
Once from FullNode.Stop, 
```
```

once from BaseFeature:

```
            this.disposableResources.Add(this.connectionManager);
```

and finally from FullNode.Stop again, but little hidden as 
```
            (this.Services.ServiceProvider as IDisposable)?.Dispose();
```

there is no check if the thing was disposed, but there really does not need to be one, we should just not be disposing things 3 times

noescape00:
Tiny PR that prevents 2 unnecessary ConnectionManager.Dispose() calls. 

I've also checked all the classes that inherit IDisposable if they dispose everything correctly. Only FullNode has a problem- it also gets disposed 3 times (2 of them are not intentional, caused by service provider automatically registering disposable objects), but the workaround already included in FullNode.Dispose() implementation, so it's fine. 

Fix #691

### DIFF
--- a/src/Stratis.Bitcoin/Base/BaseFeature.cs
+++ b/src/Stratis.Bitcoin/Base/BaseFeature.cs
@@ -175,7 +175,6 @@ namespace Stratis.Bitcoin.Base
 
             this.disposableResources.Add(this.timeSyncBehaviorState);
             this.disposableResources.Add(this.chainRepository);
-            this.disposableResources.Add(this.connectionManager);
             this.disposableResources.Add(this.nodeSettings.LoggerFactory);
 
             this.logger.LogTrace("(-)");
--- a/src/Stratis.Bitcoin/FullNode.cs
+++ b/src/Stratis.Bitcoin/FullNode.cs
@@ -215,8 +215,6 @@ namespace Stratis.Bitcoin
             // Fire INodeLifetime.Stopping.
             this.nodeLifetime.StopApplication();
 
-            this.ConnectionManager.Dispose();
-
             foreach (IDisposable dispo in this.Resources)
                 dispo.Dispose();
 


dangershony:
When does the DI dispose of the connection manager?
Multi dispose calls is not such a big problem as long as the pattern makes sure its immutable (i.e. if disposed will not dispose again)

Aprogiena:
@dangershony i am not sure i agree. on one hand, sure, it can be implemented in a way that there is no actual problem in calling the dispose many times, but it feels very wrong principally - it seems that if we dispose things many times, we don't really control those components and they are being randomly disposed somewhere, maybe one time, maybe many times, maybe 0 times ...

so conceptually i think it is a problem despite the fact that it causes no actual damage

Aprogiena:
@noescape00 as this PR does not resolve the 3x FN dispose, do we have that information that FN is disposed 3 times somewhere? like in a separated issue

noescape00:
>@noescape00 as this PR does not resolve the 3x FN dispose, do we have that information that FN is disposed 3 times somewhere? like in a separated issue

No, I only wrote it here and in slack. 
I didn't create a separate issue because it looks like writer of FN.Dispose() knew that and specifically made a workaround. 
Getting rid of that workaround would force us to include and change service provider's source code which is probably an overkill.

Aprogiena:
can you pls explain little bit more, i'm not sure i understand 
what kind of workaround?
pls show more details about why do you think it is hard to dispose FN just once

Aprogiena:
please continue in #799

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -175,7 +175,6 @@ public override void Start()
 
             this.disposableResources.Add(this.timeSyncBehaviorState);
             this.disposableResources.Add(this.chainRepository);
-            this.disposableResources.Add(this.connectionManager);
             this.disposableResources.Add(this.nodeSettings.LoggerFactory);
```
Aprogiena: So after this change, what happens is that `BaseFeature.Stop` is called first and then `ConnectionManager.Dispose` is called. I am not sure that is correct because as you can see here, we dispose logger factory here. The problem with disposing logger factory is that once it is disposed, you must not call any constructor of any object that creates logger, because that will crash. So maybe we don't have any such thing, but if we don't, it is rather coincidence. 

The correct way, imho, is to dispose the logger factory as a last thing. So there should not be any major component running (such as connection manager) after it is disposed. 

I'm not saying this is the case currently, but imagine some component still runs after LF is disposed and a new message arrives from network or a new peer is connected, some behavior is created and the logger is needed ... 

So what we probably need here is to make sure the whole shutdown series is correct, which especially means that the components are stopped and disposed in correct order.

@dangershony can you see the problem here i've tried to described?

dangershony: yes I would even dispose of this tow components manually and not inside the list, they will always get disposed last, LF must be dispose after CM, in fact perhaps we dont even dispose of LF at all, as this will anyway get cleared away for shutdown.

@noescape00 regarding your comment yeah I agree we should control the processes of shutdown however the IDisposable pattern requires the method to be able to be called more then once without throwing.

But if the DI is calling Dispose then this is out of our control and perhaps we need to make changes (not use IDsiposable as this is a component that is completely under our control) and just clean it up at node shutdown.

dangershony:
I would remove the IDisposable from ConnectionManager and call it explicitly on node shutdown.
I would also not add the `LoggerFactory` to the disposable list and explicitly dispose of it.

Aprogiena:
We have advanced much further in this conversation in #799 

I would thus rather suggest to abandon this PR (not merge it) and deal with what we discuss in #799, then we can resolve #691 as a part of that big thing.

Aprogiena:
everyone happy to not-merge this PR?

noescape00:
Yep, I'm fine with it.

PR #790, opened 2017-11-23
Added some extra data to a successfully sent transaction. (#781)

bokobza:
(cherry picked from commit b237616409e9784f1bee5058176913a877d777b3)


bokobza:
@dangershony this is the commit that was lost.

PR #791, opened 2017-11-24
 PoS/Pow consensus validator comments and refactoring.

noescape00:


src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -23,12 +27,15 @@ public class PowConsensusValidator
...
         public NBitcoin.Consensus ConsensusParams { get { return this.consensusParams; } }
 
+        /// <summary>Consensus options.</summary>
         private readonly PowConsensusOptions consensusOptions;
         public PowConsensusOptions ConsensusOptions { get { return this.consensusOptions; } }
 
+        /// <summary>Keeps track of how much time different actions took to execute and how many times they were executed.</summary>
         public ConsensusPerformanceCounter PerformanceCounter { get; }
 
         public bool UseConsensusLib { get; set; }
```
Aprogiena: missing xmldoc (just go through whole file and try to find out missing xmldocs)

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -107,9 +125,8 @@ public virtual void ContextualCheckBlock(ContextInformation context)
                 int commitpos = this.GetWitnessCommitmentIndex(block);
                 if (commitpos != -1)
                 {
-                    bool malleated = false;
-                    uint256 hashWitness = this.BlockWitnessMerkleRoot(block, ref malleated);
-             
+                    uint256 hashWitness = this.BlockWitnessMerkleRoot(block, out _);
```
Aprogiena: `out unused` would feel better

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -57,6 +69,12 @@ public virtual void CheckBlockHeader(ContextInformation context)
             context.NextWorkRequired = context.BlockValidationContext.ChainedBlock.GetWorkRequired(context.Consensus);
         }
 
+        /// <summary>
+        /// Context-dependent validity checks.
+        /// </summary>
+        /// <param name="context">The context.</param>
```
Aprogiena: i was actually waiting for this summary and param to see what is this context, how are those checks different - there is 
and i'm disappointed to see that i still don't know what this context is 

can we find out? how are those check methods different, why is this context stuff separated, what is this context? do you know? i don't know.

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -57,6 +69,12 @@ public virtual void CheckBlockHeader(ContextInformation context)
...
+        /// <exception cref="ConsensusErrors.BadTransactionNonFinal">Thrown if one or more transactions are not finalized.</exception>
+        /// <exception cref="ConsensusErrors.BadCoinbaseHeight">Thrown if coinbase doesn't start with serialized block height.</exception>
         public virtual void ContextualCheckBlock(ContextInformation context)
         {
             this.logger.LogTrace("()");
```
Aprogiena: lots of hungarian naming stuff in here, maybe elsewhere as well ... can we get rid of it?

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -162,6 +179,16 @@ public virtual void ContextualCheckBlock(ContextInformation context)
             this.logger.LogTrace("(-)[OK]");
         }
 
+        /// <summary>
+        /// Validates the UTXO set is correctly spent.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="taskScheduler">The task scheduler.</param>
```
Aprogiena: task scheduler is task scheduler, but ?? 
such comments are not very helpful
i mean without going into the code, why would ExecuteBlock needs task scheduler? 

anyway, i'd recommend to refactor this method to get rid of the task scheduler here, because looking at the code, it seems like no one uses it anyway (or did i miss that?)

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -287,6 +313,11 @@ public virtual void ExecuteBlock(ContextInformation context, TaskScheduler taskS
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Updates context's coin view.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="tx">Transaction that is used for fetching UTXO's.</param>
```
Aprogiena: i don't think this is correct
seems like the transaction here is the one which inputs are to be updated
check the code how it is used

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -313,16 +352,22 @@ public virtual void CheckBlockReward(ContextInformation context, Money nFees, Ch
             this.logger.LogTrace("(-)");
         }
 
-        protected virtual void CheckMaturity(UnspentOutputs coins, int nSpendHeight)
+        /// <summary>
+        /// Checks the maturity of UTXOs.
+        /// </summary>
+        /// <param name="coins">UTXOs.</param>
```
Aprogiena: could we maybe `UTXOs to check the maturity of.`

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -313,16 +352,22 @@ public virtual void CheckBlockReward(ContextInformation context, Money nFees, Ch
...
+        /// <param name="spendHeight">Height at which coins are considered spendable.</param>
```
Aprogiena: this seems incorrect, see the code CoinbaseMaturity is something like that you describe
this is something else, this seems to be the height at which the UTXOs are attempted to be spent

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -331,9 +376,20 @@ protected virtual void CheckMaturity(UnspentOutputs coins, int nSpendHeight)
             this.logger.LogTrace("(-)");
         }
 
-        public virtual void CheckInputs(Transaction tx, UnspentOutputSet inputs, int nSpendHeight)
+        /// <summary>
+        /// Checks that transaction's inputs are valid.
+        /// </summary>
+        /// <param name="tx">The transaction.</param>
```
Aprogiena: as above, i like more if the thing is a full sentence. 
`Transaction to check.` would feel better

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -331,9 +376,20 @@ protected virtual void CheckMaturity(UnspentOutputs coins, int nSpendHeight)
...
+        /// <param name="inputs">Inputs for the transaction.</param>
```
Aprogiena: this is imho incorrect
from the code, this seems to be UTXO set 
you have UTXO set that defines who has which coins and if a transaction has inputs that are not in UTXO set, tx is invalid

Aprogiena:
conflicts pls

noescape00:
>conflicts pls
Yep, waiting for the other PR to get merged to continue working on this one.

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
noescape00: This is called from ExecuteBlock for each block's transaction. And what it does is it just adds or replaces transaction in context's UnspentOutputSet. 
I don't see here any code where transaction's inputs get updated.

Aprogiena: I see, misunderstanding here. I did not mean to say that the transaction itself is being updated.
CoinView is being updated. What this method does is that we have previous CV state C1 and then we have transaction X, and the result is CV state C2 as in `f(C1, X) -> C2`

state C2 is exactly the same as C1 except for those outputs that are being spent using X (X's inputs), so those outputs are no longer in C2 and X's outputs are new to C2 and are not in C1

so `C2 := C1 - {X_inputs} + {X_outputs}`

Aprogiena: and again which context?

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
noescape00: Now when I thought about it I actually can't come up with a good description of this context. It contains basically everything for validation and execution of the block. 

So ho about this: 
"Context that contains variety of information regarding blocks validation and execution."

Aprogiena: sure, anything more than just "the context" is better

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -725,7 +914,7 @@ private void MerkleComputation(List<uint256> leaves, ref uint256 root, ref bool
             while (count < leaves.Count)
             {
                 uint256 h = leaves[(int)count];
-                bool matchh = count == branchpos;
+                bool match = false;
```
dangershony: Change of logic has to be reviewed carefully

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -284,426 +599,300 @@ public virtual void ExecuteBlock(ContextInformation context, TaskScheduler taskS
...
         }
 
-        private void MerkleComputation(List<uint256> leaves, ref uint256 root, ref bool pmutated, int branchpos, List<uint256> pbranch)
+        /// <summary>
+        /// Computes merkle root.
+        /// </summary>
+        /// <remarks>This implements a constant-space merkle root/path calculator, limited to 2^32 leaves.</remarks>
+        /// <param name="leaves">Merkle tree leaves.</param>
+        /// <param name="mutated"><c>true</c> if at least one leaf of the merkle tree has the same hash as any subtree. Otherwise: <c>false</c>.</param>
+        private uint256 computeMerkleRoot(List<uint256> leaves, out bool mutated)
```
Aprogiena: methods are always with uppercase first letter regardless of whether they are private/public/protected...

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -284,426 +599,300 @@ public virtual void ExecuteBlock(ContextInformation context, TaskScheduler taskS
...
         {
-            if (pbranch != null)
-                pbranch.Clear();
+            List<uint256> branch = new List<uint256>();
```
Aprogiena: we use var in 
`X x = new X();` -> `var x = new X();`

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
Aprogiena: yes and same below on new line 938

Aprogiena: btw this PR is not ready for review i think, ...

noescape00: Yep, I'll ping when it is completed. There are still a few things I have to check and improve.

noescape00: >Change of logic has to be reviewed carefully

Totally agreed. 

Regarding that change: 
branchpos and pbranch were there in order to achieve a small performance acceleration in some rare cases (so we don't have to rebuild whole tree every time) but we never used it and always filled this params with -1 and null so  
`count == branchpos` statement is always false. 

I've checked if old and new version of ComputeMerkleRoot produces the same result and it does. 
Also if new implementation would calculate merkle root differently consensus would stop advancing, so before pushing anything I'm checking if consensus advances.

Aprogiena: > I've checked if old and new version of ComputeMerkleRoot produces the same result and it does.

i'd recommend to create a set of tests for that that should pass before the change and after the change as well
or are there such tests implemented already?

noescape00: Good idea. 
Though I guess it will be a good idea to write tests for merkle root calculation after this PR so we don't make dangershony wait too long till this one is merged. 

>or are there such tests implemented already?

No, there are no tests involving this merkle root calculation code.

dangershony: > so before pushing anything I'm checking if consensus advances

Great yes thats one way to do it, tests will be also really nice to have (double check NBitcoin does not  have some already)

dangershony: >btw this PR is not ready for review i think, 

Ok @Aprogiena I will leave most of this to you to review unless you need my help let me know.

noescape00:
Ok, I think it's ready for review.

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -467,23 +425,112 @@ public void CheckAndComputeStake(ContextInformation context)
...
 
-        // Miner's coin stake reward.
-        public Money GetProofOfStakeReward(int height)
+        /// <summary>
+        /// Checks whether the coinstake timestamp meets protocol.
+        /// </summary>
+        /// <param name="blockTime">The block time.</param>
+        /// <param name="transactionTime">The transaction time.</param>
+        /// <returns><c>true</c> if block timestamp is equal to transaction timestamp, <c>false</c> otherwise.</returns>
+        private bool checkCoinStakeTimestamp(long blockTime, long transactionTime)
```
Aprogiena: capital C

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -467,23 +425,112 @@ public void CheckAndComputeStake(ContextInformation context)
...
-            if (this.IsPremine(height))
-                return this.consensusOptions.PremineReward;
+            return (blockTime == transactionTime) && ((transactionTime & StakeTimestampMask) == 0);
+        }
 
-            return this.consensusOptions.ProofOfStakeReward;
+        /// <summary>
+        /// Checks whether the future drift should be reduced after provided timestamp.
+        /// </summary>
+        /// <param name="time">Timestamp.</param>
```
Aprogiena: UNIX timestamp.

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -467,23 +425,112 @@ public void CheckAndComputeStake(ContextInformation context)
...
+        private bool IsDriftReduced(long time)
+        {
+            return time > DriftingBugFixTimestamp;
+        }
+
+        /// <summary>
+        /// Applies future drift to provided timestamp.
+        /// </summary>
+        /// <remarks>
+        /// Future drift is maximal allowed block's timestamp difference over local time.
```
Aprogiena: over adjusted time, not local time

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -467,23 +425,112 @@ public void CheckAndComputeStake(ContextInformation context)
...
+        /// If this difference is greater block won't be accepted.
+        /// </remarks>
+        /// <param name="time">Timestamp</param>
```
Aprogiena: UNIX timestamp. (note missing dot)

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
```diff
@@ -467,23 +425,112 @@ public void CheckAndComputeStake(ContextInformation context)
...
+        /// <returns>Timestamp with maximum future drift applied.</returns>
+        private long FutureDrift(long time)
+        {
+            return this.IsDriftReduced(time) ? time + 15 : time + 128 * 60 * 60;
```
Aprogiena: i'd like these constants to be commented constants not hardcoded values

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
Aprogiena: well still missing xmldoc :-)

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -22,17 +26,25 @@ public class PowConsensusValidator
...
 
+        /// <summary>Keeps track of how much time different actions took to execute and how many times they were executed.</summary>
         public ConsensusPerformanceCounter PerformanceCounter { get; }
 
         public bool UseConsensusLib { get; set; }
 
         /// <summary>Provider of time functions.</summary>
         protected readonly IDateTimeProvider dateTimeProvider;
 
+        /// <inheritdoc />
```
Aprogiena: from where?

Aprogiena: i mean no interface, no base class ... and not override

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -46,32 +58,216 @@ public PowConsensusValidator(Network network, ICheckpoints checkpoints, IDateTim
...
+
+            return signatureOperationCost;
         }
 
+        /// <summary>
+        /// Checks that transaction's inputs are valid.
+        /// </summary>
+        /// <param name="transaction">Transaction to check.</param>
+        /// <param name="inputs">Map of previous transactions that have outputs we're spending.</param>
+        /// <param name="spendHeight">Height at which coins are considered spendable.</param>
```
Aprogiena: i think this is incorrect
this seems to be the height at which we are spending
not height at which the coins are spendable
CheckMaturity call will tell us if those are spendable

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -46,32 +58,216 @@ public PowConsensusValidator(Network network, ICheckpoints checkpoints, IDateTim
...
+            this.logger.LogTrace("(-)[OK]");
+        }
+
+        /// <summary>
+        /// Context-dependent validity checks.
+        /// </summary>
+        /// <param name="context">Context that contains variety of information regarding blocks validation and execution.</param>
+        /// <exception cref="ConsensusErrors.BadTransactionNonFinal">Thrown if one or more transactions are not finalized.</exception>
+        /// <exception cref="ConsensusErrors.BadCoinbaseHeight">Thrown if coinbase doesn't start with serialized block height.</exception>
         public virtual void ContextualCheckBlock(ContextInformation context)
```
Aprogiena: empty line

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -284,23 +316,36 @@ public virtual void ExecuteBlock(ContextInformation context, TaskScheduler taskS
             this.logger.LogTrace("(-)");
         }
 
-        protected virtual void UpdateCoinView(ContextInformation context, Transaction tx)
+        /// <summary>
+        /// Updates context's <see cref="UnspentOutputSet"/>.
```
Aprogiena: i'd prefer english over reference - `Updates context's UTXO set.`

src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
noescape00: At first I had something like "over local time (usually adjusted)" but changed it to that because by local we could mean that it's adjusted already. 

Just checked IDateTimeProvider so yeah, 'local' is not applicable here.

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
noescape00: Oh. I thought constructors can inherit class summary.

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
noescape00: Hmm. Where?

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
Aprogiena: class summary is more of a big picture, constructors do have their own

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
Aprogiena: erm sorry, ignore me, it is too late :)

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -498,14 +588,15 @@ public virtual void CheckBlock(ContextInformation context)
             // checks that use witness data may be performed here.
 
             // Size limits.
-            if ((block.Transactions.Count == 0) || (block.Transactions.Count > this.ConsensusOptions.MaxBlockBaseSize) || (this.GetSize(block, TransactionOptions.None) > this.ConsensusOptions.MaxBlockBaseSize))
+            if (block.Transactions.Count == 0 || block.Transactions.Count > this.ConsensusOptions.MaxBlockBaseSize ||
```
Aprogiena: please do ADD, not REMOVE those `(` and `)`

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -498,14 +588,15 @@ public virtual void CheckBlock(ContextInformation context)
...
+                this.GetSize(block, TransactionOptions.None) > this.ConsensusOptions.MaxBlockBaseSize)
             {
                 this.logger.LogTrace("(-)[BAD_BLOCK_LEN]");
                 ConsensusErrors.BadBlockLength.Throw();
             }
 
             // First transaction must be coinbase, the rest must not be
-            if ((block.Transactions.Count == 0) || !block.Transactions[0].IsCoinBase)
+            if (block.Transactions.Count == 0 || !block.Transactions[0].IsCoinBase)
```
Aprogiena: please do ADD, not REMOVE those `(` and `)`

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -589,38 +698,38 @@ public virtual void CheckTransaction(Transaction tx)
...
 
-                vInOutPoints.Add(txin.PrevOut);
+                inOutPoints.Add(txin.PrevOut);
             }
 
-            if (tx.IsCoinBase)
+            if (transaction.IsCoinBase)
             {
-                if ((tx.Inputs[0].ScriptSig.Length < 2) || (tx.Inputs[0].ScriptSig.Length > 100))
+                if (transaction.Inputs[0].ScriptSig.Length < 2 || transaction.Inputs[0].ScriptSig.Length > 100)
```
Aprogiena: please do ADD, not REMOVE those `(` and `)`

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -633,53 +742,88 @@ public virtual void CheckTransaction(Transaction tx)
...
+            var leaves = new List<uint256>();
             leaves.Add(uint256.Zero); // The witness hash of the coinbase is 0.
             foreach (Transaction tx in block.Transactions.Skip(1))
                 leaves.Add(tx.GetWitHash());
 
             return this.ComputeMerkleRoot(leaves, ref mutated);
         }
 
+        /// <summary>
+        /// Calculates merkle root for block's trasnactions.
```
Aprogiena: typo

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -691,6 +835,17 @@ private uint256 ComputeMerkleRoot(List<uint256> leaves, ref bool mutated)
             return hash;
         }
 
+        /// <summary>
+        /// Computes merkle root.
+        /// </summary>
+        /// <param name="leaves">Merkle tree leaves.</param>
+        /// <param name="root">Merkle root</param>
```
Aprogiena: missing dot

src/Stratis.Bitcoin.Features.Consensus/PowConsensusValidator.cs
```diff
@@ -818,18 +973,29 @@ private void MerkleComputation(List<uint256> leaves, ref uint256 root, ref bool
...
+                var scriptPubKey = block.Transactions[0].Outputs[i].ScriptPubKey;
+                byte[] scriptBytes = scriptPubKey.ToBytes(true);
+
+                if ((scriptPubKey.Length >= 38) &&
+                    (scriptBytes[0] == (byte)OpcodeType.OP_RETURN) &&
+                    (scriptBytes[1] == 0x24) &&
+                    (scriptBytes[2] == 0xaa) &&
+                    (scriptBytes[3] == 0x21) &&
+                    (scriptBytes[4] == 0xa9) &&
+                    (scriptBytes[5] == 0xed))
```
Aprogiena: the original code is not equivalent with the new one
however, i'm not familiar with this code enough to be able to decide whether it matters or not
the difference is that if length of the sPK is less than 37, you don't call `.ToBytes`, which saves time 

on the other hand your code saves time by only doing `ToBytes` once

so it may be worse or better than original depending on how often the first condition is satisfied
best option here is to probably separate that first condition and only call ToBytes once only then
in which case we will have always better code

PR #792, opened 2017-11-24
Fixed the default Stratis REST API port

andrasfuchs:
Changed the default API port for the Stratis daemon from 37220 to 37221. It still uses the 37220 in Bitcoin mode.


bokobza:
Thanks for redoing this @andrasfuchs !

PR #794, opened 2017-11-24
Refactor/the node job - part 1

Aprogiena:
This moves most of NBitcoin.Protocol to full node code

- lots of files are moved
- large amount of files are changed due to new dependencies
- very minor polishing
- there are NO changes in logic or design
- some NBitcoin tests are commented out, can be ported later


Aprogiena:
ok, no problem to leave your files where they are, they can be moved later if needed

there still is NBitcoin Protocol folder, because many other components of NBitcoin rely on those files
so i tried to extract the minimum from it so that I can move the Node file to FN. so that is expected and moving the rest of the protocol is bigger task again, no need for that at the moment

Aprogiena:
now that the jeremy's monster is merged and conflicts resolved here, it is time for this one ...

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTests.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System;
-using System.IO;
 using NBitcoin;
-using NBitcoin.Protocol;
 using NBitcoin.RPC;
```
bokobza: namespaces are the wrong way around.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -1,14 +1,14 @@
-﻿#if !NOSOCKET
-
-using System;
-using System.Linq;
-using System.Threading;
 using Microsoft.Extensions.Logging;
```
bokobza: please sort namespaces.

bokobza:
> there are changes in logic or design

where?

Aprogiena:
@bokobza sorry, fixed, NO changes :)

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: tx, will fix

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTests.cs
Aprogiena: tx, will fix

PR #798, opened 2017-11-25
Implement IPeerConnector

fassadlr:
Added documentation on IPeerConnector and PeerConnector


src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -77,24 +129,32 @@ public sealed class PeerConnector : IDisposable
             this.MaximumNodeConnections = 8;
```
Aprogiena: i know this is little offtopic at this PR, but i'd prefer a well documented constant instead of hardcoded value

but again, no need for that in this particular PR, just a random comment to consider later

src/Stratis.Bitcoin/P2P/RelatedPeerConnectors.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace Stratis.Bitcoin.P2P
 {
-    public sealed class RelatedPeerConnectors : Dictionary<string, PeerConnector>
+    public sealed class RelatedPeerConnectors : Dictionary<string, IPeerConnector>
```
dangershony: Ah ok this is just to introduce the `IPeerConnector` a future PR will get ride of `RelatedPeerConnectors ` thats fine

src/Stratis.Bitcoin/P2P/PeerConnector.cs
dangershony: Correct but I am keen top push this (as generally this looks good) so @fassadlr future PR

PR #802, opened 2017-11-25
Changed some logs from Info to Debug

bokobza:
This is to get rid of the amount of unnecessary logging in the Travis build while running tests.
Like so:

image


bokobza:
no it's another merge that failed it. I'll try to find out where the error is tomorrow.
Thanks for reviewing this mate :-)

PR #807, opened 2017-11-27
Fix transaction model tests

fassadlr:


bokobza:
Which tests were failing?

fassadlr:
Hi @bokobza 6 TransactionModelTests were failing as well as 2 LoadPoolTests

bokobza:
Looks good.
However, the following tests are still failing, but only on OSX:
```
SendTransactionWithoutMempoolValidatorProcessesTransactionAndBroadcastsTransactionToConnectionManagerNodes
SendTransactionWithMempoolValidatorWithAcceptToMemoryPoolSuccessProcessesTransaction
SendTransactionWithMempoolValidatorWithAcceptToMemoryPoolFailedDoesNotProcessesTransaction
StopwatchDisposable_MeasuresPerformanceCorrectly
```

Do you have any idea of what's happening there?

fassadlr:
@bokobza these tests all have their asserts commented out:

```
SendTransactionWithoutMempoolValidatorProcessesTransactionAndBroadcastsTransactionToConnectionManagerNodes
SendTransactionWithMempoolValidatorWithAcceptToMemoryPoolSuccessProcessesTransaction
SendTransactionWithMempoolValidatorWithAcceptToMemoryPoolFailedDoesNotProcessesTransaction
```

Not sure why the stopwatch test fails :(

bokobza:
really? why does it fail on Mac though and not on the other ones?
I'm not actually sure we need these tests actually.
Can you please comment them out completely and add a TODO there.
Someone will have to take care of them at some point but let's make the build green again.
#fassadlr2020

fassadlr:
@bokobza OK so I have now commented out the those tests that did not have any asserts.

@Aprogiena do you perhaps have ANY insight as to why `StopwatchDisposable_MeasuresPerformanceCorrectly` always fail on Mac?

Any help would be appreciated. If we can get his test to pass on Travis we are golden.

Aprogiena:
@fassadlr interesting, well i'm windows guy, i know linux little bit, i have no idea about Mac at all, but if you could share the reason why it fails - like the error or exception it produces, i can maybe think about it then

fassadlr:
@Aprogiena the test fails here:

```Assert.True(diffDispDiag < epsilonMs);```

line 76.

Not sure if this helps :(

@bokobza we only nbow have one test failing!

bokobza:
like I said #fassadlr2020

Aprogiena:
it does help, are you able to check what the actual values are?
the idea of that test is pretty much described in its comment
the assumption is that all three methods will measure the time equally well
and until the machine testing it is very very slow, we assume that differences in measurements will fit within epsilonMs, which is 50 ms
so it really depends on what diffDispDiag value actually is, if it is something like 60 ms, then our assumptions are correct, just the threshold is wrongly selected and we should increase the threshold to something like 100 ms

however, if there is some fundamental problem, the actual value will be something like 394723, in which case it needs deeper inspection

PR #812, opened 2017-11-28
Remove RelatedPeerConnector and WellKnownPeerConnectorSelectors

fassadlr:


src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -220,30 +201,10 @@ private void Disconnect()
...
-                {
-                    groupFail++;
-                    continue;
-                }
-
-                break;
-            }
-
-            return peer;
+            NetworkAddress peer = this.peerAddressManager.SelectPeerToConnectTo(this.peerIntroductionType);
```
dangershony: hmm where is the code that replaces the GroupConnector?

dangershony: I would suggest to pass ConnectionMaanger as a parent to PeerConenctor (or even only pass the PeerCollection in the constructor, that way you can check a peer is not already connected.

fassadlr:
@dangershony Im closing this PR, I didn't understand the code correctly so will redo this another PR.

PR #813, opened 2017-11-28
Dont run StopwatchDisposable_MeasuresPerformanceCorrectly test in Mac OSX

fassadlr:


src/Stratis.Bitcoin.Tests/Utilities/StopwatchTest.cs
```diff
@@ -31,6 +32,11 @@ public class StopwatchTest
         [Fact]
         public void StopwatchDisposable_MeasuresPerformanceCorrectly()
         {
+            //Don't run this test in a Mac environment as it takes too long,
```
Aprogiena: space after `//` :D

bokobza:
Do we know why this is happening?

fassadlr:
@Aprogiena can you comment?

Aprogiena:
@bokobza nope :( we don't, the mac is so slow that we expect the virtual machine is somehow messed up, or possibly it is run in parallel with something cpu heavy so it does not really move, the measured values are so unreal that for sure there is something wrong.

PR #823, opened 2017-11-29
Refactor/the node job part 3

Aprogiena:
Another episode includes:
- in NetworkPeer and NetworkPeerConnection classes, remove NBitcoin based trace correlation and replace it with logs
- dispose some previously undisposed objects in those classes

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
@@ -99,9 +99,11 @@ namespace Stratis.Bitcoin.P2P.Peer
         private int cleaningUp;
         public int ListenerThreadId;
 
-        public NetworkPeerConnection(NetworkPeer peer, Socket socket, ILoggerFactory loggerFactory)
+        public NetworkPeerConnection(NetworkPeer peer, Socket socket, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
         {
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName, $"[{peer.PeerAddress.Endpoint}] ");
+            this.dateTimeProvider = dateTimeProvider;
+
             this.Peer = peer;
             this.Socket = socket;
             this.Messages = new BlockingCollection<SentMessage>(new ConcurrentQueue<SentMessage>());
@@ -437,7 +439,7 @@ namespace Stratis.Bitcoin.P2P.Peer
         public event NodeEventMessageIncoming MessageReceived;
         public event NodeEventHandler Disconnected;
 
-        public NetworkPeer()
+        public NetworkPeer(IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
         {
             // This constructor is used for testing until the Node class has an interface and can be mocked.
             this.Behaviors = new NetworkPeerBehaviorsCollection(this);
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -236,27 +230,40 @@ public void BeginListen()
...
+            this.logger.LogTrace("(-)");
         }
 
         private void EndListen(Exception unhandledException)
         {
+            this.logger.LogTrace("()");
+
             if (Interlocked.CompareExchange(ref this.cleaningUp, 1, 0) == 1)
+            {
+                this.logger.LogTrace("(-)[CLEANING_UP]");
```
dangershony: Should this be `NOT_CLEANING_UP`

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -568,14 +580,18 @@ private void OnStateChanged(NetworkPeerState previous)
...
                     }
                 }
             }
+
+            this.logger.LogTrace("(-)");
         }
 
         public void OnMessageReceived(IncomingMessage message)
         {
+            this.logger.LogTrace("({0}:'{1}')", nameof(message), message.Message.Command);
```
dangershony: `message.Message`
Message can never be null?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -637,13 +661,17 @@ private void FireFilters(IEnumerator<INetworkPeerFilter> enumerator, Payload pay
                 }
                 catch (Exception ex)
                 {
-                    this.TraceCorrelation.LogInside(() => NodeServerTrace.Error("Unhandled exception raised by a node filter (OnSendingMessage)", ex.InnerException), false);
+                    this.logger.LogError("Exception occurred: {0}", ex.InnerException.ToString());
```
dangershony: The following error logs only have `Exception occurred` when previously there was a more detailed message is that intentional?

dangershony:
Really minor stuff I think this can go in.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: correct, it can't be null

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: consider two such guards
```
if (x)
{
   exit
}

if (y)
{
  exit
}

actually do cleanup
```

with the proposed logic, you would put in both X and Y branches `trace(not cleaning up)`, but the purpose of the context is to distinguish which one actually executed

so rather than using `trace(not cleaning up)` we want something like `trace(exit because of x)` and  `trace(exit because of y)`

so we are existing here because cleanup is already in progress, thus CLEANING_UP

could be ALREADY_IN_PROGRESS or ALREADY_CLEANING or anything else, but NOT_CLEANING_UP would make less sense to me

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: yes, because due to how our NLog is set up, that information is included implicitly

PR #824, opened 2017-11-29
Refactor/the node job part 4

Aprogiena:
Depends on #823

This episode is about removing unused code and moving the code that does not belong to NetworkPeer class out. There were some methods only used by some tests, so they were moved to NodeBuilder.


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -1247,7 +1358,10 @@ public TimeSpan PingPong(CancellationToken cancellation = default(CancellationTo
                 }
 
                 DateTimeOffset after = DateTimeOffset.UtcNow;
-                return after - before;
+
+                TimeSpan res = after - before;
```
bokobza: what does `res` stand for?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -853,40 +924,52 @@ public void Disconnect(string reason = null)
...
+            this.logger.LogTrace("(-)");
         }
 
         private void DisconnectNode(string reason = null, Exception exception = null)
         {
+            this.logger.LogTrace("({0}:'{1}')", nameof(reason), reason);
+
             if (Interlocked.CompareExchange(ref this.disconnecting, 1, 0) == 1)
+            {
+                this.logger.LogTrace("(-)[DISCONNETING");
```
bokobza: DISCONNECTING - missing C

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -734,39 +766,62 @@ public Task SendMessageAsync(Payload payload)
...
+            this.logger.LogTrace("({0}:'{1}')", nameof(payload), payload);
+
             try
             {
                 SendMessageAsync(payload).Wait(cancellation);
             }
             catch (AggregateException aex)
             {
-                ExceptionDispatchInfo.Capture(aex.InnerException).Throw();
+                this.logger.LogTrace("Exception occurred: {0}", aex.InnerException.ToString());
```
bokobza: Maybe use LogError here.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -568,14 +574,18 @@ private void OnStateChanged(NetworkPeerState previous)
                     }
                     catch (TargetInvocationException ex)
                     {
-                        this.TraceCorrelation.LogInside(() => NodeServerTrace.Error("Error while StateChanged event raised", ex.InnerException));
+                        this.logger.LogError("Exception occurred: {0}", ex.InnerException.ToString());
```
bokobza: Here and for all similar occurrences of this log statement:
the InnerException might be null. Maybe log the message first and then if you really want the whole error you can use this method `public static string ExceptionToString(Exception exception)` in Util.cs.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -464,68 +462,69 @@ public NetworkPeer(NetworkAddress peerAddress, Network network, NetworkPeerConne
...
+                        this.logger.LogTrace("Outbound connection to '{0}' established.", peerAddress.Endpoint);
                     }
                 }
-                catch (OperationCanceledException)
-                {
-                    Utils.SafeCloseSocket(socket);
-                    NodeServerTrace.Information("Connection to node cancelled");
-                    this.State = NetworkPeerState.Offline;
+            }
+            catch (OperationCanceledException)
```
bokobza: How about logging what;s in the error object?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -236,27 +228,40 @@ public void BeginListen()
...
-                    {
-                    }
-                    catch (Exception ex)
-                    {
-                        unhandledException = ex;
-                    }
-                    NodeServerTrace.Information("Stop listening");
-                    this.EndListen(unhandledException);
                 }
+                catch (OperationCanceledException)
```
bokobza: How about logging the ex message?

src/Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
```diff
@@ -739,26 +739,45 @@ public void Rollback(Transaction tx)
...
+                peer.SendMessageAsync(new InvPayload(transaction));
+                peer.SendMessageAsync(new TxPayload(transaction));
+                this.PingPong(peer);
             }
         }
 
-#else
-        public void Broadcast(Transaction transaction)
+        /// <summary>
+        /// Emit a ping and wait the pong.
```
bokobza: I know you're just moving this around, but can you make this explanation a bit clearer?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -989,159 +987,6 @@ public override string ToString()
...
         }
 
-
-        /// <summary>
-        /// Get the chain of headers from the peer (thread safe).
-        /// </summary>
-        /// <param name="hashStop">The highest block wanted.</param>
-        /// <param name="cancellationToken"></param>
-        /// <returns>The chain of headers.</returns>
-        public ConcurrentChain GetChain(uint256 hashStop = null, CancellationToken cancellationToken = default(CancellationToken))
```
bokobza: All this has moved to the integration tests, is it not used in the code base but for testing?

Aprogiena: yes, this does not belong to NP class, this is just something a couple of tests need and that's it
after all, the solution compiles, tests pass ...

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: yup, will do

src/Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
bokobza: No it's fine, leave it as is for now.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: sorry i don't understand what you mean
when exception happens and it is not cancellation, it will be that catch Exception case bellow which logs the full exception
i fail to see what is that you call "error object"

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: are you sure it can be null? 
if that was true, the original code would be broken
i tried to be careful and use the same assumptions as the original code
so if it was not broken, my new code should not be broken as well and vice versa

so looking at TargetInvocationException it seems that the inner exception is set (always?)
so then there are just 3 cases, firefilters which uses Exception
and sendmessage that uses AggregateException
that AggregateException in MSDN is also said to (always?) have inner exception set, that's comes from Task.Wait

so FireFilters remain, there i think you are right, that uses Exception and calls unknown code, so we can't guarantee anything about exception coming out, i'll fix FireFilters handler then, tx

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: i would agree if we have separated logging from console output
any LogError goes to the console in our implementation
and i believe the original NBitcoin code had no console outputs, so to be equivalent with that I try not to log over Debug level unless i'm sure such an exception would not be raised under normal situation. but here aggregate exception can be thrown under very normal circumstances in which LogError would pollute console output when nothing went wrong

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: thank you

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: this code no longer exists

but in general, when i add logs and when i want to log the result value, sometimes i need to introduce a new variable because you don't want to calculate the result again

like here `return after - before;` to log that you'd need to calculate subtraction again
so instead i introduce result variable which i commonly call `res`, `res` has been in the code base for long time, so that's nothing new, stands for result, just reduces the wrist harm by typing less characters :-)

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: I meant the exception like in ` catch (OperationCanceledException ex)` and do `ex.Message` or something.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: Thanks for investigating this mate.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: ah I didn't realize that this exception was occurring regularly and doesn't constitute a non-expected exception.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: I know it stands for result lol, but I wanted you to give it a more meaningful name :-)
It doesn't exist anymore so no problem.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: i've added the log for that but not the message because i would expect operation cancelled exception to just tell me that the operation was cancelled, i'm not sure there any valuable information inside that message

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: i see, that's another discussion then :-) 
i know what you mean now, my personal preference here is that just as in setter we have this `value` name reserved (for obvious reasons though), then in very most cases, i reserve `res` for the result of the function. this preference of mine may very much be influenced by Pascal language that i used to work in ages ago where there is an implicit Result variable that you set function result to.

so when i see `res` in the code, i know this is going to be the thing to be returned, feels helpful
on the other hand, yes, you are not exactly knowing what is that (e.g. type-wise). but if i compare which one is more, to me it is knowing that that thing is the return value

this is why i (rare exceptions exist) don't use `res` for any calculations, like you would first set it to something, then change it later and then again and only then return it. it is more like when you set `res` to something, there is a chance that this will be returned, it is final in that sense that it won't contain a midstate of a computation

but overall still just a personal preference that you might not agree with, which is fine

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: How can you tell that there is no useful info in that message?
Is it not better to log it, just in case?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
bokobza: res doesn't really tell me what it is that this variable holds. 
I know it's a TimeSpan, I know it's the result, ok, but what does it represent?
In this case, naming it to something like `payloadReturnInterval` would be much more helpful.
No reason why we should name variables to be descriptive, apart from when they're the return object.

src/Stratis.Bitcoin.IntegrationTests/NodeBuilder.cs
dangershony: Its a quick a dirty method to broadcast a trx, handshake, send inv, send trx, ping to see node is still connected.
Its probably mostly used for testing

PR #825, opened 2017-11-29
Removed service provider disposing and added missing dependencies for mining feature

### LINKED ISSUE #799: [FullNode] Is disposed 3 times
Described by @noescape00 in https://github.com/stratisproject/StratisBitcoinFullNode/pull/789#issue-276450273

please continue here with the discussion

noescape00:
And also removed extra ConnectionManager.Disposed() call.
Fix #799


src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -162,6 +162,9 @@ public static IFullNodeBuilder AddMining(this IFullNodeBuilder fullNodeBuilder,
             {
                 features
                     .AddFeature<MiningFeature>()
+                    .DependOn<MempoolFeature>()
```
dangershony: Mining I would say probably also depends on Consensus and Store

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -187,7 +187,6 @@ public override void Start()
 
             this.disposableResources.Add(this.timeSyncBehaviorState);
             this.disposableResources.Add(this.chainRepository);
-            this.disposableResources.Add(this.connectionManager);
```
dangershony: Where is `LoggerFactory` getting disposed?
Can you also do the same with `LoggerFactory` remove it from this list and put it where ever we dispose LoggerFactory

src/Stratis.Bitcoin/FullNode.cs
```diff
@@ -217,12 +217,11 @@ public void Stop()
 
             this.ConnectionManager.Dispose();
 
-            foreach (IDisposable dispo in this.Resources)
-                dispo.Dispose();
+            foreach (IDisposable disposable in this.Resources)
```
dangershony: please dispose of this at the end 
`  this.ConnectionManager.Dispose();`
and right after it dispose of 
`this.nodeSettings.LoggerFactory.Dispose()`

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
noescape00: I didn't add them because Stratis.Bitcoin.Features.Miner project don't have references to them.
Would you like me to reference them if order to make change you've mentioned?

dangershony: Mempool has a reference right?
I think we should one will never mine without running a fullnode.

src/Stratis.Bitcoin/FullNode.cs
```diff
@@ -215,13 +215,14 @@ public void Stop()
...
+            // Fire the NodeFeatureExecutor.Stop.
+            this.fullNodeFeatureExecutor.Stop();
 
             foreach (IDisposable disposable in this.Resources)
                 disposable.Dispose();
 
-            // Fire the NodeFeatureExecutor.Stop.
-            this.fullNodeFeatureExecutor.Stop();
+            this.ConnectionManager.Dispose();
+            this.NodeService<NodeSettings>().LoggerFactory.Dispose();
```
dangershony: You could inject settings in the constructor similar to ConnetcionManager

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
noescape00: >Mempool has a reference right?
Yes. 
Miner references Stratis.Bitcoin, MemoryPool, RPC and Wallet.

src/Stratis.Bitcoin/FullNode.cs
noescape00: It doesn't have a constructor and we are getting CM this way:
`this.ConnectionManager = this.Services.ServiceProvider.GetService<IConnectionManager>();`

Would you like me to add constructor to full node and use injection for getting the settings?

fassadlr: Yes, just create a private field:
```
        /// <summary>Logger factory to create loggers.</summary>
        private readonly ILoggerFactory loggerFactory;
```

And set it in `Initialize`, then you can dispose of it properly in `Stop `.

noescape00: Fixed.

fassadlr:
Nice @noescape00 this will up that pesky #799 issue nicely. I'm happy with it if @dangershony s happy?

PR #827, opened 2017-11-29
Replaced Stop with Dispose for all features and node.

noescape00:
Replaced Stop with Dispose for all features and FullNode.
The reason for that is that all Stop() calls actually dispose stuff instead of stopping the feature.  

And in case of node there were Stop() and Dispose():  Stop() would do the disposing (leaving IsDisposed flag false), Dispose() would call Stop() which is a very counter-intuitive behavior. 

This is discussed in #799

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -118,7 +118,7 @@ namespace Stratis.Bitcoin.Features.Miner
         }
 
         ///<inheritdoc />
-        public override void Stop()
+        public override void Dispose()
         {
             this.powLoop?.Dispose();
             this.posLoop?.Dispose();
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -94,7 +94,7 @@ namespace Stratis.Bitcoin.Features.Miner
             }
         }
 
-        ///<inheritdoc />
+        /// <inheritdoc />
         public override void Start()
         {
             if (this.minerSettings.Mine)
@@ -117,14 +117,14 @@ namespace Stratis.Bitcoin.Features.Miner
             }
         }
 
-        ///<inheritdoc />
+        /// <inheritdoc />
         public override void Dispose()
         {
             this.powLoop?.Dispose();
             this.posLoop?.Dispose();
         }
 
-        ///<inheritdoc />
+        /// <inheritdoc />
         public override void ValidateDependencies(IFullNodeServiceProvider services)
         {
             if (services.ServiceProvider.GetService<PosMinting>() != null)
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -95,7 +95,7 @@ namespace Stratis.Bitcoin.Features.Miner
         }
 
         /// <inheritdoc />
-        public override void Start()
+        public override void Initialize()
         {
             if (this.minerSettings.Mine)
             {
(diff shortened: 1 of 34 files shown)


src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -118,7 +118,7 @@ public override void Start()
         }
 
         ///<inheritdoc />
```
Aprogiena: space after `///`

src/Stratis.Bitcoin.Features.Notifications/BlockNotificationFeature.cs
```diff
@@ -51,7 +51,7 @@ public override void Start()
             this.chainState.ConsensusTip = this.chain.Tip;
         }
 
-        public override void Stop()
+        public override void Dispose()
```
Aprogiena: if you do this, you can just make sure to inject `/// <inheritdoc/>` everywhere ...

dangershony:
Hmm I am not sure I am to happy with the name change @bokobza @fassadlr can you guys comment?

If we change to Dispose we sort of give the code a different meaning.
Stop to me feel it makes this more part of the IDisposable pattern and by using Start/Stop that give us more control.

If we did change Stope -> Dispose
Then I suggest we also change Start -> Initialize

Aprogiena:
@dangershony i know what you mean. however, to achieve what you are suggesting, we should distinguish between disposing the object and stopping. currently, what Stop usually do is that it does both and in many cases there is no stopping code, so it only disposes

if we make such distinction, then we will be have Start/Stop as well as Dispose, but as mentioned earlier, the way .NET Core chose is to only use Dispose unless the object is somehow restartable (i.e. it makes sense to do something with the object after it is stopped) and in many cases it does not make sense in our code. maybe in some classes it would - such should be separated, the rest could probably adopt this Dispose only approach

fassadlr:
@dangershony Yes I agree, we can't rename Stop and keep Start. We have to be consistent throughout.

@noescape00 A while back there was a discussion between myself and Dan to possible implement a IStartable interface which will have Start and Stop methods. This is to ensure that features get started and stopped. We haven't gotten around to do it yet but the idea was to actually force all features to start and stop. @noescape00 please note that some features' "Stop" methods actually are a teeny bit more complex, i.e. disposing of asyncloops etc and as such I also agree that we should keep Start and Stop from an architectural view point.

In the future Start and Stop logic might even get more complex and then we'll have to revert what you did now. Its just better to allow room for the future.

HOWEVER, I do feel that we need to decide on a pattern, its not good practise to mix the 2 in my opinion. Either have a Start and Stop pattern (which I prefer) Or use the Dispose pattern throught to stop asyncloops etc.

(Edit what @Aprogiena said)

noescape00:
>In the future Start and Stop logic might even get more complex and then we'll have to revert what you did now. Its just better to allow room for the future.

Not sure it will be about reverting, more like adding. 

Right now most of the code don't implement Start() and Stop() properly. In my opinion this naming convention suggests that we can do
```
Start();
Stop();
Start();
```
and everything would actually work. But most of the features are not restartable so having this names is not really logical for me.

Though I agree this PR doesn't really solves the problem because it moves from one badly implemented pattern to another slightly less badly implemented.

Aprogiena:
i agree with @noescape00 (seems like i agree with everyone here actually). so at this point it seems that very little (if any?) Stop of any class is implemented as it should (start-stop-start possible). in that case we should have these Disposes as @noescape00 implemented but we should also do what @dangershony suggested at the same time - rename Start to Initialize

dangershony:
>i.e. it makes sense to do something with the object after it is stopped

Ok yeah this convinced me, we are indeed disposing/shutting-down, not stopping a restartabe object

And seems we are in agreement to rename `Start` to `Initialize`

dangershony:
@noescape00 do you use resharper or are you as fast as @Aprogiena in typing? 👍

noescape00:
@dangershony resharper. It has an awesome renaming feature that would rename every override, reference in xmldoc and comments.

PR #829, opened 2017-11-30
updated to osx image 9 on Travis

bokobza:


bokobza:
I'll push it in if it builds successfully.

PR #830, opened 2017-11-30
Script reader possible memory leak fix.

### LINKED ISSUE #801: [ScriptReader] Undisposed -> Leak?
`NBitcoin.ScriptReader` is not `IDisposable`, but it contains inner `Stream`, often `MemoryStream`. Should not this inner stream be disposed?

noescape00:
Made ScriptReader disposable and changed every ScriptReader's usage in order to dispose it after it's not needed. 

Fix #801


fassadlr:
@noescape00 Before I merge, please confirm that you have run the node in test and main to ensure that its still syncing etc?

noescape00:
@fassadlr
Sure, I confirm. 
I always sync before pushing any commit when I do something non-cosmetic.

fassadlr:
Thanks @noescape00!

Ah I see there's some conflicts now.

noescape00:
Ok, I'm fixing it.

src/NBitcoin/Script.cs
```diff
@@ -807,8 +814,10 @@ private static BitcoinStream CreateHashWriter(HashVersion version)
 
         public IEnumerable<Op> ToOps()
         {
-            ScriptReader reader = new ScriptReader(_Script);
-            return reader.ToEnumerable();
+            using (ScriptReader reader = new ScriptReader(_Script))
+            {
+                return reader.ToEnumerable().ToList();
```
bokobza: Does `return reader.ToEnumerable();` not work here?

src/NBitcoin/ScriptReader.cs
```diff
@@ -748,9 +748,12 @@ public IEnumerable<Op> ToEnumerable()
...
             Op code;
             while((code = Read()) != null)
-            {
                 yield return code;
-            }
+        }
+
+        public void Dispose()
+        {
+            this._Inner.Dispose();
```
bokobza: what happens if the object is null?

noescape00: Hmm. 
I assumed that check in the constructor would be enough. 
But it is possible to set it to null after the creation so I'll add a null check. 
Thx for pointing that out.

src/NBitcoin/Script.cs
noescape00: No. It does git disposed before enumeration starts. 

But we can afford loading it as a whole list- scripts are small anyway and in every case that enumerable would enumerate it to the end so all of it will eventually be loaded.

bokobza: ah yes fair enough.

dangershony:
I agree its very hard to review this @noescape00 its impossible to know if you made logic changes without a lot of work from the reviewer. 
Is there anything we can do to fix this?

src/NBitcoin/ScriptReader.cs
```diff
@@ -696,46 +696,46 @@ public bool IsSmallInt
...
+    public class ScriptReader : IDisposable
     {
         private readonly Stream _Inner;
+
         public Stream Inner
         {
             get
             {
-                return _Inner;
+                return this._Inner;
```
dangershony: If you are already making some changes here then rename `_Inner` to `inner`

dangershony:
Ok after Jeremys little trick to append `?w=1` to the url on the review page, that removes all white spaces.
I am all good on this review.

src/NBitcoin/ScriptReader.cs
noescape00: Fixed.

PR #831, opened 2017-11-30
Refactor/the node job part 5

Aprogiena:
This episode is about replacing direct date time functions access with access through date time provider. There are some minor improvements here and there as well, but just few.


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -465,7 +465,7 @@ public NetworkPeer(NetworkAddress peerAddress, Network network, NetworkPeerConne
             parameters = parameters ?? new NetworkPeerConnectionParameters();
             this.Inbound = false;
             this.Behaviors = new NetworkPeerBehaviorsCollection(this);
-            this.MyVersion = parameters.CreateVersion(peerAddress.Endpoint, network);
+            this.MyVersion = parameters.CreateVersion(peerAddress.Endpoint, network, this.dateTimeProvider.GetTimeOffset());
```
bokobza: you're passing GetTimeOffset but the `CreateVersion` method was actually using  `DateTimeOffset.UtcNow` before.
Is that intentional?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
```diff
@@ -133,7 +133,7 @@ public NetworkPeer CreateConnectedNetworkPeer(Network network, IPEndPoint endpoi
         {
             var peerAddress = new NetworkAddress()
             {
-                Time = DateTimeOffset.UtcNow,
+                Time = this.dateTimeProvider.GetTimeOffset(),
```
bokobza: UtcNow to TimeOffset, is that intentional?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: Yes, that is another thing, I'd like to get rid completely of DateTimeOffset from FN, but that's different thing than introducing date time provider. So now the idea was to replace every `DateTimeOffset.UtcNow` with `this.dateTimeProvider.GetTimeOffset()` and every `DateTime.Now` or `DateTime.UtcNow` to be replaced with `this.dateTimeProvider.GetUtcTime()`.

So at this point I want to preserve the same types, just use the provider interface for delivering the actual value.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
Aprogiena: yes, same as before, same time, different delivery. UtcNow - check DTP.GetTimeOffset how it is implemented:

```
        public virtual DateTimeOffset GetTimeOffset()
        {
            return DateTimeOffset.UtcNow;
        }

```

bokobza:
ok mate that's good to go.

bokobza:
oh, let's wait for the builds, at least on appveyor.

PR #834, opened 2017-11-30
Unused 'out' optimizations

noescape00:
Tiny pr that replaces 
```
var unused;
fu(out unused)
```

with 

`fu (out var unused)`

### DIFF
--- a/src/NBitcoin/Script.cs
+++ b/src/NBitcoin/Script.cs
@@ -1018,14 +1018,12 @@ namespace NBitcoin
 
         public static bool VerifyScript(Script scriptSig, Script scriptPubKey, Transaction tx, int i, ScriptVerify scriptVerify = ScriptVerify.Standard, SigHash sigHash = SigHash.Undefined)
         {
-            ScriptError unused;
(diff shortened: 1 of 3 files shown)


src/NBitcoin/Script.cs
```diff
@@ -1088,13 +1085,11 @@ public enum BitcoinConsensusError
 
         public static bool VerifyScriptConsensus(Script scriptPubKey, Transaction tx, uint nIn, ScriptVerify flags)
         {
-            var err = BitcoinConsensusError.ERR_OK;
-            return VerifyScriptConsensus(scriptPubKey, tx, nIn, flags, out err);
+            return VerifyScriptConsensus(scriptPubKey, tx, nIn, flags, out BitcoinConsensusError unused);
```
bokobza: mmh this one is interesting.
`err` is explicitely set, but it's not used and instead set to the same value in `VerifyScriptConsensus`.
I think it's ok to remove it, as you did.

dangershony:
Looks good to me as well

Issue #124, opened 2017-04-25
Discussion: Consensus Validation Rules


dangershony:
An idea I think can be useful for the consensus code is to break many of the validation logic inside the consensus validator class in to ValidationMacros (or similar) to have smaller snippets of code.

Then on startup a node will load all its consensus validation macros to a container and when a block is received each block will pass in all the macros in order to be considered valid.
A Context will be passed in to every Macro for Macros that need previously calculated data

The benefits:
- Readability - a macro will be named based on its logic (i.e CoinBaseMaturityMacro) 
- Testability - it will be much easier to write tests for each macro
- ReUsability - POW/POS/PO..X consensus rules will be able to reuse macros  
- ChainSpecific - currently POS has hard coded checks for protocol version, that logic is Stratis specific 
- Optimization - some macros will be able to run in parallel  

I don't think there will be any performance degradation on iterating over several tens of macros.
I am also not sure Macro is the right name.

Derrick-:
Perhaps they are "Rules"

I think this is a very good idea.

dangershony:
Yeah @bokobza also suggested the word Rules I am not against that, so in the folder ValidationRules we'll have say a CoinBaseMaturityRule

NicolasDorier:
I think the consensus code is tricky and should be kept as close as possible to bitcoin core to prevent bugs.
The ConsensusValidator class already break out functions that you can reuse outside of the consensus loop.

dangershony:
I agree consensus is tricky indeed, and I think an implementation should never run on two languages so the Stratis bitcoin node will never be a replacement for the Bitcoin chain, and after we have the stratis chain running in prod for many good month we may retire the stratis c++ version in favor of the C# version.

So if consensus is not a crusial factor I believe we can safely break it smaller to Rules.

Derrick-:
Yes, the way I interpret what you'll be doing is just breaking the consensus rules into separate functions, and this also makes things more testable I hope. They will still all be called from the same place(s)

NicolasDorier:
what smaller snippet are you thinking about ? I think the methods are already fairly broken out already in the ConsensusValidator.

dangershony:
Some examples may look like this:
```
		public abstract class ConsensusRule
		{
			public abstract void Check(ContextInformation context);
		}

		public class CoinMaturityConsensusRule : ConsensusRule
		{
			public override void Check(ContextInformation context)
			{
				// checking that each trx prev output is not spending
[...]

PR #835, opened 2017-11-30
Consensus rules

dangershony:
This is the first part of what the implementation of consensus rules will potentially look like.

This code will plug in to the consensus loop a call to the rules containing some of the base rules the node validates for example:
```
            public IEnumerable<ConsensusRule> GetRules()
            {
                yield return new BlockPreviousHeaderRule();

                // rules that are inside the method ContextualCheckBlock
                yield return new Bip113ActivationRule();
                yield return new Bip34ActivationRule();
                yield return new WitnessCommitmentsRule();
                yield return new BlockSizeRule();

                // rules that are inside the method CheckBlock
                yield return new BlockMerkleRootRule();
                yield return new EnsureCoinbaseRule();
                yield return new CheckTransactionRule();
                yield return new CheckSigOpsRule();
            }
```

Note1: this rules are not yet deleted form the main execution path (they will run twice)
Note2: the test project will not compile so the build will break.

This PR is not intended to be committed yet, and still needs some styling and documentation.


Aprogiena:
How is that yield series better than just a list of rules?

Also, could we have all the new code commented as per our policies?

And finally, could we not introduce "ugly hacks for now"? Seems like that ConsensusLoop is not needed inside the rules at all except for the Tip, which can easily be part of the context that is being passed to the execution methods.

dangershony:
Thanks for having a look mate

>How is that yield series better than just a list of rules?
Not necessarily better just a way of doing it.

>Also, could we have all the new code commented as per our policies?
yeah this is still pending (see my comments at the top)

>which can easily be part of the context
you are always full of great idea, indeed I will put a tip in the context.

Aprogiena:
Ah sorry i missed the very last sentence. No problemo then :-)

As for the yield series - I find that enormously disturbing. Just the inner feeling, like you are watching some weird video and you know it is wrong (but you keep watching anyway). 

I mean, of course it is a way of doing it, but we don't need a gold medal for ugliest code. 

Enough of destructive criticism:
```
private List<ConsensusRule> DatRulzOfDatStratis = new List<ConsensusRule>()
{
    new BlockPreviousHeaderRule(),

    // Rules that are inside the method ContextualCheckBlock.
    new Bip113ActivationRule(),
    ...
};

public IEnumerable<ConsensusRule> GetRules()
{
    return this.DatRulzOfDatStratis;
}
```

or something like that.

dangershony:
Ok I consider this ready to commit, its not finished but more changes will come in following PRs and this is just getting too big.

Next changes:
- documenting further of each rule
- making distinction between execution/validation rules (execution manipulate state)
- handling of consensus errors

And then slowly we will start to move rules out of the main flow in to the rules engine.
This is further discussed in this issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/124

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
```diff
@@ -0,0 +1,59 @@
...
+        public ILogger Logger { get; set; }
+
+        /// <summary>The engine this rule belongs to.</summary>
+        public ConsensusRules Parent { get; set; }
+
+        /// <summary>
+        /// A collection of rules that the current rule depends on.
+        /// </summary>
+        /// <returns>A collection of rules this rule depends on.</returns>
+        public virtual IEnumerable<Type> Dependencies()
```
bokobza: Will that mean `the rules in this list need to be defined to be run at some point` or `the rules in this list need to have been run already`?

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// </summary>
+    public abstract class SkipValidationConsensusRule : ConsensusRule
+    {
+        /// <inheritdoc />
+        public override bool CanSkipValidation => true;
+
+        /// <inheritdoc />
+        public override IEnumerable<Type> Dependencies()
+        {
+            yield return typeof(BlockPreviousHeaderRule);
```
bokobza: This should probably go in a separate file under CommonRules, no?
Seeing `BlockPreviousHeaderRule` as a dependency makes me think that it shouldn't be here.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -0,0 +1,187 @@
...
+    /// A block  that is not valid will result in the <see cref="BlockValidationContext.Error"/> as not <c>null</c>.
+    /// </remarks>
+    public interface IConsensusRules
+    {
+        /// <summary>
+        /// Register a new rule to the engine
+        /// </summary>
+        /// <typeparam name="TRule"></typeparam>
+        /// <returns></returns>
+        ConsensusRules Register<TRule>() where TRule : ConsensusRule, new();
```
bokobza: what are you trying to enforce by having `new()`?

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                var depdendency = this.consensusRules.Keys.SingleOrDefault(w => w.GetType() == dependency);
+
+                if (depdendency == null)
+                {
+                    throw new Exception(); // todo
+                }
+            }
+
+            this.consensusRules.Add(rule, rule);
```
bokobza: `rule, rule` makes me thing that this can probably some kind of collection rather than a dictionary.
Or maybe the key should be the name of the rule?

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -0,0 +1,187 @@
...
+            // TODO: set a rule that will be invoked when a validation of a block failed.
+
+            // This will allow the creator of the blockchain to provide  
+            // an error handler rule that is unique to the current blockchain.
+            // future sidechains may have additional handling of errors that
+            // extend or replace the default current error handling code.
+        }
+
+        /// <inheritdoc />
+        public async Task ExectueAsync(BlockValidationContext blockValidationContext)
```
bokobza: ExectueAsync --> ExecuteAsync

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -233,10 +247,32 @@ public static IFullNodeBuilder UseStratisConsensus(this IFullNodeBuilder fullNod
...
                         services.AddSingleton<ConsensusSettings>();
+                        services.AddSingleton<IConsensusRules, ConsensusRules>();
+                        services.AddSingleton<IRuleRegistration, BaseConsensusRules>();
                     });
             });
 
             return fullNodeBuilder;
         }
+
+        public class BaseConsensusRules : IRuleRegistration
```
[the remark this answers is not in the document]
dangershony: Yeah agree I put it there for now but it will move in next PR, I would be happy to discuss where this should move in the dedicated issue.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -233,10 +247,32 @@ public static IFullNodeBuilder UseStratisConsensus(this IFullNodeBuilder fullNod
...
+        {
+            public IEnumerable<ConsensusRule> GetRules()
+            {
+                yield return new BlockPreviousHeaderRule();
```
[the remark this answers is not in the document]
dangershony: Did @Aprogiena put you up to this?

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
dangershony: Ah yes good point, to me it means the Rules have to exist and have run before this rule.
Some rules may update the context with stuff and later rules may depend on that updated context.
Of also rules that are cheep to execute we want to check first.

I considered using such a construction `ConsensusRule<TDependentRule>` where  TDependentRule is the dependent rule, how ever this wont work in case of more then one rule dependency.
We could also provide a few of such base classes `ConsensusRule<TDependentRule1,TDependentRule2`>

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
dangershony: yeah and I actually deleted is in my next PR.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
dangershony: I will delete this method

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
dangershony: yes perhaps `rule.GetType().FullName`

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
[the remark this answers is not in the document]
bokobza: yeah he gave me a 1.5 BTC.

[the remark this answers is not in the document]
Aprogiena: @dangershony everyone can see this as ugly, follow the herd :-P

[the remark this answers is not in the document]
MithrilMan: +1 for yield uglyness, sorry dan :)

MithrilMan:
preamble: i didn't saw yet the code (just some pieces) but read this thread

About Dependencies:
Implementing dependencies, Isn't a flag of over-engineering?
Separation of concern is good, I very like the concept of Rules and your proposed design is what usually is considered one of the best practice to organize business rules validation, but building dependency trees of validation could complicate the rule design a lot and cause lot of troubles.
If you go deeply into rules dependencies you need even to check the rule execution order and having a dependency list doesn't state how that rules should be handled

So maybe would be better to have a kind of ExecutedRules stack that you can check into the rule itself to see if a particular (named) rule has already been executed, instead of declare it on the rule body and enforce it at definition time. Even because some rules could be mandatory, some optional, and so having a mechanism to declare your needs/intention using a Dependency list isn't enough.
And how do you handle the dependant rules that hinerit from your SkipValidationConsensusRule class?
I think that having the rule itself check the "ExecutedRules" could lead to a cleaner code.

About the Rule design, you can do it in several ways: implementing an interface like you did or coding those validations as extension of the object to validate (could even be your context object), there are pros and cons for boths, I've not really a preference in this scenario yet (but again, i should read the whole code)

Another thing i'd like to add, regarding rules dependencies, or better an ExecutedRules stack (that could be a dictionary tbh), is that this way you can have multiple rules that check if a rule has been executed and if not, you could decide to trigger it's validation, or throw (i prefer to throw because it means you designed badly your rules set)

a pseudo rule example could be (following your interface pattern)
```public class Rule1 : SkipValidationConsensusRule
    {
        /// <inheritdoc />
        public override Task RunAsync(ContextInformation context)
        {
            ...
           if(context.ExecutedRules.Exists(nameof(MyNeededRule){
                go on
           }
           else{
               rule failed, missing MyNeededRule validation
           }
           ...
        }
}```

the ExecutedRules could implement a guard pattern itself, so you could call it as
`context.ExecutedRules.HasValidated<MyNeededRule>()`
that will throw if that rule didn't validated, or you could add an overload that accept a bool to force that rule execution if it isn't executed yet (but i don't like this)

oh, almost forget, i'm a lover of Fluent apis, so I'd like those rules to be chainable!
so when you define your ruleSet, you can do something as

```
myRuleSetBuilder
     .RegisterRule<MyRule1Type>()
     .RegisterRule<MyRule2Type>()
```

you could even define rule grouping in fluent way 
```
myRuleSetBuilder
     .RegisterRule<MyRule1Type>()
     .RegisterRule<MyRule2Type>()
     .RegisterRuleSet("MyRuleSetName")
          .RegisterRule<MyRule3Type>()
          .RegisterRule<MyRule4Type>()
     .EndOfRuleSet()
     .RegisterRule<MyRule5Type>()
```

about performance, we could use Roslyn to chain those dynamically if needed, but i think that this concern should rise only if really needed

my2cents

PR #837, opened 2017-12-01
Implement sub-classed peer connectors

fassadlr:
As the logic to select peers to connect to is different based on discovery, -addnode or -connect args, specific peer connectors needs to implemented to separate the logic. This also ensures that the 3 different scenarios can be better unit tested.


src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -254,7 +242,8 @@ public void AddDiscoveredNodesRequirement(NetworkPeerServices services)
         {
             this.logger.LogTrace("({0}:{1})", nameof(services), services);
 
-            this.discoveredNodeRequiredService |= services;
+            var discoveredNodeRequiredService = NetworkPeerServices.Network;
```
dangershony: `NetworkPeerServices.Network` means the node can serve blocks
This is probably very wrong, you can't know at this point if the node service is Network, only the BlockStore can do that as only when block store is present the node can serve blocks to peers

dangershony: Whats the reason behind setting this here and not in store? did you hit an issue?

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -185,21 +184,21 @@ public void PeerHandshaked(IPEndPoint endpoint, DateTimeOffset peerHandshakedAt)
         /// <inheritdoc/>
         public PeerAddress FindPeer(IPEndPoint endPoint)
         {
-            var peer = this.Peers.SingleOrDefault(p => p.Value.Match(endPoint));
+            var peer = this.Peers.SingleOrDefault(p => p.Key.Match(endPoint));
```
dangershony: I'd suggest a `.Skip(0)`

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -102,47 +107,35 @@ public sealed class PeerConnector : IPeerConnector
...
+        protected PeerConnector(NodeSettings nodeSettings, IPeerAddressManager peerAddressManager)
         {
+            Guard.NotNull(nodeSettings, nameof(nodeSettings));
             Guard.NotNull(peerAddressManager, nameof(peerAddressManager));
 
-            this.nodeLifetime = new NodeLifetime();
             this.peerAddressManager = peerAddressManager;
-            this.peerIntroductionType = peerIntroductionType;
+            this.NodeSettings = nodeSettings;
+            this.nodeLifetime = new NodeLifetime();
```
dangershony: Ouch mate... 😮

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -102,47 +107,35 @@ public sealed class PeerConnector : IPeerConnector
...
-            INodeLifetime nodeLifeTime,
-            NetworkPeerConnectionParameters parameters,
-            NetworkPeerRequirement requirements,
-            Func<IPEndPoint, byte[]> groupSelector,
-            IAsyncLoopFactory asyncLoopFactory,
-            IPeerAddressManager peerAddressManager,
-            PeerIntroductionType peerIntroductionType,
-            INetworkPeerFactory networkPeerFactory)
+        /// <summary>Constructor used by <see cref="Connection.ConnectionManager"/>.</summary>
+        protected PeerConnector(PeerConnectorContext context)
```
dangershony: If context is only used to pass types in the constructor you are only introducing complexity (both for testing and dependency injection)

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -159,13 +152,18 @@ public void RemovePeer(NetworkPeer peer)
             this.ConnectedPeers.Remove(peer);
         }
 
+        internal bool IsPeerConnected(IPEndPoint endPoint)
+        {
```
dangershony: I thought we are going to get rid of RelatedPeerConnector and pass the ConnectedNodes instead no?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -218,33 +222,7 @@ private void Disconnect()
...
-                    groupFail++;
-                    continue;
-                }
-
-                break;
-            }
-
-            return peer;
-        }
+        internal abstract NetworkAddress FindPeerToConnectTo();
```
dangershony: Protected rather then Internal.
Internal prevents later implementers to reuse this technique

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -102,47 +107,35 @@ public sealed class PeerConnector : IPeerConnector
...
-            this.peerAddressManager = peerAddressManager;
-            this.peerIntroductionType = peerIntroductionType;
-            this.Requirements = requirements;
-            this.networkPeerFactory = networkPeerFactory;
-
-            this.currentParameters = this.parentParameters.Clone();
-            this.currentParameters.TemplateBehaviors.Add(new PeerConnectorBehaviour(this));
-            this.currentParameters.ConnectCancellation = this.nodeLifetime.ApplicationStopping;
+            this.network = context.Network;
+            this.nodeLifetime = context.NodeLifetime;
```
dangershony: Ah I see you override NodeLifetime here.

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -0,0 +1,59 @@
...
+                this.peerAddressManager.AddPeer(new NetworkAddress(endPoint.MapToIpv6()), IPAddress.Loopback);
+            }
+        }
+
+        /// <inheritdoc/>
+        internal override NetworkAddress FindPeerToConnectTo()
+        {
+            PeerAddress peer = null;
+
+            foreach (var endPoint in this.NodeSettings.ConnectionManager.AddNode)
```
dangershony: Ahh excellent idea mate

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                peer = this.peerAddressManager.FindPeer(endPoint.MapToIpv6());
+                if (peer == null)
+                    continue;
+
+                // If the peer is already connected just continue.
+                if (this.IsPeerConnected(endPoint))
```
dangershony: I guess your next PR will replace the contents of `IsPeerConnected` to get rid of the group and replace it with conectionManager.ConnectedPeers

src/Stratis.Bitcoin/P2P/PeerConnectorContext.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.P2P.Peer;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.P2P
+{
+    /// <summary>Context class for <see cref="PeerConnector"/>.</summary>
+    public sealed class PeerConnectorContext
```
dangershony: I have an issue with putting a Context here, the types in this context are all singletons available in the DI.
A Context I normally use as a type that hold state for a duration of an operation.

I would propose to just use our normal pattern of passing Interfaces in to the PeerConenctor, and for testing just create a PeerTestFactory that creates a PeerConnector with all the interfaces mocked away.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                peer = this.peerAddressManager.SelectPeerToConnectTo();
+
+                if (!peer.Endpoint.Address.IsValid())
+                {
+                    peerSelectionFailed++;
+                    continue;
+                }
+
+                // If the peer exists in the -addnode collection don't
```
dangershony: AHHHH very good way to overcome this issue good job mate

dangershony:
I very much like it now, just make some changes as per comments and we are all good.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: `discoveredNodeRequiredService` was actually set to `NetworkPeerServices.Network` at field declaration. See a bit higher up in the diff.... But I'm going to revert this bit of code rather 👍

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
fassadlr: Added

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Yes this constructor is only used for unit testing ;)

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Ah I only did this for readability, as its easier to create the context once and then just pass it into the constructor. Also, PeerConnector doesn't get its constructor argument's DI-ed.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Yes the plan was to do this in the next PR 👍

src/Stratis.Bitcoin/P2P/PeerConnector.cs

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: 🥇

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: Correct

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
fassadlr: 🥇

src/Stratis.Bitcoin/P2P/PeerConnectorContext.cs
fassadlr: OK I'll revert this then 👍

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Ah no @dangershony this actually needs to be on the IPeerConnector interface as public so that I can call it from the unit tests ;)

src/Stratis.Bitcoin/P2P/PeerConnector.cs
dangershony: >Also, PeerConnector doesn't get its constructor argument's

yes but maybe we will be able to find a way to do it at some point

src/Stratis.Bitcoin/P2P/PeerConnector.cs
dangershony: ah right yes

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -100,6 +100,8 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
 
         private readonly Dictionary<NetworkPeer, PerformanceSnapshot> downloads = new Dictionary<NetworkPeer, PerformanceSnapshot>();
 
+        private NetworkPeerServices discoveredNodeRequiredService = NetworkPeerServices.Network;
```
dangershony: Ah ok yes this is about what we are looking for when discovering peers and not what are our capabilities.
yes this should probably be Network in that case.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
dangershony: If you create a test factory in a proper way you wont need this constructor for testing

PR #840, opened 2017-12-01
Refactor/the node job part 7

Aprogiena:
In this episode we care mostly about NetworkPeerServer class:
- couple of undisposed CTSs are not disposed
- add comments (except for main class comment)
- remove unused/unusful stuff
- add some logs
- simplified couple of mechanisms

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -82,8 +82,6 @@ namespace Stratis.Bitcoin.P2P.Peer
 
         public NetworkPeerCollection ConnectedNetworkPeers { get; private set; }
 
-        private List<IDisposable> resources = new List<IDisposable>();
-
         private CancellationTokenSource cancel = new CancellationTokenSource();
 
         private ulong nonce;
@@ -105,6 +103,8 @@ namespace Stratis.Bitcoin.P2P.Peer
         public event NetworkPeerServerNodeEventHandler PeerRemoved;
         public event NetworkPeerServerNodeEventHandler PeerAdded;
         public event NetworkPeerServerMessageEventHandler MessageReceived;
+
+        private readonly EventLoopMessageListener<IncomingMessage> listener;
 
         /// <summary>
         /// Initializes instance of a network peer server.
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -12,9 +12,6 @@
 
 namespace Stratis.Bitcoin.P2P.Peer
 {
-    public delegate void NetworkPeerServerNodeEventHandler(NetworkPeerServer sender, NetworkPeer peer);
-    public delegate void NetworkPeerServerMessageEventHandler(NetworkPeerServer sender, IncomingMessage message);
-
     public class NetworkPeerServer : IDisposable
```
dangershony: I suggest we rename this to `PeerListener` this will be consistent with what @fassadlr has the `PeerConnector`
We then need to think if we rename `PeerConnector` to `NetworkPeerConnector`

Aprogiena: sadly NetworkPeerListener already exists, using either NetworkPeerListener or PeerListener for this would be highly confusing at this moment

so maybe at later phase of refactoring?

dangershony: ah ok yes good point.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -369,39 +357,46 @@ private void ProcessMessageCore(IncomingMessage message)
...
-            this.MessageReceived?.Invoke(this, message);
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Callback that is called when a network status of a connected client peer changes.
+        /// <para>The peer is removed from the list of connected peers if the connection has been terminated for any reason.</para>
+        /// </summary>
+        /// <param name="peer">The connected peer.</param>
+        /// <param name="oldState">Previous state of the peer. New state of the peer is stored in its <see cref="NetworkPeer.State"/> property.</param>
```
dangershony: Perhaps the server should have its own behaviour for add/remove.
Just something to consider

Aprogiena: my strategy now is to make these guys as slim as possible and remove all the stuff we don't use, then understand it what is actually there that remained 

and ONLY THEN to somehow change the design, add new features etc.

so in general, yes, anything is possible then, but it will take time to get there

Aprogiena: in other words - if you have an idea, add it to the issue, so that we don't forget about it

Issue #597, opened 2017-10-16
[Wallet] Sending Stratis times out


dev0tion:
As reported by @mtaipe:
"Click send BITCOIN SENDs and confirms message shows  also STRATIS  sends but no confirmation and modal doesn’t close"

This happens every now and then, sometimes it does get the confirmation window, sometimes it doesn't. To me it looks like the build-transaction is working, but the send-transaction call sometimes times out, thereby not triggering the confirmation window on the UI.

dev0tion:
![](https://github.trello.services/images/mini-trello-icon.png) [[Wallet] Sending STRAT times out](https://trello.com/c/FSz0ebHu/294-wallet-sending-strat-times-out)

nopara73:
Debugged it on Friday.   
Sendtransaction works as intended, it's just if nobody propagates back the message to our node, then we cannot be sure the message was actually propagated. Therefore it times out and the transaction comes when it gets confirmed in a block.

dev0tion:
Ok, but is there any way I can catch this in the UI?

nopara73:
Yes, you got a bad request after 21 seconds of tx not arriving.  

`while (TimeSpan.FromSeconds(21) > waited)`  

`return this.StatusCode((int)HttpStatusCode.BadRequest);`

bokobza:
@nopara73 @dangershony A while ago we talked about sending the trx to the network and then returning to the called straight away (hence relying on the network to propagate it back to us).
I guess that didn't make the cut. 
Can we not reconsider it? Maybe the UI can wait for the trx to show up and show a message if nothing shows up after 21 seconds?

nopara73:
>  A while ago we talked about sending the trx to the network and then returning to the called straight away (hence relying on the network to propagate it back to us).  

What do you mean by this? I don't understand.

> Maybe the UI can wait for the trx to show up and show a message if nothing shows up after 21 seconds?  

That's how it is now, but without the message after 21 seconds.

dangershony:
We had a discussion here and decided a good approach.
We add the trx to the wallet and mark it with a flag (PendingPropagation) in the UI you can grey it out (or set a flag).
Once propagation is confirmed it will be unconfirmed but propagated (as is now). 

We will then provide the user the ability to delete propagated trx (right click and delete)

nopara73:
> We will then provide the user the ability to delete propagated trx (right click and delete)

[...]

PR #844, opened 2017-12-02
SendTransactionAsync refactoring part 1/3

noescape00:
## Motivation
Propagating algorithm is not optimal when we use light node. 
Since we don't have an ability to verify transactions on a light node we rely on other nodes to do the validation. 

**Current light node algorithm for verifying the transaction and knowing if it was propagated:**

1.  Create a transaction.
2.  Broadcast it to half of the peers that we are connected to.
3.  Wait until another half of the peers propagate it back to us.
4.  If we see out tx being propagated from any peer back to us we know that tx is valid and propagated.

**Problems of this algorithm:** 

- If we don't get transaction propagated back to us we end up with timeout exception.
- We will always get a timeout exception if we are only have 1 peer connected.
- Since we use only 50% of the peers the transaction propagates slower than it potentially can.
- If we successfully propagated tx but peers that should propagate it back to us are disconnected and no new peers got connected within a time frame of 21 second we end up with timeout exception.

## Specification
I want to change the algorithm described above to a more efficient and reliable one:

1.  Create a transaction.
2.  Broadcast it to all peers that we are connected to.
3.  Start continuously asking nodes if they have our transaction (_getdata(txid)_).
4.  If at least one peer has our tx then we assume it's valid and propagated. 

## Plan

I'm splitting that changes into 3 parts: 

Part 1: refactoring of the old approach- no algorithm changes. 
Part 2: changing the algorithm, writing tests, comparing it to the old one.  
Part 3: creating documentation for all classes involved in this. 

In the end this PRs will fix issue #597

## This PR

Refactoring: renamed some classes, optimized them; got rid of copy paste code and unused functions and classes, simplified some parts.


Aprogiena:
aww, PR with motivation AND specification section <3

Aprogiena:
Although I love the description of this PR and I think it could be great to improve how transactions are handled, this very first part of the whole process lacks value for me - I'm not sure the selected names are significantly better (I actually think InvPayload is slightly better than the new InventoryPayload due to the fact that the network message is just "inv").

However, if someone else sees the new naming as better than what there was before and approves it I'm perfectly fine with that.

noescape00:
InventoryPayload is already reverted to InvPayload so basically here we are discussing BroadcasterBehavior => BroadcastBehavior. 
If no one find new name more suitable I'm revering it back as well since I don't have a strong position on naming of that one.

Aprogiena:
so for the alg - we need issue discussing the alg before we start with works here
i can see privacy problems with this alg - if you ask about a transaction like this, it seems like you are the one interested in it ...

bokobza:
Please let me review this before merging.

noescape00:
Sure.

Btw don't mind all that info about the new algo. This particular PR is just refactoring of the old algorithm so merging it wont change the project.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -139,6 +140,15 @@ public class BuildTransactionRequest : TxFeeEstimateRequest
 
     public class SendTransactionRequest : RequestModel
     {
+        public SendTransactionRequest()
```
bokobza: What's the purpose of having this constructor defined like this?

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -139,6 +140,15 @@ public class BuildTransactionRequest : TxFeeEstimateRequest
...
+        {
+        }
+
+        public SendTransactionRequest(string transactionHex)
```
bokobza: Is there any need for this to be added?

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
```diff
@@ -1352,7 +1352,7 @@ public void SendTransactionSuccessfulReturnsWalletSendTransactionModelResponse()
 
             var mockWalletWrapper = new Mock<IBroadcastManager>();
             mockWalletWrapper.Setup(m => m.TryBroadcastAsync(transaction))
-                .Returns(Task.FromResult(Success.Yes));
+                .Returns(Task.FromResult(true));
```
bokobza: If Success.Yes == true and Success.No == false, how about Success.DontKnow?

bokobza:
OK so I've had a look at this, but I'm having a really hard time making sense of it.
Some commits have too many things in them: renaming of methods, methods moving around, file renames, all in one commit. In particular, commit 3fa82ce: all of these little changes in this commits should be separate commits.
Please try as much as you can to commit often. Changing a name? commit. moving some methods around? commit. this makes it easier for us. At the moment I'm having to open a comparison software and copy-pasting files that are not comparable on GitHub.

Also, there is no gain in doing small refactoring and changing names if we don't even know what we want ultimately.
The discussion should come first, then you can crack on with the agreed work.

Now, the priority here is to fix the fact that an exception in the TryBroadcastAsync method call in 
 SendtransactionAsync method would result in a timeout message.
After this is fixed, we can talk about architecture, then we can do the work.

So I suggest we 
1. put this PR on hold (or even close it) 
2. work on fixing the issue described **as a priority**
3. agree architecture
4. do the work

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
noescape00: Yes. Most of the SendTransactionRequest instances were created like 

`new SendTransactionRequest() {
Hex = *
}`

So this change introduces a way to create SendTransactionRequest in one line instead of 3 and make the code cleaner and more understandable.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
noescape00: Not sure. 
I should check if constructor without parameters is actually used in the code. If it's not used we can delete it.

noescape00:
>Please try as much as you can to commit often.

Ok, I'll keep it in mind for the future PR's so it's easier to review them. 

>Also, there is no gain in doing small refactoring and changing names if we don't even know what we want ultimately.

I think the strongest point I have on having this PR is that it simplifies the code and removes a lot of copy pasted code (look at master's LightWalletBroadcastManager and FullNodeBroadcastManager- they have lot's of the same code that can be moved to the base class). 

>Now, the priority here is to fix the fact that an exception in the TryBroadcastAsync method call in
SendtransactionAsync method would result in a timeout message.

The reason while it fires this exception lies is deep in the algorithm that was used. 
By changing the algorithm we will fix the exception. 
And if we fix the exception now the outcome won't be correct from the logical perspective. 
Anyway even if we fix it without fixing the algorithm this PR is still usefull because fix will be on top of current code (and version represented in this PR is cleaner so it will be easier to do the fix using this codebase instead of the old one). 


PS. 
If the reasons presented above are not convincing enough I'm totally ok with your choice of discarding this PR and doing it the way you've suggested.

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
noescape00: There is no Success.DontKnow anymore. 
In previews version if we were getting Success.DontKnow from the LightWalletBroadcastManager we would start waiting for 21 sec for tx to be propagated. If it's not we assume it's false and throw. 
In my implementation I moved that 21 sec waiting to the LightWalletBroadcastManager itself because this behavior only makes sense for the light node. And that's why I was able to get rid of the Success.DontKnow.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
bokobza: even if it's used, it's created by default when doing `new SendTransactionRequest();` ;-)

noescape00: If the class already contains a constructor that requires any parameters we can't do 
new SendTransactionRequest(); without specifically creating a parameter-less constructor. 
So that's why I added it- in case somewhere new SendTransactionRequest() is still used.

bokobza: ah yes indeed.

bokobza:
> And if we fix the exception now the outcome won't be correct from the logical perspective.

By the same token, if you fix some naming, the outcome won't be correct from the logical perspective. I'd rather concentrate on fixing an issue that some naming if at the end the result will still be incorrect.

> Anyway even if we fix it without fixing the algorithm this PR is still usefull because fix will be on top of current code (and version represented in this PR is cleaner so it will be easier to do the fix using this codebase instead of the old one).

OK fair enough, I get that. My main point here is that this PR is difficult to review and that changing too much unrelated code in 1 commit hides your intent in most cases.

But let's not drag this any longer. Please either fix the problem we know about or resubmit this PR, splitting it into multiple commits and having exactly 0 functionality changes.

noescape00:
Ok. 
I'm closing this PR, making the commits easy to review and resubmitting it again.

bokobza:
Thanks for understanding @noescape00 .

noescape00:
You are welcome @bokobza 
I should have thought about this in the first place =)

bokobza:
nah it's fine, it's us that should make the rules clearer, or point new contributors to some docs describing our preferences. We're also learning in the process :-)

PR #845, opened 2017-12-02
Fix small add connector bug

fassadlr:
This is to fix small bug as pointed out by @dangershony .

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
@@ -52,20 +52,20 @@ namespace Stratis.Bitcoin.P2P
         /// <inheritdoc/>
         public override NetworkAddress FindPeerToConnectTo()
         {
-            PeerAddress peer = null;
-
             foreach (var endPoint in this.NodeSettings.ConnectionManager.AddNode)
             {
-                peer = this.peerAddressManager.FindPeer(endPoint.MapToIpv6());
+                PeerAddress peer = this.peerAddressManager.FindPeer(endPoint.MapToIpv6());
                 if (peer == null)
                     continue;
 
                 // If the peer is already connected just continue.
-                if (this.IsPeerConnected(endPoint))
+                if (this.IsPeerConnected(endPoint.MapToIpv6()))
                     continue;
+
+                return peer.NetworkAddress;
             }
 
-            return peer.NetworkAddress;
+            return null;
         }
     }
 }
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
@@ -57,8 +57,6 @@ namespace Stratis.Bitcoin.P2P
                 PeerAddress peerAddress = this.peerAddressManager.FindPeer(endPoint);
                 if (peerAddress != null && !this.IsPeerConnected(peerAddress.NetworkAddress.Endpoint))
                     return peerAddress.NetworkAddress;
-
-                continue;
             }
 
             return null;
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
@@ -54,15 +54,11 @@ namespace Stratis.Bitcoin.P2P
         {
             foreach (var endPoint in this.NodeSettings.ConnectionManager.AddNode)
             {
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -52,20 +52,20 @@ public PeerConnectorAddNode(NodeSettings nodeSettings, IPeerAddressManager peerA
         /// <inheritdoc/>
         public override NetworkAddress FindPeerToConnectTo()
         {
-            PeerAddress peer = null;
-
             foreach (var endPoint in this.NodeSettings.ConnectionManager.AddNode)
             {
-                peer = this.peerAddressManager.FindPeer(endPoint.MapToIpv6());
+                PeerAddress peer = this.peerAddressManager.FindPeer(endPoint.MapToIpv6());
```
bokobza: You can create a variable to hold the value of `endPoint.MapToIpv6()`.

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -52,20 +52,20 @@ public PeerConnectorAddNode(NodeSettings nodeSettings, IPeerAddressManager peerA
...
                 if (peer == null)
                     continue;
 
                 // If the peer is already connected just continue.
-                if (this.IsPeerConnected(endPoint))
+                if (this.IsPeerConnected(endPoint.MapToIpv6()))
```
bokobza: Instead of having 2 `continue;` here, if there's a way to get the value of `endPoint.MapToIpv6()` from `peer`, it would be clearer to have a single statement such as 
```
if(peer == null || this.IsPeerConnected(peer.NetworkAddress))
    continue;

return peer.NetworkAddress;
```

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -52,20 +52,20 @@ public PeerConnectorAddNode(NodeSettings nodeSettings, IPeerAddressManager peerA
...
                     continue;
+
+                return peer.NetworkAddress;
             }
 
-            return peer.NetworkAddress;
+            return null;
```
dangershony: Do we handle null cases in the caller?

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
dangershony: Can't you just do `if(peer == null || this.IsPeerConnected(endPoint.MapToIpv6()))` in that case?
But to me this looks fine.

If any refactoring I would suggest to put all inside if statements 
```
if(peer != null)
  if([connected])
     return peer
```

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: OK yes, I have moved MapToIpv6() to the Match method so that it only happens in place :)

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: See the new code changes ;)

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: Yes thats correct :)

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -51,20 +51,16 @@ public PeerConnectorConnectNode(NodeSettings nodeSettings, IPeerAddressManager p
...
-                if (peer == null)
-                    continue;
+                PeerAddress peerAddress = this.peerAddressManager.FindPeer(endPoint);
+                if (peerAddress != null && !this.IsPeerConnected(peerAddress.NetworkAddress.Endpoint))
+                    return peerAddress.NetworkAddress;
 
-                // If the peer is already connected just continue.
-                if (this.IsPeerConnected(endPoint))
-                    continue;
+                continue;
```
dangershony: yeah ok `continue` is redundant then ;)

fassadlr: LOL I deserve a fine here ;)

dangershony: You will pay with a 🍺 and then with some more 🍻 after that.

Issue #659, opened 2017-10-29
[NBitcoin] Fix problematic chain functions


Aprogiena:
There are couple of wrongly implemented functions that we should aim to fix in NBitcoin.

**1) ChainBase.Contains(ChainedBlock blockIndex)**

This method is implemented as follows:
```
		public bool Contains(ChainedBlock blockIndex)
		{
			if(blockIndex == null)
				throw new ArgumentNullException("blockIndex");
			return GetBlock(blockIndex.Height) != null;
		}

```

it is thus terribly misleading to write something like `this.chain.Contains(myBlock)` because you would expect to get true if your block is on the best chain. Instead, you receive true if the best chain contains a block at the height of your block. 

**2) ChainedBlock.GetAncestor**
This is implemented as 
```
		public ChainedBlock GetAncestor(int height)
		{
			if(height > Height || height < 0)
				return null;
			ChainedBlock current = this;

			while(true)
			{
				if(current.Height == height)
					return current;
				current = current.Previous;
			}
		}
```

which is O(N) time complexity function. Bitcoin Core implements skip list to make it O(log N), which is the correct way to do it.

Similarly, `ChainedBlock.FindFork`, `ChainedBlock.FindAncestorOrSelf`, `ChainBase.FindFork(ChainBase chain)` suffer from the same problem. And all the code that is using some kind of `this.Chain.Enumerate*` should be reviewed as well.

**3) `ChainedBlock.GetLocator()`**

Similarly as in 2), this function is O(N), while in Bitcoin Core it is O(log N).

dangershony:
Indeed I agree, my intention though is not to fix NBitcoin but to move over components form there to the fullnode and change them to better fit the fullnode requirements, trying to keep NBitcoin with little changes as possible to make it easy to merge from the upstream master Nicolas is maintaining.

@bokobza  did implement a search method O(Log N) I believe in the wallet, in any we definitely can  improve on that.

Aprogiena:
Of course, that is also an option. The point was to get this fixed, no matter how :-)

As for the better implementation in the wallet, if it is the one that relies on GetLocator, then it is better than without it, but the initialization is still O(N).

dangershony:
Just sharing here the place where bitcoin does O(log N)
https://github.com/bitcoin/bitcoin/blob/master/src/chain.cpp#L83

Aprogiena:
[...]

PR #847, opened 2017-12-02
Comments and styling for ChainBase

mikedennis:
Also removed a confusing implementation of Contains which was not referenced anywhere. See #659 for details.

### DIFF
--- a/src/NBitcoin/ChainBase.cs
+++ b/src/NBitcoin/ChainBase.cs
@@ -11,14 +11,14 @@ namespace NBitcoin
         /// Gets the chained block header given a block ID (hash).
         /// </summary>
         /// <param name="id">Block ID to retreive.</param>
-        /// <returns>The block.</returns>
+        /// <returns>The chained block header.</returns>
         public abstract ChainedBlock GetBlock(uint256 id);
 
         /// <summary>
         /// Gets the chained block header at a given block height.
         /// </summary>
         /// <param name="height">Height to retrieve chained block header at.</param>
-        /// <returns>The block.</returns>
+        /// <returns>The chained block header.</returns>
         public abstract ChainedBlock GetBlock(int height);
 
         /// <summary>Gets the chained block header at the tip of the chain.</summary>
@@ -28,7 +28,7 @@ namespace NBitcoin
         public abstract int Height { get; }
 
         /// <summary>
-        /// Enumerates from start of the chain.
+        /// Enumerates chained block headers from start of the chain.
         /// </summary>
         /// <returns>An enumerable iterator.</returns>
         protected abstract IEnumerable<ChainedBlock> EnumerateFromStart();
@@ -41,7 +41,7 @@ namespace NBitcoin
         public abstract ChainedBlock SetTip(ChainedBlock chainedBlock);
 
         /// <summary>Gets the genesis block for the chain.</summary>
-        public virtual ChainedBlock Genesis { get { return GetBlock(0); } }
+        public virtual ChainedBlock Genesis { get { return this.GetBlock(0); } }
 
         /// <summary>
         /// Gets an enumerable iterator for the chain.
@@ -91,7 +91,9 @@ namespace NBitcoin
         /// </summary>
         /// <param name="header">The block header to set to tip.</param>
         /// <param name="chainedHeader">The newly chained block header for the tip.</param>
-        /// <returns>Whether the tip was set successfully. Will return <c>false</c> if the block header previous block link cannot be found in the current chain.</returns>
+        /// <returns>Whether the tip was set successfully. The method fails (and returns <c>false</c>)
+        /// if the <paramref name="header"/>'s link to a previous header does not point to any block
+        /// in the current chain.</returns>
         public bool TrySetTip(BlockHeader header, out ChainedBlock chainedHeader)
         {
             if (header == null)
@@ -159,7 +161,7 @@ namespace NBitcoin
         /// Returns the first chained block header that exists in the chain from the list of block hashes.
         /// </summary>
         /// <param name="hashes">Hash to search for.</param>
-        /// <returns>First found chained block header or <c>null</c>.</returns>
+        /// <returns>First found chained block header or <c>null</c> if not found.</returns>
         public ChainedBlock FindFork(IEnumerable<uint256> hashes)
         {
             if (hashes == null)
@@ -190,7 +192,7 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Enumerate chain after given block hash to genesis block.
+        /// Enumerate chain block headers after given block hash to genesis block.
         /// </summary>
         /// <param name="blockHash">Block hash to enumerate after.</param>
         /// <returns>Enumeration of chained block headers after given block hash.</returns>
@@ -205,9 +207,9 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Enumerates chain from the given chained block header to tip.
+        /// Enumerates chain block headers from the given chained block header to tip.
         /// </summary>
-        /// <param name="block">Chained block header to enumerate from</param>
+        /// <param name="block">Chained block header to enumerate from.</param>
         /// <returns>Enumeration of chained block headers from given chained block header to tip.</returns>
         public IEnumerable<ChainedBlock> EnumerateToTip(ChainedBlock block)
         {
@@ -218,7 +220,7 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Enumerates chain from given block hash to tip.
+        /// Enumerates chain block headers from given block hash to tip.
         /// </summary>
         /// <param name="blockHash">Block hash to enumerate from.</param>
         /// <returns>Enumeration of chained block headers from the given block hash to tip.</returns>
@@ -235,7 +237,7 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Enumerates chain after the given chained block header to genesis block.
+        /// Enumerates chain block headers after the given chained block header to genesis block.
         /// </summary>
         /// <param name="block">The chained block header to enumerate after.</param>
         /// <returns>Enumeration of chained block headers after the given block.</returns>
--- a/src/NBitcoin/ChainBase.cs
+++ b/src/NBitcoin/ChainBase.cs
@@ -8,20 +8,20 @@ namespace NBitcoin
     public abstract class ChainBase
     {
         /// <summary>
-        /// Gets the block given a block id(hash).
+        /// Gets the chained block header given a block ID (hash).
         /// </summary>
-        /// <param name="id">Block id to retreive.</param>
+        /// <param name="id">Block ID to retreive.</param>
         /// <returns>The block.</returns>
         public abstract ChainedBlock GetBlock(uint256 id);
 
         /// <summary>
-        /// Gets the block at a given block height.
+        /// Gets the chained block header at a given block height.
         /// </summary>
-        /// <param name="height">Height to retreive block at.</param>
+        /// <param name="height">Height to retrieve chained block header at.</param>
         /// <returns>The block.</returns>
         public abstract ChainedBlock GetBlock(int height);
 
-        /// <summary>Gets the block at the tip of the chain.</summary>
+        /// <summary>Gets the chained block header at the tip of the chain.</summary>
         public abstract ChainedBlock Tip { get; }
 
         /// <summary>Gets the height of the chain.</summary>
@@ -90,8 +90,8 @@ namespace NBitcoin
         /// Attempts to set the tip of this chain based upon another block header.
         /// </summary>
         /// <param name="header">The block header to set to tip.</param>
-        /// <param name="chainedHeader">The newly chained block for the tip.</param>
-        /// <returns>Whether the tip was set successfully.</returns>
+        /// <param name="chainedHeader">The newly chained block header for the tip.</param>
+        /// <returns>Whether the tip was set successfully. Will return <c>false</c> if the block header previous block link cannot be found in the current chain.</returns>
         public bool TrySetTip(BlockHeader header, out ChainedBlock chainedHeader)
         {
             if (header == null)
@@ -108,13 +108,13 @@ namespace NBitcoin
         }
 
         /// <summary>
-        /// Whether the chain contains a block with the given hash.
+        /// Whether the chain contains a chained block header with the given hash.
         /// </summary>
         /// <param name="hash">The hash to search for.</param>
-        /// <returns>Whether the chain contains the block.</returns>
+        /// <returns>Whether the chain contains the chained block header.</returns>
         public bool Contains(uint256 hash)
         {
-            ChainedBlock block = GetBlock(hash);
+            ChainedBlock block = this.GetBlock(hash);
             return block != null;
         }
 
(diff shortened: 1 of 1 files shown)


src/NBitcoin/ChainBase.cs
```diff
@@ -4,15 +4,33 @@
 
 namespace NBitcoin
 {
+    /// <summary>Base implementation for chains.</summary>
     public abstract class ChainBase
     {
+        /// <summary>
+        /// Gets the block given a block id(hash).
```
Aprogiena: i'd wrote ID as with capital I and D, I'd also put space between ID and `(`

but most importantly, here and in whole file, I think we should distinguish between Block and ChainedBlock
as Block is what represents and actual block, but ChainedBlock is imho misnamed, it is actually ChainedBlockHeader, or ChainedHeader, it is not a block 

So if we want to to keep this little confusing name ChainedBlock, we should at least make comments that make it clear that this is not a block. GetBlock is again misnamed method, it does not return a block

GetBlock should be named GetChainedBlock (optimally GetChainedBlockHeader ...) 

So I was wondering if we could at least start with comments being correct

src/NBitcoin/ChainBase.cs
```diff
@@ -4,15 +4,33 @@
...
+        /// </summary>
+        /// <param name="id">Block id to retreive.</param>
+        /// <returns>The block.</returns>
         public abstract ChainedBlock GetBlock(uint256 id);
+
+        /// <summary>
+        /// Gets the block at a given block height.
+        /// </summary>
+        /// <param name="height">Height to retreive block at.</param>
```
Aprogiena: typo `ei`, i'd rephrase `Height of the block which header to retrieve.`

src/NBitcoin/ChainBase.cs
```diff
@@ -4,15 +4,33 @@
...
         public abstract ChainedBlock Tip { get; }
 
+        /// <summary>Gets the height of the chain.</summary>
         public abstract int Height { get; }
 
+        /// <summary>
+        /// Enumerates from start of the chain.
+        /// </summary>
+        /// <returns>An enumerable iterator.</returns>
         protected abstract IEnumerable<ChainedBlock> EnumerateFromStart();
```
Aprogiena: this is not for this PR, just an idea - could we get rid of this method completely? just an area to explore

src/NBitcoin/ChainBase.cs
```diff
@@ -52,41 +75,54 @@ public ChainedBlock SetTip(ChainBase otherChain)
...
             ChainedBlock chainedHeader;
             return this.TrySetTip(header, out chainedHeader);
         }
 
+        /// <summary>
+        /// Attempts to set the tip of this chain based upon another block header.
+        /// </summary>
+        /// <param name="header">The block header to set to tip.</param>
+        /// <param name="chainedHeader">The newly chained block for the tip.</param>
+        /// <returns>Whether the tip was set successfully.</returns>
```
Aprogiena: it would be useful to know when such an operation can fail

src/NBitcoin/ChainBase.cs
```diff
@@ -52,41 +75,54 @@ public ChainedBlock SetTip(ChainBase otherChain)
...
         }
 
+        /// <summary>
+        /// Whether the chain contains a block with the given hash.
+        /// </summary>
+        /// <param name="hash">The hash to search for.</param>
+        /// <returns>Whether the chain contains the block.</returns>
         public bool Contains(uint256 hash)
         {
             ChainedBlock block = GetBlock(hash);
```
Aprogiena: `this.` missing?

src/NBitcoin/ChainBase.cs
```diff
@@ -52,41 +75,54 @@ public ChainedBlock SetTip(ChainBase otherChain)
...
-                throw new ArgumentNullException("blockIndex");
-
-            return this.GetBlock(blockIndex.Height) != null;
-        }
-
+        /// <summary>
+        /// Whether the given chain and this chain have the same tip.
+        /// </summary>
+        /// <param name="chain">The chain to compare this chain against.</param>
+        /// <returns>Whether the tips of the chain have the same block hash.</returns>
```
Aprogiena: chain`s`

src/NBitcoin/ChainBase.cs
```diff
@@ -95,11 +131,23 @@ public bool SameTip(ChainBase chain)
             return this.Tip.HashBlock == chain.Tip.HashBlock;
         }
 
+        /// <summary>
+        /// Gets the target work required for a network and a specific block height.
+        /// </summary>
+        /// <param name="network">The network to use for work computation.</param>
+        /// <param name="height">The height of the block to get work for.</param>
+        /// <returns>The target work.</returns>
```
Aprogiena: `required work` here might be little better than `target work`, or `target required work`

src/NBitcoin/ChainBase.cs
```diff
@@ -95,11 +131,23 @@ public bool SameTip(ChainBase chain)
...
             return this.GetBlock(height).GetWorkRequired(network);
         }
 
+        /// <summary>
+        /// Validates the chain on a given network.
+        /// </summary>
+        /// <param name="network">Network to validate the chain on.</param>
+        /// <param name="fullChain">Whether to validate full chain (all blocks) or just the tip.</param>
+        /// <returns>Whether the chain is valid or not.</returns>
         public bool Validate(Network network, bool fullChain = true)
```
Aprogiena: is this method used anywhere?
it seems like a heavy one, should not be used that much (at all?)

src/NBitcoin/ChainBase.cs
```diff
@@ -153,6 +200,11 @@ public ChainedBlock FindFork(IEnumerable<uint256> hashes)
             return null;
         }
 
+        /// <summary>
```
Aprogiena: there is another version of FindFork untouched by this PR, but its comment makes no sense and need rephrasing, its summary reads as `Returns the first found block.` which is confusing

src/NBitcoin/ChainBase.cs
```diff
@@ -161,6 +213,11 @@ public ChainedBlock FindFork(BlockLocator locator)
             return this.FindFork(locator.Blocks);
         }
 
+        /// <summary>
+        /// Enumerate chain after given block hash to genesis block.
+        /// </summary>
+        /// <param name="blockHash">Block hash to enumerate after.</param>
+        /// <returns>Enumeration of chain after given block hash.</returns>
         public IEnumerable<ChainedBlock> EnumerateAfter(uint256 blockHash)
```
Aprogiena: as mentioned before, these enumeration methods, if they could be checked whether they are used and whether their use is proper - they can be pretty heavy, so we should be careful

not for this PR of course

src/NBitcoin/ChainBase.cs
```diff
@@ -171,6 +228,11 @@ public IEnumerable<ChainedBlock> EnumerateAfter(uint256 blockHash)
             return this.EnumerateAfter(block);
         }
 
+        /// <summary>
+        /// Enumerates chain from the given block to tip.
+        /// </summary>
+        /// <param name="block">Block to enumerate from</param>
+        /// <returns>Enumeration of chain from given block to tip.</returns>
         public IEnumerable<ChainedBlock> EnumerateToTip(ChainedBlock block)
```
Aprogiena: "Enumeration of chain" sounds confusing to me (not only this method)
i'd be more happy with
"Enumeration of chained block headers from given block to the tip."

however, these methods seem extremely dangerous
i can see a lot of them being used in tests, in which case they should be migrated to tests where it is fine to have dangerous methods
but otherwise we should refactor if any other code is using those

again, not for this PR

src/NBitcoin/ChainBase.cs
dangershony: I would be up to renaming to `ChainedBlockHeader` even though ChainedHeader will also do the job (we don't have headers of anything else really).

mikedennis: Yes I agree confusion between Blocks and Chained Block Headers. I will change all the comments for now and we can decide what to do about ChainedBlock, ChainBase, ConcurrentChain longer term.

src/NBitcoin/ChainBase.cs
mikedennis: Sounds good thanks.

src/NBitcoin/ChainBase.cs
mikedennis: Ok sounds good.

src/NBitcoin/ChainBase.cs
mikedennis: Yes, good catch!

src/NBitcoin/ChainBase.cs
mikedennis: Sounds good.

src/NBitcoin/ChainBase.cs
mikedennis: Yes I agree. I struggled with what to comment this as, I like your suggestion.

src/NBitcoin/ChainBase.cs
mikedennis: Agreed. I find the ToEnumerable(fromTip) method confusing as well. I think it might be better to break into two separate methods rather than pass in a boolean.

src/NBitcoin/ChainBase.cs
mikedennis: No references to it. I will delete it.

src/NBitcoin/ChainBase.cs
mikedennis: I reworked this comment.

src/NBitcoin/ChainBase.cs
```diff
@@ -4,15 +4,33 @@
...
 namespace NBitcoin
 {
+    /// <summary>Base implementation for chains.</summary>
     public abstract class ChainBase
     {
+        /// <summary>
+        /// Gets the chained block header given a block ID (hash).
+        /// </summary>
+        /// <param name="id">Block ID to retreive.</param>
+        /// <returns>The block.</returns>
```
Aprogiena: good for changing the comments to `chained block headers`, but you've forgotten some returns

src/NBitcoin/ChainBase.cs
```diff
@@ -4,15 +4,33 @@
...
         public abstract ChainedBlock GetBlock(int height);
 
+        /// <summary>Gets the chained block header at the tip of the chain.</summary>
         public abstract ChainedBlock Tip { get; }
 
+        /// <summary>Gets the height of the chain.</summary>
         public abstract int Height { get; }
 
+        /// <summary>
+        /// Enumerates from start of the chain.
```
Aprogiena: Enumerates chained block headers from the start of the chain.

src/NBitcoin/ChainBase.cs
```diff
@@ -22,14 +40,14 @@ public abstract class ChainBase
...
 
-        public virtual ChainedBlock Genesis
-        {
-            get
-            {
-                return GetBlock(0);
-            }
-        }
+        /// <summary>Gets the genesis block for the chain.</summary>
+        public virtual ChainedBlock Genesis { get { return GetBlock(0); } }
```
Aprogiena: `this.` missing and we can totally add it in this PR

src/NBitcoin/ChainBase.cs
dangershony: It might also be possible to implement the IEnumerator on chained block itself, but only if it will be really beneficial.

src/NBitcoin/ChainBase.cs
```diff
@@ -52,41 +75,54 @@ public ChainedBlock SetTip(ChainBase otherChain)
...
             ChainedBlock chainedHeader;
             return this.TrySetTip(header, out chainedHeader);
         }
 
+        /// <summary>
+        /// Attempts to set the tip of this chain based upon another block header.
+        /// </summary>
+        /// <param name="header">The block header to set to tip.</param>
+        /// <param name="chainedHeader">The newly chained block header for the tip.</param>
+        /// <returns>Whether the tip was set successfully. Will return <c>false</c> if the block header previous block link cannot be found in the current chain.</returns>
```
Aprogiena: hard to read the second sentence :-)
how about
`<returns>Whether the tip was set successfully. The method fails (and returns <c>false</c>) if the <paramref name="header"/>'s link to a previous header does not point to any block in the current chain.</returns>`

src/NBitcoin/ChainBase.cs
```diff
@@ -161,6 +189,11 @@ public ChainedBlock FindFork(BlockLocator locator)
             return this.FindFork(locator.Blocks);
         }
 
+        /// <summary>
+        /// Enumerate chain after given block hash to genesis block.
+        /// </summary>
+        /// <param name="blockHash">Block hash to enumerate after.</param>
+        /// <returns>Enumeration of chained block headers after given block hash.</returns>
```
dangershony: This feels to me like it could be an extension method, perhaps it might be a good idea to see what other methods can be pulled as extensions (not this PR)

src/NBitcoin/ChainBase.cs
```diff
@@ -95,35 +131,22 @@ public bool SameTip(ChainBase chain)
...
-
-
         /// <summary>
-        /// Returns the first common block between two chains.
+        /// Returns the first common chained block header between two chains.
         /// </summary>
         /// <param name="chain">The other chain.</param>
-        /// <returns>First common block or <c>null</c>.</returns>
+        /// <returns>First common chained block header or <c>null</c>.</returns>
         public ChainedBlock FindFork(ChainBase chain)
```
Aprogiena: this method is implemented terribly, it goes through whole chain!? fortunately it is used only in tests - it should be moved to tests

please do it as a part of this PR because this method is terrible

src/NBitcoin/ChainBase.cs
```diff
@@ -133,10 +156,10 @@ public ChainedBlock FindFork(ChainBase chain)
         }
 
         /// <summary>
-        /// Returns the first found block.
+        /// Returns the first chained block header that exists in the chain from the list of block hashes.
         /// </summary>
         /// <param name="hashes">Hash to search for.</param>
-        /// <returns>First found block or <c>null</c>.</returns>
+        /// <returns>First found chained block header or <c>null</c>.</returns>
```
Aprogiena: or null ... when null? null if ...

src/NBitcoin/ChainBase.cs
```diff
@@ -179,6 +217,11 @@ public IEnumerable<ChainedBlock> EnumerateToTip(ChainedBlock block)
             return this.EnumerateToTip(block.HashBlock);
         }
 
+        /// <summary>
+        /// Enumerates chain from given block hash to tip.
```
Aprogiena: Enumerates chained block headers

src/NBitcoin/ChainBase.cs
```diff
@@ -171,6 +204,11 @@ public IEnumerable<ChainedBlock> EnumerateAfter(uint256 blockHash)
             return this.EnumerateAfter(block);
         }
 
+        /// <summary>
+        /// Enumerates chain from the given chained block header to tip.
+        /// </summary>
+        /// <param name="block">Chained block header to enumerate from</param>
```
Aprogiena: missing dot

src/NBitcoin/ChainBase.cs
```diff
@@ -171,6 +204,11 @@ public IEnumerable<ChainedBlock> EnumerateAfter(uint256 blockHash)
             return this.EnumerateAfter(block);
         }
 
+        /// <summary>
+        /// Enumerates chain from the given chained block header to tip.
```
Aprogiena: Enumerates chained block headers

src/NBitcoin/ChainBase.cs
```diff
@@ -191,6 +234,11 @@ public IEnumerable<ChainedBlock> EnumerateToTip(uint256 blockHash)
                 yield return chainedBlock;
         }
 
+        /// <summary>
+        /// Enumerates chain after the given chained block header to genesis block.
```
Aprogiena: chained block headers

PR #850, opened 2017-12-04
Fix typos in comments

youngmtool:


src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -247,7 +247,9 @@ public static IEnumerable<PeerAddress> Tried(this ConcurrentDictionary<IPEndPoin
             return tried;
         }
 
-        /// <summary>Return a random peer from a given set of peers.</summary>
+        /// <summary>
+        /// Return a random peer from a given set of peers.
+        /// </summary>
```
fassadlr: Please note that its OK to have <summary> comment blocks on one line if they are short enough ;)

PR #853, opened 2017-12-04
Refactor/the node job part 9

Aprogiena:
Note that first commit of this part is `replace filters based approach with event based`, previous commits are of previous parts.

This part includes:
- more node->peer renaming
- removing filters from the design of network peer (this required CHB refactoring, but also ONE UGLY HACK that can't be removed now, but is planned to be removed later - reasons are documented in TODO comment)
- minor polishing as always)


src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -131,154 +141,159 @@ protected override void DetachCore()
...
 
-            this.AttachedPeer.StateChanged -= this.AttachedNode_StateChanged;
+            this.AttachedPeer.MessageReceivedPriority -= this.AttachedPeer_MessageReceived;
+            this.AttachedPeer.StateChanged -= this.AttachedPeer_StateChanged;
 
             this.logger.LogTrace("(-)");
         }
 
-        private void Intercept(IncomingMessage message, Action continueInvocation)
+        private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage message)
```
fassadlr: I'm not a fan of such a big methods @Aprogiena. Can we not split the switch/case logic into separate methods here? Maybe we would want to write unit tests for the logic contained in the individual case statements? I would even split them off into classes but maybe that's a bit overkill. We need to try and stick to SOLID principles as much as we can and here we a method that is responsible for a lot of things.

WDYT?

Aprogiena: all this series of PRs is for network peer, it touched CHB here, so some work had to be done there, but further improvements of CHB are out of the scope of this PR

so for this PR, nope

otherwise, i have no problem with creating methods for each switch case as those are non trivial code blocks, so i agree. not into separate classes though, that's too much for me

but totally offtopic :)

fassadlr: OK yeah, lets put those into methods, then at least we are one step closer to making the code more modular and testable 👍

Aprogiena: yes, but pls create a separate issue for that

fassadlr: Is it going to be a lot of work to move the logic into methods?

Aprogiena: i don't know, it just does not belong here :-)
it is probably task of size 1, but still a separated task

fassadlr: heheh ok cool man! Or maybe just add a TODO: so that we know to improve that area in another PR.

Aprogiena: ok i creted #856

bokobza:
I'll have a look here as well.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -131,154 +141,159 @@ protected override void DetachCore()
...
         {
-            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(message), message.Message.Command, nameof(this.AttachedPeer), this.AttachedPeer?.RemoteSocketEndpoint);
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(peer), peer.RemoteSocketEndpoint, nameof(message), message.Message.Command);
 
-            var inv = message.Message.Payload as InvPayload;
-            if (inv != null)
+            switch (message.Message.Payload)
```
dangershony: Ahh I didn't know you could switch case on type 😱

dangershony:
Ah I approved this but @Aprogiena did you want to act on @bokobza comments here?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: ;)

Aprogiena:
I did in the future PR ...

Aprogiena:
tx

PR #854, opened 2017-12-04
SendTransactionAsync refactoring

noescape00:
Refactoring of classes that are used for broadcasting and validating transaction. 

This PR can be reviewed on commit-by-commit basis.

This PR is almost identical to closed PR 844 (except this one is separated by commits so it's much easier to review).


src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
```diff
@@ -13,13 +13,9 @@ public class FullNodeBroadcasterManager : BroadcasterManagerBase
...
 
-        /// <summary>Connection manager for managing node connections.</summary>
-        private readonly IConnectionManager connectionManager;
-
-        public FullNodeBroadcasterManager(IConnectionManager connectionManager, IMempoolValidator mempoolValidator) : base()
+        public FullNodeBroadcasterManager(IConnectionManager connectionManager, IMempoolValidator mempoolValidator) : base(connectionManager)
         {
-            this.connectionManager = connectionManager ?? throw new ArgumentNullException(nameof(connectionManager));
-            this.mempoolValidator = mempoolValidator ?? throw new ArgumentNullException(nameof(mempoolValidator));
+            this.mempoolValidator = mempoolValidator;
```
bokobza: You removed the check to check that `mempoolValidator` is not null.
Please reinstate this check by using `Guard.NotNull` instead.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -11,25 +12,27 @@ public abstract class BroadcasterManagerBase : IBroadcasterManager
     {
         public event EventHandler<TransactionBroadcastEntry> TransactionStateChanged;
 
+        /// <summary>Connection manager for managing node connections.</summary>
+        protected readonly IConnectionManager connectionManager;
+
+        public BroadcasterManagerBase(IConnectionManager connectionManager)
+        {
+            this.connectionManager = connectionManager;
```
bokobza: You moved the `mempoolValidator` here but the check for null has disappeared.
Please reinstate this check by using `Guard.NotNull`.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -58,5 +60,25 @@ public void AddOrUpdate(Transaction transaction, State state)
         }
 
         public abstract Task<Success> TryBroadcastAsync(Transaction transaction);
+
+        protected void PropagateTransactionToPeers(Transaction transaction, bool skipHalfOfThePeers = false)
```
bokobza: Can you please add some documentation on these methods?

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
```diff
@@ -21,21 +21,11 @@ public FullNodeBroadcasterManager(IConnectionManager connectionManager, IMempool
         /// <inheritdoc />
         public override async Task<Success> TryBroadcastAsync(Transaction transaction)
         {
-            if (transaction == null) throw new ArgumentNullException(nameof(transaction));
+            if (transaction == null)
```
bokobza: Please replace with a `Guard.NotNull`.

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/LightWalletBroadcasterManager.cs
```diff
@@ -16,33 +16,14 @@ public LightWalletBroadcasterManager(IConnectionManager connectionManager) : bas
...
-                if (found.State == State.CantBroadcast)
-                {
-                    this.AddOrUpdate(transaction, State.ToBroadcast);
-                }
-            }
-            else
-            {
-                this.AddOrUpdate(transaction, State.ToBroadcast);
-            }
+            if (IsPropagated(transaction))
```
bokobza: Please add `this.` in front of `IsPropagated`.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
bokobza: The FullNodeBroadcasterManager is also broadcasting to half the peers, so calling this method with `skipHalfOfThePeers = false` is a change of functionality.
Both should call only broadcast to half the peers, for now, so we don't need the parameter.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -58,5 +60,25 @@ public void AddOrUpdate(Transaction transaction, State state)
...
+        {
+            this.AddOrUpdate(transaction, State.ToBroadcast);
+
+            var invPayload = new InvPayload(transaction);
+
+            var propagateTo = skipHalfOfThePeers
```
bokobza: Is it not easier to just look at the count of connected nodes and do a `for` loop that goes to i < nodes.length/2?

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -41,21 +41,21 @@ public TransactionBroadcastEntry GetTransaction(uint256 transactionHash)
...
             }
             else
             {
-                if (oldEntry.State != state)
-                {
-                    this.Broadcasts.TryRemove(oldEntry);
-                    this.Broadcasts.Add(newEntry);
-                    this.OnTransactionStateChanged(newEntry);
-                }
+                var oldState = broadcastEntry.State;
```
bokobza: A bit confusing. How about:
 
```
if (broadcastEntry.State != state)
{
    broadcastEntry.State = state;
    this.OnTransactionStateChanged(broadcastEntry);
}
```

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -604,32 +604,15 @@ public IActionResult BuildTransaction([FromBody] BuildTransactionRequest request
...
-                        // if broadcasts doesn't contain then success
-                        var transactionEntry = this.broadcasterManager.GetTransaction(transaction.GetHash());
-                        if (transactionEntry != null && transactionEntry.State == Bitcoin.Broadcasting.State.Propagated)
-                        {
-                            return this.Json(model);
-                        }
-                        await Task.Delay(period).ConfigureAwait(false);
-                        waited += period;
-                    }
+                    throw new Exception("Transaction was not validated by mempool. Can't send the transaction.");
```
bokobza: There isn't necessarily a mempool, for example when using the light wallet.

bokobza:
Good stuff.
Commit adf6216 does change some functionality and it's not a simple refactoring woth 0 changes, but I'll take it for now. Please make this a priority in your next change.

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
noescape00: Ok. 
The reason why I did that is because we almost never have those on constructors that get created through DI. 
But I guess since it's supposed to be a refactoring the check should be there.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
noescape00: LightWalletBroadcasterManager returns true or throws. So in case of this exception it will always be a full node.

src/Stratis.Bitcoin/Broadcasting/BroadcasterBehavior.cs
```diff
@@ -13,30 +13,30 @@ namespace Stratis.Bitcoin.Broadcasting
 {
     public class BroadcasterBehavior : NetworkPeerBehavior
     {
-        protected readonly IBroadcasterManager manager;
+        protected readonly IBroadcasterManager BroadcasterManager;
```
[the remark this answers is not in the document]
noescape00: No, they don't, you're right.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
noescape00: >Both should call only broadcast to half the peers, for now, so we don't need the parameter.

In current implementation of broadcasting algorithm for light node it is used (on light node we send tx to half of the peers and expect other half to propagate it back to us. that's how we know that our tx is valid and was propagated).

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
noescape00: Yeah. It's unnecessary over-complicated. Fixing.

fassadlr:
@noescape00 we have this test failing in both release and debug builds?

`SendTransactionSuccessfulReturnsWalletSendTransactionModelResponse`

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza: And here you're assuming you know exactly what's happening inside of TryBroadcastAsync. You shouldn't. Someone could come and change the LightWalletBroadcasterManager and return false (after all it returns a bool so it's perfectly allowed) and your code suddenly wouldn't make any sense.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
bokobza: I can see that in the current implementation, the FullNodeBroadcasterManager is also broadcasting to half the peers. Are yo unot seeing the same thing?

noescape00:
>@noescape00 we have this test failing in both release and debug builds?
@fassadlr you're right. This test is failing because of this PR. Thanks, I didn't notice that because a lot of tests are failing on master. 

So I've investigated it a bit. There is a problem with a mock of IBroadcasterManager that should return true but it returns false because transaction object that was used to setup mock is different from the actual object that is used there (they have the same hashes, just different objects in memory). 

So to fix this test I can add an overload for SendTransactionAsync that accepts transaction instead of hash so here we will be avoiding creating new objects and the mock will work as planned. 

But effectively I think this test should not exist since it tests a function that would call several objects but we mock that call to be always correct. So instead of testing the logic we are testing try/catch statement inside the SendTransactionAsync.

noescape00:
@bokobza 
Ok, working on it.

src/Stratis.Bitcoin/Broadcasting/BroadcasterManagerBase.cs
noescape00: Ah, yes. 
I thought that it was a logical error and writer did not mean to do that. 
I'm removing my change because it's refactoring. Maybe I should introduce it in later PRs

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
```diff
@@ -37,7 +37,7 @@ public FullNodeBroadcasterManager(IConnectionManager connectionManager, IMempool
                 return false;
             }
 
-            this.PropagateTransactionToPeers(transaction);
+            this.PropagateTransactionToPeers(transaction, true);
```
bokobza: I had suggested removing the boolean because it's going to be always true, but I'll leave this in for now.

bokobza:
Good stuff @noescape00, let's wait for the build and then it's going in.

noescape00:
>Good stuff @noescape00, let's wait for the build and then it's going in.

Thanks. 
Awesome =)

bokobza:
Thank **you** @noescape00 .

PR #857, opened 2017-12-05
[RPC] Fix addnode call with new refactored p2p code

fassadlr:


bokobza:
Can you tell us what the problem was and how you fixed it?

fassadlr:
Oh sorry @bokobza just realized I didn't provide detail.

OK so with the peer connector refactoring I didn't take into account the addnode RPC call which is supposed to add to the ConnectionManager's AddNode collection.

This PR fixes this issue.

I hit a few snags along the way especially with how the peer connectors get constructed in the connection manager. We need to instantiate them earlier so that access to them exists **without** having to call the Start method (which instantiates them). This seemed counter intuitive as there should always be instances based off how the node was configured on start up.

BUT as the connection manager's Start is only called later in the base feature, we don't yet have all the template behaviors in the behavior collection, so I cant construct the connectors with the parameters. I needed to change the StartConnecting method of the peer connectors to take in the parameters as well as the peer address behavior mode.

This will change with the next PR where we are going to inject the connection manager into the peer connector. The parameters can then be cloned inside them. (@dangershony knows)

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -424,4 +437,4 @@ public NetworkPeer Connect(IPEndPoint endpoint)
             return node;
         }
     }
-}
+}
```
bokobza: Did you remove the extra lines at the bottom?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -385,13 +390,21 @@ public NetworkPeer FindLocalNode()
...
-            this.AddNodePeerConnector.MaximumNodeConnections++;
+            this.peerAddressManager.AddPeer(new NetworkAddress(endPoint.MapToIpv6()), IPAddress.Loopback);
+
+            if (!this.NodeSettings.ConnectionManager.AddNode.Any(p => p.Match(endPoint)))
+            {
+                this.NodeSettings.ConnectionManager.AddNode.Add(endPoint);
+                this.AddNodePeerConnector.MaximumNodeConnections++;
+            }
+            else
+                this.logger.LogTrace("The end point already exists in the add node collection.");
```
bokobza: `endpoint` is one word.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -161,38 +181,23 @@ public void Start()
             clonedParameters.TemplateBehaviors.Add(new ConnectionManagerBehavior(false, this, this.loggerFactory));
 
             // Don't start peer discovery if we have specified any nodes using the -connect arg.
-            if (!this.NodeSettings.ConnectionManager.Connect.Any())
-            {
-                if (this.Parameters.PeerAddressManagerBehaviour().Mode.HasFlag(PeerAddressManagerBehaviourMode.Discover))
-                {
-                    this.logger.LogInformation("Starting peer discovery...");
+            //if (this.Parameters.PeerAddressManagerBehaviour().Mode.HasFlag(PeerAddressManagerBehaviourMode.Discover))
```
bokobza: Do you plan to reinstate that afterwards?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -148,6 +152,22 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
 
             this.Servers = new List<NetworkPeerServer>();
             this.connectedNodes = new NetworkPeerCollection();
+
+            // Don't start the peer discovery f we have specified any nodes using the -connect arg.
```
bokobza: This comment suggests that the next piece of code deals with having the `-connect` specified, but really it's with the `-connect` not specified.

bokobza: Also, there is a missing `i` at `if`.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -148,6 +152,22 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
+            {
+                this.DiscoverNodesPeerConnector = new PeerConnectorDiscovery(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
+            }
+            else
+            {
+                // Use if we have specified any nodes using the -connect arg
+                this.ConnectNodePeerConnector = new PeerConnectorConnectNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
+            }
+
+            {
```
bokobza: Are these curly braces intentional?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -161,38 +181,23 @@ public void Start()
...
-            if (!this.NodeSettings.ConnectionManager.Connect.Any())
-            {
-                if (this.Parameters.PeerAddressManagerBehaviour().Mode.HasFlag(PeerAddressManagerBehaviourMode.Discover))
-                {
-                    this.logger.LogInformation("Starting peer discovery...");
+            //if (this.Parameters.PeerAddressManagerBehaviour().Mode.HasFlag(PeerAddressManagerBehaviourMode.Discover))
+            //{
+            //    this.logger.LogInformation("Starting peer discovery...");
 
-                    this.peerDiscoveryLoop = new PeerDiscoveryLoop(this.asyncLoopFactory, this.Network, clonedParameters, this.nodeLifetime, this.peerAddressManager, this.networkPeerFactory);
```
bokobza: This seems to have disappeared, is it intentional?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Hey mate, I think you reviewed a earlier commit here?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Same here :(

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Same here :(

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Same here :(

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -385,13 +390,21 @@ public NetworkPeer FindLocalNode()
             return this.connectedNodes.FindLocal();
         }
 
-        public void AddNodeAddress(IPEndPoint endpoint)
+        public void AddNodeAddress(IPEndPoint endPoint)
```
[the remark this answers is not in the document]
fassadlr: Have fixed

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Fixed

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: I think this was again a review comment on an earlier commit... there are no extra lines at the bottom ;)

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -62,7 +62,7 @@ public interface IPeerConnector : IDisposable
         /// If the maximum amount of connections has been reached (<see cref="MaximumNodeConnections"/>), the action gets skipped.
         /// </para>
         /// </summary>
-        void StartConnectAsync();
+        void StartConnectAsync(NetworkPeerConnectionParameters parameters, PeerAddressManagerBehaviourMode mode);
```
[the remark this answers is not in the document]

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
             this.Parameters = parameters;
             this.Parameters.ConnectCancellation = this.nodeLifetime.ApplicationStopping;
+            this.Parameters.UserAgent = $"{this.NodeSettings.Agent}:{this.GetVersion()}";
+            this.Parameters.Version = this.NodeSettings.ProtocolVersion;
 
-            this.loggerFactory = loggerFactory;
-            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
-
-            this.asyncLoopFactory = asyncLoopFactory;
+            // Don't start the peer discovery connector if the -connect arg is specified at startup.
```
bokobza: shouldn't this be:
`// Start the peer discovery connector if the -connect arg is not specified at startup.` ?

dangershony:
Yes I will review it now

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
+            if (!this.NodeSettings.ConnectionManager.Connect.Any())
```
dangershony: Generally I would prefer to keep logic in the Initialize method.
But as this is a very simple logic then I will let you decide.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
+                this.DiscoverNodesPeerConnector = new PeerConnectorDiscovery(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
 
-            this.peerAddressManager = peerAddressManager;
-            this.networkPeerFactory = networkPeerFactory;
+            // Use if we have specified the -connect arg at startup.
+            if (this.NodeSettings.ConnectionManager.Connect.Any())
```
dangershony: Just our of curiosity why not use `else`

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
+                this.ConnectNodePeerConnector = new PeerConnectorConnectNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
 
-            this.Servers = new List<NetworkPeerServer>();
-            this.connectedNodes = new NetworkPeerCollection();
+            // Use if we have specified the -addnode arg at startup.
+            this.AddNodePeerConnector = new PeerConnectorAddNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
```
dangershony: It seems all the constructor params are registered in the DI, perhaps consider injecting the PeerConnector in a disabled state and use later logic to enable those we use.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
         }
 
         public void Start()
         {
             this.logger.LogTrace("()");
 
-            this.Parameters.UserAgent = $"{this.NodeSettings.Agent}:{this.GetVersion()}";
-            this.Parameters.Version = this.NodeSettings.ProtocolVersion;
-
+            // TODO: Move connection manager to the connector next
```
dangershony: Yes good idea also it seems `ConnectionManager` can be injected in to Connector.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -131,57 +135,51 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
             {
-                // Use if we have specified any nodes using the -connect arg
-                this.ConnectNodePeerConnector = new PeerConnectorConnectNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, clonedParameters, this.peerAddressManager);
-            }
+                this.logger.LogInformation("Starting peer discovery...");
 
-            {
-                // Use if we have specified any nodes using the -addnode arg
-                this.AddNodePeerConnector = new PeerConnectorAddNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, clonedParameters, this.peerAddressManager);
+                this.peerDiscoveryLoop = new PeerDiscoveryLoop(this.asyncLoopFactory, this.Network, clonedParameters, this.nodeLifetime, this.peerAddressManager, this.networkPeerFactory);
```
dangershony: I wonder if `PeerDiscoveryLoop` can be DI and here we just call `this.peerDiscoveryLoop.DiscoverPeers()`

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -385,23 +383,45 @@ public NetworkPeer FindLocalNode()
...
+        /// </summary>
+        /// <param name="ipEndpoint">The endpoint of the peer to disconnect.</param>
+        public void RemoveNodeAddress(IPEndPoint ipEndpoint)
         {
-            this.logger.LogTrace("({0}:'{1}')", nameof(endpoint), endpoint);
+            this.logger.LogTrace("({0}:'{1}')", nameof(ipEndpoint), ipEndpoint);
 
-            NetworkPeer node = this.connectedNodes.FindByEndpoint(endpoint);
-            node?.DisconnectWithException("Requested by user");
+            NetworkPeer peer = this.connectedNodes.FindByEndpoint(ipEndpoint);
```
dangershony: Hmm I wonder do we also need to delete it from the `NodeSettings.ConnectionManager.AddNode`

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -416,12 +436,12 @@ public NetworkPeer Connect(IPEndPoint endpoint)
                 OneTry = true
             });
 
-            NetworkPeer node = this.networkPeerFactory.CreateConnectedNetworkPeer(this.Network, endpoint, cloneParameters);
+            NetworkPeer peer = this.networkPeerFactory.CreateConnectedNetworkPeer(this.Network, endpoint, cloneParameters);
```
dangershony: I think there should already be a `ConnectionManagerBehavior`  in `this.Parameters` so we need to check this logic.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -173,14 +169,20 @@ internal bool IsPeerConnected(IPEndPoint endPoint)
         }
 
         /// <inheritdoc/>
-        public void StartConnectAsync()
+        public void StartConnectAsync(NetworkPeerConnectionParameters parameters, PeerAddressManagerBehaviourMode mode)
         {
+            this.parentParameters = parameters;
+
```
dangershony: I wonder if we want to move the 
```
clonedParameters.TemplateBehaviors.Add(new ConnectionManagerBehavior(false, this, this.loggerFactory));
```
From the ConnectionManager in to each connector.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: @bokobza I've fixed this comment but it appears now in another place.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: As discussed, peer connectors will now be DI-ed.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Will do in next PR 👍

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Lets look at doing this in next PR?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Hmm I'm not sure? Can you confirm somehow?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Yes this will be done in next PR 👍

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -129,70 +121,68 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
-                this.ConnectNodePeerConnector = new PeerConnectorConnectNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, clonedParameters, this.peerAddressManager);
+                this.peerDiscoveryLoop = new PeerDiscoveryLoop(this.asyncLoopFactory, this.Network, clonedParameters, this.nodeLifetime, this.peerAddressManager, this.networkPeerFactory);
+                this.peerDiscoveryLoop.DiscoverPeers();
             }
 
+            foreach (PeerConnector peerConnector in this.PeerConnectors)
             {
-                // Use if we have specified any nodes using the -addnode arg
-                this.AddNodePeerConnector = new PeerConnectorAddNode(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, clonedParameters, this.peerAddressManager);
+                peerConnector.Initialize(this.asyncLoopFactory, this.logger, this.Network, this.networkPeerFactory, this.nodeLifetime, this.NodeSettings, this.peerAddressManager);
```
dangershony: Ah I see so we have an `Initialize` and a `StartConnecting`
That's fine I, one thing though you could let the arguments in the initialise method be injected as part of the DI.
Do that in the next PR when we inject ConnectionManager.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -129,70 +121,68 @@ public IReadOnlyNetworkPeerCollection ConnectedNodes
...
+            relatedPeerConnectors.Register("Connect", peerConnectorConnectNode);
+            relatedPeerConnectors.Register("Discovery", peerConnectorDiscovery);
 
-            this.DiscoverNodesPeerConnector?.StartConnectAsync();
-            this.ConnectNodePeerConnector?.StartConnectAsync();
-            this.AddNodePeerConnector?.StartConnectAsync();
+            peerConnectorAddNode.StartConnectAsync(clonedParameters, PeerAddressManagerBehaviourMode.AdvertiseDiscover);
+
+            // Start the connect connector if the -connect arg is specified at startup, else start the discovery peer connector.
+            if (this.NodeSettings.ConnectionManager.Connect.Any())
```
dangershony: In the next PR I would suggest you move this in to each connector WSY?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -127,31 +134,30 @@ protected PeerConnector(NodeSettings nodeSettings, IPeerAddressManager peerAddre
...
-            ILogger logger,
-            Network network,
-            INetworkPeerFactory networkPeerFactory,
-            INodeLifetime nodeLifeTime,
-            NodeSettings nodeSettings,
-            NetworkPeerConnectionParameters parameters,
-            IPeerAddressManager peerAddressManager)
+        /// <summary>Peer connector initialization as called by the <see cref="Connection.ConnectionManager"/>.</summary>
+        public void Initialize(
+           IAsyncLoopFactory asyncLoopFactory,
```
dangershony: As I suggested this can be injected by the DI.

PR #858, opened 2017-12-05
Consensus rules execution

dangershony:
This is based on PR https://github.com/stratisproject/StratisBitcoinFullNode/pull/835

Some rules are still executed twice.
This PR only introduces the Rules to to allow state changing rules.
And expose a validate only method called by miners.


fassadlr:
@dangershony is this PR OK for me to make comments on the design? Or we do discuss this? In #124 ?

bokobza:
it looks to me as though these are the new commits?
image

dangershony:
Yes Jeremy is right only last 3 commits are new.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
```diff
@@ -28,9 +27,14 @@ public virtual IEnumerable<Type> Dependencies()
...
-        /// An indicator whether this rule is allowed to skip validation when the <see cref="BlockValidationContext.SkipValidation"/> is set to <c>true</c>.
+        /// Whether the rule will be considered a rule that only does validation and does not manipulate state in any way.
+        /// When <c>true</c> rule is allowed to skip validation when the <see cref="BlockValidationContext.SkipValidation"/> is set to <c>true</c>.
         /// </summary>
-        public virtual bool CanSkipValidation => false;
+        /// <remarks>
+        /// State in this context is the manipulation of information in the consensus data store based on actions specified <see cref="Block"/> and <see cref="Transaction"/>.
+        /// This will allow to ability to run validation checks on blocks (during mining for example) without change the underline store.
+        /// </remarks>
+        public virtual bool ValidationRule => true;
```
bokobza: Not very clear what this means.
Does it mean it's a rule that is subject to validation, or that it's a validating rule?
Maybe calling thia ValidationOnlyRule will make it clearer.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -181,4 +187,36 @@ public void SetErrorHandler(ConsensusRule errorHandler)
...
+
+    public static class ConsensusRulesExtension
+    {
+        /// <summary>
+        /// Try to find a consensus rule in the rules collection.
+        /// </summary>
+        /// <typeparam name="T">The type of rule to find.</typeparam>
+        /// <param name="rules">The rules to look in.</param>
+        /// <returns>The rule or <c>null</c> if not found in the list.</returns>
+        public static T TryFindRule<T>(this IEnumerable<ConsensusRule> rules) where T : ConsensusRule
```
bokobza: With TryXXX methods we're used to an out parameter containing the result value if successful and a bool return type.
Maybe this method could be changed to something reflecting this convention.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -181,4 +187,36 @@ public void SetErrorHandler(ConsensusRule errorHandler)
...
+        {
+            return rules.OfType<T>().FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Find a consensus rule in the rules collection or throw an exception of not found.
```
bokobza: of -> if

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -181,4 +187,36 @@ public void SetErrorHandler(ConsensusRule errorHandler)
...
+        /// </summary>
+        /// <typeparam name="T">The type of rule to find.</typeparam>
+        /// <param name="rules">The rules to look in.</param>
+        /// <returns>The rule or <c>null</c> if not found in the list.</returns>
+        public static T FindRule<T>(this IEnumerable<ConsensusRule> rules) where T : ConsensusRule
```
bokobza: Why throw an exception? I'd return null.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -181,4 +187,36 @@ public void SetErrorHandler(ConsensusRule errorHandler)
             }
         }
     }
+
+    public static class ConsensusRulesExtension
```
bokobza: Please put this in a separate class.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
dangershony: Yeah I did actually called it that `ValidationOnlyRule` before and then was not happy with that name. 
This means the rules only does validation.
Hmm perhaps that name is good, any other suggestions?

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
dangershony: That's not always correct the `TryXXX` name convention means this will not throw and not always has an out, but rather may return a null so the caller knows this method is will not fail if the entity is not found.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
dangershony: The is the complementary method to `TryXXX` this will throw if not found.

bokobza:
Good stuff mate

PR #860, opened 2017-12-05
Added a test in order to ensure that every serializable object has a parameterless constructor

noescape00:
This test is added in order to prevent scenarios when devs (like me) would delete a parameterless constructor that is never referenced in the codebase thinking that it's useless and then break the node by that.


Neurosploit:
This can be done in an easier way with a single loop. Saves having to maintain a list as well.
var type = typeof(IBitcoinSerializable);
var types = AppDomain.CurrentDomain.GetAssemblies()
    .SelectMany(s => s.GetTypes())
    .Where(p => type.IsAssignableFrom(p)  && !p.IsInterface);

noescape00:
@Neurosploit 
Yep. Spitted int two for easier debugging. 

But that's a good idea, I'll fix it.

PR #864, opened 2017-12-06
Refactor/the node job part 10

Aprogiena:
In this episode we have introduced TcpListener to the NetworkPeerServer. This replaced the socket approach.

These changes simplify the whole code quite a lot, this will be even more clear in part 11. Here we also improve usability of the class so that ConnectionManager can create the instances with correct endpoints with one call and does not need to set those things later. However, the overall design is still far from perfect, so this is just another midstep.

Please ignore `unfinishedTasks`, this is temporary thing, which is replaced with much more elegant design in part 11.


Aprogiena:
@bokobza if you are going to review this PR, please please only look here https://github.com/stratisproject/StratisBitcoinFullNode/pull/864/files for the final diff, and do not check commit-by-commit

bokobza:
Mine!

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -416,26 +388,25 @@ public void Dispose()
...
+                this.ConnectedNetworkPeers.DisconnectAll();
             }
+            catch (Exception e)
+            {
+                this.logger.LogTrace("Exception occurred: {0}", e.ToString());
+            }
+
+            this.tcpListener?.Stop();
+
+            this.logger.LogTrace("Waiting for {0} unfinished tasks to complete.", this.unfinishedTasks);
```
bokobza: Did you mean to show the number of tasks we're waiting on?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -147,94 +132,77 @@ public NetworkPeerServer(Network network, ProtocolVersion version, int internalP
...
+                this.logger.LogTrace("Exception occurred: {0}", e.ToString());
                 throw;
             }
 
             this.logger.LogTrace("(-)");
         }
 
         /// <summary>
-        /// Starts accepting connections from newly connected clients.
+        /// Implements loop accepting connections from newly connected clients.
```
bokobza: you're accepting connection from newly connected clients or from new clients wishing to connect?

Aprogiena:
Ah, sorry, sure, so the whole point in all the series is to refactor what was previously known as Node and related classes. This was done because
- it was old code style
- no logs
- no comments
- basically unknown architecture to us and quite a lot of bugs were closely related to his code, so we needed to better understand it
- use of socket events, which is unnecessary 
- questionable design decisions
- no interfaces - hard ability to test

So each part is doing a bit to fix one of those things and come to the minimalistic design that will fit our needs.

So specifically this PR is mostly about `use of socket events` in the listener and next PR does that for all peers. 

The problem is that doing such incremental changes in place (instead of starting from scratch) is quite difficult and sometimes temporary hacks help to keep the speed of development on reasonable level. So I need to ask for a little bit of patience with these PRs, there are going to be imperfect stuff here and there.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: yes, that was just debug thing for me, as mentioned, pls ignore unfinished tasks, they are removed i think in part 11

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: you are right, it is confusing 

will change to `Implements loop accepting new connections.`

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -147,94 +157,77 @@ public NetworkPeerServer(Network network, ProtocolVersion version, int internalP
...
-                this.BeginAccept();
+                this.tcpListener.Server.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
+                this.tcpListener.Start();
+                this.AcceptClientsAsync();
             }
-            catch (Exception ex)
+            catch (Exception e)
             {
-                this.logger.LogTrace("Exception occurred: {0}", ex.ToString());
+                this.logger.LogTrace("Exception occurred: {0}", e.ToString());
```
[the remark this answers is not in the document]
Aprogiena: AcceptClientAsync should not produce anything, so this is just for that starting server
and yes, it is quite common - e.g. if the port is used by other application, which i believe is then handled on higher level 

in general, in sockets world, exceptions are very common everywhere

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -147,94 +157,77 @@ public NetworkPeerServer(Network network, ProtocolVersion version, int internalP
         /// Starts listening on the server's initialialized endpoint.
         /// </summary>
         /// <param name="maxIncoming">Maximal number of newly connected clients waiting to be accepted.</param>
-        public void Listen(int maxIncoming = 8)
+        public void Listen(int maxIncoming = 20)
```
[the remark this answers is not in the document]
Aprogiena: good question - i actually don't know, just 8 seemed too low
this is a setting that tells you how many new connections can be queued, so basically if your server is not in trouble, you do not expect the queue to be longer than 1 because accepting new connection is super fast, but if you are doing some calculations, CPU is on 100 %, queue may happen, then if you have 8 pending connections, new clients would be told that they were not accepted. if we increase the number, we have some very little memory cost for that and the clients just wait and there is a good chance they will be accepted 
on high performance servers, this number is set to hundreds or more, but we are not looking for such a high performance, so we do not need 100 or 1000 here. 

but it does no harm to have 100 here, the memory cost is quite low, so therefore little higher value than needed is probably better than little lower and 8 seemed little lower, although in 99.9 % cases it would be well enough ...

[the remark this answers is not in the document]
bokobza: Nice, thanks.

PR #869, opened 2017-12-08
Api and RPC integration Test fixtures

Neurosploit:
I've created separate fixtures for the api and rpc integration tests. This will speed them up tremendously from 1-2 seconds per test to 10-300miliseconds per test. 
I've also merged two of the integration tests because one was just an additional action after the other. Integration tests are expensive and this saves another 4 seconds locally for me.


Neurosploit:
Let's see how the build servers like them.

bokobza:
@Neurosploit this test is flaky, I wouldn't pay attention. We still need to re-re-refix it though.

bokobza:
Very cool, love xunit fixtures :-)

PR #871, opened 2017-12-08
[PeerConector] Inject ConnectionManager into PeerConnector instances

fassadlr:
This PR is to refactor and cleanup connection manager so that it injects itself into the peer connector instances. This is all lead up work so that we can remove the RelatedPeerConnectors.


src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -166,33 +168,48 @@ public void AddPeer(NetworkPeer peer)
             this.ConnectedPeers.Add(peer);
         }
 
+        /// <summary>Determines whether or not a connector can be started.</summary>
+        internal abstract bool OnCanStartConnectAsync { get; }
```
dangershony: I would rename this `CanStartConnect` no need for async its not returning a task.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -166,33 +168,48 @@ public void AddPeer(NetworkPeer peer)
...
         {
-            this.parentParameters = parameters;
+            if (!this.OnCanStartConnectAsync)
+                return;
 
-            this.CurrentParameters = this.parentParameters.Clone();
-            this.CurrentParameters.ConnectCancellation = this.nodeLifetime.ApplicationStopping;
-            this.CurrentParameters.PeerAddressManagerBehaviour().Mode = mode;
-            this.CurrentParameters.TemplateBehaviors.Add(new PeerConnectorBehaviour(this));
+            this.OnStartConnectAsync();
```
dangershony: No need for async either.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -166,33 +168,48 @@ public void AddPeer(NetworkPeer peer)
             this.ConnectedPeers.Add(peer);
         }
 
+        /// <summary>Determines whether or not a connector can be started.</summary>
+        internal abstract bool OnCanStartConnectAsync { get; }
+
+        /// <summary>Specific peer connector initialization for each concrete implementation of this class.</summary>
+        internal abstract void OnInitialize();
```
dangershony: You could potentially achieve the same result by making `Initialize` itself virtual, just a suggestion.

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -41,6 +41,18 @@ public override void OnInitialize()
             }
         }
 
+        /// <summary>This connector is always started </summary>
+        internal override bool OnCanStartConnectAsync
```
dangershony: Making this method `internal` will prevent future implementation of a connector, why not use `protected`?

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
```diff
@@ -38,31 +54,40 @@ public sealed class PeerDiscoveryLoop : IDisposable
...
         private readonly INetworkPeerFactory networkPeerFactory;
 
-        public PeerDiscoveryLoop(
-            IAsyncLoopFactory asyncLoopFactory,
-            Network network,
-            NetworkPeerConnectionParameters connectionParameters,
-            INodeLifetime nodeLifetime,
-            IPeerAddressManager peerAddressManager,
-            INetworkPeerFactory networkPeerFactory)
+        public PeerDiscoveryLoop(IConnectionManager connectionManager)
```
dangershony: Any reason why not to use injection? then you can prevent adding all the properties on the IConnectionManager interface.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -43,51 +46,68 @@ public interface IConnectionManager : IDisposable
...
         string GetStats();
 
+        /// <summary>
+        /// Initializes peer connection by creating the peer connectors
+        /// as well as the peer discovery loop.
+        /// </summary>
+        void Initialize();
+
+        /// <summary>Logger factory to create loggers.</summary>
+        ILoggerFactory LoggerFactory { get; }
```
dangershony: This is a problematic approach, now the IConnectionManager has a dependency on ILoggerFactory, IAsyncLoopFactory,INodeLifetime  etc...

They should be hidden away from the interface.

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -308,10 +308,11 @@ public static IFullNodeBuilder UseBaseFeature(this IFullNodeBuilder fullNodeBuil
                     services.AddSingleton<IPeerConnector>(new PeerConnectorAddNode());
                     services.AddSingleton<IPeerConnector>(new PeerConnectorConnectNode());
                     services.AddSingleton<IPeerConnector>(new PeerConnectorDiscovery());
-                });
+                })
+                ;
```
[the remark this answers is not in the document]
fassadlr: Fixed

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: I have fixed these dependencies by DI. See the new commit 👍

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Renamed

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -134,30 +139,27 @@ protected PeerConnector(NodeSettings nodeSettings, IPeerAddressManager peerAddre
...
-        public void Initialize(
-           IAsyncLoopFactory asyncLoopFactory,
-           ILogger logger,
-           Network network,
-           INetworkPeerFactory networkPeerFactory,
-           INodeLifetime nodeLifeTime,
-           NodeSettings nodeSettings,
-           IPeerAddressManager peerAddressManager)
+        /// <inheritdoc/>
+        public void Initialize(IConnectionManager connectionManager)
```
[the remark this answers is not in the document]

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: Renamed

src/Stratis.Bitcoin/P2P/PeerConnector.cs
fassadlr: I want to force the concrete implementations to implement OnInitialize() so I made it abstract :)

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
fassadlr: We now use DI to construct PeerDiscovery and inject it into the Connection Manager

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
fassadlr: You are right, but they should actually be public, not protected as the concrete implementation implements the behaviour.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
dangershony: Ah yes valid point

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -140,21 +155,9 @@ protected PeerConnector(NodeSettings nodeSettings, IPeerAddressManager peerAddre
...
-            this.networkPeerFactory = connectionManager.NetworkPeerFactory;
-            this.nodeLifetime = connectionManager.NodeLifetime;
-            this.NodeSettings = connectionManager.NodeSettings;
-            this.peerAddressManager = connectionManager.PeerAddressManager;
-
-            this.parentParameters = connectionManager.Parameters;
-            this.CurrentParameters = this.parentParameters.Clone();
-            this.CurrentParameters.ConnectCancellation = this.nodeLifetime.ApplicationStopping;
-            this.CurrentParameters.TemplateBehaviors.Add(new ConnectionManagerBehavior(false, connectionManager, this.loggerFactory));
+            this.CurrentParameters = cloneParameters(parameters);
```
dangershony: This must be a nightmare to test.
I don't see the point of the delegate, the same delegate code is used on all connectors so you could just put the code here instead of calling a delegate. 

What more is there we already use inheritance so if anything the logic could go inside  `OnInitialize()` on each child.

dangershony: You could also just have that logic in an extension method

fassadlr: @dangershony Fair enough 👍 I wanted to get away from passing connection manager to the PeerConnector... I have realized now that the delegate is a bit over kill. I could have just cloned the parameters in a method and added the connection manager behavior to it. Then just pass those parameters to the connector.

PR #872, opened 2017-12-08
Refactor/the node job part 11

Aprogiena:
In this PR:
- we introduce NetworkPeerClient, which represents a client TCP connection
- NetworkPeer gets rid of socket events, which are replaced with NetworkPeerClient
- we upgrade our peer factory to be able to produce newly introduced things
- using the newly introduced client model, we can convert various method to true async code
- NetworkPeerServer is now fully aware of all its connected clients and we can thus wait for them to gracefully disconnect on shutdown and not to fly around when this part of the node is already disposed
 - we remove `unfinishedTasks` introduced in previous PR and replace it with Task-based approach that we do not need to actively wait for


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -166,16 +164,16 @@ public class NetworkPeerConnection
...
         /// </summary>
         /// <param name="peer">Network peer the node is connected to.</param>
-        /// <param name="socket">Connected network socket to the peer.</param>
+        /// <param name="client">Connected network client to the peer.</param>
         /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="loggerFactory">Factory for creating loggers.</param>
-        public NetworkPeerConnection(NetworkPeer peer, Socket socket, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
+        public NetworkPeerConnection(NetworkPeer peer, NetworkPeerClient client, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
         {
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName, $"[{peer.PeerAddress.Endpoint}] ");
```
mikedennis: Do we want to guard against peer or peer.PeerAddress being null?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <summary>Stream to send and receive messages through established TCP connection.</summary>
+        /// <remarks>Write operations on the stream have to be protected by <see cref="writeLock"/>.</remarks>
+        public NetworkStream Stream { get; private set; }
+
+        /// <summary>Task that cares about the client once its connection is accepted up until the communication is terminated.</summary>
+        /// <remarks>The client is being processed as long as this task is not complete.</remarks>
+        public TaskCompletionSource<bool> ProcessingCompletion { get; private set; }
+
+        /// <summary><c>1</c> if the instance of the object has been disposed or disposing is in progress, <c>0</c> otherwise.</summary>
+        private int disposed;
```
mikedennis: Can we use a bool instead here? IsDisposing - pretty sure this has already been mentioned and coming in part xx :grin:

Aprogiena: we can't, https://msdn.microsoft.com/en-us/library/system.threading.interlocked%28v=vs.110%29.aspx - Interlocked.CompareExchange is not available for bool types, we are currently using this for lockless synchronization, but maybe in a future PR we won't need that sync, I'm not sure yet, but quick answer is no

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: hmm, i could maybe add guards over the whole factory interface, good idea, tx
i'll add it as a part of PR part 14 if you don't insist on having it here

mikedennis: Wow part 14... this is epic :grin: Sounds good!

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
mikedennis: Ok sounds good.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
dangershony: How many seasons more are you planing 😆

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <param name="cancellation">Cancellation token that allows aborting the operation.</param>
+        /// <exception cref="OperationCanceledException">Thrown when the connection attempt was aborted.</exception>
+        public async Task ConnectAsync(IPEndPoint endPoint, CancellationToken cancellation)
+        {
+            this.logger = this.loggerFactory.CreateLogger(this.GetType().FullName, $"[{this.Id}-{endPoint}] ");
+            this.logger.LogTrace("({0}:'{1}')", nameof(endPoint), endPoint);
+
+            try
+            {
+                await Task.Run(async () => await this.tcpClient.ConnectAsync(endPoint.Address, endPoint.Port), cancellation);
```
dangershony: `this.tcpClient.ConnectAsync` does nto take a cancellation token that's why you create a task?
Will it even matter passing cancellation in such a way?
Can we not just rely on the tcp timeout?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        private ILogger logger;
+
+        /// <summary>Unique identifier of a client.</summary>
+        public int Id { get; private set; }
+
+        /// <summary>Underlaying TCP client.</summary>
+        private TcpClient tcpClient;
+
+        /// <summary>Prevents parallel execution of multiple write operations on <see cref="Stream"/>.</summary>
+        private AsyncLock writeLock;
```
dangershony: I find `writeLock` a bit confusing, this is a plane lock and there is not read lock

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// Sends data over the established connection.
+        /// </summary>
+        /// <param name="data">Data to send.</param>
+        /// <param name="cancellation">Cancellation token that allows aborting the operation.</param>
+        /// <exception cref="OperationCanceledException">Thrown when the connection was terminated or the cancellation token was cancelled.</exception>
+        public async Task SendAsync(byte[] data, CancellationToken cancellation)
+        {
+            this.logger.LogTrace("({0}.{1}:{2})", nameof(data), nameof(data.Length), data.Length);
+
+            using (await this.writeLock.LockAsync(cancellation))
```
dangershony: Come to think about it I am not sure a lock is required here, we will only push to this client messages sequentially from a blocking queue (unless you  have plan to change that)

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: There is still a lot to do, consider we don't still don't have any interface on what was node ...

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -198,52 +197,33 @@ public void BeginListen()
...
-                                {
-                                    if (socketEventManager.SocketEvent.SocketError != SocketError.Success)
-                                        throw new SocketException((int)socketEventManager.SocketEvent.SocketError);
-
-                                    processing.Completion.SetResult(true);
-                                    processing = null;
-                                }
-                            }
+                                ProtocolVersion = this.Peer.Version,
+                                TransactionOptions = this.Peer.SupportedTransactionOptions
```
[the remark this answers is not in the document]
Aprogiena: this is protocol, how can we change it, it would hardfork?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -198,52 +197,33 @@ public void BeginListen()
...
+                            });
+
+                            byte[] bytes = ms.ToArray();
+
+                            this.Client.SendAsync(bytes, this.Cancel.Token).GetAwaiter().GetResult();
+                            this.Peer.Counter.AddWritten(bytes.Length);
+                            processing.Completion.SetResult(true);
+                            processing = null;
```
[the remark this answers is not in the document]
Aprogiena: not my code, so i haven't changed it
but it makes sense because couple of lines below you see check if processing is null

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -653,46 +628,29 @@ public NetworkPeer(NetworkAddress peerAddress, Network network, NetworkPeerConne
...
 
-                        // Otherwise wait for the socket connection to complete OR if the operation got cancelled.
-                        WaitHandle.WaitAny(new WaitHandle[] { completedEvent, cancellation.WaitHandle });
+                this.RemoteSocketEndpoint = this.Connection.Client.RemoteEndPoint;
+                this.RemoteSocketAddress = this.RemoteSocketEndpoint.Address;
+                this.RemoteSocketPort = this.RemoteSocketEndpoint.Port;
 
-                        cancellation.ThrowIfCancellationRequested();
+                this.State = NetworkPeerState.Connected;
+                this.ConnectedAt = this.dateTimeProvider.GetUtcNow();
```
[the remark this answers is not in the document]
Aprogiena: actually no one uses ConnectedAt, so no need to change anything or making sure of anything, it could even be deleted

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: it is lock to protect write operations to the stream, therefore writeLock, i think it makes perfect sense :) there is no read lock because reading is done in serial fashion, but writing can be requested by multiple threads at the time

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: yes, this is sadly imperfect API that we can't do much about
this is recommended way of introducing cancellation token to connect here
it does matter a lot because when you shutdown, you don't want to wait 30 or whatever seconds
so yes, it does matter

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: i changed that, yes, i completely removed the queue, not sure in this PR though :-/

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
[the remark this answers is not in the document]
dangershony: Nah the name will change we will plug in a NetworkOptions that will have the same data just more.
It was just a comment not related to the PR

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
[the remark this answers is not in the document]
dangershony: ah right yes there is logic depending on this

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Ah ok if this is the recommended way to override cancellations then very good

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            this.logger = this.loggerFactory.CreateLogger(this.GetType().FullName, $"[{this.Id}-{endPoint}] ");
+            this.logger.LogTrace("({0}:'{1}')", nameof(endPoint), endPoint);
+
+            try
+            {
+                await Task.Run(async () => await this.tcpClient.ConnectAsync(endPoint.Address, endPoint.Port), cancellation);
+                this.Stream = this.tcpClient.GetStream();
+            }
+            catch (Exception e)
```
dangershony: Why not ?
` catch (OperationCanceledException oce)`

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Awesome yes we let the scheduler be the queue.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+                    this.logger.LogTrace("Connection has been terminated.");
+                    this.logger.LogTrace("(-)[NO_STREAM]");
+                    throw new OperationCanceledException();
+                }
+
+                try
+                {
+                    await this.Stream.WriteAsync(data, 0, data.Length, cancellation);
+                }
+                catch (Exception e)
```
dangershony: Same here why not catch the exception type?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Initializes a new network client.
+        /// </summary>
+        /// <param name="Id">Unique identifier of a client.</param>
+        /// <param name="tcpClient">Initialized TCP client, which may or may not be already connected.</param>
+        /// <param name="loggerFactory">Factory for creating loggers.</param>
+        public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFactory)
+        {
+            this.tcpClient = tcpClient;
```
dangershony: ` this.id = Id;` no?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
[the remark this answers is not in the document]
Aprogiena: ok tx

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -81,6 +80,12 @@ public ulong Nonce
         /// <seealso cref="ProcessMessage(IncomingMessage)"/>
         private readonly EventLoopMessageListener<IncomingMessage> listener;
 
+        /// <summary>List of connected clients mapped by their unique identifiers.</summary>
+        private readonly ConcurrentDictionary<int, NetworkPeerClient> clientsById;
```
dangershony: What the purpose of this dic? beyond being a index mapping

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -164,23 +172,20 @@ public void Listen(int maxIncoming = 20)
             {
                 try
                 {
-                    int taskCount = Interlocked.Increment(ref this.unfinishedTasks);
-                    this.logger.LogTrace("Starting accepting task, there are {0} unfinished tasks now.", taskCount);
+                    this.serverCancel.Token.ThrowIfCancellationRequested();
```
dangershony: You don't need this if the condition is in the loop statement. 
` this.serverCancel.Token.ThrowIfCancellationRequested();`

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: very good eyes :)

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: yes here it makes better sense to catch like that, will fix

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: here it is little more complicated and i have a similar handling for IO and Cancel, so it seems better to me in this particular case to have it not separated, like discussed here https://stackoverflow.com/questions/136035/catch-multiple-exceptions-at-once

so here i would like to keep it as it is

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: so before i believe the node had no idea about who is connected to it
so we did have some pending processing of the clients and on shutdown we could not wait until that processing is over. simply because there was no reference to those existing clients 

with this we are aware of who is connected to us and on shutdown we wait until everyone is gracefully disconnected, so then there will be no new incoming messages after shutdown

so should help with correct shutdown

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: this method is improved in later PR, but you are correct again

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Yeah fair enough I didn't think about same logic scenarios

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
dangershony: Don't we have `ConnectedNetworkPeers` for that?

Aprogiena: Unfortunately no, the network peer is connect much earlier before it is added to ConnectedNetworkPeers. That part is also a hot candidate for being rewritten, not an easy one though.

What we need here is something that binds the server with the client connections for the whole lifetime of the clients from were first moment to the very last moment WITHOUT any gaps. ConnectedNetworkPeers does not satisfy that requirement.

PR #876, opened 2017-12-09
Upgrade NBitcoin to .NET Standard 2.0

mikedennis:
Integration tests pass and same NBitcoin tests pass as prior to update.


src/NBitcoin/Payment/BitcoinUrlBuilder.cs
```diff
@@ -230,7 +230,7 @@ private static void WriteParameters(Dictionary<string, string> parameters, Strin
                     builder.Append("&");
                 builder.Append(parameter.Key);
                 builder.Append("=");
-                builder.Append(HttpUtility.UrlEncode(parameter.Value));
+                builder.Append(System.Web.NBitcoin.HttpUtility.UrlEncode(parameter.Value));
```
fassadlr: This namespace seems weird? Does ```System.Web.NBitcoin``` exist?

mikedennis: Indeed it does... https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/Payment/HttpUtility.cs We might be able to replace it with System.Web.HttpUtility Will give it a try. I did some google searching and there seemed to be some drama in the past with accessing it in certain version of .net framework which doesn't appear to be a problem anymore. So I suspect that code was to work around it. I will run some tests and if it hangs together get rid of the offending code and update this PR.

mikedennis: I think best to leave as is. Changing it to use System.Web.HttpUtility then breaks CanParsePaymentUrl() test. It seems System.Web.HttpUtility escapes symbols slightly different which is probably harmless but since I know nothing about what actually consumes this functionality it's probably best to leave it as is so the tests pass.

dangershony:
Hi @mikedennis any special reason to do this?
We still have some components  that relay on NBitcoin being 1.6 if I am not mistaken.

mikedennis:
I don't think so unless they are external to the full node, everything I checked there seems to be at 2.0. I was mainly removing the 4.61 framework build because that errors when you try to build on linux/osx but I can revert back to 1.6 if you think we need it for something.

bokobza:
The BlockExplorer is using NBitcoin but from the NBitcoin repo, so what we do here doesn't affect it.
And when BlockExplorer is using the full node, it's using the nuget package.
I think this can be changed to .NETStandard without problem.

dangershony:
>I think this can be changed to .NETStandard without problem.

It is already .NETStandard you mean to 2.0 I guess.


yeah  4.61 and 1.6 where there for a reason block explorer and indexer (and also older versions of those use it, how much they use it is not clear)

@quantumagi  @YakupIpek can perhaps provide some info

bokobza:
@YakupIpek what do you think?
NStratis has now moved to be inside the full node solution.
From what I see, you're using the StratisBitcoinFullNode projects in `Stratis.Bitcoin.Features.AzureIndexer.csproj` which targets netstandard2.0, in `Stratis.Bitcoin.Features.AzureIndexer` which targets netstandard2.0 and net471.

Does `Stratis.Bitcoin.Features.AzureIndexer` really need to target net471? 
What do you think, can we move NStratis in the StratisBitcoinFullNode to be netstandard2.0? Do yo plan to get a new version of the Stratis.Bitcoin anytime soon?

mikedennis:
@andrasfuchs I'm not sure what exactly NETSTANDARD1_6 is being used for but in latest version of NBitcoin I notice Nicolas is no longer using that conditional compile. Probably another future PR to get rid of NETSTANDARD1_6 define and replace it with whatever NBitcoin is currently using as he doesn't seem to be using version specific ones anymore.

src/NBitcoin/Payment/BitcoinUrlBuilder.cs
Aprogiena: there, i fixed it #1003

PR #878, opened 2017-12-09
For chain searches use skip list method instead of linear searches

mikedennis:
This PR 
* Replaces linear searches on chain headers with GetAncestor which uses skip list to search for block heights which will improve performance
* Move FindFork method that was only used by tests into tests

Issue #659

### DIFF
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -186,7 +186,11 @@ namespace NBitcoin
 
         /// <summary>
         /// Finds the ancestor of this entry in the chain that matches the chained block header specified.
-        /// <remarks>Will first search assuming on existing chain if not then will assume might be a different chain and then will linearly search for the hash.</remarks>
+        /// This can be used when you not sure whether the <paramref name="chainedBlockHeader"/> is on
+        /// the same chain as this block header or not.
+        /// <remarks>Will first search assuming on same chain and looking up height using skiplist, 
+        /// if doesn't match then will assume might be a different chain and then will linearly search 
+        /// for the hash.</remarks>
         /// </summary>
         /// <param name="chainedBlockHeader">The chained block header to search for.</param>
         /// <returns>The chained block header or null if can't be found.</returns>
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using NBitcoin.BouncyCastle.Math;
 
 namespace NBitcoin
@@ -183,6 +182,20 @@ namespace NBitcoin
         public override string ToString()
         {
             return this.Height + "-" + this.HashBlock;
+        }
+
+        /// <summary>
+        /// Finds the ancestor of this entry in the chain that matches the chained block header specified.
+        /// <remarks>Will first search assuming on existing chain if not then will assume might be a different chain and then will linearly search for the hash.</remarks>
+        /// </summary>
+        /// <param name="chainedBlockHeader">The chained block header to search for.</param>
+        /// <returns>The chained block header or null if can't be found.</returns>
+        public ChainedBlock FindAncestor(ChainedBlock chainedBlockHeader)
+        {
+            ChainedBlock found = this.GetAncestor(chainedBlockHeader.Height);
+            if (found.HashBlock == chainedBlockHeader.HashBlock)
+                return found;
+            return this.FindAncestorOrSelf(chainedBlockHeader.HashBlock);
         }
 
         /// <summary>
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -311,7 +311,7 @@ namespace NBitcoin
                 pastHeight = lastBlock.Height - (consensus.DifficultyAdjustmentInterval - 1);
             }
 
-            ChainedBlock firstChainedBlock = this.EnumerateToGenesis().FirstOrDefault(o => o.Height == pastHeight);
+            ChainedBlock firstChainedBlock = this.GetAncestor((int)pastHeight);
             if (firstChainedBlock == null)
                 throw new NotSupportedException("Can only calculate work of a full chain");
 
--- a/src/NBitcoin/ChainedBlock.cs
+++ b/src/NBitcoin/ChainedBlock.cs
@@ -186,20 +186,17 @@ namespace NBitcoin
 
         /// <summary>
         /// Finds the ancestor of this entry in the chain that matches the chained block header specified.
-        /// This can be used when you not sure whether the <paramref name="chainedBlockHeader"/> is on
-        /// the same chain as this block header or not.
-        /// <remarks>Will first search assuming on same chain and looking up height using skiplist, 
-        /// if doesn't match then will assume might be a different chain and then will linearly search 
-        /// for the hash.</remarks>
+        /// <remarks>It will compare the hash at the same height in the chain to verify the chained block
+        /// header has been found.</remarks>
         /// </summary>
         /// <param name="chainedBlockHeader">The chained block header to search for.</param>
-        /// <returns>The chained block header or null if can't be found.</returns>
-        public ChainedBlock FindAncestor(ChainedBlock chainedBlockHeader)
+        /// <returns>The chained block header or <c>null</c> if can't be found.</returns>
+        public ChainedBlock FindAncestorOrSelf(ChainedBlock chainedBlockHeader)
         {
             ChainedBlock found = this.GetAncestor(chainedBlockHeader.Height);
             if ((found != null) && (found.HashBlock == chainedBlockHeader.HashBlock))
                 return found;
-            return this.FindAncestorOrSelf(chainedBlockHeader.HashBlock);
+            return null;
         }
 
         /// <summary>
@@ -457,6 +454,7 @@ namespace NBitcoin
                     lowChain = lowChain.Previous;
                     highChain = highChain.Previous;
                 }
+
                 if ((lowChain == null) || (highChain == null))
                     return null;
             }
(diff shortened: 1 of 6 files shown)


Aprogiena:
i'll take this one

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
```diff
@@ -12,10 +12,8 @@ public static ChainedBlock FindFork(this ChainedBlock newTip, ChainedBlock tip)
...
             ChainedBlock highest = newTip.Height > tip.Height ? newTip : tip;
             ChainedBlock lowest = highest == newTip ? tip : newTip;
-            while (lowest.Height != highest.Height)
-            {
-                highest = highest.Previous;
-            }
+
+            highest = highest.GetAncestor(lowest.Height);
 
             while (lowest.HashBlock != highest.HashBlock)
```
Aprogiena: and finally, also not for this PR :-), this loop can be improved by checking if Skip hashes match or not
if i understand it well, both headers should have same height of their skip headers, so comparing hashes of skip headers makes sense and if they are not equal, we can jump to skip instead of to previous

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -139,7 +139,7 @@ public virtual void ProcessBlock(Block block)
 
                 if (newTip.Height > this.walletTip.Height)
                 {
-                    ChainedBlock findTip = newTip.FindAncestorOrSelf(this.walletTip.HashBlock);
+                    ChainedBlock findTip = newTip.GetAncestor(this.walletTip.Height);
```
Aprogiena: this change is wrong, consider newTip being on chain that does not contain walletTip

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -197,7 +197,7 @@ public virtual void ProcessBlock(Block block)
                 }
                 else
                 {
-                    ChainedBlock findTip = this.walletTip.FindAncestorOrSelf(newTip.HashBlock);
+                    ChainedBlock findTip = this.walletTip.GetAncestor(newTip.Height);
```
Aprogiena: same here, this change is wrong
you can use get ancestor for this instead of FAOS, but in fact it would be better to change implementation of FAOS (and maybe add tests to it)

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
mikedennis: Yes if they aren't on the same chain this won't work. I'll revert it back to use FindAncestorOrSelf which does a linear search for hash.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
mikedennis: I'm not sure how FAOS can be improved? It's searching for a hash, I can't see the skip list being of any benefit unless maybe we first try GetAncestor and then check the hash and then if not equal (meaning, different chain) then do the linear search.

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
```diff
@@ -12,10 +12,8 @@ public static ChainedBlock FindFork(this ChainedBlock newTip, ChainedBlock tip)
         {
```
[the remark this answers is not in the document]
mikedennis: I did a search for references to this method... not used. Not sure the intention of this - where would we move it?

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
```diff
@@ -12,10 +12,8 @@ public static ChainedBlock FindFork(this ChainedBlock newTip, ChainedBlock tip)
...
             ChainedBlock highest = newTip.Height > tip.Height ? newTip : tip;
             ChainedBlock lowest = highest == newTip ? tip : newTip;
-            while (lowest.Height != highest.Height)
-            {
-                highest = highest.Previous;
-            }
+
+            highest = highest.GetAncestor(lowest.Height);
```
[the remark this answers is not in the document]
mikedennis: Yes I found highest and lowest confusing in this context as well.

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
mikedennis: I don't think that would work with hashes. What if the matching hash was in between the current block and the skip. Wouldn't you miss it then?

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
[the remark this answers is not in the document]
Aprogiena: ah ok, it is not used so we can ignore everything related to it :D we should delete it then

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
Aprogiena: yes, but i'm talking about the opposite scenario
if those hashes are different for skips, you can advanced there
but not vice versa

mikedennis: Ah ok, yeah I get it now. Yes that will work if the skip heights are the same. Cool idea but too bad this method isn't used :grin: Maybe we could do the same on one of the other FindFork methods?

Aprogiena: definitely

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: well FAOS is slow as it traverses the chain ... improve it with skip list

Aprogiena: OK, so here the improvement could be done because you have chained blocks in your hands
so what you could do is to implement a version of FAOS that works not only with hash
then you could use skiplist

src/NBitcoin/ChainedBlock.cs
```diff
@@ -185,6 +184,24 @@ public override string ToString()
...
+        /// <summary>
+        /// Finds the ancestor of this entry in the chain that matches the chained block header specified.
+        /// This can be used when you not sure whether the <paramref name="chainedBlockHeader"/> is on
+        /// the same chain as this block header or not.
+        /// <remarks>Will first search assuming on same chain and looking up height using skiplist, 
+        /// if doesn't match then will assume might be a different chain and then will linearly search 
+        /// for the hash.</remarks>
+        /// </summary>
+        /// <param name="chainedBlockHeader">The chained block header to search for.</param>
+        /// <returns>The chained block header or null if can't be found.</returns>
```
Aprogiena: `<c>null</c>`

src/NBitcoin/ChainedBlock.cs
```diff
@@ -185,6 +184,24 @@ public override string ToString()
...
+        public ChainedBlock FindAncestor(ChainedBlock chainedBlockHeader)
+        {
+            ChainedBlock found = this.GetAncestor(chainedBlockHeader.Height);
+            if ((found != null) && (found.HashBlock == chainedBlockHeader.HashBlock))
+                return found;
+            return this.FindAncestorOrSelf(chainedBlockHeader.HashBlock);
```
Aprogiena: i don't think we need this, it will return null anyway, so we can just return 

if we have ChainedBlock, this means we have Height and we have Hash, so this means that if that chained block was ancestor of `this` it could only be at that Height, can't be anywhere else

and the method should probably still be named FindAncestorOrSelf because it should be possible to return self

src/NBitcoin/ChainedBlock.cs
```diff
@@ -430,8 +447,16 @@ public ChainedBlock FindFork(ChainedBlock block)
...
+                {
+                    highChain = highChain.Skip;
+                    lowChain = lowChain.Skip;
+                }
+                else
+                {
+                    lowChain = lowChain.Previous;
+                    highChain = highChain.Previous;
+                }
                 if ((lowChain == null) || (highChain == null))
```
Aprogiena: empty line above this line

src/NBitcoin/ChainedBlock.cs
mikedennis: So I guess the difference with this method and GetAncestor() is this validates the hash. Will this work properly in the wallet sync case now? I had a hard time getting my head around what exactly was going on there.

Aprogiena: the wallet is receiving a new block from a source of blocks
the wallet maintains its own tip and the new block can be in several different states relative to this tip

it could be 
- the next block on the same chain
- the next block on a different chain (reorg)
- maybe even something completely different, like lower height block but on different chain due to reorg

but it does not really matter what is going on there, important thing is not to change the functionality, so we have two places we change code there:

1) `if (newTip.Height > this.walletTip.Height)`
here the new tip is higher than the wallet tip, so if we are on the same chain, we just need to sync the gap (if there is any)

we've had `ChainedBlock findTip = newTip.FAOS(this.walletTip.Height);` there, which attempted to go from newTip down and either walletTip was on the same chain or not, if not, it would go through the whole chain down to the genesis. but that's inefficient, we don't need to go lower than walletTip.Height to know that it is not there, so we need null here in such case

2) the else branch of 1) if with `ChainedBlock findTip = this.walletTip.GetAncestor(newTip.Height);`, so it is the same, just the higher tip is walletTip, so the very same applies here

in case 1) if the new tip is on the same chain as wallet, there is some code to sync the gap

mikedennis:
@Aprogiena Thanks for detailed explanation on wallet, That makes more sense now. I've made the changes you suggested and also found one other place we can use the skiplist version - light wallet sync.

So we are down to only two places that use the old (slow) findancestorsorself - BlockValidator in NBitcoin and ChainHeadersBehavior so definitely making progress!

PR #879, opened 2017-12-10
Bugfix: occasional problem with node being not synced with its peers

### LINKED ISSUE #862: [Network sync] Investigate occasional problem with node being not synced with its peers
The testing scenario looks like this:
- run a node with just `-testnet` argument
- run a second node that will have `-testnet -stake=1 -walletname=yourwallet -walletpassword=yourpassword -connect=firstnode -`, this will make sure that the second node is only connected to the network via the first node and that the second node is staking
- now sometimes it seems that when the second node mines a block, the first node does not accept it, it does not even advance its headers, so the situation is that the second node sends the headers to the first node, say height of the new block is `X`, and then you can see the first node to be at height `X-1` for quite a long time and then it even accepts competitive block from another peer at height `X` - this is what we want to investigate why this happens (it should not)
- this log settings can help with investigation:
dedicated target:
```
    <target name="debugInvestigateFile" xsi:type="AsyncWrapper" queueLimit="10000" overflowAction="Block" batchSize="1000">
      <target xsi:type="File" fileName="investigate.txt" layout="[${longdate:universalTime=true} ${threadid}${mdlc:item=id}] ${level:uppercase=true}: ${callsite} ${message}" encoding="utf-8" /> 
    </target>
```
rules for the target:
```
    <logger name="Stratis.Bitcoin.Base.ChainHeadersBehavior" minlevel="Trace" writeTo="debugInvestigateFile" />
    <logger name="Stratis.Bitcoin.Features.BlockStore.*" minlevel="Trace" writeTo="debugInvestigateFile" />
    <logger name="Stratis.Bitcoin.P2P.Peer.*" minlevel="Trace" writeTo="debugInvestigateFile" />
    <logger name="Stratis.Bitcoin.Features.Miner.*" minlevel="Trace" writeTo="debugInvestigateFile" />
```

noescape00:
## Overview of the problem

If we make a following setup: 
Node 1 (let's call it connector node) is connected to the network and Node 2
Node 2 (let's call it miner node) is ONLY connected to Node 1 and staking.

We'd occasionally get a bug when miner or connector would get behind the other by 1-3 blocks.

## Why this bug is happening

The problem is that one of the nodes occasionally won't announce new block's hash to the other. 

When we are announcing blocks we are checking every hash if we should indeed send it to a connected peer. If the peer we want to announce the hash to has a hash at the same height as ours we won't send it to him because we assume that those are the same hashes.
And now we found a scenario when this is not correct: when peer is a miner and he mined an orphaned block. So in this case he would have a hash at the same height as we do but those hashes will be different.

## How this PR fixes it

What I'm proposing is that instead of assuming that hashes are equal we should check if they are. And if they are equal we do what we did before- skip it. But if they are not equal we will send the hash.

More info and whole investigation of the problem can be found in #862


Aprogiena:
will check this one

bokobza:
Excellent intro to the PR @noescape00

fassadlr:
Hi @noescape00! Do we have unit tests for this specific scenario? It would awesome if we had some tests that confirms the logic change. If there are already, which ones are they?

noescape00:
@fassadlr No, we don't have tests for that since malfunctions scenario was first time discovered by @Aprogiena a few days ago.  

I can write some integration tests that would simulate this scenario. 
Shell I include them in this PR or make a separate one?

fassadlr:
@noescape00 If its not too much work, it'll be nice to include at least one test that confirms that your logic changes work?

noescape00:
>it'll be nice to include at least one test that confirms that your logic changes work?

Sure, that's exactly what I was planning to do.

fassadlr:
OK great @noescape00! 💯

Aprogiena:
I'm checking this and I believe (although I'm not sure at all) that we've [also] seen different scenarios that are not explained by the above. So my investigation here will take more time, I need to be sure that that case I have in my mind never happened despite me remembering it.

Aprogiena:
so after we checked the old logs, we've discovered that this PR is about something else than i meant in #862, it is yet to be discovered whether this PR also fixes that problem or not

Aprogiena:
OK so we agreed to separate the fix of #862 from this PR, so I continue to check this one.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -257,23 +257,35 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
-                        foundStartingHeader = true;
-                        headers.Add(chainedBlock.Header);
-                    }
                     else
                     {
-                        // Peer doesn't have this header or the prior one -- nothing will
-                        // connect, so bail out.
-                        revertToInv = true;
-                        break;
+                        ChainedBlock peersBlock = chainBehavior.PendingTip.GetAncestor(chainedBlock.Height);
```
Aprogiena: this code is non-trivial, i'd suggest adding comments to every branch and sub-branch of this execution flow

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -257,23 +257,35 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
+
+                        if (peersBlock != null)
+                        {
+                            if (peersBlock.HashBlock == chainedBlock.HashBlock)
+                                continue;
+
+                            headers.Add(chainedBlock.Header);
```
Aprogiena: are you missing `foundStartingHeader = true` here or not?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -257,23 +257,35 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
+                        }
+                        else
+                        {
+                            ChainedBlock prevPeersBlock = chainBehavior.PendingTip.GetAncestor(chainedBlock.Previous.Height);
+
+                            if (prevPeersBlock == null || prevPeersBlock.HashBlock != chainedBlock.Previous.HashBlock)
```
Aprogiena: make me happy pls with () around the conditions

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -257,23 +257,35 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
+                            {
+                                // Peer doesn't have this header or the prior one - nothing will connect, so bail out.
+                                revertToInv = true;
+                                break;
+                            }
+                            else
```
Aprogiena: when you break/return in if you don't need else

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: if you do, consider DRY and merge this case with the last case below

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
dangershony: I would suggest that we search by hash, but searching by height is much faster (now with skip list)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -257,23 +257,35 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
                     {
                         headers.Add(chainedBlock.Header);
                     }
-                    else if (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) != null)
```
dangershony: I would suggest an improvement, this code can move the the state changed even and assign to ta local variable (see the con manager behavior)
` var chainBehavior = node.Behavior<ChainHeadersBehavior>();`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
noescape00: >I would suggest that we search by hash, but searching by height is much faster (now with skip list)

We check the hash later anyway so in this case it's better to do a faster search by height imho.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
noescape00: >are you missing foundStartingHeader = true here or not?

No. This flag is for the cases when some of the hashes are good and some are bad (bad = we think peer won't validate them) so we want to start sending starting from the first one that will be approved.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: i don't understand what you mean

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: hmm my understanding of this flag was totally different, we should get in consensus on what this is first

noescape00:
Fixed everything except @dangershony's suggestion. Waiting for him to elaborate since I also didn't quite understand it.

Aprogiena:
Meanwhile we have implemented a new FindAncestorOrSelf method which we should use in this PR instead of GetAncestor. So I'd like to ask you for that one more change.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -247,34 +247,32 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
-                    {
-                        continue;
-                    }
-                    else if (chainBehavior.PendingTip.GetAncestor(chainedBlock.Previous.Height) != null)
+
+                    if (!foundStartingHeader)
                     {
-                        // Peer doesn't have this header but they do have the prior one.
-                        // Start sending headers.
+                        // Peer don't have a block at the height of our block and with the same hash?
```
Aprogiena: doesn't

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -247,34 +247,32 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
...
+                        if (chainBehavior.PendingTip.FindAncestorOrSelf(chainedBlock) == null)
+                            continue;
+
+                        // Peer don't have a block at the height of our block.Previous and with the same hash?
```
Aprogiena: doesn't

PR #881, opened 2017-12-11
Added a few coding style rules

bokobza:


Documentation/coding-style.md
```diff
@@ -59,6 +59,22 @@ The general rules:
       [EMPTY LINE HERE]
       //more code here...
    ```   
+19. We declare only one type per file.  
+    A few exceptions exist, which are permitted, although not enforced:
+    - inclusion of small behaviorless classes into the same file
```
dangershony: What does `behaviorless` mean?

bokobza: Maybe not the right work. what I mean is a class that only has properties, but no method or logic.
POCO or POJO or whatever may be more appropriate.

dangershony: I personally would prefer 
>inclusion of classes of same functionality purpose into the same file

bokobza: what's `same functionality purpose ` ?.

bokobza: How about:
- inclusion of small POCO classes if they're closely related in functionality (like the API models classes)

dangershony: But that means restricted to POCO which I don' want to restrict, remove POCO and all is good
>inclusion of types if they're closely related in functionality

bokobza: That's the thing, I don't want to have types that are related in functionality, That's how we end up with a file like Transaction.cs.

dangershony: Ah right fair enough Transaction is a good example of what we try to avoid.
Transaction (and its classes) needs to be put in its own folder.
I guess that's why you put in the rule the word small 👍

bokobza: you got it baby

bokobza: @dangershony is that ok then?
```
inclusion of small POCO classes if they're closely related in functionality (like API models classes)
```

dangershony: Ok with me yes,

PR #884, opened 2017-12-11
Attempt to fix stopwatch test

Aprogiena:
It seems that the testing environment is quite unreliable. We thus try to
- increase the epsilon to accept slightly worse results than before
- wait a bit before the test starts in case some post processing is still going on (like GC or whatever) from previous tests


bokobza:
Looks sexy,

PR #885, opened 2017-12-11
Refactor/the node job part 12

Aprogiena:
This is a short episode of our series, it includes:
- removing the two threads from NetworkPeerConnection; those were responsible for sending and receiving messages, we are now doing that in an almost async way (this is improved again later). here we remove the sent message queue completely from the code
- we are making NetworkPeerConnection IDisposable, this helps with graceful shutdown, which is improved throughout the series
- we are making sending messages async, which unfortunately causes compiler warnings in lots of code, which we compensate with temporary introduction of a new fire and forget special sending function, but using it is discouraged and an end goal is to refactor all the code using this fire and forget send message function as it is not a good practice

Overall goal of this is PR is being closer to a fully async code with a correct shutdown sequence.


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -174,170 +174,145 @@ public NetworkPeerConnection(NetworkPeer peer, NetworkPeerClient client, IDateTi
...
+                    this.Peer.DisconnectReason = new NetworkPeerDisconnectReason()
+                    {
+                        Reason = "Unexpected exception while sending a message",
+                        Exception = ex
+                    };
                 }
 
-                if (processing != null)
-                    this.Messages.Add(processing);
+                this.Cancel.Cancel();
```
mikedennis: ``` Cancel.Cancel() ``` I wonder if it might be better with calling it CancellationSource or CancelSrc or something like that so you don't end up with Cancel.Cancel() or maybe even ```Process.Cancel()```?

Aprogiena: I agree, I will rename it CancellationSource in part 15

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -154,11 +154,11 @@ public class NetworkPeerConnection
         /// <summary>Cancellation to be triggered at shutdown to abort all pending operations on the connection.</summary>
         public CancellationTokenSource Cancel { get; private set; }
 
-        /// <summary>Queue of messages to be sent to a peer over the network connection.</summary>
-        internal BlockingCollection<SentMessage> Messages;
+        /// <summary>Registration of callback routine to shutdown the connection when <see cref="Cancel"/> is cancelled.</summary>
```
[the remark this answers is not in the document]
Aprogiena: tx, i'll improve the comment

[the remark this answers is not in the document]
Aprogiena: in part 15

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -964,12 +951,12 @@ public void VersionHandshake(NetworkPeerRequirement requirements, CancellationTo
...
                 }
 
-                this.SendMessageAsync(new VerAckPayload());
+                this.SendMessageVoidAsync(new VerAckPayload());
                 listener.ReceivePayload<VerAckPayload>(cancellationToken);
                 this.State = NetworkPeerState.HandShaked;
                 if (this.Advertize && this.MyVersion.AddressFrom.Address.IsRoutable(true))
                 {
-                    this.SendMessageAsync(new AddrPayload(new NetworkAddress(this.MyVersion.AddressFrom)
+                    this.SendMessageVoidAsync(new AddrPayload(new NetworkAddress(this.MyVersion.AddressFrom)
```
dangershony: Is this correct?
Sending an empty address list (well just with our nodes address)

PR #886, opened 2017-12-11
Consensus rules move rules to the engine

dangershony:
This PR will aggregate some commits that move validation rules from PowConsensusValidator to use ConsensusRules

It will also contain some tests to check the moved rules.


src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/BlockHeaderRuleTest.cs
```diff
@@ -0,0 +1,38 @@
...
+        public async Task BlockReceived_NotNextBlock_ValidationFailAsync()
+        {
+            var testContext = TestRulesContextFactory.CreateAsync(Network.RegTest);
+            BlockHeaderRule blockHeaderRule = testContext.CreateRule<BlockHeaderRule>();
+
+            var context = new RuleContext(new BlockValidationContext(), Network.RegTest.Consensus, testContext.Chain.Tip);
+            context.BlockValidationContext.Block = new Block(new BlockHeader { HashPrevBlock = uint256.Zero });
+            var error = await Assert.ThrowsAsync<ConsensusErrorException>(async () => await blockHeaderRule.RunAsync(context));
+
+            Assert.True(error.ConsensusError == ConsensusErrors.InvalidPrevTip);
```
mikedennis: Think you might get a nicer error message if you do Assert.Equal(ConsensusErrors.InvalidPrevTip, error.ConsensusError)

bokobza:
There are a few failing tests:
image

Are they related to your change?

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/BlockHeaderRuleTest.cs
dangershony: Ah yes thanks

dangershony:
Looking at the errors now.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -494,11 +469,11 @@ public void ValidateBlock(RuleContext context, bool skipRules = false)
         /// Validates a block using the consensus rules and executes it (processes it and adds it as a tip to consensus).
         /// </summary>
         /// <param name="context">A context that contains all information required to validate the block.</param>
-        internal async Task ValidateAndExecuteBlockAsync(RuleContext context)
+        internal async Task ValidateAndExecuteBlockAsync(RuleContext context, bool skipRules = false)
```
bokobza: In the previous iteration, the default skipRules was `true` but now it's `false`.
Is that intentional?

dangershony: Yes its now derived from the caller and its still true when we come from the method `ValidateAndExecuteBlockAsync`

PR #889, opened 2017-12-12
Return wallet file path in general-info api call

### LINKED ISSUE #883: WalletFilePath returns Null
Per conversation with @mikedennis we determined the Wallet Controller isn't filling the parameter for WalletFilePath in https://github.com/stratisproject/StratisBitcoinFullNode/blob/15d3ddcbf9a0da23ab8e5c775c893c26f269f515/src/Stratis.Bitcoin.Features.Wallet/Models/WalletGeneralInfoModel.cs#L10

Resulting in general-info api request to the node returning null:
```bash
{
  "walletFilePath": null,
  "network": "Main",
  "creationTime": "1512431701",
  "isDecrypted": true,
  "lastBlockSyncedHeight": 498757,
  "chainTip": 498757,
  "isChainSynced": true,
  "connectedNodes": 8
}
```

Possibly the Name Parameter should be the name of the Wallet?

mikedennis:
This fixes #883

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
@@ -614,6 +614,14 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             var mockWalletWrapper = new Mock<IWalletManager>();
             mockWalletWrapper.Setup(w => w.GetWallet("myWallet")).Returns(wallet);
 
+            string walletFileExtension = "wallet.json";
+            string testWalletFileName = Path.ChangeExtension("myWallet", walletFileExtension);
+            string testWalletPath = "c:\\stratisnode\\" + testWalletFileName;
+            string folder = Path.GetDirectoryName(testWalletPath);
+            string[] files = new string[] { testWalletFileName };
+            mockWalletWrapper.Setup(w => w.GetWalletsFiles()).Returns((folder, files));
+            mockWalletWrapper.Setup(w => w.GetWalletFileExtension()).Returns(walletFileExtension);
+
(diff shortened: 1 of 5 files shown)


bokobza:
Hi Mike, there is already a method that does that for you.
Look at a method called `ListWalletsFiles()` in the `WalletController`.

mikedennis:
Yes I did look at that but since I have the walletname I would have had to do some filename manipulation to find the right entry which I thought would be amount the same amount of work as adding a new method to build it from the name. What do you think?

Also since we have already have this API method is there value of having the path in the general-info call at all?

bokobza:
The this.walletManager.GetWalletsFiles() call gives you the path and the list of names.
Just combine the path with the name you're looking for and you're done. I think the extension is included in the result.
Also, you're code will give you a path even if there is no wallet with this name present.

mikedennis:
Yes good point. Will do- thanks!

bokobza:
All good, just one tiny thing.

mikedennis:
@bokobza Are you keeping me in suspense? :grin: I don't think you applied your review.

bokobza:
Did I not? Oh right.
So I was saying that the this.walletManager.GetWalletsFiles() already has extension in the filenames.
like this:
```
{
    "walletsPath": "C:\\Users\\jerem\\AppData\\Roaming\\StratisNode\\stratis\\StratisTest",
    "walletsFiles": [
        "newwallet.wallet.json",
        "recovered2.wallet.json",
        "recoveredwallet.wallet.json"
    ]
}
```

mikedennis:
Are you referring to this line?
```            string searchFile = Path.ChangeExtension(request.Name, this.walletManager.GetWalletFileExtension());```

I'm just using this to build a filename to lookup against walletFiles since the request only has the walletname in it.

bokobza:
Ah yes of course!

PR #895, opened 2017-12-12
Fixed second type of scenario bug described in issue 862

noescape00:
This is fix for the original bug described in issue 862. It looks like this bug is connected to having a wrong order of headers that node announces. If headers are announced in wrong order they will be invalidated by the peer since peer can't find prev header that received one is based upon. 

For storing the hashes we use ConcurrentDictionary and apparently it does not guarantee that the order in which we've added items will be maintained. [StackOverflow proof](https://stackoverflow.com/questions/20020074/is-the-list-order-of-a-concurrentdictionary-guaranteed).

Fix: #862


Aprogiena:
I actually think that
- dictionary should not be used here, queue would be better
- there still is a race condition between producer and consumer and a lock should be introduced there instead of concurrent structure

Aprogiena:
OK, one more thing, there is a problem that RelayWorker uses ConcurrentChain, it should not, it should use consensus tip instead. 

and the comment starting `NOTE: there is a very minimal possibility a reorg would happen` should be rewritten

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,34 +115,37 @@ public void RelayWorker()
...
+                while (this.blockHashesToAnnounce.TryPeek(out uint256 blockHash))
                 {
                     // The first block that is not in disk will abort the loop.
                     if (!await this.blockRepository.ExistAsync(blockHash).ConfigureAwait(false))
                     {
-                        // NOTE: there is a very minimal possibility a reorg would happen
-                        // and post reorg blocks will now be in the 'blockHashesToAnnounce',
-                        // current logic will discard those blocks, I suspect this will be unlikely
-                        // to happen. In case it does a block will just not be relays.
+                        // There is a small possibility that a reorg happened and 'blockHashesToAnnounce'
```
Aprogiena: avoid saying small/minimal possibility/chance
we are in blockchain code, reorgs/races are normal

dangershony: Yeah thats true, I think my initial intention was there is a small possibility a reorg happened in between the block arrived and pending relay.

Lets fix this comment to be
>In cases when the node had a reorg the 'blockHashesToAnnounce' will contain blocks that are not anymore on the main chain.
etc..

dangershony:
>OK, one more thing, there is a problem that RelayWorker uses ConcurrentChain

I am not sure thats so important, on most cases ConcurrentChain will represent the best chain.
If we use chain.ConsensusTip we might have reorged but not rewinded consensus yet.
Also ConcurrentChain.Contains() is Log(1)

noescape00:
So @Aprogiena do you agree with @dangershony or shell I use

`if (this.chainState.ConsensusTip.FindAncestorOrSelf(blockHash) == null)`

instead of 

` if (!this.chain.Contains(blockHash))`

?

Aprogiena:
I need fresh brain for this, it is nontrivial to think about. Gimme tomorrow morning.

noescape00:
>I need fresh brain for this, it is nontrivial to think about. Gimme tomorrow morning.

Sure, there is no rush. 
Thanks!

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,34 +115,37 @@ public void RelayWorker()
...
                     this.logger.LogTrace("(-)[NO_BLOCKS]");
                     return;
                 }
 
                 var broadcastItems = new List<uint256>();
-                foreach (uint256 blockHash in blocks)
+
+                while (this.blockHashesToAnnounce.TryPeek(out uint256 blockHash))
                 {
                     // The first block that is not in disk will abort the loop.
```
Aprogiena: on disk

Aprogiena:
OK, so I am thinking about it now and I don't think it is handled correctly.
So imagine the following reorg. Queue status 

`1->2a->3a->2b->3b->4b`

Block store seems to delete blocks that are out of the chain.
So let's say 2a->3a is deleted from store. Then the loop will find out and it finds out it is not in the chain due to reorg and it clears the queue. broadcastItems will contain 1 only and 2b-4b won't ever be propagated. This looks like something we want to prevent.

What if we only removed those items from the queue that are not on the current chain. In this scenario we would end up with 
- 1 - ok
- 2a - not found -> remove from queue
- 3a - not found -> remove from queue
- `2b->3b->4b` - good

and we would propagate `1->2b->3b->4b`, which seems correct

Now suppose reorgs happen to A chain again with queue state 
`1->2a->3a->2b->3b->4b->4a->5a`

so in this case we don't have 2b->4b in the store, so it seems like we will propagate `1->...5a` no problem and remove B chain

so the only problem here is seem asking the concurrent chain, which can be changed at any time
if we asked block store and consensus OR block store only we should be fine

Aprogiena:
thoughts @dangershony @noescape00 ?

noescape00:
@Aprogiena 
Sounds legit. Thanks for the analysis, I'll do changes implementing it the way you've said!

noescape00:
Two new commits were added but they are showed behind the messages since they have old timestamps. 
Please take a look.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -208,14 +208,14 @@ private Task ProcessSendCmpctPayload(NetworkPeer node, SendCmpctPayload sendCmpc
         }
 
         /// <inheritdoc />
-        public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
+        public Task AnnounceBlocks(List<ChainedBlock> blocksToAnnounce)
         {
-            this.logger.LogTrace("({0}.{1}:{2})", nameof(blockHashesToAnnounce), nameof(blockHashesToAnnounce.Count), blockHashesToAnnounce?.Count);
-            Guard.NotNull(blockHashesToAnnounce, nameof(blockHashesToAnnounce));
+            this.logger.LogTrace("({0}.{1}:{2})", nameof(blocksToAnnounce), nameof(blocksToAnnounce.Count), blocksToAnnounce?.Count);
+            Guard.NotNull(blocksToAnnounce, nameof(blocksToAnnounce));
```
Aprogiena: move guard in front of the entry log and you can remove `?` in `?.Count` then

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -304,16 +296,15 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
                 // The last entry in vBlockHashesToAnnounce was our tip at some point
```
Aprogiena: fix comment

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -242,16 +242,8 @@ public Task AnnounceBlocks(List<uint256> blockHashesToAnnounce)
                 // then send all headers past that one.  If we come across any
                 // headers that aren't on chainActive, give up.
```
Aprogiena: fix this comment pls

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
```diff
@@ -170,31 +170,19 @@ public virtual string StoreName
...
-            uint256 blockHash = block.GetHash();
-            this.logger.LogTrace("({0}:'{1}')", nameof(block), blockHash);
-
-            ChainedBlock chainedBlock = this.Chain.GetBlock(blockHash);
-            if (chainedBlock == null)
-            {
-                this.logger.LogTrace("(-)[REORG]");
-                return;
-            }
+            this.logger.LogTrace("({0}:'{1}')", nameof(blockPair.Block), blockPair.ChainedBlock.HashBlock);
```
Aprogiena: `this.logger.LogTrace("({0}:'{1}')", nameof(blockPair), blockPair.ChainedBlock);`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography.X509Certificates;
```
Aprogiena: eh?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -37,7 +38,7 @@ public class BlockStoreSignaled : SignalObserver<Block>
 
         private readonly StoreSettings storeSettings;
 
-        private readonly ConcurrentDictionary<uint256, uint256> blockHashesToAnnounce;// maybe replace with a task scheduler
+        private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
```
Aprogiena: add comment here, we know what this is for now

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -73,38 +74,51 @@ protected override void OnNextCore(Block value)
...
+            ChainedBlock chainedBlock = this.chain.GetBlock(block.GetHash());
+            if (chainedBlock == null)
+            {
+                this.logger.LogTrace("(-)[REORG]");
+                return;
+            }
+
+            BlockPair blockPair = new BlockPair(block, chainedBlock);
+
             // ensure the block is written to disk before relaying
```
Aprogiena: we can fix this comment - `E` and `.`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -73,38 +74,51 @@ protected override void OnNextCore(Block value)
...
             if (this.chainState.IsInitialBlockDownload)
             {
                 this.logger.LogTrace("(-)[IBD]");
                 return;
             }
 
-            uint256 blockHash = value.GetHash();
-            this.logger.LogTrace("Block hash is '{0}'.", blockHash);
-            this.blockHashesToAnnounce.TryAdd(blockHash, blockHash);
+            this.logger.LogTrace("Block hash is '{0}'.", chainedBlock.HashBlock);
```
Aprogiena: we can log the whole chainedBlock just after we obtain it - i.e. before new BlockPair

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -73,38 +74,51 @@ protected override void OnNextCore(Block value)
...
         }
 
         /// <summary>
-        /// A loop method that continuously relays blocks found in <see cref="blockHashesToAnnounce"/> to connected peers on the network.
+        /// A loop method that continuously relays blocks found in <see cref="blocksToAnnounce"/> to connected peers on the network.
         /// </summary>
         /// <remarks>
-        /// The dictionary <see cref="blockHashesToAnnounce"/> contains
+        /// <para>
+        /// The dictionary <see cref="blocksToAnnounce"/> contains
```
Aprogiena: is not dictionary

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,35 +128,36 @@ public void RelayWorker()
...
-                if (!blocks.Any())
+                if (!this.blocksToAnnounce.Any())
                 {
                     this.logger.LogTrace("(-)[NO_BLOCKS]");
                     return;
                 }
 
-                var broadcastItems = new List<uint256>();
-                foreach (uint256 blockHash in blocks)
+                var broadcastItems = new List<ChainedBlock>(this.blocksToAnnounce.Count + 4);
```
Aprogiena: such a +4 magic is worth a comment

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,35 +128,36 @@ public void RelayWorker()
...
-                    // The first block that is not in disk will abort the loop.
-                    if (!await this.blockRepository.ExistAsync(blockHash).ConfigureAwait(false))
+                    // The first block that is not on disk will abort the loop.
+                    if (!await this.blockRepository.ExistAsync(block.HashBlock).ConfigureAwait(false))
                     {
-                        // NOTE: there is a very minimal possibility a reorg would happen
-                        // and post reorg blocks will now be in the 'blockHashesToAnnounce',
-                        // current logic will discard those blocks, I suspect this will be unlikely
-                        // to happen. In case it does a block will just not be relays.
+                        // In cases when the node had a reorg the 'blocksToAnnounce' will contain blocks
```
Aprogiena: will contain -> contain

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,35 +128,36 @@ public void RelayWorker()
...
+                        // that are not anymore on the main chain, those blocks are removed from 'blocksToAnnounce'.
 
-                        // If the hash is not on disk and not in the best chain
-                        // we assume all the following blocks are also discardable
-                        if (!this.chain.Contains(blockHash))
-                            this.blockHashesToAnnounce.Clear();
+                        // Check if the reason why we don't have a block is a reorg or it wasn't just downloaded yet.
```
Aprogiena: hasn't been downloaded yet

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -114,35 +128,36 @@ public void RelayWorker()
...
+                        if (this.chainState.ConsensusTip.FindAncestorOrSelf(block) == null)
+                        {
+                            // Remove hash that we've reorged away from.
```
Aprogiena: from ... ?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -161,7 +176,7 @@ public void RelayWorker()
                 // Announce the blocks to each of the peers.
                 var behaviours = nodes.Select(s => s.Behavior<BlockStoreBehavior>());
```
Aprogiena: unvar both lines

Aprogiena: and the var nodes above

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: lol. 
No idea how it got there

PR #902, opened 2017-12-13
Added iscoinStake in coinStake transactions 

bokobza:
* I also tidied up a few bits in the WalletManager and in the MinerController.
* I'm also assigning null to the pos async loop after it's disposed. the way it is at the moment, before we start staking, we check for nulity of the async loop. So if it's not null, staking will never restart.

This PR is better reviewed commit by commit.
@dangershony @Aprogiena @fassadlr

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
@@ -8,6 +8,7 @@ using NBitcoin;
 using Stratis.Bitcoin.Features.Miner.Models;
 using Stratis.Bitcoin.Features.Wallet;
 using Stratis.Bitcoin.Features.Wallet.Interfaces;
+using Stratis.Bitcoin.Utilities;
 using Stratis.Bitcoin.Utilities.JsonErrors;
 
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
```diff
@@ -77,39 +83,29 @@ public IActionResult GetStakingInfo()
...
                     return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, "Formatting error", string.Join(Environment.NewLine, errors));
                 }
-
-                WalletManager walletManager = this.fullNode.NodeService<IWalletManager>() as WalletManager;
-
-                Wallet.Wallet wallet = walletManager.Wallets.FirstOrDefault(w => w.Name == request.Name);
-
-                if (wallet == null)
+                
+                Wallet.Wallet wallet = this.walletManager.GetWallet(request.Name);
```
dangershony: What if wallet name is not found?

bokobza: it throws an exception

PR #908, opened 2017-12-13
 Skeleton DNS feature and StratisDnsD with test project

345James:
Created a StratisDnsD daemon project that by default runs the base feature, consensus (for RPC), API, RPC and DNS features.

- New DNS feature project and corresponding test project (with minimal unit tests to cover what is there).
- Added a reference to the DNS nuget package which will be used by the DNS feature in future commits.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Dns.Tests/GivenADnsFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Dns.Tests/GivenADnsFeature.cs
@@ -14,7 +14,7 @@ namespace Stratis.Bitcoin.Features.Dns.Tests
     {
         [Fact]
         [Trait("UnitTest", "UnitTest")]
-        public void WhenConstructorCalledAndPeerAddressManagerIsNullThenArgumentNullExceptionIsThrown()
+        public void WhenConstructorCalled_AndPeerAddressManagerIsNull_ThenArgumentNullExceptionIsThrown()
         {
             // Arrange.
             Action a = () => { new DnsFeature(null, null); };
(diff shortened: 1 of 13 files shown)


345James:
[[DNS] Skeleton DNS feature and StratisDnsD with test project](https://trello.com/c/4EOMHOf3/326-dns-skeleton-dns-feature-and-stratisdnsd-with-test-project)

src/Stratis.StratisDnsD/Program.cs
```diff
@@ -0,0 +1,52 @@
...
+        static void Main(string[] args)
+        {
+            MainAsync(args).Wait();
+        }
+
+        public static async Task MainAsync(string[] args)
+        {
+            try
+            {
+                Network network = args.Contains("-testnet") ? Network.StratisTest : Network.StratisMain;
```
bokobza: The network shouldn't be limited to Stratis.
Have a look in BreezeD, there we support stratis or bitcoin: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.BreezeD/Program.cs#L34.
So for stratis you'd do `dotnet run stratis -testnet` and bitcoin you'd do `dotnet run -testnet`.

bokobza: Just realized that this project is called Stratis.StratisDnsD so maybe your intention was indeed to create a stratis only project.
How about making it network agnostic and calling it Stratis.DnsD?

src/Stratis.Bitcoin.Features.Dns.Tests/GivenADnsFeature.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Stratis.Bitcoin.Features.Dns.Tests
+{
+    /// <summary>
+    /// Tests for the <see cref="DnsFeature"/> class.
+    /// </summary>
+    public class GivenADnsFeature
+    {
+        [Fact]
+        [Trait("UnitTest", "UnitTest")]
```
bokobza: Would naming the trait `[Trait("Dns", "UnitTest")]` be more useful?

345James: Good point, makes it easier then to just run the Dns tests.

src/Stratis.StratisDnsD/Program.cs
345James: I have just checked and I believe the decision on Monday was to limit this to only the Stratis network. The PR was created with this in mind.

bokobza: Ok cool mate, that's fair enough.

PR #909, opened 2017-12-13
Adding the ability to stop staking via api

dangershony:


src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
```diff
@@ -114,5 +113,32 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
...
+        /// Stop staking.
+        /// </summary>
+        /// <returns>An OKResult object that produces a status code 200 HTTP response.</returns>
+        [Route("stopstaking")]
+        [HttpPost]
+        public IActionResult StopStaking()
+        {
+            try
+            {
+                if (!this.ModelState.IsValid)
```
bokobza: Do you need this if there is no model?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -402,14 +408,37 @@ public IAsyncLoop Stake(WalletSecret walletSecret)
...
+            if (this.stakingLoop == null)
+            {
+                this.logger.LogTrace("(-)[NOT_MINING]");
+                return;
+            }
+
+            this.stakeCancellationTokenSource.Cancel();
+            this.stakingLoop.Dispose();
+            this.stakingLoop = null;
+            this.stakeCancellationTokenSource = null;
```
Aprogiena: how about disposing first?

src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
dangershony: You tell me, I am not sure.

src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
```diff
@@ -114,5 +113,32 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
                 return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, e.Message, e.ToString());
             }
         }
+
+        /// <summary>
+        /// Stop staking.
+        /// </summary>
+        /// <returns>An OKResult object that produces a status code 200 HTTP response.</returns>
```
Aprogiena: OKResult?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: Disposing blocks

src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
dangershony: That's the name of the type, I can add a ref=

bokobza:
I'm ok with this, @Aprogiena do you want to have a look?

src/Stratis.Bitcoin.Features.Miner/Controllers/MinerController.cs
fassadlr: Yes, you don't need to check the ModelState here as you didn't post anything to the controller method.

Neurosploit: Yes I think the modelstate.isvalid can be removed on any controller method that does not have any parameters. I think you can remove it for GetStakingInfo() as well.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: what?

bokobza: Here: https://docs.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads it says:

> The CancellationTokenSource class implements the IDisposable interface. You should be sure to call the CancellationTokenSource.Dispose method when you have finished using the cancellation token source to free any unmanaged resources it holds.

dangershony: No but I mean the dispose implementation of the async loop blocks until the loop ends.
So we have to cancel first then wait for the loop to finish calling the Dispose()
Then we can dispose the cancellation.

PR #910, opened 2017-12-13
[P2P] Improve peer selection part 1

fassadlr:
OK, so I'm going to these commits (PRs) @Aprogiena style, so that you guys can follow the logic changes more easily.

Please note that the documentation isn't very complete as the selection algorithm is going to change in future PRs.

Part 1 includes: 

- Move the peer selection logic to a PeerSelector class (IPeerSelector).
- PeerAddressManager will now have a PeerSelector instance.
- The PeerSelector will now return PeerAddress and not NetworkAddress anymore.


src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -103,10 +90,14 @@ public sealed class PeerAddressManager : IPeerAddressManager
         /// <inheritdoc />
         public DataFolder PeerFilePath { get; set; }
 
+        /// <inheritdoc />
+        public IPeerSelector Selector { get; private set; }
```
dangershony: Its easier if you call the param `PeerSelector`

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Return a random peer from a given set of peers.
+        /// </summary>
+        public static PeerAddress Random(this IEnumerable<PeerAddress> peers)
+        {
+            var chanceFactor = 1.0;
+            var random = new Random();
+
+            while (true)
```
dangershony: Cancellation?

fassadlr: Yeah I could, but this "selectability" will be dropped later on ;)

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
fassadlr: Will rename 👍

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <summary>
+        /// The address manager instance that holds the peer list to be queried. 
+        /// </summary>
+        private readonly IPeerAddressManager peerAddressManager;
+
+        /// <summary>
+        /// Constructor for the peer selector.
+        /// </summary>
+        /// <param name="peerAddressManager">The singleton peer address manager instance.</param>
+        public PeerSelector(IPeerAddressManager peerAddressManager)
```
dangershony: This is a bit of an anti pattern `IPeerAddressManager` depends on `IPeerSelector` and vice-versa.
Its normally not encouraged, not much point of using an interface now right?

fassadlr: @dangershony you are right, I will look at improving the design in the next PR. The only thing the PeerSelector class does is to work with the peer collection. I might as well just pass the collection to it.

dangershony: yes sounds right

PR #911, opened 2017-12-13
Refactor/the node job part 13

Aprogiena:
In this episode we add ability to read message from the network to NetworkPeerClient, this allows us to replace non-async Message.ReadNext reading from the network with async reading from the network and we only use Message.ReadNext for parsing the message - see TODO in ReadAndParseMessageAsync.

This allows us to have fully async processing of newly connected clients.


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
 
+        /// <summary>
+        /// Reads raw message in binary form from the connection stream.
+        /// </summary>
+        /// <param name="protocolVersion">Version of the protocol that defines the message format.</param>
+        /// <param name="cancellation">Cancellation token that allows aborting the read operation.</param>
+        /// <returns>Binary message received from the connected counterparty.</returns>
+        /// <exception cref="OperationCanceledException">Thrown if the operation was cancelled or the end of the stream was reached.</exception>
+        /// <exception cref="FormatException">Thrown if the incoming message is too big.</exception>
+        public async Task<byte[]> ReadMessageAsync(ProtocolVersion protocolVersion, CancellationToken cancellation = default(CancellationToken))
```
dangershony: I would suggest we remove the `default(CancellationToken)` to avoid scenarios where we miss use it, unless you see a scenario where we would not use cancellation when reading from the socket.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+        {
+            this.logger.LogTrace("({0}:{1})", nameof(protocolVersion), protocolVersion);
```
dangershony: I would also question the need for trace in this method.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+            byte[] messageHeader = new byte[headerSize];
+            await this.ReadBytesAsync(messageHeader, 0, headerSize, cancellation).ConfigureAwait(false);
+
+            // Then extract the length, which is the message payload size.
+            int lengthOffset = Message.CommandSize;
+            uint length = BitConverter.ToUInt32(messageHeader, lengthOffset);
+
+            // 32 MB limit on message size from Bitcoin Core.
+            if (length > 0x02000000)
+                throw new FormatException("Message payload too big (over 0x02000000 bytes)");
```
dangershony: `FormatException`? maybe we need a `ProtocolViolationException`

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+        {
+            this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(offset), offset, nameof(bytesToRead), bytesToRead);
+
+            while (bytesToRead > 0)
+            {
+                int chunkSize = await this.Stream.ReadAsync(buffer, offset, bytesToRead, cancellation).ConfigureAwait(false);
+                if (chunkSize == 0)
+                {
+                    this.logger.LogTrace("(-)[STREAM_END]");
+                    throw new OperationCanceledException();
```
dangershony: This `OperationCanceledException` will disconnect the node no?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>
+        /// Reads a raw binary message from the connection stream and formats it to a structured message.
+        /// </summary>
+        /// <param name="protocolVersion">Version of the protocol that defines the message format.</param>
+        /// <param name="cancellation">Cancellation token that allows aborting the read operation.</param>
+        /// <returns>Binary message received from the connected counterparty.</returns>
+        /// <exception cref="OperationCanceledException">Thrown if the operation was cancelled or the end of the stream was reached.</exception>
```
dangershony: This exceptions are not directly thrown from this method, but fine with me.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Ah or is that your intention..

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+        /// Seeks and reads the magic value from the connection stream.
+        /// </summary>
+        /// <param name="magic">Magic value that starts the message.</param>
+        /// <param name="cancellation">Cancellation token that allows aborting the read operation.</param>
+        /// <exception cref="OperationCanceledException">Thrown if the operation was cancelled or the end of the stream was reached.</exception>
+        /// <remarks>
+        /// Each networkm message starts with the magic value. If the connection stream is in unknown state,
+        /// the next bytes to read might not be the magic. Therefore we read from the stream until we find the magic value.
+        /// </remarks>
+        public async Task ReadMagicAsync(byte[] magic, CancellationToken cancellation)
```
dangershony: This method came from Network right? so we essentially have duplication of code?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -107,11 +116,11 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
         /// <param name="data">Data to send.</param>
         /// <param name="cancellation">Cancellation token that allows aborting the operation.</param>
         /// <exception cref="OperationCanceledException">Thrown when the connection was terminated or the cancellation token was cancelled.</exception>
-        public async Task SendAsync(byte[] data, CancellationToken cancellation)
+        public async Task SendAsync(byte[] data, CancellationToken cancellation = default(CancellationToken))
```
dangershony: Same comment about the cancellation

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: yes, yes

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: yes

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: ok i can change it in later PR

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Actually I take this back leave the cancellation here.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
```diff
@@ -145,6 +154,148 @@ public NetworkPeerClient(int Id, TcpClient tcpClient, ILoggerFactory loggerFacto
...
+        /// </summary>
+        /// <param name="buffer">Buffer to read incoming data to.</param>
+        /// <param name="offset">Position in the buffer where to write the data.</param>
+        /// <param name="bytesToRead">Number of bytes to read.</param>
+        /// <param name="cancellation">Cancellation token that allows aborting the read operation.</param>
+        /// <returns>Binary data received from the connected counterparty.</returns>
+        /// <exception cref="OperationCanceledException">Thrown if the operation was cancelled or the end of the stream was reached.</exception>
+        private async Task ReadBytesAsync(byte[] buffer, int offset, int bytesToRead, CancellationToken cancellation = default(CancellationToken))
+        {
+            this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(offset), offset, nameof(bytesToRead), bytesToRead);
```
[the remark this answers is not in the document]
Aprogiena: and if there will be a problem somewhere the first thing i'll do is add them back ...
these are for debugging

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
Aprogiena: ok, i remove it in later PR

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -168,28 +168,29 @@ public void Listen(int maxIncoming = 20)
...
-                {
-                    if (e is OperationCanceledException) this.logger.LogDebug("Shutdown detected, stop accepting connections.");
-                    else this.logger.LogDebug("Exception occurred: {0}");
 
-                    break;
+                    // This should be cheaper for the accept loop thread than just calling ProcessNewClientAsync without awaiting.
+                    Task unused = Task.Run(async () => await this.ProcessNewClientAsync(client));
                 }
             }
+            catch (OperationCanceledException)
```
dangershony: Changed your mind on casting?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -168,28 +168,29 @@ public void Listen(int maxIncoming = 20)
...
+            {
+                this.logger.LogDebug("Shutdown detected, stop accepting connections.");
+            }
+            catch (Exception e)
+            {
+                this.logger.LogDebug("Exception occurred: {0}", e.ToString());
```
dangershony: `LogError` perhaps?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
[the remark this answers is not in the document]
dangershony: yes but it will be called on each byte passing in the network, you will probably get good stats from the caller in any case.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerClient.cs
dangershony: Hmm wont it be better to improve the code in Network or better yet create a class that lives on network that has the functionality `NetowrkStreamReader`?
Later PR is fine.

Aprogiena: i am not sure this even belongs to network, you might argue that if you created completely different protocol then it would be part of the network ...

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: i am still not sure, it seems like it is possible that the socket will be disposed and it would cause the exception, this code is still under review

PR #915, opened 2017-12-14
Added IsPropagated flag to wallet transaction data and changed how SendTransaction works. 

noescape00:
Changes: 

1) Added requirement of having at least 1 connection in order to send a tx.
2) Removed waiting for propagation before returning result of SendTransaction.
3) Now it's possible to check propagation state of the transaction by looking it up in the wallet and using IsPropagated flag. 
4) Changed SendTransactionAsync to SendTransaction because now this method do not use any awaitable methods. 

Those changes were discussed in [issue 597](https://github.com/stratisproject/StratisBitcoinFullNode/issues/597) so for more info about the logic behind this changes take a look at the discussion. 

Fix: #579

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
@@ -3,26 +3,31 @@ using System.Linq;
 using System.Threading.Tasks;
 using ConcurrentCollections;
 using NBitcoin;
+using Stratis.Bitcoin.Broadcasting;
 using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Features.Wallet.Interfaces;
 using Stratis.Bitcoin.Interfaces;
-using Stratis.Bitcoin.P2P.Peer;
 using Stratis.Bitcoin.P2P.Protocol.Payloads;
 using Stratis.Bitcoin.Utilities;
 
-namespace Stratis.Bitcoin.Broadcasting
+namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
 {
     public abstract class BroadcasterManagerBase : IBroadcasterManager
     {
         public event EventHandler<TransactionBroadcastEntry> TransactionStateChanged;
 
-        /// <summary>Connection manager for managing node connections.</summary>
+        /// <summary> Connection manager for managing node connections.</summary>
         protected readonly IConnectionManager connectionManager;
 
-        public BroadcasterManagerBase(IConnectionManager connectionManager)
+        /// <summary> Wallet manager.</summary>
+        protected readonly IWalletManager walletManager;
+
+        public BroadcasterManagerBase(IConnectionManager connectionManager, IWalletManager walletManager)
         {
             Guard.NotNull(connectionManager, nameof(connectionManager));
 
             this.connectionManager = connectionManager;
+            this.walletManager = walletManager;
             this.Broadcasts = new ConcurrentHashSet<TransactionBroadcastEntry>();
         }
 
@@ -57,9 +62,11 @@ namespace Stratis.Bitcoin.Broadcasting
                 broadcastEntry.State = state;
                 this.OnTransactionStateChanged(broadcastEntry);
             }
+
+            this.walletManager.ProcessTransaction(transaction, null, null, state == State.Propagated);
         }
 
-        public abstract Task<bool> TryBroadcastAsync(Transaction transaction);
+        public abstract void BroadcastTransaction(Transaction transaction);
 
         /// <summary>
         /// Sends transaction to peers.
--- a/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
@@ -60,14 +60,14 @@ namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
             }
         }
 
-        public abstract void BroadcastTransaction(Transaction transaction);
+        public abstract Task BroadcastTransactionAsync(Transaction transaction);
 
         /// <summary>
         /// Sends transaction to peers.
         /// </summary>
         /// <param name="transaction">Transaction that will be propagated.</param>
         /// <param name="skipHalfOfThePeers">If set to <c>true</c> transaction will be send to all the peers we are connected to. Otherwise it will be sent to half of them.</param>
-        protected void PropagateTransactionToPeers(Transaction transaction, bool skipHalfOfThePeers = false)
+        protected async Task PropagateTransactionToPeersAsync(Transaction transaction, bool skipHalfOfThePeers = false)
         {
             this.AddOrUpdate(transaction, State.ToBroadcast);
 
@@ -77,7 +77,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
             int propagateToCount = skipHalfOfThePeers ? (int)Math.Ceiling(peers.Count / 2.0) : peers.Count;
 
             for (int i = 0; i < propagateToCount; ++i)
-                peers[i].SendMessageAsync(invPayload).GetAwaiter().GetResult();
+                await peers[i].SendMessageAsync(invPayload);
         }
 
         protected bool IsPropagated(Transaction transaction)
--- a/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
@@ -17,17 +17,13 @@ namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
         public event EventHandler<TransactionBroadcastEntry> TransactionStateChanged;
 
         /// <summary> Connection manager for managing node connections.</summary>
-        protected readonly IConnectionManager connectionManager;
+        private readonly IConnectionManager connectionManager;
 
-        /// <summary> Wallet manager.</summary>
-        protected readonly IWalletManager walletManager;
-
-        public BroadcasterManagerBase(IConnectionManager connectionManager, IWalletManager walletManager)
+        public BroadcasterManagerBase(IConnectionManager connectionManager)
         {
             Guard.NotNull(connectionManager, nameof(connectionManager));
 
             this.connectionManager = connectionManager;
-            this.walletManager = walletManager;
             this.Broadcasts = new ConcurrentHashSet<TransactionBroadcastEntry>();
         }
 
@@ -62,8 +58,6 @@ namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
                 broadcastEntry.State = state;
                 this.OnTransactionStateChanged(broadcastEntry);
             }
-
-            this.walletManager.ProcessTransaction(transaction, null, null, state == State.Propagated);
         }
 
         public abstract void BroadcastTransaction(Transaction transaction);
--- a/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
@@ -23,22 +23,18 @@ namespace Stratis.Bitcoin.Features.Wallet.Broadcasting
         }
 
         /// <inheritdoc />
-        public override void BroadcastTransaction(Transaction transaction)
+        public override async Task BroadcastTransactionAsync(Transaction transaction)
         {
-            if (transaction == null)
-                throw new ArgumentNullException(nameof(transaction));
+            Guard.NotNull(transaction, nameof(transaction));
 
             if (this.IsPropagated(transaction))
                 return;
 
             var state = new MempoolValidationState(false);
-            if (!this.mempoolValidator.AcceptToMemoryPool(state, transaction).GetAwaiter().GetResult())
-            {
+            if (!await this.mempoolValidator.AcceptToMemoryPool(state, transaction))
                 this.AddOrUpdate(transaction, State.CantBroadcast);
-                return;
-            }
-
-            this.PropagateTransactionToPeers(transaction, true);
+            else
+                await this.PropagateTransactionToPeersAsync(transaction, true);
         }
     }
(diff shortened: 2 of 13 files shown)


bokobza:
I'll have a look.

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -3,26 +3,31 @@
...
     {
         public event EventHandler<TransactionBroadcastEntry> TransactionStateChanged;
 
-        /// <summary>Connection manager for managing node connections.</summary>
+        /// <summary> Connection manager for managing node connections.</summary>
         protected readonly IConnectionManager connectionManager;
 
-        public BroadcasterManagerBase(IConnectionManager connectionManager)
+        /// <summary> Wallet manager.</summary>
+        protected readonly IWalletManager walletManager;
```
bokobza: Can you make this and the connectionManager above, private?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -798,6 +798,12 @@ public class TransactionData
         [JsonProperty(PropertyName = "hex", NullValueHandling = NullValueHandling.Ignore)]
         public string Hex { get; set; }
 
+        /// <summary>
+        /// Propagation state of this transaction.
+        /// </summary>
+        [JsonProperty(PropertyName = "isPropagated", NullValueHandling = NullValueHandling.Ignore)]
```
bokobza: You don't need `NullValueHandling = NullValueHandling.Ignore` here as bool is never gonna be null, but false by default.
If you want it not to show in the json, use `bool?` with `NullValueHandling = NullValueHandling.Ignore` so if you don't set it, it's not serialized.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -743,7 +743,7 @@ public void ProcessBlock(Block block, ChainedBlock chainedBlock)
         }
 
         /// <inheritdoc />
-        public void ProcessTransaction(Transaction transaction, int? blockHeight = null, Block block = null)
+        public void ProcessTransaction(Transaction transaction, int? blockHeight = null, Block block = null, bool? isPropagated = null)
```
bokobza: I would make this non-null and `true` by default (`bool isPropagated = true`).
Most of the transactions going through this method will come from blocks, where `isPropagated == true`. Then they will mostly come from other people's transactions appearing on the network, where `isPropagated == true`.
Only the transactions that we create will be `false` at first, then `true`.

This way, you only actively set isPropagated in one place, and it's to `false`, when you send a transaction.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -763,7 +763,8 @@ public void ProcessTransaction(Transaction transaction, int? blockHeight = null,
                     // Check if the outputs contain one of our addresses.
                     if (this.keysLookup.TryGetValue(utxo.ScriptPubKey, out HdAddress pubKey))
                     {
-                        this.AddTransactionToWallet(transaction.ToHex(), hash, transaction.Time, transaction.IsCoinStake, transaction.Outputs.IndexOf(utxo), utxo.Value, utxo.ScriptPubKey, blockHeight, block);
+                        this.AddTransactionToWallet(transaction.ToHex(), hash, transaction.Time, transaction.IsCoinStake, transaction.Outputs.IndexOf(utxo),
+                            utxo.Value, utxo.ScriptPubKey, blockHeight, block, isPropagated ?? block != null);
```
bokobza: Following on from my previous comment, you can now remove the condition `isPropagated ?? block != null` and just use `isPropagated`.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -819,8 +820,9 @@ public void ProcessTransaction(Transaction transaction, int? blockHeight = null,
         /// <param name="blockHeight">Height of the block.</param>
         /// <param name="block">The block containing the transaction to add.</param>
         /// <param name="transactionHex">The hexadecimal representation of the transaction.</param>
+        /// <param name="isPropagated">Propagation state of the transaction.</param>
         private void AddTransactionToWallet(string transactionHex, uint256 transactionHash, uint time, bool isCoinStake, int index, Money amount, Script script,
-            int? blockHeight = null, Block block = null)
+            int? blockHeight = null, Block block = null, bool isPropagated = false)
```
bokobza: I think true is better here as discussed above.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -878,6 +881,11 @@ public void ProcessTransaction(Transaction transaction, int? blockHeight = null,
                 {
                     foundTransaction.MerkleProof = new MerkleBlock(block, new[] { transactionHash }).PartialMerkleTree;
                 }
+
+                if (foundTransaction.IsPropagated != isPropagated)
```
bokobza: isPropagated cannot go from true to false. So you can just check 
```
if(isPropagated) 
{
    foundTransaction.IsPropagated = true;
}
```

src/Stratis.Bitcoin/Stratis.Bitcoin.csproj
```diff
@@ -17,6 +17,10 @@
     <Version>1.0.6-alpha</Version>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
   </PropertyGroup>
+
+  <ItemGroup>
```
bokobza: Where did that come from? lol

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -145,7 +145,7 @@ private void BroadcasterManager_TransactionStateChanged(object sender, Transacti
         {
             if (transactionEntry.State == State.Propagated)
             {
-                this.ProcessTransaction(transactionEntry.Transaction);
+                this.ProcessTransaction(transactionEntry.Transaction, null, null, true);
```
bokobza: This condition shouldn't matter really as the same transaction can be processed as many times as we want but ProcessTransaction is idempotent so it won't change the state of the wallet.
So, with this in mind we can:
1. remove this condition and just call ProcessTransaction
2. review what happens in AddOrUpdate, because we can always ProcessTransaction, irrespective of state.
3. rethink whether keeping a list of broadcasts and the AddOrUpdate method in BroadcastManagerBase are useful (although at this stage I'm happy to leave them there for now)

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
```diff
@@ -57,9 +62,11 @@ public void AddOrUpdate(Transaction transaction, State state)
                 broadcastEntry.State = state;
                 this.OnTransactionStateChanged(broadcastEntry);
             }
+
+            this.walletManager.ProcessTransaction(transaction, null, null, state == State.Propagated);
```
bokobza: Looking at my comment inside `BroadcasterManager_TransactionStateChanged`,  I don't think you need the walletManager here, as the OnTransactionStateChanged method can probably be called every time we have a new transaction so yo udon't need to call ProcessTransaction from here.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: Actually here, instead of `true` by default, you can have null by default, just like here, and you can set false manually when necessary.
Then you make the isPropagated field `bool?` in TransactionData. This way, if it's false, we can see it in the object, if it's propagated, the field disappears (as it's not really needed anymore).

noescape00:
@bokobza thanks for an awesome review! 
Will fix everything you've mentioned as soon as I'm home in ~4 hours (or if we're in a hurry I can do it from the laptop but you'll need to test it since I don't have my wallets for testing sending the tx on the testnet and the mainnet after changes are applied).

bokobza:
If you can and want to start, have a go, even on your laptop.
I can give you a wallet if you want 🤑

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
noescape00: Hmm. That actually makes my last commit unnecessary since it was about adding wallet manager and fixing dependency loop that was caused by it. 
But yeah, if we just set IsPropagated = true for every incoming tx it makes sense to remove wallet manager from  BroadcasterManagerBase.cs completely.

noescape00:
>If you can and want to start, have a go, even on your laptop.

Sure. I'm starting from the laptop and I'm managing to get home in 1-2 hours so I'll be able to test with my wallets =).

bokobza:
Great, thanks man.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
noescape00: >This condition shouldn't matter really as the same transaction can be processed as many times as we want but ProcessTransaction is idempotent so it won't change the state of the wallet.

Can you please elaborate? 
It looks like WalletManager.ProcessTransaction will call AddTransactionToWallet which will create new tx in the wallet OR update existing one. So if the only thing that changed is .IsPropagated flag then it's the only thing that will be updated.

bokobza: Exactly, what I'm saying is that ProcessTransaction can be called multiple times, and if there is a transaction that exists already in the wallet, it'll just update it.
so here, whether `if (transactionEntry.State == State.Propagated)` is true or false, we can call ProcessTransaction anyway.

noescape00: Makes sense, thanks!

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
noescape00: So in this case not having the IsPropagated flag will mean that it is propagated... 

Hmm. From a logical perspective I'd rather have IsNotPropagated that will be null and removed from the serialized file as soon as it's propagated. 
Are you ok with that?

noescape00: And on top we'd have a method IsPropagated in TransactionData that will check if .IsNotPropagated null or not and return true if it is.

noescape00: But I'd actually just leave IsPropagated there and not delete it after it's true. 
Removing it will make it a bit more difficult to understand the code for other devs.

noescape00:
I've reverted that commit that would move the stuff around (it was done jsut to add that wallet manager where it wasn't necessary) and pushed the last 3 with force. 

Everything else was also done basing on your comments except I've left 1 thing differently: IsPropagated is nullable (for back compatibility) but we do not set it null when it's propagated to save a few bits of disk space. 

I've tested it and it's all works as expected (when we have 2 or more connections we get IsPropagated = true when it's propagated and before it's included to the block and when we have 1 connection both IsPropagated and IsConfirmed get true at the same time because we set propagation state when we get the block with our tx)

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
```diff
@@ -22,24 +23,22 @@ public FullNodeBroadcasterManager(IConnectionManager connectionManager, IMempool
...
             if (transaction == null)
                 throw new ArgumentNullException(nameof(transaction));
 
             if (this.IsPropagated(transaction))
-                return true;
+                return;
 
             var state = new MempoolValidationState(false);
-            if (!await this.mempoolValidator.AcceptToMemoryPool(state, transaction).ConfigureAwait(false))
+            if (!this.mempoolValidator.AcceptToMemoryPool(state, transaction).GetAwaiter().GetResult())
```
bokobza: why is this not async anymore?

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
```diff
@@ -156,7 +156,8 @@ public interface IWalletManager
         /// <param name="transaction">The transaction.</param>
         /// <param name="blockHeight">The height of the block this transaction came from. Null if it was not a transaction included in a block.</param>
         /// <param name="block">The block in which this transaction was included.</param>
-        void ProcessTransaction(Transaction transaction, int? blockHeight = null, Block block = null);
+        /// <param name="isPropagated">Transaction propagation state. Null if unknown.</param>
```
bokobza: forgot to remove `Null if unknown.`

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/FullNodeBroadcasterManager.cs
noescape00: Because now BroadcastTransaction method don't have a loop that would wait for tx to be propagated back so I've changed BroadcastTransactionAsync to BroadcastTransaction. And the only async method inside the BroadcastTransaction that was left is AcceptToMemoryPool so I looked at the only usage of BroadcastTransaction and we'd await it anyway so I changed it to be awaited inside of BroadcastTransaction by running it synchronously.

src/Stratis.Bitcoin.Features.Wallet/Broadcasting/LightWalletBroadcasterManager.cs
```diff
@@ -16,30 +16,15 @@ public LightWalletBroadcasterManager(IConnectionManager connectionManager) : bas
         }
 
         /// <inheritdoc />
-        public override async Task<bool> TryBroadcastAsync(Transaction transaction)
+        public override void BroadcastTransaction(Transaction transaction)
         {
             if (transaction == null)
```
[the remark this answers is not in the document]
noescape00: Fixed

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
noescape00: Fixed

PR #916, opened 2017-12-14
[P2P] Improve peer selection part 2

fassadlr:
This is the next installment in the peer selection series:

- The PeerSelector class now only takes in the ConcurrentDictionary of peers managed by the PeerAddressManager.
- Replaced `IsNew` peer property with `Fresh` to indicate that the peer has never had a connection attempt.
- Added Connected/Attemped/Handshaked properties to PeerAddress
- The `Fresh()` peer collection extension replaces `New()`
- Added a `Handshaked` peer collection extension
- Added subtle throttling to the peer collection extensions to filters out peers where the collection condition was satisfied within the last 60 seconds.
- Added a condition to the attempted peers where peers are filtered that have been attempted more than 10 times.


src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -30,70 +30,99 @@ public interface IPeerSelector
...
+                return this.peerAddresses.Fresh().Random();
 
             return null;
         }
 
         /// <inheritdoc/>
         public IEnumerable<PeerAddress> SelectPeers()
         {
-            return this.peerAddressManager.Peers.Where(p => p.Value.Preferred).Select(p => p.Value);
+            return this.peerAddresses.Where(p => p.Value.Preferred).Select(p => p.Value);
```
dangershony: Ho do we determine `Preferred`?

fassadlr: Hi @dangershony , these will all be removed/replaced in the last PR as well as properly documented. I'm just busy "setting" up for the last logic change 👍

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -30,70 +30,99 @@ public interface IPeerSelector
...
+            this.peerAddresses = peerAddresses;
         }
 
         /// <inheritdoc/>
         public PeerAddress SelectPeer()
         {
-            if (this.peerAddressManager.Peers.Tried().Any() == true &&
-                (this.peerAddressManager.Peers.New().Any() == false || new Random().Next(2) == 0))
-                return this.peerAddressManager.Peers.Tried().Random();
+            var tried = this.peerAddresses.Attempted().Concat(this.peerAddresses.Connected());
```
dangershony: Can you simplify this?
We could break the logic apart so its more readable. 

```
var previouslyTreid = this.peerAddresses.Attempted().Concat(this.peerAddresses.Connected())
var neverTried = this.peerAddresses.Fresh()
var randomfactor =  this.random().Next(2) == 0
var neverTriedIsEmpty = !neverTried .Any();
var previouslyTreidNotEmpty = neverTried .Any();

if(previouslyTreidNotEmpty && (neverTriedIsEmpty || randomfactor))
      return tried.Random();

etc...
```

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -30,70 +30,99 @@ public interface IPeerSelector
...
+        /// The result filters out peers which satisfies the above condition within the 
+        /// last 60 seconds and that has had more than 10 failed attempts.
+        /// </para>
         /// </summary>
-        public static IEnumerable<PeerAddress> New(this ConcurrentDictionary<IPEndPoint, PeerAddress> peers)
+        public static IEnumerable<PeerAddress> Attempted(this ConcurrentDictionary<IPEndPoint, PeerAddress> peers)
         {
-            var isNew = peers.Skip(0).Where(p => p.Value.IsNew).Select(p => p.Value);
-            return isNew;
+            var result = peers.Skip(0).Where(p =>
```
dangershony: I think we should also order the list so that more recent attempts are not in the top.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: @dangershony sorry, this will also be dropped with next release...

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: @dangershony sorry, this will also be dropped with next release...

PR #919, opened 2017-12-14
Refactor/the node job part 14

Aprogiena:
Now we get async hardcore, we change two fundamental methods to be fully async - VersionHandshakeAsync and RespondToHandShakeAsync. We also try to "propagate" async-ness to wider code. We can't reasonably convert whole codebase to async in one PR, but what could quickly be done is done here.

We also removed SentMessage from code base (its use was removed in part 13).

The network peer receive task is now fully async too - ReceiveMessagesAsync. non-async SendMessage gone, OnDisconnect handler gone (can be added later if needed, but currently no one is using it, everyone uses detach core handler). CreateListener method gone.


bokobza:
Which commit does it start with?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
```diff
@@ -40,15 +40,18 @@ public NetworkPeerListener Where(Func<IncomingMessage, bool> predicate)
             Queue<IncomingMessage> pushedAside = new Queue<IncomingMessage>();
             try
             {
-                while (true)
+                using (CancellationTokenSource cancellation = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, this.Peer.Connection.Cancel.Token))
                 {
-                    IncomingMessage message = ReceiveMessage(CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, this.Peer.Connection.Cancel.Token).Token);
-                    if (this.predicates.All(p => p(message)))
+                    while (true)
```
dangershony: perhaps?
``` 
while (true)
{
       cancellation.ThrowIfCanceled() ;
```

Aprogiena:
i believe 
64f5730

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
Aprogiena: this is unchanged code and I think it just relies on ReceiveMessage throwing
and that's probably correct assumption, RM here would block, so it is its responsibility, such cancellation.throw  would not help us anything i believe, but this will be rewritten later anyway, but should work perfectly as is

PR #923, opened 2017-12-14
Refactor/the node job part 15

### LINKED ISSUE #913: [NetworkPeerClient] ConnectAsync timeout not working properly
Hi @Aprogiena, I've noticed that the timeout that gets set in the PeerConnector's ConnectAsync method isn't getting honoured:

```
    var clonedConnectParamaters = this.CurrentParameters.Clone();
    timeoutTokenSource.CancelAfter(5000);
    clonedConnectParamaters.ConnectCancellation = timeoutTokenSource.Token;

    peer = this.networkPeerFactory.CreateConnectedNetworkPeer(this.network, peerAddress.NetworkAddress, clonedConnectParamaters);
    peer.VersionHandshake(this.Requirements, timeoutTokenSource.Token);
```

It takes about 20secs normally for the connect to timeout. I had a look at the connect code in NetworkPeerClient's ConnectAsync method. I changed this line:

```
await Task.Run(async () => await this.tcpClient.ConnectAsync(endPoint.Address, endPoint.Port), cancellation);
```

to:

```
this.tcpClient.ConnectAsync(endPoint.Address, endPoint.Port).Wait(cancellation);
```

And it instantly picked up the correct timeout. I know it makes the call synchronous but can you please look at this asap as we can't have the node taking 20secs to timeout before moving onto the next peer to connect to.

Aprogiena:
This is mostly polishing episode
- various small changes and improvements requested in previous PRs
- probably most importantly - cancellation tokens were ignored in accept and connect methods, now this is fixed

Fix #913


Aprogiena:
i don't know which commit is the first here, just check the files 
anyway, there are no more parts in the queue, so the next one after the release will be clean!

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
```diff
@@ -70,11 +70,11 @@ public interface INetworkPeerFactory
...
-        /// <param name="localEndPoint">IP address and port to listen on, or <c>null</c> to listen on all available interfaces and default port.</param>
-        /// <param name="externalEndPoint">IP address and port that the server is reachable from the Internet on, or <c>null</c> to use the same value as <paramref name="localEndPoint"/>.</param>
+        /// <param name="localEndPoint">IP address and port to listen on.</param>
+        /// <param name="externalEndPoint">IP address and port that the server is reachable from the Internet on.</param>
         /// <param name="version">Version of the network protocol that the server should run.</param>
         /// <returns>Newly created network peer server, which is ready to be started.</returns>
-        NetworkPeerServer CreateNetworkPeerServer(Network network, IPEndPoint localEndPoint = null, IPEndPoint externalEndPoint = null, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION);
+        NetworkPeerServer CreateNetworkPeerServer(Network network, IPEndPoint localEndPoint, IPEndPoint externalEndPoint, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION);
```
bokobza: Is that the fix for the bug where a node would connect to itself?

Aprogiena: No, there was a wrong merge once and due to these default values it was unnoticed. So it was suggested that we remove the default values because we don't use them anyway. In my view, the default values did make sense and the initialization of such object was reasonable, but it is true that currently, it is of no use, so I was OK to remove them.

dangershony:
This looks fine to me

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
bokobza: cool thanks

bokobza:
Well done @Aprogiena .

PR #925, opened 2017-12-15
Excepton thrown if no LastBlockSyncHeight

bokobza:
There was a bug who's repro steps are the following:
In a normal situation:
1. user starts off with new installation (no folder, no blocks, and **no chain**)
2. user starts the node
3. user creates a wallet, before the chain is downloaded
4. wallet is created but LastBlockSyncHeight and LastBlockSyncHash are not yet populated (as per design, we don't know anything about the chain yet)
5. when chain is downloaded, LastBlockSyncHeight and LastBlockSyncHash are updated

If between 4. and 5., the node is stopped, the wallet won't have LastBlockSyncHeight and LastBlockSyncHash on restart, leading to an exception.
The fix presented here waits for the chain to be downloaded in the background before updating the LastBlockSyncHeight and LastBlockSyncHash in the wallets.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
@@ -233,7 +233,7 @@ namespace Stratis.Bitcoin.Features.Wallet
             }
             else
             {
-                this.UpdateWhenChainDownloaded(wallet, DateTime.Now);
+                this.UpdateWhenChainDownloaded(new[] { wallet }, DateTime.Now);
             }
 
             // Save the changes to the file and add addresses to be tracked.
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -609,7 +609,15 @@ public uint256 LastReceivedBlockHash()
...
                     .FirstOrDefault()?.LastBlockSyncedHash;
 
-                Guard.Assert(lastBlockSyncedHash != null);
+                // If details about the last block synced are not present in the wallet,
+                // find out which is the oldest wallet and set the last block synced to be the one at this date.
+                if (lastBlockSyncedHash == null)
+                {
+                    this.logger.LogWarning("There were no details about the last block synced in the wallets.");
+                    DateTimeOffset earliestWalletDate = this.Wallets.Min(c => c.CreationTime);
+                    this.UpdateWhenChainDownloaded(this.Wallets, earliestWalletDate.DateTime);
```
fassadlr: What happens if there are no wallets? Or the list is empty?

bokobza: a bit above this code, we check if there are no wallets, in which case we return chain.tip.

fassadlr: Ahhh sorry yes!

PR #929, opened 2017-12-15
Bugfix/issue 924b

### LINKED ISSUE #924: [NetworkPeerServer] Shutdown sometimes does not complete
This is just note to self, I know exactly where the problem is, just don't know what it is. I do have logs to investigate.

Aprogiena:
This is technical part 16 of the node job, but never mind :-)

This does not really fixes #924 because since the node job PR part 15 was merged such behavior has not been seen, but I've revealed what I mentioned in a comment of #924, which is leaking unfinished clients. 

This had to do with incomplete shutdown and cleanup sequence of the client.

So in this PR we 
- make couple of minor improvements in comments, logs, renaming etc.
- change the shutdown sequence of the peer clients
- and thus fix the connected client leak
- replace disconnected event with slim version
- network peer loggers are prefixed with network client ID which is very handy while debugging issues like those we fixed in this PR

We do not yet document this as the end goal is to merge NPConnection class with NPClient class into one class and outside of NP.cs file. So it will change again, but should now be done correctly.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -418,12 +418,15 @@ namespace Stratis.Bitcoin.P2P.Peer
             this.acceptTask.Wait();
 
             ICollection<NetworkPeerClient> connectedClients = this.clientsById.Values;
-            this.logger.LogInformation("Waiting for {0} connected clients to finish.", connectedClients.Count);
-            foreach (NetworkPeerClient client in connectedClients)
+            if (connectedClients.Count > 0)
             {
-                TaskCompletionSource<bool> completion = client.ProcessingCompletion;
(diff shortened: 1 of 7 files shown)


Aprogiena:
not to be merged yet

Aprogiena:
and we should be good to go

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -210,6 +210,7 @@ private void AddConnectedClient(NetworkPeerClient client)
             this.logger.LogTrace("({0}.{1}:{2})", nameof(client), nameof(client.Id), client.Id);
 
             this.clientsById.AddOrReplace(client.Id, client);
+            client.ProcessingCompletion.Task.ContinueWith(unused => this.RemoveConnectedClient(client));
```
bokobza: nice

PR #937, opened 2017-12-17
Staking in History

bokobza:
In order to be able to show a staking transaction in the History, we look for transactions that are marked as coin stake transaction + that have 1 input from our wallet to 2 outputs to our wallet.

This is better reviewed commit by commit.
Any question, please ask.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Models/WalletHistoryModel.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Models/WalletHistoryModel.cs
@@ -82,6 +82,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Models
     public enum TransactionItemType
     {
         Received,
-        Send
+        Send,
+        Mined
     }
 }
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Wallet/Models/WalletHistoryModel.cs
```diff
@@ -82,6 +82,7 @@ public class PaymentDetailModel
     public enum TransactionItemType
     {
         Received,
-        Send
+        Send,
+        Mined
```
mikedennis: Should this be 'Staked' instead of 'Mined'? I know they are used interchangeably but I like to think of Stake as POS and Mine as POW to keep things straight or does this identify POW blocks as well?

bokobza: hmm you're right, although the QT says Mined.
Well, we're doing Staked! Thanks Mike.

mikedennis:
Looks all good to me! But maybe apro can find something 😀

bokobza:
Thanks. It works well enough for now. I'll push it so the front end can work with it and worst case I'll issue a patch.

Issue #690, opened 2017-11-03
[Settings] -rpcport stopped to work


Aprogiena:
Just trying to spawn 2 instances on 1 machine and `-rpcport` on cmd line or in config does not make the rpc server to go elsewhere, still using default port.

D3m0nKingx:
Confirmed here as well, RPC port stays at the default 26174 on testnet

Aprogiena:
the whole rpc settings is very weird imho

1. rpc settings are respected only if `-server=1`, which makes no sense
2. there is this warning `WARNING: RPC bind selection (-rpcbind) was ignored because allowed ip's (-rpcallowip) were not specified, refusing to allow everyone to connect` which also makes little sense to me and is not really compatible with how bitcoin core works (or did they change it recently?)

dangershony:
No I think core is like that if you don't specify allowed ip RPC will not work (if I remember correctly)
Perhaps we need to just remove the server=1 configuration, this was there for bitcoin compatibility but I guess it was there in bitcoin to disable RPC, our node is modular based so we can just not use the RPC feature.

Aprogiena:
I think @D3m0nKingx was testing sX and Bitcoin Core and found out that it works differently. @D3m0nKingx could you add more?

But if you look at it, it is quite counterintuitive how is it now. Not very user friendly.

bokobza:
@noescape00 can we close this?

noescape00:
@bokobza sure.

PR #941, opened 2017-12-17
Rpc port setting fix

noescape00:
This fix allows specifying rpc port without enabling '-server=true'.

This is needed in order to start multiple nodes on the same machine without the need to enable -server.


Aprogiena:
which issue this fixes?

noescape00:
There is no issue for that. 
A few weeks ago it was possible to run several nodes on one machine but then rpc settings got refactored and now in order to change the RPC port you need to enable -server option.

Aprogiena:
there were about 2-3 issues for this thing

noescape00:
@Aprogiena you're right. 

[This one](https://github.com/stratisproject/StratisBitcoinFullNode/issues/690) looks like the one this PR solves.

PR #942, opened 2017-12-17
Improved console text alignment for the node stats

noescape00:
It was just a bit annoying to see the wrong padding.

stats - copy

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
@@ -97,9 +97,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
             var highestBlock = (this.blockRepository as BlockRepository)?.HighestPersistedBlock;
 
             if (highestBlock != null)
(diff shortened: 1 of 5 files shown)


bokobza:
I don't really see what's "wrong" about the previous padding. 🤔 
All I can see here is that you've made the text a bit more condensed. Is that right, or have I missed something?

noescape00:
@bokobza the lines that have hashes in them were too long for the previous formatting so that caused splitting so the leftover of the hash would go to the next line. 

34082819-c47d3996-e376-11e7-976f-a519ba5b069f

Ofc this is only true for the default console settings (lines are 80 characters long) for both linux and windows.

Aprogiena:
is anyone using console in default?

D3m0nKingx:
As in standard terminal settings? for the most part I keep mine at default, and just adjust the height/width of the window with the cursor to see the output. Also I can zoom out as well in the terminal (dunno if that's smth doable in windows tho).

bokobza:
I use cmder, which is really cool.

bokobza:
but I'm happy for this to be in.

PR #943, opened 2017-12-17
Added a bit more logic to History

bokobza:
- transactionData that are outputs of stake transactions are excluded
- after spending details for staking transactions are processed, we keep processing in case the transaction being spent itself is not a stake transaction.

In any case, I think the logic in History need to be reviewed and simplified.
The way we currently decide whether a transaction should be shown or not could be made more flat and processed on a case-by-case basis, as opposed to the way we do it now, which feels a bit like we trying to "figure it out all at once".

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
@@ -331,7 +331,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Controllers
                 // Represents a sublist of transactions associated with receive addresses + a sublist of already spent transactions associated with change addresses.
                 // In effect, we filter out 'change' transactions that are not spent, as we don't want to show these in the history.
                 List<FlatHistory> history = items.Where(t => !t.Address.IsChangeAddress() || (t.Address.IsChangeAddress() && !t.Transaction.IsSpendable())).ToList();
-                
+
                 // Represents a sublist of 'change' transactions.
                 List<FlatHistory> allchange = items.Where(t => t.Address.IsChangeAddress()).ToList();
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -340,6 +340,12 @@ public IActionResult GetHistory([FromQuery] WalletHistoryRequest request)
                     var transaction = item.Transaction;
                     var address = item.Address;
 
+                    // We don't show in history transactions that are outputs of staking transactions.
+                    if (transaction.IsCoinStake != null && transaction.IsCoinStake.Value && transaction.SpendingDetails == null)
```
mikedennis: I wonder if this might be more readable if you put some of these boolean expressions as properties in FlatHistory. Same with the logic in line 333 above: ``` !t.Address.IsChangeAddress() || (t.Address.IsChangeAddress() && !t.Transaction.IsSpendable())```  Definitely can wait until a future PR after release.

noescape00:
This change sounds reasonable but I don't think it's a good idea to completely remove the possibility of getting the whole raw history. 
Maybe consider adding some flags to the GetHistory methods so by default it will return only the items you are likely to be interested in, but you could also include all that staking txes. 
Or keep it like this and add GetRawHistory method and just filter outputs of this method in the GetHistory.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza: Hmm yeah, very good idea.
Ultimately I'd like to build the history in a simple way, like that:
if (this output is a change output) continue;
if (this output is a receive but not staking) process;
if (this output is a receive and staking) continue;

rather than have quite a few conditions that are interlocked. 
Like that It's just way easier to understand what we want to show in the history.

bokobza:
I see what you're saying here @noescape00 but if we show the whole history, it won't make sense to almost no one.
What we thought about doing at some point though is the full transactions via a browser link maybe.
This way, users that want to see more will be able to do so.

Issue #921, opened 2017-12-14
[PosMinting] Node should wait until connected to synced node


Aprogiena:
Currently the node is happy if it sees just one connected peer. However, if that peer is in IBD, the node is happy to mine a block even if it is itself well behind the tip.

fassadlr:
@Aprogiena I can look at this issue?

We can potentially add a IBD flag in `NetworkPeerRequirement`, which the miner can then check?

Aprogiena:
discuss with @noescape00 i think he was looking into this

noescape00:
@fassadlr I haven't really started digging into it so I'm ok if you take this one. 

So far it looks like whole IBD needs to be refactored and added in a behavior that both peer and node has.

noescape00:
@fassadlr let me know if you'd like to take this one or I can.

fassadlr:
HI @noescape00, In #944 I started with some work on refactoring ChainState and it seems now that this issue is starting to look related to that one. I.e. the design behind IBD and IBlockDownloadState needs to be looked at again as there is quite a dependency graph that exists. @Aprogiena mentioned that the true provider of IBD is ConsensusLoop. See his comment:

https://github.com/stratisproject/StratisBitcoinFullNode/pull/944#issuecomment-352461605

So possibly if we solve this issue we can solve the issue with ChainState construction. Lets possibly look at this together?

noescape00:
>So possibly if we solve this issue we can solve the issue with ChainState construction. Lets look at this together?

Sure. 

So let'y try outlining what we would like to achieve: 

1) We want to make a provider class of IBlockDownloadState to be the same for our node and for our view or the peer's node- that way we can implement logic behind IBlockDownloadState just once and have the ability to know if we are in IBD or any of our peers is. 
2) We don't want to pass a whole ConsensusManager just to know if we\peer is in IBD. 

Did I miss anything so far, @fassadlr?

fassadlr:
> We want to make a provider class of IBlockDownloadState to be the same for our node and for our view or the peer's node- that way we can implement logic behind IBlockDownloadState just once and have the ability to know if we are in IBD or any of our peers is.
> We don't want to pass a whole ConsensusManager just to know if we\peer is in IBD.

SPOT on @noescape00! 🥇 

Once we have solved this dependency issue, we can remove the following dependencies:
1. FullNode dependency in ChainState
2. ChainState dependency in ConsensusManager

[...]

PR #944, opened 2017-12-18
[ChainState] Pass DateTimeProvider via dependency injection to constructor

### LINKED ISSUE #904: [ChainState] use normal DI, do not pull DTP from FullNode service
Such non-standard DI gives us trouble when we have testing environment with not fully initialized FullNode.

fassadlr:
This fixes #904 .


Aprogiena:
actually, could we remove the full node dependency completely and replace it with IBlockDownloadState dependency?

fassadlr:
> actually, could we remove the full node dependency completely and replace it with IBlockDownloadState dependency?

Even better!

PS do you know what IBlockDownloadState does? There's no documentation on that interface anywhere in the code :(

Aprogiena:
its only purpose is to provide IBD check
so it should be trivial to mock in tests

fassadlr:
@Aprogiena OK wow this send me down circular dependency hell!

In BaseFeature we configure:

`services.AddSingleton<ChainState>();`

In ConsensusFeature we configure IBlockDownloadState as:

```
services.AddSingleton<ConsensusManager>()
    .AddSingleton<IBlockDownloadState, ConsensusManager>()
    .AddSingleton<INetworkDifficulty, ConsensusManager>();
```

ConsensusManager takes in ChainState:

```
public ConsensusManager(ICheckpoints checkpoints, ConsensusLoop consensusLoop = null, IDateTimeProvider dateTimeProvider = null, NodeSettings nodeSettings = null, Network network = null,
            PowConsensusValidator consensusValidator = null, ChainState chainState = null)
```

If we change ChainState to take in IBlockDownloadState:

```
public ChainState(IBlockDownloadState blockDownloadState, IDateTimeProvider dateTimeProvider, IInvalidBlockHashStore invalidBlockHashStore)
{
```

Boom! We have a circular dependency! Urgh this is not good. Something seems off here.

I dont think its good design to ask services for ConsensusManager via IBlockDownloadState ??

fassadlr:
@Aprogiena either we revisit the design here in another PR, or we just do the quick fix for now and inject FullNode.

Aprogiena:
The situation might not be as bad as it seems.
How about removing ChainState from ConsensusManager?
it seems very possible

but maybe @dangershony would know if there is not some problem with that?

Aprogiena:
I think the only problem will be when there is no consensus, then we will need to implement that interface to return false, but that seems trivial as well

fassadlr:
ConsensusManager only use ChainState in getting the ConsensusTip here:
```
public Target GetNetworkDifficulty()
{
    if ((this.ConsensusValidator?.ConsensusParams != null) && (this.ChainState?.ConsensusTip != null))
        return this.ChainState?.ConsensusTip?.GetWorkRequired(this.ConsensusValidator.ConsensusParams);
    else
        return null;
}
```

Why dont we just pass ChainState to GetNetworkDifficulty?

Aprogiena:
why don't we use this.ConsensusTip instead of this.ChainState.ConsensusTip?

fassadlr:
This is crappy design... the dependency graph just keeps going deeper. Its not right to configure a DI container for something that is essentially just a polymorphic property. This is wrong IMO:

```
services.AddSingleton<ConsensusManager>()
            .AddSingleton<IBlockDownloadState, ConsensusManager>()
            .AddSingleton<INetworkDifficulty, ConsensusManager>();
```

What we are saying here is that if we ask Services for IBlockDownloadState, give us an instance of ConsensusManager?! We are only interested in IsInitialBlockDownload.

IMO ChainState should NOT be constructed with IBlockDownloadState but rather, we should just pass IBlockDownloadState to IsInitialBlockDownload() method. That way we remove the dependency.

fassadlr:
Very much an oversimplified idea, but I think the direction is better ;)

fassadlr:
The idea (and what the intent was) of the IBlockDownloadState interface is an excellent one 👍  Its just then we need to properly implement a provider so that we can get rid of these dependencies ;)

Aprogiena:
i don't know what you mean by passing IBlockDownloadState to IsInitialBlockDownload
to me it seems that the true provider of IBD is ConsensusLoop
and that ChainState is not needed inside ConsensusManager
it seems like we can get rid of the circular dependency easily
but if you think further improvements are needed in the design, feel free to try that, but i don't understand what you mean

fassadlr:
@Aprogiena @noescape00 I'm closing this PR so that we can continue discussions in #921 and subsequently do a PR from that issue.

PR #947, opened 2017-12-18
Remove global.json and fix compiler warnings

### LINKED ISSUE #933: Support building with latest .net sdk (2.1.2)
If you only install the latest .net SDK (by itself) you can't build the full node. You either have to go back and install 2.0 or bump the version in global.json to the installed SDK version

You get this error:

"The current .NET SDK does not support targeting .NET Core 2.0. Either target .NET Core 1.1 or lower, or use a version of the .NET SDK that supports .NET Core 2.0"

This has been reported both by @D3m0nKingx  (on linux) and @MithrilMan (on Windows)

mikedennis:
Should fix #933 

It appears that global.json was introduced in order to fix a compatibility issue with VS2015 side by side with VS2017 in a pre-release version of core (see https://github.com/stratisproject/StratisBitcoinFullNode/commit/9debb6cd11e4720643f683127ad5d08d28c5b0b8). I no longer think this is an issue. I have both VS2015 and VS2017 installed side by side and I do not require a global.json file in order to build. The removal of the global.json file should allow the project to build on newer tooling sets.

I also cleaned up a couple of compiler warnings I noticed that were present with this build.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs
@@ -428,7 +428,7 @@ namespace Stratis.Bitcoin.Features.Consensus
                 }
             }
 
-            context.NextWorkRequired = StakeValidator.GetNextTargetRequired(this.stakeChain, context.BlockValidationContext.ChainedBlock.Previous, context.Consensus,
+            context.NextWorkRequired = this.StakeValidator.GetNextTargetRequired(this.stakeChain, context.BlockValidationContext.ChainedBlock.Previous, context.Consensus,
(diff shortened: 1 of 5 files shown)


bokobza:
global.json should still be there actually.
It defines which SDK to use when you run commands like `dotnet run`, `dotnet build` etc...
However, feel free to remove the projects array from within it.

mikedennis:
Yes that is the problem - it forces a requirement on tooling. If you don't have it then it uses the latest version of the SDK. It's really a hold-over from VS2015 and should have be deleted when we migrated to VS2017 see https://github.com/dotnet/project-system/issues/1184 for some history on this.

bokobza:
Fair enough, thanks for the link :-)

PR #950, opened 2017-12-19
[PeerDiscovery] Fix threading issue with discovery loop

### LINKED ISSUE #949: Race condition on peer discovery
Copying a potential bug someone reported me

```
On another note, I switched to Stratis yesterday and was able to build it and run it under the debugger. I read a lot of the P2P networking code and managed to disable Peer Server and "Advertise" so the node functions as a passive node. I also suppressed the Mining and RPC features to minimize background noise so I can focus on the peer discovery portion of the node for now. The code is rather well structured and much easier to follow so kudos there.

 

I don't know if you've seen this, but one thing I noticed is that it seems like the Async Loops are somewhat instable, especially when you jack up the Parallelism in the Peer Discovery code. It looks like the loops will stop after a while and both the node and VS debugger will freeze and I am forced to kill the dotnet process. This can be reproduced quite regularly under the debugger. After a few minutes, both the Peer Discovery Loop and the periodic Stat updates on the console stop, and it never recovers.

 

I think I might have found a potential memory corruption issue in the Peer Discovery Loop (PeerDiscoveryLoop.cs). As I increase the Parallelism in the discovery loop, I noticed I start seeing some thread-safe insert issues in line 148 at this line:

 

clonedParameters.TemplateBehaviors.Add(addressManagerBehaviour);

 

Specifically, a null pointer exception. Digging deeper, I noticed that clonedParameters was instantiated in line 127 (outside of the lambda function that references it):

 

var clonedParameters = this.currentParameters.Clone();

 

I think all the parallel discovery threads are referencing the same instance of clonedParameters and causing corruption. I moved that line inside the lambda function (e.g. line 141) and the null pointer exceptions went away.

 

However, once I got past the null pointer exceptions, I started seeing a lot of OperationCanceledException's being thrown after the call to this.networkPeerFactory.CreateConnectedNetworkPeerAsync succeeds in line 150. This tells me that the cancelation is happening in either line 151, 152 or 154:

 

await networkPeer.VersionHandshakeAsync(connectTokenSource.Token).ConfigureAwait(false);

await networkPeer.SendMessageAsync(new GetAddrPayload(), connectTokenSource.Token).ConfigureAwait(false);

connectTokenSource.Token.WaitHandle.WaitOne(TimeSpan.FromSeconds(5));

 

I'm still investigating but the code there looks pretty fragile and does not handle error cases well. I've also seen unhandled exceptions being thrown in the FullNode.PeriodicLog Async Loop, causing the stats display to stop. Looking into that one too.

 

In any case, I think Stratis is an excellent codebase to help a newbie like me explore the underlying infrastructure Bitcoin and the BlockChain, and I want to thank you and the others for making this Open Source for others to follow and learn. So far it's pretty fascinating!

 
```

fassadlr:
This fixes #949

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
@@ -124,8 +124,6 @@ namespace Stratis.Bitcoin.P2P
                     return Task.CompletedTask;
             }
 
-            var clonedParameters = this.currentParameters.Clone();
-
             Parallel.ForEach(peersToDiscover, new ParallelOptions()
             {
                 MaxDegreeOfParallelism = 2,
@@ -141,6 +139,7 @@ namespace Stratis.Bitcoin.P2P
 
                     try
                     {
+                        var clonedParameters = this.currentParameters.Clone();
                         clonedParameters.ConnectCancellation = connectTokenSource.Token;
 
                         var addressManagerBehaviour = clonedParameters.TemplateBehaviors.Find<PeerAddressManagerBehaviour>();


src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -289,10 +289,12 @@ public string GetNodeStats()
             {
                 ConnectionManagerBehavior connectionManagerBehavior = node.Behavior<ConnectionManagerBehavior>();
                 ChainHeadersBehavior chainHeadersBehavior = node.Behavior<ChainHeadersBehavior>();
+
+                var agent = node.PeerVersion == null ? node.PeerVersion.UserAgent : "[Unknown]";
```
Aprogiena: should be string not var

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
```diff
@@ -141,6 +139,7 @@ private Task DiscoverPeersAsync()
 
                     try
                     {
+                        var clonedParameters = this.currentParameters.Clone();
```
Aprogiena: also should not be var

NicolasDorier:
Actually not approved...

`this.currentParameters` should not be a field, it should be a variable passed to `DiscoverPeersAsync`.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: Fixed

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
fassadlr: Fixed

fassadlr:
@NicolasDorier I think that it is OK for us to clone this.currentParameters inside the parallel foreach as the Clone() method creates a completely new object?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -289,10 +289,12 @@ public string GetNodeStats()
             {
                 ConnectionManagerBehavior connectionManagerBehavior = node.Behavior<ConnectionManagerBehavior>();
                 ChainHeadersBehavior chainHeadersBehavior = node.Behavior<ChainHeadersBehavior>();
+
+                string agent = node.PeerVersion == null ? node.PeerVersion.UserAgent : "[Unknown]";
```
dangershony: Is that even possible?

fassadlr: Wow.

NicolasDorier:
If `DiscoverPeers` get called by two different thread might have different `this.currentParameters` in the Parallel loop.

Aprogiena:
Yes, but it is only called once during initialization - see `ConnectionManager.Initialize`, so this is only called once by one thread, so I still can't see issue with that.

NicolasDorier:
yeah, just feel uneasy about it as now I need to review clients of this method to know if there is no bug.
I would in that case add a Guard to be sure this is not called twice by mistake.
Anyway not that important.

Aprogiena:
not saying the design is perfect, but not a bug either :)

PR #951, opened 2017-12-19
[PeerConnectorDiscovery] Fix logic as peer can potentially be null

fassadlr:


src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -68,6 +68,11 @@ public override PeerAddress FindPeerToConnectTo()
                 }
 
                 peer = this.peerAddressManager.PeerSelector.SelectPeer();
+                if (peer == null)
+                {
+                    peerSelectionFailed++;
+                    continue;
+                }
 
                 if (!peer.NetworkAddress.Endpoint.Address.IsValid())
```
Aprogiena: why not 
`if ((peer == null) || !peer.NetworkAddress.Endpoint.Address.IsValid())`

fassadlr: OK so I wanted to leave the conditions separate so that I can put logging in each "fail" scenario.

Let me add this now.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -89,20 +104,27 @@ public override PeerAddress FindPeerToConnectTo()
...
                 }
 
                 break;
             }
 
+            if (peer != null)
+                this.logger.LogTrace("(-):'{0}'", peer.NetworkAddress.Endpoint);
+            else
+                this.logger.LogTrace("(-)[PEER_NULL]");
+
```
Aprogiena: it should be possible to just do 
`this.logger.LogTrace("(-):'{0}'", peer?.NetworkAddress.Endpoint);`

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -89,20 +104,27 @@ public override PeerAddress FindPeerToConnectTo()
...
                 {
+                    this.logger.LogTrace("Peer selection failed, peer exists in -connect args {0}.", peer.NetworkAddress.Endpoint);
                     peerSelectionFailed++;
                     continue;
                 }
 
                 // If the peer is already connected just continue.
                 if (this.IsPeerConnected(peer.NetworkAddress.Endpoint))
                 {
+                    this.logger.LogTrace("Peer selection failed, peer is already connected {0}.", peer.NetworkAddress.Endpoint);
```
Aprogiena: pls add `'` around endpoints -> i.e. `'{0}'`
here and above

fassadlr: Fixed

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
fassadlr: Fixed

PR #954, opened 2017-12-19
resolve some compiler warnings

Aprogiena:
just to make cleaner compilations, but some are still unresolved

### DIFF
--- a/src/Stratis.StratisD/Properties/launchSettings.json
+++ b/src/Stratis.StratisD/Properties/launchSettings.json
@@ -1,7 +1,7 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project",
+      "commandName": "Project"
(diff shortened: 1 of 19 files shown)


src/Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project"
+      "commandName": "Project",
```
bokobza: the `Stratis.StratisD Test` launch setting already has `-testnet`.

Aprogiena: sorry my mistake, i did not want to commit this file, it happens so often :/ would be great to have it in git ignore 

will remove from PR

Issue #896, opened 2017-12-13
[PosMinting] Block mined during IBD


Aprogiena:
The following happened: 

```
info: Stratis.Bitcoin.Features.Miner.PosMinting[0]
      ==================================================================
info: Stratis.Bitcoin.Features.Miner.PosMinting[0]
      Found new POS block hash 'f8230171080cb3113b6ca46c57752874b8d6c81b9591a8546d1bd4d2650ebe0d' at height 223610.
info: Stratis.Bitcoin.Features.Miner.PosMinting[0]
      ==================================================================
info: Stratis.Bitcoin.FullNode[0]
      ======Node stats====== 12/13/2017 10:14:11 agent StratisBitcoin:1.0.5
      Headers.Height:        224454   Headers.Hash:         f207188bd53a9e0e2d8793908cc5ff7c501f86fd3d1d4487d90e0b811a85c1fb
      Consensus.Height:      223610   Consensus.Hash:       f8230171080cb3113b6ca46c57752874b8d6c81b9591a8546d1bd4d2650ebe0d
      BlockStore.Height:     223609   BlockStore.Hash:      09b6f955e4a110a8a4411cbfd0470edee847316b9c6c8e94b5280a7354a640ac
      Wallet.Height:         223609   Wallet.Hash:          09b6f955e4a110a8a4411cbfd0470edee847316b9c6c8e94b5280a7354a640ac

```

This should not happen.

Aprogiena:
this is not solved, neither understood why it happened

only one clue - node made no connection for very long time when it was started, is this related?

Aprogiena:
OK I think this could be caused by #921

noescape00:
Since MaxTipAge is 24h (1300 blocks) it wasn't considered IBD (we have a check for IBD in PoSMinting). 
So even if we set MaxTipAge to smth lower it is still possible to mine a block until we're fully synced. 

Actually I think that it's ok to mine a block if our tip is synced with the network but we don't have the missing blocks. We're still sending those mined blocks to the network so why not consider mining in this state to be a correct thing?

Aprogiena:
if we don't have missing blocks we can't mine, that's not ok
you need latest utxo set to be able to mine

noescape00:
Ok. So we shell add a check if block store height reached the headers tip.

dangershony:
>Ok. So we shell add a check if block store height reached the headers tip.

Hi why is this required?

[...]

PR #956, opened 2017-12-19
IBD improvements part 1

noescape00:
## Motivation

We have several bugs [#1](https://github.com/stratisproject/StratisBitcoinFullNode/issues/921) [#2](https://github.com/stratisproject/StratisBitcoinFullNode/issues/896) that could be solved by checking if the peer is in IBD. Unfortunately at the moment we don't have any method implemented for getting this info about the peer. 

Secondly even for our node the information about IBD is spread across a few classes and some of them don't even implement an 'IBlockDownloadState' interface. Also that causes ChainState to have multiple dependencies just because of that method. 

## Implementation

I suggest improving this situation in 3 steps: 

1) Separate code that provides IBD state from the ChainState and ConsensusManager, move it into a separate IBDStateProvider; Remove unnecessary dependencies that ChainState has in order to provide IBD state and also remove other code dependencies for ChainState in cases when dependency is there only for having the access the IBD state. 
2) Make Peers provide their IBD state 
3) Fix those bugs mentioned earlier.

This PR implements the 1st step.


noescape00:
@fassadlr can you please take a look at this one?

D3m0nKingx:
Another bug possibly related to this, see #957

src/Stratis.Bitcoin.Features.Consensus/IBDStateProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Text;
+using NBitcoin;
+using Stratis.Bitcoin.Base;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Interfaces;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus
+{
+    public class IBDStateProvider : IBlockDownloadState
```
fassadlr: Can you describe the purpose of this class here? Lets not introduce new class without documentation.

fassadlr: Also, lets rename this BlockDownloadStateProvider, so that it matches IBlockDownloadState? Either that or rename IBlockDownloadState?

src/Stratis.Bitcoin/Interfaces/IBlockDownloadState.cs
```diff
@@ -1,7 +1,23 @@
...
+        /// Checks whether the node is currently in the process of initial block download.
+        /// </summary>
+        /// <returns><c>true</c> if the node is currently doing IBD, <c>false</c> otherwise.</returns>
         bool IsInitialBlockDownload();
+
+        /// <summary>
+        /// Sets last IBD status update time and result.
+        /// <para>Used in tests only.</para>
+        /// </summary>
+        /// <param name="val">New value for the IBD status, <c>true</c> means the node is considered in IBD.</param>
```
fassadlr: Can name this properly? Like blockDownloadStatus or something.

src/Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -66,39 +50,5 @@ public void MarkBlockInvalid(uint256 hashBlock, DateTime? rejectedUntil = null)
         {
             this.invalidBlockHashStore.MarkInvalid(hashBlock, rejectedUntil);
         }
-
-        /// <summary>
-        /// This method will check if the node is in a state of IBD (Initial Block Download)
-        /// </summary>
-        public bool IsInitialBlockDownload
```
fassadlr: So where did this logic move to? If Consensus is not present then IBD is meaningless.

src/Stratis.Bitcoin/Interfaces/IBlockDownloadState.cs
```diff
@@ -1,7 +1,23 @@
-﻿namespace Stratis.Bitcoin.Interfaces
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Stratis.Bitcoin.Interfaces
 {
     public interface IBlockDownloadState
```
fassadlr: Add a bit of documentation here as well.

src/Stratis.Bitcoin/Base/ChainState.cs
noescape00: IBD logic is now accessible only through the IBlockDownloadState. I've replaced all usages of that property with accessing instance of IBlockDownloadState.

src/Stratis.Bitcoin/Interfaces/IBlockDownloadState.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Interfaces/IBlockDownloadState.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.Consensus/IBDStateProvider.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
```diff
@@ -32,7 +33,7 @@ public class ConsensusStats : SignalObserver<Block>
         /// <summary>Manager of the longest fully validated chain of blocks.</summary>
         private readonly ConsensusLoop consensusLoop;
 
-        private readonly ChainState chainState;
+        private readonly IBlockDownloadState blockDownloadState;
```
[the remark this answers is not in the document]
noescape00: Fixed.

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
```diff
@@ -0,0 +1,84 @@
...
+        private readonly Network network;
+
+        /// <summary>User defined node settings.</summary>
+        private readonly NodeSettings nodeSettings;
+
+        /// <summary>Time until IBD state can be checked.</summary>
+        private DateTime lockIbdUntil;
+
+        /// <summary>A cached result of the IBD method.</summary>
+        private bool ibdCashed;
```
fassadlr: This should be ibdCached

src/Stratis.Bitcoin/Base/ChainState.cs
fassadlr: OK but there is some logic here like checking the ibdLastUpdate as well as checking that the ibdStateProvider (in this instance, Consensus) exists. If not we return false. Where is this logic in the new BlockDownloadStateProvider? Sorry if i'm missing something here?

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
```diff
@@ -0,0 +1,84 @@
...
+            this.nodeSettings = nodeSettings;
+            this.chainState = chainState;
+            this.checkpoints = checkpoints;
+            this.dateTimeProvider = DateTimeProvider.Default;
+
+            this.lockIbdUntil = DateTime.MinValue;
+        }
+
+        /// <inheritdoc />
+        public bool IsInitialBlockDownload()
```
fassadlr: Do we unit tests that explicitly test the scenarios below?

noescape00: We have some integration tests that involve IBD but no, we don't have a check for every scenario.

noescape00:
>OK but there is some logic here like checking the ibdLastUpdate as well as checking that the ibdStateProvider (in this instance, Consensus) exists. If not we return false. Where is this logic in the new BlockDownloadStateProvider? Sorry if i'm missing something here?

ibdLastUpdate was added just to prevent changing IBD state in unit tests even if logic consider changing it. This logic is saved in form of 'lockIbdUntil'. 

>as well as checking that the ibdStateProvider (in this instance, Consensus) exists.
ChainState was getting the ConsensusManager to access the chain state. Now we're getting it directly. And we're not using the consensus- instead we're using 'chainState.ConsensusTip'. 

So overall we do not need to perform the old checks now.

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
fassadlr: Can you perhaps create a few tests that verify these scenarios? This way we create confidence in the code. Thanks 💯

noescape00: Sure @fassadlr =)

src/Stratis.Bitcoin.Features.BlockStore.Tests/LoopTests/BlockStoreLoopStepBaseTest.cs
```diff
@@ -35,6 +36,7 @@ internal class FluentBlockStoreLoop : IDisposable
         private DataFolder dataFolder;
         private Mock<INodeLifetime> nodeLifeTime;
         private Mock<ILoggerFactory> loggerFactory;
+        private IBlockDownloadState blockDownloadState;
```
Aprogiena: IBD states for Initial Block Download, I'd prefer to preserve "Initial" in the name

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <summary>Time until IBD state can be checked.</summary>
+        private DateTime lockIbdUntil;
+
+        /// <summary>A cached result of the IBD method.</summary>
+        private bool ibdCached;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="BlockDownloadState"/> class.
+        /// </summary>
+        public BlockDownloadState(ChainState chainState, Network network, NodeSettings nodeSettings, ICheckpoints checkpoints)
```
Aprogiena: you can have params for free here, i know it is no longer required, just saying ...

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
```diff
@@ -0,0 +1,84 @@
...
+            this.lockIbdUntil = DateTime.MinValue;
+        }
+
+        /// <inheritdoc />
+        public bool IsInitialBlockDownload()
+        {
+            if (this.lockIbdUntil >= this.dateTimeProvider.GetUtcNow())
+                return this.ibdCached;
+
+            if (this.chainState == null)
```
Aprogiena: in original code it was 
```
            if (this.ConsensusLoop == null)
                return false;
```

i'm not sure this is equivalent (if it is, it is not obvious)
i thought IBD was false when consensus feature is not activated
but chainState will be present always, won't it?
this needs to be explained

Aprogiena:
So ChainState is in Stratis.Bitcoin, therefore everyone can use it
but if consensus is not available how can you use IBlockDownloadState? it is only initialized in use consesus feature, so it is not available

i don't think this is right

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore.Tests/LoopTests/BlockStoreLoopStepBaseTest.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Consensus/BlockDownloadState.cs
noescape00: Fixed

noescape00:
>i think it still should remain in ChainState so that it is accessible even without consensus

But it never was. 
ChainState had a IsInitialBlockDownload property (without inheriting IBlockDownloadState interface- this is bad) and all it did was checking if full node has any service that implements IBlockDownloadState. And if not we are assuming it's false. 

Why don't we do the same at any place where we maybe won't have a consensus? Try getting a service of type IBlockDownloadState and if it's null assume it's false.

Aprogiena:
because in many places you don't have the service provider, so you can't do it

i'd be ok with this only if no one was using it, but i'm not sure that it is the case

noescape00:
>i'd be ok with this only if no one was using it

I didn't remove any usages of IBD state in this PR, only changed the source. So far there was not a scenario when we need that IBD state and don't have an access to a DI or fullNode.GetFeature

noescape00:
>and then if you implement it separately, you will end with something like ChainState ... it was excatly the purpose of ChainState - to give you false when consensus is not available
and since the ChainState is always available, it works

All the usages of ChainState in order to get IBD state were unnecessary.

And overall having a IsInitialBlockDownload in ChainState without implementing an interface was a hack since on most cases DI already had smth that implemented that interface => we can't implement it on the ChainState.

fassadlr:
@noescape00, @Aprogiena might have a point here. If I look at ChainState now, after your work, the only information it now contains is IInvalidBlockHashStore and ConsensusTip. 

I maintain that it is right (better) for us to move IBD logic into its own class. It decouples it and makes it easier to test. We can then share this logic to other peers easier.

However, it does kinda leave ChainState "thin" but is that a bad thing?

At the end of the day, we need to break the dependencies ChainState creates on full node / consensus etc and isolate IBD logic, so this PR is a step in the right direction.

Ultimately @bokobza @dangershony will need to provide input here but I'm all for this change.

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class InitialBlockDownloadTest
+    {
+        private readonly ConsensusSettings consensusSettings;
+        private readonly Checkpoints checkpoints;
+        private readonly NodeSettings nodeSettings;
+        private readonly ChainState chainState;
+
+        public InitialBlockDownloadTest()
+        {
+            this.consensusSettings = new ConsensusSettings(NodeSettings.Default(), new LoggerFactory());
```
fassadlr: Change this ExtendedLoggerFcatory so that its more consistent with what the rest of the code base.

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
```diff
@@ -0,0 +1,56 @@
...
+            this.checkpoints = new Checkpoints(Network.Main, this.consensusSettings);
```
fassadlr: Maybe introduce a variable to hold the Network instance and pass that around?

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
```diff
@@ -0,0 +1,56 @@
...
+            this.nodeSettings = new NodeSettings("stratis", Network.Main);
+            this.chainState = new ChainState(new InvalidBlockHashStore(new DateTimeProvider()));
```
fassadlr: This should be DateTimeProvider.Default

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
noescape00: Done

src/Stratis.Bitcoin.Features.Consensus.Tests/InitialBlockDownloadTest.cs
noescape00: Fixed

fassadlr:
I'm happy with this PR now. But like I said, @dangershony and @bokobza will need to give input going forward.

Aprogiena:
yeah, i would like input from @bokobza because he knows about wallet and lightwallet, which is exactly the environment without the consensus

it feels strange that @noescape00 mentions that no one is actually using that property, if this is true then ok, just want some expert to confirm

noescape00:
@Aprogiena you're right. That dummy code in ChainState had perfect sense for the features that don't have a consensus. So in my last commit I've added a dummy IBD state provider that always returns false and made light wallet utilize it.

Aprogiena:
so was it used or not?

src/Stratis.Bitcoin/Interfaces/IInitialBlockDownloadState.cs
```diff
@@ -0,0 +1,26 @@
...
+    public interface IInitialBlockDownloadState
+    {
+        /// <summary>
+        /// Checks whether the node is currently in the process of initial block download.
+        /// </summary>
+        /// <returns><c>true</c> if the node is currently doing IBD, <c>false</c> otherwise.</returns>
+        bool IsInitialBlockDownload();
+
+        /// <summary>
+        /// Sets last IBD status update time and result.
```
dangershony: If its used in tests why does this method need to live on the interface?
Tests can easily override the concrete implementation.

src/Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -12,18 +12,6 @@ namespace Stratis.Bitcoin.Base
     /// </summary>
     public class ChainState
     {
-        /// <summary>The fullnode interface.</summary>
```
dangershony: I think this class now becomes obsolete

dangershony:
Hmm this is a bit of a complicated one.
IBD is a property of consensus only and its dependencies.
But its not entirely true  as a light wallet might also be in sort of an IBD (while syncing the wallet code).
This make IBD sort of a stand alone component that can be reused by features that download blocks.

I like the `IInitialBlockDownloadState` interface.

I would suggest to remove chain state all together, its too generic we use interfaces for specific functionality and not a general state class.

dangershony:
>So in my last commit I've added a dummy IBD state provide

If I remember correctly this dummy is not necessary (in wallet for example)

src/Stratis.Bitcoin/Interfaces/IInitialBlockDownloadState.cs
noescape00: It's for the integration tests and inherited from the old implementation- also don't like it being here. 
I've tried to easily remove that but it looks like we need to build a node in integration tests environment with a mock so I decided to do that in a separate PR - too much to change to remove that.

noescape00:
>so was it used or not?
>If I remember correctly this dummy is not necessary (in wallet for example)

It was used so yeah, without the dummy DI can't resolve a dependency for the light wallet. Base Feature requires it in order to provide to ChainHeadersBehavior which needs IBD to decide if 'GetHeadersPayload' messages should be ignored or not.

noescape00:
>I would suggest to remove chain state all together, its too generic we use interfaces for specific functionality and not a general state class.

Would you like me to do this as a part of this PR or maybe it will be better do do in a follow up PR in order to keep this one's size not too huge?

Aprogiena:
*needed

so i leave this one to you guys

@noescape00 for sure not for this PR to further remove ChainState
that is deep and not sure if it is that easy

noescape00:
>@noescape00 for sure not for this PR to further remove ChainState that is deep and not sure if it is that easy

Ok. 
When I'll start working on removing Chan State (probably in 3-4 PRs since I need to do step 2 and fix 3 bugs that depend on step 2) I'll consider this carefully and will discuss the ideas on how to execute this better.

PR #875, opened 2017-12-08
First miner feature tests

Neurosploit:
-Add test project for miner feature
-Tests for MinerController
-Tests for MiningRPCController
-Attempt to fix locally failing LightWalletManager tests

PR #963, opened 2017-12-20
Mining Feature Tests

Neurosploit:
See https://github.com/stratisproject/StratisBitcoinFullNode/pull/875
Messed up my branch merging with main and I don't want to be spending an entire afternoon trying to fix that so I redid everything. 
Learnt a lot from this and made some notes on what not to do next time I try this.
The commits that contains some changes when compared to the old PR are the last two in the list.


bokobza:
That's great, thanks @Neurosploit .

PR #965, opened 2017-12-20
DNS Server integrated populated with WhiteLists

345James:
Contains two tasks, one to link the Peer Address Manager to the DNS Server and another to implement the load and save of the master file from disk.
- New class WhiteListManager, to allow the DNS feature to populate the resource records in the DNS 
Server using the Peer Address Manager.
- Implemented a custom JSON Converter to handle the serializing and deserializing of the 
 IResourceRecords as these are from the nuget package and do not have parameterless constructors or setters for all of the properties.
- Bought in the functionality from the MasterFile in the DNS nuget package into the customised DnsSeedMasterFile.
- Added the DNS peer blacklist threshold and DNS hostname to the node settings.
- Added relevant unit tests

### DIFF
--- a/src/Stratis.Bitcoin.Features.Dns/DnsSeedMasterFile.cs
+++ b/src/Stratis.Bitcoin.Features.Dns/DnsSeedMasterFile.cs
@@ -1,36 +1,113 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Text;
+using System.Linq;
+using System.Text.RegularExpressions;
 using DNS.Protocol;
 using DNS.Protocol.ResourceRecords;
-using DNS.Server;
-using Stratis.Bitcoin.P2P.Peer;
+using Newtonsoft.Json;
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.Dns
 {
     /// <summary>
     /// This class defines a DNS masterfile used to cache the whitelisted peers discovered by the DNS Seed service that supports saving
     /// and loading from a stream.
+    /// This is based on 3rd party library https://github.com/kapetan/dns.
     /// </summary>
-    public class DnsSeedMasterFile : MasterFile, IMasterFile
+    public class DnsSeedMasterFile : IMasterFile
     {
         /// <summary>
-        /// Adds a <see cref="NetworkPeer"/> object to the masterfile.
+        /// Sets the default ttl.
         /// </summary>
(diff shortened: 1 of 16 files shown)


345James:
[[DNS] Create whitelist Masterfile from PeerAddressManager](https://trello.com/c/Fk6GQ0z0/328-dns-create-whitelist-masterfile-from-peeraddressmanager)

345James:
[[DNS] Periodically write cached Masterfile to disk](https://trello.com/c/7XB7pkfu/329-dns-periodically-write-cached-masterfile-to-disk)

src/Stratis.StratisDnsD/Properties/launchSettings.json
```diff
@@ -1,11 +1,11 @@
...
-      "commandName": "Project"
-    },
-    "Stratis.StratisDnsD Test": {
-      "commandName": "Project",
-      "commandLineArgs": "-testnet -dnslistenport=5399"
+    "profiles": {
+        "Stratis.StratisDnsD": {
+            "commandName": "Project"
+        },
+        "Stratis.StratisDnsD Test": {
```
bokobza: Can you create a third launch settings that has the dns stuff in, but leave the testnet one to be just -testnet?

src/Stratis.Bitcoin.Features.Dns/DnsSeedMasterFile.cs
```diff
@@ -48,7 +131,13 @@ public void Load(Stream stream)
         /// <param name="stream">The stream to write the masterfile to.</param>
         public void Save(Stream stream)
         {
-            // TODO
+            Guard.NotNull(stream, nameof(stream));
```
bokobza: Can you use FileStorage class for that, under Stratis.Bitcoin.Utilities, or is it not appropriate?
Same question for Load(Stream) above.

bokobza:
Can you please resolve this conflict as well 😊 ?

src/Stratis.Bitcoin.Features.Dns/DnsSeedMasterFile.cs
345James: We need to control the serialization and deserialization of the IResourceRecords which are from the DNS nuget package, via a JsonConverter, so its better to receive the stream. Also the load and save get a stream, making them easier to unit test without relying on a file system.

bokobza: fair enough :-)

PR #971, opened 2017-12-21
[DNS] Added round-robin ordering of responses

345paul:
- For Trello item: https://trello.com/c/5i8S6Zde
- Added the capability of re-ordering responses in a round-robin fashion to avoid same peer being queried every time.
- Added unit test.


345paul:
[[DNS] Randomise output order of IP addresses following DNS request](https://trello.com/c/5i8S6Zde/331-dns-randomise-output-order-of-ip-addresses-following-dns-request)

Issue #930, opened 2017-12-16
Blocks propagation is at least 10% slower than it can be. 


noescape00:
## Problem

When miner mines a new block (this is the most sensitive case since we want that reward and would like to make chances of making an orphaned block as low as it possible) or when we receive a block from the network it usually takes 0.8 - 1.5 seconds before we actually start announcing it to the peers. 

Average block propagation time on bitcoin network (until >50% of the network know about the block) is ~11 seconds so having that 0.8-1.5 sec delay is actually slowing the propagation by at least ~10%.

## Why is it happening?

We're adding new block's header to the list of headers for propagation and check this list every 1 sec (see BlockStoreSignaled.RelayWorker). So in case we add a new header just after the list was checked we would end up waiting 0.999 sec till we start propagating added hash. 

## Solution

I suggest replacing continuous checking if the list is empty with triggering the propagation when new hash is added.

noescape00:
@Aprogiena mind if I take this one or you have plans for this one in the next node job PRs?

Aprogiena:
this has nothing to do with node refactoring
but @dangershony had some input on this when we discussed it

Aprogiena:
in any case, do not start any work until we agree on how to proceed (and also when - some things may be more important)

noescape00:
I'll try to elaborate on my approach: 

Current implementation:
When new block appears we add it to a list of blocks, once a second we flush that list and announce hashes that this list contains. 

Suggested implementation: 

Instead of checking the list once a second we will do following: 
1) New block arrives- add it to the list. 
2) When new block is added we would attempt to start an announcing task. If the task is already running we will do nothing.
3) Announcing task will be identical to the current one logic-wise but when in the end it checks if there are still some blocks in the list and if there are the task would restart itself. 

Doing it this way will prevent announcing each hash separately- instead if while we were announcing some hashes and some more were added during that we would announce them all together. Also this approach won't announce hashes in a wrong order.

noescape00:
Please let me know if you think this should be done differently, probably we will come up with even better solution.

Aprogiena:
it seems that you want to open the pandora's box, or maybe you don't know that it is a pandora's box :-)

for the first look, it might seem logical - yes, do propagate the block as soon as possible, well obviously and i agree with that. the problem is that it won't reduce the orphaning, not if everyone improves. but there is more to it

[...]

PR #976, opened 2017-12-22
[BlockStoreSignaled] Replaced 1 second announcing loop with event-based announcing

noescape00:
Replaced an async loop with a 'producer-consumer' pattern based even system. 

Fix: #930

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
@@ -48,6 +48,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Timestamp of last announcement event.</summary>
         private DateTime lastBlockAnnounceTimeStamp;
 
+        /// <summary>Date and time information provider.</summary>
+        private readonly IDateTimeProvider dateTimeProvider;
+
         public BlockStoreSignaled(
             BlockStoreLoop blockStoreLoop,
             ConcurrentChain chain,
@@ -56,7 +59,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
             IConnectionManager connection,
             INodeLifetime nodeLifetime,
             IBlockRepository blockRepository,
-            ILoggerFactory loggerFactory)
+            ILoggerFactory loggerFactory,
+            IDateTimeProvider dateTimeProvider)
         {
             this.blocksToAnnounce = new ConcurrentQueue<ChainedBlock>();
             this.blockRepository = blockRepository;
@@ -67,6 +71,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.nodeLifetime = nodeLifetime;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.storeSettings = storeSettings;
+            this.dateTimeProvider = dateTimeProvider;
             this.blockEnqueued = new ManualResetEventSlim(false);
             this.lastBlockAnnounceTimeStamp = DateTime.MinValue;
         }
@@ -155,10 +160,11 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
                     // Make sure that at least 'FlushFrequencyMilliseconds' passed since the last announcement.
                     // This is needed in order to ensure announcing blocks in batches to reduce the overhead.
-                    int msSinceLastAnnounce = this.lastBlockAnnounceTimeStamp == DateTime.MinValue ? int.MaxValue : (int)(DateTime.Now - this.lastBlockAnnounceTimeStamp).TotalMilliseconds;
+                    int msSinceLastAnnounce = this.lastBlockAnnounceTimeStamp == DateTime.MinValue ? int.MaxValue :
+                        (int)(this.dateTimeProvider.GetUtcNow() - this.lastBlockAnnounceTimeStamp).TotalMilliseconds;
                     if (msSinceLastAnnounce < FlushFrequencyMilliseconds)
                         await Task.Delay(FlushFrequencyMilliseconds - msSinceLastAnnounce, this.nodeLifetime.ApplicationStopping);
-                    this.lastBlockAnnounceTimeStamp = DateTime.Now;
+                    this.lastBlockAnnounceTimeStamp = this.dateTimeProvider.GetUtcNow();
 
                     this.blockEnqueued.Reset();
                 }
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
@@ -16,6 +16,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
 {
     public class BlockStoreSignaled : SignalObserver<Block>
     {
+        /// <summary>Maximum time in seconds for forming a new batch.</summary>
+        private const double FlushFrequencySeconds = 0.5;
+
         private Task relayWorkerTask;
 
         private readonly IBlockRepository blockRepository;
@@ -41,11 +44,11 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Queue of chained blocks that will be announced to the peers.</summary>
         private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
 
+        /// <summary>Event slim that is set when a new block gets enqueued.</summary>
         private ManualResetEventSlim blockEnqueued;
 
+        /// <summary>Timestamp of last announcement event.</summary>
         private DateTime lastBlockAnnounceTimeStamp;
-
-        private const double FlushFrequencySeconds = 0.5;
 
         public BlockStoreSignaled(
             BlockStoreLoop blockStoreLoop,
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
@@ -19,7 +19,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Maximum time in milliseconds for forming a new batch.</summary>
         private const int FlushFrequencyMilliseconds = 500;
 
-        private Task relayWorkerTask;
+        /// <summary>Async loop that continuously runs <see cref="RelayWorkerAsync"/>.</summary>
+        private IAsyncLoop relayWorkerLoop;
 
         private readonly IBlockRepository blockRepository;
 
@@ -42,6 +43,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Queue of chained blocks that will be announced to the peers.</summary>
         private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
 
+        /// <summary>Factory for creating background async loop tasks.</summary>
+        private readonly IAsyncLoopFactory asyncLoopFactory;
+
         /// <summary>Event that is fired when a new block gets enqueued.</summary>
         private readonly ManualResetEventSlim blockEnqueued;
 
@@ -60,7 +64,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
             INodeLifetime nodeLifetime,
             IBlockRepository blockRepository,
             ILoggerFactory loggerFactory,
-            IDateTimeProvider dateTimeProvider)
+            IDateTimeProvider dateTimeProvider,
+            IAsyncLoopFactory asyncLoopFactory)
         {
             this.blocksToAnnounce = new ConcurrentQueue<ChainedBlock>();
             this.blockRepository = blockRepository;
@@ -72,6 +77,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.storeSettings = storeSettings;
             this.dateTimeProvider = dateTimeProvider;
+            this.asyncLoopFactory = asyncLoopFactory;
             this.blockEnqueued = new ManualResetEventSlim(false);
             this.lastBlockAnnounceTimeStamp = DateTime.MinValue;
         }
@@ -116,7 +122,12 @@ namespace Stratis.Bitcoin.Features.BlockStore
         {
             this.logger.LogTrace("()");
 
-            this.relayWorkerTask = Task.Run(async delegate { await this.RelayWorkerAsync(); });
+            this.relayWorkerLoop = this.asyncLoopFactory.Run("BlockStore", async token =>
+            {
+                await this.RelayWorkerAsync();
+            },
+            this.nodeLifetime.ApplicationStopping,
+            repeatEvery: TimeSpans.RunOnce);
 
             this.logger.LogTrace("(-)");
         }
@@ -167,7 +178,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     int msSinceLastAnnounce = this.lastBlockAnnounceTimeStamp == DateTime.MinValue ? int.MaxValue :
                         (int)(this.dateTimeProvider.GetUtcNow() - this.lastBlockAnnounceTimeStamp).TotalMilliseconds;
                     if (msSinceLastAnnounce < FlushFrequencyMilliseconds)
-                        await Task.Delay(FlushFrequencyMilliseconds - msSinceLastAnnounce, this.nodeLifetime.ApplicationStopping);
+                        await Task.Delay(FlushFrequencyMilliseconds - msSinceLastAnnounce, this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
                     this.lastBlockAnnounceTimeStamp = this.dateTimeProvider.GetUtcNow();
 
                     this.blockEnqueued.Reset();
@@ -252,7 +263,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <inheritdoc />
         protected override void Dispose(bool disposing)
         {
-            this.relayWorkerTask.Wait();
+            this.relayWorkerLoop.Dispose();
             this.blockEnqueued.Dispose();
 
             base.Dispose(disposing);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
@@ -147,7 +147,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
                 {
                     // Wait until a new block is added to the queue.
-                    this.blockEnqueued.Wait();
+                    this.blockEnqueued.Wait(this.nodeLifetime.ApplicationStopping);
 
                     // Make sure that at least 'FlushFrequencySeconds' seconds passed since the last announcement.
                     // This is needed in order to ensure announcing blocks in batches to reduce the overhead.
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
@@ -34,8 +34,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
-        private readonly string name;
-
         /// <summary>Global application life cycle control - triggers when application shuts down.</summary>
         private readonly INodeLifetime nodeLifetime;
 
@@ -44,8 +42,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Queue of chained blocks that will be announced to the peers.</summary>
         private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
 
-        /// <summary>Event slim that is set when a new block gets enqueued.</summary>
-        private ManualResetEventSlim blockEnqueued;
+        /// <summary>Event that is fired when a new block gets enqueued.</summary>
+        private readonly ManualResetEventSlim blockEnqueued;
 
         /// <summary>Timestamp of last announcement event.</summary>
         private DateTime lastBlockAnnounceTimeStamp;
@@ -58,8 +56,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             IConnectionManager connection,
             INodeLifetime nodeLifetime,
             IBlockRepository blockRepository,
-            ILoggerFactory loggerFactory,
-            string name = "BlockStore")
+            ILoggerFactory loggerFactory)
         {
             this.blocksToAnnounce = new ConcurrentQueue<ChainedBlock>();
             this.blockRepository = blockRepository;
@@ -67,7 +64,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.chain = chain;
             this.chainState = chainState;
             this.connection = connection;
-            this.name = name;
             this.nodeLifetime = nodeLifetime;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.storeSettings = storeSettings;
@@ -111,6 +107,16 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.logger.LogTrace("(-)");
         }
 
+        public void Initialize()
+        {
+            this.logger.LogTrace("()");
+
+            this.relayWorkerTask = Task.Run(async delegate { await this.RelayWorkerAsync(); });
+
+            this.logger.LogTrace("(-)");
+        }
+
+
         /// <summary>
         /// A loop method that continuously relays blocks found in <see cref="blocksToAnnounce"/> to connected peers on the network.
         /// </summary>
@@ -138,13 +144,11 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// </para>
         /// TODO: consider moving the relay logic to the <see cref="LoopSteps.ProcessPendingStorageStep"/>.
         /// </remarks>
-        public void RelayWorker()
+        private async Task RelayWorkerAsync()
         {
-            this.logger.LogTrace("()");
-
-            this.relayWorkerTask = Task.Factory.StartNew(async delegate
+            while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
             {
-                while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
+                try
                 {
                     // Wait until a new block is added to the queue.
                     this.blockEnqueued.Wait(this.nodeLifetime.ApplicationStopping);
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -13,10 +16,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 {
     public class BlockStoreSignaled : SignalObserver<Block>
     {
-        private readonly IAsyncLoopFactory asyncLoopFactory;
-
-        /// <summary>The async loop we need to wait upon before we can shut down this feature.</summary>
-        private IAsyncLoop asyncLoop;
+        private Task relayWorkerTask;
```
Aprogiena: pls comment all these new guys

also could you put the constant above fields?

noescape00: Fixed

dangershony:
What's the point of this PR?

noescape00:
>What's the point of this PR?

The point is to start announcing a new block as soon as it arrived instead of waiting up to 1 second before we do that. 

The reasoning behind it and the implementation are described\discussed in the [issue 930](https://github.com/stratisproject/StratisBitcoinFullNode/issues/930).

Aprogiena:
@dangershony it is a good read, totally recommended :D

Aprogiena:
also, this PR is for me to review, so don't merge!

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -41,19 +44,23 @@ public class BlockStoreSignaled : SignalObserver<Block>
         /// <summary>Queue of chained blocks that will be announced to the peers.</summary>
         private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
 
+        /// <summary>Event slim that is set when a new block gets enqueued.</summary>
```
Aprogiena: remove slim from comment, that is an implementation detail

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -128,90 +142,104 @@ public void RelayWorker()
...
-                    this.logger.LogTrace("(-)[NO_BLOCKS]");
-                    return;
-                }
+                    // Wait until a new block is added to the queue.
+                    this.blockEnqueued.Wait();
 
-                this.logger.LogTrace("There are {0} blocks in the announce queue.", announceBlockCount);
+                    // Make sure that at least 'FlushFrequencySeconds' seconds passed since the last announcement.
+                    // This is needed in order to ensure announcing blocks in batches to reduce the overhead.
+                    double secondsSinceLastAnnounce = (DateTime.Now - this.lastBlockAnnounceTimeStamp).TotalSeconds;
```
Aprogiena: int here and milliseconds

Aprogiena: we don't use date functions directly, always use IDateTimeProvider

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -128,90 +142,104 @@ public void RelayWorker()
...
-
-                int announceBlockCount = this.blocksToAnnounce.Count;
-                if (announceBlockCount == 0)
+                while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
                 {
-                    this.logger.LogTrace("(-)[NO_BLOCKS]");
-                    return;
-                }
+                    // Wait until a new block is added to the queue.
+                    this.blockEnqueued.Wait();
```
Aprogiena: you can't do that, it blocks without cancellation

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -128,90 +142,104 @@ public void RelayWorker()
         {
             this.logger.LogTrace("()");
 
-            this.asyncLoop = this.asyncLoopFactory.Run($"{this.name}.RelayWorker", async token =>
+            this.relayWorkerTask = Task.Factory.StartNew(async delegate
```
Aprogiena: create InitializeAsync which will set rWT, which will be private async Task RelayWorkerAsync
don't use Task.Factory.StartNew, you don't need this at all, it will work:

```
public void Initialize()
{
inlog
    this.relayWorkerTask = this.RelayWorkerAsync();
outlog
}
```

just like that

and also we could migrate `            this.signals.SubscribeForBlocks(this.blockStoreSignaled);` to that initialization method from BSF

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -128,90 +142,104 @@ public void RelayWorker()
...
 
-                // Initialize this list with default size of 'announceBlockCount + 4' to prevent it from autoresizing during adding new items.
-                // This +4 extra size is in case new items will be added to the queue during the loop.
-                var broadcastItems = new List<ChainedBlock>(announceBlockCount + 4);
+                    this.blockEnqueued.Reset();
 
-                while (this.blocksToAnnounce.TryPeek(out ChainedBlock block))
-                {
-                    this.logger.LogTrace("Checking if block '{0}' is on disk.", block);
+                    this.logger.LogTrace("()");
```
Aprogiena: this is an entry log, it is supposed to be on method entry
same for lambdas and delagates - at their very start, not in the middle of nowhere

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -128,90 +142,104 @@ public void RelayWorker()
...
-                foreach (BlockStoreBehavior behaviour in behaviours)
-                    await behaviour.AnnounceBlocksAsync(broadcastItems).ConfigureAwait(false);
-            },
-            this.nodeLifetime.ApplicationStopping,
-            repeatEvery: TimeSpans.Second,
-            startAfter: TimeSpans.FiveSeconds);
+                    this.logger.LogTrace("{0} blocks will be sent to {1} peers.", broadcastItems.Count, behaviours.Count());
+                    foreach (BlockStoreBehavior behaviour in behaviours)
+                        await behaviour.AnnounceBlocksAsync(broadcastItems).ConfigureAwait(false);
+                }
```
Aprogiena: here should be the exit log, but we will have no such delegate anyway, so just for future ....

noescape00:
>the major fault is that you have a blocking wait there without cancellation

Hmm, you're right. Weird I never had a node hanging on shutdown. 
Fixing.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: >this.relayWorkerTask = this.RelayWorkerAsync();

that locks current context because the task does Wait(). So we still need to start a separate task like: 
` this.relayWorkerTask = Task.Run(async delegate { await this.RelayWorkerAsync(); });`

noescape00: But I like the idea of Initialize method, will do.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -13,10 +16,10 @@ namespace Stratis.Bitcoin.Features.BlockStore
 {
     public class BlockStoreSignaled : SignalObserver<Block>
     {
-        private readonly IAsyncLoopFactory asyncLoopFactory;
+        /// <summary>Maximum time in seconds for forming a new batch.</summary>
+        private const double FlushFrequencySeconds = 0.5;
```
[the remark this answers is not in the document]
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed for the sake of code consistency.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: no :D tha does not lock current context because no blocking waits are allowed, you only wait with await and that does not block

noescape00: ManualResetEventSlim don't have an async wait method so in order to do it without starting a task we need to add a manual async version of ManualResetEventSlim or use smth else.

Aprogiena: that is true, but it is possible to construct it
alternatively we can use non-async relay worker which will be based on handle waits

noescape00: we can. But does it worth it?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -124,19 +149,39 @@ protected override void OnNextCore(Block block)
...
+                {
+                    // Wait until a new block is added to the queue.
+                    this.blockEnqueued.Wait(this.nodeLifetime.ApplicationStopping);
+
+                    // Make sure that at least 'FlushFrequencyMilliseconds' passed since the last announcement.
+                    // This is needed in order to ensure announcing blocks in batches to reduce the overhead.
+                    int msSinceLastAnnounce = this.lastBlockAnnounceTimeStamp == DateTime.MinValue ? int.MaxValue :
+                        (int)(this.dateTimeProvider.GetUtcNow() - this.lastBlockAnnounceTimeStamp).TotalMilliseconds;
+                    if (msSinceLastAnnounce < FlushFrequencyMilliseconds)
+                        await Task.Delay(FlushFrequencyMilliseconds - msSinceLastAnnounce, this.nodeLifetime.ApplicationStopping);
```
Aprogiena: every await we have in FN code should be with .ConfigureAwait(false)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -13,10 +16,10 @@ namespace Stratis.Bitcoin.Features.BlockStore
 {
     public class BlockStoreSignaled : SignalObserver<Block>
     {
-        private readonly IAsyncLoopFactory asyncLoopFactory;
+        /// <summary>Maximum time in milliseconds for forming a new batch.</summary>
+        private const int FlushFrequencyMilliseconds = 500;
 
-        /// <summary>The async loop we need to wait upon before we can shut down this feature.</summary>
-        private IAsyncLoop asyncLoop;
+        private Task relayWorkerTask;
```
Aprogiena: we could possibly comment this

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -13,10 +16,10 @@ namespace Stratis.Bitcoin.Features.BlockStore
 {
     public class BlockStoreSignaled : SignalObserver<Block>
     {
-        private readonly IAsyncLoopFactory asyncLoopFactory;
+        /// <summary>Maximum time in milliseconds for forming a new batch.</summary>
+        private const int FlushFrequencyMilliseconds = 500;
```
Aprogiena: i think 500 ms is too low, i'd go for 5 second

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: This should be an async loop not a stand alone task

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -97,15 +107,30 @@ protected override void OnNextCore(Block block)
...
+            this.blockEnqueued.Set();
+
+            this.logger.LogTrace("(-)");
+        }
+
+        public void Initialize()
+        {
+            this.logger.LogTrace("()");
+
+            this.relayWorkerTask = Task.Run(async delegate { await this.RelayWorkerAsync(); });
```
dangershony: We should avoid Task.Run in backround work thats why we have the async loop.
I know in the peer code we dont use async loop but that is an exception.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -124,19 +149,39 @@ protected override void OnNextCore(Block block)
...
 
-            this.asyncLoop = this.asyncLoopFactory.Run($"{this.name}.RelayWorker", async token =>
+            while (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
             {
                 this.logger.LogTrace("()");
 
+                try
+                {
+                    // Wait until a new block is added to the queue.
+                    this.blockEnqueued.Wait(this.nodeLifetime.ApplicationStopping);
```
dangershony: Why not use a BlockingCollection?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: i can't see how to use async loop easily here, async loop can't be called currently on demand

dangershony:
I'd like to review this as well.
Also remember that when we broadcast a block to a peer we need to have that block in store thats why we wait for store before we broadcast.
So I am not sure broadcast directly when a block is announced will actually save much time.

What we need to do is push the block that is about to be broadcasted to cache and then push it out immediately that will reduce the propagation time much more.

Aprogiena:
> Also remember that when we broadcast a block to a peer we need to have that block in store thats why we wait for store before we broadcast.

As mentioned in the earlier discussion, I think this can be skipped if we mined the block. And it will save time.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: Your are effectively starting a background loop, so you use async loop with the run once property.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: ` this.blockEnqueued.Wait()` does not block?

Aprogiena:
> What  we need to do is push the block that is about to be broadcasted to  cache and then push it out immediately that will reduce the propagation  time much more.

I agree that implementing this cache would be the next logical step. @noescape00 wanted to separate that from the speedup.

Aprogiena:
I think you should look into Task.WhenAny, which allows you to wait on multiple tasks at once. Then you probably want to look about converting manual event into task, which you can do e.g. like this https://stackoverflow.com/a/18766131/3835864

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: i think it is worth it, but anyway, if you ever want to run task on background, use Task.Run, not factory start new

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: That does not make much sense. Instead of simple Task.Run (or not even that as I suggest) you want to construct AsyncLoop that won't even loop? Why would you do that?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Agreed. But only after we have a block store cache and we won't need to skip 1 iteration of announcing some blocks when we're trying to announce something we don't have on block store yet.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: It does and I think we want to remove it.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: I don't understand. I think 5 second "timeout" would be handy even without block store cache. Block store cache only allows us to merge cases 1 and 2, the batch case is case 3, so the cache does not affect that.

noescape00: What if we just received a block and mined another one on top of that 1 sec after? We will end up waiting 4 sec before it gets propagated.

Aprogiena:
So thinking about it little more, 
- we start with empty local list of items to propagate
- we wait for item signal (with shutdown cancellation token)
- an item is added
- our task is woken up
- we remove everything from the shared queue
- we check if there is a tip
- if there is, we send and cancel any timer we have setup
- if there is not, we install 5 second timer, unless it has been set up already
- we wait for item signal OR timer signal (with shutdown cancellation token)
- either timer signals -> we send our local list, empty it, go to step 2
- new item added -> go to step 5

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Why would we remove it? The whole reason for this is to block the task until we get a signal that new block arrived.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: no, see the alg i just posted

Aprogiena:
and let's say that "moving everything from shared queue to local list" checks whether those blocks are saved on disk as it is now (with not removing those that are not)

this does not solve not waiting for disk saving in case we mined the block as i wanted, but that can wait for next pr for that cache

noescape00:
>this does not solve not waiting for disk saving in case we mined the block as i wanted, but that can wait for next pr for that cache

Exactly. It's not supposed to solve this problem. So this pr will actually speed up the propagation only after block store cache is implemented. 

But if we implement a cache and don't have this PR we still will end up with <1 sec delay.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: because 
- it is not awaitable
- it does not implement the 5 second timer

well we could leave it and add linked cancellation token with 5 second timer and we could run that with Task.Run and not have it awaitable

but it seems inferior to have it fully async and not to occupy a thread

Aprogiena:
the alg could be modified to implement that store wait skipping, but let's start with simpler version

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: Constructing an asyncloop is not so hard and that's the pattern we use, it makes it easier to dispose of the loop and puts it under our control exception handling, life time etc...

When you just do a Task,Run especially on a standalone loop it become to top long running task, you then need to manage exceptions and manage the loop.

Also see my comment on the issue we could just reduce the loop time and still achieve significant improvement.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed (except for await inside the async loop, probably don't need it here)

dangershony:
> I think this can be skipped if we mined the block

how can you do this I did not understand, the peer that got the block will still ask for it and we need to serve it from store or cache.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Replaced with IAsyncLoop and commented.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

noescape00: >Also see my comment on the issue we could just reduce the loop time and still achieve significant improvement.

This will force us to send smaller batches which creates way more overhead which is unnecessary.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: @noescape00 when we block we hold a thread from the system, in this case its just 1 thread but generally its bad practice.

Aprogiena:
> how can you do this I did not understand, the peer that got the block will still ask for it and we need to serve it from store or cache.

this optimization is not necessary to implement as there will be that cache, but if there was not idea for that cache, this would improve the time of your mined block propagation - it was expected that you will manage to store that thing you mined to disk before the peer roundtrips to ask for it ;)

noescape00:
I also don't feel like we need skipping. Adding a block to the cache will take a few ms, we can afford that.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: well I suggest not to construct async loop or Task.Run
but for single run I would go with Task.Run

dangershony:
I think we are complicating this, i would like to suggest we just reduce the loop time and use cache.
If we use the asyncloop in 0.1 it will still be quite optimized.

Also I would suggest we extend the AsyncLoop to take a trigger task, then we could externally break the wait.
Then inside AsyncLoop we can Task.WaitAny(time, trigger) and then you get, as @Aprogiena requested, external interference.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: gentlemen's agreement was to use it everywhere in FN code
but that async loop i don't like

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: > This will force us to send smaller batches which creates way more overhead which is unnecessary.

i don't agree, the alg is designed to send batch if there is a tip (which is a must imho) or to wait 5 seconds and send whatever we have if nothing new comes

Aprogiena:
> I also don't feel like we need skipping. Adding a block to the cache will take a few ms, we can afford that.

yes i've mentioned that - with cache we don't need it and for now it can wait until we implement the cache

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Ok, I'll not push new commits until we all come to consensus on what to do =)

Though I also think that creating an AsyncLoop instance for a loopless task lacks logic. But it looks like we do it all around the codebase so if we agree to remove this one we also should do it for other usages of the AsyncLoop that create RunOnce 'loops'.

Aprogiena:
>  If we use the asyncloop in 0.1 it will still be quite optimized.

that's super ugly, active waiting is even worse than blocking a thread

optimal design might be little more complicated than naive approach, but remember why we are even refactoring this, what you say here could be done with one constant replacement, and it would bad decision

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Suddenly I felt your pain x D
Convince Dan to use your approach and I'll replace this =).

dangershony:
>that's super ugly, active waiting is even worse than blocking a thread

I tend to disagree, spinning a task every 0.1 seconds to check a list is quite elegant.

>optimal design might be little more complicated than naive approach

True, but it sounds like its going to be more complexity that can be easily solved with just some param twiking

noescape00:
>I tend to disagree, spinning a task every 0.1 seconds to check a list is quite elegant.

That will create 10 times more HeadersPayload messages sent comparing to a more complicated approach that is presented in this PR.

If you and @Aprogiena agree that it's worth it- ok, let's just change the constant.

noescape00:
Or we can make it the other way: 
Start the node with that 1 sec batch forming delay and as soon as we're synced (not In IBD && Headers.Height == BlockStore.Height) change the delay to 0.1 sec.

That way we won't need to send that 1300 blocks in 10 times more HeadersPayload messages AND keep it simple as it is now AND speed up new blocks propagation time.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: :D

Aprogiena:
> I tend to disagree, spinning a task every 0.1 seconds to check a list is quite elegant.

Ehm, OK, I will give you a chance to explain that :-)

Aprogiena:
> If you and @Aprogiena agree that it's worth it- ok, let's just change the constant.
&
> Or we can make it the other way

pls don't, I am quite sure about this one, but I will let Dan try to explain why he thinks active waiting is OK here (well or anywhere). 

the fact that in rest of the code base async loops are misused for things they should not be used at all is just telling us that the code base still needs lots of work, it can't be taken as a justification of creating more mess

we need real arguments for each possibility and compare pros and cons of each then to pick the winner

dangershony:
Why will it create 10 times more header payloads?
Once a block is found we propagate the entire pending list, the new suggest approach would create a payload per block which is the same.

Why is asynchronous loop misused?

What do you mean by active waiting?
The alternative approach is to await a signaling event right? Won't this be internally probing the signaler in fixed intervals?
Buy If you want to do that then it should be part of the asyncloop I think, passing a trigger to the asyncloop loop that will trigger the loop before the time elapsed.

noescape00:
>Why will it create 10 times more header payloads?

When we are not synced yet but we're not in IBD we start announcing blocks that we've received. Approximately 1300 blocks before we've synced we start doing that and sending headers of those blocks every 1 second. Reducing it to 0.1 sec will make 10 times more header payloads sent in this scenario. 

>Once a block is found we propagate the entire pending list, the new suggest approach would create a payload per block which is the same.

This is correct for cases when we're synced and we receive 1 new block every minute or mine one.

noescape00:
>Buy If you want to do that then it should be part of the asyncloop I think, passing a trigger to the asyncloop loop that will trigger the loop before the time elapsed.

Adding functionality to asyncloop  that will allow triggering a new iteration instantly and using it in this case sounds like a good solution to me. I believe I've also suggested this as a possible solution.

dangershony:
Ah you mean the period not in IBD and not at tip yet, yes we would send more inv messages with less headers in each payload.

Ok I guess then putting a trigger in the asyncloop is also my preferred approach, then complexity is kept hidden and reusable, and I believe apro will accept that as none active waiting.

Some sort of method like `asyncLoop.Trigger()`

Aprogiena:
well, not sure i'm happy with async loop trigger

it depends on how would we implement it because i can imagine at least two different implementations and thinking about it, one seem wrong and the other seems a lot like what i just described above, although mine is still better :-)

- the first one would be that Trigger just executes the task WITHOUT affecting the main loop mechanism, this is wrong because you could get the main loop executed in the middle of execution of this manually triggered one, that's asking for trouble

- the second one is harder to implement and is much better, and it means it will affect the main loop - so here you need to cancel this
```
                            await Task.Delay(refreshRate, cancellation).ConfigureAwait(false);
```
in the middle, so you will introduce manually reset event, so either that 5 sec interval completes or the event is signalled -> then you execute the handler, this is basically what i described before except that you are executing this (wastefully) every 5 seconds if the queue is empty. well doing a wasteful check every 5 seconds is not critical, we can afford it, it is just conceptually bad (it is still active wait, see below), but I can live with that (if the timer is 5 seconds, but not less), still would consider it inferior 

> I believe apro will accept that as none active waiting.

almost, not really happy about it, but much better idea than that of 0.1 second loops


> Why is asynchronous loop misused?

i don't think AL is a good fit, because AL currently is something that gives you the ability to repeat the same task in well defined intervals. this sounds like nothing we are dealing with here. if there is nothing in the queue, we should not do any work. that's the core of this mechanism, AL is raping this in the ass


> What do you mean by active waiting?

Active waiting is basically this - https://en.wikipedia.org/wiki/Busy_waiting

it is any kind of loop that does no useful work most of the time until a certain condition is met. it is conceptually wrong for most not-super-low-level constructs because we know better today and we know better because we have operating system that provides us mechanisms to do better


> The alternative approach is to await a signaling event right? Won't this be internally probing the signaler in fixed intervals?

no!!! (yes, three exclamation marks :-)) events, unlike active waiting, is not probing at all, event object only does work when it is signaled. until then, no work is wasted, which is exactly the point


in our code base here in stratis we have a LOT of active loops, probably all of them are wrong (conceptually) and some are easy to replace and some are hard to replace. i claim that this one is easy to replace and as such, it should be replaced

note that using of async loop does not imply that we are doing active waiting. no, if there is a periodic task - imagine cron on linux, something that you want to do every period, then it is not active waiting

i think we could classify a loop as active waiting, if most of the time it does nothing but check if a condition is met. On the other hand, if most of the time a work is done and only sometimes a condition is not met, then it is not an active wait

so what we have here? a loop that propagates new blocks, once we are synced the expected time of a block is 64 seconds, so if we fire a check if we have a block every second or two or five, most of the time, we would just do nothing because the block would not be there, that for me is obvious active loop, although as we implement it, it is not consuming 100 % CPU all the time as spin-wait, but it is still conceptually wrong.


i understand that @dangershony wants to hide the mechanisms of flow control and i am ok with that, so ok, we can create a new class for that, it probably should not be async loop though. what we have here is not a loop, it is event based thing with a timer, conceptually it is totally different use case

i'd suggest to make things right, even if it means we need to implement a new class. don't misuse async loop just because we have it. we don't need to use it everywhere just because it exists. it doesn't fit.

noescape00:
> i can imagine at least two different implementations 

I was thinking about the third one: 

1) If the task is not executing: execute it right away without waiting
2) if the task is executing: wait for it to finish and start next iteration right after current one is completed.

noescape00:
We finally came to a consensus on how to execute this therefore I'm closing this one.

PR #977, opened 2017-12-23
Added IsSynced to our peer's view and preventing mining until we have at least one synced connection. 

noescape00:
Also improved our view of peer's chain tip and set startAfter to 1 sec instead of 10 for the staking loop since such long waiting time is no longer required because we have inner waiting for the right conditions to start staking. 

Fix: #921

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -255,14 +255,14 @@ namespace Stratis.Bitcoin.Features.BlockStore
                         this.logger.LogTrace("Checking is the peer '{0}' can connect header '{1}'.", peer.RemoteSocketEndpoint, chainedBlock);
 
                         // Peer doesn't have a block at the height of our block and with the same hash?
-                        if (chainBehavior.PendingTip.FindAncestorOrSelf(chainedBlock) != null)
+                        if (chainBehavior.PendingTip?.FindAncestorOrSelf(chainedBlock) != null)
                         {
                             this.logger.LogTrace("Peer '{0}' does not have header '{1}'.", peer.RemoteSocketEndpoint, chainedBlock.Previous);
                             continue;
                         }
 
                         // Peer doesn't have a block at the height of our block.Previous and with the same hash?
-                        if (chainBehavior.PendingTip.FindAncestorOrSelf(chainedBlock.Previous) == null)
+                        if (chainBehavior.PendingTip?.FindAncestorOrSelf(chainedBlock.Previous) == null)
                         {
                             // Peer doesn't have this header or the prior one - nothing will connect, so bail out.
                             this.logger.LogTrace("Neither the header nor its previous header found for peer '{0}', reverting to 'inv'.", peer.RemoteSocketEndpoint);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -257,7 +257,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                         // Peer doesn't have a block at the height of our block and with the same hash?
                         if (chainBehavior.PendingTip?.FindAncestorOrSelf(chainedBlock) != null)
                         {
-                            this.logger.LogTrace("Peer '{0}' does not have header '{1}'.", peer.RemoteSocketEndpoint, chainedBlock.Previous);
+                            this.logger.LogTrace("Peer '{0}' already has header '{1}'.", peer.RemoteSocketEndpoint, chainedBlock.Previous);
                             continue;
                         }
 
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -311,7 +311,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     ChainedBlock chainedBlock = blocksToAnnounce.Last();
                     if (chainedBlock != null)
                     {
-                        if (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
+                        if (chainBehavior.PendingTip == null || chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
                         {
                             inventoryBlockToSend.Add(chainedBlock.HashBlock);
                             this.logger.LogDebug("Sending inventory hash '{0}' to peer '{1}'.", chainedBlock.HashBlock, peer.RemoteSocketEndpoint);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -311,7 +311,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     ChainedBlock chainedBlock = blocksToAnnounce.Last();
                     if (chainedBlock != null)
                     {
-                        if (chainBehavior.PendingTip == null || chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
+                        if ((chainBehavior.PendingTip == null) || (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null))
                         {
                             inventoryBlockToSend.Add(chainedBlock.HashBlock);
                             this.logger.LogDebug("Sending inventory hash '{0}' to peer '{1}'.", chainedBlock.HashBlock, peer.RemoteSocketEndpoint);
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -414,7 +414,7 @@ namespace Stratis.Bitcoin.Features.Miner
             return this.stakingLoop;
         }
 
-        ///<inheritdoc/>
+        /// <inheritdoc/>
         public void StopStake()
         {
             this.logger.LogTrace("()");
@@ -445,26 +445,18 @@ namespace Stratis.Bitcoin.Features.Miner
 
             while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
             {
-                // Do not start staking until connected to at least one peer.
-                while (!this.connection.ConnectedNodes.Any())
+                // Wait until we have at least one connected peer who's headers are synced with ours AND until not in IBD.
+                while (!this.connection.ConnectedNodes.Any(x => x.Behavior<ChainHeadersBehavior>().IsSynced()) ||
+                       this.initialBlockDownloadState.IsInitialBlockDownload())
                 {
-                    this.logger.LogTrace("Waiting to be connected with at least one network peer...");
+                    if (this.initialBlockDownloadState.IsInitialBlockDownload())
+                        this.logger.LogTrace("Waiting for IBD to complete...");
+                    else
+                        this.logger.LogTrace("Waiting to be connected with at least one synced network peer...");
+
                     await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
                 }
 
-                // Do not start staking if we're in IBD.
-                while (this.initialBlockDownloadState.IsInitialBlockDownload())
-                {
-                    this.logger.LogTrace("Waiting for IBD to complete...");
-                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
-                }
-
-                // Wait until we have at least one connected peer who's headers are synced with ours.
-                while (!this.connection.ConnectedNodes.Any(x => x.Behavior<ChainHeadersBehavior>().IsSynced()))
-                {
-                    this.logger.LogTrace("Waiting for at least one synced network peer...");
-                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
-                }
 
                 ChainedBlock chainTip = this.chain.Tip;
                 if (chainTip != this.consensusLoop.Tip)
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -408,7 +408,7 @@ namespace Stratis.Bitcoin.Features.Miner
             },
             this.stakeCancellationTokenSource.Token,
             repeatEvery: TimeSpan.FromMilliseconds(this.minerSleep),
-            startAfter: TimeSpans.TenSeconds);
+            startAfter: TimeSpans.Second);
 
             this.logger.LogTrace("(-)");
             return this.stakingLoop;
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -19,7 +19,7 @@ using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.Miner
-{  
+{
     /// <summary>
     /// <see cref="PosMinting"/> is used in order to generate new blocks. It involves a sort of lottery, similar to proof-of-work,
     /// but the chances of winning this lottery is proportional to how many coins you are staking, not on hashing power.
@@ -356,7 +356,7 @@ namespace Stratis.Bitcoin.Features.Miner
 
             this.rpcGetStakingInfoModel = new Miner.Models.GetStakingInfoModel();
         }
-        
+
         /// <inheritdoc/>
         public IAsyncLoop Stake(WalletSecret walletSecret)
         {
@@ -413,7 +413,7 @@ namespace Stratis.Bitcoin.Features.Miner
             this.logger.LogTrace("(-)");
             return this.stakingLoop;
         }
-        
+
         ///<inheritdoc/>
         public void StopStake()
         {
@@ -445,11 +445,21 @@ namespace Stratis.Bitcoin.Features.Miner
 
             while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
             {
-                while (!this.connection.ConnectedNodes.Any() || this.initialBlockDownloadState.IsInitialBlockDownload())
+                while (!this.connection.ConnectedNodes.Any())
                 {
-                    if (!this.connection.ConnectedNodes.Any()) this.logger.LogTrace("Waiting to be connected with at least one network peer...");
-                    else this.logger.LogTrace("Waiting for IBD to complete...");
+                    this.logger.LogTrace("Waiting to be connected with at least one network peer...");
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                }
 
+                while (this.initialBlockDownloadState.IsInitialBlockDownload())
+                {
+                    this.logger.LogTrace("Waiting for IBD to complete...");
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                }
+
+                while (!this.connection.ConnectedNodes.Any(x => x.Behavior<ChainHeadersBehavior>().IsSynced()))
+                {
+                    this.logger.LogTrace("Waiting for at least one synced network peer...");
                     await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
                 }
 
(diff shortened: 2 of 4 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -311,7 +311,7 @@ private Task ProcessSendCmpctPayload(NetworkPeer node, SendCmpctPayload sendCmpc
                     ChainedBlock chainedBlock = blocksToAnnounce.Last();
                     if (chainedBlock != null)
                     {
-                        if (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
+                        if (chainBehavior.PendingTip == null || chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
```
Aprogiena: pls add ()

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -445,11 +445,24 @@ public void StopStake()
...
+                while (this.initialBlockDownloadState.IsInitialBlockDownload())
                 {
-                    if (!this.connection.ConnectedNodes.Any()) this.logger.LogTrace("Waiting to be connected with at least one network peer...");
-                    else this.logger.LogTrace("Waiting for IBD to complete...");
+                    this.logger.LogTrace("Waiting for IBD to complete...");
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                }
 
+                // Wait until we have at least one connected peer who's headers are synced with ours.
+                while (!this.connection.ConnectedNodes.Any(x => x.Behavior<ChainHeadersBehavior>().IsSynced()))
```
Aprogiena: why is this separated from the first loop?

would it be not enough, in original code, to just add the condition to Any() and change the comment?

these should not be separated loops anyway, what if IBD says true and then false again?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -408,12 +408,12 @@ public IAsyncLoop Stake(WalletSecret walletSecret)
...
             repeatEvery: TimeSpan.FromMilliseconds(this.minerSleep),
-            startAfter: TimeSpans.TenSeconds);
+            startAfter: TimeSpans.Second);
 
             this.logger.LogTrace("(-)");
             return this.stakingLoop;
         }
-        
+
         ///<inheritdoc/>
```
Aprogiena: pls add space here after `///`

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -37,7 +37,7 @@ public class ChainHeadersBehavior : NetworkPeerBehavior
         public bool AutoSync { get; set; }
 
         /// <summary>
-        /// Information about the peer's announcement of its tip using "headers" message.
+        /// Our view of the peer's headers tip.
```
Aprogiena: you should preserve the information that we get this value from "headers" message

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -180,11 +180,13 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
                         // Use the ChainState.ConsensusTip property (not Chain.Tip)
                         // if the peer is behind/equal to our best height an empty array is sent back.
 
-                        if (!this.CanRespondToGetHeaders) break;
+                        if (!this.CanRespondToGetHeaders)
```
Aprogiena: we don't have such coding style policy, so this is just polluting the PR

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -213,6 +215,10 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
                             }
                         }
 
+                        // Set our view of peer's tip equal to last header that was sent to him.
```
Aprogiena: to it

Aprogiena: the last

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -213,6 +215,10 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
...
+                        if (headers.Headers.Count != 0)
+                            this.pendingTip = this.Chain.GetBlock(headers.Headers.Last().GetHash());
```
Aprogiena: what if GetBlock returns null?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -286,7 +292,7 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
                         if (pendingTipBefore != this.pendingTip)
                             this.logger.LogTrace("Pending tip changed to '{0}'.", this.pendingTip);
 
-                        if (this.pendingTip.ChainWork > this.Chain.Tip.ChainWork)
+                        if (this.pendingTip != null && this.pendingTip.ChainWork > this.Chain.Tip.ChainWork)
```
Aprogiena: () pls

Aprogiena: but pending tip can't be null here

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -320,7 +326,7 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
                             }
                         }
 
-                        ChainedBlock chainedPendingTip = this.Chain.GetBlock(this.pendingTip.HashBlock);
+                        ChainedBlock chainedPendingTip = this.pendingTip == null ? null : this.Chain.GetBlock(this.pendingTip.HashBlock);
```
Aprogiena: again, why this?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -389,10 +395,24 @@ public void TrySync()
...
         }
 
-        private ChainedBlock GetPendingTipOrChainTip()
+        /// <summary>
+        /// Determines if the peer's headers are synced with ours.
+        /// </summary>
+        /// <remarks>
+        /// It is possible that peer is in IBD even though he has all the headers so we can't assume with 100% certainty that peer is fully synced.
+        /// </remarks>
+        public bool IsSynced()
```
Aprogiena: return value not commented

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -389,10 +395,24 @@ public void TrySync()
             this.logger.LogTrace("(-)");
         }
 
-        private ChainedBlock GetPendingTipOrChainTip()
+        /// <summary>
+        /// Determines if the peer's headers are synced with ours.
+        /// </summary>
+        /// <remarks>
+        /// It is possible that peer is in IBD even though he has all the headers so we can't assume with 100% certainty that peer is fully synced.
```
Aprogiena: it has

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -389,10 +395,24 @@ public void TrySync()
...
+
+            return (this.pendingTip.Height >= this.chainState.ConsensusTip.Height &&
+                    this.pendingTip.FindAncestorOrSelf(this.chainState.ConsensusTip) != null);
+        }
+
+        public ChainedBlock GetPendingTipOrChainTip()
         {
-            this.pendingTip = this.pendingTip ?? this.chainState.ConsensusTip ?? this.Chain.Tip;
-            return this.pendingTip;
+            return this.pendingTip ?? this.chainState.ConsensusTip ?? this.Chain.Tip;
```
Aprogiena: eh?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -291,7 +291,7 @@ public string GetNodeStats()
                     "Peer:" + (node.RemoteInfo() + ", ").PadRight(LoggingConfiguration.ColumnLength + 15) +
                     (" connected" + " (" + (connectionManagerBehavior.Inbound ? "inbound" : "outbound") + "),").PadRight(LoggingConfiguration.ColumnLength + 7) +
                     (" agent " + agent + ", ").PadRight(LoggingConfiguration.ColumnLength + 2) +
-                    " height=" + chainHeadersBehavior.PendingTip.Height);
+                    " height=" + chainHeadersBehavior.GetPendingTipOrChainTip().Height);
```
Aprogiena: again why?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -389,10 +395,24 @@ public void TrySync()
...
+        /// </summary>
+        /// <remarks>
+        /// It is possible that peer is in IBD even though he has all the headers so we can't assume with 100% certainty that peer is fully synced.
+        /// </remarks>
+        public bool IsSynced()
+        {
+            if (this.pendingTip == null)
+                return false;
+
+            return (this.pendingTip.Height >= this.chainState.ConsensusTip.Height &&
```
Aprogiena: add () pls

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -389,10 +395,24 @@ public void TrySync()
...
+                    this.pendingTip.FindAncestorOrSelf(this.chainState.ConsensusTip) != null);
```
Aprogiena: why this condition?
would it not be better to check ChainWork?

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: But it looks nicer with 'break' on the next line, isn't it? 
I can revert it if you want though.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: why? I think it can. 
Since we no longer set pending tip to be equal to our tip at the very start it is possible that pending tip could be null here.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: 1) 'GetPendingTipOrChainTip' - the name of the function tells that it will return the value, not modify it. It used to have a logical error. 
2) By setting the tip to our tip we make a false image of the peer's tip.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: Because here GetPendingTipOrChainTip() can return our tip if the pending one is null. And PendingTip now can be null if the peer didn't send us headers or we didn't send them to the peer.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Good idea. Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: >Since we no longer set pending tip to be equal to our tip at the very start it is possible that pending tip could be null here.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: This is more of personal preference. So you prefer the break on the next line. I don't mind having either. But let's say if I preferred to have it on the same line. Then if I touched the code in next PR I would revert it, then if you touched the code you would revert what I reverted and so on :-) 

Therefore I tend not to fix anything that is just my personal preference and is not part of the code policy - because exactly we could play this ping pong all the time and it just pollutes the PR. So even if you think something is nicer, if it is not in code style, rather don't fix it.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: This has some serious consequences, I did not really expect this to be part of this PR. Need to think about all changes you made again now ...

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
Aprogiena: Now how this is not misleading - if you said in above comment that we make a false image of peer's tip, this is exactly where you revert that logic again to that misleading value. It does not make sense to me. Either you want to really redefine the pending tip to hold the actual value, if we have it, or null, if we don't, then you should display here "unknown" if it is null and not this.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: OK, I agree the original code did have that misleading logic, I just did not expect it to be fixed on this PR, that what confused me everywhere.

So if we are about to fix this logic in this PR, we should really be strict (including that in ConnectionManager console display). So I will review again with this in mind that we want to fix that logic.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: OK I see, it is the new thing. So () pls here.

Aprogiena:
OK, so I now understand the part of this PR is fixing the logic of pending tip, I need to rereview this with this in mind. Did not expect it :-)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -311,7 +311,7 @@ private Task ProcessSendCmpctPayload(NetworkPeer node, SendCmpctPayload sendCmpc
                     ChainedBlock chainedBlock = blocksToAnnounce.Last();
                     if (chainedBlock != null)
                     {
-                        if (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null)
+                        if ((chainBehavior.PendingTip == null) || (chainBehavior.PendingTip.GetAncestor(chainedBlock.Height) == null))
```
Aprogiena: so now if I think about this in context of PendingTip can be null, then you need to make sure that all its uses are handled as such and you don't do that, so you should do it everywhere

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: OK, if we want to fix that, probably this method should be deleted completely

Aprogiena:
OK, I've reconsidered this and if we want to do it here, we should remove GetPendingTipOrChainTip from the code base. And we should handle null case of pendingTip everywhere as a special case.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Reasonable. 
Fixed.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: The thing is that after we're synced and connected to several nodes we get half of them at the "unknown" state because no headers were transferred between. So here its actually logical go assume that we and peer are at the same height.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: Already added before.

noescape00:
>OK, I've reconsidered this and if we want to do it here, we should remove GetPendingTipOrChainTip from the code base. And we should handle null case of pendingTip everywhere as a special case.

GetPendingTipOrChainTip is used only for the inner logic of CHB, I made it public. So I think we can get away with making it private and  handle null case of pendingTip in the only place we use it- on the console log output.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
Aprogiena: i think having unknown is much better than assumed information that may or may not be anywhere close to the real thing

Aprogiena:
i don't think making it private will help, the whole method does not make sense, it gives some more or less random number

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: Ok, reasonable. I'll turn it back to "unknown" if the pending tip is null.

noescape00:
>i don't think making it private will help, the whole method does not make sense, it gives you some more or less random number

Well, it's created just to support the inner logic so I assume the writer of this method had a good idea why we're using it instead of just our view of peer's tip even if it's null. If we decide to remove it it should be done outside of scope of this PR anyway.

Aprogiena:
well i thought whole this effort is outside of the scope of this PR, i only expected here to implement that is synced method + that extra pending tip update on getheaders

so if you can remove the extra logic for this PR, we are good and we can remake the rest completely separately

but if we want to finish it in this PR, I would suggest we really fix that logic

noescape00:
>but if we want to finish it in this PR, I would suggest we really fix that logic

Ok.

noescape00:
Actually the GetPendingTipOrChainTip is there for a very good reason. We use it or getting a locator at the sync attempt and in 2 more  places where we actually want to get our tip if pending one is null. 
So it's not possible to remove this method without copy-pasting the logic into 3 different places. 
I think we should just leave it private.

Aprogiena:
in HeadersPayload case - first use, it should not be used imho

in second use there i also think it should not be used - i even think we can now delete the TrySync from there because that logic is now implemented above in the foreach - try sync only sends GetHeaders message to the peer, which we do under that `Previous header of the new header '{0}' was not found` log
so this seems duplicate and unnecessary (and i actually probably should have removed it in my PR where i introduced this new code)


in that third case yes, there should be such a logic for those locators

so to me it seems only one use is justified, so it would be DRY

noescape00:
I've removed that GetPendingTipOrChainTip. 
I'll need to test it very well though before I can consider this to be a harmless change.

Aprogiena:
I agree but one more change is needed. 
```
                                    // If we can't connect the header we received from the peer, we might be on completely different chain or
                                    // a reorg happened recently. If we ignored it, we would have invalid view of the peer and the propagation
                                    // of blocks would not work well. So we ask the peer for headers using "getheaders" message.
                                    var getHeadersPayload = new GetHeadersPayload()
                                    {
                                        BlockLocators = pendingTipBefore.GetLocator(),
                                        HashStop = null
                                    };
```

here pendingTipBefore can be null
i think we should create private method that would either construct the whole GHP or just locator and it should be used here as well as in TrySync method, so something like
```
GetHeadersPayload getHeadersPayload = this.GetPendingTipHeadersPayload();
peer.SendMessageVoidAsync(getHeadersPayload);
```

this method would construct GHP like this:
```
return new GetHeadersPayload()
{
    BlockLocators = (this.pendingTip ?? this.chainState.ConsensusTip ?? this.Chain.Tip).GetLocator(),
    HashStop = null
};
```

noescape00:
@Aprogiena I've added comment. 
Also it looks like we should keep that TrySync. Without it I can't sync properly on a fresh node.

noescape00:
Ok, I've tested it in various different scenarios- looks good.

PR #978, opened 2017-12-23
Refactor/the node job part 17

Aprogiena:
This is another big win for the blue team, although there was a lot of blood with this one. It was supposed to be a simple and small PR that would just do one thing - remove the duplicate code that handled the handshake from the peer from NetworkPeerServer and have just NetworkPeer to handle about reading and writing messages. 

Well, that did not happen because I was revealing very ugly things (many of them need yet to be rewritten in later PRs), and I took me a week to find all bugs and problems and I just hope I get it right now.

So this PR contains
- some node-peer renaming, you can mostly ignore that on review, it is boring stuff (that affects most of the changed files though)
- NetworkPeer is Disposable, it needs to be Disposed or Disconnected!!! 💀 !!! 💀💀 !!! 💀💀💀 - we have been leaking peers all over the place, now this should be fixed everywhere
- there were different methods on how to disconnect peers, now there are just two (well three) - Dispose (2x) and Disconnect, and Dispose methods just call Disconnect, so at the core there is just one, but I think we want to have Dispose for the class to be IDisposable and sometimes we call Dispose, sometimes we call Disconnect, depending on the context - so this is just for the better readability of the code; so this disconnecting stuff is now very much simplified
- every disconnection is now with reason - you always have a reason; there is version of Dispose without reason, then default reason is used, but all code where it made sense has now a reason
- so because we want always to dispose/disconnect, we need to keep track of all network peers once they are constructed, so this is added everywhere (i'm looking at you peer connector)
- the core of this PR was supposed to be refactoring of NetworkPeerServer, so this is done - we completely remove message producer and message listener from this class as well as message processing, this logic is merged with NetworkPeer/NetworkPeerConnection functionality - this simplifies NetworkPeerServer dramatically!
- NetworkPeer shutdown sequence was redesigned and simplified - no longer we fire state change event handlers multiple times, which produces weird state transitions that should not be there in the first place


Aprogiena:
One more thing - don't try to follow this on commit-by-commit basis. All commits ARE for this PR and they are mess. Multiple times I did have to touch and rework some parts of the code, so ... just don't try that, for your own good.

dangershony:
What are you suggesting to review the entire thing?
I will need some tie and now the holidays are in so this might take some time.

Aprogiena:
I suggest you take it file by file. First, you can go with "everything else", which means all files except NetworkPeer, NetworkPeerServer and factory and peer connector files. That is just renaming and disconnect reason stuff, should be trivial. Then you can do factory and peer connector files, slightly more difficult, but still quite trivial. Then NetworkPeerServer - mostly just reduction of code, so still easy. Then finally NetworkPeer.

Aprogiena:
i would also suggest to ask a lot of question rather than trying to figure out something, there can be quite hidden reasons, so feel free to ask a lot and make this review more interactive, it is expected

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -1048,10 +1162,24 @@ public InventoryType AddSupportedOptions(InventoryType inventoryType)
             return inventoryType;
         }
 
+        /// <summary>
+        /// Disconnects the peer and cleans up.
+        /// </summary>
+        /// <param name="reason">Human readable reason for disconnecting.</param>
+        /// <param name="exception">Exception because of which the disconnection happened, or <c>null</c> if there were no exception.</param>
+        public void Dispose(string reason, Exception exception = null)
```
noescape00: I don't think that dispose should have 'reason' or 'exception' as input values. This is not a classic and expected implementation of Dispose. 

I suggest replacing every usage of this dispose with disconnect method and when the object is actually being disposed add checking if peer is connected. So it will be like: 

```
   public void Dispose()
        {
            if (this.Connected)
             this.Disconnect(reason: "Peer was disposed");

        }
```

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -214,12 +235,17 @@ public NetworkPeerConnection(NetworkPeer peer, NetworkPeerClient client, IDateTi
...
-                    this.Peer.State = NetworkPeerState.Failed;
-                    this.Peer.DisconnectReason = new NetworkPeerDisconnectReason()
+                    if (this.Peer.DisconnectReason == null)
                     {
-                        Reason = "Unexpected exception while sending a message",
-                        Exception = ex
-                    };
+                        this.Peer.DisconnectReason = new NetworkPeerDisconnectReason()
+                        {
+                            Reason = "Unexpected exception while sending a message",
```
noescape00: Add '.' here.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: Actually it looks like you already have a Dispose overload that does that so just removing 'Dispose(string reason, Exception exception = null)' would do.

noescape00: Or if you absolutely need to do both things- rename this dispose overload to "DisconnectAndDispose"

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -104,7 +92,6 @@ public NetworkPeerServer(Network network, IPEndPoint localEndpoint, IPEndPoint e
             this.dateTimeProvider = dateTimeProvider;
             this.networkPeerFactory = networkPeerFactory;
 
-            this.AllowLocalPeers = true;
             this.InboundNetworkPeerConnectionParameters = new NetworkPeerConnectionParameters();
 
             this.LocalEndpoint = Utils.EnsureIPv6(localEndpoint);
```
noescape00: Two 'param name' tags are not corresponding the variable names used in this constructor: 'localEndpoint' and 'externalEndpoint'

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -457,8 +506,8 @@ public bool IsConnected
         /// <summary>Version message payload received from the peer.</summary>
         public VersionPayload PeerVersion { get; private set; }
 
-        /// <summary>Set to <c>1</c> if the peer disconnection has been initiated, <c>0</c> otherwise.</summary>
-        private int disconnecting;
+        /// <summary>Set to <c>1</c> if the peer disconnection has been initiated, <c>0</c> otherwise.</summary>
```
noescape00: Consider changing the comment as well. If you changed 'disconnecting' to 'disconnected' then probably comment should be like: 

`Set to <c>1</c> if the peer is disconnected, <c>0</c> otherwise.`

noescape00:
Overall this looks solid to me, GJ!
Tested it- works well, didn't encounter any bugs so far.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: I was thinking about it and I have not found a perfect solution for this. I at least tried to distinguish where to use which method. 

I agree it is not optimal, I was thinking about alternative version, but then decided for this one. The alternative version was that you would specify (in constructor?) the default disconnection message for the particular peer instance. So then you would just call dispose. But then you wouldn't be able to pass exception.

If you check the code that non-parametric Dispose is doing just what you are asking for, so that is there.

I'd be happy to get more opinions on this as I'm undecided which way to take here.

Aprogiena: However, I don't hink DisconnectAndDispose is any better than just parametric Dispose.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: I have removed dots from these reasons because of how these reasons are displayed to the console. The original code also lack those dots.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
Aprogiena: I don't understand, please explain

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: I actually think the previous name was a mistake. The value of 1 does only mean that somewhere in the past it has been initiated. It does not mean the action was still in progress. So the comment was right, just the name of the variable seemed wrong.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: Note that the default Dispose is there and you can (and it is used in tests) use `using` clause for this and it will work as expected.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
noescape00: The xml doc refers param names that do not exist. Someone changed names of the constructor's variables and forgot about the xml doc changes.

Aprogiena: aah, thank you i see it now, will fix

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -261,13 +287,16 @@ public void StartReceiveMessages()
 
                     this.Peer.LastSeen = this.dateTimeProvider.GetUtcNow();
                     this.Peer.Counter.AddRead(message.MessageSize);
-                    this.Peer.OnMessageReceived(new IncomingMessage()
+
+                    var incommingMessage = new IncomingMessage()
```
fassadlr: `incomingMessage` not `incommingMessage` ;)

Aprogiena: thanks, fixed

Aprogiena:
@dangershony or @bokobza kindly please prioritize this PR, I'm now fixing new conflicts every so often as this touches quite a few files

dangershony:
Doing it now

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -143,23 +143,43 @@ public class NetworkPeerConnection : IDisposable
...
+            this.setPeerStateOnShutdown = NetworkPeerState.Offline;
             this.Peer = peer;
             this.Client = client;
             this.CancellationSource = new CancellationTokenSource();
-            this.cancelRegistration = this.CancellationSource.Token.Register(this.InitiateShutdown);
+
+            // When the cancellation source is cancelled, the registered callback is executed within 
+            // the context of the thread that invoked the cancellation. However, we need InitiateShutdown to be 
+            // called in separation of that to avoid deadlock.
+            this.cancelRegistration = this.CancellationSource.Token.Register(() => Task.Run(() => this.InitiateShutdown()));
```
dangershony: This means you lose the task of the shutdown processes.
If an exception happens in the `InitiateShutdown` where does it propagate?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -342,6 +382,18 @@ public void Dispose()
         }
     }
 
+    /// <summary>
```
dangershony: Very nice

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -417,8 +417,9 @@ public void RemoveNodeAddress(IPEndPoint ipEndpoint)
             });
 
             NetworkPeer peer = await this.NetworkPeerFactory.CreateConnectedNetworkPeerAsync(this.Network, ipEndpoint, cloneParameters).ConfigureAwait(false);
+            this.AddConnectedPeer(peer);
             this.peerAddressManager.PeerAttempted(ipEndpoint, this.dateTimeProvider.GetUtcNow());
-            await peer.VersionHandshakeAsync().ConfigureAwait(false);
+            await peer.VersionHandshakeAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
```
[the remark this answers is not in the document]
Aprogiena: no, peer class does not operate with node lifetime, nbitcoin does not have it, so components migrated from there don't have it unless you add it, i haven't added 

we can add it if we want, but it does not seem necessary, but still it is possible if we really want

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: this is reworked in part 18 as i have learned a lot about cancellation tokens over the past weeks

yes, you would lose the task, but the dispose code is depending on it, so if it crashed, you'd be fked up (node would crash), but because of that dependency, it is not needed to hold a reference to that task. anyway, reworked in p18

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerConnectionParameters.cs
```diff
@@ -46,6 +46,7 @@ public NetworkPeerConnectionParameters()
             this.SendBufferSize = 1000 * 1000;
             this.UserAgent = VersionPayload.GetNBitcoinUserAgent();
             this.PreferredTransactionOptions = NetworkOptions.TemporaryOptions;
+            this.Nonce = RandomUtils.GetUInt64();
```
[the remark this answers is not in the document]
Aprogiena: it was set somewhere else, there was some kind of problem with this and reason why i put it here, but can't remember now

i think it was just not initialized early enough or something like that

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -135,11 +135,6 @@ public NetworkPeer CreateNetworkPeerClient()
             return this.networkPeerFactory.CreateConnectedNetworkPeerAsync(Network.RegTest, "127.0.0.1:" + this.ports[0].ToString()).GetAwaiter().GetResult();
         }
 
-        public NetworkPeer CreateNodeClient(NetworkPeerConnectionParameters parameters)
```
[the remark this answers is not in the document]
Aprogiena: well if i deleted it and the code still compiles ... it wasn't used :)

PR #980, opened 2017-12-25
AsyncLoop trigger that allows manual launch of next iteration skipping the reapeatAfter delay. 

noescape00:
This PR allows manual triggering of the next iteration in case we need to. 

Also added a test.


Aprogiena:
i've mentioned my objections on slack
- dispose is missing
- linked cancellation token should not be used for this, async event should 

but one thing i did not mention and that's also important

i think we should reject any such PRs extending a functionality of the class if there is no use for that functionality. we have many many classes that do generic something and can be extended to do even more. but since we don't use that extra something, it would just be extra code in the code base, unused

so unless something is at least intended to use this new feature of async loop introduced here i would not be happy for it to be merged

PR #982, opened 2017-12-26
Revert "Added IsSynced to our peer's view and preventing mining until we have at least one synced connection. "

Aprogiena:
Reverts stratisproject/StratisBitcoinFullNode#977


Aprogiena:
@dangershony sorry for this, I thought the issue we were seeing is unrelated to this PR, but I can't reproduce it without it, so it seems it is related :-(

Aprogiena:
@noescape00 convinced me not to revert :|

PR #984, opened 2017-12-26
Async consumer queue

noescape00:
This PR adds a new utility class- AsyncConsumerQueue<T>
AsyncConsumerQueue is a queue of items that are processed when a certain condition is met or when a timer runs out. 

Primal usage of this is for cases when we need to periodically flush items (send them to peer or save to hard drive) with a certain interval. But if a certain condition is met we need to flush it right away without waiting. 

This will be used in 'BlockStoreSignaled' for announcing new headers.

### DIFF
--- a/src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
@@ -1,0 +1,269 @@
+﻿/*
+The MIT License (MIT)
+
+Copyright (c) 2014 StephenCleary
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+https://github.com/StephenCleary/AsyncEx
+ */
+
+using System;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// An async-compatible manual-reset event.
+    /// </summary>
+    public sealed class AsyncManualResetEvent
+    {
+        /// <summary>
+        /// The object used for synchronization.
+        /// </summary>
+        private readonly object mutex;
+
+        /// <summary>
+        /// The current state of the event.
+        /// </summary>
+        private TaskCompletionSource<object> tcs;
+
+        /// <summary>
+        /// Creates an async-compatible manual-reset event.
+        /// </summary>
+        /// <param name="set">Whether the manual-reset event is initially set or unset.</param>
+        public AsyncManualResetEvent(bool set)
+        {
+            this.mutex = new object();
+            this.tcs = CreateAsyncTaskSource<object>();
+            if (set)
+                this.tcs.TrySetResult(null);
+        }
+
+        /// <summary>
+        /// Creates an async-compatible manual-reset event that is initially unset.
+        /// </summary>
+        public AsyncManualResetEvent()
+            : this(false)
+        {
+        }
+
+        /// <summary>
+        /// Whether this event is currently set. This member is seldom used; code using this member has a high possibility of race conditions.
+        /// </summary>
+        public bool IsSet
+        {
+            get
+            {
+                lock (this.mutex)
+                {
+                    return this.tcs.Task.IsCompleted;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously waits for this event to be set.
+        /// </summary>
+        public Task WaitAsync()
+        {
+            lock (this.mutex)
+            {
+                return this.tcs.Task;
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously waits for this event to be set or for the wait to be canceled.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            Task waitTask = this.WaitAsync();
+            if (waitTask.IsCompleted)
+                return waitTask;
+
+            return WaitAsync(waitTask, cancellationToken);
+        }
+
+        /// <summary>
+        /// Synchronously waits for this event to be set. This method may block the calling thread.
+        /// </summary>
+        public void Wait()
+        {
+            WaitAndUnwrapException(this.WaitAsync());
+        }
+
+        /// <summary>
+        /// Synchronously waits for this event to be set. This method may block the calling thread.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public void Wait(CancellationToken cancellationToken)
+        {
+            Task ret = this.WaitAsync();
+            if (ret.IsCompleted)
+                return;
+
+            WaitAndUnwrapException(ret, cancellationToken);
+        }
+
+        /// <summary>
+        /// Sets the event, atomically completing every task returned by <see cref="O:Nito.AsyncEx.AsyncManualResetEvent.WaitAsync"/>. If the event is already set, this method does nothing.
+        /// </summary>
+        public void Set()
+        {
+            lock (this.mutex)
+            {
+                this.tcs.TrySetResult(null);
+            }
+        }
+
+        /// <summary>
+        /// Resets the event. If the event is already reset, this method does nothing.
+        /// </summary>
+        public void Reset()
+        {
+            lock (this.mutex)
+            {
+                if (this.tcs.Task.IsCompleted)
+                    this.tcs = CreateAsyncTaskSource<object>();
+            }
+        }
+
+        /// <summary>
+        /// Creates a new TCS for use with async code, and which forces its continuations to execute asynchronously.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result of the TCS.</typeparam>
+        private static TaskCompletionSource<TResult> CreateAsyncTaskSource<TResult>()
+        {
+            return new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        private static void WaitAndUnwrapException(Task task)
+        {
+            if (task == null)
+                throw new ArgumentNullException(nameof(task));
+            task.GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        /// <param name="cancellationToken">A cancellation token to observe while waiting for the task to complete.</param>
+        /// <exception cref="OperationCanceledException">The <paramref name="cancellationToken"/> was cancelled before the <paramref name="task"/> completed, or the <paramref name="task"/> raised an <see cref="OperationCanceledException"/>.</exception>
+        private static void WaitAndUnwrapException(Task task, CancellationToken cancellationToken)
+        {
+            if (task == null)
+                throw new ArgumentNullException(nameof(task));
+            try
+            {
+                task.Wait(cancellationToken);
+            }
+            catch (AggregateException ex)
+            {
+                throw PrepareForRethrow(ex.InnerException);
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously waits for the task to complete, or for the cancellation token to be canceled.
+        /// </summary>
+        /// <param name="task">The task to wait for. May not be <c>null</c>.</param>
+        /// <param name="cancellationToken">The cancellation token that cancels the wait.</param>
+        private static Task WaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            if (task == null)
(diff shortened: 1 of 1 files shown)


noescape00:
>well then you have different idea than i do :-)
i think you are mixing two things - one is triggering the event, which imho should be done on every add, that's the primary purpose of the (generic) consumer - and the other is sending the headers, but we won't send on each event, but we will process the queue

Since the queue is an inner object adding a new item to it is 'consumption' already. 
Adding triggers starting the timer for processing (if not started already) and performs a check if we should process it without waiting for the timer.

noescape00:
>we would achieve the same result, just the extra condition you have there would imho make the code harder to read
i am sure we don't need any such thing, so why to have it?

Which condition?

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            : this(false)
+        {
+        }
+
+        /// <summary>
+        /// Whether this event is currently set. This member is seldom used; code using this member has a high possibility of race conditions.
+        /// </summary>
+        public bool IsSet
+        {
+            get { lock (this.mutex) return this.tcs.Task.IsCompleted; }
```
Aprogiena: we do use {} when we use locks

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously waits for this event to be set.
+        /// </summary>
+        public Task WaitAsync()
+        {
+            lock (this.mutex)
```
Aprogiena: just like this, pls make IsSet code style consistent with this function

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously waits for this event to be set or for the wait to be canceled.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            var waitTask = this.WaitAsync();
```
Aprogiena: unvar

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            if (waitTask.IsCompleted)
+                return waitTask;
+            return waitTask.WaitAsync(cancellationToken);
```
Aprogiena: empty line after return

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            this.WaitAsync().WaitAndUnwrapException();
+        }
+
+        /// <summary>
+        /// Synchronously waits for this event to be set. This method may block the calling thread.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public void Wait(CancellationToken cancellationToken)
+        {
+            var ret = this.WaitAsync();
```
Aprogiena: unvar

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            if (ret.IsCompleted)
+                return;
```
Aprogiena: empty line after return

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            lock (this.mutex)
+            {
+                if (this.tcs.Task.IsCompleted)
+                    this.tcs = TaskExtensions.CreateAsyncTaskSource<object>();
+            }
+        }
+    }
+
+    public static class TaskExtensions
```
Aprogiena: remove this class completely

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        private static async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            using (var cancelTaskSource = new CancellationTokenTaskSource<object>(cancellationToken))
+                await await Task.WhenAny(task, cancelTaskSource.Task).ConfigureAwait(false);
+        }
+    }
+
+    internal static class ExceptionHelpers
```
Aprogiena: remove this class completely

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+                throw ExceptionHelpers.PrepareForRethrow(ex.InnerException);
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously waits for the task to complete, or for the cancellation token to be canceled.
+        /// </summary>
+        /// <param name="this">The task to wait for. May not be <c>null</c>.</param>
+        /// <param name="cancellationToken">The cancellation token that cancels the wait.</param>
+        public static Task WaitAsync(this Task @this, CancellationToken cancellationToken)
```
Aprogiena: rename `@this` to `task`

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            if (@this == null)
+                throw new ArgumentNullException(nameof(@this));
+
+            if (!cancellationToken.CanBeCanceled)
+                return @this;
```
Aprogiena: empty line after return

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
```
Aprogiena: empty line after return

Aprogiena:
> That's a good question if we need it or not.

I believe we don't.

> Ofc we can perform that check on the user side before\after we add item to the queue and if true we would use smth like queue.TriggerProcessing but it would look uglier. And imagine if we add item to the queue from different places- we would need to copy that condition.

no, you would not need to copy that condition

it seems that we still fundamentally thing differently about the design, therefore this seems to be to early to make an PR from it

> So it's better to have it here just in case it's needed. And if not- ok, just don't set any condition- that's fine.

That i consider ugly and unnecessary, just as with that manual trigger on async loop. If we won't use it, don't implement it.

src/Stratis.Bitcoin.Tests/Utilities/AsyncConsumerQueueTest.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            // 45 after a timer.
+            Assert.Equal(45, summ);
+            processedQueue.Dispose();
+        }
+
+        [Fact]
+        public async void TimerRestartsIfNotAllItemsWereConsumedAsync()
+        {
+            int summ = 0;
```
Aprogiena: i belive standard abbreviation for summation is just sum with one m

src/Stratis.Bitcoin.Tests/Utilities/AsyncConsumerQueueTest.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            for (int i = 0; i < 10; i++)
+            {
+                processedQueue.Enqueue(i);
+                await Task.Delay(10);
+            }
+
+            await Task.Delay(1000);
+
+            //All items are eventually processed.
```
Aprogiena: space after `//`

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
```diff
@@ -0,0 +1,154 @@
...
+                    if (this.executor == null || this.executor.IsCompleted)
+                        this.executor = ExecutionFlowAsync();
+                }
+
+            }).ConfigureAwait(false);
+        }
+
+        private void SetTimer()
+        {
+            this.timerAwaiterTask = Task.Delay(this.timer.Value, this.cancellation.Token);
```
Aprogiena: why would you not use Timer object, which is resetable, while cancellation is not
it seems that you again misuse cancellation for something it is not designed for
you don't need timer task at all imho

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
```diff
@@ -0,0 +1,154 @@
...
+            this.timer = timer;
+            this.restartTimerIfNotAllItemsConsumed = restartTimerIfNotAllItemsConsumed;
+
+            this.trigger = new AsyncManualResetEvent(false);
+            this.cancellation = new CancellationTokenSource();
+            this.lockObject = new object();
+        }
+
+        /// <inheritdoc />
+        public void Enqueue(T item)
```
Aprogiena: i don't think this implementation of enqueue is good

why would you run task on the background for this, it does not make sense to me

who don't you just enqueue the item and set the event on which the consumer task is waiting
that's literally two lines of code (+logs)

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// <summary>Task that waits until specified delay runs out.</summary>
+        private Task timerAwaiterTask;
+
+        /// <summary>Task that runs execution flow.</summary>
+        private Task executor;
+
+        /// <summary>Cancellation token source.</summary>
+        private CancellationTokenSource cancellation;
+
+        private object lockObject;
```
Aprogiena: this can be readonly

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            // Timer is no longer needed.
+            this.timerAwaiterTask = null;
+
+            if (this.restartTimerIfNotAllItemsConsumed &&
+                (this.timer != null) &&
+                (this.Count != 0) &&
+                !this.cancellation.IsCancellationRequested)
+            {
+                // Restart timer and recursively call execution flow.
```
Aprogiena: recursively!?

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            // The code cannot ever get here. We just return a value to work around a badly-designed API (ExceptionDispatchInfo.Throw):
+            //  https://connect.microsoft.com/VisualStudio/feedback/details/689516/exceptiondispatchinfo-api-modifications (http://www.webcitation.org/6XQ7RoJmO)
+            return exception;
+        }
+    }
+
+    /// <summary>
+    /// Holds the task for a cancellation token, as well as the token registration. The registration is disposed when this instance is disposed.
+    /// </summary>
+    public sealed class CancellationTokenTaskSource<T> : IDisposable
```
Aprogiena: we also should consider removing this class

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.
Moved functionality of this class to the ResetEvent as private functions.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.
Moved functionality of this class to the ResetEvent as private functions.

src/Stratis.Bitcoin.Tests/Utilities/AsyncConsumerQueueTest.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Tests/Utilities/AsyncConsumerQueueTest.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
noescape00: Timer is not awaitable.

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
noescape00: >why would you run task on the background for this, it does not make sense to me

because in case 'immediateProcessingTrigger' condition is heavy we don't wan't to lock current thread checking it.

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
noescape00: Yes. 
I could have made a loop here but it looks like in most cases this loop will have only 1 iteration.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: It is used in AsyncManualResetEvent so we need to keep it.

noescape00:
Closing this because a separate PR for AsyncManualResetEvent is requested. Will split this one in 2 sequential pieces.

src/Stratis.Bitcoin/Utilities/AsyncConsumerQueue.cs
Aprogiena: why does it matter?

PR #987, opened 2017-12-27
Refactor/chainheadersbehavior messagereceived

### LINKED ISSUE #856: [ChainHeadersBehavior] Refactor message received
As suggested in this comment https://github.com/stratisproject/StratisBitcoinFullNode/pull/853#discussion_r154658338

Aprogiena:
- separate message handling to methods
- make it closer to async
- add some comments

Fix #856

### DIFF
--- a/src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
+++ b/src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Connection;
@@ -110,12 +111,12 @@ namespace Stratis.Bitcoin.Base
         {
             this.logger.LogTrace("()");
 
-            this.refreshTimer = new Timer(o =>
+            this.refreshTimer = new Timer(async (o) =>
             {
                 this.logger.LogTrace("()");
 
                 if (this.AutoSync)
-                    this.TrySync();
+                    await this.TrySyncAsync().ConfigureAwait(false);
 
                 this.logger.LogTrace("(-)");
             }, null, 0, (int)TimeSpan.FromMinutes(10).TotalMilliseconds);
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -154,195 +155,260 @@ protected override void DetachCore()
...
+
+            HeadersPayload headers = new HeadersPayload();
+            ChainedBlock consensusTip = this.chainState.ConsensusTip;
+            consensusTip = this.Chain.GetBlock(consensusTip.HashBlock);
 
-                case HeadersPayload newHeaders:
+            ChainedBlock fork = this.Chain.FindFork(getHeadersPayload.BlockLocators);
+            if (fork != null)
+            {
+                if ((consensusTip == null) || (fork.Height > consensusTip.Height))
```
mikedennis: Consider adding this to the above if condition as an "and" then you can remove the second ```if (fork != null)``` below

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -154,195 +155,260 @@ protected override void DetachCore()
...
-
-                        if (!this.CanSync)
+                        if (header.Height > consensusTip.Height)
                             break;
 
-                        if (newHeaders.Headers.Count == 0)
-                        {
-                            this.logger.LogTrace("Headers payload with 0 headers was received. Assume we're synced with the peer.");
+                        headers.Headers.Add(header.Header);
+                        if ((header.HashBlock == getHeadersPayload.HashStop) || (headers.Headers.Count == 2000))
```
mikedennis: What is 2000? Consider using a constant or a comment to clarify what exactly that number represents.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -154,195 +155,260 @@ protected override void DetachCore()
...
+                        doTrySync = true;
+                        break;
+                    }
+
+                    // Now we know the previous block header and thus we can connect the new header.
+                }
+
+                tip = new ChainedBlock(header, header.GetHash(peer.Network.NetworkOptions), prev);
+                bool validated = this.Chain.GetBlock(tip.HashBlock) != null || tip.Validate(peer.Network);
+                validated &= !this.chainState.IsMarkedInvalid(tip.HashBlock);
```
mikedennis: Not a big fan of &= for boolean operations. It looked weird to me so I googled and it does seem valid (it's in C# case example) so I think it's ok but I personally would prefer putting that condition into a variable and adding an && at the end of the main condition.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -154,195 +155,260 @@ protected override void DetachCore()
...
+            }
+
+            if (pendingTipBefore != this.pendingTip)
+                this.logger.LogTrace("Pending tip changed to '{0}'.", this.pendingTip);
 
-                        ChainedBlock chainedPendingTip = this.pendingTip == null ? null : this.Chain.GetBlock(this.pendingTip.HashBlock);
-                        if (chainedPendingTip != null)
+            if ((this.pendingTip != null) && (this.pendingTip.ChainWork > this.Chain.Tip.ChainWork))
+            {
+                // Long reorganization protection on POS networks.
```
mikedennis: Do you want a trace log here? I guess the only condition you don't have a trace on is if maxReorgLength == 0, is that a condition that might be useful to have in the trace log?

Aprogiena:
@mikedennis tx for review, but lots of time you are commenting on old code. this PR is not aiming at fixing this whole class, it is only aiming to make one step towards better code.

Aprogiena:
they are good comments, but when a PR makes some moving of the code without changing it, it is probably better only to comment on what is new, not what is old and basically untouched by the PR, so i'll respond to some of those with "old code"

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: old code

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: old code

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: old code

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
Aprogiena: probably not really, either reorg protection is enabled (stratis) or disabled (bitcoin) and that's about it

but still, old code :D

Aprogiena:
OK, I think everything you've mentioned is related to old code and not related to this PR

PR #992, opened 2017-12-27
Added AsyncManualResetEvent

noescape00:
AsyncManualResetEvent is an async version of ManualResetEventSlim

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Stratis.Bitcoin.Utilities;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Utilities
+{
+    public class AsyncManualResetEventTest
+    {
+        [Fact]
+        public async void AsyncManualResetEvent_WaitAsync()
+        {
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
+
+            Task task = Task.Run(async delegate
+            {
+                await Task.Delay(500);
+                manualResetEvent.Set();
+            });
+
+            await manualResetEvent.WaitAsync();
+        }
+
+        [Fact]
+        public void AsyncManualResetEvent_CanReset()
+        {
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(true);
+
+            Assert.True(manualResetEvent.IsSet);
+
+            manualResetEvent.Reset();
+
+            Assert.False(manualResetEvent.IsSet);
+        }
+
+        [Fact]
+        public async Task AsyncManualResetEvent_IsNotCompletedAsync()
+        {
+            var mre = new AsyncManualResetEvent();
+
+            var task = mre.WaitAsync();
+
+            await Task.Delay(200);
+
+            Assert.False(task.IsCompleted);
+        }
+
+        [Fact]
+        public void AsyncManualResetEvent_AfterSet_IsCompleted()
+        {
+            var mre = new AsyncManualResetEvent();
+
+            mre.Set();
+            var task = mre.WaitAsync();
+
+            Assert.True(task.IsCompleted);
+        }
+
+        [Fact]
+        public void AsyncManualResetEvent_Set_IsCompleted()
+        {
+            var mre = new AsyncManualResetEvent(true);
+
+            var task = mre.WaitAsync();
+
+            Assert.True(task.IsCompleted);
+        }
+
+        [Fact]
+        public void AsyncManualResetEvent_MultipleWaitAsync_Set_IsCompleted()
+        {
+            var mre = new AsyncManualResetEvent(true);
+
+            var task1 = mre.WaitAsync();
+            var task2 = mre.WaitAsync();
+
+            Assert.True(task1.IsCompleted);
+            Assert.True(task2.IsCompleted);
+        }
+
+        [Fact]
+        public async Task AsyncManualResetEvent_AfterReset_IsNotCompletedAsync()
+        {
+            var mre = new AsyncManualResetEvent();
+
+            mre.Set();
+            mre.Reset();
+            var task = mre.WaitAsync();
+
+            await Task.Delay(200);
+
+            Assert.False(task.IsCompleted);
+        }
+    }
+}
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -172,6 +172,8 @@ namespace Stratis.Bitcoin.Tests.Utilities
             // Check that the list is built correctly.
             for (int i = 0; i < resultList.Count; i++)
                 Assert.Equal(i, resultList[i]);
+
+            Assert.Equal(250, resultList.Count);
         }
 
         private async Task AsyncManualResetEvent_RingTriggeringAsync_WorkerAsync(int id, List<AsyncManualResetEvent> events, List<int> resultList, CancellationTokenSource shutdown)
@@ -185,7 +187,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
                 // Prepare next item to the list, but wait a little bit before adding it.
                 int next = resultList.Last() + 1;
-                await Task.Delay(id + 1);
+                await Task.Delay(this.random.Next(10) + 1);
                 resultList.Add(next);
 
                 if (next == 250)
@@ -196,6 +198,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
                 // Select the next worker to go. This can select ourselves again.
                 int nextId = this.random.Next(events.Count);
+                Assert.False(events[nextId].IsSet);
                 events[nextId].Set();
             }
         }
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -10,24 +11,44 @@ namespace Stratis.Bitcoin.Tests.Utilities
 {
     public class AsyncManualResetEventTest
     {
+        /// <summary>
+        /// Tests that:
+        /// <list type="bullet">
+        /// <item><see cref="AsyncManualResetEvent"/> does not crush.</item>
+        /// <item>WaitAsync eventually completes.</item>
+        /// <item>WaitAsync completes when the event is set.</item>
+        /// </list>
+        /// </summary>
         [Fact]
         public async void AsyncManualResetEvent_WaitAsync()
         {
             AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
 
-            Task task = Task.Run(async delegate
+            Stopwatch stopwatch = Stopwatch.StartNew();
+
+            Task task = Task.Run(async () =>
             {
                 await Task.Delay(500);
                 manualResetEvent.Set();
             });
 
-            await manualResetEvent.WaitAsync();
+            Task mreAwaitingTask = manualResetEvent.WaitAsync();
+            await mreAwaitingTask;
+
+            stopwatch.Stop();
+
+            Assert.True(stopwatch.ElapsedMilliseconds >= 500 && stopwatch.ElapsedMilliseconds < 600);
+            Assert.True(mreAwaitingTask.Status == TaskStatus.RanToCompletion);
         }
 
         [Fact]
-        public void AsyncManualResetEvent_CanReset()
+        public void AsyncManualResetEvent_CanSetAndReset()
         {
-            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(true);
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
+
+            Assert.False(manualResetEvent.IsSet);
+
+            manualResetEvent.Set();
 
             Assert.True(manualResetEvent.IsSet);
 
@@ -93,6 +114,32 @@ namespace Stratis.Bitcoin.Tests.Utilities
             await Task.Delay(200);
 
             Assert.False(task.IsCompleted);
+        }
+
+        [Fact]
+        public async void AsyncManualResetEvent_CanBeCancelled()
+        {
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
+
+            CancellationTokenSource tokenSource = new CancellationTokenSource();
+
+            Task task = Task.Run(async () =>
+            {
+                await Task.Delay(500);
+                tokenSource.Cancel();
+            });
+
+            Task mreAwaitingTask = manualResetEvent.WaitAsync(tokenSource.Token);
+
+            try
+            {
+                await mreAwaitingTask;
+            }
+            catch (TaskCanceledException)
+            {
+            }
+
+            Assert.True(mreAwaitingTask.Status == TaskStatus.Canceled);
         }
     }
 }
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -144,12 +144,12 @@ namespace Stratis.Bitcoin.Tests.Utilities
             var cts = new CancellationTokenSource();
 
             var events = new List<AsyncManualResetEvent>();
-            for (int i = 0; i < tasksCount; ++i)
+            for (int i = 0; i < tasksCount; i++)
                 events.Add(new AsyncManualResetEvent(false));
 
             var tasks = new List<Task>();
 
-            for (int i = 0; i < tasksCount; ++i)
+            for (int i = 0; i < tasksCount; i++)
             {
                 AsyncManualResetEvent partnerEvent = i != (tasksCount - 1) ? events[i+1] : events[0];
                 tasks.Add(CountAsync(events[i], partnerEvent, cts));
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -12,8 +12,8 @@ namespace Stratis.Bitcoin.Tests.Utilities
 {
     public class AsyncManualResetEventTest
     {
-        /// <summary>Used in <see cref="AsyncManualResetEvent_RingTriggeringAsync"/>.</summary>
-        private int counter;
+        /// <summary>Source of randomness.</summary>
+        private Random random = new Random();
 
         [Fact]
         public async void AsyncManualResetEvent_WaitAsync()
@@ -139,7 +139,6 @@ namespace Stratis.Bitcoin.Tests.Utilities
         [Fact]
         public async void AsyncManualResetEvent_RingTriggeringAsync()
         {
-            this.counter = 0;
             int tasksCount = 10;
             var cts = new CancellationTokenSource();
 
@@ -149,10 +148,10 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             var tasks = new List<Task>();
 
+            List<int> resultList = new List<int>() { 0 };
             for (int i = 0; i < tasksCount; i++)
             {
-                AsyncManualResetEvent partnerEvent = i != (tasksCount - 1) ? events[i+1] : events[0];
-                tasks.Add(CountAsync(events[i], partnerEvent, cts));
+                tasks.Add(AsyncManualResetEvent_RingTriggeringAsync_WorkerAsync(i, events, resultList, cts));
             }
 
             // Trigger one event.
@@ -168,22 +167,30 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             cts.Dispose();
 
-            Assert.Equal(1000, this.counter);
+            for (int i = 0; i < resultList.Count; i++)
+                Assert.Equal(i, resultList[i]);
         }
 
-        private async Task CountAsync(AsyncManualResetEvent self, AsyncManualResetEvent partnerEvent, CancellationTokenSource shutdown)
+        private async Task AsyncManualResetEvent_RingTriggeringAsync_WorkerAsync(int id, List<AsyncManualResetEvent> events, List<int> resultList, CancellationTokenSource shutdown)
         {
+            AsyncManualResetEvent selfEvent = events[id];
             while (!shutdown.IsCancellationRequested)
             {
-                await self.WaitAsync(shutdown.Token);
+                await selfEvent.WaitAsync(shutdown.Token);
 
-                this.counter++;
+                int next = resultList.Last() + 1;
 
-                if (this.counter >= 1000)
+                await Task.Delay(id + 1);
+
+                resultList.Add(next);
+
+                if (next == 250)
                     shutdown.Cancel();
 
-                self.Reset();
-                partnerEvent.Set();
+                selfEvent.Reset();
+
+                int nextId = this.random.Next(events.Count);
+                events[nextId].Set();
             }
         }
     }
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -11,14 +11,6 @@ namespace Stratis.Bitcoin.Tests.Utilities
 {
     public class AsyncManualResetEventTest
     {
-        /// <summary>
-        /// Tests that:
-        /// <list type="bullet">
-        /// <item><see cref="AsyncManualResetEvent"/> does not crush.</item>
-        /// <item>WaitAsync eventually completes.</item>
-        /// <item>WaitAsync completes when the event is set.</item>
-        /// </list>
-        /// </summary>
         [Fact]
         public async void AsyncManualResetEvent_WaitAsync()
         {
@@ -37,7 +29,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             stopwatch.Stop();
 
-            Assert.True(stopwatch.ElapsedMilliseconds >= 500 && stopwatch.ElapsedMilliseconds < 600);
+            Assert.True(stopwatch.ElapsedMilliseconds >= 500);
             Assert.True(mreAwaitingTask.Status == TaskStatus.RanToCompletion);
         }
 
@@ -62,7 +54,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
         {
             var mre = new AsyncManualResetEvent();
 
-            var task = mre.WaitAsync();
+            Task task = mre.WaitAsync();
 
             await Task.Delay(200);
 
@@ -75,7 +67,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
             var mre = new AsyncManualResetEvent();
 
             mre.Set();
-            var task = mre.WaitAsync();
+            Task task = mre.WaitAsync();
 
             Assert.True(task.IsCompleted);
         }
@@ -85,7 +77,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
         {
             var mre = new AsyncManualResetEvent(true);
 
-            var task = mre.WaitAsync();
+            Task task = mre.WaitAsync();
 
             Assert.True(task.IsCompleted);
         }
@@ -109,7 +101,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             mre.Set();
             mre.Reset();
-            var task = mre.WaitAsync();
+            Task task = mre.WaitAsync();
 
             await Task.Delay(200);
 
@@ -117,17 +109,10 @@ namespace Stratis.Bitcoin.Tests.Utilities
         }
 
         [Fact]
-        public async void AsyncManualResetEvent_CanBeCancelled()
+        public async void AsyncManualResetEvent_CanBeCancelledAsync()
         {
-            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
-
-            CancellationTokenSource tokenSource = new CancellationTokenSource();
-
-            Task task = Task.Run(async () =>
-            {
-                await Task.Delay(500);
-                tokenSource.Cancel();
-            });
+            var manualResetEvent = new AsyncManualResetEvent(false);
+            var tokenSource = new CancellationTokenSource(500);
 
             Task mreAwaitingTask = manualResetEvent.WaitAsync(tokenSource.Token);
 
@@ -140,6 +125,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
             }
 
             Assert.True(mreAwaitingTask.Status == TaskStatus.Canceled);
+            tokenSource.Dispose();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,269 @@
...
+        private static TaskCompletionSource<TResult> CreateAsyncTaskSource<TResult>()
+        {
+            return new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        private static void WaitAndUnwrapException(Task task)
```
Aprogiena: we don't need this method, nor its other version below

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,269 @@
...
+        /// <param name="cancellationToken">The cancellation token that cancels the wait.</param>
+        private static Task WaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            if (task == null)
+                throw new ArgumentNullException(nameof(task));
+
+            if (!cancellationToken.CanBeCanceled)
+                return task;
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
```
Aprogiena: after each return there should be empty space everywhere if the code continues - i.e. there is no `}` following

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        private static void WaitAndUnwrapException(Task task)
+        {
+            if (task == null)
+                throw new ArgumentNullException(nameof(task));
```
Aprogiena: also after throw make a new empty line everywhere

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            Task waitTask = this.WaitAsync();
+            if (waitTask.IsCompleted)
+                return waitTask;
+
+            return WaitAsync(waitTask, cancellationToken);
+        }
+
+        /// <summary>
```
Aprogiena: i think we don't need this method,
 if we want to wait synchrounously, we will do what we do elsewhere - .GetAwaiter().GetResult()

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+        public void Wait()
+        {
+            WaitAndUnwrapException(this.WaitAsync());
+        }
+
+        /// <summary>
+        /// Synchronously waits for this event to be set. This method may block the calling thread.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public void Wait(CancellationToken cancellationToken)
```
Aprogiena: also not needed imho

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+        private static TaskCompletionSource<TResult> CreateAsyncTaskSource<TResult>()
+        {
+            return new TaskCompletionSource<TResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        private static void WaitAndUnwrapException(Task task)
```
Aprogiena: since i think we don't need sync Wait() methods we can delete this method

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+            task.GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Waits for the task to complete, unwrapping any exceptions.
+        /// </summary>
+        /// <param name="task">The task. May not be <c>null</c>.</param>
+        /// <param name="cancellationToken">A cancellation token to observe while waiting for the task to complete.</param>
+        /// <exception cref="OperationCanceledException">The <paramref name="cancellationToken"/> was cancelled before the <paramref name="task"/> completed, or the <paramref name="task"/> raised an <see cref="OperationCanceledException"/>.</exception>
+        private static void WaitAndUnwrapException(Task task, CancellationToken cancellationToken)
```
Aprogiena: synce i think we don't need Wait() method, we can delete this one

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+            using (var cancelTaskSource = new CancellationTokenTaskSource<object>(cancellationToken))
+                await await Task.WhenAny(task, cancelTaskSource.Task).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Attempts to prepare the exception for re-throwing by preserving the stack trace. The returned exception should be immediately thrown.
+        /// </summary>
+        /// <param name="exception">The exception. May not be <c>null</c>.</param>
+        /// <returns>The <see cref="Exception"/> that was passed into this method.</returns>
+        private static Exception PrepareForRethrow(Exception exception)
```
Aprogiena: and we can delete this one if we delete that one

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+            if (!cancellationToken.CanBeCanceled)
+                return task;
+
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
+
+            return DoWaitAsync(task, cancellationToken);
+        }
+
+        private static async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
```
Aprogiena: we can get rid of this method if we just inline its code into the above WaitAsync method

if not, we should document it

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously waits for the task to complete, or for the cancellation token to be canceled.
+        /// </summary>
+        /// <param name="task">The task to wait for. May not be <c>null</c>.</param>
+        /// <param name="cancellationToken">The cancellation token that cancels the wait.</param>
+        private static Task WaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            if (task == null)
```
Aprogiena: we will use Guards in our codebase for this, pls check whole code

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
+
+            return DoWaitAsync(task, cancellationToken);
+        }
+
+        private static async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            using (var cancelTaskSource = new CancellationTokenTaskSource<object>(cancellationToken))
+                await await Task.WhenAny(task, cancelTaskSource.Task).ConfigureAwait(false);
```
Aprogiena: the double await construct probably deserves to be commented,
it is explained here http://www.tomdupont.net/2014/09/await-await-taskwhenany.html

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
+
+            return DoWaitAsync(task, cancellationToken);
+        }
+
+        private static async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            using (var cancelTaskSource = new CancellationTokenTaskSource<object>(cancellationToken))
```
Aprogiena: the whole CancellationTokenTaskSource class seems unnecessary to me, it has one simply use that can be inlined imho

so basically we can do using over the registration instead of using over this

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,273 @@
...
+                if (this.tcs.Task.IsCompleted)
+                    this.tcs = CreateAsyncTaskSource<object>();
+            }
+        }
+
+        /// <summary>
+        /// Creates a new TCS for use with async code, and which forces its continuations to execute asynchronously.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result of the TCS.</typeparam>
+        private static TaskCompletionSource<TResult> CreateAsyncTaskSource<TResult>()
```
Aprogiena: i also think we want to convert all these static methods to nonstatic methods

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Ok. Wanted to keep it so we can use AsyncManualResetEvent as a normal ManualResetEventSlim but you're right, most likely this methods will never be used.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Documented.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed.

noescape00:
>i'd do some more changes, rip off some methods that we don't need
I'm not sure if we can go further. The class is already pretty stripped, even the default even slim's Wait() is deleted in order to keep it shorter.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,177 @@
...
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled(cancellationToken);
+
+            return DoWaitAsync(task, cancellationToken);
+        }
+
+        /// <summary>
+        /// Waits for the task to be executed or canceled.
+        /// </summary>
+        /// <remarks>
```
Aprogiena: remarks should be under param section

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// Double await construct is used here in order to catch exceptions in child tasks.
+        /// It is similar to await + unwrap but uses less allocations. <see cref="https://stackoverflow.com/questions/34816628/await-await-vs-unwrap"/>
```
Aprogiena: if you use `<see>` you need to put it into a sentence, or use `seealso` section

so for example

`For more information about double await see <see ...>.`

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,177 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+https://github.com/StephenCleary/AsyncEx
```
Aprogiena: here I would change this line to:

```
Based on the original implementation of AsyncManualResetEventfrom Stephen Cleary's AsyncEx library - https://github.com/StephenCleary/AsyncEx
```

noescape00: Fixed.

noescape00:
>however, as we changed the original implementation quite a lot, i think we absolutely need tests that will test all functionality of this object

Added unit tests.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: Fixed

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Threading.Tasks;
+using Stratis.Bitcoin.Utilities;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Utilities
+{
+    public class AsyncManualResetEventTest
+    {
+        [Fact]
+        public async void AsyncManualResetEvent_WaitAsync()
```
Aprogiena: what is the purpose of this test? (there is no comment ...)
it seems to test that
- using AMRE does not crash
- WaitAsync eventually completes

however, it does not check
- whether WaitAsync completes when the event is set

also please use lambdas over delegates to be consistent with the rest of the codebase

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                await Task.Delay(500);
+                manualResetEvent.Set();
+            });
+
+            await manualResetEvent.WaitAsync();
+        }
+
+        [Fact]
+        public void AsyncManualResetEvent_CanReset()
```
Aprogiena: we could have CanSetAndReset and test setting the event again and checking ...

noescape00: Fixed.

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
noescape00: Fixed.

noescape00:
>i'm also missing tests with cancellation tokens - so please add those

>and one more thing - once we removed static from many methods, we need to add this. when we call them

Fixed.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Detailed comparison can be found here: <see cref="https://stackoverflow.com/questions/34816628/await-await-vs-unwrap"/>.
+        /// </remarks>
+        private async Task DoWaitAsync(Task task, CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+                return;
+
+            var tcs = new TaskCompletionSource<object>();
+            using (cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken), useSynchronizationContext: false))
+                await await Task.WhenAny(task, tcs.Task).ConfigureAwait(false);
```
Aprogiena: i believe we do want `{}` around `lock` and `using` statements

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// An async-compatible manual-reset event.
+    /// </summary>
+    public sealed class AsyncManualResetEvent
+    {
+        /// <summary>
+        /// The object used for synchronization.
```
Aprogiena: this is not good enough comment (together with the comment of tcs), see e.g. block puller locks

what you want here is to describe the contract fully for both the lock and the protected object
so here 
`Lock to protect access to <see cref="tcs"/>.`

and in tcs, you should add
```
<remarks>All access to this object has to be protected by <see cref="mutex"/>.</remarks>
```

only then the reader could easily understand that the mutex is there to protect tcs and then when tcs is touched it should be protected.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Asynchronously waits for this event to be set or for the wait to be canceled.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token used to cancel the wait. If this token is already canceled, this method will first check whether the event is set.</param>
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            Task waitTask = this.WaitAsync();
+            if (waitTask.IsCompleted)
+                return waitTask;
+
+            return this.WaitAsync(waitTask, cancellationToken);
```
Aprogiena: this call does not make much sense
the body of that method should be copied here and that method should be removed
otherwise you would have that method inside this class that does not work with this class instance at all (remember it was static before -> so that's why it should be brought here)

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        /// <summary>
+        /// Sets the event, atomically completing every task returned by <see cref="O:Nito.AsyncEx.AsyncManualResetEvent.WaitAsync"/>. If the event is already set, this method does nothing.
```
Aprogiena: certainly not O:Nito.AsyncEx ....

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
Aprogiena: and then you will find you don't need DoWaitAsync either

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+            mre.Reset();
+            var task = mre.WaitAsync();
+
+            await Task.Delay(200);
+
+            Assert.False(task.IsCompleted);
+        }
+
+        [Fact]
+        public async void AsyncManualResetEvent_CanBeCancelled()
```
Aprogiena: Async suffix

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        [Fact]
+        public async Task AsyncManualResetEvent_AfterReset_IsNotCompletedAsync()
+        {
+            var mre = new AsyncManualResetEvent();
+
+            mre.Set();
+            mre.Reset();
+            var task = mre.WaitAsync();
```
Aprogiena: unvar these everywhere

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Utilities
+{
+    public class AsyncManualResetEventTest
+    {
+        /// <summary>
+        /// Tests that:
+        /// <list type="bullet">
+        /// <item><see cref="AsyncManualResetEvent"/> does not crush.</item>
```
Aprogiena: well :-) that was kind of a joke to put in the contrast what it really tests and what it does not test
to point out that maybe it was intentional to actually test what it did not test ...

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <item>WaitAsync eventually completes.</item>
```
Aprogiena: always `<see>` when you refer to a method or class or whatever

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+                await Task.Delay(500);
+                manualResetEvent.Set();
+            });
+
+            Task mreAwaitingTask = manualResetEvent.WaitAsync();
+            await mreAwaitingTask;
+
+            stopwatch.Stop();
+
+            Assert.True(stopwatch.ElapsedMilliseconds >= 500 && stopwatch.ElapsedMilliseconds < 600);
```
Aprogiena: this could blow as we've seen in my stopwatch tests
although you did allow 100 ms epsilon, so most of the time it should be ok, but better to avoid this
also use () or make two separate Asserts

Aprogiena: now i remembered that the stopwatch test is actually disabled on mac because it did not work even with higher epsilon

in general i'd prefer a test which instead of measuring time proves correctness by correct ordering, that would avoid problems with time measurement in test environment

so you can definitely replace this test with such a test of good ordering
so imagine two or more tasks (let's do two for simplicity, i'd do three for fun, but simplicity is better than fun) - A, B
and these guys want to cooperate to create a list of 100 integers from 1 to 100.

they will sync with this async event

so you create two events EA and EB, EA is set and EB is reset
you create two tasks over the same body and pass both events as the arguments, but for A, you pass events in order "EA, EB" and for B you pass events in order "EB, EA"

and you create shutdownEvent that is set to false

```
body(myEvent, theirEvent, shutdownEvent)
{
   loop until we are done
   {
       wait for myEvent + shutdownEvent
       reset myEvent
       get last number from the list
       add one to that number
       add the new number to the list
       if the newly added number is the limit to reach signal shutdownEvent
       set theirEvent
   }
}
```

the main thread should wait only for shutdownEvent and then asserts that the list is formed well

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        [Fact]
+        public async void AsyncManualResetEvent_CanBeCancelled()
+        {
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
+
+            CancellationTokenSource tokenSource = new CancellationTokenSource();
+
+            Task task = Task.Run(async () =>
```
Aprogiena: this is meeeh
use `            CancellationTokenSource tokenSource = new CancellationTokenSource(500);` instead of this charade
and for the sake of all kittens, dispose your cancellation token sources!!!

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            await Task.Delay(200);
+
+            Assert.False(task.IsCompleted);
+        }
+
+        [Fact]
+        public async void AsyncManualResetEvent_CanBeCancelled()
+        {
+            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
```
Aprogiena: here and below for CTS we use `var`

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
noescape00: I'd actually keep the DoWaitAsync and not merge it with WaitAsync because if we merge it we should make the code that is now in DoWaitAsync to be synchronous or change the signature of WaitAsync which I'd avoid.

noescape00: >the body of that method should be copied here and that method should be removed

Ok.

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
noescape00: Kk, I'll remove the comment.

src/Stratis.Bitcoin/Utilities/AsyncManualResetEvent.cs
Aprogiena: but is it not worth it to remove DWA?

Aprogiena: how about something like this, to me it looks simple enough for it to be 1 method, not 3
it remains to check if i did not do any mistakes with my quick hack coding 
```
        public async Task WaitAsync(CancellationToken cancellationToken)
        {
            Task waitTask = this.WaitAsync();
            if (waitTask.IsCompleted)
                return;

            if (!cancellationToken.CanBeCanceled)
            {
                await waitTask.ConfigureAwait(false);
                return;
            }

            var tcs = new TaskCompletionSource<object>();
            using (cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken), useSynchronizationContext: false))
            {
                await await Task.WhenAny(waitTask, tcs.Task).ConfigureAwait(false);
            }
        }
```

so if you could check this is really equivalent of the original ... then let me know if you don't think this is better

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
Aprogiena: and we could also assert that when A wakes up then EB is reset and vice versa

noescape00: I'm really not sure if we should overcomplicate this. 
Ok, I agree that we should the <600 condition just in case. But the ordering test is overkill. 

>proves correctness by correct ordering

This wasn't the intention of this test. This test tests if the event was triggered after the 500ms delay and not before, that's all.

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
noescape00: Ah, yeah. 
That was silly.

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
Aprogiena: I know, but a little more complicated test would show that the async event is reliable. Maybe it is overkill and we can think of a much simpler test (although it is really 15 minutes of coding). 

The problem with this test is that it will possibly fail in some environments. The stopwatch tests that failed on Mac was more complex, so maybe this one would work, but conceptually, such measurement is not good and can fail (imagine your machine is under heavy stress, executing 200 threads and the thread that should jump of your task.delay won't get scheduled on time, the test will fail and it should not ...). So if you feel you can think out a simpler test, let's go for a simpler, but I'd like to avoid using stopwatch.

noescape00: Fixed. 

>so imagine two or more tasks (let's do two for simplicity, i'd do three for fun, but simplicity is better than fun)
Kept it simple. Had some fun. 
Now we have 10 tasks.

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -127,5 +131,60 @@ public void AsyncManualResetEvent_MultipleWaitAsync_Set_IsCompleted()
...
+        /// </summary>
+        [Fact]
+        public async void AsyncManualResetEvent_RingTriggeringAsync()
+        {
+            this.counter = 0;
+            int tasksCount = 10;
+            var cts = new CancellationTokenSource();
+
+            var events = new List<AsyncManualResetEvent>();
+            for (int i = 0; i < tasksCount; ++i)
```
Aprogiena: no `++i` in c#

src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -127,5 +131,60 @@ public void AsyncManualResetEvent_MultipleWaitAsync_Set_IsCompleted()
...
+                events.Add(new AsyncManualResetEvent(false));
+
+            var tasks = new List<Task>();
+
+            for (int i = 0; i < tasksCount; ++i)
```
Aprogiena: no ++i

Aprogiena:
I'm happy with this now, tests seem to really test the class. So @noescape00 just pls check my last modifications if you are happy with them.

noescape00:
@Aprogiena yes, looks good to me. 

>for (int i = 0; i < resultList.Count; i++)
>                Assert.Equal(i, resultList[i]);

This assertion loop looks unnecessary to me but I'm fine with it.

PR #998, opened 2017-12-28
Bugfix/stratisx sync issue 986

### LINKED ISSUE #986: [StratisX] Can't sync from scratch from C# node
reports orphans from block 1

Aprogiena:
Somewhat controversial PR, but probably needed (the controversion here is only about whether we should support outdated mechanisms or not).

The problem is that StratisX is unable to sync from C# node. The reason for that is C# node does not implement `getblocks` network message, which is the mechanism that is no longer used by Bitcoin Core, but it is still used by StratisX. Without it, StratisX can only sync from StratisX which implements that.

So this reimplements (from the old source code) support for `getblocks`. The protocol says that in response to `getblocks`, the node should reply with up to 500 items of inventory, but if just that is done, the asking StratisX node would not continue to sync. So the C# node has to send a continuation inventory message with its tip (we are using block store tip for this). And it is important to send it just after StratisX node asks for the last block data (`getdata` message) from the last batch of those 500 items. 

Testing is still heavily in progress. 

# TESTING
We have the following testing scenarios that we need to test
**Note that for testing below, you can't use just any random C# nodes on the network, only C# nodes running this PR code counts. Contant me on slack for IP address if you don't have such a node.**

**General**
```
[x] Unit tests (apro | AppVeyor)
[x] Integration tests (apro)
```

**Stratis Testnet network**
```
[x] Try to sync StratisX from a single C# node from scratch. (apro-qt-both staking | demon-daemon)
[x] Try to sync StratisX from multiple C# nodes from scratch. (apro-qt-both staking | demon-daemon)
[x] Try to keep synced StratisX node in sync with network when it is connected only to a single C# node. (apro, qt, both staking)
[x] Try to keep synced StratisX node in sync with network when it is connected to multiple C# nodes. (apro-qt-both staking | demon-daemon)
[x] Try to sync C# node from a single StratisX node from scratch. (apro-staking)
[x] Try to sync C# node from a single C# node from scratch. (apro-both staking | noescape)
```

**Stratis Mainnet network**
```
[x] Try to sync StratisX from a single C# node from scratch. (demon-daemon)
[x] Try to sync StratisX from multiple C# nodes from scratch. (demon-daemon)
[x] Try to keep synced StratisX node in sync with network when it is connected only to a single C# node. (demon-daemon)
[x] Try to keep synced StratisX node in sync with network when it is connected to multiple C# nodes. (demon-daemon)
```

**Bitcoin Testnet or Mainnet**
```
[x] Try to sync C# node on Bitcoin mainnet or testnet. (apro - all good except for #999 but that seems completely unrelated)
```

Fix #986

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -213,8 +213,16 @@ namespace Stratis.Bitcoin.Features.BlockStore
             int count = inv.Inventory.Count;
             if (count > 0)
             {
-                this.logger.LogTrace("Setting peer's pending tip to '{0}'.", lastAddedChainedBlock);
                 ChainHeadersBehavior chainBehavior = peer.Behavior<ChainHeadersBehavior>();
+
+                // New nodes do not use "getblocks" messages and their syncing is based on "getheaders"
+                // messages. StratisX nodes do need "getblocks", but then our node tries to sync 
+                // with the modern method. So we need to disable it to prevent collisions of those two methods.
+                // However, this should be only done if the peer is syncing from us and not vice versa.
+                // And when the sync is done, we should enable it again.
+                chainBehavior.CanSync = (chainBehavior.PendingTip != null) && (this.chain.Tip.ChainWork > chainBehavior.PendingTip.ChainWork);
+
+                this.logger.LogTrace("Setting peer's pending tip to '{0}'.", lastAddedChainedBlock);
                 chainBehavior.SetPendingTip(lastAddedChainedBlock);
 
                 this.logger.LogTrace("Sending inventory with {0} block hashes.", count);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -314,7 +314,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         private async Task ProcessGetDataAsync(NetworkPeer peer, GetDataPayload getDataPayload)
         {
-            Guard.Assert(peer != null);
             this.logger.LogTrace("({0}:'{1}',{2}.{3}.{4}:{5})", nameof(peer), peer.RemoteSocketEndpoint, nameof(getDataPayload), nameof(getDataPayload.Inventory), nameof(getDataPayload.Inventory.Count), getDataPayload.Inventory.Count);
 
             // TODO: bring logic from core
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -214,16 +214,18 @@ namespace Stratis.Bitcoin.Features.BlockStore
             if (count > 0)
             {
                 ChainHeadersBehavior chainBehavior = peer.Behavior<ChainHeadersBehavior>();
+                ChainedBlock peerTip = chainBehavior.PendingTip;
 
                 // New nodes do not use "getblocks" messages and their syncing is based on "getheaders"
                 // messages. StratisX nodes do need "getblocks", but then our node tries to sync 
                 // with the modern method. So we need to disable it to prevent collisions of those two methods.
                 // However, this should be only done if the peer is syncing from us and not vice versa.
                 // And when the sync is done, we should enable it again.
-                chainBehavior.CanSync = (chainBehavior.PendingTip != null) && (this.chain.Tip.ChainWork > chainBehavior.PendingTip.ChainWork);
+                chainBehavior.CanSync = (peerTip != null) && (this.chain.Tip.ChainWork > peerTip.ChainWork);
 
                 this.logger.LogTrace("Setting peer's pending tip to '{0}'.", lastAddedChainedBlock);
-                chainBehavior.SetPendingTip(lastAddedChainedBlock);
+                int peersHeight = peerTip != null ? peerTip.Height : 0;
+                if (peerTip.Height < lastAddedChainedBlock.Height) chainBehavior.SetPendingTip(lastAddedChainedBlock);
 
                 this.logger.LogTrace("Sending inventory with {0} block hashes.", count);
                 await peer.SendMessageAsync(inv).ConfigureAwait(false);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -78,7 +78,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.loggerFactory = loggerFactory;
 
-            this.CanRespondToGetBlocksPayload = false;
+            this.CanRespondToGetBlocksPayload = true;
             this.CanRespondToGetDataPayload = true;
 
             this.PreferHeaders = false;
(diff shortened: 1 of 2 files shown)


Aprogiena:
Please let me know if you were able to complete one of the required tests, I'll mark it as finished.

Aprogiena:
also note that it is very helpful if you just repeat test that someone else have done already, often bugs are not visible with just one run

D3m0nKingx:
```[ ] Try to sync C# node from a single StratisX node from scratch.```
This test fails for me if sX is in IBD when C# node is trying to sync from it.

**Comment by Aprogiena** [comment]
That's fine, so we are almost ready.

File: src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs [hunk 1]
```diff
@@ -131,49 +141,180 @@ protected override void DetachCore()
...
+                headersToAnnounce[i] = lastBlock;
+                lastBlock = lastBlock.Previous;
+            }
+
+            // Now we compile inventory payload and we also consider hash stop given by the peer
+            // and we also check if the blocks to be announced are really present in the store.
+            bool sendContinuation = true;
+            ChainedBlock lastAddedChainedBlock = null;
+            var inv = new InvPayload();
+            for (int i = 0; i < headersToAnnounce.Length; i++)
```
mikedennis: Looks like you could use a foreach here instead of a for

Aprogiena: I could, but it felt more aesthetic to use for when we have another one just above :-)
Do we want to prefer foreaches or is it OK to leave this one as is?

mikedennis: I'm ok with using a for loop (I came from a C++ background 😁) but I think generally C#ers prefer to use foreach when iterating collections unless you are using the index for something else. At any rate just a suggestion I don't think this PR should be held up for this minor detail.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -103,18 +113,18 @@ protected override void DetachCore()
...
             try
             {
-                await this.ProcessMessageAsync(node, message).ConfigureAwait(false);
+                await this.ProcessMessageAsync(peer, message).ConfigureAwait(false);
             }
             catch (OperationCanceledException opx)
             {
                 if (!opx.CancellationToken.IsCancellationRequested)
-                    if (this.AttachedPeer?.IsConnected ?? false)
+                    if (peer.IsConnected)
```
noescape00: I think you should return the null check for peer just to be extra safe. 
It will throw if cancellation is requested after that 'IsCancellationRequested' check and there was enough time to detach the peer (detach sets peer to null).

noescape00:
Looks good to me. 
Tested it in different scenarios- it fixes the bug indeed.

Aprogiena:
So during my last test - bitcoin testnet syncing from scratch, all went fine until almost very end - #999 and #1000 appeared.

I strongly believe both are completely unrelated to this PR.

noescape00:
I can confirm: https://github.com/stratisproject/StratisBitcoinFullNode/issues/999 is unrelated to this PR, I was able to reproduce the bug with master's code.

Aprogiena:
Thanks!

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -131,49 +141,180 @@ protected override void DetachCore()
...
+                    }
+
+                    await this.ProcessGetDataAsync(peer, getDataPayload).ConfigureAwait(false);
+                    break;
+
+                case GetBlocksPayload getBlocksPayload:
+                    // TODO: this is not used in core anymore consider deleting it
+                    // However, this is required for StratisX to be able to sync from us.
+
+                    if (!this.CanRespondToGetDataPayload)
```
dangershony: This should be `CanRespondToGetBlocksPayload`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -131,49 +141,180 @@ protected override void DetachCore()
...
+            for (int i = 0; i < headersToAnnounce.Length; i++)
+            {
+                ChainedBlock chainedBlock = headersToAnnounce[i];
+                if (chainedBlock.HashBlock == getBlocksPayload.HashStop)
+                {
+                    this.logger.LogTrace("Hash stop has been reached.");
+                    break;
+                }
+
+                if (!await this.blockRepository.ExistAsync(chainedBlock.HashBlock).ConfigureAwait(false))
```
dangershony: This step is probably not necessary, we could not get here if the block was not in store right?
This line earlier means the store and all its previous blocks are in place.
```
 ChainedBlock blockStoreTip = this.chain.GetBlock(this.blockRepository.BlockHash);
```

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -183,10 +324,30 @@ private Task ProcessSendCmpctPayload(NetworkPeer node, SendCmpctPayload sendCmpc
...
-                    this.logger.LogTrace("Sending block '{0}' to peer '{1}'.", item.Hash, node?.RemoteSocketEndpoint);
+                    this.logger.LogTrace("Sending block '{0}' to peer '{1}'.", item.Hash, peer.RemoteSocketEndpoint);
 
                     //TODO strip block of witness if node does not support
-                    await node.SendMessageAsync(new BlockPayload(block.WithOptions(node.SupportedTransactionOptions))).ConfigureAwait(false);
+                    await peer.SendMessageAsync(new BlockPayload(block.WithOptions(peer.SupportedTransactionOptions))).ConfigureAwait(false);
+                }
+
+                // If the peer is syncing using "getblocks" message we are supposed to send 
+                // an "inv" message with our tip to it once we it asks for all blocks
```
dangershony: remove the `we`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -56,6 +56,16 @@ public class BlockStoreBehavior : NetworkPeerBehavior, IBlockStoreBehavior
...
 
+        /// <summary>Hash of the last block we've sent to the peer in response to "getblocks" message,
+        /// or <c>null</c> if the peer haven't used "getblocks" message or if we sent a tip to it already.</summary>
+        /// <remarks>
+        /// In case the peer is syncing using outdated "getblocks" message, we need to maintain 
+        /// the hash of the last block we sent to it in an inventory batch. Once the peer asks 
+        /// for block data of the block with this hash, we will send a continuation inventory message.
+        /// This will cause the peer to ask for more.
+        /// </remarks>
+        private uint256 getBlocksBatchLastItemHash;
```
dangershony: Shall we add a todo that this needs to go away once we fork away from X?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: of course! thanks

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: yes, tx

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: it does not really hurt to support this thing, but it will become really obsolete then ... that's a long shot issue to have, but let me add such a todo comment just above this line

```
            this.CanRespondToGetBlocksPayload = true;

```

Aprogiena: oh well it was there already:

```
                case GetBlocksPayload getBlocksPayload:
                    // TODO: this is not used in core anymore consider deleting it
                    // However, this is required for StratisX to be able to sync from us.
```

Aprogiena: so no change

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: it seems that you are right, i can't see how a block could NOT YET be in store at this point

it can be out of the store due to reorg, in which case it would make sense to stop the loop, however, that can happen even after that, so it makes no sense to check for that (again)

i'll remove it

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
dangershony: Ah cool ok I missed it.

PR #1003, opened 2017-12-29
Refactor/remove system.web.nbitcoin

Aprogiena:
removing nbitcoin's implementation of HttpUtility and fixing the test that was poorly constructed

i haven't found any example in BIP 21 with actual URI inside URI as this test does, but it seems obvious that it was poorly constructed, doesn't it?


fassadlr:
I'm reviewing this one.

Aprogiena:
See https://github.com/stratisproject/StratisBitcoinFullNode/pull/876#discussion_r155926604 for the context. What you say may be true, but has not been considered. I just wanted to free mike there. If we remove the whole thing later, that's OK as well.

Issue #991, opened 2017-12-27
[EventLoopMessageListener] Crash - cancellation disposed


Aprogiena:
```
Unhandled Exception: System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
   at System.Threading.CancellationTokenSource.get_Token()
   at NBitcoin.Protocol.EventLoopMessageListener`1.<>c__DisplayClass4_0.<<-ctor>b__0>d.MoveNext() in /home/fullnode5/StratisBitcoinFullNode/src/NBitcoin/Protocol/MessageListener.cs:line 62
```

this is subject to refactoring, so it will be changed

PR #1010, opened 2018-01-02
Feature/async queue

Aprogiena:
Async queue is a simple thread-safe queue that asynchronously executes a user-defined callback when a new item is added to the queue.

Implementation of async queue that we want for the node job part 19, which will fix issue #991 by refactoring that area of the code. 

Tests are included.

There are no changes to functionality of any existing components in this PR.

### DIFF
--- a/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// Async queue is a simple thread-safe queue that asynchronously executes 
+    /// a user-defined callback when a new item is added to the queue.
+    /// <para>
+    /// The queue guarantees that the user-defined callback is executed only once at the time. 
+    /// If another item is added to the queue, the callback is called again after the current execution 
+    /// is finished.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">Type of items to be inserted in the queue.</typeparam>
+    public class AsyncQueue<T>: IDisposable
+    {
+        /// <summary>
+        /// Represents a callback method to be executed when a new item is added to the queue.
+        /// </summary>
+        /// <param name="item">Newly added item.</param>
+        /// <param name="cancellationToken">Cancellation token that the callback method should use for its async operations to avoid blocking the queue during shutdown.</param>
+        public delegate Task OnEnqueueAsync(T item, CancellationToken cancellationToken);
+
+        /// <summary>Storage of items in the queue that are waiting to be consumed.</summary>
+        private readonly ConcurrentQueue<T> items;
+
+        /// <summary>Event that is triggered when at least one new item is waiting in the queue.</summary>
+        private readonly AsyncManualResetEvent signal;
+
+        /// <summary>Callback routine to be called when a new item is added to the queue.</summary>
+        private readonly OnEnqueueAsync onEnqueueAsync;
+
+        /// <summary>Consumer of the items in the queue which responsibility is to execute the user defined callback.</summary>
+        private readonly Task consumerTask;
+
+        /// <summary>Cancellation that is triggered when the component is disposed.</summary>
+        private readonly CancellationTokenSource cancellationTokenSource;
+
+        /// <summary>
+        /// Initializes the queue.
+        /// </summary>
+        /// <param name="onEnqueueAsync">Callback routine to be called when a new item is added to the queue.</param>
+        public AsyncQueue(OnEnqueueAsync onEnqueueAsync)
+        {
+            Guard.NotNull(onEnqueueAsync, nameof(onEnqueueAsync));
+
+            this.items = new ConcurrentQueue<T>();
+            this.signal = new AsyncManualResetEvent();
+            this.onEnqueueAsync = onEnqueueAsync;
+            this.cancellationTokenSource = new CancellationTokenSource();
+            this.consumerTask = ConsumerAsync();
+        }
+
+        /// <summary>
+        /// Add a new item to the queue and signal to the consumer task.
+        /// </summary>
+        /// <param name="item">Item to be added to the queue.</param>
+        public void Enqueue(T item)
+        {
+            this.items.Enqueue(item);
+            this.signal.Set();
+        }
+
+        /// <summary>
+        /// Consumer of the newly added items to the queue that waits for the signal 
+        /// and then executes the user-defined callback.
+        /// </summary>
+        public async Task ConsumerAsync()
+        {
+            CancellationToken cancellationToken = this.cancellationTokenSource.Token;
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    // Wait for an item to be enqueued.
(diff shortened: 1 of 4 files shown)


dangershony:
I thought @noescape will work on this.
I like the name but it's not so correct its an AsyncTrigger more then a queue.

Aprogiena:
at the heart of it, it is a queue. it is a queue because the producer enqueues items in specific order and the consumer consumes them in that order.

currently a similar mechanism is used in NetworkPeer code and there it is called EventLoopMessageListener, in which a message producer adds messages to listener queues

so the primary purpose of AsyncQueue is to replace EventLoopMessageListener, which uses one thread per listener

the optional feature that allows you to manually trigger the consumer is there in order to allow us to create a second class that could be built on the top of this one with inheritance, that will trigger the manual signal with a timer. that can be then used for the blockstore thing, which is what @noescape00 was working on (and is still kind of working on)

Aprogiena:
so basically i need this for the node job part 19
and it is possible that it will be a multi-purpose class (i.e. reused in other components as well)

@noescape00 won't be able to deliver a similar class quickly enough for me, so although there is no rush with blockstore refactoring, i would like to advance in the node job much earlier

noescape00:
>@noescape00 won't be able to deliver a similar class quickly enough for me

It's already done and I can wrap up a PR right now if you'd like. =)

noescape00:
What is the purpose if this class? 
Basically it's a wrapper around a queue that has triggers on enqueue and a manual one. 
Why manual trigger is there? It's not connected to a queue in any way, it doesn't even have an access to the queue.
 
I don't think that having in this class an anonymous function that can be manually executed is any better than letting a user do this if he want's too. 
I just don't see a scenario where having this trigger would be actually better than doing it in user code.

noescape00:
I've created a PR with classic implementation of async queue: https://github.com/stratisproject/StratisBitcoinFullNode/pull/1013

noescape00:
@dangershony we've been arguing with @Aprogiena  on the design of this thing for a week so we need a second opinion on this =)

Basically both of the approaches are used to solve same problem- make implementing a scenario where timer and trigger are involved easier. 
A concrete usecase for this is the relay worker: we wanna form batches by waiting for a second or so between announcing the headers BUT if we get a tip header it means we need to send it now because it's critical to propagate a new block right away (especially if we've mined the block). 

My position was to make this easy in the user code even at a cost of a bit more difficult class. In the end I had a class which usage is easy and straight forward: 

```
AsyncConsumerQueue<int> processedQueue = new AsyncConsumerQueue<int>(
    innerQueue =>
    {
        // Process the queue. This is the only place with the access to the 'innerQueue'
    }, item =>
    {
        // Condition that should be met in order to start processing the
        // queue right after an item as added. 
        return item == *;
    }, TimeSpan.FromMilliseconds(500)); // Timer delay. Queue processing will be invoked in 0.5 sec if the condition was not met. 
```

@Aprogiena's idea was to make the queue class as tiny as possible even at a cost of making user code more difficult and forcing user to understand queue implementation before using it.

Aprogiena:
I'm OK to remove the manual triggering from the async queue as it is not needed for me node job refactoring. It is there just as a concept to be used in the future that we can built blockstore thing on. If we don't want to do that, it is fine, I can remove the manual trigger.

Aprogiena:
@noescape00 the purpose is described above - to replace EventLoopMessageListener which is blocking queue consuming a thread.

Aprogiena:
@dangershony so i've removed the controversial manual trigger, so you should not have problem with the name of the class now :-)

@noescape00 i've removed the controversial manual trigger, so you should not have a feeling that this is supposed to be used in blockstore and compete with your implementation

Aprogiena:
For the sake of discussion on extension towards the timer trigger, i've left the original version with manual trigger in https://github.com/Aprogiena/StratisBitcoinFullNode/tree/feature/async-queue-with-manual-trigger if anyone was interested

noescape00:
>the purpose is described above - to replace EventLoopMessageListener which is blocking queue consuming a thread.

Which can also be done with AsyncConsumerQueue in more simple and clean way like this: 

```
AsyncConsumerQueue<int> processedQueue = new AsyncConsumerQueue<int>(
    innerQueue =>
    {
        NetworkPeerServer.ProcessMessageAsync()
    });
```

This will also avoid a thread consumption. So basically there is no need for another queue that can only one use case if we can use a more general class that can support exactly the behavior you require.

noescape00:
>@noescape00 i've removed the controversial manual trigger, so you should not have a feeling that this is supposed to be used in blockstore and compete with your implementation

It still says in the description that it IS supposed to. 

Actually even with this in mind I don't feel like we need this. Why bringing more narrow-usage classes to the codebase when we can have just 1 class that can be well used in many usecases including the one that made you create this PR?

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly Task consumerTask;
+
+        /// <summary>Cancellation that is triggered when the component is disposed.</summary>
+        private readonly CancellationTokenSource cancellationTokenSource;
+
+        /// <summary>
+        /// Initializes the queue.
+        /// </summary>
+        /// <param name="onEnqueueAsync">Callback routine to be called when a new item is added to the queue.</param>
+        public AsyncQueue(OnEnqueueAsync onEnqueueAsync)
```
noescape00: Why not using `Func<T, CancellationToken, Task>` instead of creating a delegate?

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        /// <summary>
+        /// Represents a callback method to be executed when a new item is added to the queue.
+        /// </summary>
+        /// <param name="item">Newly added item.</param>
+        /// <param name="cancellationToken">Cancellation token that the callback method should use for its async operations to avoid blocking the queue during shutdown.</param>
+        public delegate Task OnEnqueueAsync(T item, CancellationToken cancellationToken);
+
+        /// <summary>Storage of items in the queue that are waiting to be consumed.</summary>
+        private readonly ConcurrentQueue<T> items;
```
noescape00: Can you please make it `protected`?

Aprogiena: if needed, the further work can change this

just as there is no need for manual trigger now, it is perfectly fine for this to be private

there is no problem changing anything in the class later and extend it to either that version with manual trigger that i have in separate branch, or to whatever you have in your mind. this being private does not forbid any changes later

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: the ONLY reason for that is to have comments on the arguments of the delegate, 
I originally did have Func there and found out i lack the documentation of those parameters

i checked with MSDN and they do use delegates for the same reason (at least i believe), those delegates are commented separately and that's what i needed here

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
noescape00: Ok, reasonable.

noescape00:
We've came to an agreement that I'll build queue with timer&trigger on top of the AsyncQueue and this one should be pushed because it's required for node refactoring and is time sensitive.

dangershony:
This looks good to me.

PR #1013, opened 2018-01-02
Feature/async queue (classic)

noescape00:
Async consumer queue is a simple thread-safe queue that has a user defined 'process' function that is called when an item that meets certain condition is enqueued OR when a timer runs out. 

We can use this async queue as a base class for implementation of block store queue to resolve issue #930.

Tests are included.

There are no changes to functionality of any existing components in this PR.


noescape00:
Closed before we figure out what to do with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1010 and I improve this one.

PR #1014, opened 2018-01-02
Stratis mining test

zeptin:
There do not seem to be tests that specifically test the creation of coins in the PoW phase on StratisRegTest. As a result there are some apparent bugs or gaps in this functionality, particularly for usage in integration tests.

It is intended that this initial test be fleshed out into multiple tests that validate the functionality of coins mined on StratisRegTest end-to-end (i.e. generated coins can be used for transactions, the resulting transactions pass validation on a secondary node, the test framework's block generation correctly transitions from PoW to PoS at the right block height, etc.)

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
@@ -41,11 +41,15 @@ namespace Stratis.Bitcoin.Features.Miner
 
             base.CreateNewBlock(scriptPubKeyIn, fMineWitnessTx);
 
-            this.coinbase.Outputs[0].ScriptPubKey = new Script();
-            this.coinbase.Outputs[0].Value = Money.Zero;
-
             PosConsensusValidator posValidator = this.consensusLoop.Validator as PosConsensusValidator;
             Guard.NotNull(posValidator, nameof(posValidator));
+
+            // Without this the coinbase transaction has a zero value
+            this.coinbase.Outputs[0].Value = this.fees + posValidator.GetProofOfWorkReward(this.height);
+            this.pblocktemplate.TotalFee = this.fees;
+
+            this.UpdateHeaders();
+            this.TestBlockValidity();
 
             this.logger.LogTrace("(-)");
             return this.pblocktemplate;
--- a/src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
@@ -36,10 +36,11 @@ namespace Stratis.Bitcoin.IntegrationTests
 
                 node1.NotInIBD();
 
-                // Create the originating node's wallet
+                // Create the generating node's wallet
                 var wm1 = node1.FullNode.NodeService<IWalletManager>() as WalletManager;
                 wm1.CreateWallet("Source1", "source");
 
+                // Get the miner secret by obtaining the private key of one of the source node's addresses
                 var wallet1 = wm1.GetWalletByName("source");
                 var account1 = wallet1.GetAccountsByCoinType((CoinType)node1.FullNode.Network.Consensus.CoinType).First();
                 var address1 = account1.GetFirstUnusedReceivingAddress();
@@ -48,9 +49,10 @@ namespace Stratis.Bitcoin.IntegrationTests
                 // We can use SetDummyMinerSecret here because the private key is already in the wallet
                 node1.SetDummyMinerSecret(new BitcoinSecret(secret1.PrivateKey, node1.FullNode.Network));
 
-                // Generate a block so we have some funds to create a transaction with
+                // Generate a block so we receive some coins
                 node1.GenerateStratisWithMiner(1);
 
+                // The generated coins should immediately appear in the wallet with 1 confirmation
                 Assert.True(account1.GetSpendableAmount().ConfirmedAmount > 0);
 
                 node1.Kill();
--- a/src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using NBitcoin;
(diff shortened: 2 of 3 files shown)


Neurosploit:
Great @zeptin. I'm working on unit tests for the miner code at the moment. I'll merge that later.

src/Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
```diff
@@ -41,12 +41,16 @@ public override BlockTemplate CreateNewBlock(Script scriptPubKeyIn, bool fMineWi
...
             base.CreateNewBlock(scriptPubKeyIn, fMineWitnessTx);
 
-            this.coinbase.Outputs[0].ScriptPubKey = new Script();
-            this.coinbase.Outputs[0].Value = Money.Zero;
-
             PosConsensusValidator posValidator = this.consensusLoop.Validator as PosConsensusValidator;
             Guard.NotNull(posValidator, nameof(posValidator));
 
+            // Without this the coinbase transaction has a zero value
+            this.coinbase.Outputs[0].Value = this.fees + posValidator.GetProofOfWorkReward(this.height);
```
Aprogiena: this seems like an incompatible change, haven't tested though, but what makes you think it is a good idea?

zeptin: Considering that the functionality previously did not work at all, incompatibility is a given.

Without those lines, the Stratis POW blocks have a single transaction inside them with 0 value. This is not correct. There should be 4 Stratis coins generated per block in the POW phase.

Aprogiena:
i'll check this one

Aprogiena:
This code breaks consensus, which means the node when running to stake will try to create a block that is not compatible with the network consensus and will recognize that before it starts any real work, so it will just not stake at all. This is broken.

src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
```diff
@@ -0,0 +1,62 @@
...
+                // We can use SetDummyMinerSecret here because the private key is already in the wallet
+                node1.SetDummyMinerSecret(new BitcoinSecret(secret1.PrivateKey, node1.FullNode.Network));
+
+                // Generate a block so we receive some coins
+                node1.GenerateStratisWithMiner(1);
+
+                // The generated coins should immediately appear in the wallet with 1 confirmation
+                Assert.True(account1.GetSpendableAmount().ConfirmedAmount > 0);
+
+                node1.Kill();
```
noescape00: No need to kill it. NodeBuilder.Dispose does that.

src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Stratis.Bitcoin.IntegrationTests
+{
+    public class StratisMinerTests
+    {
+        [Fact]
+        public void CanMineStratisCoins()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create())
+            {
+                CoreNode node1 = builder.CreateStratisPosNode(true, fullNodeBuilder =>
```
noescape00: PoW node should be used. 
Also no need for specifying features, there are helpers for that.

src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
```diff
@@ -0,0 +1,62 @@
...
+                    fullNodeBuilder
+                        .UseStratisConsensus()
+                        .UseBlockStore()
+                        .UseMempool()
+                        .UseWallet()
+                        .AddPowPosMining()
+                        .AddRPC();
+                });
+
+                node1.NotInIBD();
```
noescape00: NotInIBD(); will try to get a full node reference. This will fail at this moment because you need to call ' builder.StartAll();' that will assign it.

noescape00:
Revert all the full node changes and replace the test with a fixed version, it will work:  

```
[Fact]
public void CanMineStratisCoins()
{
    {
        CoreNode node = builder.CreateStratisPowNode();
        builder.StartAll();
        node.NotInIBD();

        // Create the generating node's wallet.
        IWalletManager walletManager = node.FullNode.NodeService<IWalletManager>() as WalletManager;
        walletManager.CreateWallet("Source1", "source");

        // Get the miner secret by obtaining the private key of one of the source node's addresses.
        Wallet wallet = walletManager.GetWalletByName("source");
        HdAccount account = wallet.GetAccountsByCoinType((CoinType)node.FullNode.Network.Consensus.CoinType).First();
        HdAddress address = account.GetFirstUnusedReceivingAddress();
        ISecret secret = wallet.GetExtendedPrivateKeyForAddress("Source1", address);

        // We can use SetDummyMinerSecret here because the private key is already in the wallet.
        node.SetDummyMinerSecret(new BitcoinSecret(secret.PrivateKey, node.FullNode.Network));

        // Generate a block so we receive some coins.
        node.GenerateStratisWithMiner(1);

        TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(node));

        // The generated coins should immediately appear in the wallet with 1 confirmation.
        var spendable = account.GetSpendableAmount();
        Assert.Equal(new Money(50 * Money.COIN), spendable.ConfirmedAmount);
    }
}
```

src/Stratis.Bitcoin.Features.Miner/PosBlockAssembler.cs
noescape00: >There should be 4 Stratis coins generated per block in the POW phase.

50. 
See Network.cs : 799

src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
zeptin: I am using a somewhat different construct to the other tests, in this case the node gets started as soon as it is built. Note the 'true' argument to CreateStratisPosNode.

src/Stratis.Bitcoin.IntegrationTests/StratisMinerTests.cs
zeptin: Ok, I will amend that.

PR #1017, opened 2018-01-03
[PeerAddress] Update peer address to correctly determine when the last peer connection or handshake was made

### LINKED ISSUE #1012: [PeerAddressManager] Bug when selecting connected/handshaked peers to connect to.
After @Aprogiena's PR #953 , the node now sets the connected and handshaked time stamps in the correct order (prior to this the node was setting handshaked before the connected timestamp). Thus, PeerAddressManager needs to be updated/fixed to ensure that the logic correctly returns connected and/or handshaked peers.

PS I know how to fix this and will submit PR soon.

fassadlr:
Fixes #1012

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddress.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddress.cs
@@ -12,14 +12,6 @@ namespace Stratis.Bitcoin.P2P
     [JsonObject]
     public sealed class PeerAddress
     {
-        private const int PeerAddressLastSeen = 30;
-
-        private const int PeerMinimumFailDays = 7;
-
-        private const int PeerMaximumWeeklyAttempts = 10;
-
-        private const int PeerMaximumConnectionRetries = 3;
-
         /// <summary>EndPoint of this peer.</summary>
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -88,7 +80,8 @@ public bool Attempted
             {
                 return
                     this.LastConnectionAttempt != null &&
```
noescape00: please add () here and below

fassadlr: Fixed.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -192,12 +177,6 @@ internal void SetConnected(DateTimeOffset peerConnectedAt)
         }
 
         /// <summary>Sets the <see cref="LastConnectionHandshake"/> date.</summary>
-        /// <remarks>
```
dangershony: Comment for the `peerHandshakedAt` parameter.

PR #1019, opened 2018-01-03
IBD provider: moved test only method from interface to test-only environment mock up

noescape00:
This PR removes test-only method 'SetIsInitialBlockDownload' from the non-test codebase. 
Test implementation of 'IInitialBlockDownloadState' is added for the test environment by replacing the default service with a mock at the node build stage.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
@@ -6,7 +6,7 @@ using Stratis.Bitcoin.Features.Consensus;
 
 namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
 {
-    public class MockedInitialBlockDownloadState : InitialBlockDownloadState
+    public class InitialBlockDownloadStateMock : InitialBlockDownloadState
     {
         /// <summary>Time until IBD state can be checked.</summary>
         private DateTime lockIbdUntil;
@@ -14,7 +14,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
         /// <summary>A cached result of the IBD method.</summary>
         private bool ibdCached;
 
-        public MockedInitialBlockDownloadState(ChainState chainState, Network network, NodeSettings nodeSettings,
+        public InitialBlockDownloadStateMock(ChainState chainState, Network network, NodeSettings nodeSettings,
             ICheckpoints checkpoints) : base (chainState, network, nodeSettings, checkpoints)
         {
             this.lockIbdUntil = DateTime.MinValue;
--- a/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
@@ -12,7 +12,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
         private DateTime lockIbdUntil;
 
         /// <summary>A cached result of the IBD method.</summary>
-        private bool ibdCached;
+        private bool blockDownloadState;
 
         public InitialBlockDownloadStateMock(ChainState chainState, Network network, NodeSettings nodeSettings,
             ICheckpoints checkpoints) : base (chainState, network, nodeSettings, checkpoints)
@@ -23,7 +23,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
         public override bool IsInitialBlockDownload()
         {
             if (this.lockIbdUntil >= this.dateTimeProvider.GetUtcNow())
-                return this.ibdCached;
+                return this.blockDownloadState;
 
             return base.IsInitialBlockDownload();
         }
@@ -37,7 +37,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
         public void SetIsInitialBlockDownload(bool blockDownloadState, DateTime lockStateUntil)
         {
             this.lockIbdUntil = lockStateUntil;
-            this.ibdCached = blockDownloadState;
+            this.blockDownloadState = blockDownloadState;
         }
     }
 }
--- a/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
@@ -40,13 +40,12 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
 
         /// <summary>
         /// Sets last IBD status update time and result.
-        /// <para>Used in tests only.</para>
         /// </summary>
         /// <param name="blockDownloadState">New value for the IBD status, <c>true</c> means the node is considered in IBD.</param>
-        /// <param name="lockStateUntil">Time until IBD state won't be changed.</param>
-        public void SetIsInitialBlockDownload(bool blockDownloadState, DateTime lockStateUntil)
+        /// <param name="lockIbdUntil">Time until IBD state won't be changed.</param>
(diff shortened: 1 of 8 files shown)


fassadlr:
I'm reviewing this one 👍

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/MockedInitialBlockDownloadState.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using NBitcoin;
+using Stratis.Bitcoin.Base;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Features.Consensus;
+
+namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
+{
+    public class MockedInitialBlockDownloadState : InitialBlockDownloadState
```
fassadlr: @noescape00 I'm not too sure that mocks should inherit from the actual implementation like this? I'm used to a mocks implementing an interface and defining (custom) logic in contract members and/or methods. It'll mean possibly duplicating the logic in IsInitialBlockDownload() but is all that logic necessary for the unit tests when in fact we are mocking the IBD provider?

Also, when we call NotInIBD:

` (this.FullNode.NodeService<IInitialBlockDownloadState>() as MockedInitialBlockDownloadState).SetIsInitialBlockDownload(false, DateTime.UtcNow.AddMinutes(5));`

We don't have to cast.

**Also**, I would rename to `MockedInitialBlockDownloadState` to `InitialBlockDownloadStateMock`.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/MockedInitialBlockDownloadState.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Sets last IBD status update time and result.
+        /// <para>Used in tests only.</para>
+        /// </summary>
+        /// <param name="blockDownloadState">New value for the IBD status, <c>true</c> means the node is considered in IBD.</param>
+        /// <param name="lockStateUntil">Time until IBD state won't be changed.</param>
+        public void SetIsInitialBlockDownload(bool blockDownloadState, DateTime lockStateUntil)
+        {
+            this.lockIbdUntil = lockStateUntil;
+            this.ibdCached = blockDownloadState;
```
fassadlr: Can we maybe also rename `idbCached` to `blockDownloadState`?

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/MockedInitialBlockDownloadState.cs
noescape00: >I'm not too sure that mocks should inherit from the actual implementation like this?

Yes, that was the intention. In some tests we actually want a real IBD state, in some we would like to ignore it. So unless mocked IBD state is set this one will act exactly like real environment IBD state provider. 

> It'll mean possibly duplicating the logic in IsInitialBlockDownload() 

You mean implementing just the interface and copy paste the rest? I though about this but if we do it like that a scenario when we change full node IBD state provider and forget about the test mock is possible and we would like to avoid that. 

>We don't have to cast.

'SetIsInitialBlockDownload' is not a part of the interface so we kind of have to. Or did I miss something?

noescape00: >Also, I would rename to MockedInitialBlockDownloadState to InitialBlockDownloadStateMock.

Good idea, will do.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/MockedInitialBlockDownloadState.cs
noescape00: Sure

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/MockedInitialBlockDownloadState.cs
fassadlr: >You mean implementing just the interface and copy paste the implementation? 

Correct.

> I though about this but if we do it like that a scenario when we change full node IBD state provider and forget about the test mock is possible so we would like to avoid that.

You will still register the mock in feature registration for unit tests, as you did now, it's just that the mock will implement the interface and not inherit from the actual implementation. 

I'm not too much of a fan of making methods virtual for the sake of unit tests. Because you are implying that logic in there can be overridden in production code. That's why implementing the interface is better design for creating a mock test class. It feels funny extending production code in test code, that's my feeling on this. Yes we are duplicating logic but if we ever delete IBD provider, the actual provider will remain unchanged.

> 'SetIsInitialBlockDownload' is not a part of the interface so we kind of have to. Or did I miss something?

Sorry you are right :)

@noescape00 I'm just thinking out load here, so am open to discussion. What do you think?

noescape00: >I'm not too much of a fan of making methods virtual for the sake of unit tests. 

Hmm, makes sense. 

I think we can do it even better: avoid copy pasting the implementation code and avoid making the method virtual for the production code by just making an inner IBD state provider inside the mock.

noescape00: Fixed. 
@fassadlr let me know if you're happy with the new approach

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            this.innerBlockDownloadState = new InitialBlockDownloadState(chainState, network, nodeSettings, checkpoints);
+        }
+
+        public bool IsInitialBlockDownload()
+        {
+            if (this.lockIbdUntil >= this.dateTimeProvider.GetUtcNow())
+                return this.blockDownloadState;
+
+            return this.innerBlockDownloadState.IsInitialBlockDownload();
```
fassadlr: Dude. YES. 

Now if code changes in the actual provider we don't have to worry about changing it in here as well. High five!

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        /// <summary>
+        /// Sets last IBD status update time and result.
+        /// <para>Used in tests only.</para>
+        /// </summary>
+        /// <param name="blockDownloadState">New value for the IBD status, <c>true</c> means the node is considered in IBD.</param>
+        /// <param name="lockStateUntil">Time until IBD state won't be changed.</param>
+        public void SetIsInitialBlockDownload(bool blockDownloadState, DateTime lockStateUntil)
```
fassadlr: One last thing, lets rename `lockStateUntil` to `lockIbdUntil`  so that they match.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (this.lockIbdUntil >= this.dateTimeProvider.GetUtcNow())
+                return this.blockDownloadState;
+
+            return this.innerBlockDownloadState.IsInitialBlockDownload();
+        }
+
+        /// <summary>
+        /// Sets last IBD status update time and result.
+        /// <para>Used in tests only.</para>
```
fassadlr: Just remember to drop this comment ;)

noescape00: Oh, right. 
Now this comment obsolete since its obvious that this is only for testing =)

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/InitialBlockDownloadStateMock.cs
noescape00: Fixed

Issue #973, opened 2017-12-22
Should Propagation Algo be changed from 50% to 100% of Peers


D3m0nKingx:
>The propagation algorithm works like this:
>
>select half of the peers and send them the transaction
wait other half to propagate it back. if they did or we received a block with out tx- consider it propagated.

@noescape00 @fassadlr we can continue this matter on this thread.

Imho 100% of connected peers should have transactions broadcasted, giving more likelihood for a them to be propagated, and furthermore making it less likely for an attacker to be able to cripple the network. 

However, at this time this is theoretical and a stress test should be performed to monitor the performance/functionality of the network.

zeptin:
I am jumping in here without any context, but I think the reason for the 50% figure is because the light wallet needs to know that the transaction has been broadcast (as it does not have a mempool). If the transaction is broadcast to 100% of light wallet peers, none of them will (I believe) send it back to the light wallet node, so you have no straightforward way to track if the transaction has actually propagated until it appears in a block.

D3m0nKingx:
> If the transaction is broadcast to 100% of light wallet peers, none of them will (I believe) send it back to the light wallet node

@zeptin do you mean the whole light wallet network of peers or just the ones connected to a single node?

noescape00:
@D3m0nKingx 

All the nodes (no matter full or light) on average have a target of 8 outbound connections. 

On light wallet you don't have a consensus and a mempool so you are not able to validate even your own transaction. So if we send a tx to 50% and get it propagated back from any node from the other 50% it not only means that tx was propagated but also that the tx was validated by the network. 

If we send the tx to 100% of the peers we have no way of knowing if it even was valid.

D3m0nKingx:
Validated, meaning included in a block, which is also confirmed? not simply cause a node says "I got the transaction". 

Also wouldn't both the first half and the other half of the connected peers report back the tx was propagated/validated?  

Actually, being this is a POS system, that means the highest stakers are the only nodes that actually are reporting the official transaction validations... so not sure the logic here of using 50 / 50. Unless I'm not exactly understanding how the nodes are communicating, but I'm thinking which ever node forges the block then broadcasts it to all connected peers, and those peers broadcast it to theirs, so on and so forth.

So if block propagation is 100%, why not transactions....

noescape00:
>Validated, meaning included in a block,

[...]

PR #1020, opened 2018-01-03
Propagate tx to 100% of the peers for the full node.

noescape00:
For the full node (where we have consensus and can validate the tx) changed propagation from 50% of the peers to 100%. 

Fix: #973

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Broadcasting/BroadcasterManagerBase.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using ConcurrentCollections;
@@ -6,6 +7,7 @@ using NBitcoin;
 using Stratis.Bitcoin.Broadcasting;
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Interfaces;
+using Stratis.Bitcoin.P2P.Peer;
 using Stratis.Bitcoin.P2P.Protocol.Payloads;
 using Stratis.Bitcoin.Utilities;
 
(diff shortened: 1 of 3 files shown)


fassadlr:
@noescape00 This scenario calls for an abstract method to be implemented. I really don't like when booleans are used in this way as it's not explicit as to what the meaning of the boolean is or what it is used for.

Seeing that `BroadcasterManagerBase` is abstract, why not define an abstract method called `OnPropagateTransactionToPeersAsync` and implement the propagation to peers in `FullNodeBroadcasterManager` and `LightWalletBroadcasterManager`. So in FullNode we do all of them and in LightWallet only half. 

So in PropagateTransactionToPeersAsync you would go:

```
protected async Task PropagateTransactionToPeersAsync(Transaction transaction)
{
    this.AddOrUpdate(transaction, State.ToBroadcast);
    var invPayload = new InvPayload(transaction);
    var peers = this.connectionManager.ConnectedNodes.ToList();

    OnPropagateTransactionToPeersAsync();
}
```

And in FullNode's implementation of OnPropagateTransactionToPeersAsync():

```
    int propagateToCount = peers.Count;

    for (int i = 0; i < propagateToCount; ++i)
        await peers[i].SendMessageAsync(invPayload);
```

What do you think?

noescape00:
>What do you think?

Absolutely agree. Just wanted to separate refactoring from the algho changes.
Will do this in this PR then.

fassadlr:
> Will do this in this PR then.

Awesome 👍

noescape00:
I guess a cleaner solution is to pass the peers to 'PropagateTransactionToPeersAsync'. 
Otherwise if we implement 'OnPropagateTransactionToPeersAsync' for both light and full nodes we will have a lot of copy pasted code which I'd really like to avoid.

fassadlr:
Thats true, OK how about just implementing an abstract property call PeerPropagationCount and implement in each class? The property will just return the amount of peers to propagate to.

noescape00:
>Thats true, OK how about just implementing an abstract property call PeerPropagationCount and implement in each class? The property will just return the amount of peers to propagate to.

It's good for now but we will have to refactor it soon. 
There was an issue about propagating txes to the IBD peers so in the future we will have to sort the peers by .IsSynced for the light node so we always prioritize peers that are not in IBD. 
So I'd better just pass a list with the peers we'd like to propagate to and let LWBM and FNBM decide what peers to select for the propagation.

noescape00:
>Good job on the refactor dude. Looks much cleaner now!

Thank you for the review! =)

PR #1022, opened 2018-01-04
Code cleanup: redundant fields,variables,usings

noescape00:
Fixed some of the resharper's warnings that were pointing out redundant stuff. 
No logic has been changed.


src/Stratis.Bitcoin.Features.RPC/ModelBinders/DestinationModelBinder.cs
```diff
@@ -19,14 +19,9 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
                 return Task.CompletedTask;
             }
 
-            ValueProviderResult val = bindingContext.ValueProvider.GetValue(
-                bindingContext.ModelName);
-            if (val == null)
```
fassadlr: How come this `if` was dropped?

src/Stratis.Bitcoin.Features.RPC/ModelBinders/MoneyModelBinder.cs
```diff
@@ -13,14 +13,9 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
                 return Task.CompletedTask;
             }
 
-            ValueProviderResult val = bindingContext.ValueProvider.GetValue(
-                bindingContext.ModelName);
-            if (val == null)
```
fassadlr: The same with this one?

src/Stratis.Bitcoin.Features.RPC/ModelBinders/DestinationModelBinder.cs
noescape00: val can't be null here because it's not of a nullable type. ValueProviderResult is a struct which never can be null => therefore `if (val == null)` is always false.

src/Stratis.Bitcoin.Features.RPC/ModelBinders/MoneyModelBinder.cs
noescape00: Yup.

PR #1029, opened 2018-01-06
Feature/async queue dequeue

Aprogiena:
This PR enriches our current AsyncQueue to support blocking dequeue operation (but asynchronous!), which will allow us to use it instead of BlockingCollection, which is used in PollMessageListener, which is used by NetworkPeerListener, which I'm interested to fix (convert to full async) as a part of the node refactoring job.

The blocking dequeue is little more complex and not trivial, therefore I needed to introduce a lock into async queue, so I got rid of concurrent queue to avoid double locking.

Tests are included.

### DIFF
--- a/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
@@ -47,7 +47,7 @@ namespace Stratis.Bitcoin.Utilities
         private readonly CancellationTokenSource cancellationTokenSource;
 
         /// <summary>Number of pending dequeue operations which need to be finished before the queue can fully dispose.</summary>
-        private int unfinishedDequeueCount;
+        private volatile int unfinishedDequeueCount;
 
         /// <summary><c>true</c> if <see cref="Dispose"/> was called, <c>false</c> otherwise.</summary>
         private bool disposed;
--- a/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
@@ -53,7 +53,7 @@ namespace Stratis.Bitcoin.Utilities
         private bool disposed;
 
         /// <summary><c>true</c> if the queue operates in callback mode, <c>false</c> if it operates in blocking dequeue mode.</summary>
-        private bool callbackMode;
+        private readonly bool callbackMode;
 
         /// <summary>
         /// Initializes the queue either in blocking dequeue mode or in callback mode.
@@ -126,7 +126,7 @@ namespace Stratis.Bitcoin.Utilities
         public async Task<T> DequeueAsync(CancellationToken cancellation = default(CancellationToken))
         {
             if (this.callbackMode)
-                throw new InvalidOperationException();
+                throw new InvalidOperationException($"{nameof(DequeueAsync)} called on queue in callback mode.");
 
             // Increment the counter so that the queue's cancellation source is not disposed when we are using it.
             Interlocked.Increment(ref this.unfinishedDequeueCount);
--- a/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -18,7 +17,7 @@ namespace Stratis.Bitcoin.Utilities
     /// </para>
     /// </summary>
     /// <typeparam name="T">Type of items to be inserted in the queue.</typeparam>
-    public class AsyncQueue<T>: IDisposable
+    public class AsyncQueue<T> : IDisposable
     {
         /// <summary>
         /// Represents a callback method to be executed when a new item is added to the queue.
@@ -99,7 +98,6 @@ namespace Stratis.Bitcoin.Utilities
                 {
                     // Wait for an item to be enqueued.
                     await this.signal.WaitAsync(cancellationToken).ConfigureAwait(false);
-                    this.signal.Reset();
 
                     // Dequeue all items and execute the callback.
                     T item;
@@ -150,18 +148,8 @@ namespace Stratis.Bitcoin.Utilities
 
                         // Note that another thread could consume the message before us, 
                         // so dequeue safely and loop if nothing is available.
-                        lock (this.lockObject)
-                        {
-                            if (this.items.Count > 0)
-                            {
-                                item = this.items.Dequeue();
-
-                                if (this.items.Count == 0)
-                                    this.signal.Reset();
-
-                                return item;
-                            }
-                        }
+                        if (this.TryDequeue(out item))
+                            return item;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -93,11 +115,101 @@ public void Enqueue(T item)
...
+        /// If the queue is empty, the method waits until an item is available.
+        /// </summary>
+        /// <param name="cancellation">Cancellation token that allows aborting the wait if the queue is empty.</param>
+        /// <returns>Dequeued item from the queue.</returns>
+        /// <exception cref="OperationCanceledException">Thrown when the cancellation token is triggered or when the queue is disposed.</exception>
+        /// <exception cref="InvalidOperationException">Thrown if this method is called on a queue that operates in callback mode.</exception>
+        public async Task<T> DequeueAsync(CancellationToken cancellation = default(CancellationToken))
+        {
+            if (this.callbackMode)
+                throw new InvalidOperationException();
```
noescape00: I think it will be a good idea to let user know why we're throwing this. Maybe add a message like: 
"Dequeue is not supported if 'onEnqueue' callback is specified."

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -39,19 +46,28 @@ public class AsyncQueue<T>: IDisposable
...
         private readonly CancellationTokenSource cancellationTokenSource;
 
+        /// <summary>Number of pending dequeue operations which need to be finished before the queue can fully dispose.</summary>
+        private volatile int unfinishedDequeueCount;
+
+        /// <summary><c>true</c> if <see cref="Dispose"/> was called, <c>false</c> otherwise.</summary>
+        private bool disposed;
+
+        /// <summary><c>true</c> if the queue operates in callback mode, <c>false</c> if it operates in blocking dequeue mode.</summary>
+        private bool callbackMode;
```
noescape00: callbackMode can be made readonly

Aprogiena: tx, will fix

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: ok, will add message in this exception, tx

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -93,11 +113,95 @@ public void Enqueue(T item)
...
+                if (this.disposed)
+                    throw new OperationCanceledException();
+
+                // First check if an item is available. If it is, just return it.
+                T item;
+                if (this.TryDequeue(out item))
+                    return item;
+
+                // If the queue is empty, we need to wait until there is an item available.
+                using (var cancellationSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, this.cancellationTokenSource.Token))
```
dangershony: Why use a linked source at no point you are cancelling the cancellation.

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -93,11 +113,95 @@ public void Enqueue(T item)
...
+                {
+                    while (true)
+                    {
+                        await this.signal.WaitAsync(cancellationSource.Token).ConfigureAwait(false);
```
dangershony: If cancellation is called this will throw an Operation Cancellation exception right?
then the `Interlocked.Decrement(ref this.unfinishedDequeueCount);` will decrement by one, later bellow this might block the dispose code that uses this
```
 while (this.unfinishedDequeueCount > 0)
                Thread.Sleep(5);
```

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: i use it in tests, do you want me to remove it and explicitly use none token there in test?

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: yes cancellation will cause this to throw
then yes, decrement is called
why would this block dispose?

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
dangershony: Hmm to add a linked cancellation for tests sounds wrong, I think there is a performance hit for creating a linked token and potentially you might leak handles if not disposed correctly.

If those are not concerns (as we do dispose the linked token) then perhaps its not a big deal but we still don't have a usecase, I would prefer to remove it but wont insist.

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
dangershony: If cancellation throws you will decrements the counter only by one, what if there are more items in the queue?

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: it is not linked for the tests, the link will be there always, for the tests there is the default value, so what i'm saying is that i can remove the default value, but the link will still be there

there is no leak, we are using `using`, so no leaks

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: the decrement is for the number of pending dequeue operations (think threads), so when you enter dequeue operation you increase by 1, when you finish it, you decrease by one

this try-finally makes sure that every increase has its decrease

it has nothing to do with the content of the queue

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
dangershony: What about perf hit? we are creating a linked cancellation token for every loop in the iteration.

Aprogiena: the cancellation is needed
perf hit is minimal, this dequeue operation is supposed to be blocking, so most of the time you wait for the item to arrive

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
dangershony: Ah my bad I did not read the logic correctly, the counter represents dequeuing threads not items.

Brilliant.

I would suggest a rename in this case `unfinishedDequeueCount` to `listenerCount`

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
dangershony: Good point ok I will take that.

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
Aprogiena: it is weird because i chose such a not so nice name "unfinishedDequeueCount" in order to be so clear that it is really the number of unfinished dequeue operations :-) and its comment was supposed to make it even more clear 

listenerCount to me sounds not so explicit

dangershony: hmm I didnt read the documentation thoroughly.
How ever `unfinishedDequeueCount` is feels like its talking about items, if anything then name it `unfinishedDequeue[r]Count`

But I think listener does explain exactly the what this count is for, how many threads (or callers) are currently waiting in the `unfinishedDequeueCount` counter.
and ListenerCount sounds better then DequeuerCount.

However this is really semantics if you don't think the name change is more correct let push it as is.

Issue #1031, opened 2018-01-06
Node shutdown is slow if many -connect or -addNode options are specified


noescape00:
Node stucks on 

`Stratis.Bitcoin.Utilities.AsyncLoop.Dispose Waiting for PeerConnectorDiscovery.ConnectAsync to finish.`

if many -connect or -addNode options are specified and the node was shutted down within several seconds after it was started. 

I've added 20 -addnode statements and some logging to the PeerConnectorAddNode:OnConnectAsync() in the beginning of the foreach statement that processes those  -addNode and here is the log: 

```
[2018-01-06 21:18:52.9328 9] PeerConnectorAddNode.OnConnectAsync 13.64.76.48 - start
[2018-01-06 21:18:53.8416 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 186.188.171.179 - start
[2018-01-06 21:18:58.6498 39] PeerConnectorAddNode.<OnInitialize>b__2_0 ===ApplicationStopping token canceled===
[2018-01-06 21:18:59.4719 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 40.76.18.114 - start
[2018-01-06 21:19:00.3421 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 51.141.45.80 - start
[2018-01-06 21:19:01.0690 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 84.192.155.226 - start
[2018-01-06 21:19:01.6198 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 104.182.157.181 - start
[2018-01-06 21:19:02.0411 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 104.215.26.11 - start
[2018-01-06 21:19:02.7999 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 121.50.46.46 - start
[2018-01-06 21:19:03.5475 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 13.91.95.87 - start
[2018-01-06 21:19:04.5437 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 200.107.155.110 - start
[2018-01-06 21:19:05.1473 13] PeerConnectorAddNode+<OnConnectAsync>d__6.MoveNext 31.201.238.146 - start
[...]

PR #1032, opened 2018-01-06
Fix: slow node shutdown if many -connect or -addNode options were specified

noescape00:
Also improved some logging and peer disconnection reason (it was treated as timeout no matter what cancellation token was used- application stopping or linked one that is cancelled after 5 sec)

Fix: #1031

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnector.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnector.cs
@@ -220,8 +220,11 @@ namespace Stratis.Bitcoin.P2P
             }
             catch (OperationCanceledException timeout)
             {
-                this.logger.LogDebug("Peer {0} connection timeout.", peerAddress.NetworkAddress.Endpoint);
-                peer?.DisconnectWithException("Timeout", timeout);
+                this.logger.LogDebug(this.nodeLifetime.ApplicationStopping.IsCancellationRequested? "Peer {0} connection canceled because application is stopping." :
+                    "Peer {0} connection timeout.", peerAddress.NetworkAddress.Endpoint);
+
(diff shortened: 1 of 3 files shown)


noescape00:
@fassadlr I believe peers connectivity is your realm so could you please review this one?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -220,8 +220,11 @@ public void StartConnectAsync()
             }
             catch (OperationCanceledException timeout)
             {
-                this.logger.LogDebug("Peer {0} connection timeout.", peerAddress.NetworkAddress.Endpoint);
-                peer?.DisconnectWithException("Timeout", timeout);
+                this.logger.LogDebug(this.nodeLifetime.ApplicationStopping.IsCancellationRequested? "Peer {0} connection canceled because application is stopping." :
```
fassadlr: Can we maybe rewrite this so that we evaluate `this.nodeLifetime.ApplicationStopping.IsCancellationRequested` once? i.e.

```
if (this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
{    
    this.logger.LogDebug("Peer {0} connection canceled because application is stopping.", peerAddress.NetworkAddress.Endpoint);
    peer?.DisconnectWithException("Application stopping");
}
else
{
    this.logger.LogDebug("Peer {0} connection timeout.", peerAddress.NetworkAddress.Endpoint);
    peer?.DisconnectWithException("Timeout");
}  
```

noescape00: Sure.
Fixed.

noescape00:
Thanks, @fassadlr!

Issue #931, opened 2017-12-16
Make PeerConnectorDiscovery.MaximumNodeConnections a configurable option


noescape00:
Right now MaximumNodeConnections is set to 8 on initialization. 
Maybe it will be a good idea to make it configurable?

fassadlr:
@noescape00 I'm also for removing hard coded values in the solution. Currently we pass a lot of values via command line args... but maybe we can have settings.json file?

noescape00:
@fassadlr Yeah, having a single settings.json with all the constant's stat we're currently passing via args is a way to go. Could also make it very well structured and auto generated from the default values on the first node run. 

This is though pretty much out of scope of this issue so I'll create a new one.

PR #1033, opened 2018-01-08
MaximumNodeConnections : rename and config

noescape00:
Renamed MaximumNodeConnections  to MaxOutboundConnections since prev name was misleading and made people assume that this value is limiting overall connections count.  

Made MaxOutboundConnections configurable.
  
Fix: #931

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
@@ -40,7 +40,7 @@ namespace Stratis.Bitcoin.P2P
         /// <inheritdoc/>
         public override void OnInitialize()
         {
-            this.MaximumNodeConnections = 8;
+            this.MaxOutboundConnections = 8;
         }
 
         /// <summary>This connector is only started if there are NO peers in the -connect args.</summary>
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
@@ -40,6 +40,7 @@ namespace Stratis.Bitcoin.P2P
         /// <inheritdoc/>
         public override void OnInitialize()
         {
+            // TODO: make sure that this is moved to a new config implementation when it's ready.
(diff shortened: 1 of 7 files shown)


noescape00:
@fassadlr can you please take a look at this one?

fassadlr:
@noescape00 yo dude, there's some tests failing ;)

noescape00:
>@noescape00 yo dude, there's some tests failing ;)

Oh, my fault. 
Didn't realize that 'NodeSettings.ConfigReader' could possibly be null but apparently it's not created if we've not called NodeSettings.LoadArguments. 

Fixed.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -40,7 +40,7 @@ public sealed class PeerConnectorDiscovery : PeerConnector
         /// <inheritdoc/>
         public override void OnInitialize()
         {
-            this.MaximumNodeConnections = 8;
+            this.MaxOutboundConnections = this.NodeSettings.ConfigReader.GetOrDefault("maxOutboundConnections", 8);
```
fassadlr: Should we maybe not set this to a property on NodeSettings? I'm just worried that if we don't, this property will get "buried" i.e. a future dev or user won't know that this configurable element of the node exists, unless they go dig in peer connector code? What do you think?

noescape00: > I'm just worried that if we don't, this property will get "buried" i.e. a future dev or user won't know that this configurable element of the node exists

As far as I know Jeremy wanted to completely change the approach we take to the config file (there is a card on trello for this) so after he does that we will have a json config file initialized with default values so it will be pretty straight forward to understand what can be configured and what can't.
This should solve the problem of  all "buried" properties, not just this one

fassadlr: OK all good then @noescape00. Can we maybe then just add a TODO comment so that we remember for the future?

noescape00: Sure, fixed.

PR #1037, opened 2018-01-08
Refactor/the node job part 18

Aprogiena:
First commit of this part is `Part 18 starts here`

Then the commits are made so that you can really track the progress. 

So this PR is:
- merging NP client class with NP connection class and separation of those into a new file
- correct shutdown procedure of NP connection class
- changes in other NP classes to work with the new connection class
- two small changes in IMessageProducer and IncomingMessage


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -607,11 +337,15 @@ private NetworkPeer(bool inbound, NetworkAddress peerAddress, Network network, N
         public NetworkPeer(NetworkAddress peerAddress, Network network, NetworkPeerConnectionParameters parameters, INetworkPeerFactory networkPeerFactory, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
             : this(false, peerAddress, network, parameters, dateTimeProvider, loggerFactory)
         {
-            NetworkPeerClient client = networkPeerFactory.CreateNetworkPeerClient(parameters);
-            this.logger = loggerFactory.CreateLogger(this.GetType().FullName, $"[{client.Id}-{peerAddress.Endpoint}] ");
-            this.logger.LogTrace("()");
+            TcpClient client = new TcpClient(AddressFamily.InterNetworkV6);
```
dangershony: client is not created in the factory anymore?

Aprogiena: this is different client than it was before (NP client vs TCP client), we got rid of NP client class and merged it with NP connection class. this is TCP client, that represents the very TCP connection that we will establish to connect to remote server

PR #1043, opened 2018-01-09
Mining Feature Tests 2

Neurosploit:
- Added interfaces to some classes for testability of the mining classes.
- Added unit tests for the AssemblerFactories, MinerSettings, BlockAssemblers and PowMining classes
- Fixed several typos
- Attempted to improve the test stability of the APITest integration tests

PosMinting will be next on my list but that class is very big and needs more time from me to understand it fully before I want to start creating tests for it.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Interfaces/IPowConsensusValidator.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Interfaces/IPowConsensusValidator.cs
@@ -1,0 +1,151 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using NBitcoin;
+using Stratis.Bitcoin.Base.Deployments;
+
+namespace Stratis.Bitcoin.Features.Consensus.Interfaces
+{
+    /// <summary>
+    /// Provides functionality for verifying validity of PoW block.
+    /// </summary>
+    public interface IPowConsensusValidator
+    {
+        /// <summary>Consensus options.</summary>
+        PowConsensusOptions ConsensusOptions { get; }
+
+        /// <summary>Consensus parameters.</summary>
+        NBitcoin.Consensus ConsensusParams { get; }
+
+        /// <summary>Keeps track of how much time different actions took to execute and how many times they were executed.</summary>
+        ConsensusPerformanceCounter PerformanceCounter { get; }
+
+        /// <summary>
+        /// Calculates merkle root for block's trasnactions.
+        /// </summary>
+        /// <param name="block">Block which transactions are used for calculation.</param>
+        /// <param name="mutated"><c>true</c> if block contains repeating sequences of transactions without affecting the merkle root of a block. Otherwise: <c>false</c>.</param>
+        /// <returns>Merkle root.</returns>
+        uint256 BlockMerkleRoot(Block block, out bool mutated);
+
+        /// <summary>
+        /// Calculates merkle root for witness data.
+        /// </summary>
+        /// <param name="block">Block which transactions witness data is used for calculation.</param>
+        /// <param name="mutated"><c>true</c> if at least one leaf of the merkle tree has the same hash as any subtree. Otherwise: <c>false</c>.</param>
+        /// <returns>Merkle root.</returns>
+        uint256 BlockWitnessMerkleRoot(Block block, out bool mutated);
+
+        /// <summary>
+        /// Checks block's validity.
+        /// </summary>
+        /// <param name="context">Context that contains variety of information regarding blocks validation and execution.</param>
+        /// <exception cref="ConsensusErrors.BadMerkleRoot">Thrown block's merkle root is corrupted.</exception>
+        /// <exception cref="ConsensusErrors.BadTransactionDuplicate">Thrown if block contains duplicated transaction that don't affect merkle root.</exception>
+        /// <exception cref="ConsensusErrors.BadBlockLength">Thrown if block exceeds maximum allowed size or doesn't contain any transaction.</exception>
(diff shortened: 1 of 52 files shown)


src/Stratis.Bitcoin.Features.Consensus/Interfaces/IPowConsensusValidator.cs
```diff
@@ -0,0 +1,151 @@
...
+
+namespace Stratis.Bitcoin.Features.Consensus.Interfaces
+{
+    /// <summary>
+    /// Provides functionality for verifying validity of PoW block.
+    /// </summary>
+    public interface IPowConsensusValidator
+    {
+        /// <summary>Consensus options.</summary>
+        PowConsensusOptions ConsensusOptions { get; }
```
dangershony: Hmmm its a pity you didn't check with me first, ideally this class `IPowConsensusValidator` will be gone at some point when we use the Rules framework.

src/Stratis.Bitcoin.Features.Miner.Tests/PosBlockAssemblerTest.cs
```diff
@@ -0,0 +1,575 @@
...
+
+            return chain;
+        }
+
+        private static Transaction CreateCoinStakeTransaction(Network network, Key key, int height, uint256 prevout)
+        {
+            var coinStake = new Transaction();
+            coinStake.AddInput(new TxIn(new OutPoint(prevout, 1)));
+            coinStake.AddOutput(new TxOut(0, new Script()));
+            coinStake.AddOutput(new TxOut(network.GetReward(height), key.ScriptPubKey));
```
dangershony: I am not sure `network.GetReward(height)` is the correct method for stake trx

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -134,7 +137,7 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong generate,
                     }
                 }
 
-                this.IncrementExtraNonce(pblockTemplate.Block, chainTip, nExtraNonce);
+                nExtraNonce = this.IncrementExtraNonce(pblockTemplate.Block, chainTip, nExtraNonce);
```
dangershony: `nExtraNonce ` -> `extraNonce `

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -208,6 +212,9 @@ public ApiTestsFixture()
             this.stratisPowNode.ConfigParameters.Add("apiuri", "http://localhost:37221");
 
             this.InitializeTestWallet(this.stratisPowNode);
+            this.builder.StartAll();
+
+            Block.BlockSignature = true;
```
dangershony: ` Block.BlockSignature = true;`
Are you sure this is correct?

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -230,6 +237,7 @@ public ApiTestsFixture()
         public void Dispose()
         {
             this.builder.Dispose();
+            Block.BlockSignature = this.initialBlockSignature;
```
dangershony: Ah this is the restore for point for `Block.BlockSignature`

dangershony:
Check the comments.
I will now get the branch and try out the tests.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
Neurosploit: @dangershony The Block.BlockSignature has to do with the nodes that are started. It expects NetworkOptions.IsProofOfStake to be true for PoS otherwise the tests will fail. See https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/NetworkOptions.cs I solve this by storing the initial value and restore it when the tests are done running as anything can be using that static. That also explains the dispose.

src/Stratis.Bitcoin.Features.Consensus/Interfaces/IPowConsensusValidator.cs
Neurosploit: As discussed privately it will not be an issue when replaced with the new rules engine. It is there for testing and can be removed as soon as that is put in place. Does not have a big impact on the newly created miner tests.

src/Stratis.Bitcoin.Features.Miner.Tests/PosBlockAssemblerTest.cs
Neurosploit: This is for the blocktemplate. In the actual PosMinting it's this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height) + fees. However that is not what is coming out of the blockassembler. I can change it to Money.Zero or just 0 to more closely reflect what the assembler is actually doing. What do you think?

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
Neurosploit: Ideally we get rid of statics but that's my opinion about them.

src/Stratis.Bitcoin.Features.Miner.Tests/PosBlockAssemblerTest.cs
dangershony: hmm but then it wont pass the consensus checks right?

Neurosploit: As these are unit tests the consensus checks are mocked in this code so it will pass. It however won't pass if we run this with consensus checks enabled.

PR #1045, opened 2018-01-10
Getting the first node connection speedup.

noescape00:
This change removes 1 second delay between connection attempts until we are connected to at least 1 peer- after that set it back to normal.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnector.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnector.cs
@@ -111,7 +111,10 @@ namespace Stratis.Bitcoin.P2P
         public int BurstModeTargetConnections { get; private set; }
 
         /// <summary>Default time interval between making a connection attempt.</summary>
-        private TimeSpan DefaultConnectionAttemptInterval;
+        private readonly TimeSpan defaultConnectionInterval;
+
+        /// <summary>Burst time interval between making a connection attempt.</summary>
+        private readonly TimeSpan burstConnectionInterval;
 
         /// <summary>Parameterless constructor for dependency injection.</summary>
         protected PeerConnector(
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -129,6 +135,9 @@ public abstract class PeerConnector : IPeerConnector
             this.NodeSettings = nodeSettings;
             this.peerAddressManager = peerAddressManager;
             this.Requirements = new NetworkPeerRequirement { MinVersion = this.NodeSettings.ProtocolVersion };
+
+            this.DefaultConnectionAttemptInterval = TimeSpans.Second;
```
fassadlr: @noescape00 how about we rename this to `defaultConnectionInterval`, make it a constant and create a second constant called `burstConnectionInterval`? Then use them where appropriate?

Thoughts?

noescape00: Sounds good to me. 
Fixed. 

ps
made them readonly because it's not possible to assign constant to non-compile time value (TimeSpan.Zero)

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -212,7 +216,7 @@ public void StartConnectAsync()
                 await this.OnConnectAsync().ConfigureAwait(false);
             },
             this.nodeLifetime.ApplicationStopping,
-            repeatEvery: TimeSpan.Zero);
+            repeatEvery: burstConnectionInterval);
```
fassadlr: this.burstConnectionInterval

noescape00: Oh, right =)

PR #1047, opened 2018-01-10
WIP: verifyaddress rpc StratisD/BitcoinD

DanGould:
* Supports the 4 address types in NBitcoin
* Only returns `isvalid` field because that's needed to verify which chain is being rpc called during tumbler registration. This hasn't been a problem in the past because the rpc would typically go to bitcoind. Now we're trying to rpc Stratis.StratisD
* WIP: will also need to at a minimum return `hdkeypath` field for [RPCDestinationWallet.cs](https://github.com/BreezeHub/BreezeProject/blob/97c3c96bc238b616cabf7cd0a955e1372b871fc0/NTumbleBit/ClassicTumbler/Client/RPCDestinationWallet.cs)

Question: Why is it still conflicting after that merge?


src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
...
+        /// <returns>information about <address>. {isvalid}</returns>
+        [ActionName("validateaddress")]
+        [ActionDescription("Returns information about a bitcoin address")]
+        public JObject ValidateAddress(string address)
+        {
+            if(address == null)
+                throw new ArgumentNullException("address");
+
+            JObject res = new JObject();
+            res["isvalid"] = false; // until proven otherwise
```
fassadlr: This comment isn't really necessary as its implied by the code ;)

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
...
+        /// <param name="address">BitcoinAddress to validate.</param>
+        /// <returns>information about <address>. {isvalid}</returns>
+        [ActionName("validateaddress")]
+        [ActionDescription("Returns information about a bitcoin address")]
+        public JObject ValidateAddress(string address)
+        {
+            if(address == null)
+                throw new ArgumentNullException("address");
+
+            JObject res = new JObject();
```
fassadlr: Can use `var` here.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
...
+        /// <summary>
+        /// Return information about <address>.
+        /// </summary>
+        /// <param name="address">BitcoinAddress to validate.</param>
+        /// <returns>information about <address>. {isvalid}</returns>
+        [ActionName("validateaddress")]
+        [ActionDescription("Returns information about a bitcoin address")]
+        public JObject ValidateAddress(string address)
+        {
+            if(address == null)
```
fassadlr: Maybe also add a check for string.Empty? I think you can use string.IsNullOrEmpty

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
             return model;
         }
 
+        /// <summary>
+        /// Return information about <address>.
+        /// </summary>
+        /// <param name="address">BitcoinAddress to validate.</param>
+        /// <returns>information about <address>. {isvalid}</returns>
```
fassadlr: Also not valid markup, how about:

`/// <returns>JObject containing a boolean indicating address validity.</returns>`

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
...
+        public JObject ValidateAddress(string address)
+        {
+            if(address == null)
+                throw new ArgumentNullException("address");
+
+            JObject res = new JObject();
+            res["isvalid"] = false; // until proven otherwise
+
+            // P2PKH
+            if(BitcoinPubKeyAddress.IsValid(address, ref this.Network))
```
fassadlr: With this logic you will effectively be evaluating each address type. It'll be better to return once a valid address has been found? Maybe look into refactoring this a bit 👍

src/NBitcoin/BitcoinAddress.cs
```diff
@@ -18,23 +18,30 @@ public BitcoinScriptAddress(string base58, Network expectedNetwork)
 
         private static string Validate(string base58, ref Network expectedNetwork)
         {
-            if(base58 == null)
+            if (IsValid(base58, ref expectedNetwork))
+                return base58;
+            throw new FormatException("Invalid BitcoinScriptAddress");
+        }
+
+        public static bool IsValid(string base58, ref Network expectedNetwork)
```
dangershony: This is a bit confusing, as we pass in a network and use it in determining validity then change it if we find a different valid network.

src/NBitcoin/BitcoinSegwitAddress.cs
```diff
@@ -28,16 +35,15 @@ private static string Validate(string bech32, ref Network expectedNetwork)
...
+                    if (data.Length == 20 && witVersion == 0)
                     {
-                        expectedNetwork = network;
-                        return bech32;
+                        return true;
                     }
                 }
-                catch(Bech32FormatException) { throw; }
-                catch(FormatException) { continue; }
+                catch (Bech32FormatException) { throw; }
```
dangershony: redundant line ` catch (Bech32FormatException) { throw; }`

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
DanGould: Really? I was under the impression `else if` should only execute if the prior conditional control statements evaluate to false i.e. no valid address has been found. Else the result will be returned. I chose not to return because bitcoin/bitcoin has other fields returned in res, of which we need at least one for TumbleBit (HdAddress)

src/NBitcoin/BitcoinAddress.cs
DanGould: Agree. Refactored to only validate with the expectedNetwork param & no loop

src/NBitcoin/BitcoinSegwitAddress.cs
DanGould: I assumed you were talking about the one-after-another `FormatException`s, which have been removed.

DanGould:
i also found this, but am not sure if the class offers a solution. Wallet's serviceProvider doesn't carry the network definition. [IsBitcoinAddressAttribute.cs](https://github.com/stratisproject/StratisBitcoinFullNode/blob/a512de09f03841c74fee75368d5a42a4064569ab/src/Stratis.Bitcoin.Features.Wallet/Validations/IsBitcoinAddressAttribute.cs)

dangershony:
I am good with this @fassadlr will also approve

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -175,6 +177,44 @@ public BlockHeaderModel GetBlockHeader(string hash, bool isJsonFormat = true)
...
         }
 
+        /// <summary>
+        /// Returns information about a bitcoin address
+        /// </summary>
+        /// <param name="address">bech32 or base58 BitcoinAddress to validate.</param>
+        /// <returns>JObject containing a boolean indicating address validity</returns>
+        [ActionName("validateaddress")]
+        [ActionDescription("Returns information about a bech32 or base58 bitcoin address")]
+        public JObject ValidateAddress(string address)
```
fassadlr: @DanGould Can you please explain why you chose to return a JObject instead of a JSonResult with the boolean value in it? As far as I know JsonResults are far easier to interpret by a json ajax call etc. Please correct me if I'm wrong ;)

DanGould: I was writing this method particularly for this use case, which is why I used JObject. I can see JsonResult being friendlier for other applications. I haven't yet been able to get the noted method to work by returning JsonResult. It would make sense for that Log in (RPCArgs.cs)[https://github.com/BreezeHub/BreezeProject/blob/97c3c96bc238b616cabf7cd0a955e1372b871fc0/NTumbleBit/Configuration/RPCArgs.cs] to use a NBitcoin method for validateaddress anyway, so it probably should be migrated to JsonResult

```
try
			{
				var address = new Key().PubKey.GetAddress(network);
				var isValid = ((JObject)rpcClient.SendCommand("validateaddress", address.ToString()).Result)["isvalid"].Value<bool>();
				if(!isValid)
				{
					Logs.Configuration.LogError("The RPC Server is on a different blockchain than the one configured for tumbling");
					throw new ConfigException();
				}
			}
```

fassadlr: @DanGould Ah I see OK! 

JsonResult is definitely what should be used here but yeah it won't work if Breeze expects a JObject (fail)

Can you perhaps maybe submit a PR to change that? And then once done fix this again?

In the meantime we can push this as is but just create an issue in BreezeHub 👍

PR #1049, opened 2018-01-11
Wallet endpoint to return addresses

bokobza:
This additional endpoint returns all the addresses ocntained in a wallet account to display to the user.


src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -784,6 +784,47 @@ public IActionResult GetUnusedAddresses([FromQuery]GetUnusedAddressesModel reque
...
+                return BuildErrorResponse(this.ModelState);
+            }
+
+            try
+            {
+                Wallet wallet = this.walletManager.GetWallet(request.WalletName);
+                HdAccount account = wallet.GetAccountByCoinType(request.AccountName, this.coinType);
+                AddressesModel model = new AddressesModel();
+                
+                foreach (HdAddress address in account.GetCombinedAddresses())
```
fassadlr: How about this?

```
var addresses = account.GetCombinedAddresses().Select(a => new AddressModel(a));
model.Addresses.AddRange(addresses);
```

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -732,7 +732,7 @@ public IActionResult CreateNewAccount([FromBody]GetUnusedAccountModel request)
         /// Gets an unused address.
         /// </summary>
         /// <returns>The last created and unused address or creates a new address (in Base58 format).</returns>
-        [Route("address")]
+        [Route("newaddress")]
```
fassadlr: Won't `[Route("unusedaddress")]` be a better name perhaps?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -759,7 +759,7 @@ public IActionResult GetUnusedAddress([FromQuery]GetUnusedAddressModel request)
         /// <summary>
         /// Gets the specified number of unused addresses.
         /// </summary>
-        [Route("addresses")]
+        [Route("newaddresses")]
```
fassadlr: Same here, won't `[Route("unusedaddresses")]` be a better name perhaps?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza: yeah can do.

bokobza: yeah can do

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza: yeah adds more clarity as to how we select the new address.
same for below comment.
Thanks :-)

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -784,6 +784,45 @@ public IActionResult GetUnusedAddresses([FromQuery]GetUnusedAddressesModel reque
             }
         }
 
+        /// <summary>
+        /// Gets the specified number of unused addresses.
+        /// </summary>
+        [Route("addresses")]
+        [HttpGet]
+        public IActionResult GetAllAddresses([FromQuery]GetAllAddressesModel request)
```
dangershony: Why do we need such a method I wonder?

bokobza: because at the moment, users have no idea what addresses they have in their wallet.
All they can ever know is the first unused address.
This gives them a bit more data and POWER.

dangershony: ok cool so push when tests are done.

PR #1050, opened 2018-01-11
[Bug] Seed node addresses are not added to the peer address manager during discovery

fassadlr:
Seeing as the seed node addresses aren't being added to the peer address manager's peer list during discovery, they aren't able to be selected in the peer connector.

Also, the peer discovery loop was incorrectly adding both DNS seed nodes addresses as well as the IP seed node addresses during discovery.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
@@ -54,7 +54,15 @@ namespace Stratis.Bitcoin.P2P
             if ((this.Mode & PeerAddressManagerBehaviourMode.Discover) != 0)
             {
                 if (this.AttachedPeer.State == NetworkPeerState.Connected)
+                {
+                    //If a peer connects via the peer discovery loop but does not yet exist
+                    //in the address manager's peer list, we need to add it. 
+                    //This could be a peer address provided by one of the seed nodes.
+                    if (this.peerAddressManager.FindPeer(this.AttachedPeer.PeerAddress.Endpoint) == null)
+                        this.peerAddressManager.AddPeer(this.AttachedPeer.PeerAddress, this.AttachedPeer.RemoteSocketAddress);
+
                     this.peerAddressManager.PeerConnected(this.AttachedPeer.PeerAddress.Endpoint, this.dateTimeProvider.GetUtcNow());
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
```diff
@@ -54,7 +54,15 @@ protected override void AttachCore()
             if ((this.Mode & PeerAddressManagerBehaviourMode.Discover) != 0)
             {
                 if (this.AttachedPeer.State == NetworkPeerState.Connected)
+                {
+                    //If a peer connects via the peer discovery loop but does not yet exist
```
dangershony: hmm I didnt understand this can you explain more?
When we find addresses using address discovery we add them to the address manger right?
Then the discover peer loop will try to connect so this should already be in the address manager.

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
```diff
@@ -167,26 +171,26 @@ private Task DiscoverPeersAsync()
...
         /// </summary>
         private void AddDNSSeedNodes(List<NetworkAddress> peers)
         {
-            peers.AddRange(this.network.DNSSeeds.SelectMany(seed =>
+            foreach (var seed in this.network.DNSSeeds)
             {
                 try
                 {
-                    return seed.GetAddressNodes();
+                    var seedAddresses = seed.GetAddressNodes().Select(ip => new NetworkAddress(ip, this.network.DefaultPort));
```
dangershony: Is `DefaultPort` correct? the previous code did not specify a port.

src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
fassadlr: @dangershony OK there are 2 scenarios, peer discovery with no peers file and peer discovery with peers file. The first scenario we have to get addresses from the seed nodes, which when one them is connected to, they aren't yet in the peer address manager's list. This is why we need to add them here if they don't exist.

I had a look at the old AddressManager:

[Old Discover Loop](https://github.com/stratisproject/StratisBitcoinFullNode/commit/50f4cd76aff090a4ddcf056803485d745b31d684#diff-5eabda437d068ea8bbe924ebee5950d9R1130)

And it behaved the same way 👍

src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
fassadlr: Yes if you look at the diff you'll see we created network addresses from the IP and default port. There's a select statement:

```.Select(d => new NetworkAddress(d, this.network.DefaultPort)));```

fassadlr:
Closing this PR as per video call discussion.

PR #1053, opened 2018-01-12
[Refactor] Remove NetworkAddress from PeerAddress

fassadlr:
We only really use the EndPoint property of NetworkAddress in PeerAddress so there's no point in having NetworkAddress as a property of PeerAddress. We already serialize a private end point field anyway.

In future PR's well move closer to removing NetworkAddress altogether.


dangershony:
Please fix conflicts, also can you confirm you tried to sync a node from scratch (delete the peer file and db folders)

fassadlr:
@dangershony I confirm node fully synced :) OK to merge?

PR #1055, opened 2018-01-12
Refactor/the node job part 19

Aprogiena:
first commit of this part is `Part 19 starts here`

This PR is:
- migration of message listener and producer from NBitcoin to FN 
- removes unused NewThreadMessageListener
- replacing EventLoopMessageListener with new async implementation called CallbackMessageListener
- migration of components using EventLoopMessageListener to CallbackMessageListener
- adding cancellation tokens where they were missing in NP, which allows cancelling of more calls than before (more to come in part 20)
- PollMessageListener is not touched here, this is refactored in part 20


src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -468,14 +468,15 @@ private void OnStateChanged(NetworkPeerState previous)
         /// Processes an incoming message from the peer and calls subscribed event handlers.
         /// </summary>
         /// <param name="message">Message received from the peer.</param>
```
noescape00: Since this one was renamed from `OnMessageReceived` to `ProcessMessageAsync` I guess it will be better to change this comment too. Smth like `Message to process` maybe.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -587,14 +590,14 @@ private void CallMessageReceivedHandlers(IncomingMessage message)
...
                 this.Disconnect("Connected to self");
 
                 this.logger.LogTrace("(-)[CONNECTED_TO_SELF]");
                 return;
             }
 
-            using (CancellationTokenSource cancellationSource = CancellationTokenSource.CreateLinkedTokenSource(this.Connection.CancellationSource.Token))
+            using (CancellationTokenSource cancellationSource = CancellationTokenSource.CreateLinkedTokenSource(this.Connection.CancellationSource.Token, cancellation))
             {
                 cancellationSource.CancelAfter(TimeSpan.FromSeconds(10.0));
```
noescape00: `this.cancellationSource.Cancel...`

src/Stratis.Bitcoin/Connection/PeerBanningBehavior.cs
```diff
@@ -49,13 +49,13 @@ protected override void AttachCore()
         {
             this.logger.LogTrace("()");
 
-            var node = this.AttachedPeer;
-            if (node.State == NetworkPeerState.Connected)
+            var peer = this.AttachedPeer;
```
noescape00: unvar

noescape00: For some reason Github wont show all files changed and would mark this comment as a resolved one. The changes for PeerBanningBehavior are still there but this file is not included in the [github changed files view](https://github.com/stratisproject/StratisBitcoinFullNode/pull/1055/files) 
Odd.

Aprogiena: if it is not on the list, it means this file haven't been changed in this PR, it could be in merge of other PR while syncing, but it is unrelated to this PR

therefore offtopic and won't fix, although you are right of course

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: i can't see such a conflicting field, where is it?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: this seems fine to me, the summary claims to process an incoming message from the peer and the param message description says this is the message received from the peer, `message received` seems to have the same meaning as `incoming message`, so i can't see a problem there, makes perfect sense to me

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: oh, sry, looks like this is smth that was removed already too.

noescape00:
Ran the tests and integration tests, all good (except for `PeerConnectorAddNode_PeerAlreadyConnected_Scenario1` but that one is fixed in the current master so after merging it's all fine)

So overall looks good to me.

PR #1057, opened 2018-01-12
Tests consistency fix

noescape00:
Tests consistency improvement.


dangershony:
@Aprogiena this is yours

Aprogiena:
oh, i think you are right :)
yes, i suggested to change it like that, so it is probably mine

noescape00:
Update: travis ran this one 4 or 5 times (I was removing and adding dummy commits to trigger it)- all good. So it looks like this fixes the test.

PR #1058, opened 2018-01-13
Peer connector threads usage optimizations 

noescape00:
Fixes the bug with high threads usage.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnector.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnector.cs
@@ -212,7 +212,7 @@ namespace Stratis.Bitcoin.P2P
             {
                 if (!this.peerAddressManager.Peers.Any() || (this.ConnectedPeers.Count >= this.MaxOutboundConnections))
                 {
-                    await Task.Delay(2000, this.nodeLifetime.ApplicationStopping);
+                    await Task.Delay(2000, this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
                     return;
                 }
 
--- a/src/Stratis.Bitcoin/P2P/PeerConnector.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnector.cs
@@ -205,9 +205,6 @@ namespace Stratis.Bitcoin.P2P
 
             this.OnStartConnect();
 
-            if (this.MaxOutboundConnections == 0)
-                return;
-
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -205,13 +205,16 @@ public void StartConnectAsync()
...
+
             this.asyncLoop = this.asyncLoopFactory.Run($"{this.GetType().Name}.{nameof(this.ConnectAsync)}", async token =>
             {
-                if (!this.peerAddressManager.Peers.Any())
-                    return;
-
-                if (this.ConnectedPeers.Count >= this.MaxOutboundConnections)
+                if (!this.peerAddressManager.Peers.Any() || (this.ConnectedPeers.Count >= this.MaxOutboundConnections))
+                {
+                    await Task.Delay(2000, this.nodeLifetime.ApplicationStopping);
```
Aprogiena: configure await false

D3m0nKingx:
Running node with this new code for 4 hours now. No excessive amount of threads (varies between 100-300), CPU % around 2-70% on avg with occasional spikes to 200-400%. RAM staying around 500mb. Also the node is in a state of IBD atm, currently at 90% sync.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
noescape00: Fixed

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -205,13 +205,16 @@ public void StartConnectAsync()
 
             this.OnStartConnect();
 
+            if (this.MaxOutboundConnections == 0)
```
fassadlr: @noescape00 The issue putting the MaxOutboundConnections outside the async loop is that we can add nodes to PeerConnectorAddNode's collection via an RPC call. So if we do this, this connector never starts up and as such when we call that RPC method, it won't have any affect. Remember that PeerConnectorAddNode must always run.

noescape00: Good notice. Didn't know that we can use addNode during the runtime. 
Fixed.

fassadlr:
@noescape00 @D3m0nKingx Are you guys happy with this now? If so I'll merge 👍

noescape00:
@fassadlr sure, got for it.

PR #1059, opened 2018-01-13
Memory cache LRU

noescape00:
This PR adds a LRU memory cache with size limit and uses it in BlockStoreCache

Fix: #1030


src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Linq;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    public class MemoryCache<TKey, TValue>
+    {
+        /// <summary>Dictionary that contains cached items.</summary>
+        private readonly ConcurrentDictionary<TKey, TValue> cache;
+
+        /// <summary>Keys sorted by time they when they were accessed last time with most recent ones in the end.</summary>
```
Aprogiena: Keys sort by their last access time with most recent ones at the end.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+        private readonly LinkedList<TKey> keys;
+
+        /// <summary>Maximum items count that can be stored in the cache.</summary>
+        private readonly int sizeLimit;
+
+        /// <summary>Amount to compact the cache by when the maximum size is exceeded.</summary>
```
Aprogiena: if it is percentage then it is not amount
also naming something Percentage might invoke one to think that a value of e.g. `10` would mean 10 %
when you have a variable that has such strict bounds, it is always a good idea to explicitly document the bounds
i'd say
`Percentage of the items in the cache to be removed when the maximum item count is reached. This has to be a number within (0; 1] interval.`

and it very much makes sense to assert this boundary in the constructor as well

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+        private readonly double compactionPercentage;
+
+        /// <summary>Lock to protect access to <see cref="keys"/>.</summary>
+        private readonly object mutex;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MemoryCache{TKey, TValue}"/> class.
+        /// </summary>
+        /// <param name="sizeLimit">Maximum items count that can be stored in the cache.</param>
```
Aprogiena: if you change a comment or name of these guys, above don't forget to do it here as well

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            this.sizeLimit = sizeLimit;
+            this.compactionPercentage = compactionPercentage;
+
+            this.cache = new ConcurrentDictionary<TKey, TValue>();
+            this.keys = new LinkedList<TKey>();
+            this.mutex = new object();
+        }
+
+        /// <summary>Gets the count of the current entries for diagnostic purposes.</summary>
```
Aprogiena: try to be consistent in how you call those guys - are they entries or items?

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    return this.keys.Count;
+                }
+            }
+        }
+
+        /// <summary>Create or overwrite an entry in the cache.</summary>
+        /// <param name="key">The key.</param>
+        /// <param name="value">The value to add to the cache.</param>
+        public void AddOrUpdate(TKey key, TValue value)
```
Aprogiena: you have two structures that you want to keep in sync
you need 1 lock for both of them,
not 2 locks, otherwise weird things can happen

don't use concurrent dictionary, but normal one and protect dictionary and the linked list with the mutex in one go
don't allow your structures to get out of sync like in this method

also even if there was no list, it would be wrong, you have an atomic logic of try get value and then try update separated from it, it can happen that after you get value, it won't exist in the cache anymore

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            if (this.cache.TryGetValue(key, out TValue priorEntry))
+            {
+                this.cache.TryUpdate(key, value, priorEntry);
+
+                lock (this.mutex)
+                {
+                    this.keys.Remove(key);
```
Aprogiena: this is a big mistake, we use linked list to have this done in O(1), you are doing it in O(N)

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+            else
+            {
+                this.cache.TryAdd(key, value);
+
+                lock (this.mutex)
+                {
+                    this.keys.AddLast(key);
+                }
+
+                if (this.Count >= this.sizeLimit)
```
Aprogiena: you should keep your invariants - i.e. there should never be more than max number of items in the cache
you should first compact the structure and only add then

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// <summary>Removes the object associated with the given key.</summary>
+        /// <param name="key">Item's key that will be removed from the cache.</param>
+        public void Remove(TKey key)
+        {
+            if (this.cache.TryRemove(key, out TValue unused))
+            {
+                lock (this.mutex)
+                {
+                    this.keys.Remove(key);
```
Aprogiena: boo again ;)

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    public class MemoryCache<TKey, TValue>
```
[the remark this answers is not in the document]
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        /// <summary>Dictionary that contains cached items.</summary>
+        private readonly ConcurrentDictionary<TKey, TValue> cache;
+
+        /// <summary>Keys sorted by time they when they were accessed last time with most recent ones in the end.</summary>
+        private readonly LinkedList<TKey> keys;
+
+        /// <summary>Maximum items count that can be stored in the cache.</summary>
+        private readonly int sizeLimit;
```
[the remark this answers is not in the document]
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed by removing the Compact()

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
```diff
@@ -133,22 +94,13 @@ public void AddToCache(Block block)
             uint256 blockid = block.GetHash();
             this.logger.LogTrace("({0}:'{1}')", nameof(block), blockid);
 
-            if (!this.cache.TryGetValue(blockid, out Block existingBlock))
-                this.cache.Set(blockid, block, this.blockEntryOptions);
-
-            this.logger.LogTrace("(-)[{0}]", existingBlock != null ? "ALREADY_IN_CACHE" : "ADDED_TO_CACHE");
+            this.cache.AddOrUpdate(blockid, block);
```
Aprogiena: don't remove exit log

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    /// <summary>
+    /// Memory cache that implements the Least Recently Used (LRU) standard.
```
Aprogiena: lru is not a standard - https://en.wikipedia.org/wiki/Cache_replacement_policies

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>Dictionary that contains cached items.</summary>
+        private readonly Dictionary<TKey, LinkedListNode<CacheItem<TKey, TValue>>> cache;
+
+        /// <summary>Keys sorted by their last access time with most recent ones at the end.</summary>
+        private readonly LinkedList<CacheItem<TKey, TValue>> keys;
+
+        /// <summary>Maximum items count that can be stored in the cache.</summary>
+        private readonly int maxItemsCount;
+        
+        /// <summary>Lock to protect access to <see cref="keys"/> and <see cref="cache"/>.</summary>
```
Aprogiena: mention these contracts in remark comments of dictionary and linked list as well

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+                if (this.cache.TryGetValue(key, out node))
+                {
+                    node.Value.Value = value;
+                    this.keys.Remove(node);
+                }
+                else
+                {
+                    if (this.keys.Count == this.maxItemsCount)
+                    {
+                        // Remove 1 item.
```
Aprogiena: better comment would be `Remove the item that was not used for the longest time.`

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+                        this.cache.Remove(this.keys.First.Value.Key);
```
Aprogiena: this is hard to read and reason about,
how about extracting it to a variable at least partially

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                lock (this.mutex)
+                {
+                    return this.keys.Count;
+                }
+            }
+        }
+
+        /// <summary>Create or overwrite an entry in the cache.</summary>
+        /// <param name="key">The key.</param>
```
Aprogiena: `of the item` maybe
and you are still calling it items and entries in different comments, use items everywhere

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+                    node = new LinkedListNode<CacheItem<TKey, TValue>>(new CacheItem<TKey, TValue>(key, value));
+                    this.cache.Add(key, node);
+                }
+
+                this.keys.AddLast(node);
+            }
+        }
+
+        /// <summary>Removes the object associated with the given key.</summary>
+        /// <param name="key">Item's key that will be removed from the cache.</param>
```
Aprogiena: `Key of that item that` would be better to express that it is really the item, not just the hey that is being removed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                if (this.cache.TryGetValue(key, out LinkedListNode<CacheItem<TKey, TValue>> node))
+                {
+                    this.cache.Remove(node.Value.Key);
+                    this.keys.Remove(node);
+                }
+            }
+        }
+
+        /// <summary>Gets the item associated with this key if present.</summary>
```
Aprogiena: `an item` and `a specific key`

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="key">Item's key.</param>
+        /// <param name="value">Item assosiated with specified <paramref name="key"/>.</param>
```
Aprogiena: typo

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <returns><c>true</c> if cache contains the item; <c>false</c> otherwise.</returns>
```
Aprogiena: `;` -> `,` just for consistency with other code

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                    return true;
+                }
+            }
+
+            value = default(TValue);
+            return false;
+        }
+        
+        private class CacheItem<TKey, TValue>
```
Aprogiena: this should be the very first thing in the class, not the last
and pls document everything here

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed
Replaced word 'standard' with 'policy'

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

noescape00: Didn't comment the key,value properties because it's obvious what they represent and couldn't come up with anything better than "The key" and "The value"

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
```diff
@@ -133,22 +94,14 @@ public void AddToCache(Block block)
             uint256 blockid = block.GetHash();
             this.logger.LogTrace("({0}:'{1}')", nameof(block), blockid);
 
-            if (!this.cache.TryGetValue(blockid, out Block existingBlock))
-                this.cache.Set(blockid, block, this.blockEntryOptions);
-
-            this.logger.LogTrace("(-)[{0}]", existingBlock != null ? "ALREADY_IN_CACHE" : "ADDED_TO_CACHE");
+            this.logger.LogTrace("(-)");
```
fassadlr: I believe `this.logger.LogTrace("(-)");` should be the very last line in the method ;)

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -0,0 +1,141 @@
...
+{
+    /// <summary>
+    /// Memory cache that implements the Least Recently Used (LRU) policy.
+    /// </summary>
+    public class MemoryCache<TKey, TValue>
+    {
+        /// <summary>Cache item for the inner usage of the <see cref="MemoryCache{TKey,TValue}"/> class.</summary>
+        private class CacheItem<TKey, TValue>
+        {
+            public TKey Key { get; }
```
fassadlr: Can we not make this `Key` read only?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
Aprogiena: +1

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
Aprogiena: that should be ok

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
noescape00: Fixed

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

noescape00:
>oh yes, but not on Value property, it should be property as per gentleman's agreement

Kk, reintroduced that.

PR #1064, opened 2018-01-15
Block store signaled improvements

noescape00:
This PR removes the 1 sec delay between getting the newest block (from network or mined) and announcing it to the peers while keeping the batch headers for other cases to ensure low overhead.

Fix: #930


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -39,7 +34,19 @@ public class BlockStoreSignaled : SignalObserver<Block>
         private readonly IBlockStoreCache blockStoreCache;
 
         /// <summary>Queue of chained blocks that will be announced to the peers.</summary>
-        private readonly ConcurrentQueue<ChainedBlock> blocksToAnnounce;
+        private AsyncQueue<ChainedBlock> blocksToAnnounce;
+
+        /// <summary>Local batch of blocks that will be announced to the peers.</summary>
+        private List<ChainedBlock> localBatch;
```
Aprogiena: this and the queue readonly?

noescape00: They are created in Initialize().
I can move creation to the constructor and make them readonly but that doesn't sound very logical to me- they belong to Initialize() method.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +110,56 @@ protected override void OnNextCore(Block block)
...
+                using (await this.asyncLock.LockAsync(cancellation).ConfigureAwait(false))
+                {
+                    this.localBatch.Add(item);
+
+                    // Send batch right away if tip, if not and the timer haven't been started already- start the timer.
+                    if (item == this.chain.Tip)
+                    {
+                        this.batchTimer.Stop();
+
+                        await SendBatchAsync();
```
Aprogiena: all awaits ... configure await false ...

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +110,56 @@ protected override void OnNextCore(Block block)
...
 
+        /// <summary>Initializes the <see cref="BlockStoreSignaled"/>.</summary>
+        public void Initialize()
+        {
+            this.localBatch = new List<ChainedBlock>();
+
+            // Configure batch timer.
+            this.batchTimer = new Timer(this.batchInterval.TotalMilliseconds);
+            this.batchTimer.AutoReset = false;
+            this.batchTimer.Elapsed += async (sender, args) =>
```
Aprogiena: can we have a non-anonymous method for these guys?
i'd also create the timer inside constructor and only enabled it here

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +110,56 @@ protected override void OnNextCore(Block block)
...
+            {
+                using (await this.asyncLock.LockAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false))
+                {
+                    if (!this.batchTimer.Enabled)
+                        return;
+
+                    await SendBatchAsync();
```
Aprogiena: if this method needs lock to be called, it should have Locked suffix and that contract should be documented

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +110,56 @@ protected override void OnNextCore(Block block)
...
         /// <remarks>
         /// <para>
-        /// The queue <see cref="blocksToAnnounce"/> contains
-        /// hashes of blocks that were validated by the consensus rules.
+        /// The list <see cref="localBatch"/> contains hashes of blocks that were validated by the consensus rules.
         /// </para>
         /// <para>
         /// This block hashes need to be relayed to connected peers. A peer that does not have a block
-        /// will then ask for the entire block, that means only blocks that have been stored should be relayed.
+        /// will then ask for the entire block, that means only blocks that have been stored\cached should be relayed.
```
Aprogiena: maybe use / instead of \ just to prevent problems

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -127,82 +172,71 @@ protected override void OnNextCore(Block block)
         /// </para>
         /// TODO: consider moving the relay logic to the <see cref="LoopSteps.ProcessPendingStorageStep"/>.
```
Aprogiena: even if just todo, use para

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -48,22 +55,27 @@ public class BlockStoreSignaled : SignalObserver<Block>
...
             this.blockStoreCache = blockStoreCache;
+            this.asyncLock = new AsyncLock();
+
+            // Set interval between batches.
+            this.batchInterval = TimeSpans.Second;
+
+            this.localBatch = new List<ChainedBlock>();
+
+            // Configure batch timer.
+            this.batchTimer = new Timer(this.batchInterval.TotalMilliseconds);
```
dangershony: AsyncLoop is sort of a time, why not use the async loop is there a beneift of a Timer?
Under the hoods the async loops calls Task.Delay which triggers a timer internally when we use async loop we have more control of the code (exception, logging etc..)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +116,54 @@ protected override void OnNextCore(Block block)
...
+            this.blocksToAnnounce = new AsyncQueue<ChainedBlock>(async (item, cancellation) =>
+            {
+                using (await this.asyncLock.LockAsync(cancellation).ConfigureAwait(false))
+                {
+                    this.localBatch.Add(item);
+
+                    // Send batch right away if tip, if not and the timer haven't been started already- start the timer.
+                    if (item == this.chain.Tip)
+                    {
+                        this.batchTimer.Stop();
```
dangershony: Ah the timer is so that you can stop it from pinging, hmm..
This could easily be done with a flag.

Actually why do we need this?
Can we not just have an async loop that broadcast to peers if there are items in `this.localBatch` and if the condition `item == this.chain.Tip` we just not add the block to `this.localBatch.Add(item);` and broadcast it immediately?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: I think because we need to stop it sometimes we will cause more resources consumption by recreating cancellation tokens and creating new tasks for the timers. Btw Task.Delay uses Timer internally.
Second reason is because from the logic perspective it's not a loop if we do only 1 iteration so I didn't want to add one more misuse of async loop here. 

But we're talking about really small resources consumption differences here so I'm fine with changing timer to async loop if you'd like me to.

dangershony: Indeed for consistency I would use the AsyncLoop.
Unless we decide we want to move all our background loops to timers (which is also a possibility).

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: >we just not add the block to this.localBatch.Add(item); and broadcast it immediately?

we can't broadcast a single item if there are some items in the batch. peer will just discard the received block if it's based on a prev block that peer don't have. 

>Can we not just have an async loop that broadcast to peers 

You mean having an async loop with really small interval? That would be active waiting and cause huge resource consumption.

dangershony: >we can't broadcast a single item if there are some items in the batch

Ah yes correct then if there are items in the batch then just add to the batch and don't broadcast.

>You mean having an async loop with really small interval

The interval matters much less if we are not at tip right?
And we solved the tip problem already.
Agree though its a bigger resource consumption but its still very minor.

dangershony: I do however see the benefits of a stoppable timer. hmm.. 🤔

noescape00: >The interval matters much less if we are not at tip right?

Sure. But we really don't want to miss the moment (or delay it) when we finally hit the tip. And after we're synced we need to maintain low delays for every new block.

That was the original reason of making this PR- instead of just reducing the interval from 1 sec to smth smaller we are implementing the callbacks so we take action as soon as we have the block. 
So I'd really avoid using async loop to announce blocks. 
But async loop that will run once as a well logged solution for a timer is fine to me (even though it costs slightly more)

dangershony: >And after we're synced we need to maintain low delays for every new block

This you will anyway achieve.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +116,54 @@ protected override void OnNextCore(Block block)
...
+
+        /// <summary><see cref="batchTimer"/>'s elapsed callback.</summary>
+        private async Task OnBatchTimerRunsOutAsync()
+        {
+            using (await this.asyncLock.LockAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false))
+            {
+                if (!this.batchTimer.Enabled)
+                    return;
+
+                await SendBatchLockedAsync().ConfigureAwait(false);
```
dangershony: `this.SendBatchLockedAsync`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: using async loop here does not make sense, please dont do it

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: ok, I'll just wait for you and Dan to reach consensus on this one. I don't really mind going with any of those 2 approaches but generally I don't like the idea of using smth called loop for a 1 iteration job.

Aprogiena: using async loop here would be like using it instead of Task.Delay everywhere in the code and saying that async loop does internally use Task.Delay, so it is better

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -125,84 +173,75 @@ protected override void OnNextCore(Block block)
         /// <para>
         /// Before relaying, verify the block is still in the best chain else discard it.
         /// </para>
+        /// <para>
         /// TODO: consider moving the relay logic to the <see cref="LoopSteps.ProcessPendingStorageStep"/>.
+        /// </para>
         /// </remarks>
-        public void RelayWorker()
+        private async Task SendBatchLockedAsync()
```
dangershony: What about cancellation? when node shouts down do we know how to wait for this method to finish?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -125,84 +173,75 @@ protected override void OnNextCore(Block block)
...
 
-        /// <summary>
-        /// The async loop needs to complete its work before we can shut down this feature.
-        /// </summary>
-        internal void ShutDown()
+        /// <inheritdoc />
+        protected override void Dispose(bool disposing)
         {
-            this.asyncLoop.Dispose();
+            this.blocksToAnnounce.Dispose();
```
dangershony: We need to find a way to wait for the timer invoked method.

noescape00: we can do the disposing inside the async lock

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -509,7 +510,21 @@ public bool AddToStratisMempool(Transaction trx)
...
         {
-            return this.FullNode.Services.ServiceProvider.GetService<IPowMining>().GenerateBlocks(new ReserveScript { ReserveFullNodeScript = this.MinerSecret.ScriptPubKey }, (ulong)blockCount, uint.MaxValue);
+            List<uint256> blocks = new List<uint256>();
+
+            ConsensusLoop consensusLoop = this.FullNode.NodeService<ConsensusLoop>();
+            ConcurrentChain chain = this.FullNode.NodeService<ConcurrentChain>();
+
+            for (int i = 0; i < blockCount; i++)
+            {
+                // Wait here with a small interval. Otherwise PoSMiner will wait for this with 1 minute interval.
```
dangershony: The miner will wait for 1 min?

noescape00: Oh, it's PoW, not PoS. 

PowMining:124

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: Hmmm interesting idea, but then you might not be the next in line to get the lock.
Ah and the AsyncLock will throw anyway as it uses cancellation if we wait for the async lock with a None cancellation that might work as any tasks that are ahead of the dispose method will just throw.
Not sure that will work.

noescape00: You're right, just checked that. 
So I guess I'll just add a local task that we can await on dispose.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
noescape00: fixed the comment (PoS => PoW).

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +118,56 @@ protected override void OnNextCore(Block block)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>Initializes the <see cref="BlockStoreSignaled"/>.</summary>
+        public void Initialize()
+        {
+            this.batchTimer.Elapsed += async (sender, args) => { await this.OnBatchTimerRunsOutAsync().ConfigureAwait(false); };
```
dangershony: Ah one more question I am not sure about will the timer elapse even that the previous call did not complete?
Say timer is 1 sec and a call is taking 1.1 sec will the next call queue a task or will the timer wait time when the previous call completed.

Aprogiena:
i still want to check this, i hope to do it tomorrow
if there are any bigger concerns up until i do it, just highlight me pls

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00:
```
            // Init timer with 5ms interval that call task that does the Delay with 1 sec.
            System.Timers.Timer timer = new System.Timers.Timer(5) { AutoReset = false }; ;
            timer.Elapsed += async (sender, eventArgs) =>
            {
                await Task.Delay(1000);

                Console.WriteLine("CALL");
            };

            // Start timer 4 times with 100ms interval.
            for (int i = 0; i < 4; ++i)
            {
                timer.Start();
                Thread.Sleep(100);
            }

            Thread.Sleep(10000);
            Console.ReadLine();
```

This produces this console output:

```
CALL
CALL
CALL
CALL
```

So it looks like timer will still invoke the callback even if the previous one is not completed.

dangershony: Yes that's the difference with using a timer we don't wait completion.
Tasks will be waiting at the async lock indeed, I guess I am fine with that.
Would be nice to have a Timer that wait completion (perhaps we can modify the AsyncLoop to have an on/off button but it might be tricky)

Aprogiena: i haven't checked the code yet, but the discussion about the timer here seems unrelated to what we do (or should do) in this code, well i'll check it after lunch and see myself

noescape00: >Yes that's the difference with using a timer we don't wait completion.

In this particular case it's not that bad and I can't imagine a scenario when we will have 3 callbacks running at the same time. Actually there is just 1 occasion when we can get 2 of them queued: when we're are not synced and it's the 1st time we get a tip block. So just once per node run we can end up in scenario when we can have 2 callbacks invoked at the same time.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -104,17 +118,56 @@ protected override void OnNextCore(Block block)
...
+        /// A method that relays blocks found in <see cref="localBatch"/> to connected peers on the network.
         /// </summary>
         /// <remarks>
         /// <para>
-        /// The queue <see cref="blocksToAnnounce"/> contains
-        /// hashes of blocks that were validated by the consensus rules.
+        /// The list <see cref="localBatch"/> contains hashes of blocks that were validated by the consensus rules.
         /// </para>
         /// <para>
         /// This block hashes need to be relayed to connected peers. A peer that does not have a block
```
Aprogiena: This -> These

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
dangershony: Fair enough as we will be disabling the timer once we reached the tip.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -48,22 +58,26 @@ public class BlockStoreSignaled : SignalObserver<Block>
...
             this.connection = connection;
-            this.name = name;
             this.nodeLifetime = nodeLifetime;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.storeSettings = storeSettings;
             this.blockStoreCache = blockStoreCache;
+            this.asyncLock = new AsyncLock();
+
+            // Set interval between batches.
+            this.batchInterval = TimeSpans.Second;
```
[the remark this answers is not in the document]
noescape00: >I would use five seconds here

kk

>I would also make a constant of that

Can't make a constant out of TimeSpan

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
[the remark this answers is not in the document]
Aprogiena: don't use timespan then ;)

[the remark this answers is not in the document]
noescape00: Fixed

noescape00:
Can this one be merged or there is something else to it?

noescape00:
>i'm still working on this one, be patient please

Ok, no problem. Didn't know that =)

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -52,8 +52,8 @@ public MemoryCache(int maxItemsCount, IEqualityComparer<TKey> comparer = null)
 
             this.maxItemsCount = maxItemsCount;
 
-            this.cache = new Dictionary<TKey, LinkedListNode<CacheItem<TKey, TValue>>>(this.maxItemsCount, comparer);
-            this.keys = new LinkedList<CacheItem<TKey, TValue>>();
+            this.cache = new Dictionary<TKey, LinkedListNode<CacheItem>>(this.maxItemsCount, comparer);
+            this.keys = new LinkedList<CacheItem>();
```
dangershony: Don't use the name `mutex` its confusing, just a `lockObject`  is enough.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -105,16 +104,67 @@ protected override void OnNextCore(Block block)
...
-        /// A loop method that continuously relays blocks found in <see cref="blocksToAnnounce"/> to connected peers on the network.
+        /// Continuously dequeues items from <see cref="blocksToAnnounce"/> and sends
+        /// them  to the peers after the timer runs out or if the last item is a tip.
+        /// </summary>
+        private async Task DequeueContinuouslyAsync()
+        {
+            var batch = new List<ChainedBlock>();
+
+            Task<ChainedBlock> dequeueTask = null;
+            Task timerTask = null;
```
dangershony: Too many tasks flying around its starting to make me feel uncomfortable.

noescape00: Less then 2 at a time 🙂

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
Aprogiena: @dangershony actually i was first opposed to this solution as well, but we worked on that and it allowed us to have a lock free implementation, which is short enough that one can argue about it and check all its states, so this is actually pretty good result and i'm happy @noescape00 was persistent enough here to convince me we can do reasonably well with all those tasks around

PR #1065, opened 2018-01-15
Refactor/the node job part 20

Aprogiena:
Ole
- remove some unused methods from NetworkPeer
- refactor NetworkPeerListener (async and without fuzzy behavior) - see the particular commit for more info
- separation of IMessageListener interface from classes that implement that
- minor async fixes (configure awaits, cancellation tokens)


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
```diff
@@ -1,78 +1,72 @@
...
 using Stratis.Bitcoin.P2P.Protocol.Payloads;
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.P2P.Peer
 {
-    public class NetworkPeerListener : PollMessageListener<IncomingMessage>, IDisposable
+    /// <summary>
+    /// Message listener that waits until a specific payload is received and returns it to the caller.
+    /// </summary>
+    /// <typeparam name="T">Type of the messages that are being handled.</typeparam>
```
noescape00: There is no param with name T. Did you mean `IncomingMessage`?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
```diff
@@ -1,78 +1,72 @@
...
-                        if (this.predicates.All(p => p(message)))
-                        {
-                            if (message.Message.Payload is TPayload)
-                                return (TPayload)message.Message.Payload;
-
-                            pushedAside.Enqueue(message);
-                        }
-                    }
+                    IncomingMessage message = await this.asyncQueue.DequeueAsync(cancellation.Token).ConfigureAwait(false);
+                    if (message.Message.Payload is TPayload)
```
noescape00: You can do 

```
if (message.Message.Payload is TPayload payload)
       return payload;
```

to avoid 2nd cast.

src/Stratis.Bitcoin/P2P/Protocol/IMessageListener.cs
```diff
@@ -0,0 +1,13 @@
+﻿/// <summary>
+/// Contract for a recipient part of a consumer logic that handles incoming messages.
+/// </summary>
+/// <typeparam name="T">Type of the messages that are being handled.</typeparam>
+/// <seealso cref="MessageProducer{T}"/>
```
noescape00: You need to add `using Stratis.Bitcoin.P2P.Protocol;` OR use 
```
cref="Stratis.Bitcoin.P2P.Protocol.MessageProducer{T}"
```

so this reference is handled properly. Otherwise resharper creates a "missing reference" warning because it's not resolved.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
Aprogiena: Thank you will fix

src/Stratis.Bitcoin/P2P/Protocol/IMessageListener.cs
Aprogiena: Thank you good find will fix

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
Aprogiena: Thank you will fix

noescape00:
This looks good to me, gj! 

ps
Ran the node and integration tests- all are passing.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
```diff
@@ -1,78 +1,71 @@
...
+        /// <summary>Subscription to the message producer of the connected peer.</summary>
+        private readonly IDisposable subscription;
+
+        /// <summary>
+        /// Initializes the instance of the object and subscribes to the peer's message producer.
+        /// </summary>
+        /// <param name="peer">Connected network peer that we receive messages from.</param>
         public NetworkPeerListener(NetworkPeer peer)
         {
+            this.asyncQueue = new AsyncQueue<IncomingMessage>();
```
dangershony: I am considering we should rename `AsyncQueue` to `ConcurrentAsyncQueue`

dangershony:
All integration tests passed for @noescape00 then I will commit.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerListener.cs
Aprogiena: you might be right, but there are stronger requirements for concurrent collections and i haven't checked if this implementation satisfies them

PR #1066, opened 2018-01-16
Fix issue with SegWit not activating on Bitcoin RegTest and fix incorrect constant

### LINKED ISSUE #1027: [Consensus] Incorrect constant being used for Stratis PoW block reward
In [PosConsensusValidator.cs](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Consensus/PosConsensusValidator.cs#L503-L509) the following code exists:

```
        public override Money GetProofOfWorkReward(int height)
        {
            if (this.IsPremine(height))
                return this.consensusOptions.PremineReward;

            return this.ConsensusOptions.ProofOfWorkReward;
        }
```

The constants are incorrectly being derived from two different places (consensusOptions vs ConsensusOptions). As a result the code currently considers a block reward of 50 to be correct, which is clearly wrong for the Stratis network.

This error presumably does not cause sync problems because there are no blocks on testnet/mainnet that have a block reward higher than the correct value of 4 (disregarding fees). It is not considered invalid to have a coinbase reward lower than the maximum defined by consensus.

The fix is simple, just change ConsensusOptions to consensusOptions. And introduce consensus tests to catch errors like this going forwards.

### LINKED ISSUE #1028: [Consensus] Segwit does not activate on BTC regtest
In [Networks.cs](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/Networks.cs#L232-L234) the following constants are defined for segwit activation and timeout via BIP9:

```
            network.consensus.BIP9Deployments[BIP9Deployments.TestDummy] = new BIP9DeploymentsParameters(28, 0, 999999999);
            network.consensus.BIP9Deployments[BIP9Deployments.CSV] = new BIP9DeploymentsParameters(0, 0, 999999999);
            network.consensus.BIP9Deployments[BIP9Deployments.Segwit] = new BIP9DeploymentsParameters(1, 0, 999999999);
```

Unlike Main and TestNet, these constants denote a day in the year 2001. This means that segwit never activates on RegTest. More precisely, the BIP9 threshold state never transitions to Started.

It is probable that a similar error exists with the timeout constant for CSV activation. Both should be fixed.

There is a further error in [ThresholdConditionCache.cs](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin/Base/Deployments/ThresholdConditionCache.cs#L60) in the following code:

```
            if (pindexPrev != null)
            {
                pindexPrev = pindexPrev.GetAncestor(pindexPrev.Height - ((pindexPrev.Height + 1) % nPeriod));
            }
```

It is unclear whether there is a logical error in this code when the block height is very close to genesis, or if the introduction of the skip list affects the output of GetAncestor. Either way, this line does not appear to do what it is supposed to on regtest and the BIP9 state does not change from Defined to Started. Eventually the state changes to Failed and segwit never activates.

The expected behaviour (based on trials with a Core bitcoind) is that segwit should lock in at about block height 288, then be active by about block 432 on regtest.

There is a third apparent error. When the `pindexPrev = pindexPrev.Get...` line is commented out in the previous code sample, the BIP9 state does seem to correctly transition to Started. However, at consensus tip height 395 a BadWitnessNonceSize exception gets thrown in [WitnessCommitmentsRule.cs](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/WitnessCommitmentsRule.cs#L40-L44). It has currently not been established why this is the case.

```
                    if ((witness.PushCount != 1) || (witness.Pushes.First().Length != 32))
                    {
                        this.Logger.LogTrace("(-)[BAD_WITNESS_NONCE_SIZE]");
                        ConsensusErrors.BadWitnessNonceSize.Throw();
                    }
```

mikedennis:
Fix #1028 
Fix #1027

Also added SegWit activation parameters for Bitcoin Mainnet as per bitcoin core.

### DIFF
--- a/src/NBitcoin/BIP9Deployments.cs
+++ b/src/NBitcoin/BIP9Deployments.cs
@@ -11,6 +11,9 @@ namespace NBitcoin
 
     public class BIP9DeploymentsParameters
     {
+        /// <summary>Special flag for timeout to indicate always active.</summary>
+        public const long AlwaysActive = -1;
+
         public BIP9DeploymentsParameters(int bit, DateTimeOffset startTime, DateTimeOffset timeout)
         {
             Bit = bit;
--- a/src/Stratis.Bitcoin.IntegrationTests/SegWitTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/SegWitTests.cs
@@ -1,0 +1,74 @@
+﻿using NBitcoin;
+using Stratis.Bitcoin.Features.BlockStore;
+using Stratis.Bitcoin.Features.Consensus;
+using Stratis.Bitcoin.Features.MemoryPool;
+using Stratis.Bitcoin.Features.Notifications;
+using Stratis.Bitcoin.Features.Miner;
+using Stratis.Bitcoin.Features.Wallet;
+using Stratis.Bitcoin.Features.WatchOnlyWallet;
+using Stratis.Bitcoin.Features.Api;
+using Stratis.Bitcoin.Features.RPC;
+using Stratis.Bitcoin.Features.Wallet.Interfaces;
+using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
+using Xunit;
+using Stratis.Bitcoin.Base.Deployments;
+using NBitcoin.RPC;
+using static NBitcoin.Consensus;
+
+namespace Stratis.Bitcoin.IntegrationTests
+{
+    public class SegWitTests
+    {
+        [Fact]
+        public void TestSegwitActivation()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create(version: "0.15.1"))
+            {
+                CoreNode coreNode = builder.CreateNode(false);
+
+                coreNode.ConfigParameters.AddOrReplace("debug", "1");
+                coreNode.ConfigParameters.AddOrReplace("printtoconsole", "0");
+                coreNode.Start();
+
+                CoreNode node1 = builder.CreateStratisPowNode(true, fullNodeBuilder =>
+                {
+                    fullNodeBuilder
+                        .UseConsensus()
+                        .UseBlockStore()
+                        .UseMempool()
+                        .UseBlockNotification()
+                        .UseTransactionNotification()
+                        .AddMining()
+                        .UseWallet()
+                        .UseWatchOnlyWallet()
+                        .UseApi()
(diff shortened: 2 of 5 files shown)


src/NBitcoin/BIP9Deployments.cs
```diff
@@ -11,6 +11,9 @@ public enum BIP9Deployments : int
 
     public class BIP9DeploymentsParameters
     {
+        /// <summary>Special flag for timeout to indicate always active.</summary>
+        public const long AlwaysActive = -1;
```
dangershony: @mikedennis if you have some spare time to check and apply this to NBitcoin I am sure Nicolas would appreciate it.

src/Stratis.Bitcoin.IntegrationTests/SegWitTests.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    fullNodeBuilder
+                        .UseConsensus()
+                        .UseBlockStore()
+                        .UseMempool()
+                        .UseBlockNotification()
+                        .UseTransactionNotification()
+                        .AddMining()
+                        .UseWallet()
+                        .UseWatchOnlyWallet()
```
dangershony: Are you sure you need to use `UseWatchOnlyWallet` here?

src/Stratis.Bitcoin/Base/Deployments/ThresholdConditionCache.cs
```diff
@@ -54,6 +54,12 @@ public ThresholdState GetState(ChainedBlock pindexPrev, BIP9Deployments deployme
             var nTimeStart = this.consensus.BIP9Deployments[deployment]?.StartTime;
             var nTimeTimeout = this.consensus.BIP9Deployments[deployment]?.Timeout;
 
+            // Check if this deployment is always active.
+            if (nTimeStart == Utils.UnixTimeToDateTime(BIP9DeploymentsParameters.AlwaysActive))
```
dangershony: If you are here rename to `timeStart`

src/NBitcoin/BIP9Deployments.cs
mikedennis: 👍

PR #1067, opened 2018-01-16
[PeerSelector] Implement a seperate method to return peers for GetAddrPayload

### LINKED ISSUE #1061: [PeerSelector] GetAddrPayload should return more handshaked/connected peers during discovery.
Currently we only return a random set of peers when a node asks for addresses during peer discovery.

We should try and return a list which is comprised of 50% handshaked/connected peers and 50% attempted/fresh peers.

fassadlr:
Fix #1061


src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -82,25 +88,27 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
 
             // First check to see if there are handshaked peers. If so,
             // give them a 50% chance to be picked over all the other peers.
-            if (this.peerAddresses.Handshaked().Any())
+            var handshaked = this.peerAddresses.Handshaked();
+            if (handshaked.Any())
             {
                 int chance = this.random.Next(100);
                 if (chance <= 50)
```
345paul: I admit, I don't quite understand the 50% chance logic.  Isn't the preferred order simply handshaked, connected, fresh, attempted, as in, if you have handshaked peers then are they not preferable to anything else?  I remember us talking about handshaked vs lastseen - have they been conflated now to a single timetamp?

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -145,18 +153,69 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
...
+            // Randomly order the list of peers and return the amount asked for.
             var allPeers = this.peerAddresses.OrderBy(p => this.random.Next());
             return allPeers.Select(p => p.Value).Take(1000);
         }
+
+        /// <inheritdoc/>
+        public IEnumerable<PeerAddress> SelectPeersForGetAddrPayload(int peerCount)
+        {
+            // If there are no peers or just one then just return the list.
+            if (!this.peerAddresses.Any() || this.peerAddresses.Count == 1)
```
345paul: Would it be clearer to have these as two separate conditions?  If no peers, return Enumerable.Empty<PeerAddress>().  If 1 peer, return this.peerAddresses.First().Value.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -145,18 +153,69 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
...
+                return this.peerAddresses.Select(pa => pa.Value);
+
+            var peersToReturn = new List<PeerAddress>();
+
+            var connectedAndHandshaked = this.peerAddresses.Connected().Concat(this.peerAddresses.Handshaked()).OrderBy(p => this.random.Next());
+            var freshAndAttempted = this.peerAddresses.Attempted().Concat(this.peerAddresses.Fresh()).OrderBy(p => this.random.Next());
+
+            //If there are connected and/or handshaked peers in the address list,
+            //we need to split the list 50 / 50 between them and
+            //peers we have not yet connected to and/or that are fresh.
```
345paul: Again, as per comment before, I'm not understanding why you want to return peers that we haven't already at least gone through the handshake process (and then subsequently kept 'live' by pinging).  Up until that point, all other peers are purely speculative as to whether they are any good or not.  Would you want to return a list of peers to another node containing ones that you haven't performed 'due diligence' on yourself?  Apologies if I've missed the reason for this as I see this logic has already existed before and maybe there was a good reason for it that I can't immediately fathom.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: Hi @345paul! Ok so this method is used only when we want to connect to a single peer during normal node operation. This is not used during PeerDiscovery or GetAddrPayload. 

The idea here IS to prefer peers in the order you specified, namely, handshaked, connected, fresh and attempted. I.e first check to see if there are handshaked peers, if so, give them a 50% chance to be picked. If this fails, check if there are "connected only" peers, if so, give them a 50% chance to be picked. If all else fails pick an attempted or fresh peer randomly

The idea here is that we cant JUST connect to "good" peers. We also have to give fresh or previously attempted peers another chance. I.e. fresh peers that are potentially "good" must also be given a chance to be propagated through the network. Hence, the 50% chance given to handshaked and/or connected peers.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: Agreed 👍

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: Again, we can't JUST propagate good peers throughout the network. New peers, that are potentially good ones, also needs to be given a chance to get connected to. In this method, which is only used for GetAddrPayload, we'll "fill up" the list to be provided, first with handshaked/connected and then with attempted/fresh peers. This ensures that there's always an even (fair) distribution of peers on the network.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: @345paul the "LastSeen" rename of "LastConnectedHandshake" and its subsequently "live" updating will come after this PR is merged ;)

src/Stratis.Bitcoin/P2P/PeerSelector.cs
345paul: I would argue that you only really want to propagate good peers :)  For peers that you know about (recently connected to you or ones you received from another peer and for which you haven't yet established whether they are good or not), you have a duty to determine how 'good' they are, which could be done in the background, independent of propagating the good peers when required to do so.  I guess propagating transactions to peers should also only be done using good peers.  However, having said that, I understand your reasons for the logic of mixing the quality of the peers in the response payload.

dangershony: My question would be what is core doing?
We probably want to do the same I think this cod is responsible for address selection and logic
https://github.com/bitcoin/bitcoin/blob/5961b23898ee7c0af2626c46d5d70e80136578d3/src/net.cpp

dangershony: This is the getaddr logic 
https://github.com/bitcoin/bitcoin/blob/5961b23898ee7c0af2626c46d5d70e80136578d3/src/net_processing.cpp

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -145,18 +153,72 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
         }
 
         /// <inheritdoc/>
-        public IEnumerable<PeerAddress> SelectPeers(int peerCount)
+        public IEnumerable<PeerAddress> SelectPeersForDiscovery(int peerCount)
         {
-            // If there are no peers, just one or if the amount peers is less than
-            // the amount of peers asked for, just return the list.
+            // If there are no peers or just one then just return the list.
```
dangershony: Same comment as bellow here I would say, but not that important
```
  // If there are no peers or just one then just return the list.
            if (!this.peerAddresses.Any())
                return Enumerable.Empty<PeerAddress>();

            if (this.peerAddresses.Count == 1)
                return this.peerAddresses.Select(pa => pa.Value);
```

PR #1068, opened 2018-01-16
Refactor/the node job part 21

Aprogiena:
First commit is called `Part 21 starts here`. As you can see, this is rather smallish PR again.

- Simplify NetworkPeerCollection (removes unused Bridge)
- replaces generic IDisposable listener subscription with dedicated registration class
- refactors MessageProducer (simplification, add comments, remove unused stuff, remove dependency on Scope class, use concurrent hashset instead of lock)


src/Stratis.Bitcoin/P2P/Protocol/MessageProducer.cs
```diff
@@ -1,75 +1,93 @@
...
 
-        public void RemoveMessageListener(IMessageListener<T> listener)
+    /// <summary>
+    /// Distributor of messages to attached consumers.
+    /// </summary>
+    /// <typeparam name="T">Type of the messages that are being handled.</typeparam>
+    public class MessageProducer<T>
+    {
+        /// <summary>List of attached consumers of this producer's messages.</summary>
+        private readonly ConcurrentHashSet<IMessageListener<T>> listeners;
```
dangershony: You use ConcurrentHashSet for the additional methods that ConcurrentBag does not have (like TryAdd, TryRemove)?

dangershony:
FYI I didn't run the tests locally, if you want me to please hold on with committing.

src/Stratis.Bitcoin/P2P/Protocol/MessageProducer.cs
Aprogiena: ConcurrentBag is very tricky thing - see http://download.microsoft.com/download/B/C/F/BCFD4868-1354-45E3-B71B-B851CD78733D/PerformanceCharacteristicsOfThreadSafeCollection.pdf

it is implemented as a forest of local thread storage trees with cheap operation inside thread's tree, but expensive steal operation (thread obtaining an item that belongs to another thread)

in async environment, we have no guarantees about which thread does what, so i don't thing concurrent idea is a good choice in async environment for anything

Aprogiena: also note 
```
Other Considerations 

ConcurrentBag(T) is a bit heavyweight from the memory prespective, the reason being that ConcurrentBag(T) is not disposable but it internally consists of disposable ThreadLocal(T) objects. These ThreadLocal(T) objects, even when they are not used anymore, cannot be disposed until the ConcurrentBag(T) object is collected by GC. IsEmpty, Count, ToArray() and GetEnumerator() lock the entire data structure so that they can provide a snapshot view of the whole bag.  As such, these operations are inherently expensive and they cause concurrent Add() and Take() operations to block.  Note that by the time ToArray() or GetEnumerator() returns, the global lock is already released and so the original collection may have already been modified. 
 
```

dangershony: Ahhh nice read thanks mate

Aprogiena: From what i read there, it seems you would only used CBag if you have a very high performance app and you really dedicate threads to work with it and then it could very much outperform everything else, but as mentioned, the costs (memory, blocked threads) for that are quite high. So for normal things, it probably does not make sense.

dangershony: I will have a read but this suggest we should favour ConcurrentHashSet (or more correct is avoid ConcurrentBag)?

Aprogiena: I don't think we intend to create a high performance software, so we probably won't ever need to use CBag. Just like we switched from SocketEventArgs to TcpListener. If TcpListener is used in async environment, it provides excellent performance, but reportedly SocketEventArgs is the best performing thing available, but we just don't need it and its complexity is very high. So what we use has way enough performance for our use case and is cheap with other resources (including complexity of the code).

So yes, probably avoid CBag unless you are developing something extraordinary.

Issue #804, opened 2017-11-26
Handling critical startup exceptions. 


noescape00:
As was suggested by @aprogiena there should be a way for handling critical startup exceptions that are not compatible with continuation of node starting process. Probably a new type of exception should be introduced for that and every component's startup rebuilt.

Aprogiena:
Currently, the components are started with Start which returns void, so there is no how for the caller to see if the startup succeeded or not. We could either change the pattern to return bool and if something returned false then we would not continue with the startup. Or we can introduce a new type of exception that would be handled without reporting to the console.

The idea is not to crash to console, but give user a nice message when something bad happens due to misconfiguration or whatever is user's fault.

dangershony:
Is it not like this already?

bokobza:
https://github.com/stratisproject/StratisBitcoinFullNode/pull/1071 fixes the problem of the app getting stuck.
With this change, the app shows the exception as before, but the console stops.
I guess the way to fix this would be to internally catch the exception at the point of failure (like in the DBreeze constructor) and throw a nicer message.

bokobza:
I think we can close this, the resolution is sufficient for now.

Issue #806, opened 2017-11-27
Main net and Test net cannot be run simultaneously


D3m0nKingx:
FN cannot be initialized on both main and testnets simultaneously due to the FN requesting binding to 127.0.0.1:37221 for the swagger api, that's already in use by another instance of the Node running (either main/testnet whichever is started first)

```bash
There was a problem initializing the node. Details: 'One or more errors occurred. (Failed to bind to address http://127.0.0.1:37221: address already in use.)'
Application is shutting down...
The CancellationTokenSource has been disposed.
```

bokobza:
Hmm good one.
I guess we'd be better off defining some new ports for testing.

PR #1071, opened 2018-01-17
Running multiple nodes on the same host

bokobza:
This PR contains a few commits allowing multiple nodes to run on the same machine:

* Make sure the node terminates:
If an exception is thrown (for example the DBreeze fails to initialize), the ManualResetEventSlim is never set and the node stays waiting for it. So if for example a second node is started on a machine, it will fail whilst creating the DBreeze db but the process will keep running. This fix allows the process to terminate.

* Checking for null before disposing of Dbreeze

* Allow setting of custom data directory.
Previously, trying to start a node with a directory different than the default one would throw an exception. Now we create a directory after the one specified in the datadir argument.

* Split of ApiUri
Splitting the apiUri between ApiUri and ApiPort so that it's easier to set up a port without having to worry about setting the host.
For example, a user starting the node could just start the node with -apiport=37555.

* Removed the NOASSEMBLYCONTEXT constant from csproj files. 

Helps the following:
 https://github.com/stratisproject/Breeze/issues/322
https://github.com/stratisproject/Breeze/issues/314
https://github.com/stratisproject/StratisBitcoinFullNode/issues/806
https://github.com/stratisproject/StratisBitcoinFullNode/issues/257

### DIFF
--- a/src/Stratis.Bitcoin.Api.Tests/ApiSettingsTest.cs
+++ b/src/Stratis.Bitcoin.Api.Tests/ApiSettingsTest.cs
@@ -1,0 +1,198 @@
+﻿using System;
+using NBitcoin;
+using Stratis.Bitcoin.Builder;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Features.Api;
+using Stratis.Bitcoin.Tests;
+using Xunit;
+
+namespace Stratis.Bitcoin.Api.Tests
+{
+    /// <summary>
+    /// Tests the settings for the API features.
+    /// The default settings are:
+    /// ApiPort: 37220 for bitcoin, 37221 for Stratis.
+    /// ApiUri: http://localhost:37220 or http://localhost:37221.
+    /// </summary>
+    public class ApiSettingsTest : TestBase
+    {
+        public ApiSettingsTest()
+        {
+            // These flags are being set on an individual test case basis.
+            // Assume the default values for the static flags.
+            Transaction.TimeStamp = false;
+            Block.BlockSignature = false;
+        }
+
+        /// <summary>
+        /// Tests that if no API settings are passed and we're on the bitcoin network, the defaults settings are used.
+        /// </summary>
+        [Fact]
+        public void GivenNoApiSettingsAreProvidedAndOnBitcoinNetworkThenDefaultSettingAreUsed()
+        {
+            // Arrange.
+            Network network = Network.Main;
+            NodeSettings nodeSettings = new NodeSettings("bitcoin", network).LoadArguments(new string[0]);
+
+            // Act.
+            ApiSettings settings = new FullNodeBuilder()
+                .UseNodeSettings(nodeSettings)
+                .UseApi()
+                .Build()
+                .NodeService<ApiSettings>();
+            settings.Load(nodeSettings);
+
+            // Assert.
+            Assert.Equal(ApiSettings.DefaultBitcoinApiPort, settings.ApiPort);
+            Assert.Equal(new Uri($"{ApiSettings.DefaultApiHost}:{ApiSettings.DefaultBitcoinApiPort}"), settings.ApiUri);
+        }
+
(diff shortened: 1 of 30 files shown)


src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -9,9 +9,21 @@ namespace Stratis.Bitcoin.Features.Api
     /// </summary>
     public class ApiSettings
     {
+        /// <summary>The default port used by the API when the node runs on the bitcoin network.</summary>
+        public const int DefaultBitcoinApiPort = 37220;
```
dangershony: I think this should go in the Network class (next to rpc port)

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -174,14 +174,21 @@ public NodeSettings LoadArguments(string[] args)
...
                 this.DataDir = this.CreateDefaultDataDirectories(Path.Combine("StratisNode", this.Name), this.Network);
             }
-
-            if (!Directory.Exists(this.DataDir))
-                throw new ConfigurationException($"Data directory {this.DataDir} does not exist.");
-
+            else
+            {
+                // Create the data directories if they don't exist.
+                string directoryPath = Path.Combine(dataDir, this.Name, this.Network.Name);
```
dangershony: So now the path folder structure is
```
- data-dir (or default)
    - coin-name
        -network
            - main 
            - test
```

Right?

src/Stratis.Bitcoin/Utilities/FullNodeExtensions.cs
```diff
@@ -34,21 +34,29 @@ public static class FullNodeExtensions
...
+
                 Console.CancelKeyPress += (sender, eventArgs) =>
                 {
                     shutdown();
                     // Don't terminate the process immediately, wait for the Main thread to exit gracefully.
                     eventArgs.Cancel = true;
                 };
 
-                await node.RunAsync(cts.Token, "Application started. Press Ctrl+C to shut down.", "Application stopped.").ConfigureAwait(false);
+                try
```
dangershony: If there is a try catch here can we remove the one in program.cs?

dangershony:
I think this also fixes https://github.com/stratisproject/StratisBitcoinFullNode/issues/804

src/Stratis.Bitcoin.Api.Tests/ApiSettingsTest.cs
```diff
@@ -0,0 +1,198 @@
...
+            // Assume the default values for the static flags.
+            Transaction.TimeStamp = false;
+            Block.BlockSignature = false;
+        }
+
+        /// <summary>
+        /// Tests that if no API settings are passed and we're on the bitcoin network, the defaults settings are used.
+        /// </summary>
+        [Fact]
+        public void GivenNoApiSettingsAreProvidedAndOnBitcoinNetworkThenDefaultSettingAreUsed()
```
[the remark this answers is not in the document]
bokobza: will change

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
bokobza: This is only for the API feature though. It does not really have an obvious place in the Network class.

src/Stratis.Bitcoin/Utilities/FullNodeExtensions.cs
dangershony: Ah this is to invoke the `done.Set();`
What about putting a finally here and let the exception in program.cs be the only place to log errors?

```
                 try
                 {
                     await node.RunAsync(cts.Token, "Application started. Press Ctrl+C to shut down.", "Application stopped.").ConfigureAwait(false);
                 }
                 finally
                 {
                     done.Set();
                 }
```

PR #1077, opened 2018-01-19
Removed .NodeService<T> usages from non-test methods

noescape00:
As @dangershony said: the whole design behind .NodeService is wrong: we are supposed to inject services, not ask for them.

So in this PR .NodeService<T> usages for non-test methods are removed.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -209,7 +209,7 @@ namespace Stratis.Bitcoin.Features.Consensus
                         services.AddSingleton<IConsensusLoop, ConsensusLoop>();
                         services.AddSingleton<StakeChainStore>().AddSingleton<StakeChain, StakeChainStore>(provider => provider.GetService<StakeChainStore>());
                         services.AddSingleton<IStakeValidator, StakeValidator>();
-                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>();
+                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>().AddSingleton<IGetUnspentTransaction, ConsensusManager>();
                         services.AddSingleton<IInitialBlockDownloadState, InitialBlockDownloadState>();
                         services.AddSingleton<ConsensusController>();
                         services.AddSingleton<ConsensusStats>();
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -209,7 +209,8 @@ namespace Stratis.Bitcoin.Features.Consensus
                         services.AddSingleton<IConsensusLoop, ConsensusLoop>();
                         services.AddSingleton<StakeChainStore>().AddSingleton<StakeChain, StakeChainStore>(provider => provider.GetService<StakeChainStore>());
                         services.AddSingleton<IStakeValidator, StakeValidator>();
-                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>().AddSingleton<IGetUnspentTransaction, ConsensusManager>();
+                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>()
+                            .AddSingleton<IGetUnspentTransaction, ConsensusManager>();
                         services.AddSingleton<IInitialBlockDownloadState, InitialBlockDownloadState>();
                         services.AddSingleton<ConsensusController>();
                         services.AddSingleton<ConsensusStats>();
--- a/src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
+++ b/src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
@@ -31,12 +31,15 @@ namespace Stratis.Bitcoin.Features.RPC.Controllers
 
         private readonly INetworkDifficulty networkDifficulty;
 
+        private readonly IConsensusLoop consensusLoop;
+
         public FullNodeController(
             ILoggerFactory loggerFactory,
             IPooledTransaction pooledTransaction,
             IPooledGetUnspentTransaction pooledGetUnspentTransaction,
             IGetUnspentTransaction getUnspentTransaction,
             INetworkDifficulty networkDifficulty,
+            IConsensusLoop consensusLoop,
             IFullNode fullNode = null,
             NodeSettings nodeSettings = null,
             Network network = null,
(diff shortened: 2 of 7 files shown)


Neurosploit:
Thank you so much for this noescape! Are there any plans to do something about this.FullNode.Services.ServiceProvider as well or is that going to be allowed?

noescape00:
>Thank you so much for this noescape! 

You are welcome =) 

>Are there any plans to do something about this.FullNode.Services.ServiceProvider as well or is that going to be allowed?

this.FullNode.Services.ServiceProvider is used only in FullNode and places where we actually modify the services except for the tests (and a single usage in FullNodeController that I'll fix in the next commit for this PR- ty for reminding about that) so I guess it's fine.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -209,7 +209,7 @@ public static IFullNodeBuilder UseStratisConsensus(this IFullNodeBuilder fullNod
                         services.AddSingleton<IConsensusLoop, ConsensusLoop>();
                         services.AddSingleton<StakeChainStore>().AddSingleton<StakeChain, StakeChainStore>(provider => provider.GetService<StakeChainStore>());
                         services.AddSingleton<IStakeValidator, StakeValidator>();
-                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>();
+                        services.AddSingleton<ConsensusManager>().AddSingleton<INetworkDifficulty, ConsensusManager>().AddSingleton<IGetUnspentTransaction, ConsensusManager>();
```
mikedennis: Not sure I would put all this on the same line.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -23,8 +23,23 @@ public class FullNodeController : FeatureController
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
+        private readonly IPooledTransaction pooledTransaction;
```
mikedennis: Apro is gonna want comments 😉

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -62,7 +82,7 @@ public Task Stop()
             if (!uint256.TryParse(txid, out trxid))
                 throw new ArgumentException(nameof(txid));
 
-            Transaction trx = await this.FullNode.NodeService<IPooledTransaction>(true)?.GetTransaction(trxid);
+            Transaction trx = await this.pooledTransaction.GetTransaction(trxid);
```
mikedennis: Can the service injected be potentially null? The original code seemed to think so but not sure if that's a valid assumption or not.

noescape00: It's valid. If the default value == null then DI wont fail on the missing components and just use null. 

But there was a wrong asusmption of another type: 

`unspentOutputs = await this.FullNode.NodeService<IPooledGetUnspentTransaction>()?.GetUnspentTransactionAsync(trxid);`

'?' doesn't make any sense here since if there is no service- it will throw, not return null.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
noescape00: Fixed: added comments for those components that are commented in the codebase. I didn't touch those that are never commented because this is probably something for another PR.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
Neurosploit: For the parts in the fullnodecontroller where we expect an exception to be thrown when getting the service and it is not available the .? does not make any sense.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -23,8 +20,26 @@ public class FullNodeController : FeatureController
...
+        /// <summary>Manager of the longest fully validated chain of blocks.</summary>
+        private readonly IConsensusLoop consensusLoop;
+
         public FullNodeController(
             ILoggerFactory loggerFactory,
+            IPooledTransaction pooledTransaction,
+            IPooledGetUnspentTransaction pooledGetUnspentTransaction,
+            IGetUnspentTransaction getUnspentTransaction,
+            INetworkDifficulty networkDifficulty,
+            IConsensusLoop consensusLoop,
```
dangershony: What is this is null? `IConsensusLoop `

dangershony:
@noescape00 see my last comment, apart for that I am ready to push

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
noescape00: As far as I know we don't use RPC in conscienceless environments so it shouldn't be null. 
But I can add a default = null just in case if you want.

noescape00: Fixed.

Though if we are to run RPC in consensussless environments this whole class should be edited. Most of the methods cannot exist without a consensus dependency.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -35,11 +35,11 @@ public class FullNodeController : FeatureController
 
         public FullNodeController(
             ILoggerFactory loggerFactory,
-            IPooledTransaction pooledTransaction,
-            IPooledGetUnspentTransaction pooledGetUnspentTransaction,
-            IGetUnspentTransaction getUnspentTransaction,
-            INetworkDifficulty networkDifficulty,
-            IConsensusLoop consensusLoop,
+            IPooledTransaction pooledTransaction = null,
```
dangershony: You might also need to fix the part that calls this code, but this is for a later PR perhaps as now it wont crash.

noescape00: Yeah, will add the checks in this one.

noescape00: Fixed

dangershony:
Please push when tests passed (unless you are confident enough to not wait)

noescape00:
Let's wait for the tests just in case. 

ps
But even after they pass I can't push to the master, don't have the rights 🙂

dangershony:
>But even after they pass I can't push to the master, don't have the rights

Ah right, yeah not yet ;)

PR #1085, opened 2018-01-21
Removing unused class Bounded Dictionary 

### LINKED ISSUE #741: [StratisBitcoinFullNode] BoundedDictionary
FILE : https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/Utils/BoundedDictionary.cs

```
int _MaxItems;
		ConcurrentDictionary<TKey, TValue> _Dictionnary = new ConcurrentDictionary<TKey, TValue>();
		ConcurrentQueue<TKey> _Queue = new ConcurrentQueue<TKey>();
		public BoundedDictionary(int maxItems)
		{
			_MaxItems = maxItems;
		}
```

`int _MaxItems` should be uint, can cause infinite loop if set to -1 and Clean function

```
private void Clean()
		{
			while(_Queue.Count > _MaxItems)
			{
				TKey result;
				if(_Queue.TryDequeue(out result))
				{
					TValue result2;
					_Dictionnary.TryRemove(result, out result2);
				}
			}
		}
```

_MaxItems should be readonly  and renamed to respect naming convention

`private readonly uint maxItems = default(uint)`

degant:
Changing MaxItems in `BoundedDictionary` to `readonly uint` and also renaming as per convention. Raising PR as per the comments in issue thread. Also fixed typo in `_Dictionnary`. Added a Unit-Test as well.

EDIT: Removing the class `BoundedDictionary` as per the discussion
Fix #741


Aprogiena:
We are not using this class. I think much better fix of #741 would be to remove the file with the class.

noescape00:
Now probably Pr's title should be changed so it represents what's in it since PR's title will be the commit's description after the merge.

PR #1091, opened 2018-01-23
 Remove the "name" parameter in NodeSettings

bokobza:
Removed the parameter name as well as the property as it is not really used apart for setting the root folder where the different blockchain files are stored. Furthermore, it lead to confusion and errors as the wrong name was sometimes used for another network.

### DIFF
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -283,7 +283,7 @@ namespace NBitcoin
 
         public string Name { get; private set; }
 
-        /// <summary> The name of the root folder containing the different blockchains. </summary>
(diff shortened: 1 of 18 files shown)


src/NBitcoin/Network.cs
```diff
@@ -327,6 +283,12 @@ public PubKey AlertPubKey
 
         public string Name { get; private set; }
 
+        /// <summary> The name of the root folder containing the different blockchains. </summary>
```
dangershony: Maybe we can elaborate on the description?
"The root folder represents blockchains of the same consensus rules, for example testnet, regtest, segwit(testchain) etc.."

bokobza:
how's that?
`/// <summary> The name of the root folder containing blockchains operating with the same consensus rules (for now, this will be bitcoin or stratis). </summary>`

PR #1093, opened 2018-01-23
Added testnet API ports.

bokobza:
Added ports 38220 and 38221 to be the API ports for bitcoin testnet and stratis testnet respectively.
This allows mainnet and a testnet nodes to be run at the same time on a single machine.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
+++ b/src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
@@ -15,6 +15,12 @@ namespace Stratis.Bitcoin.Features.Api
         /// <summary>The default port used by the API when the node runs on the Stratis network.</summary>
         public const int DefaultStratisApiPort = 37221;
 
+        /// <summary>The default port used by the API when the node runs on the bitcoin testnet network.</summary>
+        public const int TestBitcoinApiPort = 38220;
+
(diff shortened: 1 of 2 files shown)


bokobza:
https://github.com/stratisproject/StratisBitcoinFullNode/issues/806

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -48,7 +54,18 @@ public void Load(NodeSettings nodeSettings)
             var apiHost = config.GetOrDefault("apiuri", DefaultApiHost);
             Uri apiUri = new Uri(apiHost);
 
-            var apiPort = config.GetOrDefault("apiport", nodeSettings.Network.IsBitcoin() ? DefaultBitcoinApiPort : DefaultStratisApiPort);
+            // Find out which port should be used for the API.
+            int port;
+            if (nodeSettings.Network.IsBitcoin())
```
dangershony: I believe we agreed this will go in Network right? so we don't need to make such checks as `Network.IsBitcoin()`

bokobza: These are only when the API feature is up. At the moment the features have their own settings, but when we redesign Network and its settings we can certainly review that.

Issue #1089, opened 2018-01-22
Miner shutdown not disposed properly 


dangershony:
```
info: Stratis.Bitcoin.Connection.ConnectionManagerBehavior[0]
      Peer '[::ffff:52.174.146.39]:26178' offline, reason: 'unknown'.
info: Stratis.Bitcoin.Connection.ConnectionManagerBehavior[0]
      Peer '[::ffff:51.141.45.80]:26178' offline, reason: 'unknown'.
info: Stratis.Bitcoin.FullNode[0]
      Waiting for PeriodicLog to finish.
info: Stratis.Bitcoin.Features.Api.ApiFeature[0]
      API stopping on URL 'http://localhost:37221/'.
info: Stratis.Bitcoin.FullNode[0]
      Waiting for PosMining.Stake to finish.
info: Stratis.Bitcoin.FullNode[0]
      PosMining.Stake stopping
crit: Stratis.Bitcoin.FullNode[0]
      PosMining.Stake threw an unhandled exception
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'LoggerFactory'.
   at Microsoft.Extensions.Logging.LoggerFactory.CreateLogger(String categoryName)
   at Stratis.Bitcoin.Features.Miner.PowBlockAssembler..ctor(IConsensusLoop consensusLoop, Network network, MempoolSchedulerLock mempoolLock, ITxMempool mempool, IDateTimeProvider dateTimeProvider, ChainedBlock chainTip, ILoggerFactory loggerFactory, AssemblerOptions options) in C:\Users\Dan\Documents\GitHub\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Miner\PowBlockAssembler.cs:line 192
   at Stratis.Bitcoin.Features.Miner.PosBlockAssembler..ctor(IConsensusLoop consensusLoop, Network network, MempoolSchedulerLock mempoolLock, ITxMempool mempool, IDateTimeProvider dateTimeProvider, StakeChain stakeChain, IStakeValidator stakeValidator, ChainedBlock chainTip, ILoggerFactory loggerFactory, AssemblerOptions options) in C:\Users\Dan\Documents\GitHub\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Miner\PosBlockAssembler.cs:line 33
[...]

PR #1098, opened 2018-01-24
Make stake progress checks thread safe

dangershony:


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -243,6 +243,15 @@ public class CoinstakeContext
         /// <summary>Loop in which the node attempts to generate new POS blocks by staking coins from its wallet.</summary>
         private IAsyncLoop stakingLoop;
 
+        /// <summary>Indicates that the stake flag is not in progress.</summary>
+        public const int StakeNotInProgress = -1;
+
+        /// <summary>Indicates that the stake flag is in progress.</summary>
+        public const int StakeInProgress = 1;
+
+        /// <summary>a flag that indicates if stake is on/off based on the <see cref="StakeInProgress"/> and <see cref="StakeNotInProgress"/> constants.</summary>
```
bokobza: * capital `a` at the beginning.
* indicates if stak`ing` is on/off

bokobza:
I fail to see why this is really required here.
We call Stake() and StopStake() only once in a while, and the probability of having 2 threads entering these methods at the same time is extremely minimal.
It doesn't hurt the code I guess, but it does make it more difficult to understand. I think even a simple flag could do the job here.

Also what happens if we start staking but the loop fails to initialize? We'd be stuck. Maybe then we need more states.

dangershony:
I believe this PR fixes the bugs reported here
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1089

>the probability of having 2 threads entering these methods at the same time is extremely minimal

This is true but from my experience, this will happen now and then, so why not make sure it wont.
 
>Also what happens if we start staking but the loop fails to initialize? We'd be stuck. Maybe then we need more states.

Yeah currently exception that are not handled will stop the miner and throw, but this is also the behaviour before this PR.
If there is an exception we just stop staking (and disable staking)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -363,7 +373,8 @@ public class CoinstakeContext
         public IAsyncLoop Stake(WalletSecret walletSecret)
```
fassadlr: Personally, I am not a big fan of "hiding" logic this. I prefer to code with intent. I.e. if we call the Stake method, we stake. We shouldnt check inside the Stake method if staking is running. Preferably when we call Stake() and we are staking, the method should throw and exception.

I know we have it like this because of the Miner controller but:

`public IActionResult StartStaking([FromBody]StartStakingRequest request)`

should check to see if we are staking before call Stake(), e.g. 

```
var staking = this.fullNode.NodeFeature<MiningFeature>(true);
if(staking.CanStake)
    staking.StartStaking(request.Name, request.Password);
```

This is more intuitive and removes the need for thread safety checks etc.

noescape00:
>I believe this PR fixes the bugs reported here
#1089

In 1089 your log shows a `ObjectDisposedException` when `PowBlockAssembler` tries to use `LoggerFactory`.
So it looks like DI tried to create an instance of `PowBlockAssembler` but when it did the logger factory was disposed already. Most likely you got this one when you launched the node, staking started (but did not reach the moment when it requires the `PowBlockAssembler` so DI still have this one unresolved) and then you shut the node down and it disposes the factory. Right after that we need that assembler and it throws. 

So to me a fix for #1089 would be the shutdown sequence fix (probably adding a Dispose() method to the miner so we can wait until the moment we can dispose other stuff)

ps
`public void StopStake()` is never called on shutdown.

fassadlr:
@noescape00 I can confirm that StopStaking is only ever called from the Controller! This is a bug.

Also, perhaps see my review comment? Thoughts?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: I agree. Having `StartStaking`\`StartStakingContinously` in the miner sounds like a more logical method name. 

We probably can rename controller's `StartStaking` to smth like RequestStartStaking as well.

noescape00: >should check to see if we are staking before call Stake(), e.g.

>if(staking.CanStake)
    staking.StartStaking(request.Name, request.Password);

>This is more intuitive and removes the need for thread safety checks etc.

But the ` public IActionResult StartStaking([FromBody]StartStakingRequest request)` can be called twice at the same time so there is a race condition and we still need the thread safety checks.

fassadlr: Yes, so that is what a possible `CanStake` property will be used for. The moment that the asyncloop starts up in the Stake() method, we set CanStake to false. I agree, we still need to make the start of the loop thread safe.

The point that I'm also trying to make is that we should expose whether or not we can can stake so that it can interrogated by whatever consumer. It shouldn't be hidden. Therefore, we should first check IF we can stake and then call Stake(). The Stake() method shouldn't be responsible for determining whether it should stake or not ;)

noescape00: >Yes, so that is what a possible CanStake property will be used for. The moment that the asyncloop starts up in the Stake() method, we set CanStake to false. I agree, we still need to make the start of the loop thread safe though.

To make it thread safe we'd need to call CanStake check and Stake() inside a lock. 
So I guess  I'm fine with the safety checks inside the Stake() since this way we won't make the caller take care of thread safety. 

>The point that I'm also trying to make is that we should expose whether or not we can can stake so that it can interrogated by whatever consumer.

Yeah, that would be nice to have.

fassadlr: > To make it thread safe we'd need to call CanStake check and Stake() inside a lock.

Hmm you do have a point here! Just looking at the code again... I see that when the node starts up, we start the MiningFeature and subsequenty PosMinting IF the stake arg was passed. So in theory we can only start and stop staking if it was enabled on node start up. Also, seeing as this a wallet call, I can't see how multiple threads will be able to call this method?

But I suppose to make it air tight we need to ensure that it can't happen. Ever!

noescape00: > So in theory we can only start and stop staking if it was enabled on node start up. 

Not really. Enabling it by using "-stake" will just make node call `MiningFeature.StartStaking` on startup. 

But when we call StartStaking from the controller there is no check for the `-stake` startup option to be == true.

>Also, seeing as this a wallet call, I can't see how multiple threads will be able to call this method?

Imagine that someone uses the wallet and presses the start staking button 2 times- 2 calls of MinerController.StartStaking. Here is the race condition so it's possible that both of the calls will pass the .CanStake check and both of them call StartStaking(). 
That's why I think it's actually a good idea to have thread safety checks inside the StartStaking() somewhere like @dangershony  did this.

dangershony:
Ahh one more thing to note we are waiting on the stake in the dispose method of the feature.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: @fassadlr there is still a possibility of a race if we don't lock the entry point to the method, this method is called form the controller and so is a high candidate for synchronisation issues.

noescape00:
Yeah, looks good to me. 

Also we should close https://github.com/stratisproject/StratisBitcoinFullNode/issues/1089 when this one is merged.

PR #1099, opened 2018-01-24
[P2P] Remove LastSeen from NetworkPeer and create on PeerAddress

### LINKED ISSUE #1062: [PeerAddress] Move LastSeen from NetworkPeer to PeerAddress
Currently LastSeen on NetworkPeer is not used at all, it is only set to. We need to "move" that field to PeerAddress as we eventually want to remove NetworkAddress entirely.

After that field has been moved, we need to update it via PingPongBehaviour. This will enable us to know how "live" a peer is.

Future PRs will then use this field to extend/improve peer selection.

fassadlr:
Fix #1062

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
@@ -70,10 +70,11 @@ namespace Stratis.Bitcoin.P2P
                     peer.SendMessageVoidAsync(addressPayload);
                 }
 
-                if (message.Message.Payload is PingPayload pingPong)
+                if (message.Message.Payload is PingPayload ping ||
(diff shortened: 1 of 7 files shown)


src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
```diff
@@ -69,6 +69,12 @@ private void AttachedPeer_MessageReceived(NetworkPeer peer, IncomingMessage mess
                     var addressPayload = new AddrPayload(endPoints.Select(p => new NetworkAddress(p)).ToArray());
                     peer.SendMessageVoidAsync(addressPayload);
                 }
+
```
dangershony: We should probably do this for both ping and pong no?

Also note `this.AttachedPeer` might be null, its safer to create a local variable.

PR #1100, opened 2018-01-24
Allow WO wallet to track incoming and outgoing tx

zeptin:
The watch-only wallet currently only seems to properly watch incoming transactions. This extends the functionality to detecting transactions being spent from the watched addresses as well.

### DIFF
--- a/src/Stratis.Bitcoin.Features.WatchOnlyWallet.Tests/WatchOnlyWalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.WatchOnlyWallet.Tests/WatchOnlyWalletManagerTest.cs
@@ -164,6 +164,49 @@ namespace Stratis.Bitcoin.Features.WatchOnlyWallet.Tests
             Assert.Equal(Money.Coins(1.60128533m), balance);
         }
 
+        [Fact]
+        [Trait("Module", "WatchOnlyWalletManager")]
+        public void Given_AWatchedAddress_ThenCanCalculateComplexRelativeBalance()
+        {
+            DataFolder dataFolder = CreateDataFolder(this);
+            
+            // Create transactions to be received.
+            string[] transactionsHex = new[]
+            {
+                "01000000018161735257ebaf9586bac9fd71955cbb833ea88a845d3a146d462c925c951c97010000001716001463a186b12a8e7c03d80eddc84c73f88e7ac41e3cffffffff02c03b4703000000001976a91454e37ad52373b08ad47dac9c909e2a86562740c088acfa4fc8fa0000000017a9148d26b9dc3a4d6105164567e37fe3a8e53fb60e958700000000",
+                "0100000001cfc7db3f85932c636170ea95f3f77ea674283f981a6a1b36c9b2fd9a8248ea2c000000006b483045022100ba70c4538193e4a228666b8db58256c6bb9225f6fe39441a949f49f6964a947d022048508c50ed4a30df1f222d3dd010c10d5a0852d47843e96277c5b730b3194843012102a31bf228b4508abe3f0aa5f91c53732dfd49233c2fef3c5cfa68f2aa12cc71b4ffffffff02b0e13403000000001976a914546f3bb7a70a077d51bed1bf6acb55dead27024388ace0c810000000000017a914e3925ce02760a2fec19871938b0ae5c740fa2b908700000000",
+                "0100000001a6391afc9adf3f385bd7261d6e120511203b60a9dacd3e99cc7010c7a22722ce0100000017160014096a4e906d7807137a10824ce68a0297a3d14e50ffffffff02e09da301000000001976a91454e37ad52373b08ad47dac9c909e2a86562740c088acb94b56e90000000017a91421649ed0b7412724abba0b27c3224beb3a7530e18700000000",
+                "010000000180f6a5f49b79029333d1ab406424da4bd5f4378b19b1f8246b496f32cf215dc6000000006a4730440220080b0dfa1a93e8f4ebf9415727f60504df95c3a7dea7fe4e4053d0bb3f54540202202e9d02bbb642d7c55d637bd9a56265f42dc1626e933031d638507223846a5e53012102a31bf228b4508abe3f0aa5f91c53732dfd49233c2fef3c5cfa68f2aa12cc71b4ffffffff02225b9101000000001976a914e15f2eb7b19ae02102a1c24452fbd3f516dca8f888ace0c810000000000017a914585cffb7dc06ac82119fba0e36e1e95b534443958700000000",
+                "01000000017f926a306647e0bf3ad12c86fca22dbce2f31161006d0723041963dbf31b17ae0100000017160014c72367e145246374dbb748fa274661cd7aece41effffffff02f0ced100000000001976a91454e37ad52373b08ad47dac9c909e2a86562740c088acf120a8d80000000017a9149bdfb11d7b870c364b4167d921339284b52ba1dc8700000000",
+                "0100000001dd19f894733ab950ae8e772ef781fe299f80f4e813ee4cb6ac067fedea3a052c000000006b483045022100d4d6570c054bcbafa8178e9074f2be32b078e0761537bfe4fb7ad1eea949ab66022022e31d6f43d9418b6dc02147aefb380a4730386364bd406365e790826400570b012102a31bf228b4508abe3f0aa5f91c53732dfd49233c2fef3c5cfa68f2aa12cc71b4ffffffff029e90bf00000000001976a914128240d302a4aadcdd08d241b54fa4ef11acb21388ace0c810000000000017a914188d767b139ef64ce4efa091c2957c6137fbe0ce8700000000"
+            };
+
+            var block = new Block();
+            foreach (var transactionHex in transactionsHex)
+            {
+                block.AddTransaction(new Transaction(transactionHex));
+            }
+
+            block.UpdateMerkleRoot();
+
+            var walletManager = new WatchOnlyWalletManager(DateTimeProvider.Default, this.LoggerFactory.Object, Network.TestNet, dataFolder);
+            walletManager.Initialize();
+            walletManager.WatchAddress("moFoZk1figjfEe1Z49GUePXy2KqYr6DioL");
+            walletManager.ProcessBlock(block);
+
+            // Manual calculation of expected result
+            // 2cea48829afdb2c9361b6a1a983f2874a67ef7f395ea7061632c93853fdbc7cf +0.55
+            // c65d21cf326f496b24f8b1198b37f4d54bda246440abd1339302799bf4a5f680 +0.275
+            // 2c053aeaed7f06acb64cee13e8f4809f29fe81f72e778eae50b93a7394f819dd +0.1375
+            // 15c4f72e80a7607d87911969a418b18cc513b70588d27694179073ccdcf966f1 -0.1375
+            // 3aeabae1ec7bd6362babf02b0bb51e4b5c183f30fbe593e170365afcdd5a6f7c -0.275
+            // 240afed9840527f55f0087f4594a245219759307b034a983378d5dda91ea08a0 -0.55 = 0
+            // Block explorer confirmation: https://testnet.smartbit.com.au/address/moFoZk1figjfEe1Z49GUePXy2KqYr6DioL
+
+            Money balance = walletManager.GetRelativeBalance("moFoZk1figjfEe1Z49GUePXy2KqYr6DioL");
+            Assert.Equal(Money.Coins(0m), balance);
+        }
+
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -36,6 +37,13 @@ public class WatchOnlyWalletManager : IWatchOnlyWalletManager
         /// <summary>Provider of date time functionality.</summary>
         private readonly IDateTimeProvider dateTimeProvider;
 
+        /// <summary>
+        /// Provides a rapid lookup of transactions appearing in the watch-only wallet.
+        /// This includes both transactions under watched addresses, as well as stored
+        /// transactions.
+        /// </summary>
+        internal Dictionary<uint256, TransactionData> txLookup;
```
noescape00: This probably can be private instead of internal

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -56,6 +64,8 @@ public void Initialize()
         {
             // load the watch only wallet into memory
             this.Wallet = this.LoadWatchOnlyWallet();
+
+            LoadTransactionLookup();
```
noescape00: `this.LoadTransactionLookup();`

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -155,11 +231,75 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
+
+        /// <summary>
+        /// Populate the transaction lookup dictionary with the current
+        /// contents of the watch only wallet's watched addresses and
+        /// transactions.
+        /// </summary>
+        private void LoadTransactionLookup()
+        {
+            var lookup = new Dictionary<uint256, TransactionData>();
+            var watchOnlyWallet = this.GetWatchOnlyWallet();
```
noescape00: unvar please (also on line 272m 274, 280, 291, 375, 383, 413)

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -155,11 +231,75 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
             }
+
+            this.Wallet.WatchedTransactions.TryGetValue(transactionHash.ToString(), out TransactionData existingWatchedTransaction);
+
+            if (existingWatchedTransaction != null)
+            {
+                // The transaction was previously stored, in an unconfirmed state.
+                // So now update the block hash and Merkle proof since it has
+                // appeared in a block.
+                existingWatchedTransaction.BlockHash = block.GetHash();
```
noescape00: I think it may be possible that `block` will be `null` here

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -117,6 +127,68 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
             var transactionHash = transaction.GetHash();
             this.logger.LogDebug($"watch only wallet received transaction - hash: {transactionHash}, coin: {this.coinType}");
 
+            // Check the transaction inputs to see if a watched address is affected.
+            foreach (TxIn input in transaction.Inputs)
+            {
+                // See if the previous transaction is in the watch-only wallet.
+                this.txLookup.TryGetValue(input.PrevOut.Hash, out TransactionData prevTransaction);
```
dangershony: Are you sure this is thread safe?

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -117,6 +127,68 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
+                            // Add the Merkle proof to the transaction.
+                            if (block != null)
+                            {
+                                newTransaction.MerkleProof = new MerkleBlock(block, new[] { transactionHash }).PartialMerkleTree;
+                            }
+
+                            addressInWallet.Transactions.TryAdd(transactionHash.ToString(), newTransaction);
+
+                            // Update the lookup cache with the new transaction information.
+                            this.txLookup.Remove(newTransaction.Id);
```
dangershony: How can it be that a trx is updated? (what can be updated)

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -181,7 +321,8 @@ public WatchOnlyWallet LoadWatchOnlyWallet()
                 Network = this.network,
                 CoinType = this.coinType,
                 CreationTime = this.dateTimeProvider.GetTimeOffset(),
-                WatchedAddresses = new ConcurrentDictionary<string, WatchedAddress>()
+                WatchedAddresses = new ConcurrentDictionary<string, WatchedAddress>(),
+                WatchedTransactions = new ConcurrentDictionary<string, TransactionData>()
```
dangershony: Nice, I think we also need such an index in the wallet itself 😉

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
dangershony: No its fine see comment above `The transaction was previously stored, in an unconfirmed state.` if its confirmed there should be a block.

noescape00: Thanks, I've missed that.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
bokobza: This is not needed here as it's already done in the WatchOnlyWallet's constructor.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
bokobza: You're inside the manager so you don't need a new watchOnlyWallet variable, you have `this.wallet`.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -155,11 +231,75 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
+        /// Populate the transaction lookup dictionary with the current
+        /// contents of the watch only wallet's watched addresses and
+        /// transactions.
+        /// </summary>
+        private void LoadTransactionLookup()
+        {
+            var lookup = new Dictionary<uint256, TransactionData>();
+            var watchOnlyWallet = this.GetWatchOnlyWallet();
+
+            foreach (var address in watchOnlyWallet.WatchedAddresses.Values)
```
bokobza: Here, I suggest moving this functionality to the WatchOnlyWallet class itself, as what's happening in this method is concerning the wallet only and therefore the manager should be the one asking the wallet for this info.
Something like `this.wallet.GetWatchedTransactions()`.

src/Stratis.Bitcoin.Features.WatchOnlyWallet.Tests/WatchOnlyWalletManagerTest.cs
```diff
@@ -132,6 +132,38 @@ public void Given_AWatchedAddress_When_ATransactionIsReceivedInABlock_ThenTransa
...
 
+        [Fact]
+        [Trait("Module", "WatchOnlyWalletManager")]
+        public void Given_AWatchedAddress_When_ATransactionIsReceivedInABlock_ThenCanCalculateRelativeBalance()
+        {
+            // Arrange.
+            DataFolder dataFolder = CreateDataFolder(this);
+
+            // Create the wallet to watch.
+            var wallet = this.CreateAndPersistAWatchOnlyWallet(dataFolder);
```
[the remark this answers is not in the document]
zeptin: Done

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
zeptin: Done

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
zeptin: Done

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -155,11 +231,75 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
+                // more watched addresses.
+                if (!lookup.TryAdd(transaction.Id, transaction))
+                {
+                    // Check to see if there is better information in
+                    // the watched transaction than the watched address.
+                    // If there is, use the watched transaction instead.
+
+                    var existingTx = lookup[transaction.Id];
+
+                    if ((existingTx.MerkleProof == null && transaction.MerkleProof != null) ||
```
bokobza: What if the `transaction.MerkleProof` is null and the `existingTx.MerkleProof` is not? 
You can support both cases if you do something like:
```
if(existingTx.MerkleProof == null)
{
    existingTx.MerkleProof = transaction.MerkleProof;
}

if(existingTx.BlockHash == null)
{
    ...
}
```

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
```diff
@@ -155,11 +231,75 @@ public void ProcessTransaction(Transaction transaction, Block block = null)
...
+                        (existingTx.BlockHash == null && transaction.BlockHash != null))
+                    {
+                        lookup.Remove(transaction.Id);
```
bokobza: you don't need to remove it from the lookup.
If you update the `existingTx` it will be updated in the list automatically.
I suggest adding some unit tests to this method.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
zeptin: If a transaction is added to the watch only wallet, it may be in an unconfirmed state initially (i.e. there is no block hash and no Merkle proof). When the transaction is later confirmed, these details get amended.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
zeptin: Actually, I think you're right. It is possible for ProcessTransaction to be called with the block parameter set to null.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
zeptin: Done

bokobza:
I'm not done with my review. To be continued...

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWalletManager.cs
bokobza: @zeptin did you see this comment?

src/Stratis.Bitcoin.Features.WatchOnlyWallet/WatchOnlyWallet.cs
```diff
@@ -56,6 +56,46 @@ public WatchOnlyWallet()
...
+                // more watched addresses.
+                if (!txDict.TryAdd(transaction.Id, transaction))
+                {
+                    // Check to see if there is better information in
+                    // the watched transaction than the watched address.
+                    // If there is, use the watched transaction instead.
+
+                    TransactionData existingTx = txDict[transaction.Id];
+
+                    if ((existingTx.MerkleProof == null && transaction.MerkleProof != null) ||
```
bokobza: see my previous comment.

bokobza:
Good to go.

PR #1102, opened 2018-01-25
Tests environmental consistency

noescape00:
Don't merge it, I'd like to trigger travis running the tests a few times.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
@@ -17,23 +17,23 @@ namespace Stratis.Bitcoin.Tests.Utilities
         [Fact]
         public async void AsyncManualResetEvent_WaitAsync()
         {
-            AsyncManualResetEvent manualResetEvent = new AsyncManualResetEvent(false);
+            var manualResetEvent = new AsyncManualResetEvent(false);
 
             Stopwatch stopwatch = Stopwatch.StartNew();
 
             Task task = Task.Run(async () =>
             {
-                await Task.Delay(500);
+                await Task.Delay(510);
                 manualResetEvent.Set();
             });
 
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Tests/Utilities/AsyncManualResetEventTest.cs
```diff
@@ -33,7 +33,6 @@ public class AsyncManualResetEventTest
             stopwatch.Stop();
 
             Assert.True(stopwatch.ElapsedMilliseconds >= 500);
-            Assert.True(mreAwaitingTask.IsCompletedSuccessfully);
```
bokobza: strange, why does this make the test fail sometimes?

noescape00: I have no idea 🙂 
Before the last fix it was a check for State == RunToCompletion and it also used to fail. This is probably something environmental because I never had this test failing on win10 machine.

noescape00: Hmm. Actually that might be a problem of the stopwatch because it's still failing

noescape00: Yeah, the problem was in stopwatch- it sometimes shows 499ms passed instead of 500ms.

noescape00:
While running the tests I've noticed that `AsyncLoopRepeatEveryIntervalCanBeChangedWhileRunningAsync` is also not very consistent so I've added fix for it as well. 
I'll rerun the tests 10+ times more just to be sure there is no problem with this 2 tests anymore and after that it's ready to merge.

noescape00:
@bokobza I've ran it ~15 times- all pass. So I think this fix finally makes it consistent on travis.

bokobza:
Great, thank you.

PR #1108, opened 2018-01-27
Bugfix: Prevent mining if node is not fully synced

noescape00:
Disable mining until the moment we're not in IBD and our block repository is  at the same height with the peer that has highest PendingTip.Height of all peers that we're connected to.

Fix: #896

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -78,6 +78,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
         BlockStoreRepositoryPerformanceCounter PerformanceCounter { get; }
 
         bool TxIndex { get; }
+
+        /// <summary>Represents the last block stored to disk.</summary>
+        ChainedBlock HighestPersistedBlock { get; }
     }
 
     public class BlockRepository : IBlockRepository
@@ -103,7 +106,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Provider of time functions.</summary>
         protected readonly IDateTimeProvider dateTimeProvider;
 
-        /// <summary>Represents the last block stored to disk.</summary>
+        /// <inheritdoc />
         public ChainedBlock HighestPersistedBlock { get; internal set; }
 
         public BlockRepository(Network network, DataFolder dataFolder, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
@@ -94,7 +94,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         public void AddNodeStats(StringBuilder benchLogs)
         {
-            var highestBlock = (this.blockRepository as BlockRepository)?.HighestPersistedBlock;
+            var highestBlock = this.blockRepository.HighestPersistedBlock;
 
             if (highestBlock != null)
                 benchLogs.AppendLine($"{this.name}.Height: ".PadRight(LoggingConfiguration.ColumnLength + 1) +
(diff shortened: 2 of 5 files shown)


dangershony:
Hi @noescape00 I have commented on the issue can you have a look?

dangershony:
I don't think this PR is completely correct so blocking the merge

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -447,17 +453,29 @@ public void StopStake()
...
-                    || this.initialBlockDownloadState.IsInitialBlockDownload())
+                bool readyToMine = !this.initialBlockDownloadState.IsInitialBlockDownload();
+
+                // If not in IBD- check if fully synced.
+                if (readyToMine)
                 {
-                    if (this.initialBlockDownloadState.IsInitialBlockDownload()) this.logger.LogTrace("Waiting for IBD to complete...");
-                    else this.logger.LogTrace("Waiting to be connected with at least one synced network peer...");
+                    // Most advanced peer's pending tip.
+                    ChainedBlock chainedBlock = this.connection.ConnectedPeers
```
fassadlr: @noescape00 Can we maybe write this better?

I see that Behaviors is an extension of NetworkPeerCollection... How about we move that whole linq statement into NetworkPeerCollection? Then you can just go:

`this.connection.ConnectedPeers.HighestPendingTip;`

That way you can write a quick unit test for it? In other words we are moving calculated logic where it belongs and making it testable 💯

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -56,7 +56,7 @@ public static BlockStoreManager BlockStoreManager(this FullNode fullNode)
 
         public static ChainedBlock HighestPersistedBlock(this FullNode fullNode)
         {
-            return (fullNode.NodeService<IBlockRepository>() as BlockRepository).HighestPersistedBlock;
+            return fullNode.NodeService<IBlockRepository>().HighestPersistedBlock;
```
fassadlr: Nice!

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -94,7 +94,7 @@ public virtual BlockStoreBehavior BlockStoreBehaviorFactory()
 
         public void AddNodeStats(StringBuilder benchLogs)
         {
-            var highestBlock = (this.blockRepository as BlockRepository)?.HighestPersistedBlock;
+            var highestBlock = this.blockRepository.HighestPersistedBlock;
```
fassadlr: ;)

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -78,6 +78,9 @@ public interface IBlockRepository : IDisposable
         BlockStoreRepositoryPerformanceCounter PerformanceCounter { get; }
 
         bool TxIndex { get; }
+
+        /// <summary>Represents the last block stored to disk.</summary>
+        ChainedBlock HighestPersistedBlock { get; }
```
fassadlr: OK please double check this, as I'm 99.9% sure that `IBlockRepository.BlockHash` IS the last block stored to disk?

noescape00: That's correct. But `BlockHash` is a hash and not a `ChainedBlock` so we can't use it

fassadlr: Oh yes! I see now ... but seeing as we have the repository in PosMinting, can't we just query it from there?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: @dangershony already convinced me to avoid using BlockRepository here so this one is no longer applicable 🙂

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
noescape00: >can't we just query it from there?

Yep, we could. Good idea. 
But BlockRepository is no longer used here so not applicable.

fassadlr:
@dangershony still needs to 👍 before we can merge.

dangershony:
@fassadlr please reads the comments in #896 to how we should fix this I thin this PR can be closed (or fixed)

PR #1111, opened 2018-01-29
[P2P] Reset connection attempts and throttle connection when there are no peers

### LINKED ISSUE #1105: [PeerAddress] Reset ConnectionAttempts after 1 day and increase filtering interval
Currently, we filter attempted peers with a 60sec interval. This needs to be increased to 1 hour.

Also, we don't currently reset the connection attempts, which means that if a peer has had more than 10 failed atttempts we will never try and connect to it again.

fassadlr:
This fixes #1105.

Also, in the scenario where there are no connections, we will employ the burst connection mode BUT if there are no peers to be selected from, the peer connector will continuously try and find peers without any delay. Therefore we need to delay the connection for a bit before trying again. This was the scenario where the logs were being flooded with entries.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddress.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddress.cs
@@ -15,13 +15,13 @@ namespace Stratis.Bitcoin.P2P
         /// <summary>
         /// The maximum amount of times a peer can be attempted within a give time frame.
         /// </summary>
-        internal static readonly int AttemptThreshold = 10;
+        internal static readonly int AttemptThreshold = 5;
 
         /// <summary>
         /// The amount of hours after which the peer's failed connection attempts
         /// will be reset.
         /// </summary>
-        internal static readonly int AttemptThresholdTime = 12;
+        internal static readonly int AttemptThresholdHours = 12;
 
         /// <summary>EndPoint of this peer.</summary>
         [JsonProperty(PropertyName = "endpoint")]
@@ -156,7 +156,7 @@ namespace Stratis.Bitcoin.P2P
         /// <para>
         /// This is reset when the amount of failed connection attempts reaches 
         /// the <see cref="PeerAddress.AttemptThreshold"/> and the last attempt was 
-        /// made more than <see cref="PeerAddress.AttemptThresholdTime"/> ago.
+        /// made more than <see cref="PeerAddress.AttemptThresholdHours"/> ago.
         /// </para>
         /// </summary>
         internal void ResetAttempts()
--- a/src/Stratis.Bitcoin/P2P/PeerAddress.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddress.cs
@@ -15,13 +15,19 @@ namespace Stratis.Bitcoin.P2P
         /// <summary>
         /// The maximum amount of times a peer can be attempted within a give time frame.
         /// </summary>
-        internal static readonly int AttemptThreshold = 5;
+        internal const int AttemptThreshold = 5;
+
+        /// <summary>
+        /// The amount of hours we will wait before selecting an attempted peer again,
+        /// if it hasn't yet reached the <see cref="AttemptThreshold"/> amount of attempts.
+        /// </summary>
+        internal const int AttempThresholdHours = 1;
 
         /// <summary>
         /// The amount of hours after which the peer's failed connection attempts
-        /// will be reset.
+        /// will be reset to zero.
         /// </summary>
-        internal static readonly int AttemptThresholdHours = 12;
+        internal const int AttemptResetThresholdHours = 12;
 
         /// <summary>EndPoint of this peer.</summary>
         [JsonProperty(PropertyName = "endpoint")]
@@ -156,7 +162,7 @@ namespace Stratis.Bitcoin.P2P
         /// <para>
         /// This is reset when the amount of failed connection attempts reaches 
         /// the <see cref="PeerAddress.AttemptThreshold"/> and the last attempt was 
-        /// made more than <see cref="PeerAddress.AttemptThresholdHours"/> ago.
+        /// made more than <see cref="PeerAddress.AttemptResetThresholdHours"/> ago.
         /// </para>
         /// </summary>
         internal void ResetAttempts()
--- a/src/Stratis.Bitcoin/P2P/PeerAddress.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddress.cs
@@ -21,7 +21,7 @@ namespace Stratis.Bitcoin.P2P
         /// The amount of hours after which the peer's failed connection attempts
         /// will be reset.
         /// </summary>
-        internal static readonly int AttempThresholdTime = 12;
+        internal static readonly int AttemptThresholdTime = 12;
 
         /// <summary>EndPoint of this peer.</summary>
         [JsonProperty(PropertyName = "endpoint")]
@@ -156,7 +156,7 @@ namespace Stratis.Bitcoin.P2P
         /// <para>
         /// This is reset when the amount of failed connection attempts reaches 
         /// the <see cref="PeerAddress.AttemptThreshold"/> and the last attempt was 
-        /// made more than <see cref="PeerAddress.AttempThresholdTime"/> ago.
+        /// made more than <see cref="PeerAddress.AttemptThresholdTime"/> ago.
         /// </para>
         /// </summary>
         internal void ResetAttempts()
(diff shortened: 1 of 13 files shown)


src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -141,27 +152,41 @@ public bool Handshaked
...
-        /// Increments <see cref="ConnectionAttempts"/> and sets the <see cref="LastConnectionAttempt"/>.
+        /// Resets the amount of <see cref="ConnectionAttempts"/>.
+        /// <para>
+        /// This is reset when the amount of failed connection attempts reaches 
+        /// the <see cref="PeerAddress.AttemptThreshold"/> and the last attempt was 
+        /// made more than <see cref="PeerAddress.AttemptThresholdTime"/> ago.
+        /// </para>
         /// </summary>
-        internal void SetAttempted(DateTimeOffset peerAttemptedAt)
+        internal void ResetAttempts()
```
dangershony: Why not public?

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -157,12 +161,22 @@ public void AddPeers(NetworkAddress[] networkAddresses, IPAddress source)
...
+        public void PeerAttempted(IPEndPoint endpoint, DateTime peerAttemptedAt)
         {
             var peer = this.FindPeer(endpoint);
             if (peer == null)
                 return;
 
+            //Reset the attempted count if:
+            //1: The last attempt was more than the threshold time ago.
+            //2: More than the threshold attempts was made.
+            if (peer.Attempted &&
```
dangershony: Do we need to check the `peer.Attempted` condition? I guess there is no harm.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -60,8 +60,6 @@ public override void OnStartConnect()
 
         public override async Task OnConnectAsync()
         {
-            this.logger.LogTrace("()");
```
dangershony: Just wondering why remove the trace?

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
```diff
@@ -123,12 +120,15 @@ public override void OnStartConnect()
                 break;
             }
 
-            this.logger.LogTrace("Peer selected: '{0}'", peer?.EndPoint);
-
-            if (peer != null)
+            //If the peer selector returns nothing, we need to wait a bit,
```
dangershony: Maybe add in to the comment more information about how long we are waiting.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -12,6 +12,17 @@ namespace Stratis.Bitcoin.P2P
...
+        /// <summary>
+        /// The maximum amount of times a peer can be attempted within a give time frame.
+        /// </summary>
+        internal static readonly int AttemptThreshold = 10;
+
+        /// <summary>
+        /// The amount of hours after which the peer's failed connection attempts
+        /// will be reset.
+        /// </summary>
+        internal static readonly int AttemptThresholdTime = 12;
```
dangershony: `AttemptThresholdTime` -> `AttemptThresholdHours`

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -12,6 +12,17 @@ namespace Stratis.Bitcoin.P2P
     [JsonObject]
     public sealed class PeerAddress
     {
+        /// <summary>
+        /// The maximum amount of times a peer can be attempted within a give time frame.
+        /// </summary>
+        internal static readonly int AttemptThreshold = 10;
```
dangershony: Perhaps 10 attempts is too much per period?

fassadlr: OK I agree! Lets make it 5?

src/Stratis.Bitcoin/P2P/PeerAddress.cs
fassadlr: Fixed

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
fassadlr: All the other set methods i.e. SetConnected / SetHandshaked are internal. I want to force the consumers outside of this assembly to go via the PeerAddressManager to set these values.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
fassadlr: I just left the logging in the areas where it became applicable, i.e. in the Ifs further down the code when peer selection fails. Otherwise, when I turned on logging, it was completely clogging up the log file.

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
dangershony: Yeah but thats the point of having method entry and exit points, its useful for threading issues, but its also up to what makes sense on if we need to put a trace at a given point or not so ok with me.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -12,6 +12,17 @@ namespace Stratis.Bitcoin.P2P
...
+        /// <summary>
+        /// The maximum amount of times a peer can be attempted within a give time frame.
+        /// </summary>
+        internal static readonly int AttemptThreshold = 5;
+
+        /// <summary>
+        /// The amount of hours after which the peer's failed connection attempts
+        /// will be reset.
+        /// </summary>
+        internal static readonly int AttemptThresholdHours = 12;
```
dangershony: Come to think about it I would reduce the threshold, do we have a mechanism to delete a peer that was not responsive for x days (or X attempts)?

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -253,7 +252,7 @@ private PeerAddress Random(IEnumerable<PeerAddress> peers)
         /// <inheritdoc/>
         public IEnumerable<PeerAddress> Attempted()
         {
-            return this.peerAddresses.Values.Where(p => p.Attempted && p.ConnectionAttempts <= 10 && p.LastConnectionAttempt < DateTime.UtcNow.AddSeconds(-60));
+            return this.peerAddresses.Values.Where(p => p.Attempted && p.ConnectionAttempts < 10 && p.LastAttempt < DateTime.UtcNow.AddHours(-1));
```
dangershony: ok so the logic as it is proposed in the PR is 
- attempted 10 times every hour
- if the threshold of 12 hours and 5 attempts is reached reset attempts to 0

Should this not be `ConnectionAttempts < PeerAddress.AttemptThreshold`
Then it will be attempt 5 once every hour for 12 hours

src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
fassadlr: Yeah, if we leave the entry and exit logs here and there aren't any issues with peer selection, we get hundreds of:

```
PeerConnector.OnConnectAsync()
PeerConnector.OnConnectAsync(-)
```

in the logs.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
fassadlr: What would reduce this to? 

No, there isn't a mechanism currently to delete peers that have been unresponsive. We did discuss this at some point but if I remember correctly, we decided against removing peers from the file?

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: >    attempted 10 times every hour
    if the threshold of 12 hours and 5 attempts is reached reset attempts to 0

Correct. BUT well spotted, I fixed that if to use the constant.

Issue #803, opened 2017-11-26
[POS] Reorg can crash


Aprogiena:
In case of long reorg, this line in PosMinting.cs can crash when GetBlock returns null
```
stakeTx.HashBlock = this.chain.GetBlock((int)set.Height).HashBlock;
```

dangershony:
Ho crap yes, we need to fix it, low priority though.

PR #1115, opened 2018-01-30
BugFix: Prevent null exception if block for minting is null(#803)

kogot:
Checking if GetBlock() return null then set hashBlock as a uint256.Zero

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -492,19 +492,18 @@ namespace Stratis.Bitcoin.Features.Miner
                 {
                     UnspentOutputs set = coinset.UnspentOutputs.FirstOrDefault(f => f?.TransactionId == infoTransaction.Transaction.Id);
                     TxOut utxo = (set != null) && (infoTransaction.Transaction.Index < set.Outputs.Length) ? set.Outputs[infoTransaction.Transaction.Index] : null;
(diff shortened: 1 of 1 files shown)


Aprogiena:
We should not use zero there, we should just skip this UTXO.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -500,11 +500,12 @@ public void StopStake()
                         utxoStakeDescription.TxOut = utxo;
                         utxoStakeDescription.OutPoint = new OutPoint(set.TransactionId, infoTransaction.Transaction.Index);
                         utxoStakeDescription.Address = infoTransaction.Address;
-                        utxoStakeDescription.HashBlock = this.chain.GetBlock((int)set.Height).HashBlock;
+                        utxoStakeDescription.HashBlock = this.chain.GetBlock((int)set.Height)?.HashBlock;
```
dangershony: This is a good I would suggest to put the line`var blockHash = 
 this.chain.GetBlock((int)set.Height)?.HashBlock;` before the `if`
then use it as part of the condition 
` if ((utxo != null) && (utxo.Value > Money.Zero) && (blockHash != null))`

Aprogiena: yeah that would be even better

kogot: Agree, it will be better.

kogot:
@dangershony  can you please review it once you have chance

PR #1122, opened 2018-01-31
Check a trx is empty for POS on the mempool

dangershony:


dangershony:
@Aprogiena just a quick glance

Aprogiena:
Looks okay, there is typo in a comment but that's.only to do, so that's fine. Then there is this unrelated removal of a method, but if tests are okay with that it should be okay. And yes the factoring is needed there, so this is rather quick fix, but hopefully okay

dangershony:
>Then there is this unrelated removal of a method
Yeah it broke the tests as did not set root folder, it was suppose to be picked by @jeremy change, but its more correct anyway to call correct regtest network.

jeremy:
s/jeremy/bokobza 😘

Issue #599, opened 2017-10-16
Block Store race condition


dangershony:
The block store will sometimes fall behind.

Because of this code in consensus
```
                this.chainState.HighestValidatedPoW = this.Tip;
                if (this.Chain.Tip.HashBlock == item.ChainedBlock?.HashBlock)
                    this.FlushAsync().GetAwaiter().GetResult();

                this.signals.SignalBlock(item.Block);
```

The problem is store uses this condition to store a block
```
                if (this.StoreTip.Height >= this.ChainState.HighestValidatedPoW?.Height)
                    break;
```

A very rare race condition can happen where `HighestValidatedPoW` is updated and store is trying to look for the next block in pending storage collection but `signals.SignalBlock` did not push the block yet to that collection.

Then the node will try to download the block.

noescape00:
Code that this issue is referring to is no longer in the repository. Was this issue resolved?

dangershony:
I believe this is still an issue as this applies to the store persistence and not relay part of the block Signaler.

noescape00:
@dangershony after investigating this one for a while I came to a conclusion that StoreBlockPuller and loopsteps related to it are redundant. Basically the only thing that BlockStore does is getting a block from signaler, reorging the repo if needed and saving blocks to the repository. And every time it actually downloads a block (once in a few thousands if we're lucky to catch it)- it's a wrong download and this block was already downloaded by the LookaheadBlockPuller which is the only block puller that does all the downloading (even though in summary it is said that LookaheadBlockPuller is used only for IBD but actually it is used always. And it's used for the light node where we can't use block store). 

The only blocks that are being downloaded by StoreBlockPuller are ones that has been already downloaded (2nd downloads caused by that race condition). 

So if we remove those unused steps we can get rid of the race condition, occasional high CPU usage (discussed in PR 1125) and make code more simple.
I'll test this thoroughly.

dangershony:
Based on PR #1132 we still need to fix the small CPU boost even on the current code base.

PR #1125, opened 2018-02-01
Minor block store improvement

noescape00:
Added `await Task.Delay(100);` at the `DownloadAndStoreBlocksAsync`:
without that we're sometimes ending up with a while() loop that will have around 3000 iterations per second that are just actively waiting for the block download. That causes a lot of the CPU consumption during the IBD (around 20% CPU consumption reduction with that delay in place).

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
@@ -264,6 +264,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 var result = await this.stepChain.ExecuteAsync(nextChainedBlock, disposeMode, cancellationToken).ConfigureAwait(false);
                 if (result == StepResult.Stop)
                     break;
+
+                if (result == StepResult.Continue)
+                    await Task.Delay(100);
             }
 
             this.logger.LogTrace("(-)");


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
```diff
@@ -264,6 +264,9 @@ private void StartLoop()
                 var result = await this.stepChain.ExecuteAsync(nextChainedBlock, disposeMode, cancellationToken).ConfigureAwait(false);
                 if (result == StepResult.Stop)
                     break;
+
+                if (result == StepResult.Continue)
```
dangershony: Ah interesting you covered all scenarios?

fassadlr: @noescape00 A little further up we check to see if the store's tip is higher than consensus. If it is we break out of this loop. (Line 254)... I.e. we are waiting for the store to catch up to consensus before we download more. Doesn't this already cover your scenario?

noescape00: During the IBD 'contunue' is returned only from ProcessPendingStorageStep.ProcessWhenInIBDAsync- here it returns continue no matter how many blocks were pushed to the block repository (those crazy iterations numbers are encountered when no blocks were pushed). And for some reason that 'tip is higher than consensus' check is passed.

I've tested it a few times - this behavior is encountered only with low peers count or when peers are slow (otherwise we get as many 'continue' as 30% of all blocks downloaded).

In a normal scenario this fix doesn't reduce the CPU usage. 

Hmm. This actually might be connected to the race condition that I was trying to find for some time https://github.com/stratisproject/StratisBitcoinFullNode/issues/599 so I'm not sure now if the delay is the right fix for the problem. I probably should investigate it more in depth to find out if that 'continue' is justified at all times (because it looks like in some cases it doesn't).

fassadlr: @noescape00 yes, Continue is only returned from the ProcessPendingStorage step. But I see its returned from both ProcessWhenInIBDAsync and ProcessWhenNotInIBDAsync.

I have a suspicion the high iteration count comes from us not adding a delay when the pending storage count is less than the processing threshold:

```
        private async Task<StepResult> ProcessWhenInIBDAsync(ProcessPendingStorageContext context)
        {
            this.logger.LogTrace("({0}:'{1}',{2}.{3}:{4})", nameof(context.NextChainedBlock), context.NextChainedBlock, nameof(this.BlockStoreLoop.PendingStorage), nameof(this.BlockStoreLoop.PendingStorage.Count), this.BlockStoreLoop.PendingStorage.Count);

            if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold)
                return StepResult.Continue;
```

Thoughts?

noescape00: Yes, that might be on option. 
But I'll also have to check how new blocks downloading is scheduled and the probability of that delay slowing down IBD overall.  

Ideally I'd switch from the async loop + inner while() with some kind of callback system like I did in blockStore signaled but taking in count the complexity of block pulling process that might be a task for later. 

Right now the plan is to tweak my server node to only connect to a local node and slow down the uploading process. I think that way I will be able to reproduce consistently what is causing lot's of 'continue' returns and start fixing that. Maybe that setup will also let me catch the race condition.

fassadlr: Yeah, maybe post the logs here, so that I can also see what is being produced when lots of those "continues" happen ;)

Issue #1090, opened 2018-01-22
[Users of NetworkPeer + NetworkPeer] Beyond NetworkPeer refactoring - separate Dispose from Disconnect to avoid cyclic calls


Aprogiena:
Currently various behaviors call disconnect/dispose on network peer and this leads to cyclic calls like 

`state changed -> on state changed callback -> dispose peer -> disconnect peer ->state changed to offline -> on state changed callback -> dispose peer`

to prevent those, we need to do following steps:
- separate disconnection from disposal (disconnection would only mean doing what is now in `NetworkPeerConnection.Disconnect`)
- allow disconnection from callbacks but not disposals (this is optional, but a good think to have to see if we fixed those everywhere)
- each component that creates network peer has to manage the list of peers and dispose properly 

kind of a big task, which won't be included in NetworkPeer refactoring series

once this is done, we can relax (remove functionality) AsyncQueue and maybe even AsyncExecutionEvent, which are now coded in a way that this cyclic calling is allowed

then we can very much refactor NetworkPeer's `Disconnect` and NetworkPeerConnection's `Shutdown`, and `Dispose`, those could be much much simpler after this is no longer needed

Aprogiena:
we would maybe want then to introduce OnDisconnect event in NetworkPeer so that those components owning the peer can get notified when this happens (i.e. completely different behavior calls disconnect on their peer) and so they can initiate disposal, but either they would manage themselves that or we would call this OnDisconnect in a non-synchronized fashion, so that we allow disposals from those

noescape00:
I think we can avoid cyclic calls without removing the ability to call Dispose from the callback. 

So what we can do is that: 
1) Save current `NetworkPeerState` when we start callback execution. 
2) After each callback is executed check current state against saved state. If they are not equal- stop current execution because newer state will run callbacks again. 

What's the point of notifying peers about a new state when this state is already outdated?

Aprogiena:
I don't think that would be safe to do, I can imagine scenarios where component relies on something being that when the peer is switched to certain state and if you just about the execution it won't happen.

So to me it seems like such a counterintuitive feature of the callback could lead to more problems than it would solve.

And also because the behaviours are not very well designed, I would rather see refactoring that would clean it up

noescape00:
>I don't think that would be safe to do, I can imagine scenarios where component relies on something being that when the peer is switched to certain state and if you just about the execution it won't happen.

[...]

PR #1126, opened 2018-02-01
Refactor/the node job part 25

Aprogiena:
Finally, the last piece of the puzzle (although we would like to see further work in related areas as described in #1090). But the node job is finished with this last part.

- network peer interfaces 
- interfaces everywhere (except two peer connector tests that already worked without interfaces, but maybe they can be changed as well to make them simpler)
- removal of unused stuff and extensions moved to base classes, simplifications, minor improvements, code style
- fix of network peer disconnection reason display
- node->peer renaming

Note that the first commit of this part is `removed unused property`

Enjoy


dangershony:
From reviewing on github this generally looks good to me, I still need to download and run the integration tests.

PR #1129, opened 2018-02-02
Keepalive Timer

bokobza:
Added a timer to keep the node going until an interval has elapsed.
* This will be used by the Breeze and FullNode.UI apps in order to provide a redundancy scenario in case the UI does not successfully close the background full node process.
* A timer with a timeout of 60s is initialized in this manner: `dotnet run -testnet -keepalive=60`
* Every time a call is made to any API endpoint, the timer is reset.


src/Stratis.StratisD/Program.cs
```diff
@@ -31,15 +33,22 @@ public static void Main(string[] args)
...
                 var node = new FullNodeBuilder()
                     .UseNodeSettings(nodeSettings)
                     .UseStratisConsensus()
                     .UseBlockStore()
                     .UseMempool()
                     .UseWallet()
                     .AddPowPosMining()
-                    .UseApi()
+                    .UseApi(null, options =>
+                    {
```
dangershony: Can we not move this in to the feature? would make the builder part much cleaner.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -58,7 +58,21 @@ public override void Initialize()
...
-            this.TryStartKeepaliveMonitor();
+            // Start the keepalive timer, if set.
+            // If the timer expires, the node will shut down.
+            if (this.apiFeatureOptions.KeepaliveTimer != null)
+            {
+                this.apiFeatureOptions.KeepaliveTimer.Elapsed += (sender, args) =>
+                {
+                    this.logger.LogInformation($"The application will shut down because the keepalive timer has elapsed.");
+
+                    this.apiFeatureOptions.KeepaliveTimer.Stop();
```
dangershony: I suggest adding this lines in the feature Dispose method as well.
```
  if (this.apiFeatureOptions.KeepaliveTimer != null)
  {
                   this.apiFeatureOptions.KeepaliveTimer.Stop();
                   this.apiFeatureOptions.KeepaliveTimer.Enabled = false;
```

Also can we call a dispose on the timer? (is the timer disposable?)

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -74,41 +88,19 @@ public override void Dispose()
...
-        public KeepaliveMonitor KeepaliveMonitor { get; private set; }
+        public Timer KeepaliveTimer { get; private set; }
 
         public void Keepalive(TimeSpan timeSpan)
         {
-            this.KeepaliveMonitor = new KeepaliveMonitor { KeepaliveInterval = timeSpan };
+            this.KeepaliveTimer = new Timer
+            {
+                AutoReset = false,
+                Interval = timeSpan.TotalSeconds * 1000
```
dangershony: Why the multiplication?
`timeSpan.TotalSeconds * 1000`

dangershony:
Minor comments looks good.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
bokobza: because the interval takes in miliseconds.

src/Stratis.StratisD/Program.cs
bokobza: no you can't that's how this is designed. You get an options object, that you need to initialize here.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
bokobza: good idea :-)

src/Stratis.StratisD/Program.cs
dangershony: Fair enough. 
Optional we could add a second method in the feature to do that.
```
 .UseKeepAliveApi(args)
```

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
dangershony: Ah in that case rename the method parameter to `keepaliveSeconds` thats how we do in other places.

src/Stratis.StratisD/Program.cs
bokobza: Don't really like passing the whole args objects to features.
Also methods that look like  .UseXXX() are really used for features and I think the keepalive should stay as an api option as 1. it depends on it 2. it really means that if the api hasn't been called, kill the app.

PR #1132, opened 2018-02-04
Block store refactoring

noescape00:
This PR refactors the Block Store.

### Why did it need a refactoring? 
1) There was a race condition that occasionally would make node start downloading blocks that were already downloaded.
2) Occasional high CPU usage when loopsteps would iterate with no delay without doing anything. 
3) Architecture behind the BlockStoreLoop was greatly over engineered (loop that contains 4 steps, 2 inner steps, steps executor etc) => therefore hardly readable and maintainable.

### How this PR improves it?

1) Code is cleaner and more readable now (3 times more compact without functionality loss).
2) Implememnted batches (they kind of were implemented before but only worked if we've received a lot of blocks in 1 iteration so basically the batches were never reaching even 50% of the target batch size).
3) Fixed race condition.
4) Fixed occasional high CPU usage.
5) Implemented faster reaction to the new blocks- trigger instead of 1 sec interval loop.
6) More correct arhitecture wise (no steps, inner steps, unnesessary contexts, step results, no download code checks in every iteraction). Previos architecture was also made to include blocks pulling into a normal scenario but in fact LookaheadBlockPuller does the downloading for all cases except for when node crashed and consensus tip was initialized ahead of block store tip.

How to review:

Basically there are only 3 classes for the review so I'd suggest checking them instead of going through the commits: StoreBlockPuller.cs, PendingStorageProcessor.cs, BlockStoreLoop.cs

Fix: #599


fassadlr:
@noescape00 I can tell straight-away that this PR will need to be done in smaller batches. @dangershony and myself spent about 2 months to rebuild the old BlockStoreLoop and we went through a massive amount of refactoring to find the sweet spot with regards to all the different areas/facets of downloading blocks and processing pending storage etc.

Prior to the current implementation we **did** have a huge while loop (similar to what you have re-introduced now) which made unit testing VERY difficult.  Hence the decision was made to refactor it into classes for that exact same reason.

>    There was a race condition that occasionally would make node start downloading blocks that were already downloaded.
Why can't we look at this issue on its own and figure out why this is happening?

>    Occasional high CPU usage when loopsteps would iterate with no delay without doing anything.
This can be easily fixed with introducing a delay (granted, some investigation will need to be done).

>    Architecture behind the BlockStoreLoop was greatly over engineered (loop that contains 4 steps, 2 inner steps, steps executor etc) => therefore hardly readable and maintainable.
As said above, classes behind each of the 4 steps was introduced so that separation exists, making unit testing easier. 

These  changes to me is removing confidence from the system as you have removed a lot of unit tests that were put there specifically for all the various scenarios. 

Having done a quick review, it seems to me that you've effectively just moved the logic we had in the individual steps, BACK into the whole loop? I.e. previously they were classes, now they are just methods?
Also, I don't see unit tests for those individual methods now anymore?

I really think we should first try and figure out what is going wrong with the current problem and identify the area in the BlockStoreLoop where the issue is THEN look at refactoring this area.

I would first like @dangershony comments on this.

noescape00:
>There was a race condition that occasionally would make node start downloading blocks that were already downloaded.

>Why can't we look at this issue on its own and figure out why this is happening?

This is happening because we have a 1 sec async loop:
1) New block from LookAheahBlockPuller is received
2) ConsensusTip is updated 
3) Loop iteration execution starts and we bump into 
```
if (this.StoreTip.Height >= this.ChainState.ConsensusTip?.Height)
                    break;
```
and continue to block steps execution
4) BlockStoreSignaled gets a new block that was signaled by the ConsensusLoop and it is added to the PendingStorage


Originally I thought that I can just replace the loop with the trigger-based system but then I've figured out that in normal scenario we never need the `BlockStoreInnerStepFindBlocks` and `BlockStoreInnerStepReadBlocks`. We only need to download new blocks in that loop in case node crashed during the prev run in a specific way: UTXO set was saved but some of the blocks that are used there were not.  So that leads to ConsensusTip being initialized ahead of the store tip and we have some blocks that we need to re-download without passing them to the consensus loop.
So effectively every single block downloaded in the `BlockStoreInnerStepFindBlocks` step (if there was no wrong shutdown) is a re-download which we would like to avoid.

noescape00:
I agree that making a PR like that without discussing the design in details was a bold decision and I shouldn't have done that. Just didn't realize that I'll end up redesigning the whole thing when I started removing download steps in order to fix the race condition.

noescape00:
Another thing that was not possible with current design is having batches of reasonable size. 
Currently the logic of the ProcessPendingStorageStep pushes blocks on every iteration if the Threshold of 10 blocks is met and the actual size of those blocks is always ignored so effectively we have a BlockRepository inserts of 1500-2000 bytes instead of our target of 5 000 000 bytes.

noescape00:
>Prior to the current implementation we did have a huge while loop (similar to what you have re-introduced now) which made unit testing VERY difficult. Hence the decision was made to refactor it into classes for that exact same reason.
>Having done a quick review, it seems to me that you've effectively just moved the logic we had in the individual steps, BACK into the whole loop? I.e. previously they were classes, now they are just methods?

Moving to methods was needed in order to let them use a shared context instead of creating a new context for every iteration (which was preventing having larger batches). 
But indeed those methods can be moved to separated classes and the solution can be optimized to allow easier unit testing while still preserving the arhitectural simplicity.

noescape00:
Basing on the review I'm ok with closing this PR. I think the solution for the race condition can be extracted from it and added as a separate PR while still having the current architecture, though I'll have to use an ugly work-around. 
So I can do the trigger-based processing instead of the async loop and cache the consensus tip on receiving new blocks (basically we can assume that the block that was signaled is the consensus tip and we can ignore the actual tip). That way we can avoid situation when consensus tip is updated but we don't have a block in the pending storage yet. 

And all the issues like batch size, high CPU and architecture improvements can be discussed in separate issues so I'll act on them only when the consensus is reached. 

Let me know what you think.

fassadlr:
@noescape00 I agree, lets close this PR... but keep your branch as reference ;)

>I think the solution for the race condition can be extracted from it and added as a separate PR while still having the current architecture, though I'll have to use an ugly work-around.

Yes, lets do this as the first PR, even its ugly. I want to see what the solution is, albeit crudely. We can then always optimize from there onwards.

The same goes for the batch size and high CPU improvements.

COOL!

PR #1133, opened 2018-02-05
[P2P] Throttle peer discovery

fassadlr:
Currently, we will ask the same peer for addresses multiple times over without any sort of throttling. This needs to change so that we don't ask the same peer for addresses during discovery within a certain time frame.


src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -69,6 +69,12 @@ public IPAddress Loopback
         [JsonProperty(PropertyName = "lastConnectionHandshake", NullValueHandling = NullValueHandling.Ignore)]
         public DateTimeOffset? LastConnectionHandshake { get; private set; }
 
+        /// <summary>
+        /// The last time this peer was discovered from.
+        /// </summary>
+        [JsonProperty(PropertyName = "lastDiscoveredFrom", NullValueHandling = NullValueHandling.Ignore)]
+        public DateTime? LastDiscoveredFrom { get; private set; }
```
dangershony: hmm I don't think we need to store this to disk, I suppose this could an in memory parameter that can be stored in the behaviour.

fassadlr: >hmm I don't think we need to store this to disk, I suppose this could an in memory parameter that can be stored in the behaviour.

Oh yes, I remember you said that we don't need to persist this to disk. I have fixed this 👍 .

>For example we could have a property called LastGetAddressBroadcastTime

Should I rename LastDiscoveredFrom to LastGetAddressBroadcastTime?

> But more importantly we need to throttle the getaddress that is received not sent, as this is provides a DoS attack possibility.

I can do this in a new PR?

fassadlr: @dangershony can you check above questions of mine? Cheers :)

dangershony: Yeah I need to review this better, I will do it soon.

fassadlr:
@dangershony I messed up branch by accident. Submitting a new PR....

PR #1135, opened 2018-02-05
[NetworkPeerServer] Limit maximum inbound connections and stall once it has been reached.

### LINKED ISSUE #1054: Potential service denial attack because we have no max inbound connections limit
We don't have a max inbound connections limit so I guess it's possible to perform a service denial attack on a C# node.

Also NetworkPeerServer: Listen(int maxIncoming = 20)
max limit is actually not used. 

In bitcoin core they have a default limit of 125 connections so I suggest doing it the same.

fassadlr:
Fix #1054

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -143,7 +143,7 @@ namespace Stratis.Bitcoin.P2P.Peer
                     if (this.ConnectedNetworkPeers.Count > MaxConnectionThreshold)
                     {
                         this.logger.LogTrace("Maximum connection threshold [{0}] reached, stalling accepting of incoming connections.", MaxConnectionThreshold);
-                        await Task.Delay(MaxConnectionThresholdStallTime).ConfigureAwait(false);
+                        await Task.Delay(MaxConnectionThresholdStallTime, this.serverCancel.Token).ConfigureAwait(false);
                         continue;
                     }
 
@@ -186,7 +186,7 @@ namespace Stratis.Bitcoin.P2P.Peer
             }
             catch (Exception e)
             {
-                this.logger.LogDebug("Exception occurred: {0}", e.ToString());
+                this.logger.LogError("Exception occurred: {0}", e.ToString());
             }
 
             this.logger.LogTrace("(-)");
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -32,8 +32,14 @@ namespace Stratis.Bitcoin.P2P.Peer
         /// <summary>The parameters that will be cloned and applied for each peer connecting to <see cref="NetworkPeerServer"/>.</summary>
         public NetworkPeerConnectionParameters InboundNetworkPeerConnectionParameters { get; set; }
 
-        /// <summary>Maximal number of inbound connection that the server is willing to handle simultaneously.</summary>
-        public int MaxConnections { get; set; }
+        /// <summary>Maximum number of inbound connection that the server is willing to handle simultaneously.</summary>
+        public const int MaxConnectionThreshold = 125;
+
+        /// <summary>
+        /// The amount of time in seconds we should stall before checking if we can accept incoming connections again.
+        /// This happens once we have reached the maximum connection threshold.
+        /// </summary>
+        public const int MaxConnectionThresholdStallTime = 60;
 
         /// <summary>IP address and port, on which the server listens to incoming connections.</summary>
         public IPEndPoint LocalEndpoint { get; private set; }
@@ -79,7 +85,6 @@ namespace Stratis.Bitcoin.P2P.Peer
             this.LocalEndpoint = Utils.EnsureIPv6(localEndpoint);
             this.ExternalEndpoint = Utils.EnsureIPv6(externalEndpoint);
 
-            this.MaxConnections = 125;
             this.Network = network;
             this.Version = version;
 
@@ -103,10 +108,9 @@ namespace Stratis.Bitcoin.P2P.Peer
         /// <summary>
         /// Starts listening on the server's initialialized endpoint.
         /// </summary>
-        /// <param name="maxIncoming">Maximal number of newly connected clients waiting to be accepted.</param>
-        public void Listen(int maxIncoming = 20)
+        public void Listen()
         {
-            this.logger.LogTrace("({0}:{1})", nameof(maxIncoming), maxIncoming);
+            this.logger.LogTrace("()");
 
             try
             {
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -136,6 +140,13 @@ public void Listen(int maxIncoming = 20)
             {
                 while (!this.serverCancel.IsCancellationRequested)
                 {
+                    if (this.ConnectedNetworkPeers.Count > MaxConnectionThreshold)
+                    {
+                        this.logger.LogTrace("Maximum connection threshold [{0}] reached, stalling accepting of incoming connections.", MaxConnectionThreshold);
+                        await Task.Delay(MaxConnectionThresholdStallTime).ConfigureAwait(false);
```
dangershony: Pass the `this.serverCancel.Token` to the Task.Delay

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -177,7 +188,7 @@ public void Listen(int maxIncoming = 20)
             {
                 this.logger.LogDebug("Exception occurred: {0}", e.ToString());
             }
-            
+
```
dangershony: Inside this catch `catch (Exception e)` should be `this.logger.LogError` I would think.
And maybe even trigger a general cancellation of the node? @Aprogiena thoughts for our next skype call.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
fassadlr: Fixed.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
fassadlr: > Inside this catch catch (Exception e) should be this.logger.LogError I would think.

I have changed this to logError 👍

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -32,8 +32,14 @@ public class NetworkPeerServer : IDisposable
...
-        /// <summary>Maximal number of inbound connection that the server is willing to handle simultaneously.</summary>
-        public int MaxConnections { get; set; }
+        /// <summary>Maximum number of inbound connection that the server is willing to handle simultaneously.</summary>
+        public const int MaxConnectionThreshold = 125;
+
+        /// <summary>
+        /// The amount of time in seconds we should stall before checking if we can accept incoming connections again.
+        /// This happens once we have reached the maximum connection threshold.
+        /// </summary>
+        public const int MaxConnectionThresholdStallTime = 60;
```
Aprogiena: This feels conceptually wrong, it basically introduces a denial of service vector. As it seems that.net implementation does not allow us to reject the connection, we should accept it, check if we are okay with that, which in this case means that the number of existing connections is reasonable, but it could also mean that the incoming peer is not banned

fassadlr: @Aprogiena are you saying that we **should** connect first, then check the connection count, **then** if it is more than the maximum, disconnect?

Aprogiena: It sounds strange, but yes. It's strange because usually you would want to keep your invariants, in this is the maximum number of connections being the maximum, but sadly this is hard to do in.net as we don't have that reject. So we'll allow one more connection over that maximum.

fassadlr: OK so in this case, we will connect, create the NetworkPeer then check the limit, then disconnect? This means that we will then always connect and reconnect when the limit has been reached...

```
while (!this.serverCancel.IsCancellationRequested)
{
    // Used to record any errors occurring in the thread pool task.
    Exception error = null;

    TcpClient tcpClient = await Task.Run(() =>
    {
        try
        {
            Task<TcpClient> acceptTask = this.tcpListener.AcceptTcpClientAsync();
            acceptTask.Wait(this.serverCancel.Token);
            return acceptTask.Result;
        }
        catch (Exception exception)
        {
            // Record the error.
            error = exception;
            return null;
        }
    }).ConfigureAwait(false);

    // Raise the error.
    if (error != null)
        throw error;

    this.logger.LogTrace("Connection accepted from client '{0}'.", tcpClient.Client.RemoteEndPoint);

    NetworkPeer networkPeer = this.networkPeerFactory.CreateNetworkPeer(this.Network, tcpClient, this.CreateNetworkPeerConnectionParameters());

    if (this.ConnectedNetworkPeers.Count > MaxConnectionThreshold)
    {
        this.logger.LogTrace("Maximum connection threshold [{0}] reached, stalling accepting of incoming connections.", MaxConnectionThreshold);
        networkPeer.Disconnect("Maximum connection threshold [{0}] reached, stalling accepting of incoming connections.", MaxConnectionThreshold);

        await Task.Delay(MaxConnectionThresholdStallTime, this.serverCancel.Token).ConfigureAwait(false);
        continue;
    }

    this.ConnectedNetworkPeers.Add(networkPeer);
    networkPeer.StateChanged.Register(this.OnStateChangedAsync);

    this.AddClientConnection(networkPeer.Connection);
}
```

Aprogiena: We may not need to create network peer, we can disconnect tcp client.

dangershony: Ah and we can send a reject payload message if thats what you mean by `reject`

Aprogiena: No I didn't mean reject message from bitcoin, I meant reject socket function

Aprogiena: Asked me on Skype about the denial of service

fassadlr: @Aprogiena see latest commit... Is this what you meant?

Aprogiena: almost, except that we need to remove this
```
await Task.Delay(MaxConnectionThresholdStallTime, this.serverCancel.Token).ConfigureAwait(false);
```

and we should then modify the comment accordingly

Aprogiena:
Yes like that

fassadlr:
@dangershony we have the approval of @Aprogiena now :) OK to approve this PR?

dangershony:
Yes we can commit I would just like to have a description of the DoS attack prevention in the comments, I will skype with him and then approve this.

Aprogiena:
now you can remove MaxConnectionThresholdStallTime

dangershony:
The possibility of DoS is from a malicious peer exhausting the inbound count and sending a peer to a recurring 60 sec sleep effectively blocking new inbound connections.

PR #1142, opened 2018-02-06
Fix: BlockStore race condition

noescape00:
This race condition was happening in this scenario:
1) New block from LookaheahBlockPuller is received
2) ConsensusTip is updated 
3) Loop iteration execution starts and we bump into 
```
if (this.StoreTip.Height >= this.ChainState.ConsensusTip?.Height)
                    break;
```
and continue to block steps execution => Find blocks step tells the StoreBlockPuller to download the missing block.
4) BlockStoreSignaled gets the block that was signaled by the ConsensusLoop and it is added to the PendingStorage

Fix: #599


noescape00:
@dangershony @fassadlr can you please review this one? 

ps:
tests are passing
BlockStoreInnerStepFindBlocks => 54: AskForMultipleBlocks never gets executed (unless node was not shutted down properly and we need to recover some blocks)

PR #1143, opened 2018-02-06
Block store high cpu usage fix

noescape00:
### Why this is happening?

`ProcessPendingStorageStep.ProcessWhenInIBDAsync` will return `StepResult.Continue` if `PendingStorage` has less than 10 blocks in it. Every time `StepResult.Continue` is returned we are starting a new iteration in `BlockStoreLoop.DownloadAndStoreBlocksAsync` which will again bump into that threshold and restart until the moment 10 blocks are finally downloaded. And so it goes over and over again => the slower the peers are the longer it takes to download enough blocks to pass the threshold => more iterations we have. 

On average I have 300000 iterations of the `DownloadAndStoreBlocksAsync:while` loop before I have 30 blocks downloaded given that I'm connected to the slow peers. 

### Fix

Returning `StepResult.Stop` instead of  `StepResult.Continue` when the threshold is not met makes the loop have a 1 second break so for now this will do. 
But it's still a quick fix and in future we will need to refactor logic behind the BlockStoreLoop so we don't start a new iteration until we have a new block that we need to act upon.

ps

I've also refactored the ProcessPendingStorageStep a bit because it had 3 methods that were 80% copy paste of each other - merged them into 1.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
@@ -54,9 +54,11 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
             }
             
             // In case of IBD do not save every single block- persist them in batches.
-            if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold && !disposeMode && 
-                this.BlockStoreLoop.InitialBlockDownloadState.IsInitialBlockDownload())
-                return StepResult.Continue;
+            if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold &&
+                !disposeMode && this.BlockStoreLoop.InitialBlockDownloadState.IsInitialBlockDownload())
+            {
+                return StepResult.Stop;
+            }
 
             while (!context.CancellationToken.IsCancellationRequested)
             {
--- a/src/Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
@@ -52,95 +52,25 @@ namespace Stratis.Bitcoin.Features.BlockStore.LoopSteps
                 this.logger.LogTrace("(-)[NOT_FOUND]:{0}", StepResult.Next);
                 return StepResult.Next;
             }
-
-            if (disposeMode)
-            {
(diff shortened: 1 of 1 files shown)


noescape00:
@fassadlr @dangershony can you please take a look at this one? 

ps:
Tests are passing
Mainnet,testnet- was able to sync from fresh node and stay in sync

dangershony:
I'll review tomorrow thanks @noescape00

src/Stratis.Bitcoin.Features.BlockStore/LoopSteps/ProcessPendingStorageStep.cs
```diff
@@ -52,95 +52,27 @@ internal ProcessPendingStorageStep(BlockStoreLoop blockStoreLoop, ILoggerFactory
...
-
-            this.logger.LogTrace("(-):{0}", StepResult.Continue);
-            return StepResult.Continue;
-        }
-
-        /// <summary>
-        /// When the node is NOT in IBD, process and push the blocks in <see cref="BlockStoreLoop.PendingStorage"/> immediately
-        /// to the block repository without checking batch size.
-        /// </summary>
-        private async Task<StepResult> ProcessWhenNotInIBDAsync(ProcessPendingStorageContext context)
```
fassadlr: @noescape00 Please check the logic here as when we are NOT in IBD we need to push to the repository immediately, without checking the batch size. That is why we had the separate methods as there are subtle differences in each scenario.

Your refactor caters for processing pending storage during IBD and disposing but not for "NotInIBD".

noescape00: here 
```
if (this.BlockStoreLoop.PendingStorage.Count < BlockStoreLoop.PendingStorageBatchThreshold &&
                !disposeMode && this.BlockStoreLoop.InitialBlockDownloadState.IsInitialBlockDownload())
            {
                return StepResult.Stop;
            }
```

we will never get into the returning stop if we're not in IBD so we will proceed to the loop that will remove all the blocks that we have in pending storage and push them immediately to the repository. 

Those lines: 

```
 if (context.PendingBlockPairsToStore.Any())
                await this.PushBlocksToRepositoryAsync(context).ConfigureAwait(false);
```

Will send the blocks to the repository no matter the batch size. 

So I don't see a problem here. 
Can you please elaborate?

fassadlr: @noescape00 Actually, I just went through the logic again and the fact that we check the batch size in the while:

```
if (context.PendingStorageBatchSize > BlockStoreLoop.MaxPendingInsertBlockSize)
await this.PushBlocksToRepositoryAsync(context).ConfigureAwait(false);
```

Doesn't actually make a difference. So yeah, the logic is good 💯

Issue #1124, opened 2018-01-31
Node shows non-existing connection


noescape00:
![err](https://user-images.githubusercontent.com/10259355/35643732-337c0776-06d8-11e8-9a34-2efc28c4ce81.png)

Right after that it disconnect:

```
info: Stratis.Bitcoin.Connection.ConnectionManagerBehavior[0]
      Peer '[::ffff:52.233.33.138]:56354' offline, reason: 'unknown'.
```

This is happening regularly and that 0 height peer with the same address as the server always has a different port on every new connection.

fassadlr:
@noescape00 I'm 99.999% sure this is happening due to this code:

[](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs#L528)

Which I think is correct. We connect, then check the nonce then disconnect if its the same.

noescape00:
@fassadlr looks like it's not the case here- Nonces are different so when the bug happens `connectedToSelf` is false. Instead we have an exception during the attempt to read the magic (ReadMagicAsync => ReadBytesAsync)- the stream contains nothing when we expect it to contain 1 byte. 

At the same time server spawns `Unexpected exception while waiting for a message`. 
So it looks server shouldn't have initiated a connection in the first place (especially taking in count that server already has an inbound connection from the same peer it is trying to connect to).

So the question is: should we allow a 2 way connectivity? (I mean when there are 2 nodes- A and B. If A connects to B should the B be allowed to connect to A so both of them have 2 connections- inbound and outbound one)

I think it we want to forbid scenarios like this (and I see no reason we we wouldn't want to) we can just check on the `ProcessInitialVersionPayloadAsync` if the peer's Nonce is equal to any peer's Nonce that we are already connected to. And if it does- disconnect. 

I think I also should investigate more in depth what is the reason for the server to initiate a connection like this on a regular basis. Maybe it has something to do with the peer discovery.

noescape00:
Ok. So far I've figured out that those connections are produced by the discovery jobs. 
I'm pretty sure that DiscoverPeersAsync() doesn't work as it was intended since it tries to discover the same peer every 10 seconds. Maybe it's something to do with the `Parallel.ForEach` inside it.

noescape00:
[...]

PR #1145, opened 2018-02-07
Peer discovery fix

noescape00:
PeerDiscoveryLoop used to use Parallel.ForEach in order to perform the peer discovery. And since awaitable methods were used in the foreach iterations the whole Parallel.ForEach was not awaited (apparently Parallel.ForEach cannot be used with awaits inside the iteration code).

That caused running peer discovery for the same set of peers every 10 seconds without awaiting prev set to be processed => we get a peer discovery connection + connection drop every 10 seconds. 

In order to fix this I've introduced a `ParallelAsync.ForEachAsync` that works exactly the same as `Parallel.ForEach` except for it can work with async code properly. 

Fix: #1124

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
@@ -31,7 +31,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             Stopwatch watch = Stopwatch.StartNew();
 
-            await this.testCollection.ForEachAsync(2, CancellationToken.None, async item =>
+            await this.testCollection.ForEachAsync(2, CancellationToken.None, async (item, cancellation) =>
             {
                 await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
                 sum += item;
@@ -50,7 +50,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
 
             Stopwatch watch = Stopwatch.StartNew();
 
-            await this.testCollection.ForEachAsync(this.testCollection.Count, CancellationToken.None, async item =>
+            await this.testCollection.ForEachAsync(this.testCollection.Count, CancellationToken.None, async (item, cancellation) =>
             {
                 await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
                 sum += item;
@@ -67,7 +67,7 @@ namespace Stratis.Bitcoin.Tests.Utilities
         {
             var emptyList = new List<int>();
 
-            await emptyList.ForEachAsync(2, CancellationToken.None, async item =>
+            await emptyList.ForEachAsync(2, CancellationToken.None, async (item, cancellation) =>
             {
                 await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
             }).ConfigureAwait(false);
--- a/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
@@ -72,5 +72,25 @@ namespace Stratis.Bitcoin.Tests.Utilities
                 await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
             }).ConfigureAwait(false);
         }
+
+
+        [Fact]
+        public async void ForEachAsync_CanBeCancelled_Async()
+        {
+            var tokenSource = new CancellationTokenSource();
+
+            int itemsProcessed = 0;
+
+            await this.testCollection.ForEachAsync(1, tokenSource.Token, async (item, cancellation) =>
+            {
+                await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
+                itemsProcessed++;
+
+                if (itemsProcessed == 3)
+                    tokenSource.Cancel();
+            }).ConfigureAwait(false);
+            
+            Assert.Equal(3, itemsProcessed);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin/Utilities/ParallelAsync.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <param name="maxDegreeOfParallelism">The maximum amount of items that can be processed simultaneously.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <param name="action">Action that is used for processing each item in the collection.</param>
+        public static async Task ForEachAsync<TSource>(this IEnumerable<TSource> collection, int maxDegreeOfParallelism, CancellationToken cancellationToken, Func<TSource, Task> action)
+        {
+            var unfinished = new List<Task>();
+
+            foreach (TSource item in collection)
+            {
+                if (cancellationToken.IsCancellationRequested)
```
fassadlr: @noescape00 Correct me if I'm wrong... if we cancel here... then won't we still execut all the unfinished tasks here?

`await Task.WhenAll(unfinished).ConfigureAwait(false);`

I would assume that if the node shuts down that we stop processing tasks? In context of peer discovery, if the nodes disposes, we want to stop peer discovery immediately.

Maybe if cancellation is requested, the we just clear the unfinished list?

noescape00: >This is really cool @noescape00 and WELL SPOTTED dude! Good job

Thanks! 🙂 

>Maybe if cancellation is requested, the we just clear the unfinished list?

well, they still will be running, we just won't await them and doing that might be dangerous- we can end up disposing something and then the task executes and bumps into a disposed object. 

I think the tasks themselves should be created with the consideration of the cancellation.

noescape00: hm, maybe to make it more obvious that tasks should consider the cancellation I should make an overload that will provide the cancellation instead of making tasks using tokens from the outside.

noescape00: Done.

src/Stratis.Bitcoin/Utilities/ParallelAsync.cs
```diff
@@ -0,0 +1,42 @@
...
+                
+                if (unfinished.Count >= maxDegreeOfParallelism)
+                {
+                    await Task.WhenAny(unfinished).ConfigureAwait(false);
+
+                    foreach (Task toRemove in unfinished.Where(x => x.IsCompleted).ToList())
+                        unfinished.Remove(toRemove);
+                }
+
+                unfinished.Add(action(item, cancellationToken));
```
fassadlr: YES, this is way better. Now, in context of Peer Discovery, we would try and discover from a peer as we check the cancellation token in the delegate code.

src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Stratis.Bitcoin.Utilities;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Utilities
+{
+    public class ParallelAsyncTest
```
fassadlr: Can we maybe add test that checks the newly added cancellation token?

noescape00: Oh yeah, sure.

noescape00: Fixed.

fassadlr:
@dangershony I'm happy with this. Have also tested the branch locally. Also tested the cancellation in the loop and it works well.

If you are happy we can push 👍

src/Stratis.Bitcoin/Utilities/ParallelAsync.cs
dangershony: I wonder if we may want to catch OperationCancellation exception on the awaits (especially the `WhenAny`) as you might end up not waiting for some tasks to complete.

noescape00: WhenAny doesn't rethrow the exception so in case one of the tasks throws- it's still perfectly fine and the rest of the tasks will be awaited.

noescape00: And as for the WhenAll- here I can use a foreach loop that will await each task in the list and if there is an exception- it will be ignored. 
Would you like me to do that?

src/Stratis.Bitcoin/Utilities/ParallelAsync.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Stratis.Bitcoin.Utilities
+{
+    public static class ParallelAsync
+    {
+        /// <summary>
+        /// Executes a foreach operation on an IEnumerable in which iterations run in parallel.
```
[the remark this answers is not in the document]
noescape00: Right, thanks.
Fixed.

src/Stratis.Bitcoin/Utilities/ParallelAsync.cs
dangershony: No that's fine as the WhenAll will wait for all tasks to complete.

noescape00: Ah, yes, you're right. Missed that 🙂

PR #1148, opened 2018-02-07
DNS seeder to use LastSeen instead of LastConnectionHandshake

dangershony:
@345paul can you please take a look as well?


345paul:
Looks good Dan.

PR #1149, opened 2018-02-08
Tests consistency on Travis

noescape00:
Removed StopWatch usages from the ParallelAsync tests & changed timings for AsyncLoopRepeatEveryIntervalCanBeChangedWhileRunningAsync & fixed race condition.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
@@ -31,8 +31,8 @@ namespace Stratis.Bitcoin.Tests.Utilities
         public async void ForEachAsync_TestDegreeOfParallelism_Async()
         {
             int sum = 0;
-
-            var delayTask = new Task(async () => { await Task.Delay(this.testCollection.Count * this.itemProcessingDelayMs).ConfigureAwait(false); });
+            
+            Stopwatch watch = Stopwatch.StartNew();
 
             await this.testCollection.ForEachAsync(2, CancellationToken.None, async (item, cancellation) =>
             {
@@ -40,7 +40,8 @@ namespace Stratis.Bitcoin.Tests.Utilities
                 await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
             }).ConfigureAwait(false);
 
-            Assert.False(delayTask.IsCompleted);
+            watch.Stop();
+            Assert.True(watch.Elapsed.TotalMilliseconds < this.testCollection.Count * this.itemProcessingDelayMs);
             Assert.Equal(this.testCollectionSum, sum);
         }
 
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
```diff
@@ -29,36 +29,33 @@ public ParallelAsyncTest()
         {
             int sum = 0;
 
-            Stopwatch watch = Stopwatch.StartNew();
+            var delayTask = new Task(async () => { await Task.Delay(this.testCollection.Count * this.itemProcessingDelayMs / 2).ConfigureAwait(false); });
 
             await this.testCollection.ForEachAsync(2, CancellationToken.None, async (item, cancellation) =>
             {
-                await Task.Delay(this.itemProcessingDelayMs).ConfigureAwait(false);
                 sum += item;
```
Aprogiena: You've got race condition here

noescape00: That's basically the intention: create a race condition where we expect something to finish first instead of using time measurements with StopWatch

Aprogiena: Unlikely, the test will fail if the race condition happens, you are asserting that the sum you are computing will equal to the sum of all items

Aprogiena: There is nothing wrong with the stopwatch, the whole test is bogus and race condition is just an extra reason like fail

the test will fail quite likely the machine with single processor and single thread execution and on multi-core machine the sample should be much larger in order for the test to have only very low chance of failing

noescape00: How is that possible if the sum is updated before the awaiting of the Task.Delay? There should be no way we execute `Assert.Equal(this.testCollectionSum, sum);` before we summarize all items.

noescape00: Though the race condition can happen in ForEachAsync_CanBeCancelled_Async, that I'll fix 🙂

Aprogiena: `sum += item` Can be executed by 2 threads in parallel, at least I believe that is the case, is it?

Aprogiena: Although the delay in the task is too low, so that's just another point where this test can fail

noescape00: >sum += item Can be executed by 2 threads in parallel, at least I believe that is the case, is it?

correct, it can.

noescape00: Still don't see where the race condition can happen which will lead to execution `Assert.Equal(this.testCollectionSum, sum);` before all the items from the list were summarized. 

Maybe you are talking about prev version where there was delay first and after that sum update? If that's the case- yes, race here can happen and that is fixed in the 1st commit of that PR.

Aprogiena: `sum += item` can be skipped due to race condition

noescape00: Fixed

src/Stratis.Bitcoin.Tests/Utilities/ParallelAsyncTest.cs
```diff
@@ -29,36 +32,33 @@ public ParallelAsyncTest()
         {
             int sum = 0;
 
-            Stopwatch watch = Stopwatch.StartNew();
+            var delayTask = new Task(async () => { await Task.Delay(this.testCollection.Count * this.itemProcessingDelayMs).ConfigureAwait(false); });
```
dangershony: This is a very weird test dude.

dangershony: If apro approves it then ok, but you could perhaps take a different approach. using a ManualResetEvent one tasks blocks it the other releases that will prove degree of parallelism .

Aprogiena: Me does not approve this, because me thinks that stopwatch is okay here, there were many other problems that are now hopefully fixed but I haven't checked carefully, but as it is just test, I am not blocking it

noescape00: >This is a very weird test dude.

Well, yeah. 
Returned the stopwatch (but now there is a larger delay so should be ok on slow environments like travis).

PR #1150, opened 2018-02-08
Added custom fees

bokobza:
Added the option to set custom fees when building a transaction.
If both feeAmount and feeType are set, feeAmount is taken into account.
If none is set, the feeType defaults to "medium".

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
@@ -1268,6 +1268,66 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
```diff
@@ -1267,6 +1267,66 @@ public void BuildTransactionWithValidRequestAllowingUnconfirmedReturnsWalletBuil
             Assert.Equal(sentTrx.GetHash(), model.TransactionId);
         }
 
+        [Fact]
+        public void BuildTransactionWithCustomFeeAmountAndFeeTypeReturnsWalletBuildTransactionModelWithFeeAmount()
```
dangershony: Can you add underscores in the method name, its not easy to read

bokobza: I'm going to leave it like this for now as the other tests in the class don't have underscores, but we'll change when we revisit naming of tests.

Issue #826, opened 2017-11-29
[FullNode] enum for node states instead of bunch of bool variables. 


noescape00:
I think it will be a good idea to have Node's state in single enum instead of several bool variables (Stopped, IsDisposed, HasExited) that don't even cover all states.

```
    public enum FullNodeState
    {
        Created,
        Initialized,
        Starting,
        Started,
        Stopping,
        Stopped,
        Disposing,
        Exited
    }
```

dangershony:
I would wait with this while @Aprogiena is refactoring the node code

noescape00:
Ok, totally agree.

PR #1154, opened 2018-02-10
Node state enum

noescape00:
State enum instead of flags. 

Fix: #826

### DIFF
--- a/src/Stratis.Bitcoin/IFullNode.cs
+++ b/src/Stratis.Bitcoin/IFullNode.cs
@@ -37,6 +37,13 @@ namespace Stratis.Bitcoin
         void Start();
 
         /// <summary>
+        /// Initializes DI services that the node needs.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin/IFullNode.cs
```diff
@@ -49,4 +52,15 @@ public interface IFullNode : IDisposable
         /// <returns></returns>
         T NodeFeature<T>(bool failWithError = false);
     }
+
+    public enum FullNodeState
```
Aprogiena: This code could be much happier, if there was a description on which state means what and how the node changes from one state to another et cetera

noescape00: Isn't is too obvious to write down? State names are very self explanatory

noescape00: Fixed.

Issue #1141, opened 2018-02-06
[shutdown] TCP bug when disposing the node


dangershony:
```
info: Stratis.Bitcoin.FullNode[0]
      Waiting for DiscoverPeersAsync to finish.
info: Stratis.Bitcoin.FullNode[0]
      Waiting for PeerConnectorAddNode.ConnectAsync to finish.
^C
C:\Users\fullnodedan\Documents\Github\StratisBitcoinFullNode\src\Stratis.StratisD>info: Stratis.Bitcoin.FullNode[0]
      PeerConnectorAddNode.ConnectAsync stopping
info: Stratis.Bitcoin.FullNode[0]
      Waiting for PeerConnectorDiscovery.ConnectAsync to finish.
fail: Stratis.Bitcoin.P2P.Peer.NetworkPeerServer[0]
      [0.0.0.0:26178] Exception occurred: System.AggregateException: One or more errors occurred. (Cannot access a disposed object.
      Object name: 'System.Net.Sockets.Socket'.) ---> System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'System.Net.Sockets.Socket'.
         at System.Net.Sockets.Socket.EndAccept(Byte[]& buffer, Int32& bytesTransferred, IAsyncResult asyncResult)
         at System.Net.Sockets.TcpListener.EndAcceptTcpClient(IAsyncResult asyncResult)
         at System.Net.Sockets.TcpListener.<>c.<AcceptTcpClientAsync>b__28_1(IAsyncResult asyncResult)
         at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
         --- End of inner exception stack trace ---
         at Stratis.Bitcoin.P2P.Peer.NetworkPeerServer.<AcceptClientsAsync>d__34.MoveNext() in C:\Users\fullnodedan\Documents\Github\StratisBitcoinFullNode\src\Stratis.Bitcoin\P2P\Peer\NetworkPeerServ
er.cs:line 160
      ---> (Inner Exception #0) System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'System.Net.Sockets.Socket'.
         at System.Net.Sockets.Socket.EndAccept(Byte[]& buffer, Int32& bytesTransferred, IAsyncResult asyncResult)
[...]

PR #1157, opened 2018-02-12
Appropriate level of exceptions handling in NetworkPeerServer

noescape00:
Error on a lower level was handled using LogError while handling was also implemented on a higher level with LogTrace. 
Removed low level handling with LogError. 

Fix: #1141

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -180,7 +180,7 @@ namespace Stratis.Bitcoin.P2P.Peer
             }
             catch (Exception e)
             {
-                this.logger.LogError("Exception occurred: {0}", e.ToString());
+                this.logger.LogDebug("Exception occurred: {0}", e.ToString());
             }
 
             this.logger.LogTrace("(-)");
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorDiscovery.cs
@@ -128,7 +128,7 @@ namespace Stratis.Bitcoin.P2P
                 await Task.Delay(2000, this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
             }
             else
-                await ConnectAsync(peer).ConfigureAwait(false);
+                await this.ConnectAsync(peer).ConfigureAwait(false);
         }
     }
 }


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -178,10 +178,6 @@ public void Listen()
             {
                 this.logger.LogDebug("Shutdown detected, stop accepting connections.");
             }
-            catch (Exception e)
-            {
-                this.logger.LogError("Exception occurred: {0}", e.ToString());
-            }
```
Aprogiena: wtf? revert this back to LogDebug as it was before
this change would crash node on shutdown

noescape00: Oh, right. 
Didn't see that we do .Wait() on it without handling

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -180,7 +180,7 @@ public void Listen()
             }
             catch (Exception e)
             {
-                this.logger.LogError("Exception occurred: {0}", e.ToString());
+                this.logger.LogDebug("Exception occurred: {0}", e.ToString());
```
dangershony: Just a quick explanation why do we expect a disposed exception?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
dangershony: >this change would crash node on shutdown  

Can you explain why?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
dangershony: What about catching an ObjectDisposedException?

Aprogiena: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1141#issuecomment-365255037

Aprogiena: See that comment. We can probably catch that specific exception, But the problem that the documentation of the function - https://msdn.microsoft.com/en-us/library/system.net.sockets.tcplistener.accepttcpclientasync%28v=vs.110%29.aspx - doesn't mention it, probably because it is just a side effect of closing the socket and not a feature of this specific method and anyway ask whether or not there can be another side effect leading to slightly different exception type and you have no guarantees that if you catch just one type that another type can be thrown because of the very same thing again

Issue #605, opened 2017-10-17
[Settings] DefaultMaxTipAge is 1 day, is it not too much?


Aprogiena:
There are couple of rules for consensus to consider itself in IBD, the last one is checking MaxTipAge, which default value is 24 hours. It seems quite a lot and the question is, for the purpose of considering itself as in IBD, is 24 hours value justified? 

On Bitcoin network with just 144 blocks per day, this could make some sense, but on Stratis network with 1350 block per day, it seems that a lower value would make better sense. 

However, currently, we use one value regardless the network, so we would need to make this part of consensus options if we wanted to preserve 24 hours for Bitcoin and have separated value for Stratis.

bokobza:
I agree that we could in theory lower the value of DefaultMaxTipAge as we have many more blocks per day.
Maybe 12h, 6h, a tenth of 24h? How do we fix that? @Aprogiena @dangershony thoughts?

In any case I think it's probably good to split the values for the different networks.
I'll look to do this with https://trello.com/c/VdIJjjtt/55-nodesettings-use-a-different-way-to-add-settings-to-the-node

Aprogiena:
The value should be different for each network. So once it is part of network options it is easy to say 24h for Bitcoin and I'd say 1 hour for Stratis should be very safe valye

dangershony:
Agreed, we need to put it in Network (or ConsensusOptions)

PR #1158, opened 2018-02-12
Move Max tip age to Network

bokobza:
This is for https://github.com/stratisproject/StratisBitcoinFullNode/issues/605
Should we change the default max tip age value for Stratis?
If yes, to what?
I think 2 hours is ok (~120 blocks).
@dangershony @Aprogiena

### DIFF
--- a/src/NBitcoin/Networks.cs
+++ b/src/NBitcoin/Networks.cs
@@ -26,6 +26,9 @@ namespace NBitcoin
             Block.BlockSignature = saveSig;
         }
 
+        /// <summary> Default value for Maximum tip age in seconds to consider node in initial block download. </summary>
+        public const int DefaultMaxTipAge = 24 * 60 * 60;
+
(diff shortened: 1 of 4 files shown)


Aprogiena:
1 hour or 2 hours should be fine

src/NBitcoin/Networks.cs
```diff
@@ -26,6 +26,9 @@ static Network()
             Block.BlockSignature = saveSig;
         }
 
+        /// <summary> Default value for Maximum tip age in seconds to consider node in initial block download. </summary>
+        public const int DefaultMaxTipAge = 24 * 60 * 60;
```
dangershony: I would suggest to add the time type to the name `DefaultMaxTipAgeSeconds`

dangershony:
Yeah sounds fine to me 2 hours.

PR #1161, opened 2018-02-13
 Removing the ContextualCheckBlock method from the validator

dangershony:
Some rules have already been implemented this code will remove them from the validator class.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -256,7 +256,7 @@ namespace Stratis.Bitcoin.Features.Consensus
                     new BlockHeaderPowContextualRule(),
 
                     // rules that are inside the method ContextualCheckBlock
-                    new Bip113ActivationRule(),
+                    new TransactionLocktimeActivationRule(),
                     new HeightInCoinbaseActivationRule(),
                     new WitnessCommitmentsRule(),
                     new BlockSizeRule(),
@@ -288,7 +288,7 @@ namespace Stratis.Bitcoin.Features.Consensus
                     new BlockHeaderPosContextualRule(),
 
                     // rules that are inside the method ContextualCheckBlock
-                    new Bip113ActivationRule(),
+                    new TransactionLocktimeActivationRule(),
                     new HeightInCoinbaseActivationRule(), 
                     new WitnessCommitmentsRule(),
                     new BlockSizeRule(),
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -257,7 +257,7 @@ namespace Stratis.Bitcoin.Features.Consensus
 
                     // rules that are inside the method ContextualCheckBlock
                     new Bip113ActivationRule(),
-                    new Bip34ActivationRule(),
+                    new HeightInCoinbaseActivationRule(),
                     new WitnessCommitmentsRule(),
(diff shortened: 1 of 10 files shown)


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Threading.Tasks;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
+{
+    /// <summary>
+    /// Context checks on a POS block.
+    /// </summary>
+    public class PosBlockContextRule : PosConsensusRule
```
dangershony: @Aprogiena so you are aware of this code (I guess its a vulnerability)

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -256,8 +256,8 @@ public IEnumerable<ConsensusRule> GetRules()
                     new BlockHeaderPowContextualRule(),
 
                     // rules that are inside the method ContextualCheckBlock
-                    new Bip113ActivationRule(),
-                    new Bip34ActivationRule(),
+                    new TransactionLocktimeActivationRule(),
+                    new CoinbaseHeighActivationRule(),
```
Aprogiena: If you rename it like that, you should make sure that we don't lose information that it is related to those BIP, so please with these lines add  a comment that says that the first one is for BIP 113 and the 2nd one is for 34

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
Aprogiena: Do you mean that block flood attack or something else?

Aprogiena: If so, then yes I have it my list

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
dangershony: yeah ok good idea, the BIP information is also present in the rule description but its a good practice to put some explanation next to each rule, agree!

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        /// <inheritdoc />
+        /// <exception cref="ConsensusErrors.HighHash"> Thrown if block doesn't have a valid PoW header.</exception>
```
noescape00: valid **PoS** header

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
noescape00: I think it will be a good idea to also add this info to the rule class summaries

dangershony: Yes it will when we implement it.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
dangershony: arghh... good spot.

Issue #1097, opened 2018-01-24
Node time not in sync with the network, display better warring message


dangershony:
My blocks got rejected when my node was an hour behind, we should display this better on the console and the UI (same as X).

```
fail: Stratis.Bitcoin.Features.Consensus.ConsensusLoop[0]
      Block 'e0e61759872e112d67306a9cdd027953ffe4069027931d607cd2e82119bb9688' rejected: block timestamp too far in the future
fail: Stratis.Bitcoin.Features.Consensus.ConsensusLoop[0]
      Marking block 'e0e61759872e112d67306a9cdd027953ffe4069027931d607cd2e82119bb9688' as invalid until 2018-01-19 03:53:04.
fail: Stratis.Bitcoin.Features.Consensus.ConsensusLoop[0]
      Block 'e0e61759872e112d67306a9cdd027953ffe4069027931d607cd2e82119bb9688' rejected: block timestamp too far in the future
fail: Stratis.Bitcoin.Features.Consensus.ConsensusLoop[0]
      Marking block 'e0e61759872e112d67306a9cdd027953ffe4069027931d607cd2e82119bb9688' as invalid until 2018-01-19 03:53:04.
info: Stratis.Bitcoin.Features.BlockStore.BlockStoreLoop[0]
      ====BlockStore Repository Stats(%)====
      Hit Count:          2
```

D3m0nKingx:
Seeing the same thing on Stratis Mainnet:

```bash
node.txt:[2018-01-24 09:04:33.2384 126] ERROR: Stratis.Bitcoin.Features.Consensus.ConsensusLoop.AcceptBlockAsync Block '29947af25f6603d72b14c06cba809ec55d4f04434a7a6538471d4029b5a7ed46' rejected: block timestamp too far in the future
node.txt:[2018-01-24 09:04:33.2384 126] ERROR: Stratis.Bitcoin.Features.Consensus.ConsensusLoop.AcceptBlockAsync Marking block '29947af25f6603d72b14c06cba809ec55d4f04434a7a6538471d4029b5a7ed46' as invalid until 2018-01-19 01:04:48.
```

dangershony:
You need to fix your computers time stamp.

To reproduce this just set the node local time one hour backwards

D3m0nKingx:
```      Local time: Wed 2018-01-24 14:21:24 CET
  Universal time: Wed 2018-01-24 13:21:24 UTC
[...]

PR #1163, opened 2018-02-14
Time sync warnings

bokobza:
* Better warning messages when system time out of sync
* Added method to explicitly states that system time is out of sync
* Prevent staking if system time is out of sync.
  Show errors in API, RPC staking calls as well as when starting the node with -stake=true

For https://github.com/stratisproject/StratisBitcoinFullNode/issues/1097

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -261,6 +261,9 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <summary>Time in milliseconds between attempts to generate PoS blocks.</summary>
         private readonly int minerSleep;
 
+        /// <summary>Time in milliseconds between attempts to generate PoS blocks, when the system time is out of sync.</summary>
+        private readonly int systemTimeOutOfSyncSleep;
+
         /// <summary>A lock for managing asynchronous access to memory pool.</summary>
         protected readonly MempoolSchedulerLock mempoolLock;
 
@@ -365,6 +368,7 @@ namespace Stratis.Bitcoin.Features.Miner
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
 
             this.minerSleep = 500; // GetArg("-minersleep", 500);
+            this.systemTimeOutOfSyncSleep = 7000;
             this.lastCoinStakeSearchTime = this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
             this.lastCoinStakeSearchPrevBlockHash = 0;
             this.targetReserveBalance = 0; // TOOD:settings.targetReserveBalance
@@ -469,7 +473,7 @@ namespace Stratis.Bitcoin.Features.Miner
                 {
                     this.logger.LogError("Staking cannot start, your system time does not match that of other nodes on the network." + Environment.NewLine
                                          + "Please adjust your system time and restart the node.");
-                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.systemTimeOutOfSyncSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
                     continue;
                 }
 
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -303,6 +303,9 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <summary>Provider of IBD state.</summary>
         private IInitialBlockDownloadState initialBlockDownloadState;
 
+        /// <summary>State of time synchronization feature that stores collected data samples.</summary>
+        private readonly ITimeSyncBehaviorState timeSyncBehaviorState;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="PosMinting"/> class.
         /// </summary>
@@ -321,6 +324,7 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <param name="mempool">Memory pool of pending transactions.</param>
         /// <param name="wallet">A manager providing operations on wallets.</param>
         /// <param name="asyncLoopFactory">Factory for creating background async loop tasks.</param>
+        /// <param name="timeSyncBehaviorState">State of time synchronization feature that stores collected data samples.</param>
         /// <param name="loggerFactory">Factory for creating loggers.</param>
         public PosMinting(
             IConsensusLoop consensusLoop,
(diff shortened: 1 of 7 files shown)


src/Stratis.Bitcoin.Features.RPC/RPCMiddleware.cs
```diff
@@ -54,6 +55,12 @@ public RPCMiddleware(RequestDelegate next, IRPCAuthorization authorization, ILog
                 httpContext.Response.ContentType = "application/json";
                 await httpContext.Response.WriteAsync(response.ToString(Formatting.Indented));
             }
+            else if (ex is ConfigurationException)
```
dangershony: A suggested improvement is to use a switch case on each type (like we do in some of the behaviours)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -458,6 +464,15 @@ public void StopStake()
 
             while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
             {
+                // Prevent mining if the system time is not in sync with that of other members on the network.
+                if (this.timeSyncBehaviorState.IsSystemTimeOutOfSync())
+                {
+                    this.logger.LogError("Staking cannot start, your system time does not match that of other nodes on the network." + Environment.NewLine
+                                         + "Please adjust your system time and restart the node.");
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                    continue;
```
Aprogiena: I don't think this is correct. What this does is that every half a second it displays an error on a console. That seems like overkill and it will be mixed with those other messages that you display from other components so it will be quite messy. Instead of this continue maybe you would like to Implement some kind of mechanism that would actually terminate stakingLoop , Or maybe at least just wait a lot longer than half a second

src/Stratis.Bitcoin.Features.RPC/RPCMiddleware.cs
Aprogiena: +1

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
```diff
@@ -27,6 +27,12 @@ public interface ITimeSyncBehaviorState
         /// <c>false</c> if the sample comes from a peer that our node connected to.</param>
         /// <returns><c>true</c> if the sample was added to the mix, <c>false</c> otherwise.</returns>
         bool AddTimeData(IPAddress peerAddress, TimeSpan offsetSample, bool isInboundConnection);
+
+        /// <summary>
+        /// Gets a value indicating whether the system time is not in sync and needs adjustment.
+        /// </summary>
+        /// <returns><true/> if the system time is not in sync and needs adjustment, <c>false</c> otherwise.</returns>
```
Aprogiena: `<c>true</c>`

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
```diff
@@ -27,6 +27,12 @@ public interface ITimeSyncBehaviorState
...
+        bool IsSystemTimeOutOfSync();
```
Aprogiena: But while the property with public get

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
```diff
@@ -167,6 +173,9 @@ public struct TimestampOffsetSample
         /// otherwise the node may not perform correctly on the network.</summary>
         private IAsyncLoop warningLoop;
 
+        /// <summary> A value indicating whether the system time is not in sync and needs adjustment. </summary>
+        private bool isSystemTimeOutOfSync;
```
Aprogiena: why not using `WarningLoopStarted`  that is already there doing exactly this?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
bokobza: That's the intended purpose: making it really clear to the user that his node isn't staking. 
If we just stop staking and show a message that says as much, this message will quickly disappear amidst the other logs.
We could extend it to be every 5 seconds I guess.

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
bokobza: fixed

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
bokobza: Fixed. I can't believe I created such a method.

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
bokobza: While they're being set up at the same time, I don't like to rely on a flag showing something that's a result of a certain state. 
Here, the system is out of sync, therefore the loop starts. I don't want to look at whether the loop has started to conclude that we're out of sync. I'd rather have another flag that explicitly tells me that.

Aprogiena: Then you should probably just rename `WarningLoopStarted` to `IsSystemTimeOutOfSync`, Because that was the purpose of that property, now we have 2 things doing absolutely the same, and one is not used, so maybe just remove it

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: There are other things happening every 5 seconds so you probably want to have something different maybe 7 seconds, maybe 11 seconds, not 5, and no less than that

src/Stratis.Bitcoin/Base/TimeSyncBehavior.cs
bokobza: Good idea :-)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
bokobza: gotcha

src/Stratis.Bitcoin.Features.RPC/RPCMiddleware.cs
Aprogiena: It especially doesn't look nice when a lot of code is repeated

why don't We have
```
httpContext.Response.ContentType = "application/json";
await httpContext.Response.WriteAsync(response.ToString(Formatting.Indented));
```
These 2 guys just once?

bokobza: I agree. But as far as the switch is concerned, I don't think it's an improvement.

Aprogiena: It should be catch clause not switch anyway, but whatever, feel free to leave it for this PR

bokobza:
@Aprogiena did you approve this?

Aprogiena:
I think it is okay and it can go, but note that my current reviews are superficial, so don't rely on me, but otherwise it looks good

bokobza:
Thanks!

PR #1164, opened 2018-02-14
Network Peer Behaviors minor refactoring

noescape00:
This PR simplifies the way behaviors receive payloads from the peer. 

Old way: 
1. Subscribe to all messages received by peer
2. Make a wrapper that would handle exceptions 
3. use switch() to get only those types of payloads that we're interested in
4. Finally process the particular payload.

New way: 
Subscribe for a particular payload with method that is supposed to process it. 

img


dangershony:
Minor refactor you say 🤔 😄

noescape00:
>Minor refactor you say

You'll only need to review changes in NetworkPeerBehavior. Everything else use new tools added there and all changes follow the same pattern 🙂

Aprogiena:
As I mentioned elsewhere, I don't like it

noescape00:
>As I mentioned elsewhere, I don't like it

you did here: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1090

Those problems (most of them) that were mentioned during the skype call will be fixed in the next commit.

Aprogiena:
Besides those things that you said you were fix, I don't like the new design of this, but I leave this to others to review if they are happy with it

dangershony:
I will review this next I didn't forget @noescape00

src/Stratis.Bitcoin/P2P/Protocol/Behaviors/NetworkPeerBehavior.cs
```diff
@@ -37,16 +50,75 @@ public void Attach(INetworkPeer peer)
...
 
+        private async Task OnMessageReceivedAsync(INetworkPeer peer, IncomingMessage message)
+        {
+            Type payloadType = message.Message.Payload.GetType();
+
+            foreach (var subscription in this.subscriptions.Where(x => x.Key == payloadType))
+                await subscription.Value(peer, message.Message.Payload, message.Length).ConfigureAwait(false);
+        }
+
+        protected void SubscribeToPayload<T>(OnPayloadReceivedCompact<T> callback) where T : Payload
```
dangershony: You can add a `OnPayloadReceivedCompact` delegate to an ` Dictionary<Type, OnPayloadReceived<object>>`?

noescape00: Yeah, sure. Casting to `object` is probably the only way to keep different generic types in one structure.

noescape00: Ah, I probably misunderstood the question. 
It's not possible because those are different types BUT we have a wrapper: 

```
this.subscriptions.Add(typeof(T), (peer, payload, lenght) => callback(peer, payload as T));
```

So here we actually create a OnPayloadReceived which calls OnPayloadReceivedCompact (which is the same but without the `length` which is just ignored)

And this wrapper OnPayloadReceived is added to the `Dictionary<Type, OnPayloadReceived<object>>`

noescape00: This looks like a good architectural decision to me but if you find it over-engineered or ugly I don't mind having only `OnPayloadReceived` and implementing a `length` input parameter for every callback which will just be ignored in most cases.

bokobza:
@noescape00 @Aprogiena @dangershony what do we do?

noescape00:
@bokobza I'm fine with discarding it since there are more important tasks to do. 

But if you think that it's a good idea to subscribe to a particular payload instead of to all messages and then sorting them out and doing copy-pasted handling of that- then I'll update/rework this pr.

bokobza:
Ok thanks guys :-)

PR #1165, opened 2018-02-14
 Remove any validation from the validator, its all done in rule now.

dangershony:


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -7,30 +7,29 @@
 
 namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
 {
+    /// <summary>
+    /// Check for merkle tree malleability (CVE-2012-2459): repeating sequences
+    /// of transactions in a block without affecting the merkle root of a block,
+    /// while still invalidating it.
+    /// </summary>
```
Aprogiena: This rule does more than that, the primary intention of this is to check if the merkle root  is actually correctly calculated and only then it checks for that vulnerability.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -7,30 +7,29 @@
...
-                Block block = context.BlockValidationContext.Block;
+            Block block = context.BlockValidationContext.Block;
 
-                uint256 hashMerkleRoot2 = BlockMerkleRoot(block, out bool mutated);
-                if (block.Header.HashMerkleRoot != hashMerkleRoot2)
-                {
-                    this.Logger.LogTrace("(-)[BAD_MERKLE_ROOT]");
-                    ConsensusErrors.BadMerkleRoot.Throw();
-                }
+            uint256 hashMerkleRoot2 = BlockMerkleRoot(block, out bool mutated);
```
Aprogiena: Are you sure you should not check context.CheckMerkleRoot before you calculate anything and apply.for both of the rules not only for checking whether the value was calculated correctly

Aprogiena: By the way why are those calculation methods static and public

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (BlockStake.IsProofOfWork(block))
+            {
+                bool res = block.BlockSignatur.IsEmpty();
+                this.Logger.LogTrace("(-)[POW]:{0}", res);
+                return res;
+            }
+
+            if (block.BlockSignatur.IsEmpty())
+            {
+                this.Logger.LogTrace("(-)[EMPTY]:false");
```
Aprogiena: All these logs are exit logs, they should not be because you don't have any entry logs, in general logging in this part of the coded cares about rules is not good

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
dangershony: I am not too sure but I realised I copied it wrongly form the source when I came to delete the source 
https://github.com/stratisproject/StratisBitcoinFullNode/pull/1165/files#diff-8e73c7145e2315ef48d28613641256baL389

But come to think about it we probably dont need to do this check if the flag is false (so it was wrong before)

dangershony: Ah but don't mistake the `context.CheckMerkleRoot` with the checkpoint validation indicator, this rule `BlockMerkleRootRule` is a validation rule so if we are under checkpoint/assumvalid it will not be called.

dangershony: >By the way why are those calculation methods static and public

The logic in the method `ComputeMerkleRoot` is called from another Rule

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
dangershony: Yeah its a good question about logging, do we log inside rules?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
Aprogiena: I am not mistaking CMR with Skip validation, it is that there are 2 ways of checking Merkle root, and only one is skipped

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
Aprogiena: We can definitely look inside rules, but but if you have an exit log should have an entry log. Also you would like to have instance logger and not to pass it from another class

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
dangershony: Yeah I will add the second check I think this will be correct.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CheckPowTransactionRule.cs
```diff
@@ -5,7 +5,19 @@
...
+    /// <exception cref="ConsensusErrors.BadTransactionNoOutput">Thrown if transaction has no outputs.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionOversize">Thrown if transaction size is greater than maximum allowed size of a block.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionNegativeOutput">Thrown if at least one transaction output has negative value.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionTooLargeOutput">Thrown if at least one transaction output value is greater than maximum allowed one.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionTooLargeTotalOutput">Thrown if sum of all transaction outputs is greater than maximum allowed one.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionDuplicateInputs">Thrown if any of transaction inputs are duplicate.</exception>
+    /// <exception cref="ConsensusErrors.BadCoinbaseSize">Thrown if coinbase transaction is too small or too big.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionNullPrevout">Thrown if transaction contains a null prevout.</exception>
```
noescape00: I think this should be moved to the RunAsync's comment since it's not the class that can throw those exceptions but the inner method.

Aprogiena: Absolutely

dangershony: Fair point

Issue #868, opened 2017-12-08
[Tests] NBitcoin tests that currently fail


mikedennis:
These ones fail all the time (are from RPC Client Tests and REST Client Tests):

- CanGetBlockHeader
- CanGetRawMemPool
- CanGetTransaction
- CanGetTxOutAsyncFromRPC
- CanGetTxOutFromRPC
- CanGetUTXOsMempool
- CanUseAsyncRPC
- EstimateFeeRate
- TestFundRawTransaction

These ones I was able to get to pass by running alone but failed originally when running with full batch:

- CheckBlockProofOfStake
- IndexTheFullChain
- RawTransactionIsConformsToRPC

mikedennis:
Some of the RPC test failures appear to be related to code commented out in this commit https://github.com/stratisproject/StratisBitcoinFullNode/pull/794

For example CanGetTxOutFromRPC needs this code https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin.Tests/NodeBuilder.cs#L538

dangershony:
If this is because of the removal of Node then perhaps @Aprogiena can comment, NetworkPeer (previously Node) is now much more flexible for testing as it makes more uses Interfaces.
@Neurosploit might also be interested in this.

Neurosploit:
Thanks for the heads up @dangershony.

Aprogiena:
interfaces on NP will be implemented soon (TM)

bokobza:
All good now :-)

PR #1170, opened 2018-02-27
Nbitcoin tests

bokobza:
Fixed 4 NBitcoin test out of 7.
The 3 that are still failing are:
CanGetBlockHeader
CanGetUTXOsMempool
CanGetTransaction
in the class `RestClientTests`.
Not really sure what to do with them as the code they rely on has been moved to NetworkPeer and is no longer working.
So I'm creating this PR in the meantime.
Maybe @Aprogiena or @fassadlr can give us some insight as to the best way forward.

### DIFF
--- a/src/NBitcoin.Tests/ColoredCoinsTests.cs
+++ b/src/NBitcoin.Tests/ColoredCoinsTests.cs
@@ -110,7 +110,7 @@ namespace NBitcoin.Tests
         [Trait("UnitTest", "UnitTest")]
         public void CanParseColoredAddress()
         {
-            var address = new BitcoinPubKeyAddress("16UwLL9Risc3QfPqBUvKofHmBQ7wMtjvM");
+            var address = new BitcoinPubKeyAddress("16UwLL9Risc3QfPqBUvKofHmBQ7wMtjvM", Network.Main);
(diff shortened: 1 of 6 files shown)


Aprogiena:
@bokobza 

> CanGetBlockHeader

This test seems to check that "headers" REST API request works - https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#blockheaders

> CanGetUTXOsMempool

This is again about rest API, this time it is https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#query-utxo-set

> CanGetTransaction
And this one is https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md#transactions


These tests are not testing C# node functionality. I believe we do not implement rest API. These tests run bitcoin core node and the test implements client side for rest API.

So it depends whether we are interested to have this rest API client, in which case we might be interested to fix those test. But I don't believe that is the case. So unless we are interested in rest API, it is okay to just delete the whole file with these tests.

bokobza:
Thanks @Aprogiena

bokobza:
https://github.com/stratisproject/StratisBitcoinFullNode/issues/868

PR #1187, opened 2018-03-07
Network Peer Server refactoring

noescape00:
This is first PR in series of PRs that will lead to fixing https://github.com/stratisproject/StratisBitcoinFullNode/issues/1090

In this part NetworkPeerCollection was removed from the NetworkPeerServer

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
@@ -154,12 +155,12 @@ namespace Stratis.Bitcoin.P2P.Peer
                         tcpClient.Close();
                         continue;
                     }
-                    
+
                     this.logger.LogTrace("Connection accepted from client '{0}'.", tcpClient.Client.RemoteEndPoint);
 
                     INetworkPeer networkPeer = this.networkPeerFactory.CreateNetworkPeer(tcpClient, this.CreateNetworkPeerConnectionParameters());
 
-                    this.peersById.AddOrReplace(networkPeer.Connection.Id, networkPeer);
+                    this.peersById.TryAdd(networkPeer.Connection.Id, networkPeer);
 
                     networkPeer.StateChanged.Register(this.OnStateChangedAsync);
                 }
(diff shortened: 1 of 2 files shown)


Aprogiena:
I have checked the code, it looks okay to me, but I have not done any tests. I'm not approving this just because of that. If anyone can run the tests to confirm that it does not break anything that I'm happy for this to be merged.

dangershony:
I will run the tests now

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -229,7 +192,8 @@ private Task OnStateChangedAsync(INetworkPeer peer, NetworkPeerState oldState)
                 || (peer.State == NetworkPeerState.Offline))
             {
                 peer.StateChanged.Unregister(this.OnStateChangedAsync);
-                this.ConnectedNetworkPeers.Remove(peer, "Peer disconnected");
+                this.peersById.TryRemove(peer.Connection.Id, out INetworkPeer unused);
+                peer.Dispose("Peer disconnected");
```
dangershony: I assume peer was already disposed if we reached this point so why call it again?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -243,35 +207,35 @@ public void Dispose()
...
             this.logger.LogTrace("Waiting for accepting task to complete.");
             this.acceptTask.Wait();
 
-            ICollection<NetworkPeerConnection> connections = this.connectionsById.Values;
-            if (connections.Count > 0)
+            try
             {
-                this.logger.LogInformation("Waiting for {0} connected clients to finish.", connections.Count);
-                foreach (NetworkPeerConnection connection in connections)
+                if (this.peersById.Count > 0)
```
dangershony: Why not do as before and use `if (peers .Count > 0)`

noescape00: Good point. Will fix

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
dangershony: Right ok so inside the collection Dispose was called so all good.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
noescape00: Done

PR #1193, opened 2018-03-09
Agent prefix

bokobza:


src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -236,10 +236,15 @@ public NodeSettings LoadConfiguration()
...
             this.MinRelayTxFeeRate = new FeeRate(config.GetOrDefault("minrelaytxfee", this.Network.MinRelayTxFee));
             this.Logger.LogDebug("MinRelayTxFeeRate set to {0}.", this.MinRelayTxFeeRate);
-
             this.SyncTimeEnabled = config.GetOrDefault<bool>("synctime", true);
             this.Logger.LogDebug("Time synchronization with peers is {0}.", this.SyncTimeEnabled ? "enabled" : "disabled");
 
+            // Add a prefix set by the user to the agent. This will allow people running nodes to
+            // identify themselves if they wish. The prefix is limited to 10 characters.
+            string agentPrefix = config.GetOrDefault("agentprefix", string.Empty);
+            agentPrefix = agentPrefix.Substring(0, Math.Min(10, agentPrefix.Length));
```
dangershony: 10 is a bit too short I think we have a parameter called `MaxSubversionLength` that limits the agents length so if we take the default current test `StratisBitcoin:000.000.000` we could allow more characters in the prefix how about 100 characters?

bokobza: I think 10 is enough to identify yourself.
If we allow 100 it's going to look horrible.

PR #1195, opened 2018-03-11
NetworkPeer refactoring- separate Dispose from Disconnect to avoid cyclic calls 

noescape00:
Fix: #1090

This PR improves architecture if the NetworkPeer in order to prevent cyclic calls, fixes other components that create peers and fixes a few minor bugs.

PR was done with @Aprogiena


Aprogiena:
We have spent a lot of time reworking the design of all components that create and use peers. However, at the same time we wanted the PR to be as small as possible. So there is still a lot to do, it is kind of another series of the node job, which I could not really completed.

This PR focuses on simplification of disconnect and dispose of peers and cancelling cyclic dependencies.

At this point you're just starting the big testing, which is absolutely needed in this PR must not be merged until an extreme effort interesting is done. It is quite likely that we will discover some problems we will need to at couple of fixes to this PR.

I approve this PR conceptually, but I cannot approve it fully, which would mean that I am sure it does not break anything, because I have not done any testing. This is why any help is welcome here, if you can test it. We especially need tests of this version on public IP address with a lot of clients connecting to it

dangershony:
I will focus on this

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -182,14 +211,26 @@ public bool IsConnected
...
+        private int disposed;
+        
+        /// <summary>
+        /// Async context to allow to recognize whether <see cref="onDisconnected"/> callback execution is scheduled in this async context.
+        /// <para>
+        /// It is not <c>null</c> if one of the following callbacks is in progress: <see cref="StateChanged"/>, <see cref="MessageReceived"/>,
+        /// set to <c>null</c> otherwise.
+        /// </para>
+        /// </summary>
+        private readonly AsyncLocal<DisconnectedExecutionAsyncContext> onDisconnectedAsyncContext;
```
dangershony: hmmm... Is this absolutely necessary, we are introducing more levels of complexity here, I assume this will follow a message during its propagation inside the behaviors and that way we can know if disconnect was called from a behaviour or not.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -206,6 +247,9 @@ public bool IsConnected
         /// <inheritdoc/>
         public MessageProducer<IncomingMessage> MessageProducer { get { return this.Connection.MessageProducer; } }
 
+        /// <summary>Callback that is invoked when peer has finished disconnecting, or <c>null</c> when no notification after the disconnection is required.</summary>
+        private readonly Action<INetworkPeer> onDisconnected;
```
dangershony: No other component is using this apart for the helper right?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -268,43 +333,60 @@ public NetworkPeer(IPEndPoint peerEndPoint, Network network, NetworkPeerConnecti
...
                 if ((newState == NetworkPeerState.Failed) || (newState == NetworkPeerState.Offline))
+                {
                     this.logger.LogTrace("Communication with the peer has been closed.");
+
+                    this.ExecuteDisconnectedCallbackWhenSafe();
+                }
+            }
+            else if (previous != newState)
+            {
+                this.logger.LogDebug("Illegal transition from {0} to {1} occurred.", previous, newState);
```
dangershony: This should be much less forgiving it indicates a design error if we hit this `else`

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -236,7 +240,7 @@ public void AddDiscoveredNodesRequirement(NetworkPeerServices services)
             if ((peerConnector != null) && !peerConnector.Requirements.RequiredServices.HasFlag(services))
             {
                 peerConnector.Requirements.RequiredServices |= NetworkPeerServices.NODE_WITNESS;
-                foreach (INetworkPeer peer in peerConnector.ConnectedPeers)
+                foreach (INetworkPeer peer in peerConnector.ConnectorPeers)
```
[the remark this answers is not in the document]
noescape00: Because there were 2 variables: `connectedPeers` and `ConnectedPeers`. One of them was renamed so they are more distinguishable.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: Correct

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: Agreed. Will change to Error level.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: Yes. Since we want to prevent callback call from inside of another callback this is needed.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -134,6 +134,9 @@ public IReadOnlyNetworkPeerCollection ConnectedPeers
 
         public List<NetworkPeerServer> Servers { get; }
 
+        /// <summary>Maintains a list of connected peers and ensures their proper disposal.</summary>
+        private readonly ConnectedPeersHelper connectedPeersHelper;
```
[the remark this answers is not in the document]
noescape00: I think that would be a bit misleading since we already have a `NetworkPeerCollection`. But maybe functionality of the helper can be moved to a `NetworkPeerCollection` class.

Also I don't think that `Collection` will be an appropriate name for the functionality presented in the helper since it doesn't implement `IEnumerable`

src/Stratis.Bitcoin/P2P/Peer/ConnectedPeersHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.P2P.Peer
+{
+    /// <summary>Maintains a list of connected peers and ensures their proper disposal.</summary>
+    public class ConnectedPeersHelper : IDisposable
```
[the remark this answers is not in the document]
noescape00: >I think we can discuss this component a bit further, to my understanding this is just sort of a bridge between a disconnect event on a peer and the call to dispose that peer.

Yes, correct. 

>this could possibly just live in side the peer itself why does it need to be part of the manager?
Because that way we won't be able to implement IDisposable design properly which suggests that by the end of Dispose() call all threads used by the component should be completed. Instead we would have something like OnDisposeCompleted callback and when Dispose() call is finished component won't actually be disposed.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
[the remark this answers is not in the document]
Aprogiena: At 1st we called that network peer disposed helper, because its primary objective was to take care of network peer disposals. However it also maintains a list of all connected peers, at which point the 1st name we selected seemed an appropriate.

Then as no escape says, there is network peer collection already which is used by connection manager, which purpose is to provide a collection of connected peers all other components. For example if another component was to set a message to all connected peers it should use connection managers peer collection.

This is why we did not want to name it is a collection because then we would have 2 collections with completely different purpose and it could be confusing which is which. So this helper class can  be used by any component that actually creates new peers. If a company does that did no longer it needs to track those newly created peers itself and care about their disposal.

So this is the purpose of this class, feel free to come up with better name, we struggled with that ourselves. I agree that since it doesn't really implement enumeration, it is probably not a collection anyway.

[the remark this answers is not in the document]
Aprogiena: I don't think the functionality of this class should be part of the collection, especially when considering that the collection for connection manager must not dispose the peers.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: Alternative solution would be to have a context parameter that everyone would be required to pass to every sub- function. That doesn't seem better at all. Do you see any other alternative?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
Aprogiena: Yes, but not always, which means that without any change just raising this to error level would produce fake errors. I believe during shutdown of the peer and its connection some components may try to set the new state to off-line while the state is already in failed state or vice versa, which seems perfectly valid to me in terms of that the transition should not happen but it should not be reported as an error.

Aprogiena: So if you want to change this, we should exclude transition from off-line to error and error to off-line.

dangershony: >Yes, but not always
I assumed any state transaction in the table is allowed and all others are invalid (and state can be changed only internally

Aprogiena: For that we would have to match off-line and failed states into one

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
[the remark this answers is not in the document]
dangershony: How about something like PeerLifetime or PeerDisposeMonitor etc...
Also instead of using a CD you could just inject the exiting `NetworkPeerCollection` and not need to mange in internal list.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
```diff
@@ -32,16 +33,18 @@ public interface INetworkPeerFactory
...
 
         /// <summary>
         /// Creates a new network peer which is connected to a specified counterparty.
         /// </summary>
         /// <param name="peerEndPoint">Address and port of the counterparty to connect to.</param>
         /// <param name="parameters">Parameters specifying how the connection with the counterparty should be established, or <c>null</c> to use default parameters.</param>
+        /// <param name="connectedPeersHelper">Maintains a list of connected peers and ensures their proper disposal. Or <c>null</c> if case disposal should be handled from user code.</param>
         /// <returns>Network peer connected to the specified counterparty.</returns>
-        Task<INetworkPeer> CreateConnectedNetworkPeerAsync(IPEndPoint peerEndPoint, NetworkPeerConnectionParameters parameters = null);
+        Task<INetworkPeer> CreateConnectedNetworkPeerAsync(IPEndPoint peerEndPoint, NetworkPeerConnectionParameters parameters = null, ConnectedPeersHelper connectedPeersHelper = null);
```
dangershony: `ConnectedPeersHelper` is a singleton that lives through out the node lifetime so you could just inject it to the factory.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -159,6 +162,7 @@ public IReadOnlyNetworkPeerCollection ConnectedPeers
             this.PeerConnectors = peerConnectors;
             this.peerDiscovery = peerDiscovery;
             this.ConnectionSettings = connectionSettings;
+            this.connectedPeersHelper = new ConnectedPeersHelper(this.loggerFactory);
```
dangershony: If we don't inject it to the factory (see my other comment then here it can definitely be injected.

dangershony: Hmm as this is a utility class then perhaps no need to inject it but it could be an internal citizen of the factory.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
```diff
@@ -22,8 +22,9 @@ public interface INetworkPeerFactory
         /// </summary>
         /// <param name="client">Already connected network client.</param>
         /// <param name="parameters">Parameters of the established connection, or <c>null</c> to use default parameters.</param>
+        /// <param name="connectedPeersHelper">Maintains a list of connected peers and ensures their proper disposal. Or <c>null</c> if case disposal should be handled from user code.</param>
         /// <returns>New network peer that is connected via the established connection.</returns>
-        INetworkPeer CreateNetworkPeer(TcpClient client, NetworkPeerConnectionParameters parameters = null);
+        INetworkPeer CreateNetworkPeer(TcpClient client, NetworkPeerConnectionParameters parameters = null, ConnectedPeersHelper connectedPeersHelper = null);
```
dangershony: Why to allow null helpers?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
noescape00: That was the first implementation of the ConnectedPeersHelper (add as a singleton, inject to the factory and add a bool `disposeOnDisconnect` that will make the factory use the helper, dispose the helper on ConnectionManager disposal).

Don't remember why we didn't stick with that solution, but for sure I don't mind it being that way.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -419,6 +518,13 @@ public NetworkPeer(IPEndPoint peerEndPoint, Network network, NetworkPeerConnecti
                 this.logger.LogTrace("(-)[EXCEPTION_CALLBACKS]");
                 throw;
             }
+            finally
+            {
+                if (this.onDisconnectedAsyncContext.Value.DisconnectCallbackRequested)
+                    this.onDisconnected(this);
```
dangershony: Effectively the only thing this does is delegate the dispose to the AsyncQueue, can we not write such a delegated dispose internally to the NetworkPeer?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
noescape00: In case disposal should be handled in the user code. 
Right now only peer discovery uses that.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
noescape00: We can and it is exactly what it used to be before. 
And that leaded to the wrong implementation of IDisposable (by the end if Dispose() call the peer was not disposed)

In most cases Dispose() on network peer should be called from a separated task to avoid locks

dangershony: Hmm it will be a bit of an overhead as each peer will have to have its own dispose task waiting in the background so I see the reason of using a shared dispose queue. but this can be more specific to the NetworkPeer perhaps instead of injecting an OnDispose delegate just pass the DisposeHelper itself

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: Yes, that's also an option. But that way we will need to make the factory IDisposable

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
```diff
@@ -130,24 +147,36 @@ public INetworkPeer CreateNetworkPeer(TcpClient client, NetworkPeerConnectionPar
...
+            IPEndPoint peerEndPoint,
+            NetworkPeerConnectionParameters parameters = null,
+            ConnectedPeersHelper connectedPeersHelper = null)
         {
             Guard.NotNull(peerEndPoint, nameof(peerEndPoint));
 
-            var peer = new NetworkPeer(peerEndPoint, this.network, parameters, this, this.dateTimeProvider, this.loggerFactory);
+            Action<INetworkPeer> onDisconnected = null;
+
+            if (connectedPeersHelper != null)
```
dangershony: This is a bit of a hack, as I suggested earlier we can inject the PeerLifetime (or what ever we call it maybe PeerDisposer) directly to the peer itself.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
dangershony: Ah its not IDisposable I see, good point.
This feel more like a factory type then a connection manager type so we need to think about it.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
dangershony: Ok I am struggling to grasp why the DisposeHelper needs to be an external component to the peer.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
dangershony: Right so if we just call `Task.Run(()=> this.DisposeInternal())` will that not work?

noescape00: It will work. But in order to actually dispose the peer we will need to: 
```
peer.Dispose()
peer.DisposeCompleted.Wait()
```

Which goes against the Dispose pattern

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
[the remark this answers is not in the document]
Aprogiena: Why would we use a heavier collection for something we only need dictionary for? That doesn't make sense.

As explained above no longer this component is only about disposing, it also tracks all connected peers of the owning component. So peer dispose monitor is not very good I think.

Peer lifetime also doesn't resonate with me. But in general I agree that we did struggle with naming this class. But so far no better ideas in my head.

Any ideas @noescape00?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
Aprogiena: It is not a Singleton. Multiple instances exist. As explained above this class is used when a component wants to create network peers and does not want to manage the disposal and their tracking. So each such component will have its helper for this tracking and disposing.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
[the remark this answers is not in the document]
noescape00: `PeerDisposalHelper` as it was before, maybe?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
Aprogiena: Each component has to have its own instance of this helper, because when the component is being destroyed it only wants to destroy those peers that it owns, not all peers.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
Aprogiena: As explained above, if the component does not want to track and dispose its peers, it can use the helper. This is optional however because at least one component in our code manage disposal itself and does not need to track peers at all.

It would probably better for me to make a Skype call to explain this design.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerFactory.cs
Aprogiena: There's a fundamental misunderstanding of the purpose of this component, let's have Skype call for that.

Aprogiena:
`ConnectedPeersHelper ` Class comment is obviously completely insufficient. It should describe the whole story of this class, why is it needed, who is supposed to use it, et cetera. It should be at least 10 lines of comments.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
dangershony: What do you mean goes against the pattern?

noescape00: Dispose() suggests that by the end of it's call all threads that this object used are closed. 
If we do 
```
peer.Dispose()
peer.DisposeCompleted.Wait()
```

that basically means that we didn't finish disposing after exiting the Dispose() method and we would like to avoid that.

dangershony:
Ok had a Skype with @Aprogiena am happy with this PR I just want to changes.
As suggested by Justin lets rename `ConnectedPeersHelper` to `PeerDisposer` or `NetworkPeerDisposer`

Please also add more comments describing the purpose of that class 
> i.e the creator of a component needs to be also the disposer of it and a peer instance can not dispose itself so the NetworkPeerDisposer is responsible to disposing peer instance.

Aprogiena:
i'm ok with `NetworkPeerDisposer`

dangershony:
I have been running this branch since yesterday and staking all good

noescape00:
Renamed ConnectedPeersHelper to NetworkPeerDisposer and added a detailed class comment that describes purpose of the class and it's usage.

Aprogiena:
don't merge yet pls

dangershony:
>don't merge yet pls

Yes sir, just push when you are happy.

Aprogiena:
yeah, we just wanted to fix that comment

@noescape00 when you feel you are happy, we can merge

Issue #1167, opened 2018-02-26
[mining feature] remove POS loop


Aprogiena:
`posLoop` Should be removed

`Dispose` Should call `StopStaking`, Which should use `this.posMinting?.StopStake();`

dangershony:
>`posLoop` Should be removed 

And replaced with what?

Aprogiena:
Nothing, it is disposed elsewhere. Look into the code, the same loop is assigned to 2 variables and there is little bit of mess where one is disposed in the other isn't

majikandy:
I can look at this one

dangershony:
Ah ok yeah (I though you meant replace the loop itself) we don't use that variable any way.
Go for this change @majikandy

majikandy:
Can someone assign it to me? I don’t seem to have that permission

Aprogiena:
I still can't see you on the list, I do see other new developers there, but you seem to be forgotten

majikandy:
@StratisIain is this something you can fix?  apparently I am no on the list to be able to be assigned to github issues

PR #1197, opened 2018-03-12
Remove pos loop var

majikandy:


majikandy:
Fix:#1167

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -148,15 +137,15 @@ public override void Initialize()
         {
             if (this.minerSettings.Mine)
             {
-                string minto = this.minerSettings.MineAddress;
-                // if (string.IsNullOrEmpty(minto)) ;
+                string mineToAddress = this.minerSettings.MineAddress;
```
dangershony: This code should eventually go away as its not secure to pass the password on the command line (we should use swagger api or command cli instead)

PR #1198, opened 2018-03-12
Fix for wrong height reward calculation

stepkillah:
Hi, it's my first commit to this repo, let me know if something wrong. This problem occurs when I'm tried to create a private PoW/PoS chain with different parameters.
Not sure if this is correct, but `PosMinting` calculate the reward for current blockchain height (let's say `N`), but when it validates reward in `PosConesnsusValidator` -> `CheckBlockReward` method - its uses newly created block height (`N+1`). As far as Stratis has a constant reward for PoS - 1 coin, it will work properly, but in case if reward depends on the block - it probably won't work

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -783,8 +783,8 @@ namespace Stratis.Bitcoin.Features.Miner
             }
 
             this.logger.LogTrace("Worker #{0} found the kernel.", workersResult.KernelFoundIndex);
-			//Get reward for newly created block
-            long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
+	        //Get reward for newly created block
+	        long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
             if (reward <= 0)
             {
                 // TODO: This can't happen unless we remove reward for mined block.
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -784,7 +784,7 @@ namespace Stratis.Bitcoin.Features.Miner
 
             this.logger.LogTrace("Worker #{0} found the kernel.", workersResult.KernelFoundIndex);
             // Get reward for newly created block.
-            long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -783,8 +783,8 @@ private void CheckStake(Block block, ChainedBlock chainTip)
             }
 
             this.logger.LogTrace("Worker #{0} found the kernel.", workersResult.KernelFoundIndex);
-
-            long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height);
+	        //Get reward for newly created block
```
Aprogiena: – don't use Tabs, use spaces
– in-line comment should start with space
– comment sentence should and with `.`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -783,8 +783,8 @@ private void CheckStake(Block block, ChainedBlock chainTip)
...
+	        long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
```
Aprogiena: – don't use Tabs, use spaces
– Use Ctrl+K+D to let visual studio format your code properly, which would mean spaces around the +

Aprogiena:
From formal point of view – i.e. code style – I have added comments to your commit.

From the factual point of view, we need to think about it.

Aprogiena:
Okay, so from the factual point of view, I think you're right, Because height of our new block is the current tip +1.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -783,8 +783,8 @@ private void CheckStake(Block block, ChainedBlock chainTip)
             }
 
             this.logger.LogTrace("Worker #{0} found the kernel.", workersResult.KernelFoundIndex);
-	        //Get reward for newly created block
-	        long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
+            // Get reward for newly created block.
+            long reward = fees + this.posConsensusValidator.GetProofOfStakeReward(chainTip.Height+1);
```
Aprogiena: `chainTip.Height + 1`

As mentioned before, Ctrl+K+D in visual studio should do this for you

bokobza:
Good stuff @stepkillah, thank you!

dangershony:
Excellent good spot and congrats on your first commit

Issue #1120, opened 2018-01-31
[peerbanning] Banned peers need to be remove from the peer address store


dangershony:
Banned peers should not be propagated to peers so when a peer is banned it needs to be removed from store.

Aprogiena:
And also when we receive a list of new addresses from a peer, we should check the list against the list of banned nodes, So that we do not introduce those removed nodes again

dangershony:
Perhaps the right way to do this is to add a flag on the address manager file that a peer is banned instead of creating a new store for panned peers

Aprogiena:
You can do that, if you extend the format of that file to include that information and some other information like reason for the ban and duration of that or something, then yes you wouldn't need new store.

dangershony:
Well if this data is anyway going to be stored then we save additional complexity (adding a new store) by reusing the existing store, so unless we will have requirements to delete banned peers (or none banned peers) after a time period I think this is a good approach.

Aprogiena:
The problem with the peer manager file is that it is not a real store as far as I know, so it can get corrupted in case of power failure, unlike Dbreeze database. But that's something else, I'm jumping to another topic.

So yes I agree that you can have one store and it will be fine

PR #1201, opened 2018-03-13
Banned peers need to be removed from the peer address store (#1120)

ferdeen:
Fix: #1120

 * PeerAddressManager is now injected into PeerBanning.  This is to add banned peers in one central store, and to disk.

  * Loading Peers from disk, into the PeersAddressManager.Peers resets any expired banned Peers.

  * IPeerBanning.IsBanned also resets banned Peers, if BanUntil has expired.

  * PeerBanning has a redundant class - NodeSettings - injected into it.  I have taken this out and altered related unit tests.

  * New tests exist in PeerBanningTest to cover this fix.


src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -162,6 +163,33 @@ public bool Handshaked
...
+        public bool? IsBanned { get; set; }
+
+        /// <summary>
+        /// Reason for the banning the Peer.
+        /// <para>
+        /// This is set in <see cref="PeerBanning"/>.
+        /// </para>
+        /// </summary>
+        [JsonProperty(PropertyName = "bannedreason", NullValueHandling = NullValueHandling.Ignore)]
+        public string BannedReason { get; set; }
```
dangershony: Hmm do we need to keep the ban reason in store? this is just an arbitrary string right?

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -162,6 +163,33 @@ public bool Handshaked
...
         public DateTime? LastDiscoveredFrom { get; private set; }
 
+        /// <summary>
+        /// Determine whether the peer has been banned.
+        /// <para>
+        /// This is set in <see cref="PeerBanning"/>.
+        /// </para>
+        /// </summary>
+        [JsonProperty(PropertyName = "isbanned", NullValueHandling = NullValueHandling.Ignore)]
+        public bool? IsBanned { get; set; }
```
dangershony: I am not sure we need this property at all in store as it enough to have the `BanUntil ` to know if a peer is banned and for how long, how ever it might be useful to know when the ban was created.

So perhaps we only need to properties  `BanDate` and `BanUntil` and maybe also reason (but not sure yet about as this will be available in the logs if needed)

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -238,5 +240,25 @@ public void Dispose()
         {
             this.SavePeers();
         }
+
+        /// <summary>
+        /// When the PeerAddressManager saves peers to disk, check if any of
+        /// banned Peers have expired.  If so, then reset the Peers.
+        /// </summary>
+        private void ResetBannedPeers()
```
dangershony: We probably dont need to reset banned peers we just explode banned peers in the selection algorithm.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -131,7 +83,14 @@ public void BanPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = nul
...
                 this.logger.LogDebug("Peer '{0}' banned for reason '{1}'.", endpoint, reason);
-                this.banStore.BanPeer(endpoint, this.dateTimeProvider.GetUtcNow().AddSeconds(banTimeSeconds));
+
+                PeerAddress peerToAdd = PeerAddress.Create(endpoint, endpoint.Address);
+
+                peerToAdd.IsBanned = true;
+                peerToAdd.BannedReason = reason;
+                peerToAdd.BanUntil = this.dateTimeProvider.GetUtcNow().AddSeconds(banTimeSeconds);
+
+                this.peerAddressManager.Peers.TryAdd(peerToAdd.Endpoint, peerToAdd);
```
dangershony: Its probably not such a good idea to have the `Peers` dictionary as public this means any outside component can modify/add peers, this is effectively the store so perhaps we should protect it (maybe expose the values only) @fassadlr any reason its a public property?

Adding to this list has to be done in one place only I would suggest to change the method on IPeerAddressManager.Add to return the newly created PeerAddress so the caller can then set the BannedUntil property

dangershony:
Perhaps as this PR we should also filter out banned peers from the IPeerSelector methods @fassadlr

src/Stratis.Bitcoin/P2P/PeerAddress.cs
ferdeen: Ok, I'll change to this.  As we have the banned reason in the logs I'll remove the BannedReason property.

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
ferdeen: The reset was to make sure the data persisted to the Peers.json file is correct.  So if the BanUntil has passed we need a way null out BanDate and BanUntil in the json file.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
ferdeen: I'll remove as stored in the log as you said.

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
dangershony: hmm ok I see, it just to reduce the size of the file where do we call this from?

src/Stratis.Bitcoin/Connection/PeerBanning.cs
dangershony: I just made this change in another PR

ferdeen: Can you send over a link or commit details and I'll take a look, cheers.

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
ferdeen: When we load the peers.json file - PeerAddressManager.LoadPeers().

src/Stratis.Bitcoin/Connection/PeerBanning.cs
dangershony: yeah #1204

ferdeen: Ta

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
dangershony: Ok so if we call this both onLoad and onSave this is fine I guess

src/Stratis.Bitcoin/Connection/PeerBanning.cs
ferdeen: Merged in #1204 to my branch and changed the code for this issue accordingly.  Looks a lot cleaner now.  Will create a new PR.  Cheers.

Aprogiena:
@dangershony What is the discussion here? I can see this is probably not supposed to be a pull request ready to be merged, right?

ferdeen:
@Aprogiena The commit was upstream to my local bug branch - not too sure why is showed up here..

Anyhow, have made the changes based on this thread and commited to my local bug branch and can see the commit here.  

What is the process now, will the reviewers see my new changes ?

Cheers.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -128,11 +81,16 @@ public void BanPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = nul
                 }
             }
 
-            if (banPeer)
-            {
-                this.logger.LogDebug("Peer '{0}' banned for reason '{1}'.", endpoint, reason);
-                this.banStore.BanPeer(endpoint, this.dateTimeProvider.GetUtcNow().AddSeconds(banTimeSeconds));
-            }
+            if (!banPeer) return;
```
dangershony: Please put log traces inside this `if`s

dangershony:
Hi @Aprogiena the main question is the content of the store currently we only store BanUntil, and BanDate (we dont store ban reason) anything else you think we need to store to disk?

src/Stratis.Bitcoin/Connection/PeerBanning.cs
ferdeen: Sorry I don't follow.

If banPeer is false (line 84) then return, as no need to go to AddressManager to get the ban details.  Otherwise carry on.  

I have added a LogDebug at line 93 - to log BanDate and BanUntil values (if the Peer is found).  Should I change this to LogTrace ?

dangershony:
I had some more thought about this and I realise we don't have any way to keep a bad behaviour counter. 
If a node has misbehaved for what ever reason and accumulated such behaviour we may want to be able to  maintain a bad behaviour counter and ban only after a certain score was has elapsed.

To do this we could add a BanScore to the store which becomes invalid once the BanUntil is set.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
Aprogiena: I will explain on the call

Aprogiena:
Please consider describing the commits based on what is in each comment and not based on how the full PR is called

src/Stratis.Bitcoin.Tests/P2P/PeerSelectorTests.cs
```diff
@@ -468,5 +468,151 @@ public void PeerSelector_ReturnPeersForDiscovery_Scenario1()
             var peers = peerAddressManager.PeerSelector.SelectPeersForDiscovery(5);
             Assert.Equal(2, peers.Count());
         }
+
+        /// <summary>
+        /// Ensures that a particular peer is returned from the attempted peers 
+        /// set and ignores banned peers.
+        ///
```
Aprogiena: Do not use empty lines as they are discarded in XML, use `<para>` instead

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -76,21 +75,24 @@ public void BanPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = nul
...
+                return;
+            }
 
-            peerAddress.BanDate = this.dateTimeProvider.GetUtcNow();
+            peerAddress.BanTimeStamp = this.dateTimeProvider.GetUtcNow();
             peerAddress.BanUntil = this.dateTimeProvider.GetUtcNow().AddSeconds(banTimeSeconds);
+            peerAddress.BanReason = reason;
 
-            this.logger.LogDebug("Peer '{0}' banned for reason '{1}'.", endpoint, reason);
+            this.logger.LogDebug("Peer '{0}' banned for reason '{1}', until {2}.", endpoint, reason, peerAddress.BanUntil.ToString());
```
Aprogiena: No need for ToString unless you want to specify special format

src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -105,7 +107,12 @@ public bool IsBanned(IPEndPoint endpoint)
             PeerAddress peerAddress = this.peerAddressManager.FindPeer(endpoint);
 
             if (peerAddress == null)
+            {
+                this.logger.LogTrace("(-)[PEERNOTFOUND]");
                 return false;
+            }
+
+            this.logger.LogTrace("(-)");
```
Aprogiena: Here and on line 111, the exit log should contain the return value, so for example here it should look like:

```
    this.logger.LogTrace("(-):{0}", res);
```

On line 111

```
    this.logger.LogTrace("(-)[PEER_NOT_FOUND]:false");
```

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -80,23 +80,58 @@ public IPAddress Loopback
         public DateTime? LastSeen { get; private set; }
 
         /// <summary>
-        /// The UTC Date when a Peer is banned.
-        /// <para>
-        /// This is set in <see cref="PeerBanning"/>, against <see cref="PeerAddress"/>.
-        /// </para>
+        /// UTC DateTime when a peer is banned.
```
Aprogiena: UTC timestamp when ...
or
Time in UTC when ...

Otherwise you would need to use `<see>` if you wanted to reference the DateTime class

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -80,23 +80,58 @@ public IPAddress Loopback
...
+        /// <remarks>
+        /// This is set in <see cref="PeerBanning"/>.
+        /// </remarks>
+        [JsonProperty(PropertyName = "banreason", NullValueHandling = NullValueHandling.Ignore)]
+        public string BanReason { get; set; }
+
+        /// <summary>
+        /// Maintain a count of bad behaviour.  
+        /// </summary>
+        /// <para>
```
Aprogiena: `<para>` Is part of summary or remarks, it is not section by itself

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -80,23 +80,58 @@ public IPAddress Loopback
...
+        /// <summary>
+        /// <c>True</c> if <see cref="BanUntil"/> is in the future.
+        /// </summary>
+        [JsonIgnore]
+        public bool IsBanned
+        {
+            get
+            {
+                return 
+                    (this.BanUntil > DateTime.UtcNow.Date);
```
Aprogiena: Remove this method, because otherwise you would need date time provider to be injected. Never use date time UTC now directly.

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -137,11 +137,22 @@ public void LoadPeers()
             {
                 // Ensure that any address already in store is mapped.
                 peer.Endpoint = peer.Endpoint.MapToIpv6();
-                
-                this.peers.TryAdd(peer.Endpoint, peer);
-            });
 
-            ResetBannedPeers();
+                // If no longer banned reset ban details.
+                if (peer.BanUntil.HasValue && peer.BanUntil < this.dateTimeProvider.GetUtcNow())
```
Aprogiena: See rule 23 of coding style

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -137,11 +137,22 @@ public void LoadPeers()
...
+                    peer.BanTimeStamp = null;
+                    peer.BanUntil = null;
+                    peer.BanReason = string.Empty;
+
+                    this.logger.LogTrace("{0} no longer banned.", peer.Endpoint);
+                }
+
+                if (!this.peers.TryAdd(peer.Endpoint, peer))
+                {
+                    this.peers.AddOrUpdate(peer.Endpoint, peer, (key, oldValue) => peer);
```
Aprogiena: Why would you not call add or update in the 1st place

Aprogiena:
@dangershony Are you okay to finished the review once updated?

ferdeen:
@Aprogiena thanks for your comments.  I have just done another commit.  

@dangershony when your free please review.  

Cheers

dangershony:
@Aprogiena thanks for having a look I will take it from here.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -79,6 +79,45 @@ public IPAddress Loopback
...
+        [JsonProperty(PropertyName = "banuntil", NullValueHandling = NullValueHandling.Ignore)]
+        public DateTime? BanUntil { get; set; }
+
+        /// <summary>
+        /// Reason for banning the peer.
+        /// <remarks>
+        /// This is set in <see cref="PeerBanning"/>.
+        /// </remarks>
+        [JsonProperty(PropertyName = "banreason", NullValueHandling = NullValueHandling.Ignore)]
+        public string BanReason { get; set; }
```
dangershony: Did we decide we want to store the ban reason?

ferdeen: Yes, I had a chat with @Aprogiena on this, and he said sizing wouldn't be an issue.

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -137,8 +137,18 @@ public void LoadPeers()
             {
                 // Ensure that any address already in store is mapped.
                 peer.Endpoint = peer.Endpoint.MapToIpv6();
-                
-                this.peers.TryAdd(peer.Endpoint, peer);
+
+                // If no longer banned reset ban details.
+                if (peer.BanUntil.HasValue && peer.BanUntil < this.dateTimeProvider.GetUtcNow())
```
dangershony: So we only clear this list on startup, that's fine I guess as it indirectly means we clear it on shutdown.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -196,6 +197,7 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
         public IEnumerable<PeerAddress> SelectPeersForDiscovery(int peerCount)
         {
             var discoverable = this.peerAddresses.Values.Where(p => p.LastDiscoveredFrom < this.dateTimeProvider.GetUtcNow().AddHours(-PeerSelector.DiscoveryThresholdHours));
+
```
dangershony: This method does not filter banned peers?

fassadlr: @majikandy Didn't you fix this? We discussed this the other day.. unless its done somewhere else now?

dangershony:
Ok I think we need make sure we filter banned peers when
- we receive addresses from getaddr 
- when we try to connect to random peers (outbound)
- when a connection is made to us (inbound)

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: @majikandy Oh no wait that's another PR, sorry.

ferdeen: I can fix SelectedPeersForDiscovery as follows :- (this is on my branch)

image

majikandy: >@majikandy Didn't you fix this? We discussed this the other day.. unless its done somewhere else now?

I added the filter for !IsSelf in this PR which is not merged yet https://github.com/stratisproject/StratisBitcoinFullNode/pull/1218

ferdeen:
@dangershony to answer the below : - 

**We receive addresses from getaddr**  

PeerAddressManagerBehaviour.OnMessageReceivedAsync() makes a call to peerAddressManager.PeerSelector.SelectPeersForGetAddrPayload (line 71)

PeerSelector.SelectPeersForGetAddrPayload() calls this.Connected() & this.Attempted() (lines 219-220).
both methods include the logic to ignore banned peers.

**When we try to connect to random peers (outbound)**

As you mentioned, already covered in peer selector.

**When a connection is made to us (inbound)**

As mentioned by @noescape00 : 

> We have PeerBanningBehavior for that already. If a server receives a connection from a peer that is banned we will disconnect during the process of attaching the behaviors
> so the banning itself is implemented here.

I've added code logic for `SelectPeersForDiscovery`, mentioned above, and pushed for review.

Cheers.

Aprogiena:
>> We receive addresses from getaddr
>
>PeerAddressManagerBehaviour.OnMessageReceivedAsync() makes a call to >peerAddressManager.PeerSelector.SelectPeersForGetAddrPayload (line 71)
>
>PeerSelector.SelectPeersForGetAddrPayload() calls this.Connected() & this.Attempted() (lines 219-220).
>both methods include the logic to ignore banned peers.

This is the opposite direction, node is the one that sent get address message and we received address payload as an answer, which leads to this code:

```
                if ((this.Mode & PeerAddressManagerBehaviourMode.Discover) != 0)
                {
                    if (message.Message.Payload is AddrPayload addr)
                        this.peerAddressManager.AddPeers(addr.Addresses.Select(a => a.Endpoint).ToArray(), peer.RemoteSocketAddress);
                }
```

In which there is no check whether the received addresses are banned or not.

Aprogiena:
Okay if that is the case then it is fine

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -207,7 +208,8 @@ public IEnumerable<PeerAddress> SelectPeersForDiscovery(int peerCount)
         {
             IEnumerable<PeerAddress> discoverable = this.peerAddresses.Values
                 .Where(p => p.LastDiscoveredFrom < this.dateTimeProvider.GetUtcNow().AddHours(-PeerSelector.DiscoveryThresholdHours))
-                .Where(p => !this.selfEndpointTracker.IsSelf(p.Endpoint));
+                .Where(p => !this.selfEndpointTracker.IsSelf(p.Endpoint))
+                .Where(p => !IsBanned(p));
```
dangershony: One last change `!this.IsBanned(p)` also in other places

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -273,27 +275,40 @@ private PeerAddress Random(IEnumerable<PeerAddress> peers)
...
         public IEnumerable<PeerAddress> Handshaked()
         {
-            return this.peerAddresses.Values.Where(p => p.Handshaked && p.LastConnectionHandshake < this.dateTimeProvider.GetUtcNow().AddSeconds(-60));
+            return this.peerAddresses.Values.Where(p => p.Handshaked && 
+                                                        p.LastConnectionHandshake < this.dateTimeProvider.GetUtcNow().AddSeconds(-60) &&
+                                                        !IsBanned(p));
+        }
+
+        /// <summary>
+        /// <c>True</c> if <see cref="PeerAddress.BanUntil"/> is in the future.
```
Aprogiena: `<c>true</c>`

but most importantly, the correct comment should say "if the address is banned" which is the real purpose of the method. what you say is how you implement it

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -273,27 +275,40 @@ private PeerAddress Random(IEnumerable<PeerAddress> peers)
...
+        /// </summary>
+        private bool IsBanned(PeerAddress peerAddress)
+        {
+            return (peerAddress.BanUntil > this.dateTimeProvider.GetUtcNow());
```
Aprogiena: no need for those extra `( ... )` here

Issue #1153, opened 2018-02-09
[PeerAddressManager] make sure all addresses are IPv6


dangershony:
In the PeerAddresManager we need to make sure all addresses that are added are in the IPv6 format (or if IPv4 then in an embedded format IPv4 represented in an IPv6 address.

From discussion with @345paul 
>so what we need to do is verify we always use IPv6 and convert any instance of IPv4 to an embedded format.

>yes mate - it's in the AddPeer of PeerAddressManager.  It should check to see what kind of IPEndPoint it is given and convert it as necessary

>which actually calls PeerAddress.Create under the hood - so I guess in there's best to fix it

PR #1204, opened 2018-03-13
Store only IPv6 or IPv4 in embedded format

dangershony:
Fix: #1153 

This is one for @345paul and @fassadlr


345paul:
Looks ok.  I think the peer file was ok, but it doesn't hurt to ensure it is converted on load.  The AddPeer method looks fine.

dangershony:
I did review all places but your suggestion (in private) to do it in the Create method is actually better updating PR

345paul:
Looks good.  I see you've put a change in to make the ConcurrentDictionary private and expose a simple List generated from the dictionary on access of the property.  I assume that you wanted a simpler list exposed for other reasons, as I'm not sure this is relevant to the issue being fixed?

dangershony:
Yes it came up in another PR so I just fixed it here, but this is the right place as it will prevent any other components from adding to the ConcurrentDictionary none embedded ip addresses

dangershony:
Also the reason for using the `this.peers.Select(item => item.Value).ToList()` can be found here https://arbel.net/2013/02/03/best-practices-for-using-concurrentdictionary/

Issue #1205, opened 2018-03-13
Staking exception because collection was modified


noescape00:
```
[2018-03-13 20:02:50.8818 39] ERROR: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0+<<StartAsync>b__0>d.MoveNext PosMining.Stake threw an unhandled exception: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at Stratis.Bitcoin.Features.Miner.PosMinting.<StakeAndSignBlockAsync>d__46.MoveNext() in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Miner\PosMinting.cs:line 668
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Stratis.Bitcoin.Features.Miner.PosMinting.<GenerateBlocksAsync>d__44.MoveNext() in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Miner\PosMinting.cs:line 556
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Stratis.Bitcoin.Features.Miner.PosMinting.<>c__DisplayClass42_0.<<Stake>b__0>d.MoveNext() in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Miner\PosMinting.cs:line 426
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1209, opened 2018-03-14
Fix: Staking exception because collection was modified

noescape00:
Fix: #1205

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -665,12 +665,12 @@ namespace Stratis.Bitcoin.Features.Miner
 
                     // We have to make sure that we have no future timestamps in
                     // our transactions set.
-                    foreach (Transaction transaction in block.Transactions)
+                    for (int i = block.Transactions.Count - 1; i >= 0; i--)
                     {
-                        if (transaction.Time > block.Header.Time)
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -665,12 +665,12 @@ private void CheckStake(Block block, ChainedBlock chainTip)
 
                     // We have to make sure that we have no future timestamps in
                     // our transactions set.
-                    foreach (Transaction transaction in block.Transactions)
+                    for (int i = block.Transactions.Count - 1; i >= 0; i--)
```
dangershony: Hmm I would have done it differently but this is also good

noescape00: You mean using a 2nd list like `toBeRemoved` or list.RemoveAll(x => x.timestamp > ...)? 

First solution will consume more memory and 2nd one will force us to get rid of the trace log. 
Or you had a different one in mind?

dangershony: Yeah using as second list, it wont consume more memory as you are not creating new instances just moving them to a new list. but this is good as well no problem.

noescape00: Ok, cool. 
Though I don't mind changing it if you'd like.

PR #1211, opened 2018-03-14
Attempt dns seeds only once

dangershony:


src/Stratis.Bitcoin/P2P/PeerDiscoveryLoop.cs
```diff
@@ -62,6 +62,9 @@ public sealed class PeerDiscovery : IPeerDiscovery
         /// <summary>Factory for creating P2P network peers.</summary>
         private readonly INetworkPeerFactory networkPeerFactory;
 
+        /// <summary>Indicates the dns and seed nodes where attempted.</summary>
```
fassadlr: 'were' not 'where' ;)

dangershony: ahh thanks mate

PR #1218, opened 2018-03-15
Connection to self

### LINKED ISSUE #1084: [PeerConnector + NetworkPeer] Connection to self
If `peers.json`  file contains information about the node itself, it will try to connect to itself.

1) PeerConnector - should probably not allow this to happen, it should not open the connection and it should not try to discover other nodes from itself

2) NetworkPeer - should close that connection (at least it seems to me that the code is intended to do that, but it just doesn't work for some reason)

majikandy:
Fix #1084 

SelfEndpointTracker to store Endpoints known to be self (via the nonce check between peer versions)

Check in PeerConnector to not connect if known to be self.


src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// Tracker for endpoints known to be self.
+    /// </summary>
+    public interface ISelfEndpointTracker
+    {
+        /// <summary>Adds an endpoint to the currently known list.</summary>
+        /// <param name="ipEndPoint">The endpoint to add.</param>
+        void Add(IPEndPoint ipEndPoint);
+
+        /// <summary>Checks if endpoint is known to be itself against the pruned dictionary.</summary>
+        /// <param name="ipEndPoint"></param>
```
Aprogiena: Something is missing here

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -226,6 +231,13 @@ public void StartConnectAsync()
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(peerAddress), peerAddress.Endpoint);
 
+            if (this.selfEndpointTracker.IsSelf(peerAddress.Endpoint))
+            {
+                this.logger.LogTrace($"{peerAddress.Endpoint} is self. Therefore not connecting.");
```
Aprogiena: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md#interpolated-strings

src/Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "commandLineArgs": "-testnet -addnode=127.0.0.2"
```
Aprogiena: You probably don't want to change this file

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    /// <summary>
+    /// Tracker for endpoints known to be self.
+    /// </summary>
+    public class SelfEndpointTracker : ISelfEndpointTracker
+    {
+        /// <summary>Stored endpoints expire after this many hours and need to be readded.
+        /// <para>So that changes in Endpoint Address don't prevent a connection to another node
+        /// in the event they have the Endpoint Address that used to belong to this node</para>
+        /// </summary>
```
Aprogiena: This does not make much sense to me. We don't really store this information to disk, therefore it will be forgotten after the node restarts. And during the lifetime of the node it would not change it end points.

src/Stratis.StratisD/Properties/launchSettings.json
majikandy: correct, whoops, undoing

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
majikandy: it is :)

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
majikandy: A node running on a dynamic IP address - what happens if the ISP issues a new IP address?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
majikandy: ok

Aprogiena:
Although this is an improvement to previous state, it is not complete to fix what was left in issue #1084. I like the idea of the node reacting to detection of connection to itself. This way we can to prevent the difficult task to open a list of all our external endpoints.

This modification does not affect PeerDiscoveryLoop, so it seems that the Discovery component will still connect to itself even after 1st attempt when it is detected.

Most importantly,It does not change the database of peers, therefore after the node is restarted, it will again connect to itself. I think this is crucial at should be added to this PR, because without it, it is very weak.

So I suggest that peer discovery is also affected and that upon detection the peer information is removed from the peer database, so that it won't be used after the node is restarted. Maybe the information should not be removed from the database, just a special flat should be said that this peer represents our own node and we should not connect to it. That is probably even better then removing the peer from the database, because it will prevent the entry to reappear later.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
Aprogiena: That is a good question, but this does not seem to be an answer. As mentioned in the PR comment, I like the fact that we do not need to detect our external endpoints. Therefore suddenly this task is much easier. The disadvantage, however, is that we have no how to react to change of IP address.

We could try to look forward IP address is to other nodes, but then we will lose the benefit of this approach. Therefore I would probably discard the problem of dynamic IP address, because it is unlikely that if we change IP address then on our old IP address another node will exist. So it seems okay not to connect there, pretending that it is our own endpoint.

dangershony:
ok I see the point of using a singleton interface, this can be injected to the discovery loop to check connection to self.

So to add to apro suggestion in peerdiscoveryloop we can remove the self ip addresses from the db I think keeping a flag is a bit too much work for a maybe one extra short connection now and then

Aprogiena:
> I think keeping a flag is a bit too much work for a maybe one extra short connection now and then

I disagree with that, but it is not critical.

Having it in the discovery loop, however, seems important.

majikandy:
added the check to peer discovery.

>So to add to apro suggestion in peerdiscoveryloop we can remove the self ip addresses from the db I think keeping a flag is a bit too much work for a maybe one extra short connection now and then

I will filter inside PeerSelector.SelectPeer  too  but in terms of removing from db @fassadlr suggests this is a bad idea to actually remove from db as this would propogate round and effectively ban yourself from the network

src/Stratis.StratisD/Properties/launchSettings.json
Aprogiena: Here we go again

Aprogiena:
I went briefly through the code and it looks good to me, but as usual I have done no tests, so I can't approve and someone else needs to do the testing.

dangershony:
I am on this now

src/Stratis.Bitcoin.Tests/P2P/PeerSelectorTests.cs
```diff
@@ -358,41 +377,41 @@ public void PeerSelector_ReturnPeersForGetAddrPayload_Scenario3()
...
 
-            var peers = peerAddressManager.PeerSelector.SelectPeersForGetAddrPayload(15);
+            IEnumerable<PeerAddress> peers = peerAddressManager.PeerSelector.SelectPeersForGetAddrPayload(15);
             Assert.Equal(15, peers.Count());
             Assert.Equal(7, peers.Count(p => p.Attempted));
             Assert.Equal(8, peers.Count(p => p.Fresh));
         }
 
         /// <summary>
+        /// <para>
```
dangershony: In the future so you know comments on unit-tests are optional

majikandy: Yep cool - I was just following the style of the tests already there.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class SelfEndpointTracker : ISelfEndpointTracker
+    {
+        /// <summary>Hashset to hold the endpoints currently known to be itself.</summary>
+        private readonly ConcurrentHashSet<IPEndPoint> knownSelfEndpoints = new ConcurrentHashSet<IPEndPoint>();
+
+        /// <inheritdoc/>
+        public void Add(IPEndPoint ipEndPoint)
+        {
+            this.knownSelfEndpoints.Add(ipEndPoint);
```
dangershony: Dont you need a TryAdd here?

majikandy: ConcurrentHashSet only uses Try for remove, Add will just add and if it is already there it doesn't matter, it can't really fail.

dangershony: Super ok let push

PR #1223, opened 2018-03-17
Replace R/W lock with AsyncLock in CachedCoinView

noescape00:
Fix: #1176


dangershony:
Ok we need to discuss this on conf call

Aprogiena:
Sure, any time

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -176,9 +170,8 @@ private CachedCoinView(IDateTimeProvider dateTimeProvider, ILoggerFactory logger
             UnspentOutputs[] outputs = new UnspentOutputs[txIds.Length];
             List<int> miss = new List<int>();
             List<uint256> missedTxIds = new List<uint256>();
-            using (this.lockobj.LockRead())
+            using (await this.lockobj.LockAsync().ConfigureAwait(false))
```
dangershony: I am ready to check this in, but suddenly wonder if its not such a bad idea to combine the lock.

PR #1225, opened 2018-03-17
Clean POS code in nbitcoin

dangershony:
The POS code in NBitcoin is not used anymore (its all moved to the fullnode) so this PR deletes that code (and tests)


src/NBitcoin/Block.cs
```diff
@@ -348,9 +348,6 @@ public void UpdateMerkleRoot()
         /// <returns></returns>
         public bool Check(Consensus consensus)
```
Aprogiena: I think you can completely delete this whole method

src/NBitcoin/BlockValidator.cs
```diff
@@ -1,940 +0,0 @@
...
-                var data = ops.ElementAt(1).PushData;
-                if (!ScriptEvaluationContext.IsCompressedOrUncompressedPubKey(data))
-                    return false;
-                return new PubKey(data).Verify(block.GetHash(), new ECDSASignature(block.BlockSignatur.Signature));
-            }
-
-            return false;
-        }
-
-        public static bool IsCanonicalBlockSignature(Block block, bool checkLowS)
```
Aprogiena: This is something that I found in stratisX and it is not explicitly used in C# node, so I was wondering if it is implicitly checked with the current signature verification. But seeing that here, maybe we are missing this check compared to stratisX

Aprogiena:
Looks awesome. I haven't checked if something was broken by this, no tests done, but it looks great. Just please consider my 2 comments.

src/NBitcoin/BlockValidator.cs
dangershony: Yeah we have a TODO rule in place for this so I did not forget.
https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs

That's why I feel its safe to delete from NBitcoin

src/NBitcoin/Block.cs
dangershony: Ah yeah good spot, its going away.

PR #1229, opened 2018-03-21
 Added methods to remove transactions

bokobza:
- added a method to remove specific unconfirmed transactions 
- added a method to remove all transactions (use with caution!) from a wallet

When called from the API, users also have the option to sync their wallet after the removal.
The time of sync is chosen automatically from the earliest removed transaction.


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1250,6 +1250,86 @@ public DateTimeOffset GetOldestWalletCreationTime()
...
 
+        /// <inheritdoc />
+        public HashSet<(uint256, DateTimeOffset)> RemoveTransactionsByIds(string walletName, IEnumerable<uint256> transactionsIds)
+        {
+            Guard.NotNull(transactionsIds, nameof(transactionsIds));
+            Guard.NotEmpty(walletName, nameof(walletName));
+
+            List<uint256> idsToRemove = transactionsIds.ToList();
+            Wallet wallet = this.GetWallet(walletName);
+            IEnumerable<HdAccount> accounts = wallet.GetAccountsByCoinType(this.coinType);
```
dangershony: This code `wallet.GetAccountsByCoinType()`  is inside the lock in other places so I wonder if it needs to be inside the lock as well

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1250,6 +1250,86 @@ public DateTimeOffset GetOldestWalletCreationTime()
...
+            return result;
+        }
+
+        /// <inheritdoc />
+        public HashSet<(uint256, DateTimeOffset)> RemoveAllTransactions(string walletName)
+        {
+            Guard.NotEmpty(walletName, nameof(walletName));
+
+            Wallet wallet = this.GetWallet(walletName);
+            IEnumerable<HdAccount> accounts = wallet.GetAccountsByCoinType(this.coinType);
```
dangershony: Same here check if it needs to be inside a lock

PR #1238, opened 2018-03-23
Spend coin straight after mining should fail

majikandy:


src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
```diff
@@ -0,0 +1,131 @@
...
+            var walletException = (WalletException)this.caughtException;
+            walletException.Message.Should().Be("No spendable transactions found.");
+
+            // reset for future checks
+            this.caughtException = null;
+        }
+
+        private void the_transaction_should_be_accepted_by_the_mempool()
+        {
+            this.caughtException.Should().BeNull();
```
dangershony: Also you can check if the trx is in the memppol, use the rpc method (or query the memppol directly) and get the list of trxids and compare it to our.

Same goes for the other method `the_transaction_should_be_rejected_from_the_mempool` check the mempool is empty

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
```diff
@@ -0,0 +1,131 @@
...
+using FluentAssertions;
+using NBitcoin;
+using Stratis.Bitcoin.Features.Consensus;
+using Stratis.Bitcoin.Features.Wallet;
+using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
+using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests.BlockStore
+{
+    public partial class ProofOfWorkSpendingSpecification
```
dangershony: Hmm this test seems to be very general `ProofOfWorkSpending` you might find that other cases need to create more then one nodes.
If this is a very specific specification then my preference is to keep related code in the same file as the test (maybe an internal class?).

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
```diff
@@ -0,0 +1,131 @@
...
+            var wallet = this.stratisBitcoinNode.FullNode.WalletManager().GetWalletByName("mywallet");
+
+            var extendedPrivateKey = wallet.GetExtendedPrivateKeyForAddress("123456", address).PrivateKey;
+
+            this.stratisBitcoinNode.SetDummyMinerSecret(new BitcoinSecret(extendedPrivateKey,
+                this.stratisBitcoinNode.FullNode.Network));
+
+            this.stratisBitcoinNode.GenerateStratisWithMiner(targetHeight);
+
+            // wait for block repo for block sync to work
```
dangershony: >wait for block store to sync (or catch-up)

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            // Build Transaction 
+            // ====================
+            // send coins to next self address
+            var sendtoAddress = this.stratisBitcoinNode.FullNode.WalletManager()
+                .GetUnusedAddresses(new WalletAccountReference("mywallet", "account 0"), 2).ElementAt(1);
+
+            try
+            {
+                this.stratisBitcoinNode.FullNode.WalletTransactionHandler().BuildTransaction(
```
dangershony: I think you need to build and send (not just build)

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
majikandy: Yep, will do

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSpecification_Steps.cs
majikandy: The separate partial is literally just to make the file you read most of the time (The specification) as clean as it possibly can be - the counter argument being hiding the code away in the steps file - but on balance I like as many tricks as possible to keep the readable file readable. Happy to hear others opinions on this though.

PR #1239, opened 2018-03-24
Fixing integration tests

dangershony:
With this PR the wallet tests are more stable, however more work will go in to this PR

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -210,7 +210,6 @@ namespace Stratis.Bitcoin.Features.Miner
                         services.AddSingleton<MinerController>();
                         services.AddSingleton<MiningRPCController>();
                         services.AddSingleton<MinerSettings>(new MinerSettings(setup));
-	                    services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
 					});
             });
 
@@ -241,9 +240,7 @@ namespace Stratis.Bitcoin.Features.Miner
                         services.AddSingleton<IAssemblerFactory, PosAssemblerFactory>();
                         services.AddSingleton<MinerController>();
                         services.AddSingleton<MiningRPCController>();
-                        services.AddSingleton<IWalletManager, WalletManager>();
                         services.AddSingleton<MinerSettings>(new MinerSettings(setup));
-                        services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
 					});
             });
 
--- a/src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
@@ -245,6 +245,7 @@ namespace Stratis.Bitcoin.IntegrationTests
 
                 // remove node 2
                 stratisNodeSync.CreateRPCClient().RemoveNode(stratisNode2.Endpoint);
+                TestHelper.WaitLoop(() => !TestHelper.IsNodeConnected(stratisNode2));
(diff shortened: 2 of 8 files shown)


src/Stratis.Bitcoin.IntegrationTests/TestHelper.cs
```diff
@@ -43,5 +44,11 @@ public static void TriggerSync(CoreNode node)
             foreach (var connectedPeer in node.FullNode.ConnectionManager.ConnectedPeers)
                 connectedPeer.Behavior<ChainHeadersBehavior>().TrySyncAsync().GetAwaiter().GetResult();
         }
+
+        public static bool IsNodeConnected(CoreNode node)
+        {
+            if (node.FullNode.ConnectionManager.ConnectedPeers.Any()) return true;
```
bokobza: You could just `return node.FullNode.ConnectionManager.ConnectedPeers.Any()`

src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
```diff
@@ -245,6 +245,7 @@ public void BlockStoreCanReorg()
 
                 // remove node 2
                 stratisNodeSync.CreateRPCClient().RemoveNode(stratisNode2.Endpoint);
+                TestHelper.WaitLoop(() => !TestHelper.IsNodeConnected(stratisNode2));
```
bokobza: What happens if the node doesn't disconnect at all?
Would it be a good idea to put a one-minute timeout with an error?

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -241,9 +240,7 @@ public static IFullNodeBuilder AddPowPosMining(this IFullNodeBuilder fullNodeBui
                         services.AddSingleton<IAssemblerFactory, PosAssemblerFactory>();
                         services.AddSingleton<MinerController>();
                         services.AddSingleton<MiningRPCController>();
-                        services.AddSingleton<IWalletManager, WalletManager>();
```
bokobza: I think you need the walletManager for mining, don't you?

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -60,6 +60,8 @@ public void CanStratisSyncFromCore()
 
                 //Now check if Core connect to stratis
                 stratisNode.CreateRPCClient().RemoveNode(coreNode.Endpoint);
+                TestHelper.WaitLoop(() => coreNode.CreateRPCClient().GetPeersInfo().Length == 0);
```
bokobza: Why not use `TestHelper.IsNodeConnected(coreNode)`?

src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
dangershony: Yeah The method WaitLoop will throw after a timeout

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
dangershony: Well WalletFeature is a dependency so it should already be registered in the DI so this call is actually redundant

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
dangershony: Yeah I tried then realized this is a bitcoin core node.

bokobza: ah right, fair enough 😆

PR #1241, opened 2018-03-25
Mempool tests moving to mempool project

dangershony:


bokobza:
Are these just unit tests?

dangershony:
Yes just tests that are not heavy and can run on appvoyar so I moved them out of integration tests

Issue #1228, opened 2018-03-21
[Wallet Bug] consensus stopes when the wallet looks for a block in store


dangershony:
![image](https://user-images.githubusercontent.com/7487930/37716112-a0384876-2d15-11e8-8857-695b8afb9322.png)

![image](https://user-images.githubusercontent.com/7487930/37716086-908461c6-2d15-11e8-82e0-9f66cd019e9e.png)

Aprogiena:
how do we know this is wallet related? or do we know the cause already?

dangershony:
I think this is not an issue anymore.

PR #1244, opened 2018-03-26
Wallet performance improvements numero uno

bokobza:
- Improved performance of when and how the lookup dictionary for addresses is loaded
- Added look-up on unspent outputs
- Changed from saving the wallet after every change to saving the wallet when
 1. a transaction is received from the network
 2. a block is done being processed and had some transactions affecting the wallet within it

- no benchmarking was done but the gain in speed is noticeable.

Kudos to @noescape00 for creating blocks with crazy transactions.
For reference, take a look at these blocks on Stratis Testnet:
333235
333445
333449
333946
334042
334044
334046
334048
334051
334052
334055
334056
334081
334084
334088
334094
334100

@dangershony

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
@@ -97,9 +97,11 @@ namespace Stratis.Bitcoin.Features.Wallet
 
         public uint256 WalletTipHash { get; set; }
 
-        // TODO: a second lookup dictionary is proposed to lookup for spent outputs
-        // every time we find a trx that credits we need to add it to this lookup
-        // private Dictionary<OutPoint, TransactionData> outpointLookup;
+        // In order to allow faster look-ups of transactions affecting the wallets' addresses,
+        // we keep a couple of objects in memory:
+        // 1. the list of unspent outputs for checking whether inputs from a transaction are being spent by our wallet and
+        // 2. the list of addresses contained in our wallet for checking whether a transaction is being paid to the wallet.
+        private Dictionary<OutPoint, TransactionData> outpointLookup;
         internal Dictionary<Script, HdAddress> keysLookup;
 
         public WalletManager(
@@ -147,6 +149,7 @@ namespace Stratis.Bitcoin.Features.Wallet
             }
 
             this.keysLookup = new Dictionary<Script, HdAddress>();
+            this.outpointLookup = new Dictionary<OutPoint, TransactionData>();
         }
 
         private void BroadcasterManager_TransactionStateChanged(object sender, TransactionBroadcastEntry transactionEntry)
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1194,23 +1181,38 @@ public void LoadKeysLookupLock()
...
+                            this.keysLookup[address.Pubkey] = address;
                     }
                 }
+            }
+        }
+
+        /// <summary>
+        /// Update the keys and transactions we're tracking in memory for faster lookups.
+        /// </summary>
+        public void UpdateKeysLookupLock(IEnumerable<HdAddress> addresses)
```
dangershony: I would suggest (for the sake of avoiding double locks) to make this methods private `UpdateKeysLookupLock` `LoadKeysLookupLock` (if needed for unit testing then a public Internal method) and remove the lock from inside then make sure the callers are holding the lock, there are only 6 places where this method is called.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -791,9 +794,13 @@ public bool ProcessTransaction(Transaction transaction, int? blockHeight = null,
                 }
 
                 // Check the inputs - include those that have a reference to a transaction containing one of our scripts and the same index.
-                foreach (TxIn input in transaction.Inputs.Where(txIn => this.keysLookup.Values.Distinct().SelectMany(v => v.Transactions).Any(trackedTx => trackedTx.Id == txIn.PrevOut.Hash && trackedTx.Index == txIn.PrevOut.N)))
+                foreach (TxIn input in transaction.Inputs)
                 {
-                    TransactionData tTx = this.keysLookup.Values.Distinct().SelectMany(v => v.Transactions).Single(trackedTx => trackedTx.Id == input.PrevOut.Hash && trackedTx.Index == input.PrevOut.N);
+                    this.outpointLookup.TryGetValue(input.PrevOut, out TransactionData tTx);
```
dangershony: Why not?
```
if( !this.outpointLookup.TryGetValue(input.PrevOut, out TransactionData tTx)
{
   continue;
}
```

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1215,6 +1221,12 @@ public void LoadKeysLookupLock()
                         this.keysLookup[address.ScriptPubKey] = address;
                         if (address.Pubkey != null)
                             this.keysLookup[address.Pubkey] = address;
+
+                        foreach (var unspentTransaction in address.UnspentTransactions())
+                        {
+                            this.outpointLookup[new OutPoint(unspentTransaction.Id, unspentTransaction.Index)] = unspentTransaction;
+                        }
+
```
dangershony: space

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
dangershony: Leave this to the wallet refactor when we make it stand alone I think this is good as it is now.

dangershony:
I think this will also fix: #1228

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: yep

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: :space_invader:

bokobza:
This is the first batch of improvements, more will come in the near future.

noescape00:
Given that wallet is being updated every time it's changed do we need the async loop that saves it every 5 minutes? 
WalletManager.cs:177

bokobza:
@noescape00 yeah this is to update the lastBlockSyncHeight and lastBlockSyncHash, even if there was no change in the wallet's addresses.

noescape00:
Ah, makes perfect sense.

PR #1245, opened 2018-03-26
[Tests] Build a POS/POW integration test

fassadlr:
As part of this task I also refactored / cleaned up the NodeRunners.

For the POS/POW integration test, we first premine 98mil coins which we then use to stake with. The test passes once some new coins has been minted and the staking wallet's balance updated.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
@@ -210,8 +210,8 @@ namespace Stratis.Bitcoin.Features.Miner
                         services.AddSingleton<MinerController>();
                         services.AddSingleton<MiningRPCController>();
                         services.AddSingleton<MinerSettings>(new MinerSettings(setup));
-	                    services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
-					});
+                        services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
+                    });
             });
 
             return fullNodeBuilder;
@@ -244,7 +244,7 @@ namespace Stratis.Bitcoin.Features.Miner
                         services.AddSingleton<IWalletManager, WalletManager>();
                         services.AddSingleton<MinerSettings>(new MinerSettings(setup));
                         services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
-					});
+                    });
             });
 
             return fullNodeBuilder;
--- a/src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
@@ -4,7 +4,6 @@ using System.Diagnostics;
 using System.Linq;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
-using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Features.BlockStore;
 using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
 using Stratis.Bitcoin.Utilities;
@@ -148,9 +147,9 @@ namespace Stratis.Bitcoin.IntegrationTests
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNodeSync = builder.CreateStratisPowNode();
-                var stratisNode1 = builder.CreateStratisPowNode();
-                var stratisNode2 = builder.CreateStratisPowNode();
+                var stratisNodeSync = builder.CreateStratisBitcoinPowNode();
+                var stratisNode1 = builder.CreateStratisBitcoinPowNode();
+                var stratisNode2 = builder.CreateStratisBitcoinPowNode();
                 builder.StartAll();
                 stratisNodeSync.NotInIBD();
                 stratisNode1.NotInIBD();
@@ -190,7 +189,7 @@ namespace Stratis.Bitcoin.IntegrationTests
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNodeSync = builder.CreateStratisPowNode();
+                var stratisNodeSync = builder.CreateStratisBitcoinPowNode();
                 builder.StartAll();
                 stratisNodeSync.NotInIBD();
 
@@ -206,7 +205,7 @@ namespace Stratis.Bitcoin.IntegrationTests
                 // stop the node it will persist the chain with the reset tip
                 stratisNodeSync.FullNode.Dispose();
 
-                var newNodeInstance = builder.CloneStratisNode(stratisNodeSync);
+                var newNodeInstance = builder.CloneStratisBitcoinPowNode(stratisNodeSync);
 
                 // load the node, this should hit the block store recover code
                 newNodeInstance.Start();
@@ -222,9 +221,9 @@ namespace Stratis.Bitcoin.IntegrationTests
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNodeSync = builder.CreateStratisPowNode();
-                var stratisNode1 = builder.CreateStratisPowNode();
-                var stratisNode2 = builder.CreateStratisPowNode();
+                var stratisNodeSync = builder.CreateStratisBitcoinPowNode();
+                var stratisNode1 = builder.CreateStratisBitcoinPowNode();
+                var stratisNode2 = builder.CreateStratisBitcoinPowNode();
                 builder.StartAll();
                 stratisNodeSync.NotInIBD();
                 stratisNode1.NotInIBD();
@@ -273,8 +272,8 @@ namespace Stratis.Bitcoin.IntegrationTests
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNode1 = builder.CreateStratisPowNode();
-                var stratisNode2 = builder.CreateStratisPowNode();
+                var stratisNode1 = builder.CreateStratisBitcoinPowNode();
+                var stratisNode2 = builder.CreateStratisBitcoinPowNode();
                 builder.StartAll();
                 stratisNode1.NotInIBD();
                 stratisNode2.NotInIBD();
--- a/src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
@@ -172,9 +172,9 @@ namespace Stratis.Bitcoin.IntegrationTests
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNode = builder.CreateStratisPowNode();
-                var coreNode1 = builder.CreateNode();
-                var coreNode2 = builder.CreateNode();
+                var stratisNode = builder.CreateStratisBitcoinPowNode();
+                var coreNode1 = builder.CreateBitcoinNode();
+                var coreNode2 = builder.CreateBitcoinNode();
                 builder.StartAll();
 
                 //Core1 discovers 10 blocks, sends to stratis
--- a/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
@@ -12,7 +12,6 @@ using NBitcoin.DataEncoders;
 using NBitcoin.Protocol;
 using NBitcoin.RPC;
 using Stratis.Bitcoin.Configuration.Logging;
-using Stratis.Bitcoin.Features.Consensus;
 using Stratis.Bitcoin.Features.MemoryPool;
 using Stratis.Bitcoin.Features.Miner;
 using Stratis.Bitcoin.Features.Miner.Interfaces;
@@ -76,13 +75,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
         /// <summary>Get stratis full node if possible.</summary>
         public FullNode FullNode
         {
-            get
-            {
-                if (this.runner is StratisBitcoinPosRunner)
-                    return ((StratisBitcoinPosRunner)this.runner).FullNode;
-
-                return ((StratisBitcoinPowRunner)this.runner).FullNode;
-            }
+            get { return this.runner.FullNode; }
         }
 
         private void CleanFolder()
@@ -90,7 +83,7 @@ namespace Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers
             NodeBuilder.CleanupTestFolder(this.Folder);
         }
 
-        public void Sync(CoreNode node, bool keepConnection = false)
+        public void SyncFrom(CoreNode node, bool keepConnection = false)
         {
             var rpc = this.CreateRPCClient();
             var rpc1 = node.CreateRPCClient();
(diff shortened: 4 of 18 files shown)


src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -210,7 +210,8 @@ public static IFullNodeBuilder AddMining(this IFullNodeBuilder fullNodeBuilder,
                         services.AddSingleton<MinerController>();
                         services.AddSingleton<MiningRPCController>();
                         services.AddSingleton<MinerSettings>(new MinerSettings(setup));
-					});
+                        services.AddSingleton<ITimeSyncBehaviorState, TimeSyncBehaviorState>();
```
dangershony: This was deleted intentionally

src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
```diff
@@ -148,9 +147,9 @@ public void BlockBroadcastInv()
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNodeSync = builder.CreateStratisPowNode();
-                var stratisNode1 = builder.CreateStratisPowNode();
-                var stratisNode2 = builder.CreateStratisPowNode();
+                var stratisNodeSync = builder.CreateStratisBitcoinPowNode();
```
dangershony: hmm this name is problematic `CreateStratisBitcoinPowNode` we need to come up with a better way to distinguish the framework name and the token.

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
fassadlr: Sorry, yeah fixed now.

src/Stratis.Bitcoin.IntegrationTests/BlockStoreTests.cs
fassadlr: I agree @dangershony ! I'll name the nodes with more intent.

fassadlr: OK should be better now :)

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
```diff
@@ -172,9 +172,9 @@ public void CanHandleReorgs()
         {
             using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var stratisNode = builder.CreateStratisPowNode();
-                var coreNode1 = builder.CreateNode();
-                var coreNode2 = builder.CreateNode();
+                var stratisNode = builder.CreateBitcoinPowNode();
+                var coreNode1 = builder.CreateBitcoinCoreNode();
+                var coreNode2 = builder.CreateBitcoinCoreNode();
```
dangershony: hmm so your solution is to default to a statis (c#) node and if we use core then just name it core hmmmm...

How about we actually also use core  for stratis and core will be the code name for the tech.
Then the token can follow


so it will be 
CreateStratisCoreBitcoinNode
CreateBitcoinCoreBitcoinNode

and for the stratis network 
CreateStratisCoreStratisNode
CreateBitcoinCoreStratisNode (this is invalid ofcourse unless we actually also use stratisx)

WDYT

@Aprogiena @bokobza thoughts?

dangershony: I would even go one step further and renaming the namespace to `Stratiscore.Bitcoin` instead of Stratis.Bitcoin`

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -76,21 +75,15 @@ public CoreNode(string folder, INodeRunner runner, NodeBuilder builder, Network
...
         public FullNode FullNode
         {
-            get
-            {
-                if (this.runner is StratisBitcoinPosRunner)
-                    return ((StratisBitcoinPosRunner)this.runner).FullNode;
-
-                return ((StratisBitcoinPowRunner)this.runner).FullNode;
-            }
+            get { return this.runner.FullNode; }
```
dangershony: ah ofcourse

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -105,9 +104,10 @@ public class NodeBuilder : IDisposable
 
         public NodeConfigParameters ConfigParameters { get; }
 
+        private List<IDisposable> disposables;
         private int last;
         private string root;
-        private List<IDisposable> disposables;
+        private bool skipRules = false;
```
dangershony: why do you need this property?

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -29,6 +32,12 @@ public BitcoinCoreRunner(string bitcoinD)
 
         private Process process;
 
+        public FullNode FullNode
+        {
+            get { return null; }
```
dangershony: `get { return null; }` hu?

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
Aprogiena: I am missing context, not sure what the question is ...

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -210,15 +276,34 @@ public static IFullNodeBuilder MockIBD(this IFullNodeBuilder fullNodeBuilder)
...
+
+                        services.AddSingleton<IInitialBlockDownloadState, InitialBlockDownloadStateMock>();
+                    });
+                }
+            });
+
+            return fullNodeBuilder;
+        }
+
+        public static IFullNodeBuilder MockRules(this IFullNodeBuilder fullNodeBuilder, IRuleRegistration ruleRegistration)
```
dangershony: Hu what? why do you need to replace rules?

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
fassadlr: @dangershony I reckon the node runners were initially called StratisBitcoinPowNode to follow the daemon's naming convention.

I only named it BitcoinCore because its the only runner which starts up a process too execute the Bitcoin.exe, the rest are all effectively Stratis nodes.

I think we should go with:

CreateBitcoinCore (Starts up the Bitcoin.exe in a process)
CreateStratisCorePowNode (Starts up a PoW Stratis node)
CreateStratisCorePosNode (Starts up a PoS Stratis node)
CreateStratisBitcoinPowNode (Starts up a Stratis node mining bitcoin)

src/Stratis.Bitcoin.IntegrationTests/PoSMintingTests.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace Stratis.Bitcoin.IntegrationTests
+{
+    public sealed class PoSMintingTests
+    {
+        [Fact]
+        public void ProofOfStake_MintNewCoinsViaStaking()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create())
+            {
+                var nodeGenerateCoins = builder.SkipRules().CreateStratisPowNode();
```
dangershony: This almost defeats the purpose of the test no?

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeBuilder.cs
fassadlr: It is used to determine whether or not the node should skip consensus rules. Do a find all references, you;ll see that we can now set up the node in the integration tests like that.

`var node = nodeBuilder.SkipRules().CreateStratisPosNode()'

We then inject a fake rule registration class.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: As part of the refactor I wanted all the runners to implement the interface, but because bitcoin core starts up the bitcoin.exe we don't have a FullNode. I can change this to throw a `NotImplementedException`?

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: We cant mine x blocks with the ACTUAL miner as the block times are too short after each other, so consensus fails when we use  `GenerateStratisWithMiner`. This way we can inject the rules in cases where want to test mining 100 blocks for example but we dont want to wait 100 minutes.

src/Stratis.Bitcoin.IntegrationTests/PoSMintingTests.cs
fassadlr: No, see comment above, this is is to quickly generate x amount of blocks. Its either this or we have to use the now obsolete `GenerateStratis` method.

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
dangershony: >I am missing context
We currently start two type of codebase the C++ bitcoincore and the C# stratis 
But we also have distinction of token, all of our tests currently use the bitcoin network, but we will be wanting to write tests for the stratis network as well.

But then again the distinction is more about POW/POS so fassa is sort of correct that we need to make the distinction like that.

But this is not so correct 
>CreateStratisCorePowNode (Starts up a PoW Stratis node)
>CreateStratisCorePosNode (Starts up a PoS Stratis node)

A POS node is always also a POW node how you choose to mine is up to you.

CreateBitcoincorePowNode (c++ on bitcoin network)
CreateStratiscorePowNode  (c# node bitcoin network)
CreateStratiscorePosNode   (c# node stratis network)

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
dangershony: ah yeah thats core, yeah throw is better.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
dangershony: Hmm I think this is the wrong approach, in POW there is a flag to ignore difficulty, probably this is the same thing we need to do for POS how ever how exactly I am not sure.
We can add a flag to the code that checks the next target and reduce the target spacing requirements to be much lower, so effectively not have to wait so long between blocks.

@Aprogiena WDYT how can we achieve fast POS blocks in stratisregtest

Aprogiena: @dangershony pls ask me tomorrow and don't forget about it :)

dangershony: yeah good point, @fassadlr tomorrow I met apro we will discuss how we can make the POS mining faster while we still maintain the rules rules.

src/Stratis.Bitcoin.IntegrationTests/PoSMintingTests.cs
dangershony: If I understand correctly you are removing all the rules apart for one.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -135,65 +154,112 @@ public static IFullNode BuildStakingNode(string dir, bool staking = true)
...
+    }
 
-            if (callback != null)
-            {
-                var builder = new FullNodeBuilder().UseNodeSettings(args);
+    public sealed class PosTestRuleRegistration : IRuleRegistration
+    {
+        public IEnumerable<ConsensusRule> GetRules()
+        {
+            return new List<ConsensusRule>() { new PosTestRule() };
```
dangershony: There is only one rule added, so this means all other rules are not run then the measurement of the test is not so accurate.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: @dangershony if we can get around this by changing the network's block time (or completely zero-ing) it in StratisRegTest, then this will work really great.

dangershony: So the solution is as following:
POS blocks expect a block every 64 seconds, so we can use a new DateTimeProvider and override the time to be 1 day ago, then every time a POS block is found the datetime will increment the current time bu 64 seconds and that will allow creating many POS blocks very fast.

So we should not remove any rules at all for this to work.

Aprogiena: > 1 day ago

Or more, it can even be one week or one month

fassadlr: @dangershony OK, this is a good solution! I'll implement a custom DTP for this (the) tests. I will do this on this PR so that we keep record of the discussion.

fassadlr:
@dangershony @Aprogiena So i implemented a custom DTP for block creation and whilst we can set the block creation time to whatever we want via this, we cannot get past this rule:

`ConsensusErrors.BlockTimeBeforeTrx.Throw();`

as the Transaction's time stamp is set on construction.

Some suggestions to get around this:

1> Re-implement overriding the rules on block creation/validation in unit tests.
2> Move Transaction out of NBitcoin to Stratis so that we can inject a DTP into it.
3> Put an interface on Transaction.
4> Set Transaction's timestamp in the block assembler via DTP.

I think number 4 might be the easiest?

fassadlr:
Actually, number 4 might work if we do it here? In PowBlockAssembler:

```
protected virtual void CreateCoinbase()
        {
            // Create coinbase transaction.
            // Set the coin base with zero money.
            // Once we have the fee we can update the amount.
            this.coinbase = new Transaction();
            this.coinbase.Time = (uint)this.dateTimeProvider.GetTimeOffset().ToUnixTimeSeconds();
            this.coinbase.AddInput(TxIn.CreateCoinbase(this.ChainTip.Height + 1));
            this.coinbase.AddOutput(new TxOut(Money.Zero, this.scriptPubKeyIn));
```

Aprogiena:
` (uint)this.dateTimeProvider.GetTimeOffset().ToUnixTimeSeconds();`

don't use `GetTimeOffset` (like ever if possible, use `GetUtcNow` instead if you really don't need offset time)

but here, `GetTime` should be used which already gives you unix timestamp


Otherwise yes, transaction time is by default set with out date time provider as it is done in Nbitcoin and the suggested modification might help.

dangershony:
Good solution yes, we are getting closer.
@fassadlr I would also suggest if you have time look at the new way we write integration tests and perhaps migrate to that.

will:
bitcoin is dumb though so

Issue #1250, opened 2018-03-26
[RF] Transaction relaying is allowed once in 20 years


noescape00:
So we have the `SendTrickleAsync` in `MempoolBehavior` which relays the transactions to the peers. Inside it it has 

```
  if (!this.CanSend)
            {
                this.logger.LogTrace("(-)[NO_SEND]");
                return;
            }
```

And the getter for the CanSend has this: 

```
if (this.NextInvSend < this.manager.DateTimeProvider.GetTime())
                {
                    sendTrickle = true;
                    // Use half the delay for outbound peers, as there is less privacy concern for them.
                    this.NextInvSend = this.manager.DateTimeProvider.GetTime() + TimeSpan.TicksPerMinute;
                }
```

`NextInvSend` is the next timestamp after which we can relay the tx. It is a unix timestamp. 
So this line
` this.NextInvSend = this.manager.DateTimeProvider.GetTime() + TimeSpan.TicksPerMinute;`
is:

NextUnixTime = CurrentUnixTime + 600000000.

600000000 seconds is ~20 years.

dangershony:
Yeah if we already have a delay in the loop we don't need another delay logic in `CanSend` I agree.
And I think this two elements can become ConcurrentDictionaries  `inventoryTxToSend` and `filterInventoryKnown` then we only use the mempool lock when calling the mempool.

dangershony:
@Aprogiena had a good comment on the closed PR us the existing constant

dangershony:
This issue is going to be used for rarefaction mempool behaviour as per comments above.

mikedennis:
@dangershony @noescape00 Just to confirm what's involved here:

In mempool behaviour replace locking done with ```this.manager.MempoolLock.WriteAsync(()```

With individuals locks around access to:
```
this.inventoryTxToSend
this.filterInventoryKnown
```
Question: Are there any existing tests that will test this?

noescape00:
>In mempool behaviour replace locking done with this.manager.MempoolLock.WriteAsync(()
>With individuals locks around access to:

Yes. But also this issue suggests fixing general design problems: 

1) We have 2 different delays- one in async loop and one in CanSend
2) CanSend getter actually changes the value which is not an expected behavior 
3) Potentially the same logic can be used for tx relaying and tx propagation (but that might be a separated task since it's a big one)

mikedennis:
[...]

Issue #1242, opened 2018-03-26
Transactions are not being relayed by the FullNode


maciejzaleski:
**Setup**
- RelayNode is connected to the network
- OtherNode is connected to RelayNode only

When sending transaction from OtherNode to RelayNode the transaction is received by the RelayNode but it is not propagated to any other node on the network.

dangershony:
@noescape00 you on this right?

noescape00:
Yes. Looks like there are several issues that prevent relaying the transactions in the setup that @maciejzaleski has

PR #1251, opened 2018-03-26
Fix: Transaction relaying is allowed once in 20 years 

noescape00:
Changed delay from 600000000 seconds to 10 seconds (10 seconds constant was used here because in MempoolSignaled we have an AsyncLoop with that constant that calls SendTrickleAsync() which uses CanSend)

@maciejzaleski has confirmed that this fix resolves the problem for him that was reported in #1242

Fix: #1250
Fix: #1242

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -140,7 +140,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
                 {
                     sendTrickle = true;
                     // Use half the delay for outbound peers, as there is less privacy concern for them.
-                    this.NextInvSend = this.manager.DateTimeProvider.GetTime() + TimeSpan.TicksPerMinute;
+                    this.NextInvSend = this.manager.DateTimeProvider.GetTime() + 10;
                     // TODO: PoissonNextSend(nNow, InventoryBroadcastInterval >> !pto->fInbound);
                 }
 


dangershony:
Arghh wowo good catch, I wonder now if 10 seconds is not very little.

dangershony:
On core I see this code

```
            // Check whether periodic sends should happen
            bool fSendTrickle = pto->fWhitelisted;
            if (pto->nNextInvSend < nNow) {
                fSendTrickle = true;
                // Use half the delay for outbound peers, as there is less privacy concern for them.
                pto->nNextInvSend = PoissonNextSend(nNow, INVENTORY_BROADCAST_INTERVAL >> !pto->fInbound);
            }

            // Time to send but the peer has requested we not relay transactions.
            if (fSendTrickle) {
                LOCK(pto->cs_filter);
                if (!pto->fRelayTxes) pto->setInventoryTxToSend.clear();
            }

```
```
int64_t PoissonNextSend(int64_t nNow, int average_interval_seconds) {
    return nNow + (int64_t)(log1p(GetRand(1ULL << 48) * -0.0000000000000035527136788 /* -1/2^48 */) * average_interval_seconds * -1000000.0 + 0.5);
}

```

dangershony:
```
/** Average delay between trickled inventory transmissions in seconds.
 *  Blocks and whitelisted receivers bypass this, outbound peers get half this delay. */
static const unsigned int INVENTORY_BROADCAST_INTERVAL = 5;
```

Aprogiena:
Well, 1st of all, there is no need for this to be stored as a UNIX timestamp. Data time should be used. 2nd, we should have such a constant defined, not using hardcoded values. And then we can ask how often… So then if bitcoin core uses 5 seconds, then 5 seconds or 10 seconds should be fine.

So I can actually see `InventoryBroadcastInterval` constantly defined in the code, so I don't understand, why this fix does not use it…

Issue #1107, opened 2018-01-27
Wallet.GetSpendableTransactions is not threadsafe


noescape00:
`GetSpendableTransactions` is used by PeriodicLog (WalletFeature (or light wallet) => AddFeatureStats => GetSpendableTransactionsInWallet). So if we start calculating confirmed balance in the periodic log and at the same time we send the tx (which updates UTXO set that is used by `GetSpendableTransactionsInWallet`) we get the following unhanded exception:

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.ConcatIterator`1.MoveNext()
   at Stratis.Bitcoin.Features.Wallet.HdAccount.<GetSpendableTransactions>d__40.MoveNext() in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Wallet\Wallet.cs:line 621
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.SparseArrayBuilder`1.ReserveOrAdd(IEnumerable`1 items)
   at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Stratis.Bitcoin.Features.Wallet.WalletManager.GetSpendableTransactionsInWallet(String walletName, Int32 confirmations) in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Wallet\WalletManager.cs:line 641
   at Stratis.Bitcoin.Features.Wallet.WalletFeature.AddFeatureStats(StringBuilder benchLog) in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin.Features.Wallet\WalletFeature.cs:line 99
   at Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__73_0(CancellationToken cancellation) in C:\Users\user\Desktop\StratisRepository\src\Stratis.Bitcoin\FullNode.cs:line 232
[...]

PR #1253, opened 2018-03-27
Move get balance to the wallet manager

bokobza:
Moved the retrieval of balances to the wallet manager.
This is to avoid having the controller directly access the wallet objects, and to avoid race conditions.

Sorry about the extra commits appearing here.
The changes are in commit 91acda5.


bokobza:
Fixes https://github.com/stratisproject/FullNodeUI/issues/39

bokobza:
Fixes https://github.com/stratisproject/StratisBitcoinFullNode/issues/1107

PR #1259, opened 2018-03-27
Utilised test spec for txn input & output wallet integration tests.

ferdeen:
* Moved Integration wallet tests into a new folder

* Completed Wallet_can_send_and_receive_many_transaction_inputs_and_ouputs test

* Refactored above test and Wallet_can_send_one_transaction_with_many_outputs to use the test specification pattern and use fluent assertions.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/Program.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Program.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Security.Cryptography;
+using Stratis.Bitcoin.IntegrationTests.Wallet;
 using Xunit;
 
 namespace Stratis.Bitcoin.IntegrationTests
@@ -12,6 +12,7 @@ namespace Stratis.Bitcoin.IntegrationTests
         public static void Main(string[] args)
         {
             RunAllTestsOf<WalletTests>();
+            RunAllTestsOf<WalletTestsManyTxnInputsAndOutputsSpecification>();
             RunAllTestsOf<NodeSyncTests>();
         }
 
--- a/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
@@ -1,0 +1,25 @@
+﻿using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class WalletTestsManyTxnInputsAndOutputsSpecification 
+    {
+        [Fact]
+        public void Wallet_can_send_one_transaction_with_many_outputs()
+        {
+            Given(a_stratis_sender_and_receiver_node_and_their_wallets);
+            And(a_block_is_minded);
+            Then(many_transaction_inputs_go_to_the_sender);
+        }
+
+        [Fact]
+        public void Wallet_can_send_and_recieve_many_transaction_inputs_and_ouputs()
+        {
+            Given(a_stratis_sender_and_receiver_node_and_their_wallets);
+            And(a_block_is_minded);
+            When(many_transaction_inputs_go_to_the_sender);
+            When(many_transaction_outputs_go_back_to_the_receiver);
+            Then(the_receiver_has_many_inputs);
+        }
+    }
+}
--- a/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
@@ -5,17 +5,16 @@ using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
 using Microsoft.Extensions.DependencyInjection;
 using NBitcoin;
 using Stratis.Bitcoin.Features.Consensus;
-using Stratis.Bitcoin.Features.Wallet;
 using Stratis.Bitcoin.Features.Wallet.Controllers;
 using Stratis.Bitcoin.Features.Wallet.Interfaces;
 using Stratis.Bitcoin.Features.Wallet.Models;
 using Xunit;
 using System;
-using Stratis.Bitcoin.Utilities;
-using System.Text;
 
-namespace Stratis.Bitcoin.IntegrationTests
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
 {
+    using Stratis.Bitcoin.Features.Wallet;
+
     public class WalletTests : IDisposable
     {
         private bool initialBlockSignature;
@@ -92,193 +91,7 @@ namespace Stratis.Bitcoin.IntegrationTests
 
                 TestHelper.WaitLoop(() => maturity + 6 == stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
             }
-        }
-
-        [Fact]
-        public void WalletCanSendOneTransactionWithManyOutputs()
-        { 
-            using (NodeBuilder builder = NodeBuilder.Create())
-            {
-                CoreNode stratisSender = builder.CreateStratisPowNode();
-                CoreNode stratisReceiver = builder.CreateStratisPowNode();
-
-                builder.StartAll();
-                stratisSender.NotInIBD();
-                stratisReceiver.NotInIBD();
-
-                stratisSender.FullNode.WalletManager().CreateWallet("123456", "mywallet");
-                stratisReceiver.FullNode.WalletManager().CreateWallet("123456", "mywallet");
-
-                HdAddress addr = stratisSender.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
-                Wallet wallet = stratisSender.FullNode.WalletManager().GetWalletByName("mywallet");
-                Key key = wallet.GetExtendedPrivateKeyForAddress("123456", addr).PrivateKey;
-
-                stratisSender.SetDummyMinerSecret(new BitcoinSecret(key, stratisSender.FullNode.Network));
-                int maturity = (int)stratisSender.FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
-                stratisSender.GenerateStratisWithMiner(maturity + 51);
-
-                // Wait for block repo for block sync to work.
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-                
-                Assert.Equal(Money.COIN * 150 * 50, stratisSender.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount));
-
-                // Sync both nodes.
-                stratisSender.CreateRPCClient().AddNode(stratisReceiver.Endpoint, true);
-                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisReceiver, stratisSender));
-
-                // Get 50 unused addresses from the receiver.
-                IEnumerable<HdAddress> recevierAddresses = stratisReceiver.FullNode.WalletManager()
-                    .GetUnusedAddresses(new WalletAccountReference("mywallet", "account 0"), 50);
-
-                List<Recipient> recipients = recevierAddresses.Select(address => new Recipient
-                    {
-                        ScriptPubKey = address.ScriptPubKey,
-                        Amount = Money.COIN
-                    })
-                    .ToList();
-
-                var transactionBuildContext = new TransactionBuildContext(
-                    new WalletAccountReference("mywallet", "account 0"), recipients, "123456")
-                    {
-                        FeeType = FeeType.Medium,
-                        MinConfirmations = 101
-                    };
-
-                Transaction transaction = stratisSender.FullNode.WalletTransactionHandler().BuildTransaction(transactionBuildContext);
-                Assert.Equal(51, transaction.Outputs.Count);
-
-                // Broadcast to the other node.
-                stratisSender.FullNode.NodeService<WalletController>().SendTransaction(new SendTransactionRequest(transaction.ToHex()));
-
-                // Wait for the trx's to arrive.
-                TestHelper.WaitLoop(() => stratisReceiver.CreateRPCClient().GetRawMempool().Length > 0);
-                TestHelper.WaitLoop(() => stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Any());
-
-                Assert.Equal(Money.COIN * 50, stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount));
-                Assert.Null(stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);  
-
-                // Generate new blocks so the trx is confirmed.
-                stratisSender.GenerateStratisWithMiner(1);
-
-                // Wait for block repo for block sync to work.
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisReceiver, stratisSender));
-
-                // Confirm trx's have been committed to the block.
-                Assert.Equal(maturity + 52 , stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
-            }
-        }
-
-        [Fact]
-        public void Wallet_can_send_one_transaction_with_many_outputs_then_send_many_inputs_to_one_transaction()
-        {
-            using (NodeBuilder builder = NodeBuilder.Create())
-            {
-                CoreNode stratisSender = builder.CreateStratisPowNode();
-                CoreNode stratisReceiver = builder.CreateStratisPowNode();
-
-                builder.StartAll();
-                stratisSender.NotInIBD();
-                stratisReceiver.NotInIBD();
-
-                stratisSender.FullNode.WalletManager().CreateWallet("123456", "mywallet");
-                stratisReceiver.FullNode.WalletManager().CreateWallet("123456", "mywallet");
-
-                HdAddress addrSender = stratisSender.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
-                Wallet walletSender = stratisSender.FullNode.WalletManager().GetWalletByName("mywallet");
-                Key keySender = walletSender.GetExtendedPrivateKeyForAddress("123456", addrSender).PrivateKey;
-
-                stratisSender.SetDummyMinerSecret(new BitcoinSecret(keySender, stratisSender.FullNode.Network));
-                int maturity = (int)stratisSender.FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
-                stratisSender.GenerateStratisWithMiner(maturity + 51);
-
-                // Wait for block repo for block sync to work.
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-
-                Assert.Equal(Money.COIN * 150 * 50, stratisSender.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount));
-
-                // Sync both nodes.
-                stratisSender.CreateRPCClient().AddNode(stratisReceiver.Endpoint, true);
-                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisReceiver, stratisSender));
-
-                // Get 50 unused addresses from the receiver.
-                IEnumerable<HdAddress> recevierAddresses = stratisReceiver.FullNode.WalletManager()
-                    .GetUnusedAddresses(new WalletAccountReference("mywallet", "account 0"), 50);
-
-                List<Recipient> recipients = recevierAddresses.Select(address => new Recipient
-                {
-                    ScriptPubKey = address.ScriptPubKey,
-                    Amount = Money.COIN
-                }).ToList();
-
-                var transactionBuildContext = new TransactionBuildContext(
-                    new WalletAccountReference("mywallet", "account 0"), recipients, "123456")
-                {
-                    FeeType = FeeType.Medium,
-                    MinConfirmations = 101
-                };
-
-                Transaction transaction = stratisSender.FullNode.WalletTransactionHandler().BuildTransaction(transactionBuildContext);
-                Assert.Equal(51, transaction.Outputs.Count);
-
-                // Broadcast to the other node.
-                stratisSender.FullNode.NodeService<WalletController>().SendTransaction(new SendTransactionRequest(transaction.ToHex()));
-
-                // Wait for the trx's to arrive.
-                TestHelper.WaitLoop(() => stratisReceiver.CreateRPCClient().GetRawMempool().Length > 0);
-                TestHelper.WaitLoop(() => stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Any());
-
-                long totalAmount = stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount);
-                Assert.Equal(Money.COIN * 50, totalAmount);
-
-                Assert.Null(stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
-
-                // Generate new blocks so the trx is confirmed.
-                stratisSender.GenerateStratisWithMiner(1);
-
-                // Wait for block repo for block sync to work.
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisReceiver, stratisSender));
-
-                // Confirm trx's have been committed to the block.
-                Assert.Equal(maturity + 52, stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
-
-                // Now send many inputs from receviever back to sender.         
-                var total = stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount);
-                Assert.Equal(5000000000, total);
-
-                // send coins to the receiver.
-                var sendto = stratisSender.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
-
-                transactionBuildContext = new TransactionBuildContext(
-                        new WalletAccountReference("mywallet", "account 0"),
-                        new[] {
-                            new Recipient {
-                                Amount = total - Money.COIN,
-                                ScriptPubKey = sendto.ScriptPubKey
-                            }
-                        }.ToList(), "123456");
-
-                // Check receiver has the correct inputs.
-                var trx = stratisReceiver.FullNode.WalletTransactionHandler().BuildTransaction(transactionBuildContext);
-                Assert.Equal(50, trx.Inputs.Count);
-
-                // Broadcast.
-                stratisReceiver.FullNode.NodeService<WalletController>().SendTransaction(new SendTransactionRequest(trx.ToHex()));
-
-                // Wait for the trx to arrive.
-                TestHelper.WaitLoop(() => stratisSender.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Any());
-
-                long receivetotal = stratisSender.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount);
-                Assert.Equal(747500000000, receivetotal);
-
-                // wait for block repo for block sync to work.
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisSender, stratisReceiver));
-
-                TestHelper.WaitLoop(() => 3 == stratisSender.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
-            }
-        }
+        }        
 
         [Fact]
         public void CanMineAndSendToAddress()
--- a/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification_Steps.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification_Steps.cs
@@ -1,0 +1,174 @@
+﻿using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
+using Stratis.Bitcoin.IntegrationTests.TestFramework;
+using Xunit.Abstractions;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    using Stratis.Bitcoin.Features.Consensus;
+    using Stratis.Bitcoin.Features.Wallet;
+    using NBitcoin;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Stratis.Bitcoin.Features.Wallet.Controllers;
+    using Stratis.Bitcoin.Features.Wallet.Models;
+    using System;
+    using FluentAssertions;
+
+    public partial class WalletTestsManyTxnInputsAndOutputsSpecification : BddSpecification
+    {
+        private NodeBuilder nodeBuilder;
+        private const string WalletName = "mywallet";
+        private const string WalletPassword = "123456";
+        private const string WalletAccountName = "account 0";
+        private CoreNode stratisSender;
+        private CoreNode stratisReceiver;
+        private Key keySender;
+        private TransactionBuildContext transactionBuildContext;
+        private int CoinBaseMaturity;
+
+        // Output helper allows test step names to be logged.
+        public WalletTestsManyTxnInputsAndOutputsSpecification(ITestOutputHelper outputHelper) : base(outputHelper) { }
+
+        protected override void BeforeTest()
+        {
+            this.nodeBuilder = NodeBuilder.Create();
+        }
+
+        protected override void AfterTest()
+        {
+            this.nodeBuilder.Dispose();
+        }
+
+        private void a_stratis_sender_and_receiver_node_and_their_wallets()
+        {
+            this.stratisSender = this.nodeBuilder.CreateStratisPowNode();
+            this.stratisReceiver = this.nodeBuilder.CreateStratisPowNode();
+
+            this.nodeBuilder.StartAll();
+            this.stratisSender.NotInIBD();
+            this.stratisReceiver.NotInIBD();
+
+            this.stratisSender.FullNode.WalletManager().CreateWallet(WalletPassword, WalletName);
+            this.stratisReceiver.FullNode.WalletManager().CreateWallet(WalletPassword, WalletName);
+
+            HdAddress addrSender = this.stratisSender.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference(WalletName, WalletAccountName));
+            Wallet walletSender = this.stratisSender.FullNode.WalletManager().GetWalletByName(WalletName);
+            this.keySender = walletSender.GetExtendedPrivateKeyForAddress(WalletPassword, addrSender).PrivateKey;
+        }
+
+        private void a_block_is_minded()
+        {
+            this.stratisSender.SetDummyMinerSecret(new BitcoinSecret(this.keySender, this.stratisSender.FullNode.Network));
+            this.CoinBaseMaturity = (int)this.stratisSender.FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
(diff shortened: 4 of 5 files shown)


src/Stratis.Bitcoin.IntegrationTests/Program.cs
```diff
@@ -12,6 +12,7 @@ internal class Program
         public static void Main(string[] args)
         {
             RunAllTestsOf<WalletTests>();
+            RunAllTestsOf<WalletTestsManyTxnInputsAndOutputsSpecification>();
```
majikandy: is this needed?

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        [Fact]
+        public void Wallet_can_send_and_receive_many_transaction_inputs_and_ouputs()
+        {
+            Given(a_stratis_sender_and_receiver_node_and_their_wallets);
+            And(a_block_is_mined);
+            When(a_transaction_is_sent_to_receiver_via_fifty_outputs);
+            And(the_transaction_is_recevied);
+            Then(the_recevier_sends_back_a_transaction_using_all_the_outputs);
```
majikandy: A Then shouldn't be a doing action, just an assert of some kind

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public partial class WalletTestsManyTxnInputsAndOutputsSpecification 
+    {
+        [Fact]
+        public void Wallet_can_send_one_transaction_with_many_outputs()
+        {
+            Given(a_stratis_sender_and_receiver_node_and_their_wallets);
+            And(a_block_is_mined);
+            When(a_transaction_is_sent_to_receiver_via_fifty_outputs);
+            Then(the_transaction_is_recevied);
```
majikandy: typo received

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        [Fact]
+        public void Wallet_can_send_and_receive_many_transaction_inputs_and_ouputs()
```
majikandy: In this situation where the first 4 lines of this test are the FULL previous test,  I think it makes more sense to just follow on the first test and do another When Then..

eg
Given(a_stratis_sender_and_receiver_node_and_their_wallets);
And(a_block_is_mined);
When(a_transaction_is_sent_to_receiver_via_fifty_outputs);
Then(the_transaction_is_received);
When(the_receiver_sends_back_from_those_fifty_outputs)
Then(the_transaction_is_received_by_sender);

Now it gets confusing - so perhaps node1 and node2 or nodeA and nodeB might be clearer

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
majikandy: If you want to keep as a separate test.... perhaps...

Given(a_stratis_node_with_unspent_outputs_across_fifty_addresses)
When(transaction_requiring_all_fifty_unspent_outputs_is_sent_to_one_address)
Then(the_transaction_is_received)

// the above given could call all the steps from the other test, or another method of setting it up

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            Given(a_stratis_sender_and_receiver_node_and_their_wallets);
+            And(a_block_is_mined);
+            When(a_transaction_is_sent_to_receiver_via_fifty_outputs);
+            And(the_transaction_is_recevied);
+            Then(the_recevier_sends_back_a_transaction_using_all_the_outputs);
+        }
+    }
+}
```
majikandy: how about adding another test to do multiple inputs to multiple outputs?

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
majikandy: Is "transaction received" the right domain language?  is it more like the "funds are received" or "balance is updated". I am thinking because the transaction is the thing on the block chain. @dangershony thoughts?

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTestsManyTxnInputsAndOutputsSpecification.cs
```diff
@@ -0,0 +1,24 @@
...
+            When(funds_are_sent_to_receiver_via_fifty_txn_outputs);
+            Then(the_funds_are_received);
+        }
+
+        [Fact]
+        public void Wallet_can_receive_funds_using_many_txn_inputs()
+        {
+            Given(sender_sends_funds_with_fifty_txn_outputs);
+            When(the_receiver_creates_a_txn_using_fifty_txn_inputs_and_sends_back);
+            Then(the_sender_receives_the_funds);
```
majikandy: sender receives still isn't logical

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            this.nodeOne = this.nodeBuilder.CreateStratisPowNode();
+            this.nodeTwo = this.nodeBuilder.CreateStratisPowNode();
+
+            this.nodeBuilder.StartAll();
+            this.nodeOne.NotInIBD();
+            this.nodeTwo.NotInIBD();
+
+            this.nodeOne.FullNode.WalletManager().CreateWallet(WalletPassword, WalletName);
+            this.nodeTwo.FullNode.WalletManager().CreateWallet(WalletPassword, WalletName);
```
majikandy: All of us appear to be repeating this bit of code everywhere, perhaps something like a NodeNetworkBuilder  could be useful (perhaps when writing next test between us)

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            HdAddress addrSender = this.nodeOne.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference(WalletName, WalletAccountName));
+            Wallet walletSender = this.nodeOne.FullNode.WalletManager().GetWalletByName(WalletName);
+            this.keySender = walletSender.GetExtendedPrivateKeyForAddress(WalletPassword, addrSender).PrivateKey;
+        }
+
+        private void node1_sends_funds_to_node2_TO_fifty_addresses()
+        {
+            // Mine block
```
majikandy: there is a method in SharedSteps you can use for this - new it up in the Before test as a field and then use it here

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,198 @@
...
+            this.nodeOne.SetDummyMinerSecret(new BitcoinSecret(this.keySender, this.nodeOne.FullNode.Network));
+            this.CoinBaseMaturity = (int)this.nodeOne.FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
+            this.nodeOne.GenerateStratisWithMiner(this.CoinBaseMaturity + 51);
+
+            // Wait for block repo for block sync to work.
+            TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(this.nodeOne));
```
majikandy: shared steps again - lose the comment when have that here

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,198 @@
...
+            this.nodeOne.FullNode.WalletManager().
+                GetSpendableTransactionsInWallet("mywallet").
+                Sum(s => s.Transaction.Amount).
+                Should().Be(Money.COIN * 150 * 50);
+
+            // Sync both nodes.
+            this.nodeOne.CreateRPCClient().AddNode(this.nodeTwo.Endpoint, true);
+            TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(this.nodeTwo, this.nodeOne));
+
+            // Get 50 unused addresses from the receiver.
```
majikandy: comment appears to be same as line of code

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,198 @@
...
+            List<Recipient> recipients = recevierAddresses.Select(address => new Recipient
+            {
+                ScriptPubKey = address.ScriptPubKey,
+                Amount = Money.COIN
+            }).ToList();
+
+            this.transactionBuildContext = new TransactionBuildContext(new WalletAccountReference(WalletName, WalletAccountName), recipients, WalletPassword)
+            {
+                FeeType = FeeType.Medium,
+                MinConfirmations = 101
```
majikandy: why 101 :)

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Done.

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+            else
+            {
+                reward = (consensusValidator.GetProofOfWorkReward(subsidyHalvingInterval - 1) * subsidyHalvingInterval)
+                            + (consensusValidator.GetProofOfWorkReward(blockCount) * (blockCount - (subsidyHalvingInterval + 1)));
+            }
+
+            return reward;
+        }
+    }
+}
```
majikandy: This implementation will only work to block 300 ?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private void two_connected_nodes()
+        {
+            this.nodes = this.nodeGroupBuilder
+                .StratisPowNode(NodeOne).Start().NotInIBD().WithWallet(WalletName, WalletPassword)
+                .StratisPowNode(NodeTwo).Start().NotInIBD().WithWallet(WalletName, WalletPassword)
+                .WithConnections()
+                    .Connect(NodeOne, NodeTwo)
+                    .AndNoMoreConnections()
+                .Build();
```
majikandy: nice :)

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        private void node1_sends_funds_to_node2_TO_fifty_addresses()
+        {
+            // Mine block
+            this.CoinBaseMaturity = (int)this.nodes[NodeOne].FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
+
+            this.sharedSteps.MineBlocks(this.CoinBaseMaturity + 50
```
majikandy: why +50 out of curiousity?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+                , WalletName
+                , WalletPassword);
+
+            // Sync both nodes.
+            this.nodes[NodeOne].CreateRPCClient().AddNode(this.nodes[NodeTwo].Endpoint, true);
+
+            this.sharedSteps.WaitForBlockStoreToSync(this.nodes[NodeTwo], this.nodes[NodeOne]);
+
+            // Get 50 unused addresses from node two.
+            IEnumerable<HdAddress> recevierAddresses = this.nodes[NodeTwo].FullNode.WalletManager()
```
majikandy: typo  recevierAddreses

perhaps you could lose the comment if you put   fiftyReceiveAddresses as the variableName?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+            // Mine block
+            this.CoinBaseMaturity = (int)this.nodes[NodeOne].FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
+
+            this.sharedSteps.MineBlocks(this.CoinBaseMaturity + 50
+                , this.nodes[NodeOne]
+                , WalletAccountName
+                , WalletName
+                , WalletPassword);
+
+            // Sync both nodes.
```
majikandy: feels like the 2 lines of code grouped without a blank line negates the need for this comment

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+                ScriptPubKey = address.ScriptPubKey,
+                Amount = Money.COIN
+            }).ToList();
+
+            this.transactionBuildContext = SharedSteps.CreateTransactionBuildContext(WalletName
+                , WalletAccountName
+                , WalletPassword
+                , recipients
+                , FeeType.Medium
+                , 101);
```
majikandy: I think I changed this to 1 in another test and it made no difference to the test

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            Transaction transaction = this.nodes[NodeOne].FullNode.WalletTransactionHandler().
+                BuildTransaction(this.transactionBuildContext);
+
+            transaction.Outputs.Count.Should().Be(51);
```
majikandy: why 51 and not 50?  change address?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+            this.stratisReceiverTotalAmount = this.nodes[NodeTwo].FullNode.WalletManager().
+                GetSpendableTransactionsInWallet(WalletName).
+                Sum(s => s.Transaction.Amount);
+
+            this.stratisReceiverTotalAmount.Should().Be(5000000000);
+
+            this.nodes[NodeTwo].FullNode.WalletManager().
+                GetSpendableTransactionsInWallet(WalletName).
+                First().Transaction.BlockHeight.
+                Should().BeNull();
```
majikandy: why should first transaction blockheight be null ?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+                SendTransaction(new SendTransactionRequest(transaction.ToHex()));
+
+            // Wait for the transactions to arrive.
+            TestHelper.WaitLoop(() => this.nodes[NodeTwo].CreateRPCClient().GetRawMempool().Length > 0);
+            TestHelper.WaitLoop(() => this.nodes[NodeTwo].FullNode.WalletManager().GetSpendableTransactionsInWallet(WalletName).Any());
+        }
+
+        private void node2_receives_the_funds()
+        {
+            this.stratisReceiverTotalAmount = this.nodes[NodeTwo].FullNode.WalletManager().
```
majikandy: you can inline this

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+                First().Transaction.BlockHeight.
+                Should().BeNull();
+
+            this.nodes[NodeTwo].FullNode.WalletManager().
+                GetSpendableTransactionsInWallet(WalletName).
+                Count().Should().Be(50);
+
+            this.nodes[NodeOne].GenerateStratisWithMiner(1);
+
+            this.sharedSteps.WaitForBlockStoreToSync(this.nodes[NodeOne]);
```
majikandy: won't the next line negate the need for this line?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+            Transaction transaction = this.nodes[NodeTwo].FullNode.WalletTransactionHandler()
+                        .BuildTransaction(this.transactionBuildContext);
+
+            transaction.Inputs.Count.Should().Be(50);
+
+            // Broadcast.
+            this.nodes[NodeTwo].FullNode.NodeService<WalletController>()
+                .SendTransaction(new SendTransactionRequest(transaction.ToHex()));
+
+            // Wait for the transaction to arrive.
```
majikandy: this should be in the next method I think

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+            TestHelper.WaitLoop(() => this.nodes[NodeOne].FullNode.WalletManager()
+                                        .GetSpendableTransactionsInWallet(WalletName).Any());
```
majikandy: Is this check correct?  Doesn't it have some spendable transactions already before it receives this lot?

majikandy: probably was zero I guess  - perhaps we check that first before making it, to make sure this check is definitely correct?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,189 @@
...
+            this.nodes[NodeOne].FullNode.WalletManager()
+                .GetSpendableTransactionsInWallet(WalletName)
+                .Sum(s => s.Transaction.Amount).Should().Be(745000000000);
+
+            this.nodes[NodeTwo].FullNode.WalletManager()
+                .GetSpendableTransactionsInWallet(WalletName)
+                .Count().Should().Be(1);
+
+            this.sharedSteps.WaitForBlockStoreToSync(this.nodes[NodeOne]);
+            this.sharedSteps.WaitForBlockStoreToSync(this.nodes[NodeTwo], this.nodes[NodeOne]);
```
majikandy: as above, not sure both of these lines needed

src/Stratis.BitcoinD/Stratis.BitcoinD.csproj
```diff
@@ -14,6 +14,8 @@
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <NoWarn>1701;1702;1705;IDE0008;</NoWarn>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
majikandy: I'm unsure about this - I remember your chat with @noescape00  about this - are we committing this one to master ?

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
ferdeen: Alter the method to cater for this and any block size.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Need a large amount to cover sending from 50 addresses.  I got to that figure using trial and error.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Lost the comment and changed to nodeTwoAddresses.  Reads better.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Dropped the comment.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: I couldn't drop it.  I'll check with Dan once his back - what minConfirmations is all about.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: This was the count after building the transaction.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Because nodetwo is not in sync with nodeone.  Before and after test.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Why ? It's used in node2_sends_funds_to_node1_FROM_fifty_addresses().

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Your'e correct!  I have removed it now.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Done.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: This was zero.  So the Any() check is good.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Done.

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using NBitcoin;
+using Stratis.Bitcoin.Features.Consensus;
+using Stratis.Bitcoin.Features.Consensus.Interfaces;
+using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
+
+namespace Stratis.Bitcoin.IntegrationTests.Utilities.Extensions
+{
+    public static class CoreNodeExtensions
+    {
+        public static Money CalculateProofOfWorkReward(this CoreNode node, int blockCount)
```
majikandy: This still won't work will it?  The reward is related to the blockheight - is that used in the implementation? I can't see it?

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var consensusValidator = node.FullNode.NodeService<IPowConsensusValidator>() as PowConsensusValidator;
+
+            var groupedRewardsBySubsidyHalving = Enumerable.Range(0, blockCount)
```
majikandy: here, you did 0-blockcount. This implies to me that this will only work when using it to validate a reward when you started mining at from 0

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            return rewardsPerGroup.Sum();
+        }
+
+        private static IEnumerable<IGrouping<int, TSource>> GroupBy<TSource> (this IEnumerable<TSource> source, int itemsPerGroup)
+        {
+            return source.Zip(Enumerable.Range(0, source.Count()),
+                              (s, r) => new { Group = r / itemsPerGroup, Item = s })
+                         .GroupBy(i => i.Group, g => g.Item)
+                         .ToList();
```
majikandy: struggling to follow this code

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,178 @@
...
+        private const string WalletAccountName = "account 0";
+        private IDictionary<string, CoreNode> nodes;
+        private NodeGroupBuilder nodeGroupBuilder;
+        private const string NodeOne = "one";
+        private const string NodeTwo = "two";
+        private long nodeTwoTotalAmount;
+        private TransactionBuildContext transactionBuildContext;
+        private int CoinBaseMaturity;
+
+        // Output helper allows test step names to be logged.
```
majikandy: this is forced by the base class now, so probably we can lose this comment

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
majikandy: You expected there to be 51 and not 50?  why?  from what I see you pass 50 recipient addresses in

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,178 @@
...
+                BuildTransaction(this.transactionBuildContext);
+
+            transaction.Outputs.Count.Should().Be(51);
+
+            this.nodes[NodeOne].FullNode.NodeService<WalletController>().
+                SendTransaction(new SendTransactionRequest(transaction.ToHex()));
+
+            TestHelper.WaitLoop(() => this.nodes[NodeTwo].CreateRPCClient().GetRawMempool().Length > 0);
+            TestHelper.WaitLoop(() => this.nodes[NodeTwo].FullNode.WalletManager().GetSpendableTransactionsInWallet(WalletName).Any());
+        }
```
majikandy: so you are checking it got to the mempool of the second node? that feels like it should be part of the Then and not the When?

majikandy: TBH, you could argue part of the Send step, but perhaps more so if you were checking the mem pool on the sending node instead of receiving here.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,178 @@
...
+                }.ToList(), WalletPassword);
+
+            Transaction transaction = this.nodes[NodeTwo].FullNode.WalletTransactionHandler()
+                        .BuildTransaction(this.transactionBuildContext);
+
+            transaction.Inputs.Count.Should().Be(50);
+
+            this.nodes[NodeTwo].FullNode.NodeService<WalletController>()
+                .SendTransaction(new SendTransactionRequest(transaction.ToHex()));
+        }
```
majikandy: you didn't do the mempool check here like you did on the other send.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: This is 50 for nodeTwo addresses and the extra output is change - confirmed this with Dan.

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
ferdeen: This is just building up a sequence of numbers to work out each reward per group - I tested if zero blocks are mined and there is no side effect on this function.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
ferdeen: Dropped the mempool check as it's not adding anything to the test.

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSteps.cs
```diff
@@ -82,7 +82,11 @@ private void spending_the_coins_from_original_block()
             try
             {
                 this.lastTransaction = this.sendingStratisBitcoinNode.FullNode.WalletTransactionHandler()
-                    .BuildTransaction(SharedSteps.CreateTransactionBuildContext(SendingWalletName, AccountName, WalletPassword, sendtoAddress.ScriptPubKey, Money.COIN * 1, FeeType.Medium, 101));
+                    .BuildTransaction(SharedSteps.CreateTransactionBuildContext(SendingWalletName
+                        , AccountName
```
majikandy: I find these `,` at the start of the new line a bit strange - probably doesn't break the policy but just looks slightly odd to me - like when you do sql statements so you can comment lines out etc, except you can't/wouldn't here - wouldn't end of prev line be more in line with most of the rest of codebase?

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ReorgToLongestChainSteps.cs
```diff
@@ -83,7 +83,11 @@ private void bob_creates_a_transaction_and_broadcasts()
         {
             var nodeCReceivingAddress = this.GetSecondUnusedAddressToAvoidClashWithMiningAddress(this.nodes[Charlie]);
 
-            var transactionBuildContext = SharedSteps.CreateTransactionBuildContext(WalletZero, AccountZero, WalletPassword, nodeCReceivingAddress.ScriptPubKey, Money.COIN * 1, FeeType.Medium, minConfirmations: 1);
+            var transactionBuildContext = SharedSteps.CreateTransactionBuildContext(WalletZero
```
majikandy: I would also suggest that if it a new line per parameter, perhaps the first one should be too? Just an idea

src/Stratis.Bitcoin.IntegrationTests/Utilities/Extensions/CoreNodeExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+                .Partition(consensusValidator.ConsensusParams.SubsidyHalvingInterval);
+
+            var rewardsPerGroup = new List<Money>();
+
+            foreach(var groupedReward in groupedRewards)
+            { 
+                rewardsPerGroup.Add(groupedReward.Count() * consensusValidator.GetProofOfWorkReward(groupedReward.First()));
+            }
+
+            return rewardsPerGroup.Sum();
```
majikandy: could lines 22 -> 29 have been something like `groupedRewards.Sum(g => g.Count() * onsensusValidator.GetProofOfWorkReward(x));`   ?

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ReorgToLongestChainSteps.cs
```diff
@@ -83,7 +83,17 @@ private void bob_creates_a_transaction_and_broadcasts()
         {
             var nodeCReceivingAddress = this.GetSecondUnusedAddressToAvoidClashWithMiningAddress(this.nodes[Charlie]);
 
-            var transactionBuildContext = SharedSteps.CreateTransactionBuildContext(WalletZero, AccountZero, WalletPassword, nodeCReceivingAddress.ScriptPubKey, Money.COIN * 1, FeeType.Medium, minConfirmations: 1);
+            var transactionBuildContext = SharedSteps.CreateTransactionBuildContext(
+                WalletZero,
+                AccountZero,
+                WalletPassword
+                , new[] {
```
majikandy: shocking

ferdeen: Ha! Put the comma in the correct place.

src/Stratis.BitcoinD/Stratis.BitcoinD.csproj
```diff
@@ -14,6 +14,8 @@
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <NoWarn>1701;1702;1705;IDE0008;</NoWarn>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsAsErrors>NU1605</WarningsAsErrors>
```
majikandy: I thought you were gonna completely revert this file ?

majikandy: but you haven't reverted it

ferdeen: Just done a revert git command line and it's done the same as my change - so not too sure what gains we made here.

   >git checkout 693743e~1 src\Stratis.BitcoinD\Stratis.BitcoinD.csproj

image

src/Stratis.BitcoinD/Stratis.BitcoinD.csproj
majikandy: you still need to revert this file right ?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -0,0 +1,151 @@
...
+                    {
+                        Amount = this.nodeTwoBalance - Money.COIN,
+                        ScriptPubKey = sendToNodeOne.ScriptPubKey
+                    }
+                }.ToList(), 
+                WalletPassword);
+
+            var transaction = this.nodes[NodeTwo].FullNode.WalletTransactionHandler().BuildTransaction(this.transactionBuildContext);
+
+            this.transactionFee = this.transactionFee = this.nodes[NodeTwo].GetFee(this.transactionBuildContext);
```
fassadlr: Extra ` this.transactionFee`

ferdeen: Cool, just removed it.

PR #1261, opened 2018-03-28
Switch from the chain if it's no longer represented by any peer that we're connected to

noescape00:
This PR is the first one for #229

It introduces a new component `BestChainSelector` that handles situations when a peer disconnects and selects the best chain tip from the available ones in case disconnected peer was the only one who backed our current best chain. 

ps
Reviewed already by @Aprogiena


dangershony:
Ok this looks good to me lets push it.
@noescape00 one thing I would add is to create an integration test that will check this behaviour with 3 nodes.

Not urgent

PR #1264, opened 2018-03-30
Wallet Thread and error message fixes

bokobza:
Fixes a thread safety issue when building a transaction. 
Also fixes https://github.com/stratisproject/FullNodeUI/issues/39

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
@@ -115,7 +115,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Interfaces
         /// <param name="accountReference">The name of the wallet and account.</param>
         /// <param name="count">The number of addresses to create.</param>
         /// <param name="isChange">A value indicating whether or not the addresses to get should be receiving or change addresses.</param>
-        /// <returns>A list of addresses.</returns>
+        /// <returns>A list of unused addresses. New addresses will be created as necessary.</returns>
         IEnumerable<HdAddress> GetUnusedAddresses(WalletAccountReference accountReference, int count, bool isChange = false);
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -421,20 +433,34 @@ public IEnumerable<HdAddress> GetUnusedAddresses(WalletAccountReference accountR
...
                     generated = true;
                 }
 
-                addresses = account
-                    .ExternalAddresses
-                    .Where(acc => !acc.Transactions.Any())
-                    .OrderBy(x => x.Index)
-                    .Take(count);
+                if (isChange)
+                {
```
dangershony: you already did this enumeration at line 436 why do it again?

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
```diff
@@ -102,7 +102,21 @@ public interface IWalletManager
         /// <returns>An unused address or a newly created address, in Base58 format.</returns>
         HdAddress GetUnusedAddress(WalletAccountReference accountReference);
 
-        IEnumerable<HdAddress> GetUnusedAddresses(WalletAccountReference accountReference, int count);
+        /// <summary>
+        /// Gets the first change address that contains no transaction.
```
dangershony: If no addresses are found this will also create new derived addresses in the wallet.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -446,35 +472,7 @@ public IEnumerable<HdAddress> GetUnusedAddresses(WalletAccountReference accountR
...
-        /// <inheritdoc />
-        public HdAddress GetOrCreateChangeAddress(HdAccount account)
-        {
-            this.logger.LogTrace("()");
-            HdAddress changeAddress = null;
-
-            lock (this.lockObject)
-            {
-                // Get an address to send the change to.
-                changeAddress = account.GetFirstUnusedChangeAddress();
```
dangershony: Maybe this method can be deletes? `GetFirstUnusedChangeAddress`

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: I guess yeah. I didn't do it because if say you have 7 unused addresses and you want to receive 3, unusedAddresses will return all 7 addresses.
But you can return only the first 3 (.Take(3)). I'll change it.

PR #1266, opened 2018-04-01
Tests for PosMinting

Neurosploit:
Tests GetDifficulty and GetNetworkWeight.
Resolved some typos and attempted to add some more documentation on getdifficulty.
Removed GetMatureBalance and used GetUtxoStakeDescriptionsSuitableForStaking.

The main method is just too big of a chunk to setup. That would need some additional refactoring to become testable. Not sure how yet. Testing the private methods in the class seperately would help quite a lot then we don't need a monolith of a test. to test out each scenario.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
+++ b/src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Moq;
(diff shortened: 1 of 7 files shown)


Neurosploit:
Furthermore I wonder if https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs#L1083 should be a zero or a one.

Neurosploit:
Hmm looks like this needs some stability.. i'll have a look where this is coming from.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -704,12 +704,12 @@ private void CheckStake(Block block, ChainedBlock chainTip)
             // Mark coinstake transaction.
             coinstakeContext.CoinstakeTx.Outputs.Add(new TxOut(Money.Zero, new Script()));
 
-            long balance = this.GetMatureBalance(utxoStakeDescriptions).Satoshi;
+            long balance = this.GetUtxoStakeDescriptionsSuitableForStaking(utxoStakeDescriptions, coinstakeContext.CoinstakeTx.Time, long.MaxValue).Sum(b => b.TxOut.Value);
```
noescape00: Before this change we used to get mature balance (mature balance is the balance we can spend, but some of utxos that are being used in it's calculation may be not suitable for staking because they dont meet the coinstake age maturity). 
This balance was needed in order to calculate target value for the `GetUtxoStakeDescriptionsSuitableForStaking`. 

With the new implementation you are likely to reserve more than is required. 
For example: we have 10 coins and 3 of them are confirmed just in a few blocks. In the current code those 3 coins are being cut off by the first `GetUtxoStakeDescriptionsSuitableForStaking` and after that you cut off some more coins for the reserve balance. But what we wanted to do is to just make sure that we have enough coins suitable for spending with the sum of reserve balance. 

So the logic effectively became incorrect after that optimization.

src/Stratis.Bitcoin.Features.Miner/Interfaces/IPosMinting.cs
```diff
@@ -36,25 +36,24 @@ public interface IPosMinting
...
+        /// It has a fixed 24-block difficulty adjustment window for which it determines the average difficulty during the block window 
+        /// using a linear adjustment (directly scaling by the ratio of actual and target time).
+        /// The algorithm then uses the block header bits to calculate a maximum shift that allows the difficulty to shift either up or down. 
+        /// It can shift in either direction to the same fixed factor.
+        /// The algorithm was ported from <see cref="https://github.com/stratisproject/stratisX/blob/47851b7337f528f52ec20e86dca7dcead8191cf5/src/rpcblockchain.cpp#L16"/>
+        /// which in turn was ported from darkcoin and/or bitcoin <see cref="https://github.com/bitcoin/bitcoin/blob/0.15/src/rpc/blockchain.cpp#L49"/>.
+        /// Additional information on general difficulty calculations: <see cref="https://en.bitcoin.it/wiki/Difficulty"/> and 
+        /// for some description regarding possible adjustments to make <see cref="http://web.archive.org/web/20180209161330/https://github.com/zcash/zcash/issues/147#issuecomment-202829008"/>.
```
noescape00: https://bitcoin.stackexchange.com/questions/30467/what-are-the-equations-to-convert-between-bits-and-difficulty

The comment is incorrect. It doesn't calculate the difficulty, it's just encoding 32 bits difficulty (`block.Header.Bits`) into 4 bits representation with some data loss. 

The difficulty retargeting algorithm is in the `StakeValidator.cs:76`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -1188,7 +1175,7 @@ public double GetNetworkWeight()
         }
 
         /// <inheritdoc/>
-        public Models.GetStakingInfoModel GetGetStakingInfoModel()
+        public Models.GetStakingInfoModel GetStakingInfoModel()
```
noescape00: Before it was called `GetGetStakingInfoModel` for the consistency. So the pattern for rpc is : `get` + `command name` + `model`. The command here is `getstakinginfo` so that's where this double 'get' came from.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -0,0 +1,476 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
noescape00: redundant namespace

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -0,0 +1,476 @@
...
+using Stratis.Bitcoin.Tests.Logging;
+using Stratis.Bitcoin.Utilities;
+using Xunit;
+
+namespace Stratis.Bitcoin.Features.Miner.Tests
+{
+    public class PosMintingTest : LogsTestBase, IDisposable
+    {
+        private PosMinting posMinting;
+        private bool initialBlockSignature;
```
noescape00: almost all fields can be made readonly

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -0,0 +1,476 @@
...
+            this.timeSyncBehaviorState = new Mock<ITimeSyncBehaviorState>();
+
+            this.consensusLoop.Setup(c => c.Validator)
+                .Returns(this.consensusValidator.Object);
+
+            this.cancellationTokenSource = new CancellationTokenSource();
+            this.nodeLifetime.Setup(n => n.ApplicationStopping)
+                .Returns(this.cancellationTokenSource.Token);
+
+            this.posMinting = InitializePosMinting();
```
noescape00: `this.InitializePosMinting();`

bokobza:
@noescape00 @Neurosploit I'm rebuilding this commit and I'll check in if successful.

bokobza:
Not passing tests because it's not merged. 
There are no changes to the core code so I'll merge it anyway.
Thanks @Neurosploit !

PR #1268, opened 2018-04-02
Draw more attention to the existence of the getting started guide

StratisDeveloper:


StratisDeveloper:
A developer who wants to use the SFN will look at the readme to understand how to use it. 

"more info here" is not sufficient for someone to find the getting started guide. So lets improve that with a header.

dangershony:
Yeah sounds good

mwherman2000:
What is the link to the Getting Started Guide? ...seriously

PR #1276, opened 2018-04-06
Payload command length check

noescape00:
This pr adds a check for the `commandName` of the payload. 

I believe this is needed because if a new payload is created (let's say by someone who builds on top of the node and implements a new protocol message) and it's command name doesn't meet the limitations set by the protocol the communication with this payload won't be possible, however no error will be spawned while trying to use a payload with long command name and the node will run normally.


noescape00:
>can you not use the Guard NotEmpty method

No, ` Guard.NotEmpty` checks for null or "". Here we need to make sure that it's shorter than 12 characters.

PR #1279, opened 2018-04-07
Flag to make sure GetAddrPayload is only sent once.

### LINKED ISSUE #637: Implement getaddr similar to stratisX (and core)
This is how its done in core
```
    else if (strCommand == NetMsgType::GETADDR)
    {
        // This asymmetric behavior for inbound and outbound connections was introduced
        // to prevent a fingerprinting attack: an attacker can send specific fake addresses
        // to users' AddrMan and later request them by sending getaddr messages.
        // Making nodes which are behind NAT and can only make outgoing connections ignore
        // the getaddr message mitigates the attack.
        if (!pfrom->fInbound) {
            LogPrint(BCLog::NET, "Ignoring \"getaddr\" from outbound connection. peer=%d\n", pfrom->GetId());
            return true;
        }

        // Only send one GetAddr response per connection to reduce resource waste
        //  and discourage addr stamping of INV announcements.
        if (pfrom->fSentAddr) {
            LogPrint(BCLog::NET, "Ignoring repeated \"getaddr\". peer=%d\n", pfrom->GetId());
            return true;
        }
        pfrom->fSentAddr = true;

        pfrom->vAddrToSend.clear();
        std::vector<CAddress> vAddr = connman.GetAddresses();
        FastRandomContext insecure_rand;
        for (const CAddress &addr : vAddr)
            pfrom->PushAddress(addr, insecure_rand);
    }
```

ferdeen:
Fix #637.

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
@@ -38,7 +38,7 @@ namespace Stratis.Bitcoin.P2P
         public int PeersToDiscover { get; set; }
 
         /// <summary>
-        /// Flat to make sure <see cref="GetAddrPayload"/> is only sent once.
+        /// Flag to make sure <see cref="GetAddrPayload"/> is only sent once.
         /// </summary>
         private bool SentAddress;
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
```diff
@@ -37,6 +37,11 @@ public sealed class PeerAddressManagerBehaviour : NetworkPeerBehavior
         /// </summary>
         public int PeersToDiscover { get; set; }
 
+        /// <summary>
+        /// Flag to make sure <see cref="GetAddrPayload"/> is only sent once.
+        /// </summary>
+        private bool SentAddress;
```
noescape00: `SentAddress` => `sentAddress`

PR #1280, opened 2018-04-08
Add ability to specify agent version

sondreb:
Add the ability to specify both user agent and user agent version, allowing consumers of the Stratis.Bitcoin library to specify version based on external criteria and not strongly bounded to the assembly version of Stratis.Bitcoin.

Since the agent can already be specified, it would make sense to be allowed to set the version as well, which might come from the executing assembly version.

### DIFF
--- a/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
+++ b/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
@@ -41,10 +41,12 @@ namespace Stratis.Bitcoin.Configuration
         /// <param name="innerNetwork">Specification of the network the node runs on - regtest/testnet/mainnet.</param>
         /// <param name="protocolVersion">Supported protocol version for which to create the configuration.</param>
         /// <param name="agent">The nodes user agent that will be shared with peers.</param>
+        /// <param name="agentVersion">The nodes user agent version that will be shared with peers.</param>
         public NodeSettings(Network innerNetwork = null, ProtocolVersion protocolVersion = SupportedProtocolVersion, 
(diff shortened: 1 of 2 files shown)


dangershony:
Interesting, so in case you made changes to other features and you wanted to reflect that in the version this will be useful.
The problem I see here is the current version represents the underline protocol which you use an nuget package to reference and you do indeed want to keep  the version of protocol you are executing, if you fork the code just change the agent yourself.

One could always use the agent text available to specify higher level versioning 

so you would end up with 
`my-agent:high-level-version:p2p-version`
my-agent:high-level-version can be injected by the user.

You could easily  right now do this
`explorer-0.1-:1.1.0`

Is this not enough?

sondreb:
Someone that uses the NuGet could use an agent prefix, but I would expect most to want to be able to specify a single version to differentiate between what versions their nodes are running. Let me know if you don't want this, and I'll cancel it :-)

bokobza:
At the moment we have an `agentPrefix`, as well as an `agent`.
The `agentPrefix` can be set in the command line and is prepended to the `agent` (see https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin/Configuration/NodeSettings.cs#L250).
A version can already either:
* be added to the `agentPrefix` if the node is started via the command line, 
* be included in the `agent` if the node is started programmatically.

I think there are already enough way to include a `version`, without us needing to add an `agentVersion`.

dangershony:
@sondreb I think we have enough options so yes please close this.
If you want some tasks to work on we have some very interesting bounties you could select find me on discord and we can take it form there.

bokobza:
Closing this.

PR #1281, opened 2018-04-08
Fixed unconfirmed trx bug

bokobza:
There was a bug whereby transactions sent to an address would not get confirmed.
The outgoing transaction would get registered in the wallet file, but not updated once it was confirmed in a block.

The explanation is as follows: 
The wallet manager maintains in memory a dictionary of unspent transactions, which makes checking whether incoming transactions affects the wallet or not, much easier and much faster.
The problem was that this list was marking an spent transactions as spent, even before it was confirmed. So when the transaction would show up in a block as confirmed, the manager would not even process it, and the "unconfirmed" status of the transaction would never move to "confirmed".

The resolution is that this dictionary is now keeping a list of all the transactions, unspent or not. Lookup is still very fast, and since transaction processing is idempotent, the transactions are processed properly.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
@@ -433,10 +433,10 @@ namespace Stratis.Bitcoin.Features.Wallet
                 // Get the account.
                 HdAccount account = wallet.GetAccountByCoinType(accountReference.AccountName, this.coinType);
 
-                List<HdAddress> unusedAddresses = isChange ? 
-                    account.InternalAddresses.Where(acc => !acc.Transactions.Any()).ToList() : 
+                List<HdAddress> unusedAddresses = isChange ?
+                    account.InternalAddresses.Where(acc => !acc.Transactions.Any()).ToList() :
                     account.ExternalAddresses.Where(acc => !acc.Transactions.Any()).ToList();
-                
+
                 int diff = unusedAddresses.Count - count;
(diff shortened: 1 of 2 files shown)


dangershony:
So the problem was on startup?

dev0tion:
This happened every time a transaction was sent Dan

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1010,7 +1010,6 @@ private void AddTransactionToWallet(Transaction transaction, TxOut utxo, int? bl
 
                 spentTransaction.SpendingDetails = spendingDetails;
                 spentTransaction.MerkleProof = null;
-                this.RemoveInputKeysLookupLock(spentTransaction);
```
dangershony: FYI just noticed the params in the signature `uint256 spendingTransactionId, int? spendingTransactionIndex` they are taken form a TransactionDetails class that is already found at the caller
`  if (!this.outpointLookup.TryGetValue(input.PrevOut, out TransactionData tTx))`
You could just pass that class inside to the method instead of looking it up again.

dangershony:
@bokobza please remember to merge to master

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: indeed, nice one. Next PR. ;-)

PR #1290, opened 2018-04-10
[Tests] POW / POS Integation Test

fassadlr:
As part of this test I refactored the Api tests so that each test creates a node. This way it'll be easier to "convert" them to BDD.

Also, @Aprogiena  and @dangershony  please see the changes in the POW block assembler with regards to setting the coinbase and block header time stamps.

For the test I implemented a `FastCoinsDateTimeProvider` which starts at some point in the past and generate's blocks "forward" in time from there. Let me know you thoughts.

(This PR replaces #1245)


Aprogiena:
Yes I think coinstake time is now created correctly

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            this.posReceiverAddress = this.nodes[PosStaker].FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference(PosWallet, PosWalletAccount));
+            var wallet = this.nodes[PosStaker].FullNode.WalletManager().GetWalletByName(PosWallet);
+            this.posReceiverPrivateKey = wallet.GetExtendedPrivateKeyForAddress(PosWalletPassword, this.posReceiverAddress).PrivateKey;
+
+            this.nodes[PosStaker].SetDummyMinerSecret(new BitcoinSecret(this.posReceiverPrivateKey, this.nodes[PosStaker].FullNode.Network));
+            this.nodeGroupBuilder.WithConnections().Connect(PowMiner, PosStaker);
+        }
+
+        private void create_tx_to_send_million_coins_from_pow_wallet_to_pos_node_wallet()
```
majikandy: Since this step actually sends it too, perhaps a rename to    `sends_a_million_coins_from_pow_wallet_to_pos_wallet`

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+                FeeType.Medium,
+                101);
+
+            context.OverrideFeeRate = new FeeRate(Money.Satoshis(20000));
+
+            var sendTransaction = CreateSendTransaction();
+            this.nodes[PowMiner].FullNode.NodeService<WalletController>().SendTransaction(new SendTransactionRequest(sendTransaction.ToHex()));
+        }
+
+        private Transaction CreateSendTransaction()
```
majikandy: do you like this private method close to its usage like this, or perhaps pull to the bottom of the file?

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+            transaction.AddOutput(new TxOut(new Money(100000000000000), this.posReceiverAddress.ScriptPubKey));
+            transaction.Sign(this.powSenderPrivateKey, new[] { coin });
+
+            return transaction;
+        }
+
+        private void pow_wallet_broadcasts_tx_of_million_coins_and_pos_wallet_receives()
+        {
+            // Wait for the coins to arrive
+            TestHelper.WaitLoop(() => this.nodes[PosStaker].CreateRPCClient().GetRawMempool().Length > 0);
```
majikandy: I think we have something like this in the shared steps - if we don't perhaps we should and it can be properly named and lose the comment

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+            TestHelper.WaitLoop(() => this.nodes[PosStaker].FullNode.WalletManager().GetSpendableTransactionsInWallet(PosWallet).Any());
```
majikandy: @ferdeen had some extension methods in his PR which might clean these lines up a lot, but I don't think that is quite merged in yet

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+            TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(this.nodes[PosStaker]));
+
+            // Ensure that the wallet on the proof-of-stake node reflects the coins sent (1 000 000)
```
majikandy: the name of the step basically covers this assert line in terms of needing this comment or not? - as long as the step is overall short enough, I feel you wouldn't need the comment

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,177 @@
...
+            var received = this.nodes[PosStaker].FullNode.WalletManager().GetSpendableTransactionsInWallet(PosWallet);
+            received.Sum(s => s.Transaction.Amount).Should().Be(Money.COIN * 1000000);
+        }
+
+        private void pos_node_mines_a_further_ten_blocks()
```
majikandy: if that 10 represents maturity of blocks earning though staking, it feels like the name here could reflect that and negate the need for the extensive comment block?

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSpecification.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Stratis.Bitcoin.IntegrationTests.Miners
+{
+    public partial class ProofOfStakeMintCoinsSpecification : BddSpecification
+    {
+        [Fact]
+        public void Staking_wallet_can_mint_new_coins()
+        {
+            Given(a_proof_of_work_node_with_wallet);
+            And(it_mines_genesis_and_premine_blocks);
```
majikandy: These two steps together feel like they could be their own GWT (Given When Then) in another scenario with a `Then 98 million coins are mined`.  The rest of this test could start from that point as a given.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSpecification.cs
```diff
@@ -0,0 +1,23 @@
...
+            And(mine_coins_to_maturity);
+            And(a_proof_of_stake_node_with_wallet);
+            And(it_syncs_with_proof_work_node);
+            And(create_tx_to_send_million_coins_from_pow_wallet_to_pos_node_wallet);
+            And(pow_wallet_broadcasts_tx_of_million_coins_and_pos_wallet_receives);
+            And(pos_node_mines_a_further_ten_blocks);
+            When(pos_node_starts_staking);
+            Then(pos_node_wallet_has_earned_coins_through_staking);
+        }
```
majikandy: Overall, I like the detail - perhaps some tense standardisation and perhaps this is overgranular - but at this stage with us all new to this expressive style, being overgranular is probably not a bad thing at this stage.

As a fairly new developer to the codebase, seeing  this high level overview of the test at this level immediately gives me some understanding of what is going on.  And to be able to drill in and see the details of how these things are achieved is really nice when it is compartmentalised like this.  
@ferdeen @monsieurleberre @maciejzaleski , as new developers, do you find this easy to read and get the higher level bigger picture of what this test (and that part of the system) is about?

In many ways, I don't need to drill in, (once we know the inside is accurate) - reading the integration test steps alone starts to give me the bigger picture of the system.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/FastCoinsDateTimeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+            return this.startFrom.ToUnixTimestamp();
+        }
+
+        public DateTime GetUtcNow()
+        {
+            return DateTime.UtcNow;
+        }
+
+        /// <summary>
+        /// This gets called when the Transaction's time gets set.
```
majikandy: I do wonder about summaries like this where it says where something gets called from - on the face of it, it seems useful to know that information, but, without going to the other code, we wouldn't know if it really does or doesn't get called from there.  Therefore a summary like this creates potential work for the reader/maintainer and can easily be inaccurate.   (For reference, In [Clean Code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882), Uncle Bob describes something similar to this in the "Bad comments" section of the book, regarding a properties file getting default values when file not found, yet nothing in the code right there showing if true or not.  Just like this part of the code doesn't show where it is called from, without hunting that down, we cannot know if this is true or not. Furthermore, if that changes later to not call this, would we remember to come and update this comment. It sounds like a huge overhead in code maintenance for such a simple thing, but my gut feel is that saying where something is called from, should solely be in the code alone and not the comments.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: Renamed 👍

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: Actually I moved this into the step 👍

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: OK yes, I used the shared step to test if the node synced.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: I've removed the comment 👍

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/FastCoinsDateTimeProvider.cs
fassadlr: @majikandy OK yeah I agree with you, I actually put this there for my own reference as this DTP substitutes the actual DTP of the node for this integration test. I needed to keep track of what DTP method was called when, at block creation. I will update the comment to be more specific as to the function of this class.

majikandy: Ah, I missed that this was in tests, I saw a summary and assumed it was prod code

src/Stratis.Bitcoin.IntegrationTests/Builders/NodeGroupBuilder.cs
```diff
@@ -33,6 +32,24 @@ public NodeGroupBuilder StratisPowNode(string nodeName)
             return this;
         }
 
+        public NodeGroupBuilder CreateStratisPowMiningNode(string nodeName)
```
majikandy: Shall we update the bitcoin pow method on line 29 from `StratisPowNode(` to `CreateStratisBitcoinPowNode(` or maybe even `CreateBitcoinPowNode(` ?

fassadlr: We will get to renaming these runners properly soon!

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
majikandy: reads well

src/Stratis.Bitcoin.IntegrationTests/Builders/NodeGroupBuilder.cs
majikandy: ok

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -192,12 +267,99 @@ public static FullNode BuildFullNode(NodeSettings args, Action<IFullNodeBuilder>
...
+                    builder.SubstituteDateTimeProviderFor<MiningFeature>(new GenerateCoinsFastDateTimeProvider());
+
+                node = (FullNode)builder.Build();
+            }
+
+            return node;
+        }
     }
 
     public static class FullNodeTestBuilderExtension
```
ferdeen: Hi, maybe move the extension class into it's own folder.  My [PR](https://github.com/stratisproject/StratisBitcoinFullNode/pull/1259) (which isn't merged to master yet) has an extension class - `CoreNodeExtension`.  

`CoreNodeExtension` lives in a \Utilities\Extensions folder.

image

ferdeen: If there are no other types of code in the \Utilities folder then maybe just an \Extensions folder.  I added Utilities as I copied an example from another project in the solution.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -89,40 +66,35 @@ public void CanGetGeneralInfoViaAPI()
         [Fact]
         public void CanStartStakingViaAPI()
         {
-            try
+            using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var fullNode = this.apiTestsFixture.stratisStakeNode.FullNode;
-                var apiURI = fullNode.NodeService<ApiSettings>().ApiUri;
+                var posNode = builder.CreateStratisPosApiNode();
+                posNode.Start();
```
ferdeen: Maybe reduce a line - pass in `start `param as true ? `var posNode = builder.CreateStratisPosApiNode(true);`

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -89,40 +66,35 @@ public void CanGetGeneralInfoViaAPI()
...
 
-                using (client = new HttpClient())
+                using (var httpClient = new HttpClient())
                 {
-                    WalletManager walletManager = fullNode.NodeService<IWalletManager>() as WalletManager;
+                    var walletManager = posNode.FullNode.NodeService<IWalletManager>() as WalletManager;
 
                     // create the wallet
                     var model = new StartStakingRequest { Name = "apitest", Password = "123456" };
                     var mnemonic = walletManager.CreateWallet(model.Password, model.Name);
```
ferdeen: Comment and mnemonic var not required.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -89,40 +66,35 @@ public void CanGetGeneralInfoViaAPI()
...
-            try
+            using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var fullNode = this.apiTestsFixture.stratisStakeNode.FullNode;
-                var apiURI = fullNode.NodeService<ApiSettings>().ApiUri;
+                var posNode = builder.CreateStratisPosApiNode();
+                posNode.Start();
 
-                Assert.NotNull(fullNode.NodeService<IPosMinting>(true));
+                var apiURI = posNode.FullNode.NodeService<ApiSettings>().ApiUri;
```
ferdeen: Can move this line close to where apiURI is being used.  Something like 

image

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -89,40 +66,35 @@ public void CanGetGeneralInfoViaAPI()
...
-                    var response = client.PostAsync(apiURI + "api/miner/startstaking", content).GetAwaiter().GetResult();
+                    var response = httpClient.PostAsync(apiURI + "api/miner/startstaking", content).GetAwaiter().GetResult();
                     Assert.Equal(System.Net.HttpStatusCode.OK, response.StatusCode);
 
                     var responseText = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
                     Assert.Equal("", responseText);
 
-                    MiningRPCController controller = fullNode.NodeService<MiningRPCController>();
+                    MiningRPCController controller = posNode.FullNode.NodeService<MiningRPCController>();
                     GetStakingInfoModel info = controller.GetStakingInfo();
```
ferdeen: Lines above 91 you declare variables with the var key word.  Here you use typed variables - do you think it might look better to keep to be consistent throughout the method ? 

Even reduce down the lines by getting rid of `controller` variable 

image

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -132,26 +104,22 @@ public void CanStartStakingViaAPI()
         [Fact]
         public void CanCallRPCMethodViaRPCsCallByNameAPI()
         {
-            try
+            using (NodeBuilder builder = NodeBuilder.Create())
             {
-                var fullNode = this.apiTestsFixture.stratisPowNode.FullNode;
-                var apiURI = fullNode.NodeService<ApiSettings>().ApiUri;
+                var posNode = builder.CreateStratisPosApiNode();
+                posNode.Start();
```
ferdeen: Maybe reduce a line - pass in startparam as true ? `var posNode = builder.CreateStratisPosApiNode(true)`

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -132,26 +104,22 @@ public void CanStartStakingViaAPI()
...
+                var apiURI = posNode.FullNode.NodeService<ApiSettings>().ApiUri;
 
-                    var response = client.GetStringAsync(apiURI + "api/rpc/callbyname?methodName=getblockhash&height=0").GetAwaiter().GetResult();
+                using (var httpClient = new HttpClient())
+                {
+                    httpClient.DefaultRequestHeaders.Accept.Clear();
+                    httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
 
-                    Assert.Equal("\"0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206\"", response);
+                    var response = httpClient.GetStringAsync(apiURI + "api/rpc/callbyname?methodName=getblockhash&height=0").GetAwaiter().GetResult();
```
ferdeen: Bring the `apiURI` variable closer to where it's used 

image

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -160,99 +128,22 @@ public void CanCallRPCMethodViaRPCsCallByNameAPI()
...
-                var fullNode = this.apiTestsFixture.stratisPowNode.FullNode;
-                var apiURI = fullNode.NodeService<ApiSettings>().ApiUri;
+                var posNode = builder.CreateStratisPosApiNode();
+                posNode.Start();
 
-                using (client = new HttpClient())
-                {
-                    client.DefaultRequestHeaders.Accept.Clear();
-                    client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+                var apiURI = posNode.FullNode.NodeService<ApiSettings>().ApiUri;
```
ferdeen: same as above, bring closer.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -43,16 +46,17 @@ public void Kill()
...
-        public void Start(string dataDir)
+        public void Start(string dataDir, bool mineCoinsFast)
         {
-            this.process = Process.Start(new FileInfo(this.bitcoinD).FullName,
-                $"-conf=bitcoin.conf -datadir={dataDir} -debug=net");
+            this.process = Process.Start(new FileInfo(this.bitcoinD).FullName, $"-conf=bitcoin.conf -datadir={dataDir} -debug=net");
         }
     }
 
     public class StratisBitcoinPosRunner : INodeRunner
```
ferdeen: Notice a lot of duplicate code in NodeRunner.cs.  At some point good to pull out the common code into a base class.  Something like this for staking logic.

image

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,167 @@
...
+        private Key posReceiverPrivateKey;
+
+        private const string PowWallet = "powwallet";
+        private const string PowWalletAccount = "account 0";
+        private const string PowWalletPassword = "password";
+
+        private const string PosWallet = "poswallet";
+        private const string PosWalletAccount = "account 0";
+        private const string PosWalletPassword = "password";
+
```
ferdeen: if the account and password are the same can remove two of the constants and share the other two.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,167 @@
...
+                    .MineCoinsFast()
+                    .Start()
+                    .NotInIBD()
+                    .WithWallet(PowWallet, PowWalletPassword)
+                    .Build();
+
+            this.powSenderAddress = this.nodes[PowMiner].FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference(PowWallet, PowWalletAccount));
+            var wallet = this.nodes[PowMiner].FullNode.WalletManager().GetWalletByName(PowWallet);
+            this.powSenderPrivateKey = wallet.GetExtendedPrivateKeyForAddress(PowWalletPassword, this.powSenderAddress).PrivateKey;
+        }
```
ferdeen: The above three lines are repeated further down, maybe move into it's own method.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,167 @@
...
+            };
+
+            foreach (var item in unspent.OrderByDescending(a => a.Transaction.Amount))
+            {
+                coins.Add(new Coin(item.Transaction.Id, (uint)item.Transaction.Index, item.Transaction.Amount, item.Transaction.ScriptPubKey));
+            }
+
+            var coin = coins.First();
+            var txIn = transaction.AddInput(new TxIn(coin.Outpoint, this.powSenderAddress.ScriptPubKey));
+            transaction.AddOutput(new TxOut(new Money(9699999999995400), this.powSenderAddress.ScriptPubKey));
```
ferdeen: I am not too sure this would work.  The statement at line 116 is the same as just getting the first item in the list - unless this is not consistent when the list is being built.  Also the line could be moved closer to where it's used.

So something like this 

image

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: OK Ive added this to a \Extensions folder.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Fixed

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
fassadlr: Done

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: Yeah we will do this soon!

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: Done

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: @ferdeen they are similar but they set different variables ;)

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -93,23 +94,25 @@ public static FullNode BuildFullNode(NodeSettings args, Action<IFullNodeBuilder>
...
                     .UseBlockStore()
                     .UseMempool()
                     .UseWallet()
                     .AddPowPosMining()
                     .AddRPC()
-                    .MockIBD()
-                    .Build();
+                    .MockIBD();
+
+                if (this.mineCoinsFast)
```
dangershony: Create a new runner and remove this flag `mineCoinsFast`

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -192,33 +262,70 @@ public static FullNode BuildFullNode(NodeSettings args, Action<IFullNodeBuilder>
...
-                }
-            });
-
-            return fullNodeBuilder;
+                var builder = new FullNodeBuilder()
+                    .UseNodeSettings(args)
+                    .UsePosConsensus()
+                    .UseBlockStore()
+                    .UseMempool()
+                    .AddMining()
```
dangershony: use here `AddPosPowMining`

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            this.sharedSteps.WaitForNodeToSync(this.nodes[PosStaker]);
+
+            var received = this.nodes[PosStaker].FullNode.WalletManager().GetSpendableTransactionsInWallet(PosWallet);
+            received.Sum(s => s.Transaction.Amount).Should().Be(Money.COIN * 1000000);
+        }
+
+        private void pos_node_mines_ten_blocks_more_ensuring_they_can_be_staked()
+        {
+            this.nodes[PosStaker].GenerateStratisWithMiner(10);
```
dangershony: Ideally put the ConsensusOptions.Maturity (instead of had coding 10)

src/Stratis.Bitcoin.IntegrationTests/SharedSteps.cs
```diff
@@ -50,14 +50,31 @@ public void MineBlocks(int blockCount, CoreNode node, string accountName, string
...
+            var wallet = node.FullNode.WalletManager().GetWalletByName(walletName);
+            var extendedPrivateKey = wallet.GetExtendedPrivateKeyForAddress(walletPassword, unusedAddress).PrivateKey;
+
+            node.SetDummyMinerSecret(new BitcoinSecret(extendedPrivateKey, node.FullNode.Network));
+            node.GenerateStratisWithMiner(2);
+
+            this.WaitForNodeToSync(node);
+
+            var spendable = node.FullNode.WalletManager().GetSpendableTransactionsInWallet(walletName);
+            spendable.Sum(s => s.Transaction.Amount).Should().Be(Money.COIN * 98000004);
```
dangershony: Also here you could reuse the consensus reward properties (ideally)

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: Fixed!

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
fassadlr: Fixed!

src/Stratis.Bitcoin.IntegrationTests/SharedSteps.cs
fassadlr: Fixed!

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
fassadlr: We can only use `AddMining` as `AddPosPowMining` uses the PosAssemblerFactory which uses `PosMinting`.

zeptin: `PosAssemblerFactory` does not necessarily use `PosMinting`, but if you use it to make a PoW block the coinbase transaction is hardcoded to be empty. This is a bug (1113#).

PR #1292, opened 2018-04-11
Future drift fix

noescape00:
This pr is basically created by @Aprogiena

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosFutureDriftRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosFutureDriftRule.cs
@@ -9,17 +9,25 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
     /// </summary>
     public class PosFutureDriftRule : PosConsensusRule
     {
+        /// <summary>New future drift in seconds after the hardfork.</summary>
+        private const int NewFutureDriftSeconds = 15;
+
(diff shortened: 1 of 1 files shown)


Aprogiena:
The mistake was quite innocent as the result of the mistake was that such an invalid block would not be banned properly as we intended before. Instead it's bad would expire 5 days in history, which basically would mean it would not be banned to the future at all.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosFutureDriftRule.cs
```diff
@@ -9,17 +9,25 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
...
             Block block = context.BlockValidationContext.Block;
 
+            long adjustedTime = this.Parent.DateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
+
             // Check timestamp.
-            if (block.Header.Time > this.FutureDrift(this.Parent.DateTimeProvider.GetAdjustedTimeAsUnixTimestamp()))
+            if (block.Header.Time > adjustedTime + this.GetFutureDrift(adjustedTime))
             {
                 // The block can be valid only after its time minus the future drift.
-                context.BlockValidationContext.RejectUntil = Utils.UnixTimeToDateTime(block.Header.Time - this.FutureDrift(0)).UtcDateTime;
```
dangershony: hoo that was the mistake `this.FutureDrift(0)`

PR #1296, opened 2018-04-12
[Tests] #makethebuildgreen 

fassadlr:


bokobza:
Did you figure out what changed and made these tests fail?

bokobza:
@fassadlr that's what I thought, but were some tests added or changed?

fassadlr:
@bokobza sorry dude, I wouldnt know as I havent been on the full node for a few months, so I can't tell you off the bat what tests were added and at point they started failing. I just saw red and investigated and saw that its the same old issues with static flags. I updated the test to use Stratis networks as they expect the IsProofStake flag to be true and they all passed again.

PR #1297, opened 2018-04-12
Remove project dependency between unit tests

majikandy:
Copied the shared code so that each project has their own copy of the methods in use, so that the project references between test projects are removed (except Stratis.Bitcoin.Tests, that is the next task)


Neurosploit:
@dangershony I think it could be a good idea create a single test utilities project at some point with some of these commonly used methods if they can be shared. What do you think?

majikandy:
@Neurosploit I tend to agree, I just changed this PR to do that.  The new assembly is Stratis.Tests.Common - this could later be nugetted and shared across other projects too (where relevant)

src/Stratis.Bitcoin.Features.RPC.Tests/Controller/FullNodeControllerTest.cs
```diff
@@ -687,5 +687,52 @@ private byte[] GetBytes(uint compact)
                 (byte)(compact)
             };
         }
+
+        public class TestReadOnlyNetworkPeerCollection : IReadOnlyNetworkPeerCollection
```
bokobza: Where is this used?

src/Stratis.Tests.Common/Stratis.Tests.Common.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.2.0" />
+    <PackageReference Include="Moq" Version="4.8.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Stratis.Bitcoin.Features.Wallet\Stratis.Bitcoin.Features.Wallet.csproj" />
```
bokobza: The wallet helpers class is used by a few wallet projects but it's fine if it's in side the wallet.tests project.
Having a reference to the wallet feature defeats the purpose I think.
If we ever want to have Tests.Common as a nuget, it'll be dependant on many other packages.

majikandy: Perhaps yes the wallet tests helper class in here is a step too far. I did initially duplicate the re-used code to break the dependency between unit tests. As only a small part of it was actually being used and the majority was used in the main wallet test only. Right now essentially only those 30 lines of code is shared code, not the full 500.   If there is a desire to reference one unit test project from another, then we could just collapse the unit test projects into 1 and use folders for the separation and grouping of the tests per assembly. I raised an issue to discuss this,  becasue a single unit test project can be very effective.  It helps VS cope better too when you do that too.

src/Stratis.Bitcoin.Features.RPC.Tests/Controller/FullNodeControllerTest.cs
majikandy: Somewhere up in that test class I think, copied it here because it was used from the reference that was taken away but it can probably just as easily be a mock instead.

src/Stratis.Tests.Common/Stratis.Tests.Common.csproj
fassadlr: Guys, this is a tricky one... I'm not really sure why we would ever want to Nuget Tests.Common? But then again, Tests.Common should only contain Stratis agnostic code, if you want to purist about it (and if going to go down that routeO. I think for now the best think would just have to be to duplicate the wallet helper code out the to test assemblies that require them. That, or we have a Wallet.Tests.Common assembly but that seems like overkill. At the end of the day we can't have unit test assemblies referencing each other.

majikandy: My preference is collapsing the unit tests into a single unit test project. Then the wallet code can be used by any test that needs it. If we need to keep them separate for another reason, there is no real harm in duplicating the used code as per a couple of commits ago on this branch. Yes is breaks DRY but it is really quite minor considering the tests themselves are the protection layer around that code.  The idea of having a nugetted test common was really only about having the shared Stratis toolset that any test project can add, and wallettesthelper slipping in here in hindsight was a mistake.

fassadlr: >My preference is collapsing the unit tests into a single unit test project.

I cannot imagine how we are going coordinate such an undertaking at this point with all the open PRs and multiple other projects (Smart Contracts / ICO / SideChains) tracking the FN master repo. So I don't think this will be possible at this time. But we still need to de-reference those test assemblies regardless. So whilst it not optimal, duplicating the wallet helper code to the various assemblies that require them, is at least a step in the right direction ;)

majikandy: Ah yes, fair enough. Ok I’ll do that then so we can proceed, later if we want to consolidate that code into something shared, it really wouldn’t be that hard.

bokobza: `So whilst it not optimal, duplicating the wallet helper code to the various assemblies that require them, is at least a step in the right direction ;)`
Nope, it's not.
Why can't we have tests projects referencing each other?
I don't want to see duplication of code, certainly when there are alternatives.

fassadlr:
I personally dont have an issue with duplicating test helpers in test assemblies as its not production code. But I do understand that its not very good practise with regards to maintaining the code.

If you feel strongly about not duplicating the code then we need to move the wallet helpers to Tests.Common. 

The problem with having unit test projects referencing each other, is that Visual Studio and VSTS executes teste multiple times cause the referenced test assemblies gets copied into bin folders as well which then gets picked up the runner and executes them. 

E.g. If unit test project A references B and C, then VSTS for example will run tests in A,B and C. If unit project D references A,B itll run D,A and B. So in this scenario B gets run twice.

bokobza:
I've never seen this happening in Visual Studio but maybe in VSTS, execution is different.
How do you run the tests there?

bokobza:
` That, or we have a Wallet.Tests.Common assembly but that seems like overkill. At the end of the day we can't have unit test assemblies referencing each other.`
That's my favorite option. Thanks @majikandy for pointing it out, I'd missed it.

bokobza:
Good stuff, thanks!

Neurosploit:
@bokobza @fassadlr I've also never seen that happen. Usually you don't put any unit tests in projects you want have for shared code so you can reduce maintenance and repeat yourself less often. Only with class inheritance I've seen tests show up in the test runner multiple times.

Issue #1307, opened 2018-04-18
Logs (log4net) folder does not respect the -datadir when working folder different from .dll


sondreb:
When launching the daemon from the .csproj location using PowerShell and specifying the working directory, the wallet files (written with FileStorage) is written relative to the working folder (which I would expect is correct), while the Logs folder is written relative to the .dll files, which is wrong.

```
Start-Process -FilePath 'dotnet' -WorkingDirectory 'C:\src\Stratis.StratisD' -ArgumentList 'run -testnet -datadir="StratStore1"'
```

I have fixed the issue and will supply a PR related to this bug report. The tests supplied are to verify that the implementation code is valid, but the tests doesn't test the actual code due to not very testable. I considered refactored code to make it test-able, but decided not to to make the PR smaller.

Aprogiena:
would it not make better sense to ensure that LogPath (and other paths in DataFolder) are rooted?

Aprogiena:
what i mean is that it makes better sense to me to address the issue in DataFolder rather then in each end code when it is used

sondreb:
Could be an idea to root all the paths in DataFolder, but that would be a bigger change that would need additional integration testing to verify. Perhaps accept this PR, and then we could investigate if rooting all the paths would make sense?

Aprogiena:
If there is issue with one path, it can very much happen with any other path. Fixing it at 1 place for everything seems to be better approach than just to fix every particular instance one by one. So I would suggest to go that way instead of this PR, but feel free to wait for opinions of others.

sondreb:
The issue is really with path for log files, produced by log4net, currently can end up somewhere other than the normal data files.

Changing DataFolder to root all the paths is likely a major refactoring job, considering there are hundreds of usage of DataFolder and many File IO operations done throughout the code base. If even a single one of those read/write operations changes position, it can be very problematic.

This does though bring up a potential new PR, and that is to rewrite all code that uses System.IO directly and make sure every interaction with the file system goes through the FileStorage class, as of right now it's spread a bit throughout the code base. Then when all IO is handled through the FileStorage abstraction, it is possible to be more certain that there won't be an issues. I wouldn't start rooting the paths unless that was done first.

Aprogiena:
[...]

PR #1308, opened 2018-04-18
Fix the path for logging when -datadir and working folder specified

sondreb:
- Included unit test verifies the implemented logic, but not the actual implementation. #1307

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Configuration/Logging/LoggingConfigurationTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Configuration/Logging/LoggingConfigurationTest.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Configuration.Logging
+{
+    public class LoggingConfigurationTest
+    {
+        /// <summary>
+        /// The test will verify the bug #1307, but does not test the implementation directly.
+        /// </summary>
+        [Fact]
+        public void PartialPathShouldBePrefixed()
+        {
+            var logPath = "StratStore1";
+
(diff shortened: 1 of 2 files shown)


Aprogiena:
Please see my comments in #1307

bokobza:
From what I see when running the node, the Datafolder paths are always rooted so I'm not sure what this fixes.
If this is to support the case where a non-rooted DataFolder is passed, I'm not sure this is necessary.

Also the tests seems to be testing the system.IO.Path class rather than any of our classes. Is that intentional?

bokobza:
I stand corrected, they're not if you specify the datadir in the command line.
If there's no root in the DataFolder paths, the current directory will be used, so the questions remain.

bokobza:
@sondreb @Aprogiena  I found the issue.
In NodeSettings.cs if datadir is not null, we create a folder for it, but the DataDir property itself stays root-less, so all the paths in DataFolder are without a root.

Have a look:
```
// Setting the data directory.
if (this.DataDir == null)
{
	this.DataDir = this.CreateDefaultDataDirectories(Path.Combine("StratisNode", this.Network.RootFolderName), this.Network);
}
else
{
	// Create the data directories if they don't exist.
	string directoryPath = Path.Combine(this.DataDir, this.Network.RootFolderName, this.Network.Name);
	---> Directory.CreateDirectory(directoryPath);
    ---> this.DataDir = directoryPath;
	this.Logger.LogDebug("Data directory initialized with path {0}.", directoryPath);
}
```

The fix is to replace
```
Directory.CreateDirectory(directoryPath);
this.DataDir = directoryPath;
```
by 
`this.DataDir = Directory.CreateDirectory(directoryPath).FullName;`

and it should be ok.
@sondreb I'll let you do it.

Aprogiena:
So as we merged the other PR fixing this, we can probably close this, right?

bokobza:
yep.

PR #1309, opened 2018-04-19
Fix consensus test project tests not able to run

Neurosploit:
I just pulled all latest changes into my main branch which is equal with the stratis main branch and I was unable to run the consensus tests. Turns out the sdk was missing from the project file.
Added it back and now they run again.


bokobza:
Thanks mate!

PR #1314, opened 2018-04-22
Added example logfile

noescape00:
This is an example logging config file that enables details logs for every part of the codebase. Logs are kept for 14 days.


bokobza:
Excellent stuff

PR #1324, opened 2018-04-23
Convert api test to bdd style

majikandy:
Converted the test to the expressive style - at this stage the test wasn't improved. But this has shown that there are some holes in the tests that could do with improvement. 

I am open to including them in this PR but wanted to present this change as is without modifying coverage too much.


src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
```diff
@@ -0,0 +1,40 @@
...
+        [Fact]
+        public void getgeneralinfo_returns_non_empty_json()
+        {
+            Given(a_proof_of_work_node_api);
+            When(getting_general_info);
+            Then(data_starting_with_wallet_file_path_is_returned);
+        }
+
+        [Fact]
+        public void startstaking_starts_Staking()
```
ferdeen: Test method is not clear for this test.

fassadlr:
Hey Andy, can you please merge latest from master so that we can fix the conflicts and that the tests pass?

majikandy:
Yeah, I'm on it - quite a change in the original I converted so I am going through line by line - shouldn't be too long

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -1,140 +0,0 @@
...
-namespace Stratis.Bitcoin.IntegrationTests.Api
-{
-    public class APITests : IDisposable
-    {
-        private bool initialBlockSignature;
-        private bool initialTimeStamp;
-
-        public APITests()
-        {
-            this.initialBlockSignature = Transaction.TimeStamp;
```
fassadlr: Please add this to the `BeforeTest()` method so that we can stop a possible static flag issue.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
```diff
@@ -1,140 +0,0 @@
...
-            this.initialTimeStamp = Block.BlockSignature;
-
-            Transaction.TimeStamp = true;
-            Block.BlockSignature = true;
-        }
-
-        public void Dispose()
-        {
-            Transaction.TimeStamp = this.initialBlockSignature;
```
fassadlr: Please add this to the `AfterTest()` method.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification_Steps.cs
```diff
@@ -0,0 +1,133 @@
...
+                this.httpClient = null;
+            }
+
+            this.nodeBuilder.Dispose();
+            this.nodeGroupBuilder.Dispose();
+        }
+
+        private void a_proof_of_stake_node_api()
+        {
+            // TODO BEFORE PR APPROVAL: Perhaps the nodegroup builder below instead
```
fassadlr: Yeah use the new `NodeGroupBuilder`.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification_Steps.cs
```diff
@@ -0,0 +1,133 @@
...
+        private void getting_general_info()
+        {
+            this.httpClient.DefaultRequestHeaders.Accept.Clear();
+            this.httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(JsonContentType));
+            this.response = this.httpClient.GetStringAsync(this.apiUri + "api/wallet/general-info?name=testapi").GetAwaiter().GetResult();
+        }
+
+        private void data_starting_with_wallet_file_path_is_returned()
+        {
+            // TODO BEFORE PR APPROVAL: improve this check - change method name to information_about_the_wallet_is_returned and implement
```
fassadlr: Can rename and remove TODO.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification_Steps.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        private void the_blockhash_is_returned()
+        {
+            this.response.Should().Be("\"93925104d664314f581bc7ecb7b4bad07bcfabd1cfce4256dbd2faddcf53bd1f\"");
+        }
+
+        private void non_empty_list_returned()
+        {
+            // TODO BEFORE PR APPROVAL: improve this check  - change method name to the_available_rpc_methods_are_listed and implement
```
fassadlr: Can rename and remove TODO.

src/Stratis.Bitcoin.IntegrationTests/APITests.cs
majikandy: same as below

majikandy:
I've made the changes for the static flags and then a few other cleanups and merged the latest master into this (but master unfortunately now has a couple of unit test failures again).

majikandy:
Ok those were just fixed by @Neurosploit so I have merged into this PR

majikandy:
Yes @bokobza

bokobza:
I haven't had a look at these new way of doing tests at all but generally speaking:

* Why are the specification tests method names all lower case (e.g. `getgeneralinfo_returns_json_starting_with_wallet_path`)?
* Why is there an underscore after steps (e.g ApiSpecification_Steps.cs)?
* What's described as steps are not really expressing or representing steps at all. Some are states, some are actions or steps, some are expected results.
* There are partial classes for no real reason. It seems the purpose of the split is to separate the class in "steps" and "specification", but in reality is adding a bit of confusion and inaccuracy. 
* ApiTests has changed to ApiSpecification. That's also incorrect naming. The specification belong in a spec document, but Tests is more appropriate here. Maybe ApiSpecificationTests is the best name.

majikandy:
>Why are the specification tests method names all lower case (e.g. getgeneralinfo_returns_json_starting_with_wallet_path)?

snake case... Party to be different and easily identifyable and partly so that it is easier to read and closer to english.  I would be happy capitalising first letters, in this context I was thinking that the RPC call itself getgeneralinfo is lowercase so that is why no capital at the start

>Why is there an underscore after steps (e.g ApiSpecification_Steps.cs)?

This was to jar the eye and draw attention to the two classes being tightly coupled as one

>What's described as steps are not really expressing or representing steps at all. Some are states, some are actions or steps, some are expected results.

The intention is that they are execution steps, no matter whether used by the Specifications as Given When or Then actions.  Yes there is state in there too which is sort of a compromise to keep the Specification file as clean as humanly possible for easy reading with as little gunk as possible.

>There are partial classes for no real reason. It seems the purpose of the split is to separate the class in "steps" and "specification", but in reality is adding a bit of confusion and inaccuracy.

What is the confusion? That you didn't realise they are just the same class? Perhaps that is an initial confusion but once you realise, that goes away maybe?

>ApiTests has changed to ApiSpecification. That's also incorrect naming. The specification belong in a spec document, but Tests is more appropriate here. Maybe ApiSpecificationTests is the best name.

It was incorrect against the policy prior to introducing this style - calling it specification is a BDD thing of trying to make you think about this being a specification for the system and not just codey tests.  Also at this stage where there is a mixture of the two styles, having the ones that are BDD style stand out is not a bad thing, but would need a little flexibility adding to the policy I agree.

majikandy:
Region is the other option yes, and you are right it does make the naming of the other file go away.

Generally I don't like regions because they only seem to exist to hide stuff, and if you are hiding it makes me think "why are you hiding it".



The other specs so far have the 2 separate files, perhaps I will do this one as region as you say and then we have both ways in the integration tests and we can decide?

>but the test method names should start with an upper case 

ok

>and have an upper case after the underscore.

currently :  getgeneralinfo_returns_json_starting_with_wallet_path()

I think the following looks ok

Get_general_info_returns_json_starting_with_wallet_path()

but to me...

`Get_General_Info_Returns_Json_Starting_With_Wallet_Path()`

just looks like bad english as you would never write..

`Get General Info Returns Json Starting With Wallet Path`

and
`GetGeneralInfoReturnsJsonStartingWithWalletPath()`

just looks like very classic code, and the idea is just to use every trick in the book to improve readability for both coders and even non coders.



Another reason for snake case is that in an expressive test method name is that it allows you to emphasise words with capitals to make them leap off the page in your face... Eg...    X_happening_to_y_makes_z_NOT_happen  which @dangershony, @ferdeen and I did in a few of the other bdd style tests.


I would be quite up for emphasising that it is a scenario too ... and keeping it Capitalised snake case.
Perhaps   ScenarioAttribute : FactAttribute  in the base  and then [Scenario] over the method


>no underscore in the class name

You mean filename, but ok

bokobza:
GitHub won't collapse it indeed.

src/Stratis.Bitcoin.IntegrationTests/TestFramework/BddSpecification.cs
```diff
@@ -11,16 +12,24 @@ public abstract class BddSpecification : IDisposable
...
             this.startOfTestTime = DateTime.UtcNow;
+
+            this.StoreStaticFlagPreTestValues();
             this.BeforeTest();
         }
 
         public void Dispose()
         {
             this.AfterTest();
+            this.RestoreStaticFlagValues();
```
majikandy: @monsieurleberre see this - the static flags isolator would be useful here

bokobza:
Let's do Pascal snake case then :-)

majikandy:
Pascal snake is like worst of both :)  (IMO)

bokobza:
I mean this one `Get_general_info_returns_json_starting_with_wallet_path()`

Issue #1315, opened 2018-04-22
Checkpoints are always disabled


noescape00:
If `settings.UseCheckpoints` is `false` we won't load the checkpoints:

Checkpoints.cs: 157

```
if (!settings.UseCheckpoints) this.checkpoints = new Dictionary<int, CheckpointInfo>();
```

The problem is that we call the checkpoints constructor and check that `settings.UseCheckpoints` before we load the value in ConsensusSettings: 41

```
this.UseCheckpoints = config.GetOrDefault<bool>("checkpoints", true);
```

So basically checkpoints are always disabled no matter what.

ps
I also suggest adding test for the use-checkpoints switch after the issue is fixed.

noescape00:
@dangershony

Aprogiena:
How is that possible? It worked before well, have we changed something?

dangershony:
@Neurosploit all yours

dangershony:
This was fixed closing

PR #1329, opened 2018-04-23
 Checkpoints are always disabled (#1315)

Neurosploit:
Fix for https://github.com/stratisproject/StratisBitcoinFullNode/issues/1315 and unit tested the whole class.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
@@ -134,46 +134,54 @@ namespace Stratis.Bitcoin.Features.Consensus
         };
 
         /// <summary>Checkpoints for the specific instance of the class and its network.</summary>
-        private readonly Dictionary<int, CheckpointInfo> checkpoints;
+        private readonly Dictionary<int, CheckpointInfo> networkSpecificCheckpoints;
+
+        /// <summary>Empty list of checkpoints for the specific instance of the class in case the use of checkpoints is disabled in the consensus settings.</summary>
+        private readonly Dictionary<int, CheckpointInfo> emptyCheckpoints = new Dictionary<int, CheckpointInfo>();
+
+        /// <summary>Consensus settings for the full node.</summary>
+        private ConsensusSettings ConsensusSettings { get; }
 
         /// <summary>
         /// Initializes a new instance of the object.
         /// </summary>
         public Checkpoints()
         {
-            this.checkpoints = new Dictionary<int, CheckpointInfo>();
+            this.networkSpecificCheckpoints = new Dictionary<int, CheckpointInfo>();
         }
 
         /// <summary>
         /// Initializes a new instance of the object.
         /// </summary>
         /// <param name="network">Specification of the network the node runs on - regtest/testnet/mainnet/stratis test/main.</param>
-        /// <param name="settings">Consensus settings for node - used to see if checkpoints have been disabled or not.</param>
-        public Checkpoints(Network network, ConsensusSettings settings)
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -134,46 +134,54 @@ public class Checkpoints : ICheckpoints
         };
 
         /// <summary>Checkpoints for the specific instance of the class and its network.</summary>
-        private readonly Dictionary<int, CheckpointInfo> checkpoints;
+        private readonly Dictionary<int, CheckpointInfo> networkSpecificCheckpoints;
+
+        /// <summary>Empty list of checkpoints for the specific instance of the class in case the use of checkpoints is disabled in the consensus settings.</summary>
+        private readonly Dictionary<int, CheckpointInfo> emptyCheckpoints = new Dictionary<int, CheckpointInfo>();
```
bokobza: We usually avoid instantiating objects outside of the ctor.
Also it seems this object is never filled (empty, I get it), and then only returned in `GetCheckpoints()`. Is it really necessary to have?

Neurosploit: I could just return a new empty one instead.

src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -182,8 +190,16 @@ public bool CheckHardened(int height, uint256 hash)
...
         {
             CheckpointInfo checkpoint;
-            this.checkpoints.TryGetValue(height, out checkpoint);
+            this.GetCheckpoints().TryGetValue(height, out checkpoint);
             return checkpoint;
         }
+
+        private Dictionary<int, CheckpointInfo> GetCheckpoints()
+        {
+            if (this.ConsensusSettings == null)
```
bokobza:
```
if (this.ConsensusSettings == null || !this.ConsensusSettings.UseCheckpoints)
    return this.emptyCheckpoints;

return this.networkSpecificCheckpoints;
```
seems more condensed to me.

Neurosploit: Yep

noescape00:
Yes, sure. Will take a look in couple of hours.

dangershony:
Looks good to me.
@Neurosploit eventually the checkpoints themselves need to go to the network file 
If you are up for it to do another PR after this to move checkpoints to Network and add them to the network builder.

src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -182,8 +187,16 @@ public bool CheckHardened(int height, uint256 hash)
...
         {
             CheckpointInfo checkpoint;
-            this.checkpoints.TryGetValue(height, out checkpoint);
+            this.GetCheckpoints().TryGetValue(height, out checkpoint);
             return checkpoint;
         }
+
+        private Dictionary<int, CheckpointInfo> GetCheckpoints()
+        {
+            if (this.ConsensusSettings == null || !this.ConsensusSettings.UseCheckpoints)
```
noescape00: Can the `ConsensusSettings ` be null here?

noescape00:
Reviewed it, looks nice!

src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
Neurosploit: Yes it can if you call the parameterless constructor.

PR #1333, opened 2018-04-24
Blockstore api

monsieurleberre:
First REST function for the blockstore


src/Stratis.Bitcoin.Features.BlockStore/Models/BlockByHashQueryModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Stratis.Bitcoin.Features.BlockStore.Models
+{
+    public abstract class QueryModelBase
```
bokobza: For consistency, look at how we created and named those in https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs

src/NBitcoin/Block.cs
```diff
@@ -428,6 +429,23 @@ public static Block ParseJson(string json)
             return blk;
         }
 
+        public string ToJson(bool indented = false)
```
bokobza: I actually really think we should create a BlockModel that is only used by the controller.
The controller would then get the block and map it to a new blockModel object. This way there is no disruption to the Block object.
As for future proofing of Block vs SmartBlock serialization, we can always have custom serializers in place.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            this.blockStoreCache = blockStoreCache;
+            this.connectionManager = connectionManager;
+            this.network = network;
+            this.chain = chain;
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.dateTimeProvider = dateTimeProvider;
+        }
+
+        [Route("getblock")]
```
bokobza: `block` is enough as the endpoint name.

src/Stratis.Bitcoin.Features.BlockStore/Models/BlockByHashQueryModel.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public bool OutputJson { get; set; }
```
bokobza: Json should really be the default. Maybe `ShowAsHex`can be better? What do you think?

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
bokobza: Please add a comment on this method.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+                return BuildErrorResponse(this.ModelState);
+            }
+
+            this.logger.LogTrace("({0}:'{1}')", nameof(ObjectByHashQueryModel.Hash), query.Hash);
+
+            try
+            {
+                var block = await this.blockStoreCache.GetBlockAsync(uint256.Parse(query.Hash)).ConfigureAwait(false);
+                if(block == null) return new NotFoundObjectResult("Block not found");
+                return query.OutputJson
```
bokobza: I think a BlockModel here is a better option.

monsieurleberre:
This will be done using a Model instead of changing the NBitcoin.Block code

PR #1334, opened 2018-04-24
Resolve merge block assembler unit tests

Neurosploit:
This resolves the failing Pos/Pow block assembler unit tests after merging several commits into master lost some changes.


bokobza:
Yep, that looks good!

PR #1335, opened 2018-04-25
Blockstore api with model

monsieurleberre:
First REST function for the blockstore


src/Stratis.Bitcoin.Features.BlockStore/Models/Block.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+using NBitcoin;
+using NBitcoin.RPC;
+
+namespace Stratis.Bitcoin.Features.BlockStore.Models
+{
+    public class Block
```
bokobza: To keep in line with the other model objects in the solution, please call this object BlockModel.
Also, I'm afraid implicitly converting a Block to a Block might lead to confusion, even if they're on different namespaces.

src/Stratis.Bitcoin.Features.BlockStore/Models/Block.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+using NBitcoin;
+using NBitcoin.RPC;
```
bokobza: Is this necessary here?

src/Stratis.Bitcoin.Features.BlockStore/Models/ObjectByHashRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Stratis.Bitcoin.Features.BlockStore.Models
+{
+    public abstract class RequestBase
+    {
+        public bool OutputJson { get; set; }
+    }
+    public class ObjectByHashRequest : RequestBase
```
bokobza: Please add a new line here.

src/Stratis.Bitcoin.Features.BlockStore/Models/Block.cs
```diff
@@ -0,0 +1,50 @@
...
+                PreviousBlockHash = block.Header.HashPrevBlock.ToString(),
+                MerkleRoot = block.Header.HashMerkleRoot.ToString(),
+                Difficulty = block.Header.Bits.Difficulty,
+                Transactions = block.Transactions.Select(t => t.GetHash().ToString()).ToArray()
+            };
+            return blockModel;
+        }
+    }
+
+    internal static class BlockExtension
```
bokobza: I think having a way to create a BlockModel using its constructor like in `BlockModel blockModel = new BlockModel(block);` is more straightforward than doing it via an extension method. What do you think?

src/Stratis.Bitcoin.Features.BlockStore/Models/ObjectByHashRequest.cs
bokobza: Not sure this will always return an object. Like if we have a `DELETE /api/block?hash=dddddddd`.

src/Stratis.Bitcoin.Features.BlockStore/Models/Block.cs
```diff
@@ -0,0 +1,50 @@
...
+        public int Version { get; set; }
+        public string Bits { get; set; }
+        public DateTimeOffset Time { get; set; }
+        public string[] Transactions { get; set; }
+        public double Difficulty { get; set; }
+        public string MerkleRoot { get; set; }
+        public string PreviousBlockHash { get; set; }
+        public uint Nonce { get; set; }
+
+        public static implicit operator Block(NBitcoin.Block block)
```
bokobza: This operator and the extension method below can be replaced by a ctor taking in an NBitcoin.Block.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+        public BlockStoreController(
+            ILoggerFactory loggerFactory,
+            IBlockStoreCache blockStoreCache,
+            IConnectionManager connectionManager,
+            Network network,
+            ConcurrentChain chain,
+            IBroadcasterManager broadcasterManager,
+            IDateTimeProvider dateTimeProvider)
+        {
+            this.blockStoreCache = blockStoreCache;
```
bokobza: Are you using connectionManager, network, chain and dateTimeProvider?

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Features.BlockStore.Models;
+using Stratis.Bitcoin.Interfaces;
+using Stratis.Bitcoin.Utilities;
+using Stratis.Bitcoin.Utilities.JsonErrors;
+
+namespace Stratis.Bitcoin.Features.BlockStore.Controllers
+{
+    /// <summary>
+    /// Controller providing operations on a wallet.
```
bokobza: "on the block store"

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Runtime.Serialization.Json;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+using NBitcoin.RPC;
```
bokobza: I think a few usings can be removed.

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
```diff
@@ -0,0 +1,99 @@
...
+            this.connectionManager = connectionManager;
+            this.network = network;
+            this.chain = chain;
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.dateTimeProvider = dateTimeProvider;
+        }
+
+        [Route("block")]
+        [HttpGet]
+        public async Task<IActionResult> GetBlockAsync([FromQuery] ObjectByHashRequest query)
```
bokobza: Maybe call it `GetBlockByHashAsync` if we're going to potentially have a `GetBlockByTransactionAsync`.
Or, instead of having two methods, you could have 1 method `/api/block` that does both.
For that you'd have to change the `ObjectByHashRequest` to be `BlockRequest` (that has both a block hash and a trx hash), and the same controller method can see whether a block hash has been passed or a trx hash has been passed and act appropriately.

monsieurleberre: I will call it GetBlockByHashAsync because I think having 1 single method for 2 things means we need to try one first then if it returns null, try the other one. Potentially every request for transaction hash will happen after a request for block hash (or vice versa), even when you know what you are querying for

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
monsieurleberre: definitely ! thanks

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
bokobza: That' fine but then you'll need 2 endpoints, right?
you can have one method for 2 things if the request has `hash` and `trxHash` for example.
So you would do:
```
if query.Hash
    --find a block by hash
else if query.trxHash
    -- find a block by trx hash
```

monsieurleberre: OK that is good, so for now no changes, and when we want to be able to search by transactionHash, then we'll do it this way

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
monsieurleberre: No I am not, sorry...

src/Stratis.Bitcoin.Features.BlockStore/Controllers/BlockStoreController.cs
bokobza: Perfect

noescape00:
Why do we need API method for get block if it's already implemented in RPC?

bokobza:
Do we?
Ultimately I'd like to have all the RPC methods moved/reproduced to use the API. It's more ubiquitous.

noescape00:
>Do we?

Not entirely sure. We have client part for getblock, probably not the server. 

>Ultimately I'd like to have all the RPC methods moved/reproduced to use the API.

That's a good idea.

PR #1340, opened 2018-04-25
Static Flags Issue - Store previous value to restore things as they were before the test

monsieurleberre:
an easy one :)


bokobza:
Nice one!

PR #1342, opened 2018-04-26
[Integration] Fix the POS test

fassadlr:
Fix test by implementing a better date time provider that is more in line with the actual node works. I.e. we don't do any time manipulation on the block header and transaction level.

Also, needed to fix the StaticFlagIsolator as we should only use it to store the initial flag values and reset them after the test has completed.


fassadlr:
Need to fix something...

PR #1343, opened 2018-04-26
[Tests] Fix POW/POS integration test

fassadlr:
Implement a fast coins date time provider thats more in line with the node's provider.

This PR ONLY fixes the POW/POS integration tests. I'm still looking into the others.


bokobza:
That looks pretty good, thanks mate.

PR #1345, opened 2018-04-26
Rooted datadir

bokobza:
This is in relation to https://github.com/stratisproject/StratisBitcoinFullNode/pull/1308 and https://github.com/stratisproject/StratisBitcoinFullNode/issues/1307.
@Aprogiena


Aprogiena:
I'm not sure about the changes in the tests, I am not familiar with those tests. But the main change looks good, makes perfect sense. However, no tests from me, so someone should check that it fixes the problem.

bokobza:
The problem was that since we didn't have a property giving the root path of DataFolder, people used other values that happens to be the same, like AddressManagerFilePath.
So I added a root path and fixed the tests.

bokobza:
@monsieurleberre @ferdeen @majikandy can one of you please have a look at this?

majikandy:
Looks fine to me

bokobza:
cool stuff, thanks guys.

PR #1347, opened 2018-04-27
[SmartContracts] Refactor Block Assemblers Part #1

fassadlr:
This is the first PR refactoring the block assemblers to implement a better inheritance strategy, primarily to facilitate integrating the smart contract feature into the full node. This is also key to moving smart contracts to proof of stake after alpha release.

Currently the POS assembler inherits from the POW assembler. In the smart contract branch we also inherit from the POW assembler. We need to move base logic to a abstract base block assembler so that all the different block assemblers are logically more modular.

The next PR will rename the block assemblers to block builders and then we will move the assemblers to individual folder and create the base class.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -314,7 +314,6 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <param name="network">Specification of the network the node runs on - regtest/testnet/mainnet.</param>
         /// <param name="connectionManager">Provider of information about the node's connection to it's network peers.</param>
         /// <param name="dateTimeProvider">Provides date time functionality.</param>
-        /// <param name="blockAssemblerFactory">Provides an interface for creating block templates of different types.</param>
         /// <param name="initialBlockDownloadState">Provider of IBD state.</param>
         /// <param name="nodeLifetime">Global application life cycle control - triggers when application shuts down.</param>
         /// <param name="coinView">Consensus' view of UTXO set.</param>
@@ -515,7 +514,7 @@ namespace Stratis.Bitcoin.Features.Miner
                 {
                     UnspentOutputs set = coinset.UnspentOutputs.FirstOrDefault(f => f?.TransactionId == infoTransaction.Transaction.Id);
                     TxOut utxo = (set != null) && (infoTransaction.Transaction.Index < set.Outputs.Length) ? set.Outputs[infoTransaction.Transaction.Index] : null;
-                    uint256 hashBlock = set != null ? this.chain.GetBlock((int) set.Height)?.HashBlock : null;
+                    uint256 hashBlock = set != null ? this.chain.GetBlock((int)set.Height)?.HashBlock : null;
 
                     if ((utxo != null) && (utxo.Value > Money.Zero) && (hashBlock != null))
                     {
--- a/src/Stratis.Bitcoin.Features.Miner/PowBlockAssembler.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PowBlockAssembler.cs
@@ -9,7 +9,6 @@ using Stratis.Bitcoin.Features.Consensus.Interfaces;
 using Stratis.Bitcoin.Features.MemoryPool;
 using Stratis.Bitcoin.Features.MemoryPool.Interfaces;
 using Stratis.Bitcoin.Utilities;
-using Stratis.Bitcoin.Utilities.Extensions;
 
 namespace Stratis.Bitcoin.Features.Miner
 {
@@ -31,7 +30,7 @@ namespace Stratis.Bitcoin.Features.Miner
         public FeeRate BlockMinFeeRate = new FeeRate(PowMining.DefaultBlockMinTxFee);
 
         public bool IsProofOfStake = false;
-    };
+    }
 
     public class BlockTemplate
     {
--- a/src/Stratis.Bitcoin.Features.Miner/PowMining.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PowMining.cs
@@ -28,12 +28,10 @@ namespace Stratis.Bitcoin.Features.Miner
 
     public class PowMining : IPowMining
     {
-        /// <summary>The async loop we need to wait upon before we can shut down this feature.</summary>
-        private IAsyncLoop asyncLoop;
-
         /// <summary>Factory for creating background async loop tasks.</summary>
         private readonly IAsyncLoopFactory asyncLoopFactory;
 
+        /// <summary>Thread safe chain of block headers from genesis.</summary>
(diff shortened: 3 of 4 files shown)


Aprogiena:
yeah, i want to check this one

src/Stratis.Bitcoin.Features.Miner/PowBlockAssembler.cs
```diff
@@ -31,9 +31,7 @@ public class AssemblerOptions
         public FeeRate BlockMinFeeRate = new FeeRate(PowMining.DefaultBlockMinTxFee);
 
         public bool IsProofOfStake = false;
-    }
-
-;
+    };
```
noescape00: `;` is redundant

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -534,14 +528,24 @@ public void StopStake()
...
+
                         this.logger.LogTrace("UTXO '{0}' with value {1} might be available for staking.", utxoStakeDescription.OutPoint, utxo.Value);
                     }
                 }
 
                 this.logger.LogTrace("Wallet contains {0} coins.", new Money(totalBalance));
 
                 if (blockTemplate == null)
-                    blockTemplate = this.blockAssemblerFactory.Create(chainTip, new AssemblerOptions() { IsProofOfStake = true }).CreateNewBlock(new Script());
+                    blockTemplate = new PosBlockAssembler(
```
noescape00: I think it will be a good idea to create a single PosBlockAssembler somewhere in constructor of PoSMinting or when staking starts and reuse it instead of recreating PosBlockAssembler every time consensus advances.

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -27,6 +28,20 @@ public ReserveScript(Script reserveFullNodeScript)
 
     public class PowMining : IPowMining
     {
+        /// <summary>The async loop we need to wait upon before we can shut down this feature.</summary>
+        private IAsyncLoop asyncLoop;
+
+        /// <summary>Factory for creating background async loop tasks.</summary>
+        private readonly IAsyncLoopFactory asyncLoopFactory;
+
+        private readonly ConcurrentChain chain;
```
noescape00: Comment is missing.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -335,7 +332,6 @@ public class CoinstakeContext
             Network network,
             IConnectionManager connectionManager,
             IDateTimeProvider dateTimeProvider,
-            IAssemblerFactory blockAssemblerFactory,
```
noescape00: Please remove `param name="blockAssemblerFactory"` from the comment

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -27,6 +28,20 @@ public ReserveScript(Script reserveFullNodeScript)
 
     public class PowMining : IPowMining
     {
+        /// <summary>The async loop we need to wait upon before we can shut down this feature.</summary>
+        private IAsyncLoop asyncLoop;
```
noescape00: maybe name it smth like `miningLoop` for the consistency? (in PoSMinting it's called stakingLoop so it makes sense to follow the same pattern here)

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -41,64 +56,69 @@ public class PowMining : IPowMining
...
-        private const int InnerLoopCount = 0x10000;
-
-        /// <summary>Manager of the longest fully validated chain of blocks.</summary>
-        private readonly IConsensusLoop consensusLoop;
-
-        private readonly ConcurrentChain chain;
+        private uint256 hashPrevBlock;
 
-        private readonly Network network;
+        private const int innerLoopCount = 0x10000;
```
noescape00: Can you please add the first capital letter? 

>CodingStyle: 12. We use PascalCasing to name all our constant local variables and fields.

src/Stratis.Bitcoin.Features.Miner/Stratis.Bitcoin.Features.Miner.csproj
```diff
@@ -23,6 +23,10 @@
     <EmbeddedResource Remove="Stratis.Bitcoin.Features.Miner\**" />
     <None Remove="Stratis.Bitcoin.Features.Miner\**" />
   </ItemGroup>
+
+  <ItemGroup>
```
noescape00: This change is probably not needed since AssemblerFactory was removed.

noescape00:
Yes. All unit tests and IntegrationTests.Miner - all are passing

Aprogiena:
How about syncing on mainnets Of bitcoin and Stratis?

noescape00:
Stratis mainnet- synced
BTC mainnet- synced first 40gb- all good.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I agree, the previous version of this code created a assembler every time which was wrong. This upgrade remove the factory, but did not address the main problem. Fixing the chain tip is only needed when we create the new block. Therefore it should not go into the constructor.

Aprogiena:
Basically I agree with the points that Ivan raised.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fassadlr: Removed!

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
fassadlr: Renamed (was gonna do variable renaming in next PR but we can do this one here.) 👍

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
fassadlr: Fixed.

src/Stratis.Bitcoin.Features.Miner/Stratis.Bitcoin.Features.Miner.csproj
fassadlr: Removed

fassadlr:
@Aprogiena @noescape00 I have run all test and have synced main net and test net :)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
noescape00: Yes, sure 🙂

Issue #1341, opened 2018-04-25
[RF] Move checkpoints to network


Neurosploit:
Move the network specific checkpoints to Network and add them to the network builder so they are more closely to where they are supposed to be. This also makes the checkpoints class really tidy. 

As mentioned in #1329.

PR #1350, opened 2018-04-27
Move checkpoints to network (#1341)

Neurosploit:
This PR moves the checkpoints to the network file and adds methods to include them in the nodebuilder.
See https://github.com/stratisproject/StratisBitcoinFullNode/issues/1341

@dangershony please have a look. I'm still unsure where to move CheckpointInfo to.

### DIFF
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -66,7 +66,7 @@ namespace NBitcoin
         WITNESS_PUBKEY_ADDRESS,
         WITNESS_SCRIPT_ADDRESS
     }
-    
+
     public enum BuriedDeployments : int
     {
         /// <summary>
@@ -83,7 +83,7 @@ namespace NBitcoin
         /// Strict DER signature
         /// </summary>
         BIP66
-    }    
+    }
 
     public class Consensus
     {
@@ -131,7 +131,7 @@ namespace NBitcoin
                 get { return this.parameters[(int) index]; }
                 set { this.parameters[(int) index] = value; }
             }
-        }       
+        }
 
         public Consensus()
         {
@@ -330,6 +330,12 @@ namespace NBitcoin
             if (GetNetwork(builder.Name) != null)
                 throw new InvalidOperationException("The network " + builder.Name + " is already registered.");
 
+            if (builder.Genesis == null)
+                throw new InvalidOperationException("A genesis block needs to be provided.");
+
+            if (builder.Consensus == null)
+                throw new InvalidOperationException("A consensus needs to be provided.");
+
             Network network = new Network();
             network.Name = builder.Name;
             network.RootFolderName = builder.RootFolderName;
@@ -339,8 +345,7 @@ namespace NBitcoin
             network.DefaultPort = builder.Port;
             network.RPCPort = builder.RPCPort;
             network.genesis = builder.Genesis;
-            if(network.consensus != null && network.genesis != null)
-                network.consensus.HashGenesisBlock = network.genesis.GetHash();
+            network.consensus.HashGenesisBlock = network.genesis.GetHash();
 
             foreach (DNSSeedData seed in builder.Seeds)
             {
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -242,7 +242,7 @@ namespace NBitcoin
         {
             get
             {
-                return consensus.NetworkOptions;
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin.Features.Consensus/Checkpoints.cs
```diff
@@ -69,82 +45,17 @@ public CheckpointInfo(uint256 hash, uint256 stakeModifierV2 = null)
...
-        };
-
-        /// <summary>Checkpoints for the specific instance of the class and its network.</summary>
-        private readonly Dictionary<int, CheckpointInfo> networkSpecificCheckpoints;
+        /// <summary>The current network. </summary>
+        private Network network;
 
         /// <summary>Consensus settings for the full node.</summary>
-        private ConsensusSettings ConsensusSettings { get; }
+        private ConsensusSettings consensusSettings { get; }
```
mikedennis: I think this should be ConsensusSettings according to coding standard but confirming in slack.

dangershony: Its private so change is correct, should start with lower case.

src/NBitcoin/Network.cs
```diff
@@ -336,7 +363,8 @@ internal static Network Register(NetworkBuilder builder)
             network.DefaultPort = builder.Port;
             network.RPCPort = builder.RPCPort;
             network.genesis = builder.Genesis;
-            network.consensus.HashGenesisBlock = network.genesis.GetHash();
+            if(network.consensus != null && network.genesis != null)
```
dangershony: This `if` is not correct both `consensus` and `genesis` should be set so we need to throw if they are not set.

dangershony:
I have only one comment, very good change.

Neurosploit:
@mikedennis I will move it to a seperate file. Dan didn't have any preferences.

src/NBitcoin/Networks.cs
```diff
@@ -155,6 +155,32 @@ private static Network InitMain()
...
+                { 225430, new CheckpointInfo(new uint256("0x00000000000001c108384350f74090433e7fcf79a606b8e797f065b130575932")) },
+                { 250000, new CheckpointInfo(new uint256("0x000000000000003887df1f29024b06fc2200b55f8af8f35453d7be294df2d214")) },
+                { 279000, new CheckpointInfo(new uint256("0x0000000000000001ae8c72a0b0c301f67e3afca10e819efa9041e458e9bd7e40")) },
+                { 295000, new CheckpointInfo(new uint256("0x00000000000000004d9b4ef50f0f9d686fd69db2e03af35a100370c64632a983")) },
+                // Our own new checkpoints.
+                { 486000, new CheckpointInfo(new uint256("0x000000000000000000a2a8104d61651f76c666b70754d6e9346176385f7afa24")) },
+                { 491800, new CheckpointInfo(new uint256("0x000000000000000000d80de1f855902b50941bc3a3d0f71064d9613fd3943dc4")) },
+            };
+
+            foreach (var checkpoint in checkpoints)
```
noescape00: unvar please here and in the similar foreaches in this file

src/NBitcoin/Networks.cs
```diff
@@ -155,6 +155,32 @@ private static Network InitMain()
...
+            {
+                network.checkpoints.Add(checkpoint.Key, checkpoint.Value);
```
noescape00: Maybe make Network.checkpoints non-readonly and just assign this dictionary to it (probably through a new method like SetCheckpoints or smth)?

Neurosploit: Instead I will follow the other lines in this file and remove the foreach completely.

noescape00: Nice!

Aprogiena:
@noescape00 Have you tried to sync?

noescape00:
No. But I doubt that there will be any problem with that. 
Can sync if you want.

Aprogiena:
@noescape00 Yes please, always recheck all test plus basic syncing when we are doing reviews, at least up until a point when this is guaranteed to be checked automatically by our QA. Only Skip these tests in case of very small trivial PR's that cannot damage anything.

noescape00:
Well, I was almost 100% that this PR will work as intended. After all- it's just a movement of checkpoints to the network. 

Ran the tests- testnet and mainnet synced successfully.

Aprogiena:
then we can merge, tx

Issue #1230, opened 2018-03-21
Wallet file might not be written correctly to disk when FullNode crushes


maciejzaleski:
My FullNode has been abruptly stopped by Windows restarting to install updates. This has resulted in a incomplete content in the wallet file.
As a result when starting the node again I get an exception  message; [Incomplete wallet file.txt](https://github.com/stratisproject/StratisBitcoinFullNode/files/1834324/Incomplete.wallet.file.txt)

**Proposed solution**
 - Write the content of the wallet file to *.partial file while deserializing
 - Move currently used wallet file to *.bak file
 - Move *.partial file to the correct wallet file name

Since move operation is really a file rename it should be much quicker than writing file to disk it have less chance of failing then the current approach.
In case of the FullNode start issues we can always discard broken wallet file and use backup.

bokobza:
Nice one @maciejzaleski.
I'll have a look.

bokobza:
Fixed in #1351 .

Issue #1231, opened 2018-03-21
Node is not syncing


maciejzaleski:
This seems to happen in the following scenario:
 - User has started the node for the first time
 - User has restored the wallet
 - Before the wallet restore is completed user stops the FullNode
 - When starting the full node again it will thrown an exception [Wallet with no balance.txt](https://github.com/stratisproject/StratisBitcoinFullNode/files/1834317/Wallet.with.no.balance.txt)
 - The node is connected to peers and peers have higher blockheigh than the tested node

**From Console**
> ======Connection======
> Peer:[::ffff:13.64.76.48]:26178,         connected:outbound,        height:343122,        agent:/Stratis:2.0.0.3/
> Peer:[::ffff:94.102.63.207]:26178,       connected:outbound,        height:343127,         agent:StratisBitcoin:1.0.10
> Peer:[::ffff:54.37.46.84]:26178,         connected:outbound,        height:343127,        agent:/Stratis:2.0.0.3/
> Peer:[::ffff:10.0.0.1]:51150,            connected:inbound,         height:343117,         agent:StratisBitcoin:1.0.10
> Peer:[::ffff:51.141.45.80]:26178,        connected:outbound,        height:343127,        agent:/Stratis:2.0.0.3/
> Peer:[::ffff:40.113.118.138]:26178,      connected:outbound,        height:343127,        agent:/Stratis:2.0.0.3/

**From Swagger**
>  "consensusHeight": 343117,

bokobza:
@maciejzaleski Did this happen when you were using the UI?
> Before the wallet restore is completed user stops the FullNode
How did the full node stop?

bokobza:
Haven't been able to reproduce this.
However, the separation of writing (to a temp file) and reading of a file in #1351 should help with the error described above.
@maciejzaleski feel free to have a lok, test and close if you're happy.

maciejzaleski:
This problem has now been resolved.

PR #1351, opened 2018-04-27
Protect file from corruption while saving

bokobza:
This should fix https://github.com/stratisproject/StratisBitcoinFullNode/issues/1230 and hopefully https://github.com/stratisproject/StratisBitcoinFullNode/issues/1231.

I added a call to `File.Replace` when saving the file so that the file never gets in a corrupted state.
This call is atomic in the sense that the file is either replaced successfully or not replaced at all.

Additionally, I added a parameter to the `FileStorage.SaveFile` method where a backup file (.bak) will be saved every time the file is saved. So the backup will have the previous version, and the file will have the new version.

More here: https://msdn.microsoft.com/en-us/library/windows/desktop/hh802690%28v=vs.85%29.aspx

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/FileStorageTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/FileStorageTest.cs
@@ -227,6 +227,103 @@ namespace Stratis.Bitcoin.Tests.Utilities
             Assert.False(loadedObjects.Any());
         }
 
+        [Fact]
+        public void GivenSaveWithBackupIsCalled_WhenTheFileExistsandWasChanged_ThenABackupIsSaved()
+        {
+            // Arrange
+            TestObject testObject = new TestObject { Property1 = "prop1", Property2 = "prop2" };
+            string dir = this.GetFolderPathForTestExecution();
+            FileStorage<TestObject> fileStorage = new FileStorage<TestObject>(dir);
+            fileStorage.SaveToFile(testObject, "savedTestObject.json", true);
+            testObject.Property1 = testObject.Property1 + "-changed";
+            fileStorage.SaveToFile(testObject, "savedTestObject.json", true);
+
+            // Act
+            bool isFileExists = fileStorage.Exists("savedTestObject.json");
+            bool isBackupFileExists = fileStorage.Exists("savedTestObject.json.bak");
+            
+            // Assert
+            Assert.True(isFileExists);
+            Assert.True(isBackupFileExists);
+
+            TestObject loadedByFileName = fileStorage.LoadByFileName("savedTestObject.json");
+            TestObject loadedBackupByFileName = fileStorage.LoadByFileName("savedTestObject.json.bak");
+
+            Assert.Equal("prop1", loadedBackupByFileName.Property1);
+            Assert.Equal("prop2", loadedBackupByFileName.Property2);
+            Assert.Equal("prop1-changed", loadedByFileName.Property1);
+            Assert.Equal("prop2", loadedByFileName.Property2);
+        }
+
+        [Fact]
+        public void GivenSaveWithBackupIsCalled_WhenTheFileDidntExist_ThenTwoIdenticalFilesAreSaved()
+        {
+            // Arrange
+            TestObject testObject = new TestObject { Property1 = "prop1", Property2 = "prop2" };
+            string dir = this.GetFolderPathForTestExecution();
+            FileStorage<TestObject> fileStorage = new FileStorage<TestObject>(dir);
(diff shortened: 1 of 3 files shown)


Aprogiena:
The ultimate solution to this is to use DBreeze, but for now using file replace is probably good enough (although documentation does not really guarantees File.Replace to be atomic). However, you also have this 

```
 // If the file does not exist yet, create it.
+            if (!File.Exists(filePath))
+            {
+                File.WriteAllText(filePath, JsonConvert.SerializeObject(toSave, Formatting.Indented));
```

so you expect File.WriteAllText to be atomic, which it probably is not.

fassadlr:
@bokobza Why don't we just always save a back up file? Or, in what situations would we NOT want to save a back  up?

If we dont always want to save a back up file via the wallet manager then in what instances do we want to?

bokobza:
@Aprogiena The documentation states: 
> The needed atomicity here is that the changes either are completely applied or not applied at all, as an inconsistent state would render the file corrupt. 


> Why don't we just always save a back up file?

@fassadlr oh just because it's a few more operations. Also, when you call a method called SaveFile, you're not expecting a backup to be created for you by default. It's there if you want to use it though.
We could use a backup feature for example if we want to save a backup of a newly created wallet. This way if the wallet gets corrupted, we can just delete it and use the backup.


For File.WriteAllText, true, it's not guaranteed to be atomic. However, here we use it 1. on the creation of the file and 2. to create the temp file (when the file already exists).
In both occasions, if it fails, no big deal. In the first instance, the probability that the system will crash the first time a file is created is very slim, and in the second instance, only the temp file will be corrupted, the main file staying at is it.

I can do 2 things:
1. on creation, create a tmp file and then rename it -> if it fails while creating the tmp file, no big deal.
2. find another way to write the file atomically. I've searched for an out-of-the-box solution but couldn't find any.

In any case I'll also add a lock around the save operation.

Aprogiena:
@bokobza Okay, I'm happy with replace file.

So just change write all text. Of course the probability is low, but that's not a good argument, it will just happen later. Simply avoid any chance of corruption. Since it is not possible to write the file atomically, just use option one as you suggested.

bokobza:
you got it baby.

bokobza:
Works on Mac too.

src/Stratis.Bitcoin.Tests/Utilities/FileStorageTest.cs
```diff
@@ -227,6 +227,103 @@ public void GivenLoadByFileExtensionIsCalled_WhenFilesDontExist_ThenNoObjectsIsR
...
+
+            // Assert
+            Assert.True(isFileExists);
+            Assert.False(isBackupFileExists);
+            TestObject loadedByFileName = fileStorage.LoadByFileName("savedTestObject.json");
+            Assert.Equal("prop1-changed", loadedByFileName.Property1);
+            Assert.Equal("prop2", loadedByFileName.Property2);
+
+        }
+
```
monsieurleberre: Not sure it adds much value, but maybe a separate test to check the default behavior is false, and a test to see what happens when the false is used explicitly ...

src/Stratis.Bitcoin/Utilities/FileStorage.cs
```diff
@@ -29,18 +29,37 @@ public FileStorage(string folderPath)
...
         /// <param name="fileName">Name of the file to be saved.</param>
-        public void SaveToFile(T toSave, string fileName)
+        /// <param name="saveBackupFile">A value indicating whether to save a backup of the file.</param>
+        public void SaveToFile(T toSave, string fileName, bool saveBackupFile = false)
         {
             Guard.NotEmpty(fileName, nameof(fileName));
             Guard.NotNull(toSave, nameof(toSave));
 
             string filePath = Path.Combine(this.FolderPath, fileName);
+            string tempFilePath = $"{filePath}.temp";
```
monsieurleberre: I would get a random temp name using something like System.IO.Path.GetTempFileName(); so you reduce the chances of writing to a file that is already being locked or used in some way ...

monsieurleberre:
My suggestions are only suggestions, I am happy for that to go in with or without them :) 🍃 🦄 🏈

src/Stratis.Bitcoin/Utilities/FileStorage.cs
bokobza: That's a great suggestion. I'm keen to keep the temp file in the same folder as the principal so GetTempFileName won't work for us, but I'll find a way to create unique temp file names (probably using ticks) .

monsieurleberre:
bang on 💸

PR #1352, opened 2018-04-27
Added the NLog file to the solution

bokobza:
This way the file is more accessible to developers.

### DIFF
--- a/src/Stratis.Bitcoin/Configuration/Logging/NLog.config
+++ b/src/Stratis.Bitcoin/Configuration/Logging/NLog.config
@@ -1,0 +1,139 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- This is an example logging config file that enables details logs for every part of the codebase. Logs are kept for 14 days. -->
+<!-- For an explanation on how to use this file, go to https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md. -->
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true">
+  <targets> 
+    <!-- API -->
+    <target name="debugApiFile" xsi:type="AsyncWrapper" queueLimit="10000" overflowAction="Block" batchSize="1000">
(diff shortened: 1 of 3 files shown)


noescape00:
Can you please add this to the `Stratis.Bitcoin.csproj`?

```
  <ItemGroup>
    <ContentWithTargetPath Include="Configuration\Logging\NLog.config">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <TargetPath>NLog.config</TargetPath>
    </ContentWithTargetPath>
  </ItemGroup>
```

It will automatically copy nlog.config to the build directory in case the config file is not present there.
This will basically make full logging a default option. If the logging is not needed for some reason developer can just edit the file in the build dir to log nothing. 
This will solve the problem of noticing some bug and not being able to investigate it. 

Also since the file was copied from `\Documentation` I guess it will be a good idea to remove it from the original folder to avoid duplicates.

Aprogiena:
That is very interesting suggestion. At 1st I was like no way, but then I like it!

bokobza:
> Can you please add this to the Stratis.Bitcoin.csproj?

I don't really want to enable this by default, it logs 100MB/min. It's there for reference only.

What I've done, however, is add a config file that logs everything in StratisD. It's 'Off' by default and will be saved in the right folder so that devs can just edit it and be on their way.

noescape00:
>I don't really want to enable this by default, it logs 100MB/min. It's there for reference only.

Ok, makes sense. 
If it would log only on debug level and we had a good trace\debug separation so just a few MB/min is generated would you reconsider?

bokobza:
Yeah possibly. Another problem is that if we keep a cop of the file in Stratis.Bitcoin, Nlog.config is copied to every bin folder of every feature so it gets a bit messy. 
I think we should only copy the NLog.config file in the console apps.
what do you think?

Aprogiena:
definitely not needed everywhere

bokobza:
Right, I'm going to merge this like this. let's reconsider later whether to set the `minLevel` to `debug` by default. @noescape00

noescape00:
Ok, sounds good. 

Btw you didn't delete the config file from the original folder before merging so now we have it duplicated.

Issue #1278, opened 2018-04-06
Merkle calculation issues


noescape00:
There are 2 issues:

1) `BlockMerkleRootRule` has it's implementation of the merkle tree calculation which is a copy paste from the PowConsensusValidator. So there are 2 copies of the same functionality and the one that is not covered by tests is being used. 

2) `BlockMerkleRootRule` throws in case merkle tree was mutated (`ConsensusErrors.BadTransactionDuplicate.Throw();`). This exception is being handled by `ConsensusLoop.AcceptBlockAsync` and the block is marked as invalid. The problem here is that we can't ban such a block because a valid block with the same hash might exist so there is an attack where invalid block is created in order to force the network to ban a valid block. More info here: https://bitcointalk.org/index.php?topic=102395.0

So instead of banning the block we should just discard it and ban the peer that provided such a block.

Created with @Aprogiena

dangershony:
1. Very good point the tests in `MerkeRootComputationTest` should point to the `BlockMerkleRootRule` rule instead of the methods in PowConsensusValidator, how did get slip by me I am not sure.

2. So we probably need to create a new consensus error type for this behaviour.

Very good @noescape00

noescape00:
>So we probably need to create a new consensus error type for this behaviour.

We can do it like that (create a type of errors that doesn't lead to banning the block) or do it the same way as `InvalidPrevTip` or `BadWitnessNonceSize` errors are handled in the consensus loop. Those 2 have custom handling and do not ban the block or peer so probably we can add there `BadTransactionDuplicate` exception and just ban the peer but avoid banning the block.

Aprogiena:
just add that if for now

dangershony:
>just add that if for now
What?

noescape00:
Also there is one more issue with mutated blocks that was discovered just today: if the block is below the last checkpoint we check the hash but don't check if it's mutated so we will download and save a mutated block.

I suggest we don't allow validation skip for the BlockMerkleRootRule.

PR #1353, opened 2018-04-28
Ban peers sending blocks with duplicate transactions

monsieurleberre:
trying to solve the issues raised in
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1278

I think some of the previous unit tests were using a whitelisted peer where it should have been a normal one, making it less relevant, so I changed it, please let me know if that is a mistake

there is still the issue mentioned at the end by @noescape00 for blocks below the last checkpoint


src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -1,12 +1,16 @@
-﻿using System.IO;
+﻿using System;
+using System.Collections.Generic;
```
noescape00: Following usings are redundant:

```
using System.IO;
using Stratis.Bitcoin.Features.Consensus.Rules.CommonRules;
using Stratis.Bitcoin.Configuration;
using FluentAssertions;
```

src/Stratis.Bitcoin.Features.Consensus.Tests/TestChainFactory.cs
```diff
@@ -120,34 +125,70 @@ internal class TestChainFactory
...
-                Transaction txCoinbase = newBlock.Block.Transactions[0];
-                txCoinbase.Inputs[0] = TxIn.CreateCoinbase(nHeight);
-                newBlock.Block.UpdateMerkleRoot();
+        private static async Task<BlockTemplate> MineBlockAsync(TestChainContext testChainContext, Script scriptPubKey,
+            MempoolSchedulerLock mempoolLock, TxMempool mempool, bool getMutatedBlock = false)
+        {
+            BlockTemplate newBlock = CreateBlockTemplate(testChainContext, scriptPubKey, mempoolLock, mempool);
 
-                var maxTries = int.MaxValue;
+            if(getMutatedBlock) newBlock.Block.Transactions.Add(newBlock.Block.Transactions[0]);
```
noescape00: You are duplicating a coinbase tx. 

There can be only one coinbase and coinstake tx so the block fails several consensus rules, not just mutated merkle tree one.

In order to create a mutated block that exploits CVE-2012-2459 you need to create a fully valid block with a few normal tx and duplicate 2 of them, after that update merkle root and check if it's the same as before the duplication.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -395,6 +395,14 @@ public void Stop()
                         return;
                     }
 
+                    if (blockValidationContext.Error == ConsensusErrors.BadTransactionDuplicate)
+                    {
+                        int banDuration = blockValidationContext.BanDurationSeconds == BlockValidationContext.BanDurationDefaultBan ? this.connectionManager.ConnectionSettings.BanTimeSeconds : blockValidationContext.BanDurationSeconds;
+                        this.peerBanning.BanPeer(blockValidationContext.Peer, banDuration, $"Invalid block received: {blockValidationContext.Error.Message}");
```
noescape00: peer can be null here

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -26,24 +26,22 @@ public class BlockMerkleRootRule : ConsensusRule
...
-                Block block = context.BlockValidationContext.Block;
+            if (!context.CheckMerkleRoot) return Task.CompletedTask;
 
-                uint256 hashMerkleRoot2 = BlockMerkleRoot(block, out bool mutated);
-                if (block.Header.HashMerkleRoot != hashMerkleRoot2)
-                {
-                    this.Logger.LogTrace("(-)[BAD_MERKLE_ROOT]");
-                    ConsensusErrors.BadMerkleRoot.Throw();
-                }
+            var block = context.BlockValidationContext.Block;
```
noescape00: unvar please

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +68,58 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
...
                 int level;
 
-                // For each of the lower bits in count that are 0, do 1 step. Each
-                // corresponds to an inner value that existed before processing the
+                // For each of the lower bits in processedLeavesCount that are 0, do 1 step. Each
+                // corresponds to an subTreeHash value that existed before processing the
                 // current leaf, and each needs a hash to combine it.
-                for (level = 0; (count & (((uint)1) << level)) == 0; level++)
+                for ( level = 0; (processedLeavesCount & (((uint)1) << level)) == 0; level++)
+                //for (level = 0; (processedLeavesCount % ((2^level) -1)) == 0; level++)
```
noescape00: Please remove commented out code

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +68,58 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
...
-                    Buffer.BlockCopy(h.ToBytes(), 0, hash, 32, 32);
-                    h = Hashes.Hash256(hash);
+                    Buffer.BlockCopy(subTreeHashes[level].ToBytes(), 0, hash, 0, 32);
+                    Buffer.BlockCopy(currentLeaveHash.ToBytes(), 0, hash, 32, 32);
+                    currentLeaveHash = Hashes.Hash256(hash);
                 }
 
-                // Store the resulting hash at inner position level.
-                inner[level] = h;
+                // Store the resulting hash at subTreeHashes  position level.
```
noescape00: Please replace 2 spaces with one after the `subTreeHashes `

Aprogiena:
After little more thinking, I realised that it is not that simple and if we want to move rule, any rule, we need to think much more about the consequences. Good rule of thumb here would be to keep the rules in the same order as bitcoin core has it.

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
monsieurleberre: OK I will get that cleaned

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
monsieurleberre: I left it here on purpose because I thought it was really not obvious what this loop was doing...

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
monsieurleberre: oh nooooo, I thought var was allowed when the type was obvious :(

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
monsieurleberre: thanks!

src/Stratis.Bitcoin.Features.Consensus.Tests/TestChainFactory.cs
monsieurleberre: That is probably true, I think I will try to do exactly that, but for the purpose of these tests I think they are already good enough to prove the code works as expected 
image
The tx-duplicate is the first error appearing in the block validation and the code goes to the correct path during the test. I will try to build a more realistic block today, but if we don't decide that rules can be run in any order, I think this is already a proof no?

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -397,9 +397,12 @@ public void Stop()
 
                     if (blockValidationContext.Error == ConsensusErrors.BadTransactionDuplicate)
                     {
-                        int banDuration = blockValidationContext.BanDurationSeconds == BlockValidationContext.BanDurationDefaultBan ? this.connectionManager.ConnectionSettings.BanTimeSeconds : blockValidationContext.BanDurationSeconds;
-                        this.peerBanning.BanPeer(blockValidationContext.Peer, banDuration, $"Invalid block received: {blockValidationContext.Error.Message}");
                         this.logger.LogTrace("(-)[BAD_TX_DUP]");
+                        if (blockValidationContext.Peer == null) return;
```
noescape00: I think it will probably be a good idea to have a smaller check on line 419 in order to avoid duplicated code. 

So instead of that block you can do on line 419:
```
if (blockValidationContext.Error == ConsensusErrors.BadTransactionDuplicate)
{
 //log     
 return;
}
```

just before we ban the block.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
noescape00: It's a bit different. 
If the type is on the right- use var. If not- explicit type.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
Aprogiena: or just add 
`if (blockValidationContext.Error != ConsensusErrors.BadTransactionDuplicate)`
there and make no return (the method ends normally)

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
noescape00: Oh, ok. 

So probably it will be a good idea to replace the commented out code with a text comment so it's obvious that this line is an explanation, not an alternative implementation.

Aprogiena: if something is not clear, make a normal comment instead

Aprogiena: but there is comment already and it seems that the rewrite of `(processedLeavesCount % ((2^level) -1)) == 0` does not make much sense, we are doing an action for every unset bit inside of `count` up until the first set bit, while modulo operation is logically something else

this is more like
count is `101101000`, then you execute the loop 3 times (as count ands with 3 zero bits)

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
monsieurleberre: OK Not sure what is the agreed solution reading that now : last time I discussed with @noescape00 we wanted to put that logic here to not have to do this bit
image
So maybe to avoid code duplication I can simply methodise that logic and reuse the method twice ?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
monsieurleberre: haha so I definitely don't get that way of writing "%" but it feels like it is equivalent though. I just find it easier to imagine the loop when I read my comment than when I see operations on bits

Aprogiena: just try to think what happens in your for:
```
for (level = 0; (processedLeavesCount % ((2^level) -1)) == 0; level++)
```

so imagine level is 0, then you have `(pLC % ((2 ^ 0) -1 ))` which is `pLC % (1 - 1)`, which means you are dividing by zero

src/Stratis.Bitcoin.Features.Consensus.Tests/TestChainFactory.cs
dangershony: You may be right but rules are such an important part of the code I would suggest to try to create proper invalid merkle trx.

dangershony:
>Good rule of thumb here would be to keep the rules in the same order as bitcoin core has it.

Yes unless we have a good reason and have properly discussed moving a rule.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -395,6 +395,17 @@ public void Stop()
                         return;
                     }
 
+                    if (blockValidationContext.Error == ConsensusErrors.BadTransactionDuplicate)
+                    {
+                        this.logger.LogTrace("(-)[BAD_TX_DUP]");
```
dangershony: There is already logging inside the rule no?

monsieurleberre: Yes, true, I have now removed that

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
Aprogiena: it is not clearer, it is wrong, pls remove it

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -395,16 +397,22 @@ public void Stop()
...
+    
+                    if (blockValidationContext.Peer != null && blockValidationContext.CanBanPeers)
                     {
                         int banDuration = blockValidationContext.BanDurationSeconds == BlockValidationContext.BanDurationDefaultBan ? this.connectionManager.ConnectionSettings.BanTimeSeconds : blockValidationContext.BanDurationSeconds;
                         this.peerBanning.BanPeer(blockValidationContext.Peer, banDuration, $"Invalid block received: {blockValidationContext.Error.Message}");
                     }
 
+                    if (blockValidationContext.Error == ConsensusErrors.BadTransactionDuplicate)
+                    {
+                        return;
```
noescape00: Can you please add exit log here?

monsieurleberre: Can I ask you why ?
I honestly think there is already way too much nesting in this method, and I don't want to do more unless it is necessary.

Aprogiena: Please read https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/logging-style.md. You created exit point from the function without proper logging it. The most elegant solution here is to invert the condition and then just follow through to the function end, which contains the log. Alternative solution would be as you have it but with correct logging and that would also be okay.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +73,60 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
...
-                // For each of the lower bits in count that are 0, do 1 step. Each
-                // corresponds to an inner value that existed before processing the
+                // For each of the lower bits in processedLeavesCount that are 0, do 1 step. Each
+                // corresponds to an subTreeHash value that existed before processing the
                 // current leaf, and each needs a hash to combine it.
-                for (level = 0; (count & (((uint)1) << level)) == 0; level++)
+                // The following loop could also be read like this:
+                // for (level = 0; (processedLeavesCount % ((2^level) -1)) == 0; level++)
+                // but bitwise operations are faster    
+                for ( level = 0; (processedLeavesCount & (((uint)1) << level)) == 0; level++)
```
Aprogiena: No space After (

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +73,60 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
...
 
-                // For each of the lower bits in count that are 0, do 1 step. Each
-                // corresponds to an inner value that existed before processing the
+                // For each of the lower bits in processedLeavesCount that are 0, do 1 step. Each
+                // corresponds to an subTreeHash value that existed before processing the
                 // current leaf, and each needs a hash to combine it.
-                for (level = 0; (count & (((uint)1) << level)) == 0; level++)
+                // The following loop could also be read like this:
+                // for (level = 0; (processedLeavesCount % ((2^level) -1)) == 0; level++)
+                // but bitwise operations are faster
```
Aprogiena: This is wrong I thought I explained that please remove it

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -139,16 +139,19 @@ public static uint256 ComputeMerkleRoot(List<uint256> leaves, out bool mutated)
                 // Level is the level (counted from the bottom) up to which we've sweeped.
                 int level = 0;
 
-                // As long as bit number level in count is zero, skip it. It means there
+                // As long as bit number level in processedLeavesCount is zero, skip it. It means there
                 // is nothing left at this level.
-                while ((count & (((uint)1) << level)) == 0)
+                // the following while can also be read like this
+                // while ((processedLeavesCount % ((2^level) -1)) == 0)
+                // but bitwise operations are faster.
```
Aprogiena: This is wrong, please remove it

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -399,12 +399,19 @@ public void Stop()
                     this.Chain.SetTip(this.Tip);
                     this.logger.LogTrace("Chain reverted back to block '{0}'.", this.Tip);
 
-                    if ((blockValidationContext.Peer != null) && (blockValidationContext.BanDurationSeconds != BlockValidationContext.BanDurationNoBan))
+                    bool peerShouldGetBanned =
```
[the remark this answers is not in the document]
monsieurleberre: there was a free line here so I decided to use it :)

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
dangershony: Ok I would like a proper explanation on this logic @monsieurleberre please understand exactly the low level with apro and then describe it in a comment inline (inside the method).

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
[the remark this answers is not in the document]
Aprogiena: Please don't

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
Aprogiena: The original comment seems to be very much sufficient

Aprogiena: @dangershony For more context please see discussion starting here https://github.com/stratisproject/StratisBitcoinFullNode/pull/1353#discussion_r184953228

monsieurleberre: OK I removed it as it is not strictly equivalent you are right, it still helped me a lot to understand the loop though to get a rough understanding of this condition / loop

Aprogiena: Thank you. But I think you're still misunderstanding. It is not that it is not strictly equivalent, it is completely different. So if what you're saying is true and it helped you understand, then you understand it wrongly.

Take a look
```
 for (level = 0; (count & (((uint)1) << level)) == 0; level++)
```

```
for (level = 0; (processedLeavesCount % ((2^level) -1)) == 0; level++) 
```

So let's say our `count` is `1011000 (bin) = 0x58 = 88 (dec)`, what do we have? 

1st iteration: `level = 0`
original: `(0x58 & (1 << 0)) == 0` is TRUE -> loop is taken
your code: `(88 % (1 - 1)) == 0` crashes

2nd iteration: `level = 1`
original: `(0x58 & (1 << 1)) == 0` is TRUE -> loop is taken
your code: `(88 % (2 - 1)) == 0` is TRUE (by accident because modulo 1 is always 0 for everything, so it is not possible for your code to return FALSE in second iteration)

3nd iteration: `level = 2`
original: `(0x58 & (1 << 2)) == 0` is TRUE -> loop is taken
your code: `(88 % (4 - 1)) == 0` is FALSE -> loop is NOT taken and would end

4nd iteration: `level = 3`
original: `(0x58 & (1 << 3)) == 0` is TRUE -> loop NOT taken and ends
your code: `(88 % (8 - 1)) == 0` is FALSE -> loop is NOT taken and would end


now if you try number `1000010`, your second iteration would do something else than original and you can literally find number that behaves differently on every level

there is NO relation between your code and original, it is completely wrong

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +73,57 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
         /// <param name="mutated"><c>true</c> if at least one leaf of the merkle tree has the same hash as any subtree. Otherwise: <c>false</c>.</param>
         public static uint256 ComputeMerkleRoot(List<uint256> leaves, out bool mutated)
         {
-            var branch = new List<uint256>();
-
             mutated = false;
-            if (leaves.Count == 0)
-                return uint256.Zero;
+            if (leaves.Count == 0) return uint256.Zero;
```
Aprogiena: I was telling @noescape the same thing. Try to avoid changes in code style for which we don't have rule. This leads to subjective changes because you like this way of writing the code more than what was there before, but if for example @noescape was about to edit this code later, he prefer the other way and he would change it back. So we would up in an infinite loop of changes. So the golden rule is, if there is no rule in the official code style policy, leave it as is.

But you don't need to fix it back here, it is just for the future.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -70,63 +73,57 @@ public static uint256 BlockMerkleRoot(Block block, out bool mutated)
...
+                processedLeavesCount++;
                 int level;
 
-                // For each of the lower bits in count that are 0, do 1 step. Each
-                // corresponds to an inner value that existed before processing the
+                // For each of the lower bits in processedLeavesCount that are 0, do 1 step. Each
+                // corresponds to an subTreeHash value that existed before processing the
                 // current leaf, and each needs a hash to combine it.
-                for (level = 0; (count & (((uint)1) << level)) == 0; level++)
+                for ( level = 0; (processedLeavesCount & (((uint)1) << level)) == 0; level++)
```
Aprogiena: Please remove the space after the 1st (

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
monsieurleberre: OK, thanks, I probably did not even do it on purpose... I'll keep that in mind

dangershony:
@monsieurleberre please fix the conflict

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
```diff
@@ -18,31 +18,34 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
...
+                ConsensusErrors.BadMerkleRoot.Throw();
+            }
 
-                uint256 hashMerkleRoot2 = BlockMerkleRoot(block, out bool mutated);
-                if (block.Header.HashMerkleRoot != hashMerkleRoot2)
-                {
-                    this.Logger.LogTrace("(-)[BAD_MERKLE_ROOT]");
-                    ConsensusErrors.BadMerkleRoot.Throw();
-                }
+            if (!mutated) return Task.CompletedTask;
```
dangershony: I don't agree with this change much
Its much simpler to read
```
                if (mutated)	
                {	
                    this.Logger.LogTrace("(-)[BAD_TX_DUP]");	
                    ConsensusErrors.BadTransactionDuplicate.Throw();	
                }
```
But no big deal, feels like changing code for the sake of changing try to avoid that.

src/Stratis.Bitcoin.Tests.Common/Transaction.cs
```diff
@@ -0,0 +1,20 @@
+﻿using FluentAssertions;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Tests.Common
+{
+    public class Transactions
+    {
+        public static Transaction BuildNewTransactionFromExistingTransaction(Transaction inputTransaction, int index = 0)
+        {
+            var transaction = new Transaction();
```
dangershony: There is a property on transaction called Clone is that not better?

dangershony:
Ah @Aprogiena if you are happy then push after the conflict is merged.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BlockMerkleRootRule.cs
Aprogiena: +1

Aprogiena:
I agree with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1353#discussion_r185796429

otherwise it is OK, but wait for @noescape00

noescape00:
>otherwise it is OK, but wait for @noescape00

I believe @monsieurleberre is still working on the integration test that will imitate a real life exploit for that bu so I'll wait for that one.

monsieurleberre:
OK Dan let me quickly change according to your comments and resolve the conflict if I can.
I was thinking of adding the integration test in a separate PR, it is currently showing no banning and chain progress when syncing from the malicious peer. I hope that my test is not realistic and that you guys can help me :)

src/Stratis.Bitcoin.Tests.Common/Transaction.cs
monsieurleberre: I think this one takes an existing transaction to use as inputs, not really cloning it...

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -16,124 +15,200 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests
...
-            block.Header.HashPrevBlock = context.Chain.Tip.HashBlock;
-            await context.Consensus.AcceptBlockAsync(new BlockValidationContext { Block = block, Peer = peer });
-
-            Assert.True(context.PeerBanning.IsBanned(peer));
+            minerScriptPubKey = new Key().ScriptPubKey;
         }
 
-        [Fact]
-        public async Task NodeIsSynced_PeerSendsABadBlockAndPeerIsConnected_ThePeerGetsBanned_Async()
+        private async Task<(TestChainContext context, IPEndPoint peerEndPoint)> InitialiseContextAndPeerEndpoint()
```
noescape00: Please add `Async` suffic

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -16,124 +15,200 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests
...
-            var peer = new Mock<INetworkPeer>();
-            peer.Setup(p => p.Behavior<ConnectionManagerBehavior>()).Returns(connectionManagerBehavior);
+            return (context, peerEndPoint);
+        }
 
-            context.MockReadOnlyNodesCollection.Setup(s => s.FindByEndpoint(It.IsAny<IPEndPoint>())).Returns(peer.Object);
+        [Fact]
+        public async Task NodeIsSynced_PeerSendsABlockWithBadPrevHashAndPeerDisconnected_ThePeerGetsBanned_Async()
+        {
+            await NodeIsSynced_PeerSendsABadBlockAndPeerDisconnected_ThePeerGetsBanned_Async(
```
noescape00: `.this` is missing here and at 52, 72, 79, 109, 116, 135,142, 166,173

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -16,124 +15,200 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests
...
+
+            await context.Consensus.AcceptBlockAsync(blockValidationContext);
 
             // wait 1 sec for ban to expire.
             Thread.Sleep(1000);
 
             Assert.False(context.PeerBanning.IsBanned(peerEndPoint));
         }
 
+        private static async Task<Block> Mine2BlocksAndCreateABlockWithBadPrevHash(TestChainContext context)
```
noescape00: `Async` suffix is missing

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -16,124 +15,200 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests
...
+        {
+            var blocks = await TestChainFactory.MineBlocksAsync(context, 2, minerScriptPubKey);
+            
+            var block = blocks.First();
+            block.Header.HashPrevBlock = context.Chain.Tip.HashBlock;
+            return block;
+        }
+
+        private static async Task<Block> MineAMutatedBlock(TestChainContext context)
```
noescape00: `Async` suffix is missing

PR #1358, opened 2018-04-30
Add Dashboard to API. Add a test feature showing how new features wil…

StratisDeveloper:
…l implement dashboard.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
+++ b/src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
@@ -1,0 +1,34 @@
+﻿using System.Net;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Features.Dashboard.Controllers
+{
+    /// <summary>
+    /// Controller providing HTML Dashboard
+    /// </summary>
+    [Route("dashboards")] //declares the endpoint
+    public class DashboardController : Controller
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DashboardController"/> class.
+        /// </summary>
+        public DashboardController()
+        {
+        }
+
+        /// <summary>
+        /// Returns a web page to act as a dashboard
+        /// </summary>
+        /// <returns>text/html content</returns>
+        [HttpGet]
+        [Route("nodedashboard")] // the endpoint name
+		public IActionResult NodeDashboard()
+        {
+            return Content("Hello World", "text/html");
+        }
+
+
+    }
+}
--- a/src/Stratis.Bitcoin.Features.Dashboard/Dashboard.cs
+++ b/src/Stratis.Bitcoin.Features.Dashboard/Dashboard.cs
@@ -1,0 +1,50 @@
+﻿using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.DependencyInjection;
+using Stratis.Bitcoin.Builder;
+using Stratis.Bitcoin.Builder.Feature;
+using Stratis.Bitcoin.Features.Api;
+using Stratis.Bitcoin.Features.Dashboard.Controllers; 
+
+namespace Stratis.Bitcoin.Features.Dashboard
+{
+    /// <summary>
+    /// Dashboard Feature
+    /// </summary>
+	public class Dashboard : FullNodeFeature 
+    {
+
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger; 
+
+		public Dashboard(ILoggerFactory loggerFactory)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName); 
+        }
+
+        public override void Initialize()
+        {
+			this.logger.LogInformation("Dashboard initialized");
+        }
+    }
+
+    /// <summary>
+    /// A class providing extension methods for <see cref="IFullNodeBuilder"/>.
+    /// </summary>
+	public static class FullNodeBuilderDashboardExtension
+    {
+		public static IFullNodeBuilder UseDashboard(this IFullNodeBuilder fullNodeBuilder)
+        {
+            fullNodeBuilder.ConfigureFeature(features =>
+            {
+                features
+					.AddFeature<Dashboard>()
+                    .DependOn<ApiFeature>() // API endpoints are going to be required to display data and make method calls from the dashboard
+                    .FeatureServices(services =>
+                    {
+					services.AddSingleton<DashboardController>();
+                    });
+            });
+            return fullNodeBuilder;
+        }
+    }
+}
--- a/src/Stratis.Bitcoin.Features.Dashboard/Stratis.Bitcoin.Features.Dashboard.csproj
+++ b/src/Stratis.Bitcoin.Features.Dashboard/Stratis.Bitcoin.Features.Dashboard.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="Controllers\" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\NBitcoin\NBitcoin.csproj" />
+    <ProjectReference Include="..\Stratis.Bitcoin\Stratis.Bitcoin.csproj" />
+    <ProjectReference Include="..\Stratis.Bitcoin.Features.Api\Stratis.Bitcoin.Features.Api.csproj" />
+  </ItemGroup>
+</Project>
--- a/src/Stratis.Bitcoin.FullNode.sln
+++ b/src/Stratis.Bitcoin.FullNode.sln
@@ -91,7 +91,13 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.Bitcoin.Tests.Walle
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.Bitcoin.Tests.Common", "Stratis.Bitcoin.Tests.Common\Stratis.Bitcoin.Tests.Common.csproj", "{2B3B9594-5074-48CE-BFC5-DA02B56C16CA}"
 EndProject
+<<<<<<< Updated upstream
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.Bitcoin.Features.TestFeature", "Stratis.Bitcoin.Features.TestFeature\Stratis.Bitcoin.Features.TestFeature.csproj", "{9D13B526-920C-449A-B397-D0906C0FB467}"
+=======
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.Bitcoin.Features.Dashboard", "Stratis.Bitcoin.Features.Dashboard\Stratis.Bitcoin.Features.Dashboard.csproj", "{CBF35BCA-659B-4288-AED4-5EF4103CB2F4}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Stratis.StratisNodeWithHtmlUI", "Stratis.StratisNodeWithHtmlUI\Stratis.StratisNodeWithHtmlUI.csproj", "{6C958708-F059-41F3-9BA6-493CE57969AD}"
+>>>>>>> Stashed changes
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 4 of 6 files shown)


src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Serves up the requested file from the embedded resources
+        /// </summary>
+        /// <returns>Web response with text content</returns>
+        [HttpGet]
+        [Route("static/{file_name}")]
+        public ActionResult Static(string file_name)
+        {
+            var assembly = Assembly.GetEntryAssembly();
```
StratisDeveloper: We need to jump through some hoops here to read the files we want to serve. The files we want to serve are in the Features.Api project, but the hosting of files depends on the application using the Feature.Api project.  Therefore we compile the static files as "embedded resources" and need to read them from the assembly to serve them.

src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,49 @@
...
+            assembly = Assembly.Load(new AssemblyName("Stratis.Bitcoin.Features.Api"));
+            var resourceStream = assembly.GetManifestResourceStream("Stratis.Bitcoin.Features.Api.dashboard_static_files." + file_name);
+            using (var reader = new StreamReader(resourceStream, Encoding.UTF8))
```
StratisDeveloper: Possible issue here with serving unicode?

src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                if (file_name.EndsWith(".html")) {
+                    return Content(reader.ReadToEnd(), "text/html");
```
StratisDeveloper: Not a great solution, but something else is forcing the return type to json in most cases, probably something to do with swagger

src/Stratis.Bitcoin.Features.Api/MvcBuilderExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static IMvcBuilder AddControllers(this IMvcBuilder builder, IServiceColle
                 builder.AddApplicationPart(controllerType.ServiceType.GetTypeInfo().Assembly);
             }
 
-            builder.AddApplicationPart(typeof(Controllers.NodeController).Assembly);
+            builder.AddApplicationPart(typeof(Stratis.Bitcoin.Controllers.NodeController).Assembly);
```
StratisDeveloper: needed to disambiguate from new Dashboard Controller

src/Stratis.Bitcoin.Features.Api/Stratis.Bitcoin.Features.Api.csproj
```diff
@@ -29,9 +29,21 @@
...
+
+  <ItemGroup>
+    <Folder Include="dashboard_static_files\" />
+    <Folder Include="Controllers\" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="dashboard_static_files\*.*" />
+  </ItemGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="dashboard_static_files\*.*" />
```
StratisDeveloper: This is where we mark the directory's contents as embedded resources. Probably won't work on subfolders?

src/Stratis.Bitcoin.Features.Api/dashboard_static_files/index.html
```diff
@@ -0,0 +1,42 @@
...
+                $('#content').show();
+            }
+        });
+        var $container = $("#content");
+        load_url = "../../node/status"
+        $container.load(load_url);
+        var refreshId = setInterval(function()
+        {
+            $container.load(load_url);
+        }, 10000);
```
StratisDeveloper: Refresh every 10 seconds

src/Stratis.Bitcoin.Features.Api/dashboard_static_files/index.html
```diff
@@ -0,0 +1,42 @@
...
+                $('#loading').hide();
+                $('#content').show();
+            },
+            success: function() {
+                $('#loading').hide();
+                $('#content').show();
+            }
+        });
+        var $container = $("#content");
+        load_url = "../../node/status"
```
StratisDeveloper: Get json from /node/status api endpoint

src/Stratis.Bitcoin.Features.Api/dashboard_static_files/index.html
```diff
@@ -0,0 +1,42 @@
+﻿<!DOCTYPE HTML>
+<head>
+    <title>Stratis Dashboard</title>
+    <script type="text/javascript" src="//ajax.googleapis.com/ajax/libs/jquery/1.8.0/jquery.min.js"></script>
```
StratisDeveloper: Simple example of how html/javascript static files can render a dashboard.

src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
StratisDeveloper: Alternate approaches: 

Using MVC views instead of serving static files. The problem with this is that some dashboard builders may not want to use mvc, and instead use angular or something similar. So, perhaps make both options available - an MVC dashboard and a static pages dashboard.

Serving Static files using .UseStaticFiles() - This would work easily if the files weren't in separate assemblies from StratisD. Instead, to use this way, you have to still embed the files, but then you have to override the file provider used by the component that serves the static files. I was unable to get this working.

dangershony:
First comment I would have is I'd like to break this PR in two, lets start with the `DashboardController` presented in the API feature, if you can remove the TestFeature for and we focus on the `DashboardController`

I noticed you serve any static files form the dashboard, I would suggest no to take such a generic approach right now and just serve a dedicated html file.

StratisDeveloper:
Can you review the DashboardController part and simply ignore any file that includes the word "TestFeature" ?

ferdeen:
Hi,  

Guessing this is to do with the dashboard bounty ? 

Have you considered the ApiExplorer in ASP.NET Core - or is this not the same thing ? 

[ApiExplorer in ASP.NET Core](https://andrewlock.net/introduction-to-the-apiexplorer-in-asp-net-core/)

Cheers,

Ferdeen

StratisDeveloper:
@ferdeen The API Explorer seems roughly equivalent to the swagger API.

What we are seeking to make is a presentation layer for certain features, as per this spec https://github.com/StratisDevelopmentFoundation/StratisDevelopmentFoundation.github.io/issues/9

So for example, instead of using the FullNodeUI, you could use an html layer on top of the Wallet feature to accomplish the same tasks. The upside of this is that someone who has no idea about electron UIs could use their web developing skills to make a wallet, or to have the node display certain information, etc.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -124,6 +125,7 @@ public static IFullNodeBuilder UseApi(this IFullNodeBuilder fullNodeBuilder, Act
                         services.AddSingleton(fullNodeBuilder);
                         services.AddSingleton(options);
                         services.AddSingleton<ApiSettings>(new ApiSettings(setup));
+                        services.AddSingleton<DashboardController>();
```
bokobza: I don't think this controller should be included in this feature as we don't want to have the Dashboard enabled by default.
ApiFeature is the feature enabling APIs, and I think the dashboard should be a Feature in itself.

src/Stratis.Bitcoin.Features.TestFeature/Controllers/TestFeatureController.cs
```diff
@@ -0,0 +1,60 @@
...
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Features.TestFeature.Controllers
+{
+
+    /// <summary>
+    /// Controller providing TestFeature operation
+    /// </summary>
+    [Route("api/[controller]")]
+    public class TestFeatureController : Controller
```
bokobza: `TestFeatureController` sounds like it's a feature that tests something. I'm being picky but maybe `DummyFeatureController` or `ExampleFeatureController` is more self-explicit.

src/Stratis.Bitcoin.Features.TestFeature/TestFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// A class providing extension methods for <see cref="IFullNodeBuilder"/>.
+    /// </summary>
+    public static class FullNodeBuilderTestFeatureExtension
+    {
+        public static IFullNodeBuilder UseTestFeature(this IFullNodeBuilder fullNodeBuilder)
+        {
+            fullNodeBuilder.ConfigureFeature(features =>
+            {
+                features
+                    .AddFeature<TestFeature>()
```
bokobza: If the dashboard controller moves to its own feature (which I think it should), you would it to add it here.

src/Stratis.StratisD/Program.cs
```diff
@@ -39,6 +40,7 @@ public static void Main(string[] args)
                     .UseBlockStore()
                     .UseMempool()
                     .UseWallet()
+                    .UseTestFeature()
```
bokobza: You can create a new console app showing off this new feature. It's a bit early to add it to StratisD. :stuck_out_tongue_winking_eye:

bokobza:
I haven't reviewed the serving of files yet, but I have some first comments.

src/Stratis.Bitcoin.Features.Api/dashboard_static_files/index.html
sondreb: How come jquery 1.8.0 is being used and not the latest? There are continuously found various exploits in older versions of jquery that are fixed, and 1.8.0 includes the 11974 issue reported here, which has a lot of XSS issues: https://bugs.jquery.com/ticket/11974

mikeolteanu:
For context, this pull request was meant as a way to get early feedback on how to structure the dashboard. Specifically, with regards to how static files are served, and where the code to serve them should be placed within the structure.

Unfortunately, the pull request usually means I'm asking for the code to be included in the Stratis code base. However, in this case I'm only seeking a code review.

StratisDeveloper:
For context, this pull request was meant as a way to get early feedback on how to structure the dashboard. Specifically, with regards to how static files are served, and where the code to serve them should be placed within the structure.

Unfortunately, the pull request usually means I'm asking for the code to be included in the Stratis code base. However, in this case I'm only seeking a code review.

src/Stratis.Bitcoin.Features.TestFeature/Controllers/TestFeatureController.cs
StratisDeveloper: TestFeature is included as a way to demonstrate how new feature specific dashboards might look in the code, for purposes of this code review. Do you think there should be an ExampleFeature that lives in the stratis project as an example for feature developers? I think it probably doesn't belong there.

bokobza: This comment is just about naming. :-)
I find the `Test` prefix confusing. Maybe a prefix such as `Dummy` or `Example` will be more clear.

src/Stratis.Bitcoin.Features.Api/dashboard_static_files/index.html
dangershony: @StratisDeveloper yes please upgrade and it should be as simple as updating the nuget

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
dangershony: Yes thats true, and it will be a feature that depends on the api feature.

src/Stratis.Bitcoin.Features.TestFeature/Controllers/TestFeatureController.cs
dangershony: I think this should just be called DashboardController 

So I think we should have a new DashboardFeature that depends on ApiFeature
And the DashboardFeature will have internally a DashboardController that will serve some basic node info.

src/Stratis.StratisD/Program.cs
dangershony: Yes perhaps call it StratisUiD

dangershony:
I am fine with asking for review, but if we break it to may small PR then it can already go in the main repo.

So I would suggest as first PR (either fix this one or create a new one) with the following components.
- A new Feature called DashboardFeature that has a dependency on ApiFeature
- A DashboardController that just return a "Hello Stratis World" in string format. (I would suggest trying to use razor but the most simple solution is fine with me)
- A new daemon (perhaps called StratisUiD) that starts the node with the dashboard feature.

Then the next PR we can think on how to display the UI and after that on the feature menu.

src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Serves up the requested file from the embedded resources
+        /// </summary>
+        /// <returns>Web response with text content</returns>
+        [HttpGet]
+        [Route("static/{file_name}")]
+        public ActionResult Static(string file_name)
+        {
+            var assembly = Assembly.GetEntryAssembly();
+            assembly = Assembly.Load(new AssemblyName("Stratis.Bitcoin.Features.Api"));
```
majikandy: Could this be the assembly of a type you know will be in that assembly rather than loading by string assembly name?  If not, then perhaps a constant for the assembly name

src/Stratis.Bitcoin.Features.Api/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,49 @@
...
+            var resourceStream = assembly.GetManifestResourceStream("Stratis.Bitcoin.Features.Api.dashboard_static_files." + file_name);
```
majikandy: perhaps a constant for the prefix and shouldn't file_name be fileName?  or is it an accepted practice to use snake case for route params?

StratisDeveloper:
Thanks for all the feedback, closing this, see #1406 for first part of dashboard.

PR #1366, opened 2018-05-01
Fix mining loop issue and BDD integration test folder creation

fassadlr:


src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -146,25 +139,51 @@ public INetworkPeer CreateNetworkPeerClient()
...
+            this.runner.Start(this.DataFolder);
+
+            this.State = CoreNodeState.Starting;
+
+            if (this.runner is BitcoinCoreRunner)
+                StartBitcoinCoreRunner();
+            else
+                StartStratisRunner();
+
+            this.State = CoreNodeState.Running;
```
bokobza: Looks like you're setting this inside StartBitcoinCoreRunner as well.

src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -53,12 +54,11 @@ public CoreNode(string folder, INodeRunner runner, NodeBuilder builder, Network
             this.runner = runner;
             this.Folder = folder;
             this.State = CoreNodeState.Stopped;
-            if (cleanfolders)
-                this.CleanFolder();
 
-            Directory.CreateDirectory(folder);
             this.DataFolder = Path.Combine(folder, "data");
```
bokobza: not sure why the DataFolder is even set manually. We should create a node using -datadir and Datafolder should be set automatically.

bokobza:
I'm finding this hard to review as well. Methods have changed AND moved place so I'm ending up coparing 2 methods thinking they're the same, when in fact they're not.
Multiple commits can make this a 5 minutes job instead of a much longer and uncertain one.

PR #1373, opened 2018-05-02
[Integration Tests] Refactor node runners.

fassadlr:


src/Stratis.Bitcoin.IntegrationTests/EnvironmentMockUpHelpers/NodeRunners.cs
```diff
@@ -47,46 +47,45 @@ public void Start(string dataDir)
         }
     }
 
-    public class StratisBitcoinPosRunner : INodeRunner
+    public abstract class StratisRunner : INodeRunner
```
bokobza: Does StratisRunner mean that it's running one of Stratis' networks?
If not then maybe calling this `NodeRunner` is better?

fassadlr: OK I renamed to NodeRunner. I'll look at removing INodeRunner in next commit ;)

bokobza:
I like.

PR #1386, opened 2018-05-04
Block puller bug fix

noescape00:
This is a quick fix for a bug that lead to consensus stopping any advancement. 

If the total score of all peers is too low we were stopping the quality adjustments. This means that in some cases we were able to ask a peer for a block it doesn't even have and we will never reassign the download job to any other peer because we won't reduce the quality score of that peer on stalling. 

Please note that this is a temporary quick fix and in the nearest future whole BlockPuller will be redesigned.


Aprogiena:
I think this is a good attempt for a quick fix of those 2 problems we have seen with advancing.

Aprogiena:
It should be noted that this change will pretty much disable disconnecting off peers during IBD, so it is really just a temporary solution until the big refactoring

src/Stratis.Bitcoin/BlockPulling/QualityScore.cs
```diff
@@ -143,46 +143,6 @@ public double CalculateQualityAdjustment(long blockDownloadTimeMs, int blockSize
             if (timePerKb < 2 * avgTimePerKb) res = avgTimePerKb / timePerKb;
             else if (Math.Abs(avgTimePerKb) >= 0.00001) res = -timePerKb / (2 * avgTimePerKb);
 
-            if ((res < 0) && this.IsPenaltyDiscarded())
```
dangershony: should you keep this code?
```
        if (res < 0)	
                res = 0;
```

PR #1389, opened 2018-05-04
[Integration Tests] Fix TestSegwit test and remove callbacks from test runners.

fassadlr:


src/Stratis.Bitcoin.IntegrationTests/BlockStore/RetrieveFromBlockStoreSteps.cs
```diff
@@ -48,7 +48,7 @@ public RetrieveFromBlockStoreSpecification(ITestOutputHelper output) : base(outp
 
         protected override void BeforeTest()
         {
-            this.builder = NodeBuilder.Create(caller: this.CurrentTest.DisplayName);
+            this.builder = NodeBuilder.Create(testFolder: this.CurrentTest.DisplayName);
```
bokobza: I'm happier with `caller` as a param name.

fassadlr: OK fixed

Issue #1338, opened 2018-04-25
Stop staking bug


dangershony:
Got this bug when calling the api method `StopStaking`
![image](https://user-images.githubusercontent.com/7487930/39252689-1cbbf6d8-489e-11e8-8eb2-5b7a06040b7a.png)

I wonder if its related to a double stake output
![image](https://user-images.githubusercontent.com/7487930/39252733-3642d0d6-489e-11e8-9d34-f99cb6232949.png)

mikedennis:
PR 1392 does not fix this as I was unable to figure out how to reproduce this. In examining the code the dispose on the staking should have waited on the staking task to complete before setting the cancellation token to null so not sure how it got into the above state where it was executing a staking task with a null cancellation token.

mwherman2000:
Mine is also reproducible: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1430

PR #1392, opened 2018-05-06
Fix some staking issues

### LINKED ISSUE #1357: [POS Mining] incorrect logic due to missing return
there should be return after this line https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs#L383

mikedennis:
This should fix #1357 and #1338

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -447,7 +447,7 @@ namespace Stratis.Bitcoin.Features.Miner
             }
 
             this.stakeCancellationTokenSource?.Cancel();  
-            this.logger.LogTrace("Disposing of staking loop.");
+            this.logger.LogTrace("Disposing staking loop.");
             this.stakingLoop?.Dispose();
             this.stakingLoop = null;
             this.stakeCancellationTokenSource?.Dispose();
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -386,6 +386,7 @@ namespace Stratis.Bitcoin.Features.Miner
             if (Interlocked.CompareExchange(ref this.stakeProgressFlag, StakeInProgress, StakeNotInProgress) == StakeInProgress)
             {
                 this.logger.LogTrace("(-)[ALREADY_MINING]");
+                return;
             }
 
             this.rpcGetStakingInfoModel.Enabled = true;
@@ -463,7 +464,7 @@ namespace Stratis.Bitcoin.Features.Miner
 
             BlockTemplate blockTemplate = null;
 
-            while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
+            while (this.stakeCancellationTokenSource != null && !this.stakeCancellationTokenSource.Token.IsCancellationRequested)
             {
                 // Prevent mining if the system time is not in sync with that of other members on the network.
                 if (this.timeSyncBehaviorState.IsSystemTimeOutOfSync)
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -446,17 +446,10 @@ namespace Stratis.Bitcoin.Features.Miner
                 return;
             }
 
-            this.stakeCancellationTokenSource.Cancel();
-            this.logger.LogTrace("Waiting for staking loop to cancel...");
-            this.stakingLoop?.RunningTask?.GetAwaiter().OnCompleted(() =>
-            {
-                this.logger.LogTrace("Staking loop cancelled.");                              
-            });
+            this.stakeCancellationTokenSource.Cancel();  
             this.logger.LogTrace("Disposing of staking loop.");
             this.stakingLoop?.Dispose();
-            this.stakingLoop = null;
             this.stakeCancellationTokenSource?.Dispose();
-            this.stakeCancellationTokenSource = null;
             this.rpcGetStakingInfoModel = new Miner.Models.GetStakingInfoModel();
             this.rpcGetStakingInfoModel.Enabled = false;
 
@@ -470,7 +463,7 @@ namespace Stratis.Bitcoin.Features.Miner
 
             BlockTemplate blockTemplate = null;
 
-            while (this.stakeCancellationTokenSource != null && !this.stakeCancellationTokenSource.Token.IsCancellationRequested)
+            while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
             {
                 // Prevent mining if the system time is not in sync with that of other members on the network.
                 if (this.timeSyncBehaviorState.IsSystemTimeOutOfSync)
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -447,9 +447,15 @@ namespace Stratis.Bitcoin.Features.Miner
             }
 
             this.stakeCancellationTokenSource.Cancel();
-            this.stakingLoop.Dispose();
+            this.logger.LogTrace("Waiting for staking loop to cancel...");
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -463,7 +464,7 @@ public void StopStake()
 
             BlockTemplate blockTemplate = null;
 
-            while (!this.stakeCancellationTokenSource.Token.IsCancellationRequested)
+            while (this.stakeCancellationTokenSource != null && !this.stakeCancellationTokenSource.Token.IsCancellationRequested)
```
Aprogiena: This is not correct. The correct solution is to wait for this method to complete before setting the cancellation source to null.

mikedennis: Yes, that would be cleaner - working on it...

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -446,9 +447,15 @@ public void StopStake()
             }
 
             this.stakeCancellationTokenSource.Cancel();
-            this.stakingLoop.Dispose();
+            this.logger.LogTrace("Waiting for staking loop to cancel...");
+            this.stakingLoop?.RunningTask?.GetAwaiter().OnCompleted(() =>
+            {
+                this.logger.LogTrace("Staking loop cancelled.");                              
+            });
```
Aprogiena: We do not really need these 5 lines

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -446,9 +447,15 @@ public void StopStake()
...
+            this.logger.LogTrace("Disposing of staking loop.");
+            this.stakingLoop?.Dispose();
```
Aprogiena: These 2 are sufficient (together with the cancellation token source questionmark)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: So now you can remove it

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -446,17 +446,10 @@ public void StopStake()
                 return;
             }
 
-            this.stakeCancellationTokenSource.Cancel();
-            this.logger.LogTrace("Waiting for staking loop to cancel...");
-            this.stakingLoop?.RunningTask?.GetAwaiter().OnCompleted(() =>
-            {
-                this.logger.LogTrace("Staking loop cancelled.");                              
-            });
+            this.stakeCancellationTokenSource.Cancel();
```
Aprogiena: although if it is possible on line 452 that sCTS is null, it is also possible here
so is it possible or not?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -446,17 +446,10 @@ public void StopStake()
...
-            this.stakingLoop?.RunningTask?.GetAwaiter().OnCompleted(() =>
-            {
-                this.logger.LogTrace("Staking loop cancelled.");                              
-            });
+            this.stakeCancellationTokenSource.Cancel();  
             this.logger.LogTrace("Disposing of staking loop.");
             this.stakingLoop?.Dispose();
-            this.stakingLoop = null;
             this.stakeCancellationTokenSource?.Dispose();
-            this.stakeCancellationTokenSource = null;
```
Aprogiena: why removing assignment to null? whoever created the code probably wanted it to be "cleaned" probably because it will actually cause that exception in case something is forgotton (otherwise it would just work improperly silently) - i'd say keep the logic, just fix the bug

mikedennis: I get the feeling my dense brain is missing something obvious 😁 I'll do some reading up on CTS as haven't really worked much with it before.

Aprogiena: i'm aiming here for just minimal changes to the code, don't overthink it :-)

mikedennis: After digging more into this, I'm not sure why the task is still running.  Calling Dispose() on the staking loop should have waited on the task so it should be completed.
```
       /// <summary>
        /// Wait for the loop task to complete.
        /// </summary>
        public void Dispose()
        {
            if (!this.RunningTask.IsCanceled)
            {
                this.logger.LogInformation("Waiting for {0} to finish.", this.Name);
                this.RunningTask.Wait();
            }
        }
```

mikedennis: TBH my suggestion is we get rid of the disposal of the cancellationtokensource altogether from this code and let the garbage collector take care of it. This guy suggests that is the best way unless you are using the waithandle.  https://stackoverflow.com/questions/29779842/is-code-that-disposes-cancellationtokensource-while-tasks-are-canceling-correct

Aprogiena: As a matter of good practice, when something is disposable, we dispose it regardless of its internal implementation. Otherwise we would need to argue why we are not disposing it which is quite hard and could lead to errors.

I have checked the code and found out the design here is wrong and there is a race condition of the staking status, which is being controlled by interlocked operations. But let's discard this finding for this PR.

To finish with this PR, let's add questionmark after the cancellation token source in this code in stop staking and restore the line that assigned null to it.

If you want to dig deeper, let's make a Skype call.In that case please contact me on slack.

mikedennis:
Ok sounds good - made the changes requested. Still not sure I have an explanation for why the issue is happening to start with but maybe the race condition you found has something to do with it.

Aprogiena:
The race conditions should be unrelated I hope, but I'm happy with the PR as it is now. Although we should create a new issue for the race condition and rework the design here.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -445,10 +446,11 @@ public void StopStake()
                 return;
             }
 
-            this.stakeCancellationTokenSource.Cancel();
-            this.stakingLoop.Dispose();
+            this.stakeCancellationTokenSource?.Cancel();  
+            this.logger.LogTrace("Disposing of staking loop.");
```
noescape00: Probably having "Disposing staking loop." will be more correct

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -386,6 +386,7 @@ public void Stake(WalletSecret walletSecret)
             if (Interlocked.CompareExchange(ref this.stakeProgressFlag, StakeInProgress, StakeNotInProgress) == StakeInProgress)
```
majikandy: Since this block of code is being changed, perhaps this if condition could be given more meaning?   My guess (from not knowing what this code does) but merely from the log message that it means   if (this.StakingInProgress())

Aprogiena: This is a good comment, but it should be part of https://github.com/stratisproject/StratisBitcoinFullNode/issues/1397

It does not really belong here. Needless to say that `StakingInProgress` would not be a good name. So we need to think about a good name, but 1st we need to fix the race condition, which may actually cause this code to be removed or 
rewritten significantly.

majikandy: >but 1st we need to fix the race condition, which may actually cause this code to be removed or
rewritten significantly.

Sounds good - but can I actually put this comment on #1397 ?

Aprogiena: Absolutely, it is most relevant there

Aprogiena:
This has been approved already, but after last commit we have Travis CI failure. That's not good. @bokobza @fassadlr @dangershony can you please help here? It does not seem that it is related to the actual commit. Are we doing green builds or not yet (I thought we are, if we are not, ignore this message I guess)?

PR #1398, opened 2018-05-08
Create code review policy

dangershony:


Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
...
+changes to the code, even small changes as incrementing an integer, can result in critical consensus bugs and lose of funds if not reviewed properly.  
+Some PRs are more severe then others in importance (for example changes to tests may not require the same level of checks as changes to the consensus code).  
+For that reason we have a PR review policy with 3 tags, A PR must be associated with a pr review policy.
+
+#### Requirements
+
+The requirement for each pr policy
+
+*pr-review-high*
+- Code must be review by at least two core devs
```
Neurosploit: reviewed

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
+### Code Review Policy
+
+#### How to code review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from types to suggestions in design.
```
Neurosploit: typo's?

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
...
+#### How to code review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from types to suggestions in design.   
+
+#### Why are code reviews important?  
+It's essential that PRs are properly code-reviewed,   
+changes to the code, even small changes as incrementing an integer, can result in critical consensus bugs and lose of funds if not reviewed properly.  
+Some PRs are more severe then others in importance (for example changes to tests may not require the same level of checks as changes to the consensus code).  
+For that reason we have a PR review policy with 3 tags, A PR must be associated with a pr review policy.
```
Neurosploit: if we capitalize PR do it for all mentions

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Requirements
+
+The requirement for each pr policy
+
+*pr-review-high*
+- Code must be review by at least two core devs
+- All tests must pass (unit test and Integration tests)
```
Neurosploit: capitalize unit or write integration in lowercase

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
...
+- All 4 chains need to have been synced (to a certain height)
+
+
+*pr-review-medium*
+- Code must be review by at least two core devs
+- All tests must pass (unit test and Integration tests) 
+
+*pr-review-low*
+- Code must be review by at one core devs
+- unit test must pass
```
Neurosploit: Unit tests

dangershony:
Hmm apro you are very strict. 
I think your medium suggestion is too high of a requirement, medium is enough to run tests no need to sync chains, if you think chains need to be synced just propose to upgrade to high alternatively lets add another options `pr-review-critical` - similar to your `pr-review-high`.

I agree perhaps author needs to suggest a level, but we need to provide guidance on when the levels need to be set.

- code changes to consensus/wallet/store - high (critical)
- code changes rpc/api/config/connection/protocol - medium/high
- code changes to tests - medium/low

Aprogiena:
p2p (i.e. protocol) should be rarely something else than high
you still can do trivial changes everywhere, but most of the time, p2p is just as critical as consensus
while store seems to be just a store ... not so critical from my point of view

mining, block assembly - also critical

Aprogiena:
and yes, i'm strict as i want to be rather on the safe side

majikandy:
>All tests must pass (Unit-test and Integration-tests) (and have to be run by the author of the code, and BOTH of the authorative reviewers)

I think we should rely on the build for both of these.  

Including them as though then individuals reviewing the code has to run them is a mistake and sets a bad precedent that without putting these that we might not.

Obviously I cannot argue it is a bad idea for reviewers to run them -  by all means do so.... but We have continuous build servers for a reason, and at the very least we should have a mechanism to kick off  integration tests on a remote machine / build server so the reviewer can do it that way if we aren't doing it continuously.

However, it is 2018, we should be doing this continuously. No exceptions.

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
+### Code Review Policy
+
+#### How to code review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from typos to suggestions in design.   
+
+#### Why are code reviews important?  
+It's essential that PRs are properly code-reviewed,   
+changes to the code, even small changes as incrementing an integer, can result in critical consensus bugs and lose of funds if not reviewed properly.
```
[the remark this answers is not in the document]
majikandy: That example would fail a test, so I think that is bad example and represents that we have a dangerous codebase where you can increment an integer and lose funds.

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,30 @@
...
+For that reason we have a PR review policy with 3 tags, A PR must be associated with a PR review policy.
+
+#### Requirements
+
+The requirement for each PR policy
+
+*pr-review-high*
+- Code must be reviewed by at least two core devs
+- All tests must pass (Unit-test and Integration-tests)
+- All 4 chains need to have been synced (to a certain height)
```
majikandy: Can we be explicit what these 4 chains are please.

Aprogiena:
@majikandy You are completely right, the problem is that ever since I on this project, we haven't been able to make integration tests reliable enough that they could be put on automation. These requirements for developers and the reviewers to run integration tests manually is necessary only because of that and it would not exist if we did have the perfect setup with working automation. The ultimate goal should be as you say and we should make every effort to get there. But until we are there, we still need to care about the quality assurance manually. The same holds for the syncing with chains – optimally, in the future, we will have a dedicated test environment that will just this for us and will only out that in case of the problem. This will lead to a massive savings of everyone's time as well as a great boost in confidence in the quality of our products.

Documentation/codereview-policy.md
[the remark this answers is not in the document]
Aprogiena: Again, in a perfect world yes, but we don't have hundred percent test coverage, so in practice it might not fail a test.

bokobza:
Maybe we can have a few words on what to look for in code reviews?
something like that:

Look out for:
* **Correctness in business logic**: check whether the change makes sense from a pure business logic perspective.
* **Changes in interface/architectural design**: a mistake here may lead to something that previously worked may not work anymore.
* **Code robustness**: check whether the new code is solid and defensive, i.e checks for nullity.
* **Tests**: make sure tests have been added to support the code change. If not, figure out why.
* **Code styling**: make sure our [code style policy](https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/coding-style.md) is respected.
* **Typos and grammar**: make sure the code reads smoothly and is in British English. 
* **Issue reference**: check whether the PR references an issue and make sure both are linked.
* etc...

Documentation/codereview-policy.md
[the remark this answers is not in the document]
dangershony: I think this is good that devs feel its dangerous to make changes, they will be extra careful.
And its very hard to cover all aspects to the code.

dangershony:
@Aprogiena I am gonna leave low as it is (only unit tests) as this is for cases where typos and really small changes can be used. this is covered by the automated CI.

I will add another option and call it critical as you suggested.

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+For that reason we have a PR review policy with 3 tags. A PR must be associated with a PR review policy.
+
+#### Requirements
+
+The requirement for each PR policy
+
+*pr-review-critical*
+- Code must be reviewed by at least two core devs
+- All tests must pass (Unit-test and Integration-tests) (run by auther and the two reviewers)
+- All 4 chains (Bitcoin main/test, Stratis main/test) need to have been synced (to an agreed height) (run by auther and 2 reviewers)
```
Aprogiena: `either`

And we should probably specify the heights exactly.

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+
+*pr-review-high*
+- Code must be reviewed by at least two core devs
+- All tests must pass (Unit-test and Integration-tests) (run by auther and the two reviewers)
```
Aprogiena: `either`

Documentation/codereview-policy.md
dangershony: I was thinking that could be discussed in the review itself

Aprogiena: It would just distract both the coder and the reviewer, so I think it's better to just define it once in the policy.

noescape00: >It would just distract both the coder and the reviewer, so I think it's better to just define it once in the policy.

I agree. 

I think for stratis main and test syncing till the tip is required. 
For BTC test and main- hard to tell- potentially we want to sync until the last checkpoint or the last BIP activation but that still might take a very long time.

dangershony: My take is If you find this to be an issue we can revisit

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### How to review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from typos to suggestions in design.   
+
+#### Why are code reviews important?  
+It's essential that PRs are properly code-reviewed.   
+Changes to the code, even small changes as incrementing an integer, can result in critical consensus bugs and loss of funds if not reviewed properly.  
+Some PRs are more severe than others in importance (for example changes to tests may not require the same level of checks as changes to the consensus code).
```
majikandy: I don’t like this part about devaluing the importance of tests. I would suggest comparing something else with the importance of consensus code. Ie pick another area to compare as being less important, eg wallet button repositioning

Aprogiena: Is not devaluing the importance of tests. The tests are important, but no one will ever lose money just because the test is broken in a way that it gives false-positive result. And if it is broken in a way that it gives false-negative then it will just prevent PR being merged. So basically you cannot make a mistake in a test that will immediately break something for the users. So obviously the test code is less critical.

Documentation/codereview-policy.md
bokobza: I think he meant `author`.

Documentation/codereview-policy.md
majikandy: What I meant was....   Perhaps this would be better...

Some PRs are more severe than others in importance (for example changes to consensus code need a high level of checks).

I don't see any need to compare prod code to tests.

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
+### Code Review Policy
+
+#### How to review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from typos to suggestions in design.   
+
+#### Why are code reviews important?  
+It's essential that PRs are properly code-reviewed.
```
majikandy: This line isn't an answer to the question above and could be removed merged into the title... eg
`#### Why it is essential that PRs are properly code-reviewed?`

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
+### Code Review Policy
+
+#### How to review pull requests  
+
+This is a guide for core developers and contributors who want to help with code reviewing PRs on the fullnode.  
+We welcome anyone to help with code reviews, from typos to suggestions in design.   
+
+#### Why are code reviews important?  
+It's essential that PRs are properly code-reviewed.   
+Changes to the code, even small changes as incrementing an integer, can result in critical consensus bugs and loss of funds if not reviewed properly.
```
majikandy: This could be emphasised even more by saying `if not tested properly` instead of `if not reviewed and tested properly`

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+- Code must be reviewed by at least one core dev
+- Unit-test must pass (automated CI tests)
+
+
+#### What to look for in code reviews
+
+* **Correctness in business logic**: check whether the change makes sense from a pure business logic perspective.
+* **Changes in interface/architectural design**: a mistake here may lead to something that previously worked may not work anymore.
+* **Code robustness**: check whether the new code is solid and defensive, i.e checks for nullity.
+* **Tests**: make sure tests have been added to support the code change. If not, figure out why.
```
majikandy: Should this say    `ask why`  rather than `figure out why` ?

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+- All tests must pass (Unit-test and Integration-tests) (run by author and the two reviewers)
+- All 4 chains (Bitcoin main/test, Stratis main/test) need to have been synced (to an agreed height) (run by author and 2 reviewers)
+
+*pr-review-high*
+- Code must be reviewed by at least two core devs
+- All tests must pass (Unit-test and Integration-tests) (run by author and the two reviewers)
+- All 4 chains (Bitcoin main/test, Stratis main/test) need to have been synced (to an agreed height)
+
+*pr-review-medium*
+- Code must be reviewed by at least two core dev (run by author and reviewr)
```
majikandy: typo -  `reviewr` to `reviewer`

majikandy: `dev` -> `developers`?

Documentation/codereview-policy.md
```diff
@@ -0,0 +1,45 @@
...
+- All tests must pass (Unit-test and Integration-tests) 
+
+*pr-review-low*
+- Code must be reviewed by at least one core dev
+- Unit-test must pass (automated CI tests)
+
```
majikandy: missing full stops on the above 2

majikandy: ooops - ignore that, I was comparing to the block at the bottom that all have full stops

noescape00:
Update: I think for the pr-review-low we should add that the creator of the PR also have to run integration tests to avoid situations like this: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1469 (simple PR with pr-review-low caused node to crash at startup)

noescape00:
@fassadlr @bokobza @ferdeen @dangershony @Aprogiena 
this PR is not merged for a month already. 
Do we agreed on this or something has to be updated?

dangershony:
Yes we agree I will push it.

Issue #1320, opened 2018-04-23
[RF] Extend chained header 


dangershony:
Extend chained header with pointer to the block data and information about block state.

noescape00:
There will be 2 new properties that will be contained in the ChainedHeader. They are only in-memory and are not persisted to the hard drive:

1. Pointer to the Block instance that this header is representing. Can be Null in case we don't have a block data downloaded. 
2. BlockState (a new enum should be introduced). 

BlockState enum: 

- HeaderOnly - we only have a header of the block and the header is valid (after we introduce ProvenHeader this state means that we have valid proven header).
- BlockDataRequired - state that means that we are interested in downloading a block that is being represented by that header. This happens when header is a part of the chain that can potentially replace our consensus tip because its tip's total chainwork is greater comparing to what we currently have.
- BlockDataAvailable- we have block data for the header.
- PartiallyValidated
- FullyValidated
-

PR #1401, opened 2018-05-09
Extending ChainedHeader

dangershony:
Fix: #1320

### DIFF
--- a/src/NBitcoin/ChainedHeader.cs
+++ b/src/NBitcoin/ChainedHeader.cs
@@ -5,6 +5,60 @@ using NBitcoin.BouncyCastle.Math;
 namespace NBitcoin
 {
     /// <summary>
+    /// Represent the avilability state of a block in regards to it's header.
+    /// </summary>
+    public enum BlockDataAvailabilityState
+    {
+        /// <summary>
+        /// A BlockHeader is found and is a valid header.
+        /// </summary>
+        HeaderOnly,
+
+        /// <summary>
+        /// We are interested in downloading a block that is being represented by that header. 
+        /// This happens when a header is a part of the chain that can potentially replace our consensus tip 
+        /// because its tip's total chainwork is greater comparing to what we currently have.
+        /// </summary>
+        BlockDataRequired,
+
+        /// <summary>
+        /// Block is downloaded, if the Block property is null the block needs to be asked from store.
+        /// </summary>
+        BlockDataAvailable
+    }
+
+    /// <summary>
+    /// Represent the validation level of a block.
+    /// </summary>
+    public enum ValidationState
+    {
+        /// <summary>
+        /// No validation was performed.
+        /// </summary>
+        Unknown,
+
+        /// <summary>
+        /// We have a valid block header.
+        /// </summary>
+        HeaderValidated,
+
+        /// <summary>
+        /// No validation required because the block is assumed to be valid.
+        /// </summary>
+        AssumedValid,
+
+        /// <summary>
+        /// Validated using all rules that don't require db state changes.
+        /// </summary>
+        PartiallyValidated,
+
+        /// <summary>
+        /// Validated using all the rules.
+        /// </summary>
+        FullyValidated
+    }
+
+    /// <summary>
     /// A BlockHeader chained with all its ancestors.
     /// </summary>
     public class ChainedHeader
@@ -36,6 +90,14 @@ namespace NBitcoin
         /// <summary>Total amount of work in the chain up to and including this block.</summary>
         public uint256 ChainWork { get { return Target.ToUInt256(this.chainWork); } }
 
+        public BlockDataAvailabilityState BlockDataAvailability { get; set; }
+
+        public ValidationState BlockValidationState { get; set; }
+
+        public Block Block { get; set; }
+
+        public List<ChainedHeader> Next { get; private set; }
+
         /// <summary>
         /// Constructs a chained block.
         /// </summary>
@@ -51,6 +113,7 @@ namespace NBitcoin
                 this.Height = previous.Height + 1;
 
             this.Previous = previous;
+            this.Next = new List<ChainedHeader>();
 
             if (previous == null)
             {
@@ -80,6 +143,7 @@ namespace NBitcoin
             this.Header = header ?? throw new ArgumentNullException("header");
             this.Height = height;
             this.HashBlock = headerHash;
+            this.Next = new List<ChainedHeader>();
             this.CalculateChainWork();
         }
 
@@ -413,7 +477,7 @@ namespace NBitcoin
         /// <returns>Whether proof of work is valid.</returns>
         public bool CheckProofOfWorkAndTarget(Consensus consensus)
         {
-            return (this.Height == 0) || (this.Header.CheckProofOfWork(consensus) && this.Header.Bits == GetWorkRequired(consensus));
+            return (this.Height == 0) || (this.Header.CheckProofOfWork(consensus) && this.Header.Bits == this.GetWorkRequired(consensus));
         }
 
         /// <summary>


src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
 
 namespace NBitcoin
 {
+    /// <summary>
+    /// Represent the avilability state of a block in regards to it's header.
+    /// </summary>
+    public enum BlockDataAvailabilityState
+    {
+        /// <summary>
+        /// A BlockHeader is found and is a valid header.
```
bokobza: Either use `<see cref="BlockHeader"/>` or `block header` in the documentation.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// </summary>
+        HeaderOnly,
+
+        /// <summary>
+        /// We are interested in downloading a block that is being represented by that header. 
+        /// This happens when a header is a part of the chain that can potentially replace our consensus tip 
+        /// because its tip's total chainwork is greater comparing to what we currently have.
+        /// </summary>
+        BlockDataRequired,
+
```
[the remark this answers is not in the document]
bokobza: I think this should be `BlockRequired`.
BlockData is BlockHeader + Block. Here we want to download the Block (we have the BlockHeader already), so BlockRequired is probably better suited.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
 
 namespace NBitcoin
 {
+    /// <summary>
+    /// Represent the avilability state of a block in regards to it's header.
```
bokobza: *availability
*its

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// We are interested in downloading a block that is being represented by that header. 
+        /// This happens when a header is a part of the chain that can potentially replace our consensus tip 
+        /// because its tip's total chainwork is greater comparing to what we currently have.
+        /// </summary>
+        BlockDataRequired,
+
+        /// <summary>
+        /// Block is downloaded, if the Block property is null the block needs to be asked from store.
+        /// </summary>
+        BlockDataAvailable
```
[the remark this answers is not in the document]
bokobza: Or `BlockAvailable`.

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
bokobza: I'm thinking "BlockData is BlockHeader + Block" because we have the HeaderOnly enum member present so BlockData is not only about a Block.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+    }
+
+    /// <summary>
+    /// Represent the validation level of a block.
```
bokobza: *Represents

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// Validated using all rules that don't require db state changes.
+        /// </summary>
+        PartiallyValidated,
+
+        /// <summary>
+        /// Validated using all the rules.
+        /// </summary>
+        FullyValidated
+    }
+
```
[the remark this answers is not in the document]
bokobza: I prefer BlockDataValidationState with:
Unknown,
HeaderOnly (is that what it means?),
Assumed,
Partial,
Full

src/NBitcoin/ChainedHeader.cs
```diff
@@ -36,6 +90,14 @@ public class ChainedHeader
         /// <summary>Total amount of work in the chain up to and including this block.</summary>
         public uint256 ChainWork { get { return Target.ToUInt256(this.chainWork); } }
 
+        public BlockDataAvailabilityState BlockDataAvailability { get; set; }
```
bokobza: Please add some documentation about what these are about.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -80,6 +143,7 @@ public ChainedHeader(BlockHeader header, uint256 headerHash, int height)
             this.Header = header ?? throw new ArgumentNullException("header");
             this.Height = height;
             this.HashBlock = headerHash;
+            this.Next = new List<ChainedHeader>();
```
noescape00: I guess it will be a good optimization to initialize a list with capacity of 1 since in most of the cases ChainedHeader will have a single Next pointer. 
Same suggestion for the line 116

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
noescape00: >HeaderOnly (is that what it means?)

It means we have a valid header (or proven header) but we don't have block data for it.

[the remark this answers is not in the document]
noescape00: I wouldn't change the AssumedValid to Assumed since AssumeValid is a well know term which suddenly reminds of Checkpoints and AssumeValid hash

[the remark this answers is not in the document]
dangershony: I dont think I would change any of the enums they all feel rightly named.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -80,6 +143,7 @@ public ChainedHeader(BlockHeader header, uint256 headerHash, int height)
...
             this.CalculateChainWork();
         }
```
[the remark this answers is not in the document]
dangershony: hmm very minor but ok I will add another const

src/NBitcoin/ChainedHeader.cs
dangershony: There is an open discussion about not needing to document every property as it will be repetitive.
I can copy past the comment from the enum itself but feels useless.

Aprogiena: @dangershony I think we can achieve an agreement in that if something is used in many components in the same way, we don't need that comment to be repeated.

This case is completely different and not documenting the property that is unique inside of the class would be a big mistake. This is not about repeating what each state value means, that information does not belong here.

Aprogiena: Okay, I understand, in which case I would suggest we introduce `<inheritdoc/>` here, or `<inheritdocfromtype/>`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
 
 namespace NBitcoin
 {
+    /// <summary>
+    /// Represents the state of the availability of a block.
+    /// </summary>
+    public enum BlockDataAvailabilityState
+    {
+        /// <summary>
+        /// A <see cref="BlockHeader"/> is found and is a valid header.
```
Aprogiena: I would not say "found". I would also not say that it is a valid header, because these states are about availability of the data. This means that this particular state represents having data of the header.

Aprogiena: I would also say that this is an initial value and that is because before we have full data of the header, we are unable to construct chained header.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+
+        /// <summary>
+        /// We are interested in downloading the <see cref="Block"/> that is being represented by the current <see cref="BlockHeader"/>. 
+        /// This happens when a header is a part of the chain that can potentially replace our consensus tip 
+        /// because its tip's total chain work is greater comparing to what we currently have.
+        /// </summary>
+        BlockRequired,
+
+        /// <summary>
+        /// The <see cref="Block"/> is downloaded, if the Block property is null the block needs to be asked from store.
```
Aprogiena: You don't really have a block property here. So actually this part of the comment should be on the actual property inside of chained header, not here. By itself this value means that the block data has been downloaded and is still reachable. It is possible that a block was downloaded by light wallet and it was analysed and it is no longer needed and therefore the block availability state will be changed to header only. But here you should not mention anything about the store of the block data.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// </summary>
+        Unknown,
+
+        /// <summary>
+        /// We have a valid block header.
+        /// </summary>
+        HeaderValidated,
+
+        /// <summary>
+        /// No validation required because the block is assumed to be valid.
```
Aprogiena: It is rather minimum validation, not no validation.I would maybe add an information that this state is used for blocks before the last checkpoint or for blocks that are on the chain of assume valid block.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// </summary>
+        AssumedValid,
+
+        /// <summary>
+        /// Validated using all rules that don't require db state changes.
```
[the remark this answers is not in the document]
Aprogiena: However, for a block that was represented using headers with assume valid state, this state does not mean that it was validated using all the rules, but that the rules were skipped because of the assumption of validity.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,60 @@
...
+        /// </summary>
+        PartiallyValidated,
+
+        /// <summary>
+        /// Validated using all the rules.
```
Aprogiena: Same problem with assume valid here – this state actually represents that we treat this block as if it was validated using all the rules, although in reality it might not be validated because the rules were skipped.

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
Aprogiena: I don't mind either variant.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -413,7 +492,7 @@ public bool CheckProofOfWorkAndTarget(Network network)
         /// <returns>Whether proof of work is valid.</returns>
         public bool CheckProofOfWorkAndTarget(Consensus consensus)
         {
-            return (this.Height == 0) || (this.Header.CheckProofOfWork(consensus) && this.Header.Bits == GetWorkRequired(consensus));
+            return (this.Height == 0) || (this.Header.CheckProofOfWork(consensus) && this.Header.Bits == this.GetWorkRequired(consensus));
```
Aprogiena: As we are making this more beautiful, please add brackets.

src/NBitcoin/ChainedHeader.cs
Aprogiena: For a second I was thinking whether partially validated and fully validated should not be renamed to partially trusted and fully trusted because the node itself may have not actually validated the blocks. However, then I realised that the validation was actually done, just not by the node itself, but someone else – some other node or a developer – and the owner of the node declared that the validation was done.

So I'm happy with the naming, and I still believe we just need to describe the actual meaning in more detail and then we are fine.

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
dangershony: It wont necessarily be inputs though, I think its more about the state of the chain, partial validation will not change any state (for example outputs will not be marked as spent) this will only do validation, in case of smart contracts its not clear yet.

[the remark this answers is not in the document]
Aprogiena: connecting inputs (newly called UpdateCoins) means that we check that inputs are UTXOs (this can only be done after we rewind the chain to the fork point) and we spend them while creating new UTXOs from new outputs, that is the last step. Bitcoin core does a few extra checks in this phase - like sigop count and some BIP30/34 thingy, but more or less, it is just pure input connecting.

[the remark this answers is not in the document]
Aprogiena: Though there is one more thing I need to check. So wait for it.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
 
 namespace NBitcoin
 {
+    /// <summary>
+    /// Represents the state of the availability of a block.
```
Aprogiena: can you check with Justin on grammar correctness of this pls?

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+    public enum BlockDataAvailabilityState
+    {
+        /// <summary>
+        /// A <see cref="BlockHeader"/> is presented, the block data is not available.
+        /// </summary>
+        HeaderOnly,
+
+        /// <summary>
+        /// We are interested in downloading the <see cref="Block"/> that is being represented by the current <see cref="BlockHeader"/>. 
+        /// This happens when a header is a part of the chain that can potentially replace our consensus tip
```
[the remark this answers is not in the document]
Aprogiena: i can't of like "can potentially replace", but i see you are aiming for little more correct wording, i agree with that as well

therefore i'd try

`This happens when we don't have block which is represented by this header and the header is a part of a chain that can potentially replace our consensus tip because its chain work is greater than our consensus tip's chain work.`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        BlockRequired,
+
+        /// <summary>
+        /// The <see cref="Block"/> was downloaded and is available, but it may not be reachable directly but via a store.
+        /// </summary>
+        BlockAvailable
+    }
+
+    /// <summary>
+    /// Represents the validation level of a block.
```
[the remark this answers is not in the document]
Aprogiena: agree, status here is probably better, as it is not strictly a level, but rather more context is included

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+
+        /// <summary>
+        /// We have a valid block header.
+        /// </summary>
+        HeaderValidated,
+
+        /// <summary>
+        /// Minimum validation required because the block is assumed to be valid.
+        /// This state is used for blocks before the last checkpoint or for blocks that are on the chain of assume valid block.
+        /// Once a block is fully validated this state will change to FullyValidated.
```
Aprogiena: use `<see>`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        Unknown,
+
+        /// <summary>
+        /// We have a valid block header.
+        /// </summary>
+        HeaderValidated,
+
+        /// <summary>
+        /// Minimum validation required because the block is assumed to be valid.
+        /// This state is used for blocks before the last checkpoint or for blocks that are on the chain of assume valid block.
```
Aprogiena: => `... for block headers before the last checkpoint, which are on the correct chain, or for block headers that are ...`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        /// </summary>
+        Unknown,
+
+        /// <summary>
+        /// We have a valid block header.
+        /// </summary>
+        HeaderValidated,
+
+        /// <summary>
+        /// Minimum validation required because the block is assumed to be valid.
```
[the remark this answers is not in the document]
Aprogiena: yes => `Blocks represented by headers with this state are assumed to be valid and only minimal validation is required for them when they are downloaded.`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        AssumedValid,
+
+        /// <summary>
+        /// Validated using all rules that don't require change of state.
+        /// </summary>
+        PartiallyValidated,
+
+        /// <summary>
+        /// Validated using all the rules.
+        /// Some rules validation may be skipped for blocks marked as AssumedValid.
```
Aprogiena: `<see>`

src/NBitcoin/ChainedHeader.cs
```diff
@@ -36,17 +93,32 @@ public class ChainedHeader
         /// <summary>Total amount of work in the chain up to and including this block.</summary>
         public uint256 ChainWork { get { return Target.ToUInt256(this.chainWork); } }
 
+        /// <summary>
+        /// Represents the state of the availability of a block.
```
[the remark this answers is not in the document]
Aprogiena: or use that `inherittypedoc` so you don't repeat it (otherwise consider fixing on type as well)

src/NBitcoin/ChainedHeader.cs
```diff
@@ -36,17 +93,32 @@ public class ChainedHeader
...
+        /// </summary>
+        public BlockDataAvailabilityState BlockDataAvailability { get; set; }
+
+        /// <summary>
+        /// Represents the validation level of a block.
```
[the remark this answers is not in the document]
Aprogiena: or again use inherit

src/NBitcoin/ChainedHeader.cs
```diff
@@ -75,14 +147,24 @@ public ChainedHeader(BlockHeader header, uint256 headerHash, ChainedHeader previ
...
             this.Height = height;
-            this.HashBlock = headerHash;
             this.CalculateChainWork();
         }
 
+        /// <summary>
+        /// Constructs a chained header at the start of a chain.
+        /// </summary>
+        /// <param name="header">The header for the block.</param>
+        /// <param name="headerHash">The hash computed according to NetworkOptions.</param>
```
Aprogiena: use `<see>`

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
dangershony: inherittypedoc  wont work on a none override type

dangershony:
I will make some of the proposed fixes but I suggest we push this and have Justin review all the comments in this class at a later stage.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+
+        /// <summary>
+        /// Minimum validation required because the block is assumed to be valid.
+        /// This state is used for blocks before the last checkpoint or for blocks that are on the chain of assume valid block.
+        /// Once a block is fully validated this state will change to FullyValidated.
+        /// </summary>
+        AssumedValid,
+
+        /// <summary>
+        /// Validated using all rules that don't require change of state.
```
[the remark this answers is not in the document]
dangershony: No I am avoiding specific tech, coinview is a name of a class representing UTXO state.

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
dangershony: Ah I found this little baby
```
        /// <inheritdoc cref="ValidationState" />
        public ValidationState BlockValidationState { get; set; }
```

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
noescape00: I think in case when a component is well known and came from BTC core it's ok to use particular name such as a coinview. It is more understandable this way because `don't require change of state.` doesn't say much. Which state? What is a state and when can it be changed?

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
 
 namespace NBitcoin
 {
+    /// <summary>
+    /// Represents the availability state of a block.
+    /// </summary>
+    public enum BlockDataAvailabilityState
+    {
+        /// <summary>
+        /// A <see cref="BlockHeader"/> is presented, the block data is not available.
```
bokobza: Do you mean to say `is present`?

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        /// </summary>
+        HeaderOnly,
+
+        /// <summary>
+        /// We are interested in downloading the <see cref="Block"/> that is being represented by the current <see cref="BlockHeader"/>. 
+        /// This happens when we don't have block which is represented by this header and the header is a part of a chain that
+        /// can potentially replace our consensus tip because its chain work is greater than our consensus tip's chain work.
+        /// </summary>
+        BlockRequired,
```
bokobza: so here the block is not available but the header is. Is it not like HeaderOnly then?

bokobza:
Happy for you to proceed but
`and have Justin review all the comments in this class at a later stage.`
What do you mean?

noescape00:
I agree that we can leave grammar mistakes and beautification to Justin but most of it is not about grammar but about tech details which should be written only by those who write/review the code.

src/NBitcoin/ChainedHeader.cs
Aprogiena: the distinction is that HO we don't want to download (it is not interesting for us), while this one is
so yes, we still have only the header, but we feel differently about it, so the actual physical availability have not changed, but its context did

src/NBitcoin/ChainedHeader.cs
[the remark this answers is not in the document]
Aprogiena: i'd just add
`/// Some rules validation may be skipped for blocks marked as <see cref="AssumedValid"/>. `

and i'm ok with this

src/NBitcoin/ChainedHeader.cs
```diff
@@ -4,6 +4,63 @@
...
+        AssumedValid,
+
+        /// <summary>
+        /// Validated using all rules that don't require change of state.
+        /// </summary>
+        PartiallyValidated,
+
+        /// <summary>
+        /// Validated using all the rules.
+        /// Some rules validation may be skipped for blocks marked as <see cref="AssumedValid"/>.
```
Aprogiena: add previously 
-> ``/// Some rules validation may be skipped for blocks previously marked as <see cref="AssumedValid"/>. ``

dangershony:
@bokobza what I mean is fix grammatical errors but possibly also even elaborate on a component (talking to the developer if need be) the dev that is writing the comment may not realise its missing more explanations.

bokobza:
We can do grammar ourselves and ask Justin for hep if needs be.
The questions we had are not about grammar but about meaning.
Let's just push this in so you can continue with your work, but let's keep an eye on the names of these elements as well as the associated comments. 
I'm sure their obvious meaning will emerge and we can readjust. :metal:

bokobza:
Jeremy's approach to commenting.

"That will do"
Alt Text

dangershony:
This is a fairly simple change if the rest of our PRs are going to be like this we are in for a long winter.

PR #1411, opened 2018-05-15
add pid, blockStore height to NodeStatus

DanGould:
`GET:http://localhost:37221/api/Node/status` now responds:
```
{
  "agent": "StratisBitcoin",
  "version": "1.1.2.0",
  "network": "StratisMain",
  "pid": 8284,
  "consensusHeight": 195519,
  "blockStoreHeight": 195519,
  "inboundPeers": [],
  "outboundPeers": [
    {
      "version": "/Stratis:2.0.0.3/",
      "remoteSocketEndpoint": "[::ffff:113.167.183.194]:16178",
      "tipHeight": 538405
    }
  ],
  "enabledFeatures": [
    "Stratis.Bitcoin.Base.BaseFeature",
    "Stratis.Bitcoin.Features.Consensus.ConsensusFeature",
    "Stratis.Bitcoin.Features.BlockStore.BlockStoreFeature",
    "Stratis.Bitcoin.Features.MemoryPool.MempoolFeature",
    "Stratis.Bitcoin.Features.Wallet.WalletFeature",
    "Stratis.Bitcoin.Features.Miner.MiningFeature",
    "Stratis.Bitcoin.Features.Api.ApiFeature",
    "Stratis.Bitcoin.Features.RPC.RPCFeature"
  ],
  "dataDirectoryPath": "C:\\Users\\dcsle\\AppData\\Roaming\\StratisNode\\stratis\\StratisMain",
  "runningTime": "00:00:52.5545221"
}
```

Design decision to use provider courtesy of noescape0

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
@@ -186,6 +186,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                         services.AddSingleton<IBlockStoreCache, BlockStoreCache>();
                         services.AddSingleton<StoreBlockPuller>();
                         services.AddSingleton<BlockStoreLoop>();
+                        services.AddSingleton<IStoreStateProvider>(x => x.GetService<BlockStoreLoop>());
                         services.AddSingleton<BlockStoreManager>();
                         services.AddSingleton<BlockStoreSignaled>();
                         services.AddSingleton<StoreSettings>(new StoreSettings(setup));
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
@@ -16,7 +16,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
     /// <summary>
     /// The BlockStoreLoop simultaneously finds and downloads blocks and stores them in the BlockRepository.
     /// </summary>
-    public class BlockStoreLoop
+    public class BlockStoreLoop : IStoreStateProvider
     {
         /// <summary>Factory for creating background async loop tasks.</summary>
         private readonly IAsyncLoopFactory asyncLoopFactory;
@@ -85,7 +85,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         private readonly StoreSettings storeSettings;
 
         /// <summary>The highest stored block in the repository.</summary>
-        internal ChainedHeader StoreTip { get; private set; }
+        public ChainedHeader StoreTip { get; private set; }
 
         /// <summary>Public constructor for unit testing.</summary>
         public BlockStoreLoop(IAsyncLoopFactory asyncLoopFactory,
--- a/src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
+++ b/src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
@@ -27,6 +27,9 @@ namespace Stratis.Bitcoin.Controllers.Models
         /// <summary>The network the current node is running on.</summary>
         public string Network { get; set; }
 
+        /// <summary> The processId of the node.</summary>
+        public int PID { get; set; }
+
         /// <summary>The height of the consensus.</summary>
         public int ConsensusHeight { get; set; }
 
--- a/src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
+++ b/src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
@@ -33,6 +33,9 @@ namespace Stratis.Bitcoin.Controllers.Models
         /// <summary>The height of the consensus.</summary>
         public int ConsensusHeight { get; set; }
 
+        /// <summary> The height of the full blocks downloaded to BlockStore </summary>
+        public int BlockStoreHeight { get; set; }
+
         /// <summary>A collection of inbound peers.</summary>
         public List<ConnectedPeerModel> InboundPeers { get; set; }
 
--- a/src/Stratis.Bitcoin/Controllers/NodeController.cs
+++ b/src/Stratis.Bitcoin/Controllers/NodeController.cs
@@ -5,6 +5,7 @@ using Stratis.Bitcoin.Builder.Feature;
 using Stratis.Bitcoin.Configuration;
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Controllers.Models;
+using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.P2P.Peer;
 using Stratis.Bitcoin.Utilities;
 using System.Diagnostics;
@@ -32,7 +33,9 @@ namespace Stratis.Bitcoin.Controllers
         /// <summary>The connection manager.</summary>
         private readonly IConnectionManager connectionManager;
 
-        public NodeController(IFullNode fullNode, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider, IChainState chainState, NodeSettings nodeSettings, IConnectionManager connectionManager)
+        private readonly IStoreStateProvider storeStateProvider;
+
+        public NodeController(IFullNode fullNode, ILoggerFactory loggerFactory, IDateTimeProvider dateTimeProvider, IChainState chainState, IStoreStateProvider storeStateProvider, NodeSettings nodeSettings, IConnectionManager connectionManager)
         {
             this.fullNode = fullNode;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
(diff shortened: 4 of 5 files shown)


src/Stratis.Bitcoin/Interfaces/IStoreStateProvider.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Interfaces
+{
+    public interface IStoreStateProvider
```
noescape00: Add comment please

src/Stratis.Bitcoin/Controllers/NodeController.cs
```diff
@@ -53,8 +58,10 @@ public IActionResult Status()
             {
                 Version = this.fullNode.Version?.ToString() ?? "0",
                 Agent = this.nodeSettings.Agent,
+                PID = Process.GetCurrentProcess().Id,
```
noescape00: We tent to have meaningful names for variables so probably something pike `ProcessId` will be better here?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreLoop.cs
```diff
@@ -16,7 +16,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
     /// <summary>
     /// The BlockStoreLoop simultaneously finds and downloads blocks and stores them in the BlockRepository.
     /// </summary>
-    public class BlockStoreLoop
+    public class BlockStoreLoop : IStoreStateProvider
```
noescape00: Probably making an interface for the whole `BlockStoreLoop` would be a better idea but I'm fine with the quick approach since block store loop will be refactored soon anyway

bokobza:
I think it's more correct to use `IBlockRepository.HighestPersistedBlock` (like the console does) rather than the BlockStoreLoop's `StoreTip`.

Now, counting on the BlockStore to be present is a bit problematic, because it won't always be there. Look in BreezeD's Program.cs file.

I think here the best thing to do at the moment is to look at fullNode.Services to see if you have a type of BlockRepository, and if you do, call `IBlockRepository.HighestPersistedBlock` using reflection. 

It's not the most elegant, but maybe in the future we can have the Features publish their own status. So you'd do 
```
foreach (IFullNodeFeature feature in this.fullNode.Services.Features)
{
    model.FeatureDetails.Add(feature.Name, feature.GetStatusDetails());
    model.EnabledFeatures.Add(feature.GetType().ToString());
}
```
where 
1. feature.GetStatusDetails() returns a property bag so you have any arbitrary object that can be deserialized as JSON
2. model.FeatureDetails is a dictionary <name of feature, feature details object>.

src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
```diff
@@ -27,9 +27,15 @@ public StatusModel()
         /// <summary>The network the current node is running on.</summary>
         public string Network { get; set; }
 
+        /// <summary> The processId of the node.</summary>
+        public int PID { get; set; }
```
Aprogiena:
```
        /// <summary>System identifier of the node's process.</summary>
        public int ProcessId { get; set; }
```

src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
```diff
@@ -27,9 +27,15 @@ public StatusModel()
...
+
         /// <summary>The height of the consensus.</summary>
         public int ConsensusHeight { get; set; }
 
+        /// <summary> The height of the full blocks downloaded to BlockStore </summary>
```
Aprogiena: `/// <summary>Height of the last block in the block store.</summary>`

Note:
- no extra space at start or end
- dot at the end
- either use `<see>` if you want to talk about name of specific component (or namespace - i.e. `<see cref="BlockStore"/>` would be for namespace, but you would probably need to use full path as you have no usings in this file -> `<see cref="Stratis.Bitcoin.Features.BlockStore"/>`), or use English description without see

src/Stratis.Bitcoin/Interfaces/IStoreStateProvider.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        ChainedHeader StoreTip { get; }
```
Aprogiena: and here

bokobza:
but the Stratis.Bitcoin is not supposed to know whether we have a store or not.
If you do that then you open the door to having all sorts of things in Stratis.Bitcoin that may or may not have been initialized.
What about having the features publish their own statuses?

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -8,11 +8,12 @@
 using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.BlockStore
 {
-    /// <summary>
+    /// <summary>Ch
```
Aprogiena: ?

src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
```diff
@@ -27,13 +27,14 @@ public StatusModel()
...
-        public int PID { get; set; }
+        /// <summary>System identifier of the node's process.</summary>
+        public int ProcessId { get; set; }
 
         /// <summary>The height of the consensus.</summary>
         public int ConsensusHeight { get; set; }
 
-        /// <summary> The height of the full blocks downloaded to BlockStore </summary>
+        /// <summary>Height of the highestfull block in persistent storage.</summary>
+        /// <see cref="Stratis.Bitcoin.Features.BlockRepository.HighestPersistedBlock.Height"/>
```
Aprogiena: That would be `<seealso>`

src/Stratis.Bitcoin/Controllers/Models/StatusModel.cs
```diff
@@ -27,13 +27,14 @@ public StatusModel()
...
-        /// <summary> The processId of the node.</summary>
-        public int PID { get; set; }
+        /// <summary>System identifier of the node's process.</summary>
+        public int ProcessId { get; set; }
 
         /// <summary>The height of the consensus.</summary>
         public int ConsensusHeight { get; set; }
 
-        /// <summary> The height of the full blocks downloaded to BlockStore </summary>
+        /// <summary>Height of the highestfull block in persistent storage.</summary>
```
Aprogiena: typo highestfull

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -119,6 +119,11 @@ public Task<uint256> GetTrxBlockIdAsync(uint256 trxid)
             return this.blockRepository.GetTrxBlockIdAsync(trxid);
         }
 
+        public ChainedHeader GetHighestPersistedBlock()
```
Aprogiena: `<inheritdoc/>`

DanGould: Does this work even though It's a method & not a field of the name HighestPersistedBlock?

src/Stratis.Bitcoin/Controllers/NodeController.cs
```diff
@@ -58,14 +58,18 @@ public IActionResult Status()
...
+                ProcessId = Process.GetCurrentProcess().Id,
                 Network = this.fullNode.Network.Name,
                 ConsensusHeight = this.chainState.ConsensusTip.Height,
-                BlockStoreHeight = storeStateProvider.StoreTip.Height,
                 DataDirectoryPath = this.nodeSettings.DataDir,
                 RunningTime = this.dateTimeProvider.GetUtcNow() - this.fullNode.StartTime
             };
 
+            // Include BlockStore Height if enabled
+            foreach (var blockStore in this.fullNode.Services.Features)
```
Aprogiena: that iteration variable is not a block store it is a feature that might not be block store, so it is kind of misleading to call it block store

src/Stratis.Bitcoin/Interfaces/IBlockStore.cs
```diff
@@ -8,5 +8,8 @@ public interface IBlockStore
         Task<Transaction> GetTrxAsync(uint256 trxid);
 
         Task<uint256> GetTrxBlockIdAsync(uint256 trxid);
+
+        // <summary>Fetch the last block stored to disk.</summary>
```
Aprogiena: Triple /

DanGould:
Did this fail because the loop casts as IFullNodeFeature so feature never `is IBlockStore`?

bokobza:
It failed because Mac.

PR #1414, opened 2018-05-16
Pull out coinview validation rule

majikandy:
This is the conversion of the final rule in the Validator and removal of the validator.

https://github.com/stratisproject/StratisBitcoinFullNode/issues/124

Despite going down the route of breaking the Uxto CoinviewRule into lots of subrules that run per transaction.  @dangershony and I decided to go back to the idea of pulling it out as just one rule (at least in the first instance).

The code review on that would have been a challenge, so to reduce that complexity and make it much easier in this first instance, the starting point was to rename the PowConsensusValidator as PowCoinviewRule and the equivalent same change for Pos.

Then, changing all the usages of IPos/PowValidators to use the rules engine (or a rule from the rules engine) - allowing us to remove the remaining validator code completely.

This PR should be labelled as critical.

Some of the documentation was moved around to avoid deleting it completely when deleting the 2 interfaces, so I want to make sure this is the in the right place too.


majikandy:
Note: I have 2 integration tests on this currently failing locally in Release mode but passing in debug which implies some kind of timing issue potentially - but the same happens on master for me so it shouldn't be related to this change.

`Statking_wallet_can_mint_new_coins` and `Sending_Coins_Before_Maturity_Are_Not_Included_In_The_Wallets_Transaction_History`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -232,7 +232,7 @@ public class CoinstakeContext
         private readonly IWalletManager walletManager;
 
         /// <summary>Provides value for PoS reward and checks PoS kernel.</summary>
-        private readonly IPosConsensusValidator posConsensusValidator;
+        //private readonly IPosConsensusValidator posConsensusValidator;
```
noescape00: Remove commented code please if it's not needed

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -373,7 +373,7 @@ public class CoinstakeContext
             this.targetReserveBalance = 0; // TOOD:settings.targetReserveBalance
             this.stakeProgressFlag = StakeNotInProgress;
 
-            this.posConsensusValidator = consensusLoop.Validator as IPosConsensusValidator;
+            //this.posConsensusValidator = consensusLoop.Validator as IPosConsensusValidator;
```
noescape00: Remove commented code please if it's not needed

majikandy: Good catch - what shall I do with the following lines in NLog config....

I just did a global search and found it.....

<logger name="Stratis.Bitcoin.Features.Consensus.PowConsensusValidator" minlevel="Trace" writeTo="debugConsensusValidationFile" />

majikandy: since there is this line below it - I think we can delete the above lines for pos and pow?

`<logger name="Stratis.Bitcoin.Features.Consensus.Rules.*" minlevel="Trace" writeTo="debugConsensusValidationFile" />`

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
```diff
@@ -18,22 +18,20 @@ public class ConsensusManager : INetworkDifficulty, IGetUnspentTransaction
...
         }
 
         public Target GetNetworkDifficulty()
         {
-            if ((this.ConsensusValidator?.ConsensusParams != null) && (this.ConsensusLoop?.Tip != null))
-                return this.ConsensusLoop?.Tip?.GetWorkRequired(this.ConsensusValidator.ConsensusParams);
+            var powCoinviewRule = this.ConsensusLoop.ConsensusRules.GetRule<PowCoinviewRule>();
+            if ((powCoinviewRule.ConsensusParams != null) && (this.ConsensusLoop?.Tip != null))
+                return this.ConsensusLoop?.Tip?.GetWorkRequired(powCoinviewRule.ConsensusParams);
             else
```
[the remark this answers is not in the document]
majikandy: it is,  I'll remove

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
majikandy: agreed

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
```
noescape00: Redundant

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -348,7 +343,7 @@ public void Stop()
 
             using (await this.consensusLock.LockAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false))
             {
-                blockValidationContext.RuleContext = new RuleContext(blockValidationContext, this.Validator.ConsensusParams, this.Tip);
+                blockValidationContext.RuleContext = new RuleContext(blockValidationContext, this.Chain.Network.Consensus, this.Tip);
```
noescape00: Maybe here it will make more sense to have Network taken from the CI instead of the Chain?

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
```diff
@@ -18,24 +18,22 @@ public class ConsensusManager : INetworkDifficulty, IGetUnspentTransaction
...
-            this.ConsensusValidator = consensusValidator;
         }
 
         public Target GetNetworkDifficulty()
         {
-            if ((this.ConsensusValidator?.ConsensusParams != null) && (this.ConsensusLoop?.Tip != null))
-                return this.ConsensusLoop?.Tip?.GetWorkRequired(this.ConsensusValidator.ConsensusParams);
-            else
-                return null;
+            var powCoinviewRule = this.ConsensusLoop.ConsensusRules.GetRule<PowCoinviewRule>();
```
noescape00: Won't this work?

```
public Target GetNetworkDifficulty()
{
    return this.Network.Consensus != null ? this.ConsensusLoop?.Tip?.GetWorkRequired(this.Network.Consensus) : null;
}
```

src/Stratis.Bitcoin.Features.Consensus/PowCoinviewRule.cs
```diff
@@ -5,83 +5,58 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
-using NBitcoin.Crypto;
 using Stratis.Bitcoin.Base.Deployments;
 using Stratis.Bitcoin.Features.Consensus.Interfaces;
```
noescape00: Redundant nasmespace

src/Stratis.Bitcoin.Features.Consensus/PowCoinviewRule.cs
```diff
@@ -5,83 +5,58 @@
...
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="PowConsensusValidator"/> class.
-        /// </summary>
-        /// <param name="network">Specification of the network the node runs on - regtest/testnet/mainnet.</param>
-        /// <param name="checkpoints">Provider of block header hash checkpoints.</param>
-        /// <param name="dateTimeProvider">Provider of time functions.</param>
-        /// <param name="loggerFactory">Factory for creating loggers.</param>
-        public PowConsensusValidator(Network network, ICheckpoints checkpoints, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
+        public override void Initialize()
```
noescape00: No comment, no logs (For PoS rule.Initialize too)

src/Stratis.Bitcoin.Features.Consensus/Rules/IConsensusRules.cs
```diff
@@ -47,6 +52,13 @@ public interface IConsensusRules
         /// <param name="ruleContext">A context that holds information about the current validated block.</param>
         /// <returns>The processing task.</returns>
         Task ValidateAsync(RuleContext ruleContext);
+
+        /// <summary>
+        /// Gets the consensus rule that is assignable to the supplied generic type.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
```
noescape00: empty <returns>

src/Stratis.Bitcoin.IntegrationTests/Extensions/CoreNodeExtensions.cs
```diff
@@ -12,13 +13,16 @@ public static class CoreNodeExtensions
     {
         public static Money GetProofOfWorkRewardForMinedBlocks(this CoreNode node, int numberOfBlocks)
         {
-            var powValidator = node.FullNode.NodeService<IPowConsensusValidator>();
-            var halvingInterval = powValidator.ConsensusParams.SubsidyHalvingInterval;
+            var coinviewRule = node.FullNode.NodeService<IRuleRegistration>().GetRules().OfType<PowCoinviewRule>().Single();
```
noescape00: Unvar please

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
```diff
@@ -18,24 +18,22 @@ public class ConsensusManager : INetworkDifficulty, IGetUnspentTransaction
 
         public Network Network { get; private set; }
 
-        public IPowConsensusValidator ConsensusValidator { get; private set; }
 
-        public ConsensusManager(IConsensusLoop consensusLoop = null, IDateTimeProvider dateTimeProvider = null, NodeSettings nodeSettings = null, Network network = null,
-            IPowConsensusValidator consensusValidator = null)
+        public ConsensusManager(IConsensusLoop consensusLoop = null, IDateTimeProvider dateTimeProvider = null, NodeSettings nodeSettings = null, Network network = null)
```
noescape00: None of them can be null

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
```diff
@@ -18,24 +18,22 @@ public class ConsensusManager : INetworkDifficulty, IGetUnspentTransaction
 
         public Network Network { get; private set; }
 
-        public IPowConsensusValidator ConsensusValidator { get; private set; }
```
noescape00: empty line

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
noescape00: Also `this.Network.Consensus` can't be null

src/Stratis.Bitcoin.IntegrationTests/Extensions/CoreNodeExtensions.cs
majikandy: type is on the right - therefore complies with policy  (.... I'll duck for cover)

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: Maybe, I don't know definitively either way, so I'll take whoever's advice on this. Just let me know

src/Stratis.Bitcoin.IntegrationTests/Extensions/CoreNodeExtensions.cs
noescape00: only if type is after `new` or `as Type` or `(Type)x`

`var x = y<Type1>.Smth<Type2>().DoMagic<Type3>().Single()` doesn't count.

Aprogiena: This is not about Var or not here. A puppy died when this line was written.

dangershony: `var` is fine here

majikandy: >This is not about Var or not here. A puppy died when this line was written.
we have a little wrapper for it , GetRule.
However, probably in this context in the test it can probably just new up the rule directly

src/Stratis.Bitcoin.Features.Consensus/ConsensusManager.cs
majikandy: The only thing changed on this constructor was the removal of the consensusValidator - but more than happy to un optional them all if that is the right course of action

src/Stratis.Bitcoin.Features.Consensus/PosCoinviewRule.cs
```diff
@@ -45,92 +20,77 @@ public class PosConsensusValidator : PowConsensusValidator, IPosConsensusValidat
...
+            // Compute and store the stake proofs.
+            this.CheckAndComputeStake(context);
 
-            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
-            this.StakeValidator = stakeValidator;
-            this.stakeChain = stakeChain;
-            this.consensusOptions = network.Consensus.Option<PosConsensusOptions>();
+            base.RunAsync(context);
+
+            this.stakeChain.Set(context.BlockValidationContext.ChainedHeader, context.Stake.BlockStake);
```
noescape00: .Set here is 

```
public void Set(ChainedHeader chainedHeader, BlockStake blockStake)
{
    this.SetAsync(chainedHeader, blockStake).GetAwaiter().GetResult();
}
```

So since we are calling it from an async method maybe it will be a good idea to remove .Set at all and use SetAsync instead?

src/Stratis.Bitcoin.IntegrationTests/Extensions/CoreNodeExtensions.cs
dangershony: 😄 it was an evil puppy

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PosFutureDriftRuleTest.cs
```diff
@@ -21,7 +21,7 @@ public PosFutureDriftRuleTest()
         [Fact]
         public async Task RunAsync_HeaderTimestampTooNew_WithoutReducedDrift_ThrowsBlockTimestampTooFarConsensusErrorAsync()
         {
-            var futureDriftTimestamp = (PosConsensusValidator.DriftingBugFixTimestamp - 100);
+            var futureDriftTimestamp = (PosCoinviewRule.DriftingBugFixTimestamp - 100);
```
majikandy: Does `DriftingBugFixTimestamp` now below somewhere else rather than in the coinview rule? Any suggestions?

noescape00: Yes, in the `PosFutureDriftRule`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -4,39 +4,13 @@
 using Stratis.Bitcoin.Features.Consensus.Interfaces;
 using Stratis.Bitcoin.Utilities;
 
-namespace Stratis.Bitcoin.Features.Consensus
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
 {
     /// <summary>
-    /// Provides functionality for verifying validity of PoS block.
+    /// Proof of stake override for the coinview rules - BIP68, MaxSigOps and BlockReward checks
```
noescape00: `.` is missing

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -45,92 +19,81 @@ public class PosConsensusValidator : PowConsensusValidator, IPosConsensusValidat
...
-        public IStakeValidator StakeValidator { get; }
+        public IStakeValidator StakeValidator { get; set; }
 
         /// <summary>Database of stake related data for the current blockchain.</summary>
-        private readonly IStakeChain stakeChain;
+        private IStakeChain stakeChain;
+
+        private PosConsensusOptions posConsensusOptions;
+
+        public override void Initialize()
```
noescape00: Please add inheritdoc

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,87 +1,63 @@
...
-using Stratis.Bitcoin.Features.Consensus.Interfaces;
-using Stratis.Bitcoin.Features.Consensus.Rules.CommonRules;
 using Stratis.Bitcoin.Utilities;
 
-namespace Stratis.Bitcoin.Features.Consensus
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
 {
     /// <summary>
-    /// Provides functionality for verifying validity of PoW block.
+    /// The proof of work coinview update rules. Validates the UTXO set is correctly spent.
```
noescape00: It's not only correctly spent but this rule also creating new outputs. It's an analog if BTC Core's `ConnectInputs`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,87 +1,63 @@
...
 
-        /// <summary>Instance logger.</summary>
-        private readonly ILogger logger;
-
-        /// <summary>Provider of block header hash checkpoints.</summary>
-        protected readonly ICheckpoints Checkpoints;
-
         /// <summary>Consensus parameters.</summary>
-        public NBitcoin.Consensus ConsensusParams { get; }
+        public NBitcoin.Consensus ConsensusParams { get; set; }
```
noescape00: This and 29 should not be public. Why would other component load consensus params from some rule if they are already in the Network?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,87 +1,63 @@
...
 
             Block block = context.BlockValidationContext.Block;
             ChainedHeader index = context.BlockValidationContext.ChainedHeader;
             DeploymentFlags flags = context.Flags;
             UnspentOutputSet view = context.Set;
 
-            this.PerformanceCounter.AddProcessedBlocks(1);
-            taskScheduler = taskScheduler ?? TaskScheduler.Default;
+            this.Parent.PerformanceCounter.AddProcessedBlocks(1);
+            var taskScheduler = TaskScheduler.Default;
```
noescape00: This variable is not needed. 

`checkInput.Start(taskScheduler);` : 118 - this can be called without any params- default Task Scheduler will be used

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,87 +1,63 @@
...
             long sigOpsCost = 0;
             Money fees = Money.Zero;
             var checkInputs = new List<Task<bool>>();
             for (int txIndex = 0; txIndex < block.Transactions.Count; txIndex++)
             {
-                this.PerformanceCounter.AddProcessedTransactions(1);
+                this.Parent.PerformanceCounter.AddProcessedTransactions(1);
                 Transaction tx = block.Transactions[txIndex];
                 if (!context.SkipValidation)
                 {
```
noescape00: Empty line after that is redundant

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -128,14 +104,14 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
                         var txData = new PrecomputedTransactionData(tx);
```
noescape00: On line 97 before the Throw- please add a log

src/Stratis.Bitcoin.Features.Consensus/Rules/IConsensusRules.cs
```diff
@@ -19,6 +19,11 @@ public interface IConsensusRules
         /// </summary>
         IEnumerable<ConsensusRuleDescriptor> Rules { get; }
 
+        /// <summary>
+        /// Keeps track of how much time different actions took to execute and how many times they were executed.
```
noescape00: After you've added a comment here there is no need to have an exact same comment in `ConsensusRules.cs`. Please add inheritdoc for the performance counter in `ConsensusRules.cs`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: Yes, now I know it is in network, you are right.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: ah cool

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: ok

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
majikandy: ok

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -172,6 +172,11 @@ public void SetErrorHandler(ConsensusRule errorHandler)
                 }
             }
         }
+
```
noescape00: Please add inheritdoc

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
```diff
@@ -17,8 +17,9 @@ public abstract class ConsensusRule
         /// <summary>Instance logger.</summary>
         public ILogger Logger { get; set; }
 
+        //TODO: should this be IConsensusRules?
```
noescape00: I think it should. But in order to do that you need to make IConsensusRules interface to reflect every member of the actual ConsensusRules

@dangershony

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
```diff
@@ -17,8 +17,9 @@ public abstract class ConsensusRule
...
         /// <summary>The engine this rule belongs to.</summary>
-        public ConsensusRules Parent { get; set; }
+        public ConsensusRules Parent { get; set; }
```
noescape00: Plase remove space in the end of the line

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -10,6 +10,7 @@
 using Stratis.Bitcoin.Features.Consensus;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
 using Stratis.Bitcoin.Features.Consensus.Interfaces;
```
noescape00: Namespace is redundant.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -734,7 +734,7 @@ private void CreateMempoolEntry(MempoolValidationContext context, long acceptTim
             // be mined yet.
             // Must keep pool.cs for this unless we change CheckSequenceLocks to take a
             // CoinsViewCache instead of create its own
-            if (!CheckSequenceLocks(this.network, this.chain.Tip, context, PowConsensusValidator.StandardLocktimeVerifyFlags, context.LockPoints))
+            if (!CheckSequenceLocks(this.network, this.chain.Tip, context, PowCoinviewRule.StandardLocktimeVerifyFlags, context.LockPoints))
```
noescape00: @dangershony, should the StandardLocktimeVerifyFlags be inside a rule? Maybe there is a better place for this? 
This variable is not used in the rule, only in block assembler and mempoolvalidator

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -1043,7 +1043,7 @@ private void CheckAllInputs(MempoolValidationContext context)
             Transaction tx = context.Transaction;
             if (!context.Transaction.IsCoinBase)
             {
-                this.consensusValidator.CheckInputs(context.Transaction, context.View.Set, this.chain.Height + 1);
+                this.consensusRules.GetRule<PowCoinviewRule>().CheckInputs(context.Transaction, context.View.Set, this.chain.Height + 1);
```
noescape00: Logic here is changed. Before when we were getting PowCoinviewRule and we are running PoS node we would get an instance which has overriden CheckInputs method. But now we are getting it always from PowCoinviewRule so even in case we are running PoS node we will still be checking according to the PoW rules for CheckInputs

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -4,133 +4,93 @@
...
 
-        /// <summary>Drifting Bug Fix, hardfork on Sat, 19 Nov 2016 00:00:00 GMT.</summary>
-        public const long DriftingBugFixTimestamp = 1479513600;
-
-        /// <summary>Instance logger.</summary>
-        private readonly ILogger logger;
-
         /// <summary>Provides functionality for checking validity of PoS blocks.</summary>
-        public IStakeValidator StakeValidator { get; }
+        public IStakeValidator StakeValidator { get; set; }
```
noescape00: StakeValidator should be a private member

src/Stratis.Bitcoin.Features.Consensus/Rules/IConsensusRules.cs
majikandy: Cool

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
majikandy: how did you even find that ?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
majikandy: I think the confusion is that it is called `PowCoinviewRule` but since when running a PoS node, only the `PosCoinviewRule` is registered, and that inherits from PowCoinviewRule - I believe it would actually get the PosCoinviewRule as expected and use the `CheckInputs` from there.    Did you actually experience this behaviour when debugging it or just from eyeballing the code?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -4,133 +4,93 @@
...
-    /// <item>Coinbase 'scriptSig' starts with serialized block height value. This means that coinbase transaction commits to the height of the block it appears in.</item>
-    /// </list>
-    /// </remarks>
-    public class PosConsensusValidator : PowConsensusValidator, IPosConsensusValidator
+    [ExecutionRule]
+    public class PosCoinviewRule : PowCoinviewRule
     {
         /// <summary>PoS block's timestamp mask.</summary>
         /// <remarks>Used to decrease granularity of timestamp. Supposed to be 2^n-1.</remarks>
         public const uint StakeTimestampMask = 0x0000000F;
```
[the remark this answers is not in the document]
majikandy: moving to BlockHeaderPosContextualRule as it is used there.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -785,8 +781,9 @@ private void CheckStake(Block block, ChainedHeader chainTip)
             }
 
             this.logger.LogTrace("Worker #{0} found the kernel.", workersResult.KernelFoundIndex);
+            //DAN
```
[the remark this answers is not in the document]
majikandy: that was temporary to ask @dangershony something

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
majikandy: agree

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: @dangershony mentioned that `ConsensusLoop` will go anyway, so said to leave it as this for now.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
majikandy: let's keep the TODO for now then

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
majikandy: moving to MempoolValidator so that it ends up in a mempool rule eventually

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
Aprogiena: private member, not private setter

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            this.rules = new FullNodeBuilderConsensusExtension.PosConsensusRulesRegistration().GetRules();
+        }
+
+        [Fact]
+        public void GetRules_ForPOS_ReturnsListOfRegisteredPowRules()
+        {
+            this.rules.Count().Should().Be(22);
+
+            this.rules.ElementAt(0).Should().BeOfType<BlockHeaderRule>();
```
dangershony: ah interesting this is a good idea.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,102 +1,72 @@
...
-        private readonly Network network;
+        /// <inheritdoc />
+        public override void Initialize()
+        {
+            this.Logger.LogTrace("()");
 
-        /// <summary>Provider of time functions.</summary>
-        protected readonly IDateTimeProvider dateTimeProvider;
+            this.ConsensusParams = this.Parent.Network.Consensus;
+            this.ConsensusOptions = this.ConsensusParams.Option<PowConsensusOptions>();
```
dangershony: I was hoping to get rid of this TODO but happy to fist push the change.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
dangershony: I prefer to keep the todo for now as we still need to revisit how we measure download speed

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -73,7 +74,7 @@ public class BlockValidationContext
     /// <remarks>
     /// <para>
     /// Blocks are coming from <see cref="ILookaheadBlockPuller"/> or Miner/Staker and get validated by
-    /// either the <see cref="PowConsensusValidator"/> for PoW or the <see cref="PosConsensusValidator"/> for PoS.
+    /// either the <see cref="PowCoinviewRule"/> for PoW or the <see cref="PosCoinviewRule"/> for PoS.
```
dangershony: "And get validated by the IConsensusRules engine"

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -175,7 +171,7 @@ public class ConsensusLoop : IConsensusLoop
             IStakeChain stakeChain = null)
         {
             Guard.NotNull(asyncLoopFactory, nameof(asyncLoopFactory));
-            Guard.NotNull(validator, nameof(validator));
+            Guard.NotNull(consensusRules, nameof(consensusRules));
```
dangershony: This is a duplicate

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -459,14 +454,7 @@ public void ValidateBlock(RuleContext context)
...
-            // Validate the UTXO set is correctly spent.
-            this.logger.LogTrace("Executing block.");
-            using (new StopwatchDisposable(o => this.Validator.PerformanceCounter.AddBlockProcessingTime(o)))
-            {
-                this.Validator.ExecuteBlock(context);
-            }
+            await this.ConsensusRules.ValidateAndExecuteAsync(context);
 
             // Persist the changes to the coinview. This will likely only be stored in memory,
             // unless the coinview treashold is reached.
```
dangershony: I suppose this should go in its own rule as well (or at the end of the CoinViewRule)
` await this.UTXOSet.SaveChangesAsync` we can wait with this for another PR.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -127,19 +127,20 @@ public class MempoolValidator : IMempoolValidator
 
         /// <summary>Coin view of the memory pool.</summary>
         private readonly CoinView coinView;
+        private readonly IConsensusRules consensusRules;
```
dangershony: Space and a comment to keep in line with the rest of the class you can use ` /// <inheritdoc seeref="IConsensusRules" />`

src/Stratis.Bitcoin.IntegrationTests/Program.cs
```diff
@@ -2,17 +2,21 @@
...
 
 namespace Stratis.Bitcoin.IntegrationTests
 {
     internal class Program
     {
         public static void Main(string[] args)
         {
-            RunAllTestsOf<WalletTests>();
-            RunAllTestsOf<NodeSyncTests>();
+            new ProofOfStakeMintCoinsSpecification(new TestOutputHelper()).Staking_wallet_can_mint_new_coins();
```
dangershony: Undo this file

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
dangershony: Andy did also suggest that and in order to have minimal changes for this PR I preferred to not use a base.
If we do a base there will be other parts in the coinview rule we can fix.
So this is fine for now and besides this will be refactored when we do the mempool rules.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRule.cs
dangershony: CTRL + R , CTRL + W

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: agree, next PR

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: done, I also changed the line there that points to the coinview rules to....

/// See either the <see cref="FullNodeBuilderConsensusExtension.PowConsensusRulesRegistration"/> for PoW or the <see cref="FullNodeBuilderConsensusExtension.PowConsensusRulesRegistration"/> for PoS.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: it is

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: ok

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
majikandy: done

majikandy:
All requested changes now done afaik. 

I have successfully synced testnet and had it running for a while now, it is on the correct chain as per chainz. 

I have also staked for a while and managed to mine 5 blocks with no issues.

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
Aprogiena: Is it? I'm kind of missing the purpose of this test.Can you explain?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -0,0 +1,221 @@
...
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
+{
+    /// <summary>
+    /// Proof of stake override for the coinview rules - BIP68, MaxSigOps and BlockReward checks.
+    /// </summary>
+    [ExecutionRule]
+    public class PosCoinviewRule : PowCoinviewRule
+    {
+        /// <summary>Provides functionality for checking validity of PoS blocks.</summary>
+        public IStakeValidator StakeValidator { get; private set; }
```
Aprogiena: I still think this guy should be private

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,102 +1,72 @@
...
         /// <summary>Consensus parameters.</summary>
-        public NBitcoin.Consensus ConsensusParams { get; }
+        public NBitcoin.Consensus ConsensusParams { get; private set; }
 
         /// <summary>Consensus options.</summary>
-        public PowConsensusOptions ConsensusOptions { get; }
-
-        /// <summary>Keeps track of how much time different actions took to execute and how many times they were executed.</summary>
-        public ConsensusPerformanceCounter PerformanceCounter { get; }
+        public PowConsensusOptions ConsensusOptions { get; private set; }
```
Aprogiena: Again, I think this should be private

Aprogiena: And so should be the consensus parameters above

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -155,13 +128,15 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
                 bool passed = checkInputs.All(c => c.GetAwaiter().GetResult());
```
Aprogiena: Didn't we want to convert this to async?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -151,32 +154,32 @@ public class MempoolValidator : IMempoolValidator
 
         private Network network;
 
+
```
Aprogiena: Empty line?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
Aprogiena: I think that because of the minimal changes we will have plenty of bad code inserted in the code base and it will not get fixed later because we just forget about it. It is not the case of this particular code which you are correct will be Refactored for sure, but in other cases and afraid no one holds the to do list, so I would expect many things to be just forgotten.

src/Stratis.Bitcoin.Features.Miner.Tests/PosBlockAssemblerTest.cs
```diff
@@ -440,23 +442,20 @@ private void SetupConsensusLoop()
                 {
                     this.callbackRuleContext = c;
                 }).Verifiable();
-
-            this.consensusLoop.Setup(c => c.Validator)
-                .Returns(this.validator.Object);
         }
 
-        private void SetupValidator()
+        private void SetupRulesEngine(ConcurrentChain chain)
```
Aprogiena: What happened to the indentation here?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
majikandy: I absolutely agree, we need to take that list and promote it to the board, and tag it is sp2 so that they are candidates for the next sprint

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
Aprogiena: The problem is that this test does not guarantee the correct order. It is just long-winded way how to copy the original list.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
Aprogiena: It looks like some of the comments were already overwritten. We would probably need to go back in emails or I don't know how to recover the list. There should be an issue created with a list of follow-ups immediately when you decide to skip some comment, otherwise it gets forgotten.

majikandy: what do you mean by skip some comment? the PR comment? or code comment?

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
majikandy: Why won't it guarantee the order?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
Aprogiena: was either commit comment or something, but i can rememember at least a couple of comments that were not implemented nor said not to be implemented

this is one of the things
https://github.com/stratisproject/StratisBitcoinFullNode/pull/1414#discussion_r189283428

others i can't even find anymore :-( we spent quite a lot of time on a review making a lot of comments and it looks like they disappeared or something

Aprogiena: oh, here is another one https://github.com/stratisproject/StratisBitcoinFullNode/pull/1414#discussion_r189021413

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
Aprogiena: because it is only a copy of the list, it has no logic to know whether the main code order was correct

Aprogiena: basically it is a test that is fit to the code, so that current code passes it, as such it is very much not testing anything as if the current order was wrong, you would not get test failed from it

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
Aprogiena: ok, found them now, basically just ignored, they did not disappear, just hard to find - https://github.com/stratisproject/StratisBitcoinFullNode/pull/1414#discussion_r189019667

https://github.com/stratisproject/StratisBitcoinFullNode/pull/1414#discussion_r189020082

maybe there are more, not sure

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: I thought you just meant private set which is what I did, but yes it can be made a fully private member, as only the inheriting PosCoinviewRule uses it and that can get it direct from the Parent so I'll change it to do that.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: I didn't know we wanted to do this, but I will.

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
dangershony: It is testing that the rules have been correctly ordered on the Registration class, if someone breaks the order by mistake this will fail (its already better then nothing) testing on the engine itself can also be a good idea.

However we have to remember this will be network specific and later rules will move to the Network.Consensus class.

I am fine with this as is right now as it holds some value.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
dangershony: This is available on the DI so private/public does not matter much.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
dangershony: Well we can yes, we will be changing behaviour from how it was before.
In fact I think the script verification should go in its own rule but I did not want too many changes in behaviour.

src/Stratis.Bitcoin.Features.Miner.Tests/PosBlockAssemblerTest.cs
majikandy: It lost its way somehow :) Fixed in next commit

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
dangershony: Yes @majikandy please make sure you have responded to every comment either that you agree (and made the change) or start a discussion so nothing is forgotten.

src/Stratis.Bitcoin.Features.Consensus/PosCoinviewRule.cs
dangershony: @majikandy please implement this as well.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
majikandy: I'll remove, but worth noting there is commented code just above this is not good either and needs promoting to an issue/task.

However this isn't relevant to your comment

```
        // TODO: Implement Later with CheckRateLimit()
        //private readonly FreeLimiterSection freeLimiter;

        //private class FreeLimiterSection
        //{
        //  public double FreeCount;
        //  public long LastTime;
        //}
```

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
majikandy: Yep, agreed

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
Aprogiena: In my opinion it is not testing that the rules have been correctly ordered. I don't mind having this test inside, it is just not doing what you think it is doing.

majikandy: You keep saying that is it not testing they are correctly ordered, but it is testing they are correctly ordered. Arguably we could move this onto the Regsitration itself and check that the IConsensusRules (of pow and pos) each have their Rules correctly set.

There is no proof within this test that the system in general even chooses to use this list of rules, however it does prove that it if does that the rules are set in the correct order.

Can you suggest something better instead of this?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
Aprogiena: Agree

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
Aprogiena: I still can't see how is it testing the correct order. In my opinion it is only testing that the list in one file is in the same order as list in another file, nothing else. So the main value of this test is to prevent changing the main code without thinking twice. And that is because you will have failed test and maybe you will think twice. Quite likely so, you will just fix the test list in the same way you changed the main code list, therefore this test is quite weak in achieving even that.

I don't think the current design allows you to test the correct order. Some improvements could be made to have some weaker guarantees between some of the rules, but the full correct ordering is probably difficult to achieve this way. It makes much better sense to create a comment that argues about ordering and explains why the ordering selected as it is.

src/Stratis.Bitcoin.Features.Consensus/PosCoinviewRule.cs
noescape00: Pretty much, yes

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
majikandy: I cannot accept that a comment is an improvement over a unit test.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
majikandy: yep this is a private member now

src/Stratis.Bitcoin.Features.Consensus/PosCoinviewRule.cs
majikandy: cool, thanks - that will be in the next commit

majikandy: I'm also changing the PowCoinviewRule  part where it was:-

`bool passed = checkInputs.All(c => c.GetAwaiter().GetResult());`

to

`bool passed = (await Task.WhenAll(checkInputs)).All(b => b);`

Then the PosCoinviewRule call down to base can be awaited too.

src/Stratis.Bitcoin.Features.Consensus.Tests/PosConsensusRulesRegistrationTests.cs
majikandy: Created https://github.com/stratisproject/StratisBitcoinFullNode/issues/1435  to reflect this discussion for improvement

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
majikandy: created issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/1438

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -128,18 +101,18 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
                         var txData = new PrecomputedTransactionData(tx);
                         for (int inputIndex = 0; inputIndex < tx.Inputs.Count; inputIndex++)
```
majikandy: Created issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/1432  for a potential improvement to break this down into sub rules or sub component that run inside the transaction loop but are individually testable/reusable and more SOLID compliant than now.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -1,102 +1,72 @@
...
             var checkInputs = new List<Task<bool>>();
             for (int txIndex = 0; txIndex < block.Transactions.Count; txIndex++)
             {
-                this.PerformanceCounter.AddProcessedTransactions(1);
+                this.Parent.PerformanceCounter.AddProcessedTransactions(1);
                 Transaction tx = block.Transactions[txIndex];
                 if (!context.SkipValidation)
                 {
-                    
+                    // TODO: Simplify this condition.
```
majikandy: See issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/1433  for the improvement suggestion

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -117,8 +87,11 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
...
                     sigOpsCost += this.GetTransactionSignatureOperationCost(tx, view, flags);
-                    if (sigOpsCost > this.ConsensusOptions.MaxBlockSigopsCost)
+                    if (sigOpsCost > this.consensusOptions.MaxBlockSigopsCost)
+                    {
+                        this.Logger.LogTrace("(-)[BAD_BLOCK_SIG_OPS]");
                         ConsensusErrors.BadBlockSigOps.Throw();
+                    }
 
                     // TODO: Simplify this condition.
                     if (!tx.IsCoinBase && (!context.IsPoS || (context.IsPoS && !tx.IsCoinStake)))
```
majikandy: see issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/1433 for the improvement suggestion

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -37,20 +37,18 @@ public override void Initialize()
...
             this.Logger.LogTrace("()");
 
             // Compute and store the stake proofs.
             this.CheckAndComputeStake(context);
 
-            base.RunAsync(context);
+            await base.RunAsync(context);
 
-            this.stakeChain.Set(context.BlockValidationContext.ChainedHeader, context.Stake.BlockStake);
+            await this.stakeChain.SetAsync(context.BlockValidationContext.ChainedHeader, context.Stake.BlockStake);
```
Aprogiena: In both cases, please add configure await with false

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -125,7 +125,7 @@ public override Task RunAsync(RuleContext context)
             {
                 this.CheckBlockReward(context, fees, index.Height, block);
 
-                bool passed = checkInputs.All(c => c.GetAwaiter().GetResult());
+                bool passed = (await Task.WhenAll(checkInputs)).All(b => b);
```
Aprogiena: As above, you won't configure await with false

More importantly, however, it seems to me that you are losing a performance here. If I read the original code correctly,It will iterate through the tasks as long as they return true one by one, which means that is as 1 of the inputs is incorrect, the evaluation stops waiting for other tasks and we will return from the method

With your new implementation, this does not seem to hold as we will always wait for tasks to be completed before we iterate over their results.

Please correct me if I'm wrong because my linq–fu is weak.

I think we could very much improve the code even compared to the original version if we introduced the cancellation token and prevent the tasks to even start evaluation should the cancellation token be cancelled before they do so. This could be out of the scope of the PR, in which case please create an issue for that, however here we should definitely not make the things worse in this PR, so please replicate the original performance. Of course only if I read that correctly, if not please explain why it is the case because I would like to know.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
dangershony: I guess you never really saw spaghetti code.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
majikandy: Yes, good catch, my bad.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: I don't seem to be able to find out definitively either way but yes it would appear that we would be waiting for all to finish before checking All whereas the original seems like it could shortcut. But then it comes down to whether the parallelism of the Task.When all outweighs that (perhaps in most cases all the tasks pass so awaiting all would be better?)

I can either return it to the original or perhaps

```
            foreach (Task<bool> checkInput in checkInputs)
            {
                if (await checkInput.ConfigureAwait(false) == false)
                {
                    ConsensusErrors.BadTransactionScriptError.Throw();
                }
            }
```
What do you think?

Aprogiena: Yes the for each is a simple solution, please do that and later we can improve using the cancellation token.

Please preserve the exit log, however:
```
 this.Logger.LogTrace("(-)[BAD_TX_SCRIPT]");
```

It is not better to wait for all every time because you need to think about the worst-case. So we want to quit checking as soon as possible to prevent denial of service attacks. This will be fully mitigated only when we later implement that cancellation token, but even here with just the for each instead of wait all we are at least no worse than original code.

majikandy: Yes of  course re the log line, I was just kinda psuedoing it

majikandy: Issue raised for cancellation token:  https://github.com/stratisproject/StratisBitcoinFullNode/issues/1446

Aprogiena: tx

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -152,16 +125,28 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
...
+                {
+                    var result = await checkInput;
+                    if(!result)
+                    {
+                        this.Logger.LogTrace("(-)[BAD_TX_SCRIPT]");
+                        ConsensusErrors.BadTransactionScriptError.Throw();
+                    }
+                }
+
+                foreach (Task<bool> checkInput in checkInputs)
```
Aprogiena: erm, double foreach?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -152,16 +125,28 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
             {
                 this.CheckBlockReward(context, fees, index.Height, block);
 
-                bool passed = checkInputs.All(c => c.GetAwaiter().GetResult());
-                if (!passed)
+                foreach (Task<bool> checkInput in checkInputs)
+                {
+                    var result = await checkInput;
+                    if(!result)
```
Aprogiena: space after if

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
```diff
@@ -152,16 +125,28 @@ public virtual void ExecuteBlock(RuleContext context, TaskScheduler taskSchedule
             {
                 this.CheckBlockReward(context, fees, index.Height, block);
 
-                bool passed = checkInputs.All(c => c.GetAwaiter().GetResult());
-                if (!passed)
+                foreach (Task<bool> checkInput in checkInputs)
+                {
+                    var result = await checkInput;
```
Aprogiena: ConfigureAwait

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PowCoinviewRule.cs
majikandy: what a weird rule :) 

Done.

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PowCoinViewRuleTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        public PowCoinViewRuleTests()
+        {
+            this.rule = new PowCoinViewRule();
+        }
+
+        [Fact]
+        public void RunAsync_ValidatingATransactionThatIsNotCoinBaseButStillHasUnspentOutputsWithoutInput_ThrowsBadTransactionMissingInput()
+        {
+            this.GivenACoinbaseTransactionFromAPreviousBlock();
```
noescape00: >ATransactionThatIsNotCoinBaseButStillHasUnspentOutputs

but a tx doesn't have any outputs: `new Transaction { Inputs = { new TxIn() } }`

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PowCoinViewRuleTests.cs
```diff
@@ -0,0 +1,137 @@
...
+using Stratis.Bitcoin.Features.Consensus.CoinViews;
+using Stratis.Bitcoin.Features.Consensus.Rules;
+using Stratis.Bitcoin.Features.Consensus.Rules.CommonRules;
+using Stratis.Bitcoin.Utilities;
+using Xunit;
+
+namespace Stratis.Bitcoin.Features.Consensus.Tests.Rules.CommonRules
+{
+    /// <summary>
+    /// These tests only cover the first part of BIP68 and not the MaxSigOps, coinview update or scripts verify or calculate block rewards
```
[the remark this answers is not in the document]
Aprogiena: This is not really a new component, so adding any kind of tests should be sufficient as it is improvements to previous state. Full coverage might be too huge to include in one PR.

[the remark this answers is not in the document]
majikandy: Also it was really hard to get in there and test, but if we break this down into sub components it will get much easier to get tests around it. The tests I added were originally for the first sub rule when I had it broken down that way, but when the decision was made to defer that change, I put it back around this rule as a whole but obviously it didn't cover the whole rule, just only the first part. 

Ideally I wanted to cover all, it was just a big task to retro fit so we opted for line by line matching the new code to old and relying on manual testing and the review for the remaining parts.  Otherwise, I would have loved to have got the coverage.

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PowCoinViewRuleTests.cs
majikandy: My understanding of the first BIP68 check was that it is a transaction, that looks like a coinbase but it is not the first transaction and therefore not the coinbase so it gets rejected.

More than happy to take your advice on improving this to be more realistic

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/ConsensusRuleUnitTestBase.cs
```diff
@@ -56,20 +57,27 @@ protected ConsensusRuleUnitTestBase(Network network)
             };
         }
 
-        protected static void AddBlocksToChain(ConcurrentChain chain, int blockAmount)
+        protected void AddBlocksToChain(ConcurrentChain chain, int blockAmount)
         {
             var nonce = RandomUtils.GetUInt32();
             var prevBlockHash = chain.Tip.HashBlock;
+
+            this.ruleContext.Set = new UnspentOutputSet();
```
majikandy: by the way, these preparations were for the UpdateCoinViewRule that I had started writing using the unit test base that neurosploit put in.

However, that didn't end up making it into this PR as I couldn't quite get it working - however it still seemed reasonable to leave this coinview setup inside the base as it didn't have one previously and whilst not strictly needed for the tests in place now, it is still a realistic setup of the engine - So I kept the code here breaking YAGNI.  

If you think I should undo this file, I am happy to.

Issue #1395, opened 2018-05-08
Create additional Integration Tests


bokobza:
4 tests here:
* User tries to double-send and fails
* Create 2 tests that tests a transaction size limit:
  -  a test that has a big transaction but valid
  -  a test that has a big transaction but invalid (small size increment from the previous test)
* Create a transaction that uses sequence locks

noescape00:
@bokobza @fassadlr 
Which of those tests were implemented in 1424? Can you please update the issue?

fassadlr:
@noescape00 it was @thefazzer you needed to tag ;)

noescape00:
Ah, yes. Sorry 🙂

bokobza:
sending_transaction_over_policy_byte_limit
sending_transaction_near_policy_byte_limit
sending_transaction_with_double_spend_mined_trx
sending_transaction_with_double_spend_in_mempool

✔️ User tries to double-send and fails
Create 2 tests that tests a transaction size limit:
✔️ a test that has a big transaction but valid
✔️a test that has a big transaction but invalid (small size increment from the previous test)
❔  Create a transaction that uses sequence locks

herbepau:
_**MinerCreateBlockRelativeTimeLockedAsync**_ in **_MinerTests.cs_** already covers sequence locks (nSequence - relative time locks vs absolute time locks - nLockTime).  Accompanying this test are three other tests that also use and test against sequence locks: MinerCreateBlockNonFinalTxsInMempoolAsync, MinerCreateBlockAbsoluteHeightLockedAsync and MinerCreateBlockAbsoluteTimeLockedAsync.

PR #1424, opened 2018-05-18
Integration tests for transaction size limit and double spend failure.

thefazzer:
Relates to https://github.com/stratisproject/StratisBitcoinFullNode/issues/1395.


src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class SendingTransactionOverPolicyByteLimit
```
majikandy: In Given When Then,  the `When` should be the last action you take before then asserting the side effects of the system using `Then`s.

This test blurs that a bit too much as that last action you can take (sending the transaction) is happening in the Then and the When/And also has an assert that should actually be part of the Givens.   It is fairly reasonable to have an assert at the end of a Given (like the asserting the tx is in the range 90-100kb as you did) but usually you only do this when writing the test to prove it to yourself and ensure the setup is correct, still I do think that is ok to keep as the next reader/writer of the code will like to see that too if not clear from the action of creating the tx that size. 

There perhaps something like the following would be a better logical breaking up of the Given When Then

```
Given two connected nodes
And node 1 builds a transaction just UNDER the 100kb limit
When sending the transaction
Then the transaction is successfully mined and propagated
```

```
Given two connected nodes
And node 1 builds a transaction OVER the 100kb limit
When sending the transaction
Then the transaction fails with oversize transaction error
```


You might be able to make use of sending

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        private void node1_fails_with_oversize_tx_wallet_error()
+        {
+            this.caughtException.Should().BeOfType<WalletException>().Which.Message.Should().Contain("Transaction's size is too high");
+        }
+
+        private void node1_succeeds_sending_tx_to_node2()
+        {
+            this.caughtException.Should().BeNull();
```
majikandy: you can actually check it makes it to the other node beyond just checking not null.  You can check the mempool on the other node or the block if you are mining after the transaction send

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        private void MineBlocks(CoreNode node)
+        {
+            this.sharedSteps.MineBlocks(this.CoinBaseMaturity * 2, node, WalletAccountName, WalletName, WalletPassword);
```
majikandy: not sure you need to mine coinbase maturity * 2. Probably just maturity + 1 so that the coins from the first block can be spent

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class SendingTxWithDoubleSpend
+    {
+        [Fact]
+        public void sending_tx_with_double_spend()
```
majikandy: As with the other test, the boundaries need to change here between what is a Given, When and Then, line 3 looks like the When here and line 2 a Given, I'll inline those comments below actually

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
```diff
@@ -0,0 +1,16 @@
...
+        {
+            Given(wallets_with_coins);
+            When(coins_first_sent_to_receiving_wallet);
```
majikandy: `When` -> `And`

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
```diff
@@ -0,0 +1,16 @@
...
+            Then(attempt_made_to_spend_same_coins);
```
majikandy: `Then` to `When`

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+using Stratis.Bitcoin.Features.Wallet.Controllers;
+using Stratis.Bitcoin.Features.Wallet.Models;
+using Stratis.Bitcoin.IntegrationTests.EnvironmentMockUpHelpers;
+using Stratis.Bitcoin.IntegrationTests.TestFramework;
+using Stratis.Bitcoin.Utilities.JsonErrors;
+using Xunit.Abstractions;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class SendingTxWithDoubleSpend : BddSpecification
```
majikandy: I would say Transaction here instead of Tx since it is the test title

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+            mnemonic2.Words.Length.Should().Equals(12);
+
+            var addr = this.stratisSender.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
+            var wallet = this.stratisSender.FullNode.WalletManager().GetWalletByName("mywallet");
+            var key = wallet.GetExtendedPrivateKeyForAddress("123456", addr).PrivateKey;
+
+            this.stratisSender.SetDummyMinerSecret(new BitcoinSecret(key, this.stratisSender.FullNode.Network));
+            var maturity = (int)this.stratisSender.FullNode.Network.Consensus.Option<PowConsensusOptions>().CoinbaseMaturity;
+
+            this.stratisSender.GenerateStratis(maturity + 5);
```
majikandy: I seem to remember hearing we were suppsoed to use GenerateStratisWithMiner method rather than this

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+            this.stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight.Should().BeNull();
+        }
+
+        private void attempt_made_to_spend_same_coins()
+        {
+            //create double spend transaction
+            var sendtoDoubleSpend = this.stratisReceiver.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
+            sendtoDoubleSpend.Should().NotBeSameAs(this.sendto);
+
+            var trxCloned = this.transaction.Clone();
```
majikandy: mixing tx and trx and transaction in this class to all mean the same

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            var receivetotal = this.stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount);
+            receivetotal.Should().Equals(Money.COIN * 100);
+            this.stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight.Should().BeNull();
+        }
+
+        private void attempt_made_to_spend_same_coins()
+        {
+            //create double spend transaction
+            var sendtoDoubleSpend = this.stratisReceiver.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
```
majikandy: this is an address is it? the variable name seems wrong

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+            total.Should().Equals(Money.COIN * 105 * 50);
+ 
+            // sync both nodes
+            this.stratisSender.CreateRPCClient().AddNode(stratisReceiver.Endpoint, true);
+            TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(stratisReceiver, stratisSender));
+        }
+
+        private void coins_first_sent_to_receiving_wallet()
+        {
+            this.sendto = this.stratisReceiver.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
```
majikandy: this field needs a better name

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -0,0 +1,104 @@
...
+            var receivetotal = this.stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").Sum(s => s.Transaction.Amount);
+            receivetotal.Should().Equals(Money.COIN * 100);
+            this.stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight.Should().BeNull();
+        }
+
+        private void attempt_made_to_spend_same_coins()
+        {
+            //create double spend transaction
+            var sendtoDoubleSpend = this.stratisReceiver.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
+            sendtoDoubleSpend.Should().NotBeSameAs(this.sendto);
```
majikandy: I would guess this line could never fail so not needed?

bokobza:
@thefazzer Can you please address the above comments?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification.cs
thefazzer: Agreed.  Have followed the Given->And->When->Then pattern

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
thefazzer: Oversize transactions fail in the wallet prior to send due to NBitcoin policy constraints (StandardTransactionPolicy::MaxTransactionSize = 100000)

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
thefazzer: Agreed - done

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
thefazzer: Done

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
thefazzer: changed to unusedAddress

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
thefazzer: changed to receivingAddress

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
thefazzer: Agreed - removed

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
majikandy: Yes but this one is used in the success case right ?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
majikandy: *2 means you are mining 300 blocks so that the first 150 blocks are spendable, are you sure you need 7500 bitcoins for the test?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification.cs
```diff
@@ -0,0 +1,26 @@
...
+            When(sending_the_transaction);
+            Then(node1_succeeds_sending_tx_to_node2);
+        }
+
+        [Fact]
+        public void sending_transaction_over_policy_byte_limit()
+        {
+            Given(two_connected_nodes);
+            And(node1_builds_oversize_tx_to_send_to_node2);
+            Then(node1_fails_with_oversize_transaction_wallet_error);
```
dangershony: Should this be a wallet error? or does it mean the wallet raised the error but it failed to get in the mempool.

thefazzer: Yes we get a WalletException: Could not build the transaction. Details: Transaction's size is too high. Actual value is 107344, but the maximum is 100000
I understand the Wallet gets this restriction from NBitcoin StandardTransactionPolicy::MaxTransactionSize = 100000

dangershony: Ok so the failure comes from the mempool I take it?
Hmm thats not exactly the test I wanted, I want to see a trx arrive to the mempool from a peer and fails.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public partial class SendingTransactionWithDoubleSpend
+    {
+        [Fact]
+        public void sending_transaction_with_double_spend()
+        {
+            Given(wallets_with_coins);
+            And(coins_first_sent_to_receiving_wallet);
+            When(two_transactions_attempt_to_spend_same_unspent_outputs);
+            Then(mempool_rejects_doublespending_transaction);
```
dangershony: Just confirm this failure comes from the mempool (the method `AcceptToMempool`)

thefazzer: Correct.  When broadcast to other node w/SendTransaction a txn-mempool-conflict is thrown

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        private void node1_builds_oversize_tx_to_send_to_node2()
+        {
+            Node1BuildsTransactionToSendToNode2(2900);
+            
+        }
+
+        private void sending_the_transaction()
+        {
+            this.nodes[NodeOne].FullNode.NodeService<WalletController>().SendTransaction(new SendTransactionRequest(this.transaction.ToHex()));
```
dangershony: does ToHex() not take a network?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class SendingTransactionWithDoubleSpend
+    {
+        [Fact]
+        public void sending_transaction_with_double_spend()
```
dangershony: ok yes this is correct I would like to ask you make one more tests similar to this one that checks the double spent is confirmed.

So create the same test but mine the trx in to a block.
Should be just one more line of code and a different error message

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
thefazzer: AcceptToMempool prevalidation added + check of mempool to confirm 2x spent transaction not present

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
thefazzer: Done

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTransactionOverPolicyByteLimitSpecification_Steps.cs
```diff
@@ -58,14 +60,26 @@ private void two_connected_nodes()
...
 
+        private void mempool_of_receiver_node2_is_empty()
+        {
+            this.nodes[NodeTwo].FullNode.MempoolManager().GetMempoolAsync().Result.Should().BeEmpty();
+        }
+
+        private void mempool_of_node2_has_received_transaction()
+        {
+            List<uint256> mempool = this.nodes[NodeOne].FullNode.MempoolManager().GetMempoolAsync().Result;
+            TestHelper.WaitLoop(() => mempool.Any());
```
dangershony: ah you have to put the entire line in the waitloop
```
TestHelper.WaitLoop(() => this.nodes[NodeOne].FullNode.MempoolManager().GetMempoolAsync().Result.Any());
```

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification.cs
thefazzer: I now check it is mined and removed from mempool on the receiving node

When trying to double spend MempoolValidator throws error an checking the coinview but handles this internally, doesn't populate errors collection or propagate.  
Note this is why I don't call SendTransaction as it wraps AcceptToMemoryPool and will swallow error result

PR #1429, opened 2018-05-20
Ported listaccounts RPC call to the API

MSinteur317:


bokobza:
Yep, that's great, thanks @MSinteur317 !

PR #1444, opened 2018-05-22
Move BDD around to Tests.Common

monsieurleberre:
Resharper did the work :)
I would like to get this change in so we can publish a Test.Common nuget package that we can reuse in other projects, like Sidechains for instance


bokobza:
Looks good to me, just waiting on these tests passing + integration tests.

PR #1448, opened 2018-05-23
Fixed test to be order agnostic.

majikandy:
Test was getting argument exception when it was run on its own.


bokobza:
Thanks for also correcting the typo there ;-)

Issue #1375, opened 2018-05-02
Mempool behaviour SendMempoolPayloadAsync doesn't send anything


mikedennis:
Please correct me if I'm wrong but to me it looks like this code doesn't work:

https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs#L301-L314

```
            List<TxMempoolInfo> sends = await this.manager.MempoolLock.WriteAsync(() =>
            {
                List<TxMempoolInfo> ret = new List<TxMempoolInfo>();
                foreach (TxMempoolInfo txinfo in vtxinfo)
                {
                    uint256 hash = txinfo.Trx.GetHash();
                    this.inventoryTxToSend.Remove(hash);
                    if (filterrate != Money.Zero)
                        if (txinfo.FeeRate.FeePerK < filterrate)
                            continue;
                    this.filterInventoryKnown.TryAdd(hash, hash);
                }
                return ret;
            });
```

To me it seems like it's missing a step of adding it to sends after it removes it from ```this.inventoryTxToSend``` 

Unless I'm misunderstanding how this is suppose to work?

dangershony:
Yes this code is incomplete, so it needs to be done and also we need to compare with StratisX and bitcoin core behaviour for vulnerabilities.

mikedennis:
Ok I will look at this as part of the work I'm doing on mempool behaviour as it is some of the code I will be touching anyways.  Thanks!

dangershony:
yes please.

PR #1449, opened 2018-05-24
In mempool behaviour use own locks instead of mempool manager locks

mikedennis:
This PR addresses one of the issues in #1250 in regards to refactoring mempool behaviour to use it's own locks instead of relying on mempool manager locks. There is one occurrence that required leaving the mempool manager lock. Initially I tried removing it but that caused peers to be dropped, so it must be still required.
 
This PR also fixes the issue #1375 by actually adding the inventory items to the list of items that is sent.

I have also improved logging support in mempool behaviour which I needed in order to troubleshoot this PR, as well as added some high level logging into mempool orphans.

I tested this PR on both fully synced Bitcoin and Stratis Testnets by running the node fully synced for several hours with full mempool logging and examining the logs for any anomalies. Things appear to 
behave as expected.

Integration and unit tests were also run with no new failures identified.


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -291,20 +296,25 @@ public override object Clone()
...
                     uint256 hash = txinfo.Trx.GetHash();
                     this.inventoryTxToSend.Remove(hash);
                     if (filterrate != Money.Zero)
                         if (txinfo.FeeRate.FeePerK < filterrate)
+                        {
+                            this.logger.LogTrace("Fee too low, transaction ID '{0}' not added to inventory list.", hash);
                             continue;
+                        }
                     this.filterInventoryKnown.TryAdd(hash, hash);
+                    this.logger.LogTrace("Added transaction ID '{0}' to known inventory filter.", hash);
```
dangershony: One trace is enough for both lists

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -291,20 +296,25 @@ public override object Clone()
             //  filterrate = pto->minFeeFilter;
             //}
 
-            List<TxMempoolInfo> sends = await this.manager.MempoolLock.WriteAsync(() =>
+            List<TxMempoolInfo> sends = new List<TxMempoolInfo>();
```
dangershony: Please add an issue to implement the above code
`if (connman.OutboundTargetReached(false) && !pfrom->fWhitelisted)`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -508,14 +525,30 @@ public Task RelayTransaction(uint256 hash)
             // find all behaviours then start an exclusive task
             // to add the hash to each local collection
             IEnumerable<MempoolBehavior> behaviours = peers.Select(s => s.Behavior<MempoolBehavior>());
-            this.logger.LogTrace("(-)");
-            return this.manager.MempoolLock.WriteAsync(() =>
+            return this.manager.MempoolLock.WriteAsync(() =>
```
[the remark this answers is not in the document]
dangershony: Ah but we need a global lock hmmm...

I suggest if we don't have a global lock anymore and we use `Hashset<>` then perhaps create a new method on each behaviour that will do the following code inside the lock lock.
```
 if (!mempoolBehavior.filterInventoryKnown.ContainsKey(hash))
    mempoolBehavior.inventoryTxToSend.TryAdd(hash, hash);
```

Also try to keep tracing outside of the locks.

dangershony:
I made an initial review but as @Aprogiena said perhaps a revisit of this logic is needed, my comments still apply though.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
mikedennis: If I remove the mempool manager locking here I had issues with all the peers getting dropped with "pong timeouts".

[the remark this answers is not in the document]
dangershony: Hu can you elaborate? 
If we need a video chat let me know.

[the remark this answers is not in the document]
dangershony: If you removed the global lock then you will get race conditions as each behaviour will try to do the same (it might be you got internal exceptions that where not logged) 🤔 

You have to either use a global lock (global to all behaviours) or lock each behaviour separately (move the code to a method and put a lock around it and call that method for each behaviour).   

if you use CD then you wont have an issue, but @Aprogiena might argue against CD even though I think this is quite a simple use of collections to justify it (unless you use hashset).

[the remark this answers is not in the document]
mikedennis: Ok that makes sense. I'll wait to see Apro's opinion and if he doesn't recommend something else go with the CD and see how that works. Thanks!

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -65,6 +65,11 @@ public class MempoolBehavior : NetworkPeerBehavior
         /// </summary>
         private readonly Dictionary<uint256, uint256> filterInventoryKnown;
 
+        /// <summary>
+        /// Locking object for memory pool behaviour.
+        /// </summary>
+        private readonly AsyncLock asyncLock = new AsyncLock();
```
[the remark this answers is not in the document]
noescape00: Yes, don't use asynclock, use `lock` keyword. 
Also please don't use dictionary/concurrentDictionary, use hashset

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
noescape00: var?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -508,14 +525,30 @@ public Task RelayTransaction(uint256 hash)
...
+            return this.manager.MempoolLock.WriteAsync(() => 
             {
                 foreach (MempoolBehavior mempoolBehavior in behaviours)
                 {
+                    this.logger.LogTrace("Attempting to relaying transaction ID '{0}' to peer '{1}'.", hash, mempoolBehavior?.AttachedPeer.RemoteSocketEndpoint);
                     if (mempoolBehavior?.AttachedPeer.PeerVersion.Relay ?? false)
-                        if (!mempoolBehavior.filterInventoryKnown.ContainsKey(hash))
-                            mempoolBehavior.inventoryTxToSend.TryAdd(hash, hash);
+                    {
+                        using (mempoolBehavior.asyncLock.Lock())
```
noescape00: Please don't use private members of other instances of the same class. 
Expose a public interface for handling such situations.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
noescape00: Leave it for now as you have it, it will be fixed in the mempool behavior refactoring later on. 

As mentioned before- use lock keyworkd and hashset for the inner datastructures.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -543,44 +576,55 @@ public Task RelayTransaction(uint256 hash)
...
+                if (!this.inventoryTxToSend.Keys.Any())
+                {
+                    this.logger.LogTrace("(-)[NO_TXS]");
+                    return;
+                }
             }
 
-            List<uint256> sends = await this.manager.MempoolLock.WriteAsync(() =>
-            {
+            List<uint256> sends = new List<uint256>();
```
noescape00: var?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -508,14 +525,30 @@ public Task RelayTransaction(uint256 hash)
...
                     if (mempoolBehavior?.AttachedPeer.PeerVersion.Relay ?? false)
-                        if (!mempoolBehavior.filterInventoryKnown.ContainsKey(hash))
-                            mempoolBehavior.inventoryTxToSend.TryAdd(hash, hash);
+                    {
+                        using (mempoolBehavior.asyncLock.Lock())
+                        {
+                            if (!mempoolBehavior.filterInventoryKnown.ContainsKey(hash))
+                            {
+                                mempoolBehavior.inventoryTxToSend.TryAdd(hash, hash);
+                                this.logger.LogTrace("Added transaction ID '{0}' to inventory to send for peer '{1}'.", hash, mempoolBehavior?.AttachedPeer.RemoteSocketEndpoint);
```
noescape00: `for peer` => `to peer`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -508,14 +525,30 @@ public Task RelayTransaction(uint256 hash)
...
+                            }
+                            else
+                            {
+                                this.logger.LogTrace("Transaction ID '{0}' already exists in inventory known filter for peer '{1}'.", hash, mempoolBehavior?.AttachedPeer.RemoteSocketEndpoint);
```
noescape00: `for peer` => `of peer`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -543,44 +576,55 @@ public Task RelayTransaction(uint256 hash)
 
             // before locking an exclusive task
             // check if there is anything to processes
-            if (!await this.manager.MempoolLock.ReadAsync(() => this.inventoryTxToSend.Keys.Any()))
-            {
-                this.logger.LogTrace("(-)[NO_TXS]");
-                return;
+            using (this.asyncLock.Lock())
```
noescape00: Please delete 2 comments above that line. 
Merge this lock block with the one after that.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -151,6 +150,7 @@ public class OrphanTx
         /// <returns>Whether the transaction id is present.</returns>
         public async Task<bool> AlreadyHaveAsync(uint256 trxid)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}'", nameof(trxid), trxid);
```
noescape00: Please rename mempoolLogger to logger

noescape00: Also change this line to 
`this.logger.LogTrace("({0}:'{1}')", nameof(trxid), trxid);`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -180,6 +183,7 @@ public class OrphanTx
         /// <param name="tx">The new transaction received.</param>
         public async Task ProcessesOrphansAsync(MempoolBehavior behavior, Transaction tx)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}', {2}:'{3}'", nameof(behavior), behavior?.AttachedPeer?.RemoteSocketEndpoint, nameof(tx), tx?.GetHash());
```
noescape00: `"({0}:'{1}',{2}:'{3}')"`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -159,11 +159,14 @@ public class OrphanTx
                     // might be now valid, e.g. due to a nLockTime'd tx becoming valid,
                     // or a double-spend. Reset the rejects filter and give those
                     // txs a second chance.
+                    this.mempoolLogger.LogTrace("Executing task to clear rejected transactions.");
                     this.hashRecentRejectsChainTip = this.chain.Tip.HashBlock;
                     this.recentRejects.Clear();
                 });
             }
 
+            this.mempoolLogger.LogTrace("(-)");
```
noescape00: Please log the exit value

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -259,15 +265,18 @@ public class OrphanTx
         /// <returns>Whether the transaction was added to orphans.</returns>
         public async Task<bool> ProcessesOrphansMissingInputsAsync(INetworkPeer from, Transaction tx)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}', {2}:'{3}'", nameof(from), from?.RemoteSocketEndpoint, nameof(tx), tx?.GetHash());
```
noescape00: `"{0}:'{1}', {2}:'{3}'"` => `"({0}:'{1}',{2}:'{3}')"`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
noescape00: Attacked peer and behavior can't be null here

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -259,15 +265,18 @@ public class OrphanTx
...
             // It may be the case that the orphans parents have all been rejected
             var rejectedParents = await this.MempoolLock.ReadAsync(() =>
             {
+                this.mempoolLogger.LogTrace("Executing task to query if orphans parents have all been rejected.");
```
noescape00: Please add exit log with the return value

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -259,15 +265,18 @@ public class OrphanTx
...
                 return tx.Inputs.Any(txin => this.recentRejects.ContainsKey(txin.PrevOut.Hash));
             });
 
             if (rejectedParents)
             {
                 this.mempoolLogger.LogInformation($"not keeping orphan with rejected parents {tx.GetHash()}");
+                this.mempoolLogger.LogTrace("(-):false[REJECT_PARENTS_ORPH]");
```
noescape00: `(-):false[REJECT_PARENTS_ORPH]` => `(-)[REJECT_PARENTS_ORPH]:false`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -299,6 +309,7 @@ public class OrphanTx
         /// <returns>The number of transactions evicted.</returns>
         public async Task<int> LimitOrphanTxSizeAsync(int maxOrphanTx)
         {
+            this.mempoolLogger.LogTrace("{0}:{1}", nameof(maxOrphanTx), maxOrphanTx);
```
noescape00: `{0}:{1}` => `({0}:{1})`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -347,11 +360,15 @@ public class OrphanTx
         /// <returns>Whether the orphan transaction was added.</returns>
         public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
         {
+            this.mempoolLogger.LogTrace("{0}:{1}, {2}:'{3}'", nameof(nodeId), nodeId, nameof(tx), tx?.GetHash());
```
noescape00: add () and remove space please

noescape00: tx can't be null here

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
noescape00: from and tx can't be null here

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -347,11 +360,15 @@ public class OrphanTx
...
             return this.MempoolLock.WriteAsync(() =>
             {
                 uint256 hash = tx.GetHash();
                 if (this.mapOrphanTransactions.ContainsKey(hash))
+                {
+                    this.mempoolLogger.LogTrace("(-):false[DUP_ORPH]");
```
noescape00: `false[DUP_ORPH]` => `[DUP_ORPH]false`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -364,6 +381,7 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
                 if (sz >= this.consensusValidator.ConsensusOptions.MaxStandardTxWeight)
                 {
                     this.mempoolLogger.LogInformation($"ignoring large orphan tx (size: {sz}, hash: {hash})");
+                    this.mempoolLogger.LogTrace("(-):false[LARGE_ORPH]");
```
noescape00: `false[LARGE_ORPH]` => `[LARGE_ORPH]false`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -347,11 +360,15 @@ public class OrphanTx
         /// <returns>Whether the orphan transaction was added.</returns>
         public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
         {
+            this.mempoolLogger.LogTrace("{0}:{1}, {2}:'{3}'", nameof(nodeId), nodeId, nameof(tx), tx?.GetHash());
             return this.MempoolLock.WriteAsync(() =>
```
noescape00: Please assign the result to bool, add exit log with this bool

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -402,11 +421,15 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
         /// <returns>Whether erased.</returns>
         private Task<bool> EraseOrphanTx(uint256 hash)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}'", nameof(hash), hash);
             return this.MempoolLock.WriteAsync(() =>
             {
                 OrphanTx it = this.mapOrphanTransactions.TryGet(hash);
                 if (it == null)
+                {
+                    this.mempoolLogger.LogTrace("(-):false[NOTFOUND_ORPH]");
```
noescape00: `false[NOTFOUND_ORPH]` => `[NOTFOUND_ORPH]false`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -402,11 +421,15 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
         /// <returns>Whether erased.</returns>
         private Task<bool> EraseOrphanTx(uint256 hash)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}'", nameof(hash), hash);
             return this.MempoolLock.WriteAsync(() =>
```
noescape00: exit log is missing

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -402,11 +421,15 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
         /// <returns>Whether erased.</returns>
         private Task<bool> EraseOrphanTx(uint256 hash)
         {
+            this.mempoolLogger.LogTrace("{0}:'{1}'", nameof(hash), hash);
```
noescape00: add ()

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -427,8 +452,11 @@ private Task<bool> EraseOrphanTx(uint256 hash)
         /// <param name="peer">Peer node id</param>
         public Task EraseOrphansFor(ulong peer)
         {
+            this.mempoolLogger.LogTrace("{0}:{1}", nameof(peer), peer);
+            this.mempoolLogger.LogTrace("(-)");
```
noescape00: please remove the line

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -427,8 +452,11 @@ private Task<bool> EraseOrphanTx(uint256 hash)
         /// <param name="peer">Peer node id</param>
         public Task EraseOrphansFor(ulong peer)
         {
+            this.mempoolLogger.LogTrace("{0}:{1}", nameof(peer), peer);
```
noescape00: add ()

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -195,6 +195,7 @@ public class MempoolValidator : IMempoolValidator
         /// <inheritdoc />
         public async Task<bool> AcceptToMemoryPoolWithTime(MempoolValidationState state, Transaction tx)
         {
+            this.logger.LogTrace("{0}:'{1}', {2}:'{3}'", nameof(state), state?.ToString(), nameof(tx), tx?.GetHash());
```
noescape00: add (), remove spaces

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -203,15 +204,18 @@ public class MempoolValidator : IMempoolValidator
                 //    BOOST_FOREACH(const uint256& hashTx, vHashTxToUncache)
                 //        pcoinsTip->Uncache(hashTx);
                 //}
+                this.logger.LogTrace("(-):true");
                 return true;
             }
             catch (MempoolErrorException)
             {
+                this.logger.LogTrace("(-):false[MEMPOOL_EXCEPTION]");
```
noescape00: false[MEMPOOL_EXCEPTION] => [MEMPOOL_EXCEPTION]false

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -203,15 +204,18 @@ public class MempoolValidator : IMempoolValidator
...
                 return false;
             }
             catch (ConsensusErrorException consensusError)
             {
                 state.Error = new MempoolError(consensusError.ConsensusError);
+                this.logger.LogTrace("(-):false[CONSENSUS_EXCEPTION]");
```
noescape00: [CONSENSUS_EXCEPTION]false

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
dangershony: What do you mean leave it for now? using the mempool lock?
I prefer if we did not leave it and try to fix it in this PR.

I suggest to create new method on the behavior that will lock the hashset

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
dangershony: Why not? this is a very general suggestion you make @noescape00 stick to the review.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
dangershony: I think previously AttachedPeer could be null, not sure this is not the case still.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
dangershony: While you are here rename this ugly `it` param.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
mikedennis: Not sure this logging makes any sense 😁 I think it's safely covered by the block just below it, so just got rid of it

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
mikedennis: This is returns a task so exit may have not occurred yet.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
mikedennis: This is also returning a task.

mikedennis:
Don't review new changes yet - I still need to move to hashsets and figure out to do with the lock that's still being debated.

mikedennis:
I changed the collections from dictionary's to hashsets. So I think that I'm pretty much caught up on changes suggested other than the one last remaining mempool manager lock. Let me know if you guys think we should change that in this PR as well.

Going to run my nodes for a while and see if anything breaks but so far so good. My bitcoin node has already accepted several mempool tx's so at least the basics seem to be working.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
Aprogiena: The suggestion is to leave that one use of the original lock and attached as Mike reported that there is some problem with that. There are 20 other problems with this code and it is out of scope of this PR to fix them. You would basically need to completely rewrite the whole thing. I believe we can take this small PR which improves the situation slightly but also leaves many problems unresolved.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
Aprogiena: It cannot be null, the network peer refactoring that happened a while ago guarantees that all callbacks are finished before detaching.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
Aprogiena: Yes that is why you create a variable for the result and you will assign the task to it, only then you create exit log and only then you return from the method.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -88,7 +88,7 @@ public class MempoolBehavior : NetworkPeerBehavior
             IInitialBlockDownloadState initialBlockDownloadState,
             Signals.Signals signals,
             ILogger logger)
-        {
+        {
```
Aprogiena: Don't need extra spaces here.

Aprogiena:
As mentioned above, I don't think the global lock should be resolved by this PR.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
dangershony: Indeed there are but I want to get rid of the scheduler lock and I believe this can be done, so @mikedennis can you please attempt to fix this as I described earlier? (each behaviour will do its own locking on its dictionaries) and see if you still get nodes dropping.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
dangershony: ok yes so it cannot be null remove the `.?`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -298,20 +304,24 @@ public override object Clone()
             //  filterrate = pto->minFeeFilter;
             //}
 
-            List<TxMempoolInfo> sends = await this.manager.MempoolLock.WriteAsync(() =>
+            var sends = new List<TxMempoolInfo>();
+            lock (this.lockObject)
```
dangershony: Perhaps this could be done with minimal time inside the lock?
You can do the logic in the foreach outside of the lock (say with local lists) and the also do the tracing, then inside the lock just copy/remove from the shared resource.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -375,11 +385,14 @@ public override object Clone()
...
             // add to known inventory
-            await this.manager.MempoolLock.WriteAsync(() =>
+            lock (this.lockObject)
             {
                 foreach (InventoryVector inventoryVector in send.Inventory)
-                    this.filterInventoryKnown.TryAdd(inventoryVector.Hash, inventoryVector.Hash);
-            });
+                {
+                    this.filterInventoryKnown.Add(inventoryVector.Hash);
+                    this.logger.LogTrace("Added inventory transaction ID '{0}' to known inventory filter.", inventoryVector.Hash);
```
dangershony: Try to avoid tracing inside a lock, is this trace really needed?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -515,14 +532,30 @@ public Task RelayTransaction(uint256 hash)
...
-            return this.manager.MempoolLock.WriteAsync(() =>
+            return this.manager.MempoolLock.WriteAsync(() => 
             {
                 foreach (MempoolBehavior mempoolBehavior in behaviours)
                 {
+                    this.logger.LogTrace("Attempting to relaying transaction ID '{0}' to peer '{1}'.", hash, mempoolBehavior?.AttachedPeer.RemoteSocketEndpoint);
                     if (mempoolBehavior?.AttachedPeer.PeerVersion.Relay ?? false)
-                        if (!mempoolBehavior.filterInventoryKnown.ContainsKey(hash))
-                            mempoolBehavior.inventoryTxToSend.TryAdd(hash, hash);
+                    {
```
dangershony: All the code form this ` if (mempoolBehavior?.AttachedPeer.PeerVersion.Relay ?? false)` should go in its own private method.
`mempoolBehavior.TryRelay()` then you can use the lock locally.

If this current behaviours does not drop nodes then all is ok.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -151,17 +152,21 @@ public class OrphanTx
...
                     this.hashRecentRejectsChainTip = this.chain.Tip.HashBlock;
                     this.recentRejects.Clear();
                 });
-            }
+            }            
 
             // Use pcoinsTip->HaveCoinsInCache as a quick approximation to exclude
             // requesting or processing some txs which have already been included in a block
-            return await this.MempoolLock.ReadAsync(() =>
+            bool isTxPresent = await this.MempoolLock.ReadAsync(() =>
```
dangershony: I think you only need the mempool lock for this `this.memPool.Exists(trxid)` and this should be available on the manager.

dangershony:
One of the main reasons for this PR is to not use the mempool lock other then mempool operations and we still are, so please either do it in this PR or a following PR immediately after up to you @mikedennis

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
mikedennis: I'll see if I can do some tracing outside so as to not slow down the lock.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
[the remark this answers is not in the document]
mikedennis: I will give it another try and see if I can get rid of this one as well with the strategy you suggested. Thanks!

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
Aprogiena: I don't think this is a good advice. It is perfectly fine to have the logger inside the lock because it cost you exactly nothing when the logging is not enabled.

Aprogiena:
In that case it should be rewritten completely. I don't think this PR can solve that.

mikedennis:
I've addressed the remaining code review issues. I also have removed that last mempool lock and things seem to be hanging together. I've been running on bitcoin testnet for several hours and haven't observed any strange behavior or nodes dropping.

dangershony:
@mikedennis did you sync all the way to the tip and observer transactions going on to the mempool? (just confirming)

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
dangershony: It's still bad practice and you should avoid it if you can, especially when the lock is used for quick dictionary access.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -438,13 +451,17 @@ public override object Clone()
             uint256 trxHash = trx.GetHash();
 
             // add to local filter
-            await this.manager.MempoolLock.WriteAsync(() => this.filterInventoryKnown.TryAdd(trxHash, trxHash));
+            lock (this.lockObject)
+            {
+                this.filterInventoryKnown.Add(trxHash);
+                this.logger.LogTrace("Added transaction ID '{0}' to known inventory filter.", trxHash);
```
dangershony: Trace can go one line outside of the lock.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -151,17 +151,20 @@ public class OrphanTx
...
                 });
             }
 
             // Use pcoinsTip->HaveCoinsInCache as a quick approximation to exclude
             // requesting or processing some txs which have already been included in a block
-            return await this.MempoolLock.ReadAsync(() =>
-                                this.recentRejects.ContainsKey(trxid) ||
-                                this.memPool.Exists(trxid) ||
-                                this.mapOrphanTransactions.ContainsKey(trxid));
+            bool isTxPresent =  this.recentRejects.ContainsKey(trxid) ||
```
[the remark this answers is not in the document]
mikedennis: Just to confirm - add a local lock for the recenRejects,mapOrphanTransactions collections?

mikedennis:
@dangershony Yes I ran fully synced on both stratis and bitcoin testsnets. Not much for mempool txs on stratis but did see a couple go through, but lots of action on bitcoin. Things seem to be behaving.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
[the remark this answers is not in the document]
mikedennis: Ok - I've done that. We may want to review the locking in general in mempool orphans as there may be other instances that can use the local lock instead of the mempool manager one. Next PR perhaps?

dangershony:
Ok great I will try to run the networks myself today.

dangershony:
Hi @mikedennis I made some changes to help you out, can you run the latest version instead (and tests)?

mikedennis:
@dangershony thanks for the help!

I reviewed your additions and fixed up a couple of minor typos, but nothing that should change any functionality. I also ran synced bitcoin testnet node for a couple of hours and it seems to be accepting tx to memory pool and adding orphans etc with no nodes dropping or anything. Looking good!

I ran the integration tests and no new failures (only the ones that always fail for me). I've also started up my stratis testnet node but haven't gotten to the point of having any txs yet so haven't been able to verify it yet but it also is behaving as expected so far.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -473,7 +429,7 @@ public override object Clone()
             }
             else if (state.MissingInputs)
             {
-                await this.orphans.ProcessesOrphansMissingInputsAsync(peer, trx).ConfigureAwait(false);
+                this.orphans.ProcessesOrphansMissingInputsAsync(peer, trx);
```
Aprogiena: You can't do that, it violates our shutdown sequence

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -611,32 +572,36 @@ public void RelayTransaction(uint256 hash)
                         this.logger.LogTrace("Transaction ID '{0}' not added to inventory list, exists in known inventory filter.", hash);
                         continue;
                     }
-
-                    // Not in the mempool anymore? don't bother sending it.
```
Aprogiena: Is it okay to remove this code? Why?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -611,32 +572,36 @@ public void RelayTransaction(uint256 hash)
...
 
                 this.logger.LogTrace("Transaction inventory list created.");
             }
 
-            if (sends.Any())
+            List<uint256> findInMempool = transactionsToSend.ToList();
+            foreach (uint256 hash in findInMempool)
+            {
+                // Not in the mempool anymore? don't bother sending it.
+                TxMempoolInfo txInfo = await this.mempoolManager.InfoAsync(hash);
```
Aprogiena: Configure await?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -265,14 +274,16 @@ public class OrphanTx
         /// <param name="from">Source node for transaction.</param>
         /// <param name="tx">Transaction to add.</param>
         /// <returns>Whether the transaction was added to orphans.</returns>
-        public async Task<bool> ProcessesOrphansMissingInputsAsync(INetworkPeer from, Transaction tx)
+        public bool ProcessesOrphansMissingInputsAsync(INetworkPeer from, Transaction tx)
```
Aprogiena: not async anymore

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
dangershony: AH yes ops

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
dangershony: No it was async only for the mempool lock, other then that nothing else needed asyncness

Aprogiena: That was not question. Remove the suffix

dangershony:
Hi @mikedennis please pull again the PR I made some more changes so the orphan stats are correct.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
dangershony: Done

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -124,6 +124,34 @@ public class OrphanTx
...
+        }
+
+        /// <summary>
+        /// Remove transactions form the orphan list.
+        /// </summary>
+        public void RemoveForBlock(List<Transaction> transactionsToRemove)
+        {
+            lock (this.lockObject)
+            {
+                lock (this.lockObject)
```
mikedennis: Is this right? two locks

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -258,7 +258,7 @@ public override object Clone()
             //  return true;
             //}
 
-            List<TxMempoolInfo> transactionsInMempol = await this.mempoolManager.InfoAllAsync();
+            List<TxMempoolInfo> transactionsInMempol = await this.mempoolManager.InfoAllAsync().ConfigureAwait(false);
```
mikedennis: transactionsInMempool?

mikedennis:
@dangershony Ran the node with new changes - didn't see any issues. Also ran all integrations tests again. They are also passing.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
Aprogiena: No.

dangershony: ops

dangershony:
OK we need another candidate to run networks and tests (review is high)

mikedennis:
It might have been doing this before but I noticed lots of orphans when syncing:
``` info: Stratis.Bitcoin.FullNode[0]
      ======Node stats====== 06/07/2018 23:56:52 agent StratisBitcoin:1.1.7
      Headers.Height:      1324552  Headers.Hash:     0000000000037c4fd2349de326eef290d7756f215bc18af84671f6812ab0d3e1
      BlockStore.Height:   1324436  BlockStore.Hash:  0000000000000129e00571f3dcc3d13089d9c332243859b5434c4970e91aaba2
      Consensus.Height:    1324442  Consensus.Hash:   000000000000010f14d5417b09233e9cc112ced6d74104f1dc7ba815c625497e
      Wallet.Height:       No Wallet

      ======Mempool======
      MempoolSize: 21   DynamicSize: 36 kb  OrphanSize: 49

```

dangershony:
Yes thats fine, before the orphans where not correctly displayed so it was always zero.

If the orphans number is changed (up/down) then I beleive it ok.

@noescape00 will sync netowrks as well (Ivan most important is to run on a fully downloaded chain to see mempool trx) so less imporatnt to sync from genesis.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
noescape00: This wasn't fixed

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -133,7 +118,51 @@ public class OrphanTx
         /// <returns>A list of orphan transactions.</returns>
         public List<OrphanTx> OrphansList() // for testing
```
noescape00: maybe make it internal if it's for tests only?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -380,55 +473,81 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
...
+                    if (erase.NodeId == peer)
+                    {
+                        erased += this.EraseOrphanTxLock(erase.Tx.GetHash()) ? 1 : 0;
+                    }
+                }
 
                 if (erased > 0)
-                    this.mempoolLogger.LogInformation($"Erased {erased} orphan tx from peer {peer}");
+                    this.logger.LogInformation("Erased {0} orphan tx from peer {1}", erased, peer);
+            };
```
noescape00: redundant `;`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -380,55 +473,81 @@ public Task<bool> AddOrphanTx(ulong nodeId, Transaction tx)
...
+            this.logger.LogTrace("(-):true");
+            return true;
         }
 
         /// <summary>
         /// Erase all orphans for a specific peer node.
         /// </summary>
         /// <param name="peer">Peer node id</param>
-        public Task EraseOrphansFor(ulong peer)
+        public void EraseOrphansFor(ulong peer)
```
noescape00: peer => peerId

dangershony:
Thanks @noescape00

PR #1451, opened 2018-05-24
Guard improvements

### LINKED ISSUE #1450: Guard improvements
We have numerous lines like this in our codebase:

```
```

First parameter is the variable and the 2nd is it's name. Variable name is used ONLY when guard throws. 

So the point is to introduce an override of the NotNull that will do exactly the same but accept only 1 value- the variable. 

Example snippet: 

```
static void Main()
{
    string variableThatIsNull = null;

    try
    {
        Guard.NotNull(variableThatIsNull);
    }
    catch (Exception e)
    {
        Console.WriteLine(e);
    }
            
    Console.Read();
}

public static class Guard
{
    public static void NotNull(object value)
    {
        // throw if the value is null
        if (ReferenceEquals(value, null))
        {
            string name = GetOriginalVariableName(value);
            throw new ArgumentNullException(name);
        }
    }

    public static string GetOriginalVariableName(object obj)
    {
        StackFrame stackFrame = new StackTrace(true).GetFrame(2);
        var file = new StreamReader(stackFrame.GetFileName());
        for (int i = 0; i < stackFrame.GetFileLineNumber() - 1; i++)
            file.ReadLine();
        string varName = file.ReadLine().Split(new char[] { '(', ')' })[1];
        return varName;
    }
}
```

That will save us some time writing those `nameof(something)` all the time.

noescape00:
Only `Guard` and `GuardTests` have some new code that needs to be reviewed

Fix #1450

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
@@ -66,8 +66,8 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// <param name="loggerFactory">Factory to be used to create logger for the puller.</param>
         public DBreezeCoinView(Network network, string folder, IDateTimeProvider dateTimeProvider, ILoggerFactory loggerFactory)
         {
-            Guard.NotNull(network, nameof(network));
-            Guard.NotEmpty(folder, nameof(folder));
+            Guard.NotNull(network);
+            Guard.NotEmpty(folder);
 
             // Create the coinview folder if it does not exist.
             Directory.CreateDirectory(folder);
@@ -391,6 +391,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
             Task task = Task.Run(() =>
             {
                 this.logger.LogTrace("({0}.Count():{1})", nameof(blocklist), blocklist.Count());
+                this.logger.LogTrace("({0}.Count():{1})");
 
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
@@ -391,7 +391,6 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
             Task task = Task.Run(() =>
             {
                 this.logger.LogTrace("({0}.Count():{1})", nameof(blocklist), blocklist.Count());
-                this.logger.LogTrace("({0}.Count():{1})");
 
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
--- a/src/Stratis.Bitcoin/Utilities/Guard.cs
+++ b/src/Stratis.Bitcoin/Utilities/Guard.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Diagnostics;
+using System.IO;
 
 namespace Stratis.Bitcoin.Utilities
 {
@@ -47,6 +49,21 @@ namespace Stratis.Bitcoin.Utilities
         }
 
         /// <summary>
+        /// Checks an object is not null.
+        /// </summary>
+        /// <param name="value">The object.</param>
+        /// <exception cref="ArgumentNullException">An exception if the object passed is null.</exception>
+        public static void NotNull(object value)
+        {
+            // throw if the value is null
+            if (ReferenceEquals(value, null))
+            {
+                string name = GetOriginalVariableName(value);
+                throw new ArgumentNullException(name);
+            }
+        }
+
+        /// <summary>
         /// Checks a <see cref="string"/> is not null or empty.
         /// </summary>
         /// <param name="value">The string to check.</param>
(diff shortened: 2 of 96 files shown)


src/Stratis.Bitcoin/Utilities/Guard.cs
```diff
@@ -46,6 +48,23 @@ public static T NotNull<T>(T value, string parameterName)
...
 
+        /// <summary>
+        /// Checks an object is not null.
+        /// </summary>
+        /// <param name="value">The object.</param>
+        /// <exception cref="ArgumentNullException">An exception if the object passed is null.</exception>
+        public static T NotNull<T>(T value)
+        {
+            // throw if the value is null
+            if (ReferenceEquals(value, null))
```
majikandy: did you do this just incase == is overloaded?  ie rather than `value == null` ?

src/Stratis.Bitcoin/Utilities/Guard.cs
```diff
@@ -63,5 +82,38 @@ public static string NotEmpty(string value, string parameterName)
...
+        }
+
+        /// <summary>
+        /// Gets the name of the original variable.
+        /// </summary>
+        /// <param name="obj">The variable to get the name of.</param>
+        private static string GetOriginalVariableName(object obj)
+        {
+            StackFrame stackFrame = new StackTrace(true).GetFrame(2);
+            var file = new StreamReader(stackFrame.GetFileName());
```
majikandy: do you need to dispose this StreamReader?  ie put in a using?

src/Stratis.Bitcoin/Utilities/Guard.cs
```diff
@@ -63,5 +82,38 @@ public static string NotEmpty(string value, string parameterName)
...
+                string name = GetOriginalVariableName(value);
+                throw new ArgumentException($"The string parameter {name} cannot be empty.");
+            }
+        }
+
+        /// <summary>
+        /// Gets the name of the original variable.
+        /// </summary>
+        /// <param name="obj">The variable to get the name of.</param>
+        private static string GetOriginalVariableName(object obj)
```
majikandy: This appears to be cumbersome work just to get the variable name when ArgumentNullException with a magic string is quite well handled by the IDE and even suggests you change it to nameof

screenshot 2018-05-24 21 03 02

Especially as this Guard method is called a lot in the code, is this too much and unnecessary just to get a slightly better looking Guard at the top of the methods.

Or are there some other advantages that make it worth it ?

src/Stratis.Bitcoin/Utilities/Guard.cs
```diff
@@ -63,5 +82,38 @@ public static string NotEmpty(string value, string parameterName)
...
+        {
+            StackFrame stackFrame = new StackTrace(true).GetFrame(2);
+            var file = new StreamReader(stackFrame.GetFileName());
+            for (int i = 0; i < stackFrame.GetFileLineNumber() - 1; i++)
+                file.ReadLine();
+            string varName = file.ReadLine().Split(new char[] { '(', ')' })[1];
+            return varName;
```
majikandy: are we allowed to use var in the name here ? 😺

src/Stratis.Bitcoin/Utilities/Guard.cs
noescape00: To keep it consistent with the other overload that uses `ReferenceEquals`

src/Stratis.Bitcoin/Utilities/Guard.cs
noescape00: Yeah, good catch. 

Fixed

majikandy:
Ah ok I was missing something,  it only happens when it is null,  which is not very often.  So ok yes the cleaner looking code is worth it.

However, even better if we can remove them completely

src/Stratis.Bitcoin/Utilities/Guard.cs
noescape00: Lol.
But changed to `name`

src/Stratis.Bitcoin/Utilities/Guard.cs
noescape00: >and even suggests you change it to nameof

what? I'm not sure what you've meant by that example.

>is this too much and unnecessary just to get a slightly better looking Guard at the top of the methods.

Yep, all of that to get better looking Guard. 
Also what do you mean by "too much"? No additional CPU is being consumed.

majikandy: Yes, I made a mistake on that one, I saw the method that got the variable name and made the assumption it was called every time. Apologies.

So yes, I do agree with it looking cleaner at the top of the method and therefore worth it.

noescape00:
>Ah ok I was missing something, it only happens when it is null, which is not very often.

If something is null and we use Guard we usually just stop the whole node because we never handle exceptions thrown by Guard intentionally.

So doing all that nasty stuff is justified- that code is never executed unless we are about to throw and shut down the node.

majikandy:
>So doing all that nasty stuff is justified- that code is never executed unless we are about to throw and shut down the node.

Yep, I agree with that now. Tired brain.

noescape00:
>Yep, I agree with that now. Tired brain.

No problem, mate 🙂

src/Stratis.Bitcoin/Utilities/Guard.cs
bokobza: Yep, exactly that @majikandy . We have many overloaded == operators in the code and we can't guarantee that they handle null correctly.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -391,6 +391,7 @@ public Task GetStakeAsync(IEnumerable<StakeItem> blocklist)
             Task task = Task.Run(() =>
             {
                 this.logger.LogTrace("({0}.Count():{1})", nameof(blocklist), blocklist.Count());
+                this.logger.LogTrace("({0}.Count():{1})");
```
bokobza: is this intentional?

Aprogiena:
I have decided for level medium because it touches so many files

noescape00:
Ok, it turns out getting a variable name from a stack trace is not the best idea- there is no guarantee that the stack trace will be the same on different platforms and build modes. 
When you compile in release mode, the compiler will remove certain methods and insert their content into all of the call sites (called inlining).

See: https://stackoverflow.com/questions/7585010/exception-stack-trace-difference-between-debug-and-release-mode
See: http://bartdesmet.net/blogs/bart/archive/2007/02/19/inlining-yes-it-happens.aspx

So that might lead to trying to get a wrong stack trace frame that actually doesn't contain the variable name. 

Therefore closing the issue: implemented approach is not reliable when compiling for release.

PR #1459, opened 2018-05-30
[RF][SmartContracts] Create CoinviewRule base class

fassadlr:
Currently `PosCoinviewRule` inherits and overrides some members and/or methods of `PowCoinviewRule`. As these two classes basically share a lot of logic I want to introduce a base abstract class which holds this logic. Then instead of having virtual overrides for methods like `GetProofOfWork` we implement concrete implementations for each network type, be it Pos or Pow. 

The current implementation breaks some SOLID principles like Liskov as if you had to remove `PosCoinviewRule`'s override of `GetProofOfWorkReward` the node wouldn't certain be concerned otherwise. We need to be more explicit as to each implementation.

As we also require SmartContractCoinviewRule we cannot currently implement this due to the `PosCoinviewRule` inheriting from `PowCoinviewRule` approach.

**There are NO logic changes in this PR but instead only the shared logic was moved to `CoinviewRule`.**

**I confirm a full Testnet sync and staking on Testnet works.**

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
@@ -1,0 +1,408 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Base.Deployments;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
+{
+    public abstract class CoinviewRule : ConsensusRule
+    {
+        /// <summary>Consensus options.</summary>
+        public PowConsensusOptions PowConsensusOptions { get; private set; }
+
+        public override void Initialize()
+        {
+            this.Logger.LogTrace("()");
+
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -0,0 +1,408 @@
...
+            Money fees = Money.Zero;
+            var checkInputs = new List<Task<bool>>();
+            for (int txIndex = 0; txIndex < block.Transactions.Count; txIndex++)
+            {
+                this.Parent.PerformanceCounter.AddProcessedTransactions(1);
+                Transaction tx = block.Transactions[txIndex];
+                if (!context.SkipValidation)
+                {
+                    // TODO: Simplify this condition.
+                    if (!tx.IsCoinBase && (!context.IsPoS || (context.IsPoS && !tx.IsCoinStake)))
```
dangershony: If we already have two implementations this logic should be an abstract in the core class (so we dont have to mention POS in a POW network).

fassadlr: Ah yes @dangershony well spotted! I didn't want to make any logic changes in this PR but this is definitely something that we can look at in a PR after this one 👍

fassadlr:
Closing this as will redo for easier reviewing.

PR #1464, opened 2018-05-31
[RF][SmartContracts] Create CoinviewRule base class

fassadlr:
[RF][SmartContracts] Create CoinviewRule base class

**This PR should be reviewed commit by commit.**

Currently `PosCoinviewRule` inherits and overrides some members and/or methods of `PowCoinviewRule`. As these two classes basically share a lot of logic I want to introduce a base abstract class which holds this logic. Then instead of having virtual overrides for methods like `GetProofOfWork` we implement concrete implementations for each network type, be it Pos or Pow. 

The current implementation breaks some SOLID principles like Liskov as if you had to remove `PosCoinviewRule`'s override of `GetProofOfWorkReward` the node wouldn't certain be concerned otherwise. We need to be more explicit as to each implementation.

As we also require SmartContractCoinviewRule we cannot currently implement this due to the `PosCoinviewRule` inheriting from `PowCoinviewRule` approach.

**There are NO logic changes in this PR but instead only the shared logic was moved to `CoinviewRule`.**

**I confirm a full Testnet sync and staking on Testnet works.**


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -98,7 +96,7 @@ protected override void UpdateCoinView(RuleContext context, Transaction transact
         }
 
         /// <inheritdoc />
-        protected override void CheckMaturity(UnspentOutputs coins, int spendHeight)
+        public override void OnCheckMaturity(UnspentOutputs coins, int spendHeight)
```
dangershony: I wonder why add the `On` keyword, On would indicate to me that an even happened (like OnClick)

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -267,18 +267,7 @@ public virtual void CheckInputs(Transaction transaction, UnspentOutputSet inputs
...
-            // Force block reward to zero when right shift is undefined.
-            if (halvings >= 64)
-                return 0;
-
-            Money subsidy = this.consensusOptions.ProofOfWorkReward;
-            // Subsidy is cut in half every 210,000 blocks which will occur approximately every 4 years.
-            subsidy >>= halvings;
-            return subsidy;
-        }
+        public abstract Money GetProofOfWorkReward(int height);
```
dangershony: Rename to `GetBlockReward`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
fassadlr: I have now removed these :+1

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -0,0 +1,450 @@
...
+namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
+{
+    /// <summary>
+    /// The proof of work coinview update rules. Validates the UTXO set is correctly spent and creating new outputs.
+    /// </summary>
+    /// <exception cref="ConsensusErrors.BadTransactionMissingInput">Thrown if transaction tries to spend inputs that are missing.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionNonFinal">Thrown if transaction's height or time is lower then provided by SequenceLock for this block.</exception>
+    /// <exception cref="ConsensusErrors.BadBlockSigOps">Thrown if signature operation cost is greater then maximum block signature operation cost.</exception>
+    /// <exception cref="ConsensusErrors.BadTransactionScriptError">Thrown if not all inputs are valid (no double spends, scripts & sigs, amounts).</exception>
+    [ExecutionRule]
```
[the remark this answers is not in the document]
fassadlr: Ah yes this needs to be removed and added to PowCoinviewRule.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
fassadlr: Done

fassadlr:
@dangershony I confirm that I have synced all the network chainz and that staking is working as per normal.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -287,7 +287,7 @@ public IEnumerable<ConsensusRule> GetRules()
                     // rules that require the store to be loaded (coinview)
                     new LoadCoinviewRule(),
                     new TransactionDuplicationActivationRule(), // implements BIP30
-                    new PowCoinViewRule() // implements BIP68, MaxSigOps and BlockReward calculation
+                    new PowCoinviewRule() // implements BIP68, MaxSigOps and BlockReward calculation
```
dangershony: You might want to also rename `PosCoinViewRule`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -0,0 +1,442 @@
...
+            return (int)bms.Counter.WrittenBytes;
+        }
+
+        /// <summary>
+        /// Calculates merkle root for witness data.
+        /// </summary>
+        /// <param name="block">Block which transactions witness data is used for calculation.</param>
+        /// <param name="mutated"><c>true</c> if at least one leaf of the merkle tree has the same hash as any subtree. Otherwise: <c>false</c>.</param>
+        /// <returns>Merkle root.</returns>
+        public uint256 BlockWitnessMerkleRoot(Block block, out bool mutated)
```
dangershony: This method is not used maybe delete it

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -0,0 +1,442 @@
...
+            return BlockMerkleRootRule.ComputeMerkleRoot(leaves, out mutated);
+        }
+
+        /// <summary>
+        /// Calculates merkle root for block's trasnactions.
+        /// </summary>
+        /// <param name="block">Block which transactions are used for calculation.</param>
+        /// <param name="mutated"><c>true</c> if block contains repeating sequences of transactions without affecting the merkle root of a block. Otherwise: <c>false</c>.</param>
+        /// <returns>Merkle root.</returns>
+        public uint256 BlockMerkleRoot(Block block, out bool mutated)
```
dangershony: This method is not used maybe delete it

dangershony:
Looks good, please just rename back to `GetProofOfWorkReward` my mistake and we are good.
I will sync networks now.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
fassadlr: Done

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
fassadlr: Done

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
fassadlr: Done

PR #1465, opened 2018-05-31
Small Network changes

bokobza:
Just a small PR so that the next PR on Network will be more foicused and easier to digest for reviewer.

* Moved DNSSeedData and Consensus to their own classes.
* Made Network.AlertPubKey into a property
* Make it possible to specify a ConsensusFactory when getting a block hash.
Also has the ince side effect of avoiding to init Bitcoin Main everytime we init another Network
* Removed unused `parent` object in BuriedDeploymentsArray and BIP9DeploymentsArray


src/NBitcoin/Networks.cs
```diff
@@ -170,7 +170,7 @@ private static Network InitTest()
                 RootFolderName = BitcoinRootFolderName,
                 DefaultConfigFilename = BitcoinDefaultConfigFilename,
                 Magic = 0x0709110B,
-                alertPubKeyArray = Encoders.Hex.DecodeData("04302390343f91cc401d56d68b123028bf52e5fca1939df127f63c6467cdf9c8e2c14b61104cf817d0b780da337893ecc4aaff1309e536162dabbdb45200ca2b0a"),
```
monsieurleberre: I can't see the alertPubKey setting in all the other networks constructors, is it on purpose ?

bokobza: yeah only bitcoin main and bitcoin test have it.

src/NBitcoin.Tests/NetworkTests.cs
```diff
@@ -76,6 +66,10 @@ public void BitcoinMainnetIsInitializedCorrectly()
         {
             Network network = Network.Main;
 
+            Assert.Equal(15, network.Checkpoints.Count);
```
noescape00: I'm not sure if assert.Equal is the right approach here. 
In case we add a new valid checkpoint the test will fail. 

I guess the purpose of the test is to make sure that checkpoints were loaded correctly and none were removed so maybe here and in another similar asserts it will be a good idea to check if the number of checkpoints is less than what we have now? If it's greater- new checkpoint was added and we shouldn't fail the test.

src/NBitcoin/Block.cs
```diff
@@ -106,6 +106,11 @@ public virtual void ReadWrite(BitcoinStream stream)
         #endregion
 
         public virtual uint256 GetHash()
+        {
+            return this.GetHash(null);
```
noescape00: If GetHash's input param is nullable why not just have 

```
public virtual uint256 GetHash(ConsensusFactory consensusFactory = null)
```

and from `GetHash()` call that overload?

src/NBitcoin/Block.cs
```diff
@@ -280,6 +285,12 @@ public uint256 GetHash()
             return this.header.GetHash();
         }
 
+        public uint256 GetHash(ConsensusFactory consensusFactory)
+        {
+            // Block's hash is his header's hash.
```
noescape00: his => its

src/NBitcoin/BlockStake.cs
```diff
@@ -220,12 +226,12 @@ public override uint256 GetHash()
         }
 
         /// <summary>
-        /// Generate a has based on the X13 algorithms.
+        /// Generate a hash based on the X13 algorithms.
         /// </summary>
-        /// <returns></returns>
-        public override uint256 GetPoWHash()
+        /// <returns>A hash</returns>
```
noescape00: `.` is missing

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System;
+using NBitcoin.BouncyCastle.Math;
+
+namespace NBitcoin
+{
+    public enum BuriedDeployments : int
+    {
+        /// <summary>
+        /// Height in coinbase
```
noescape00: Dots in the summaries of this enum are missing.

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System;
+using NBitcoin.BouncyCastle.Math;
+
+namespace NBitcoin
+{
+    public enum BuriedDeployments : int
```
noescape00: ` : int` is redundant

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// </summary>
+        public class ConsensusOptions
+        {
+        }
+
+        public ConsensusOptions Options { get; set; }
+
+        public class BuriedDeploymentsArray
+        {
+            readonly int[] heights;
```
noescape00: missing `private` modifier

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            public BuriedDeploymentsArray()
+            {
+                this.heights = new int[Enum.GetValues(typeof(BuriedDeployments)).Length];
+            }
+
+            public int this[BuriedDeployments index]
```
noescape00:
```
public int this[BuriedDeployments index]
{
    get => this.heights[(int)index];
    set => this.heights[(int)index] = value;
}
```

probably looks nicer (but I don't have a strong point here so ignore if you like it how it is)

same for line 60 and 106

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                get { return this.heights[(int)index]; }
+                set { this.heights[(int)index] = value; }
+            }
+        }
+
+        public class BIP9DeploymentsArray
+        {
+            readonly BIP9DeploymentsParameters[] parameters;
```
noescape00: missing `private` modifier

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,168 @@
...
+            this.BuriedDeployments = new BuriedDeploymentsArray();
+            this.BIP9Deployments = new BIP9DeploymentsArray();
+
+            this.ConsensusFactory = new ConsensusFactory()
+            {
+                Consensus = this
+            };
+        }
+
+        public BuriedDeploymentsArray BuriedDeployments { get; set; }
```
noescape00: Please add comments for this and the following fields

bokobza:
So I've gone ahead and done it in this PR.
I split the networks to their own file.
Like this 
image


The Network class is now abstract and most of its properties have protected setters.
This way someone wanting to create a Network can do `public class JeremyChain : Network` and set up all the properties, or even do `public class JeremyChain : StratisMain` and only change a few properties.

src/NBitcoin/DNSSeedData.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Net;
+
+namespace NBitcoin
+{
+    public class DNSSeedData
```
noescape00: Please comment this class

src/NBitcoin.Tests/NetworkTests.cs
bokobza: yes I want the test to fail if you update checkpoints. Basically I want to make sure that you've 1. loaded checkpoints 2. loaded the right set of checkpoints.

src/NBitcoin/Network.cs
```diff
@@ -324,32 +108,45 @@ public byte[] MagicBytes
             }
         }
 
+        public uint GenesisTime { get; protected set; }
```
noescape00: Please comment those

src/NBitcoin.Tests/NetworkTests.cs
noescape00: Ok, makes sense. 

Just feels slightly weird that you can do a good change and it will fail the test.

noescape00:
In some of my comments I've requested adding comments- I know that this code was moved and it wasn't commented before but we have to start commenting it at some point. 

Network will be the first thing anyone will touch when they want to use our code to create a custom blockchain so it would make sense to at least have most difficult parts commented (mostly stuff in Consensus.cs is challenging)

src/NBitcoin/Block.cs
bokobza: Because there are some places in the code (in Moq methods) that are using GetHash() in an expression and expressions don't play well with optional arguments.
So I was receiving an error saying `An expression tree may not contain a call or invocation that uses optional arguments`.

To illustrate what this is, let me show you a small example:
```
void Main()
{
	MyClass myclass1 = new MyClass();	
	Console.WriteLine(myclass1.MethodWithExpression(c => myclass1.Method1()));		
}

class MyClass
{	
	public string Method1(string first = null)
	{
		return first;
	}

	public object MethodWithExpression(Expression<Func<string, string>> myExpression)
	{		
		return myExpression.Body;
	}
}

```
This piece of code will show the error I described above around `myclass1.Method1()`.
I think the reason why it's not supported is because since the Expression is just representing the lambda expression without compiling it, it doesn't want to deal with creating a tree that would accomodate all overloads (as the compiler would do for a normal method with optional arguments).

bokobza:
You're right @noescape00 , I should have put some comments there.
I wanted to move quickly and since this code has just been copied over, I didn't bother with documentation.
I'll sort it out now.

src/NBitcoin.Tests/NetworkTests.cs
bokobza: Not really: I'm just testing every property in the Network so any change made to it should also update the tests.

src/NBitcoin/Consensus.cs
bokobza: quite 👍

src/NBitcoin/Block.cs
bokobza: yep

src/NBitcoin/BlockStake.cs
bokobza: fixed

src/NBitcoin/Consensus.cs
bokobza: 👍

src/NBitcoin/Consensus.cs
bokobza: 😎

src/NBitcoin/Consensus.cs
bokobza: 💎

src/NBitcoin/Consensus.cs
bokobza: 🕵️

src/NBitcoin/Consensus.cs
dangershony: 🏄

src/NBitcoin.Tests/NetworkTests.cs
dangershony: To me we don't even need a test around checkpoints, its metadata of a network.
If someone is changing that its intentional and the test will obviously fail.
Same goes to other metadata of network (magic bytes, port)

monsieurleberre:
OK for me

src/NBitcoin/Networks.cs
dangershony: I think its deprecated in bitcoin core so not really needed.

src/NBitcoin/Consensus.cs
dangershony: I totally agree comments will be nice on each of the elements, however we might have time constraints on this PR (for sidechains).
So am happy to create an issue and do it as a separate PR.

src/NBitcoin/BlockStake.cs
```diff
@@ -195,6 +195,12 @@ public class PosBlockHeader : BlockHeader
 
         /// <inheritdoc />
         public override uint256 GetHash()
+        {
+            return this.GetHash(null);
+        }
+
+        /// <inheritdoc />
+        public override uint256 GetHash(ConsensusFactory consensusFactory)
```
dangershony: I think we should wait with this change for now its not necessarily for the Network changes.
Stratis header is the same as bitcoin so it will work fine and I would try to avoid passing the factory in the method.

This is probably a deeper changes that needs more thinking.
An idea is to pass the ConsensusFacotyr to the Block/BlockHeader/Transaction constructors (as they are anyway created by the factory.

Or move the GetHAsh/GetPowHash in to the factory itself.

src/NBitcoin/BlockStake.cs
```diff
@@ -220,12 +226,12 @@ public override uint256 GetHash()
         }
 
         /// <summary>
-        /// Generate a has based on the X13 algorithms.
+        /// Generate a hash based on the X13 algorithms.
         /// </summary>
-        /// <returns></returns>
-        public override uint256 GetPoWHash()
+        /// <returns>A hash</returns>
+        public override uint256 GetPoWHash(ConsensusFactory consensusFactory = null)
```
dangershony: Same goes for this change.

src/NBitcoin/Network.cs
```diff
@@ -324,32 +108,45 @@ public byte[] MagicBytes
...
+
+        public uint GenesisNonce { get; protected set; }
```
dangershony: This fields are probably not required to be part of network
GenesisNonce /GenesisBits/GenesisVersion /GenesisReward

src/NBitcoin/Networks/StratisTest.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace NBitcoin.Networks
+{
+    public class StratisTest : StratisMain
+    {
+        public StratisTest()
+        {
+            // The message start string is designed to be unlikely to occur in normal data.
+            // The characters are rarely used upper ASCII, not valid as UTF-8, and produce
+            // a large 4-byte int at any alignment.
+            var messageStart = new byte[4];
```
dangershony: We should maybe make magic bytes simpler? we dont really need the byte array created in advance.

src/NBitcoin/BlockStake.cs
dangershony: I'll create an issue to address this.

src/NBitcoin/Consensus.cs
```diff
@@ -0,0 +1,169 @@
...
+        public TimeSpan PowTargetSpacing { get; set; }
+
+        public bool PowAllowMinDifficultyBlocks { get; set; }
+
+        public bool PowNoRetargeting { get; set; }
+
+        public uint256 HashGenesisBlock { get; set; }
+
+        /// <summary> The minimum amount of work the best chain should have. </summary>
+        public uint256 MinimumChainWork { get; set; }
```
noescape00: I've noticed that it's used only in IBD- if consensus chainwork is less- we are in IBD. 
What is the reason to have such a parameter? Is it useful?

dangershony: I guess we could take it out, it was here since forever (no harm though)

PR #1466, opened 2018-05-31
Chained tree part 2

dangershony:


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -12,9 +12,23 @@ namespace Stratis.Bitcoin.Consensus
     /// <summary>
     /// TODO comment this interface
     /// </summary>
-    public interface IChainedHeaderValidator
+    public interface IChainedHeaderTreeValidator
```
noescape00: Why renamed? New name suggests that tree is being validated

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -12,9 +12,23 @@ namespace Stratis.Bitcoin.Consensus
...
     {
-        void Validate(ChainedHeader chainedHeader);
+        /// <summary>
+        /// Validation of a header that was seen for the first time.
```
noescape00: `Validation of a header that was seen for the first time.`

So the header that was seen before can't be validated using this method?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -12,9 +12,23 @@ namespace Stratis.Bitcoin.Consensus
...
+        /// </summary>
+        /// <param name="chainedHeader">The chained header to be validated.</param>
+        void HeaderValidation(ChainedHeader chainedHeader);
```
noescape00: `ValidateHeader` maybe?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -12,9 +12,23 @@ namespace Stratis.Bitcoin.Consensus
...
+
+        /// <summary>
+        /// Validate the block signature (for POS) & merkle tree, throws if validation fails. 
+        /// </summary>
+        /// <remarks>
+        /// Throws different exceptions in case any of them failed.
+        /// </remarks>
+        /// <param name="block">The block that is going to be validated.</param>
+        /// <param name="chainedHeader">The chained header of the block that will be validated.</param>
```
noescape00: No xml doc for exceptions that can be thrown

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -118,6 +132,142 @@ public void Initialize(ChainedHeader consensusTip)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Remove a peer and the entire branch on the tree that it claims unless the headers are before our consensus tip or are claimed by other peers.
+        /// </summary>
+        /// <param name="networkPeerId">The peer id that is removed.</param>
```
noescape00: throws xml doc missing

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -118,6 +132,142 @@ public void Initialize(ChainedHeader consensusTip)
...
+            return availableBlocks;
+        }
+
+        public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeader), chainedHeader);
+
+            ChainedHeader claimedHeader = this.RemoveUnclaimedBranch(chainedHeader, true);
+
+            List<int> peersToBan = new List<int>();
```
noescape00: var

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: I don't know them yet ;)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Hmm good point, it was rename to suggest this is a validator that is used by the tree, however this validator might also validate blocks, but I get the confusion I'll revert.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: It could but it wont (according to the code), but the comment does no suggest a validated header cannot be validated again.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: yes

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -143,7 +143,7 @@ public void DisconnectPeer(int networkPeerId)
             if (!this.peerTipsByPeerId.TryGetValue(networkPeerId, out uint256 peerTipHash)) 
             {
                 this.logger.LogTrace("(-)[PEER_TIP_NOT_FOUND]");
-                throw new ConnectHeaderException();
+                return;
```
dangershony: Are we sure? last dicussion with @Aprogiena we concluded a peer must exist if Disconnect is called but we where not sure.

noescape00: This was done in case peer connects and disconnects right away. 
Another reason is in case partial or full validation fails and we remove the whole subtree so peerTipsByPeerId won't have a tip that is on an invalid chain. Old code would throw there.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -216,35 +214,55 @@ public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
...
         /// </summary>
-        /// <remarks>
-        /// The <see cref="discoveredPeers"/> will be modified in the processes.
-        /// </remarks>
-        /// <param name="discoveredPeers">List of peers that where discovered.</param>
-        /// <param name="chainedHeader">The chained header to start from.</param>
-        private void RemoveNextClaimsRecursive(List<int> discoveredPeers, ChainedHeader chainedHeader)
+        /// <param name="startHeader">The chained header to start from.</param>
+        /// <returns>List of peer Ids for banning.</returns>
+        private List<int> RemoveNextClaims(ChainedHeader startHeader)
```
dangershony: Brilliant

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -216,35 +214,55 @@ public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
...
+            
+            this.logger.LogTrace("(-)");
+            return peersToBan;
+        }
+
+        private void DisconnectChainHeader(ChainedHeader header)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(header), header);
+
+            header.Previous.Next.Remove(header);
```
dangershony: Excellent I like this method.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -216,35 +214,55 @@ public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
...
+
+                foreach (ChainedHeader nextHeader in header.Next)
+                    headersToProcess.Push(nextHeader);
+                
+                if (this.peerIdsByTipHash.TryGetValue(header.HashBlock, out HashSet<int> peers))
                 {
-                    this.peerTipsByPeerId.Remove(peer);
+                    foreach (int peerId in peers)
+                    {
+                        if (peerId != LocalPeerId)
```
dangershony: If a peer failed validation we probably want to remove it even if its claimed by us no?

noescape00: We will claim something that is not fully validated only during a reorg. In case reorg fails we will rewind to prev tip and remove the claim. It will be done by the ConsensusManager

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -51,11 +51,22 @@ public interface IChainedHeaderValidator
     /// </remarks>
     public sealed class ChainedHeaderTree
     {
+        /// <inheritdoc cref="Network"/>
```
dangershony: Nice try ;)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -250,7 +271,7 @@ public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
         /// and return all the peers that where claiming next headers.
         /// </summary>
         /// <param name="startHeader">The chained header to start from.</param>
-        /// <returns>List of peer Ids for banning.</returns>
+        /// <returns>List of peer Ids that were claiming headers on removed chains.</returns>
```
dangershony: This is a bit confusing I would be happy with just
>List of peer Ids that were claiming headers

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -14,7 +14,22 @@ namespace Stratis.Bitcoin.Consensus
     /// </summary>
     public interface IChainedHeaderValidator
     {
-        void Validate(ChainedHeader chainedHeader);
+        /// <summary>
+        /// Validation of a header that was seen for the first time.
```
Aprogiena: The validator doesn't care whether it is first-time or not.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -14,7 +14,22 @@ namespace Stratis.Bitcoin.Consensus
...
+        /// </summary>
+        /// <param name="chainedHeader">The chained header to be validated.</param>
+        void ValidateHeader(ChainedHeader chainedHeader);
+
+        /// <summary>
+        /// Validate the block signature (for POS) and merkle tree, throws if validation fails.
```
Aprogiena: Maybe put this into the remarks as it is implementation specific thing and that the describe it on higher level.

Verifies that the block data corresponds to the chain header.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: Validates the block header.

Should probably be enough

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -299,14 +320,21 @@ private void DisconnectChainHeader(ChainedHeader header)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Handles situation when blocks the data is downloaded for a given chained header.
```
dangershony: >Handles situation when a block data is downloaded for a given chained header.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: In fact that level of description should go in the implementation itself `(for POS) and merkle tree,`

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: This is some extra useful information

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: yes

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Yes but you would probably (and currently thats the case) always call this on new headers but not on headers there where already validated.

Its sort of similar to the comment bellow
> This validation represents minimal required validation for every block that we download.  
but you could call `VerifyBlockIntegrity` on blocks that where not downloaded (read form disk)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
             }
 
+            // Initialize local tip claim with consensus tip.
             this.AddOrReplacePeerTip(LocalPeerId, consensusTip.HashBlock);
+            
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>Gets the consensus tip.</summary>
+        private ChainedHeader GetConsensusTip()
```
dangershony: Whats the reason not to use ChainState?

noescape00: 1 dependency less

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>
+        /// Handles situation when partial validation for block data for a given <see cref="ChainedHeader"/> was successful.
+        /// </summary>
+        /// <remarks>
+        /// In case partial validation was successful we want to partially validate all the next blocks for which we have block data for.
+        /// <para>
+        /// If block that was just partially validated has more cumulative chainwork than our consensus tip we want to switch our consensus tip to this block.
```
dangershony: This sentence is not so clear.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Also it's a protection against chainstate.consensus tip being changed from some place we didnt want it to be changed

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: What's unclear about it?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+            {
+                this.logger.LogTrace("(-)[HEADER_NOT_FOUND]");
+                return null;
+            }
+
+            chainedHeader.BlockValidationState = ValidationState.PartiallyValidated;
+
+            if (chainedHeader.ChainWork > this.GetConsensusTip().ChainWork)
+            {
+                // A better tip that was partially validated is found. Set our node's claim on the tip to avoid that chain removal in
```
dangershony: We have to discuss this.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+        /// Handles situation when block data was considered to be invalid
+        /// for a given header during the partial or full validation.
+        /// </summary>
+        /// <param name="chainedHeader">Chained header which block data failed the validation.</param>
+        /// <returns>List of peer Ids that were claiming chain that contains an invalid block. Such peers should be banned.</returns>
+        public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeader), chainedHeader);
+
+            List<int> peersToBan = this.RemoveSubtree(chainedHeader);
```
dangershony: `RemoveSubBranches`

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+            List<int> peerIdsToResync = this.FindPeersToResync(newConsensusTip);
+
+            // Remove block data for the headers that are too far from the consensus tip.
+            this.CleanOldBlockDataFromMemory(newConsensusTip);
+
+            this.logger.LogTrace("(-)");
+            return peerIdsToResync;
+        }
+
+        /// <summary>Checks each peer's tip if it violates max reorg rule.</summary>
```
dangershony: The description does not sound correct (or the method name is not correct)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+        /// <param name="consensusTip">Consensus tip.</param>
+        /// <returns>List of peers which tips violate max reorg rule.</returns>
+        private List<int> FindPeersToResync(ChainedHeader consensusTip)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(consensusTip), consensusTip);
+
+            var peerIdsToResync = new List<int>();
+            uint maxReorgLength = this.chainState.MaxReorgLength;
```
dangershony: `MaxReorgLength` should be available in settings.

noescape00: we can't do that: `(this.network.Consensus.Options as PowConsensusOptions).MaxReorgLength;`
because it's in the consensus namespace

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Fixed

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+            {
+                foreach (KeyValuePair<int, uint256> peerIdToTipHash in this.peerTipsByPeerId)
+                {
+                    ChainedHeader peerTip = this.chainedHeadersByHash[peerIdToTipHash.Value];
+                    int peerId = peerIdToTipHash.Key;
+
+                    ChainedHeader fork = peerTip.FindFork(consensusTip);
+
+                    // Do nothing in case peer's tip is on our consensus chain.
+                    if ((fork != consensusTip) && (fork != peerTip))
```
dangershony: This condition is not clear can we try to write it better?
Perhaps an en extension method `bool IsOnSameChain(header, out fork)`

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+                }
+            }
+
+            this.logger.LogTrace("(-):*.{0}={1}", nameof(peerIdsToResync.Count), peerIdsToResync.Count);
+            return peerIdsToResync;
+        }
+
+        /// <summary>Cleans the block data for chained headers that are old. This data will still exist in the block store if it is enabled.</summary>
+        /// <param name="consensusTip">Consensus tip.</param>
+        private void CleanOldBlockDataFromMemory(ChainedHeader consensusTip)
```
dangershony: I am starting to reconsider `ChainedHeader.Block` this should be available in its own Interface.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -113,11 +134,331 @@ public void Initialize(ChainedHeader consensusTip)
...
+                ChainedHeader header = headersToProcess.Pop();
+
+                foreach (ChainedHeader nextHeader in header.Next)
+                    headersToProcess.Push(nextHeader);
+                
+                if (this.peerIdsByTipHash.TryGetValue(header.HashBlock, out HashSet<int> peers))
+                {
+                    foreach (int peerId in peers)
+                    {
+                        // There was a partially validated chain that was better than our consensus tip, we've started full validation
```
dangershony: Ok we need to discuss this as well.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -338,6 +340,23 @@ private List<int> FindPeersToResync(ChainedHeader consensusTip)
             return peerIdsToResync;
         }
 
+        /// <summary>Find the fork between two headers and return the fork if the headers are not on the same chain.</summary>
+        private ChainedHeader FindForkNotOnConensusChain(ChainedHeader peerTip, ChainedHeader consensusTip)
```
Aprogiena: You have mismatch of the summary and method name. It should be named FindForkIfChainedHeadersNotOnSameChain and your parameters should be named header1 and header2 as to be agnostic of what they represent.

dangershony: ok sounds fine I will fix

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -449,18 +449,20 @@ private void DisconnectChainHeader(ChainedHeader header)
...
+        /// Finds the header and verifies block integrity.
         /// </summary>
-        /// <param name="block">Block data.</param>
-        /// <param name="chainedHeader">If the function succeeds this is set to chained header which block data was downloaded.</param>
-        /// <returns><c>true</c> in case partial validation is required for the downloaded block, <c>false</c> otherwise.</returns>
+        /// <param name="block">The block.</param>
+        /// <returns>Chained header for a given block.</returns>
         /// <exception cref="BlockDownloadedForMissingChainedHeaderException">Thrown when block data is presented for a chained block that doesn't exist.</exception>
-        public bool BlockDataDownloaded(Block block, out ChainedHeader chainedHeader)
+        public ChainedHeader FindHeaderAndVerifyBlockIntegrity(Block block)
```
dangershony: Awesome

PR #1472, opened 2018-06-04
Verbose error logging for finickey endpoint

DanGould:


bokobza:
Agree with the change of error type (you need to fix a test), and for the stack trace in the logs.
But for the message returned to the caller, I think it's better to have the ToString method as the stackTrace might be too long and not very digestable by the front-end.

bokobza:
Good stuff 😄

PR #1473, opened 2018-06-04
Settings fix

noescape00:
It looks like a giant pr but all the changes are the same- just every Settings usage was touched. 

Changelog: 
1) Settings are now loaded when DI resolves setting class
2) Removed callbacks from features that exposed inner logic to the most high level interfaces


bokobza:
Mine!

noescape00:
@bokobza I've explained logic behind this PR to Dan but since you'd like to take a look at it I'll outline them here. 
Here are the problems I have with the current design:

1.

In every *Settings class we have callbacks. 
```
public DnsSettings(Action<DnsSettings> callback)
    : this()
{
    this.callback = callback;
}
```

Those are callbacks that are called after the settings are loaded- at the end of `Load` method. 

Now in the *Feature classes during adding a singleton we propagate this callback from the parameter of the method to the settings's constructor. 

```
public static IFullNodeBuilder UseDns(this IFullNodeBuilder fullNodeBuilder, Action<DnsSettings> setup = null)
        {
            LoggingConfiguration.RegisterFeatureNamespace<DnsFeature>("dns");

            fullNodeBuilder.ConfigureFeature(features =>
            {
                features
                .AddFeature<DnsFeature>()
                .FeatureServices(services =>
                {
                    services.AddSingleton<IMasterFile, DnsSeedMasterFile>();
                    services.AddSingleton<IDnsServer, DnsSeedServer>();
                    services.AddSingleton<DnsSettings>(new DnsSettings(setup));
                    ...
```

the question is: why do we expose such a low level logic to the high level interface of the FN builder? 
This callback mechanic is used only in the tests and dns feature. 
What's wrong with just adding a commandline argument? 

We can always achieve overriding of some setting by just adding a command line argument- node settings reader will always use last specified value in case there are many. 

2. 

Now instead of loading settings at the moment DI resolves settings class (when instance is first time accessed) we have an additional level of complexity- Load methods that are called by the FN feature executor:
```
this.Execute(service => service.LoadConfiguration());
```

and this leads to the kind of mistakes we had after yesterday's PR- wrong launch sequence when settings class was accessed before Load was executed. 

I don't see why we need an additional step at the node's initialization which just loads the settings from commandline args and config. 

Also that leads to more lines in tests- first create the settings class and after call Load on it. 


What I did in this PR is the removal of those Load methods- moved settings loading back to the constructor so it's executed when DI resolves the class and removed callbacks.

src/Stratis.Bitcoin.Features.RPC/RpcSettings.cs
```diff
@@ -36,24 +36,16 @@ public class RpcSettings
...
             this.AllowIp = new List<IPAddress>();
-        }
 
-        /// <summary> Initializes an instance of the object.</summary>
-        /// <param name="callback">The callback to call after the settings have been loaded.</param>
-        public RpcSettings(Action<RpcSettings> callback)
-            : this()
-        {
-            this.callback = callback;
+            this.Load(nodeSettings);
```
bokobza: Can you bring the content of the Load(NodeSettings) method inside the ctor?

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
```diff
@@ -22,38 +22,14 @@ public class WalletSettings
...
-        /// <summary>
-        /// A callback allow changing the default settings.
-        /// </summary>
-        private readonly Action<WalletSettings> callback;
-
-        /// <summary>
-        /// Initializes an instance of the object.
-        /// </summary>
-        public WalletSettings()
+        public WalletSettings() : this(NodeSettings.Default())
```
bokobza: Are we using this constructor?

src/Stratis.Bitcoin.Tests/Consensus/CheckpointsTest.cs
```diff
@@ -131,7 +132,7 @@ public void GetCheckPoint_CheckpointExists_PoWChain_ReturnsCheckpoint()
         [Fact]
         public void GetCheckPoint_CheckpointExists_PoSChain_ReturnsCheckpoint()
         {
-            var checkpoints = new Checkpoints(Network.StratisMain, new ConsensusSettings() { UseCheckpoints = true });
+            var checkpoints = new Checkpoints(Network.StratisMain, new ConsensusSettings(new NodeSettings()) { UseCheckpoints = true });
```
bokobza: Instead of having to specify `new NodeSettings()` inside every *Settings instantiation, would it make sense to have a parameterless ctor call `new NodeSettings()` or `NodeSettings.Default()`?

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCSettingsTest.cs
```diff
@@ -1,44 +0,0 @@
-﻿using NBitcoin;
```
bokobza: Shall we keep this test but use the settings init with RPCsettings?

src/Stratis.Bitcoin.Features.Miner.Tests/MinerSettingsTest.cs
```diff
@@ -1,99 +0,0 @@
-﻿using System;
```
bokobza: Can you keep these tests, but just remove the callbacks?

src/Stratis.Bitcoin.Features.Dns/DnsSettings.cs
```diff
@@ -35,40 +34,17 @@ public class DnsSettings
...
-        }
-
-        /// <summary>
-        /// Constructs this object whilst providing a callback to override/constrain/extend 
-        /// the settings provided by the Load method.
-        /// </summary>
-        /// <param name="callback">The callback used to override/constrain/extend the settings provided by the Load method.</param>
-        public DnsSettings(Action<DnsSettings> callback)
-            : this()
+        public DnsSettings() : this(NodeSettings.Default())
```
bokobza: Is this ctor used?

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/ConsensusRulesTest.cs
```diff
@@ -20,11 +21,11 @@ public ConsensusRulesTest()
         [Fact]
         public void Constructor_InitializesClass()
         {
-            this.consensusSettings = new ConsensusSettings
-            {
-                BlockAssumedValid = null,
-                UseCheckpoints = true
-            };
+            this.consensusSettings = new ConsensusSettings(new NodeSettings());
```
bokobza: you no like
```
 this.consensusSettings = new ConsensusSettings(new NodeSettings())
    {
        BlockAssumedValid = null,
        UseCheckpoints = true
    };
```
?

src/Stratis.Bitcoin.Features.Consensus/ConsensusOptions.cs
```diff
@@ -39,8 +34,6 @@ public PosConsensusOptions()
             this.ProofOfStakeReward = Money.COIN;
             this.PremineReward = Money.Coins(98000000);
             this.PremineHeight = 2;
-            this.StakeMinAge = 60;
```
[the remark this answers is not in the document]
noescape00: Those are from PoS v1 and v2. @Aprogiena was clearing them out when he joined, some code pieces are still sticking around.

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -37,33 +37,20 @@ public class ApiSettings
...
-        {
-            this.callback = callback;
-        }
-
-        /// <summary>
-        /// Loads the API related settings from the application configuration.
-        /// </summary>
-        /// <param name="nodeSettings">Application configuration.</param>
-        public void Load(NodeSettings nodeSettings)
+        public ApiSettings(NodeSettings nodeSettings)
```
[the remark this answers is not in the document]
noescape00: Agreed. Fixed

src/Stratis.Bitcoin.Features.RPC/RpcSettings.cs
noescape00: Ah, missed this one, ty.
Fixed

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
noescape00: No. Was used in the tests before, forgot to remove it after fixing the tests

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCSettingsTest.cs
noescape00: This test is basically testing if callback was called. If we keep it but without a callback it won't make much sense imho

src/Stratis.Bitcoin.Features.Dns/DnsSettings.cs
noescape00: No. Removed

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/ConsensusRulesTest.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Miner.Tests/MinerSettingsTest.cs
noescape00: Those are testing callback + inner logic (in case mining disabled- don't load other settings). 
Not sure if those are valid tests. But I don't mind keeping them if you'd like

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
noescape00: Actually, this parameterless constructor is still needed for the tests- without it you can't mock the class

noescape00: Same for dns settings constructor

noescape00:
@bokobza all fixes except for those removed tests are applied. Thanks for review! 🙂 
Let me know if you think that those tests make sense (imho they don't test anything)- I'll return them.

src/Stratis.Bitcoin.Tests/Consensus/CheckpointsTest.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Consensus/ConsensusOptions.cs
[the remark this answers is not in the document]
bokobza: fair enough

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
bokobza: ah yeah good point

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCSettingsTest.cs
bokobza: ok

src/Stratis.Bitcoin.Features.Miner.Tests/MinerSettingsTest.cs
bokobza: yeah please

bokobza:
If the tests can test that the settings are applied properly, there is still some value in my opinion.

bokobza:
Let's wait for a second review.

noescape00:
Reintroduced the tests

>If the tests can test that the settings are applied properly, there is still some value in my opinion.

Fair point

>Let's wait for a second review.

Ok

PR #1474, opened 2018-06-04
Mine genesis block

bokobza:
Added a method to mine a genesis block. 🔢

### DIFF
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -232,7 +232,7 @@ namespace NBitcoin
 
         /// <summary>
         /// Mines a new genesis block, to use with a new network.
-        /// Typically, 3 such genesis blocks need to be created when bootstrapping a network: for Main, Test and Reg.
+        /// Typically, 3 such genesis blocks need to be created when bootstrapping a new coin: for Main, Test and Reg networks.
         /// </summary>
         /// <param name="consensusFactory">
         /// The consensus factory used to create transactions and blocks. 
@@ -474,7 +474,7 @@ namespace NBitcoin
                             candidate = new BitcoinWitScriptAddress(str, network);
 
                         if (candidate is T)
-                            return (T)(object)candidate;
+                            return (T)candidate;
                     }
                     catch (Bech32FormatException)
                     {
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -244,6 +244,9 @@ namespace NBitcoin
         /// </param>
         /// <param name="target">
         /// The difficulty target under which the hash of the block need to be. 
+        /// Some more details: As an example, the target for the Stratis Main network is 00000fffffffffffffffffffffffffffffffffffffffffffffffffffffffffff.
+        /// To make it harder to mine the genesis block, have more zeros at the beginning (keeping the length the same). This will make the target smaller, so finding a number under it will be more difficult.
+        /// To make it easier to mine the genesis block ,do the opposite. Example of an easy one: 00ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff.
         /// Make the Test and Reg targets ones easier to find than the Main one so that you don't wait too long to mine the genesis block.
         /// </param>
         /// <param name="genesisReward">
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -205,7 +205,7 @@ namespace NBitcoin
         /// <summary>
         /// Register an immutable <see cref="Network"/> instance so it is queryable through <see cref="GetNetwork(string)"/> and <see cref="GetNetworks()"/>.
         /// </summary>
-        internal static Network Register(Network network)
+        public static Network Register(Network network)
         {
             IEnumerable<string> networkNames = network.AdditionalNames != null ? new[] { network.Name }.Concat(network.AdditionalNames) : new[] { network.Name };
 
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -321,9 +321,9 @@ namespace NBitcoin
             genesis.Header.HashPrevBlock = uint256.Zero;
             genesis.UpdateMerkleRoot();
 
-            // Iterate over the nonce until the block header hash is under the target.
-            uint256 hashTarget = target.ToUInt256();
-            while (genesis.Header.GetHash() >= hashTarget)
+            // Iterate over the nonce until the proof-of-work is valid.
+            // This will mean the block header hash is under the target.
+            while (!genesis.CheckProofOfWork())
             {
                 genesis.Header.Nonce++;
                 if (genesis.Header.Nonce == 0)
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -241,6 +241,7 @@ namespace NBitcoin
         /// <param name="coinbaseText">
         /// Traditionally a news headline from the day of the launch, but could be any string or link.
         /// This will be inserted in the input coinbase transaction script.
+        /// It should be shorter than 92 characters.
         /// </param>
         /// <param name="target">
         /// The difficulty target under which the hash of the block need to be. 
@@ -281,6 +282,9 @@ namespace NBitcoin
 
             if (string.IsNullOrEmpty(coinbaseText))
                 throw new ArgumentException($"Parameter '{nameof(coinbaseText)}' cannot be null. Use a news headline or any other appropriate string.");
+
+            if (coinbaseText.Length >= 92)
+                throw new ArgumentException($"Parameter '{nameof(coinbaseText)}' should be shorter than 92 characters.");
 
             if (genesisReward == null)
                 throw new ArgumentException($"Parameter '{nameof(genesisReward)}' cannot be null. Example use: 'Money.Coins(50m)'.");
--- a/src/NBitcoin/Network.cs
+++ b/src/NBitcoin/Network.cs
@@ -38,7 +38,7 @@ namespace NBitcoin
         private static readonly ConcurrentDictionary<string, Network> NetworksContainer = new ConcurrentDictionary<string, Network>();
 
         protected Block Genesis;
-        
+
         protected Network()
         {
             this.Consensus = new Consensus();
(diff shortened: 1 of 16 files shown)


src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
             return network;
         }
 
+        /// <summary>
+        /// Mines a new genesis block, to use with a new network.
+        /// Typically, 3 such genesis blocks need to be created when bootstrapping a network: for Main, Test and Reg.
```
noescape00: =>

```
Typically, 3 such genesis blocks need to be created when bootstrapping a new coin: for Main, Test and Reg networks.
```

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        /// <param name="pszTimestamp">
+        /// Traditionally a news headline from the day of the launch, but could be any string or link.
+        /// This will be inserted in the input coinbase transaction script.
+        /// </param>
+        /// <param name="target">
+        /// The difficulty target under which the hash of the block need to be. 
+        /// Make the Test and Reg targets ones easier to find than the Main one so that you don't wait too long to mine the genesis block.
+        /// </param>
+        /// <param name="genesisReward">
+        /// Specify how many coins to put in the genesis transaction's output. These coins are unspendable.
```
noescape00: Not sure if this value should be exposed. Why not just default it to 50 coins or smth? 

Btw I didn't know about the genesis reward and why it's unspendable: 

quote:
This is how the Bitcoin client worked when Satoshi created it: The client maintains a block database and a transaction database. When it finds that its block database is empty, it inserts the genesis block into its block database to get things started. This block includes a transaction sending 50 BTC to 1A1z..., but the client does not insert this transaction into its transaction database. So even though this transaction is part of the block chain, if the client sees a transaction spending this 50 BTC, it won't be able to find the 50 BTC transaction in its transaction database, and the spending transaction will be rejected. In other words, the genesis block's transaction isn't considered to be a "real transaction" by the original Bitcoin client.

(Current versions of Bitcoin-Qt handle the block/transaction databases in a totally different way, so the genesis block's transaction is now just a weird special case in the code.)

This could easily have been an oversight, or Satoshi could have intended this transaction to be unspendable. This quirk wasn't discovered until after Satoshi disappeared, so no one had a chance to ask him.

from https://www.reddit.com/r/Bitcoin/comments/1nc13r/the_first_50btc_block_reward_cant_be_spend_why/

src/NBitcoin/Network.cs
```diff
@@ -378,7 +474,7 @@ public static IBitcoinString Parse(string str, Network expectedNetwork)
                             candidate = new BitcoinWitScriptAddress(str, network);
 
                         if (candidate is T)
-                            return (T) (object) candidate;
+                            return (T)(object)candidate;
```
noescape00: casting to object is redundant

src/NBitcoin/Network.cs
bokobza: ahah yeah that's interesting.
I guess that's one of these things that has now become "tradition".

src/NBitcoin/Network.cs
bokobza: why did you pick this one lol.
ok, fixed

src/NBitcoin/Network.cs
bokobza: For the 50 coins, we just let the user decide. Doesn't really matter either way. 
In Stratis Test and Reg we set it to 0 for example.

src/NBitcoin/Network.cs
noescape00: >why did you pick this one lol.

Well, had to find at least something what is wrong with the PR 😂

src/NBitcoin/Network.cs
noescape00: Makes sense. 
Just wanted to avoid users thinking on what value to use when setting up a new network. 
But since comment says that it's fine.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
 
+        /// <summary>
+        /// Mines a new genesis block, to use with a new network.
+        /// Typically, 3 such genesis blocks need to be created when bootstrapping a new coin: for Main, Test and Reg networks.
+        /// </summary>
+        /// <param name="consensusFactory">
+        /// The consensus factory used to create transactions and blocks. 
+        /// Use <see cref="PosConsensusFactory"/> for proof-of-stake based networks.
+        /// </param>
+        /// <param name="pszTimestamp">
```
Aprogiena: We don't use Hungarians, no need to introduce them in the new code. Just timestamp would do, but if you really want to be explicit about type feel free to go with timestampString, but I would not.

Also, do we really want to college timestamp? Is it rather not Coinbase text?

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        /// </param>
+        /// <param name="genesisReward">
+        /// Specify how many coins to put in the genesis transaction's output. These coins are unspendable.
+        /// </param>
+        /// <param name="version">
+        /// The version of the transaction and the block header set in the genesis block. 
+        /// </param>
+        /// <example>
+        /// The following example shows the creation of a genesis block.
+        /// </example>
```
Aprogiena: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/example

The closing tag should be after the closing tag of code.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        /// BlockHeader header = genesis.Header;
+        /// Console.WriteLine("Make a note of the following values:");
+        /// Console.WriteLine("bits: " + header.Bits);
+        /// Console.WriteLine("nonce: " + header.Nonce);
+        /// Console.WriteLine("time: " + header.Time);
+        /// Console.WriteLine("version: " + header.Version);
+        /// Console.WriteLine("hash: " + header.GetHash());
+        /// Console.WriteLine("merkleroot: " + header.HashMerkleRoot);
+        /// </code>
+        /// <returns> A genesis block. </returns>
```
Aprogiena: Unnecessary extra spaces. We don't really use them elsewhere. If you want the separation between the text and the text, feel free to use what you used above – new line.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        public static Block MineGenesisBlock(ConsensusFactory consensusFactory, string pszTimestamp, Target target, Money genesisReward, int version = 1)
+        {
+            if (target == null)
+                throw new ArgumentException("Parameter 'target' cannot be null. Example use: new Target(new uint256(\"0000ffff00000000000000000000000000000000000000000000000000000000\"))");
```
Aprogiena: At least with the parameter names, please use name of, so that the string is updated when the argument gets renamed.

Aprogiena: Same below.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+                throw new ArgumentException("Parameter 'consensusFactory' cannot be null. Use 'new ConsensusFactory()' for Bitcoin-like proof-of-work blockchains" +
+                                            "and 'new PosConsensusFactory()' for Stratis-like proof-of-stake blockchains.");
+
+            if (string.IsNullOrEmpty(pszTimestamp))
+                throw new ArgumentException("Parameter 'pszTimestamp' cannot be null. Use a news headline or any other appropriate string.");
+
+            if (genesisReward == null)
+                throw new ArgumentException("Parameter 'genesisReward' cannot be null. Example use: 'Money.Coins(50m)'.");
+
+            DateTimeOffset time = DateTimeOffset.Now;
```
Aprogiena: Any particular reason why not to use date time provider here?

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+            Transaction txNew = consensusFactory.CreateTransaction();
+            txNew.Version = (uint)version;
+            txNew.Time = unixTime;
+            txNew.AddInput(new TxIn()
+            {
+                ScriptSig = new Script(
+                    Op.GetPushOp(0),
+                    new Op()
+                    {
+                        Code = (OpcodeType)0x1,
```
Aprogiena: Don't we have better construction for this somewhere? This is just full of magic constants. If there is no better way to write it, at least, every line to say what does it really mean.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        /// Mines a new genesis block, to use with a new network.
+        /// Typically, 3 such genesis blocks need to be created when bootstrapping a new coin: for Main, Test and Reg networks.
+        /// </summary>
+        /// <param name="consensusFactory">
+        /// The consensus factory used to create transactions and blocks. 
+        /// Use <see cref="PosConsensusFactory"/> for proof-of-stake based networks.
+        /// </param>
+        /// <param name="pszTimestamp">
+        /// Traditionally a news headline from the day of the launch, but could be any string or link.
+        /// This will be inserted in the input coinbase transaction script.
```
Aprogiena: For sure there is some kind of size limit for this argument. It should be mentioned in the comment.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+        /// </param>
+        /// <param name="target">
+        /// The difficulty target under which the hash of the block need to be. 
+        /// Make the Test and Reg targets ones easier to find than the Main one so that you don't wait too long to mine the genesis block.
```
Aprogiena: Same here – it should be explained, which values are meaningful.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,102 @@ internal static Network Register(Network network)
...
+            genesis.UpdateMerkleRoot();
+
+            // Iterate over the nonce until the block header hash is under the target.
+            uint256 hashTarget = target.ToUInt256();
+            while (genesis.Header.GetHash() >= hashTarget)
+            {
+                genesis.Header.Nonce++;
+                if (genesis.Header.Nonce == 0)
+                    genesis.Header.Time++;
+            }
```
Aprogiena: It might be a good idea if the nonce value comes as an optional argument and we do this hash finding loop only if it is not given – i.e. it has default value of 0. Such argument would allow us to precompute the correct nonce in the way that the creation of the Genesis can then be done instantly.

src/NBitcoin/Network.cs
bokobza: yeah it's a weird name. I was just keeping in line with bitcoin and the names we have in our current CreateGenesis methods.

bokobza: I can do coinbaseText

src/NBitcoin/Network.cs
bokobza: ah yeah missed that, thanks

src/NBitcoin/Network.cs
bokobza: changed

src/NBitcoin/Network.cs
bokobza: yep

src/NBitcoin/Network.cs
bokobza: because the DatetimeProvider is not available in NBitcoin.

src/NBitcoin/Network.cs
bokobza: So this method is to be executed on its own, just to mine the genesis block.
Then, the creator of the network should look at the data included in this block and use it in the network settings.
As such, it's not a replacement for the CreateGenesisBlock methods, whose signature is `Block CreateStratisGenesisBlock(ConsensusFactory consensusFactory, uint nTime, uint nNonce, uint nBits, int nVersion, Money genesisReward)`.

Now thinking about it, I could add nonce and time as optional params.
I'll try to see how it looks.

src/NBitcoin/Network.cs
bokobza: I'm already providing a code sample and telling the user to make it easier for Test and Reg.
How much more help should I give to someone who wants to start a coin but doesn't understand how to make a target easier or harder 🤣

src/NBitcoin/Network.cs
bokobza: So I had this before at 90 characters (I read that online) but the only limitation that I know of is the ScriptSig that's limited to 10,000 bytes.

src/NBitcoin/Network.cs
bokobza: in

src/NBitcoin/Network.cs
bokobza: in

src/NBitcoin/Network.cs
bokobza: in

src/NBitcoin/Network.cs
bokobza: in

src/NBitcoin/Network.cs
Aprogiena: I think it is more like 200. I think there is consensus rule for that with some kind of limit around 200.

src/NBitcoin/Network.cs
Aprogiena: I would appreciate little more info, but it is not required if you don't want it.

src/NBitcoin/Network.cs
Aprogiena: I leave this decision to you, it was just an idea, maybe use it in different way than I suggested.

src/NBitcoin/Network.cs
bokobza: added

src/NBitcoin/Network.cs
bokobza: Found it https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CheckPowTransactionRule.cs#L96

ScriptSig is max 100 so I'll set the limit under that.

bokobza:
this is ready for review for now.

src/NBitcoin/Network.cs
Aprogiena: that is the one! i thought it was 200 :-) wroong!

bokobza: good catch tho 🐟

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,109 @@ internal static Network Register(Network network)
...
+            genesis.Header.Bits = target;
+            genesis.Header.Nonce = 0;
+            genesis.Header.Version = version;
+            genesis.Transactions.Add(txNew);
+            genesis.Header.HashPrevBlock = uint256.Zero;
+            genesis.UpdateMerkleRoot();
+
+            // Iterate over the nonce until the block header hash is under the target.
+            uint256 hashTarget = target.ToUInt256();
+            while (genesis.Header.GetHash() >= hashTarget)
```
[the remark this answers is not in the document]
bokobza: I saw that but CheckProofOfWork requires Consensus, which I don't have at this stage.
It's not actually used so I'll check if it can be removed.

src/NBitcoin/Network.cs
```diff
@@ -230,6 +230,109 @@ internal static Network Register(Network network)
             return network;
         }
 
+        /// <summary>
+        /// Mines a new genesis block, to use with a new network.
```
dangershony: Beautiful summery

bokobza: thanks bruh

PR #1475, opened 2018-06-05
One rewind data instance per block when close to chain tip

noescape00:
Flush coinview when consensus tip is close to the chain tip to ensure that we have 1 rewind data per block. 
Previously we were creating 1 rewind data for batch of blocks so during a reorg at the tip we were able to call a rewind on consensus tip and it would rewind a lot of blocks (up to 5000 if reorg happened when we just synced).

This is PR 2 from the list of PRs for the big refasctoring


src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -441,6 +439,22 @@ public void Stop()
...
+        /// <remarks>
+        /// For blockchains with max reorg property flush is required when consensus tip is less than max reorg blocks behind the chain tip.
+        /// If there is no max reorg property - flush is required when consensus tip timestamp is less than two weeks behind the adjusted time.
+        /// </remarks>
+        private bool FlushRequired()
+        {
+            if (this.chainState.MaxReorgLength != 0)
+                return this.Chain.Height - this.Tip.Height < this.chainState.MaxReorgLength;
+
+            int twoWeeksSeconds = 2 * 24 * 60 * 60;
```
dangershony: `twoWeeksSeconds` -> `twoWeeksAsSeconds`

dangershony: Two weeks might be too long no? even 1 day is enough.

dangershony:
Ah I did not run tests yet (or sync chains).

noescape00:
>Ah I did not run tests yet (or sync chains).

Kk. 

Ps
For me tests are passing (except for those tests that fail for 2 weeks already), stratis main and test synced.

src/Stratis.Bitcoin.Features.Consensus/ConsensusLoop.cs
```diff
@@ -441,6 +439,22 @@ public void Stop()
...
+        /// <remarks>
+        /// For blockchains with max reorg property flush is required when consensus tip is less than max reorg blocks behind the chain tip.
+        /// If there is no max reorg property - flush is required when consensus tip timestamp is less than two weeks behind the adjusted time.
+        /// </remarks>
+        private bool FlushRequired()
+        {
+            if (this.chainState.MaxReorgLength != 0)
+                return this.Chain.Height - this.Tip.Height < this.chainState.MaxReorgLength;
+
+            int twoWeeksAsSeconds = 2 * 24 * 60 * 60;
```
Aprogiena: This should be a constant on the class with comment. It should be explained why this number was selected.

noescape00: Fixed

Issue #1436, opened 2018-05-21
[RF] Remove old consensus manager


dangershony:

dangershony:
The current code will go in the consensus loop.

PR #1476, opened 2018-06-05
Consensus manager removal

noescape00:
Moved logic from consensus manager to consensus loop

Fix: #1436


Aprogiena:
I don't see anything wrong at 1st glance, but I have not run anything as usual and this PR needs just running everything more than anything else.

dangershony:
I think its fairly safe to commit this PR as is.

bokobza:
I think the 2 interfaces here are not necessary:
`INetworkDifficulty, IGetUnspentTransaction`.
INetworkDifficulty is only used there and IGetUnspentTransaction is also used in MempoolManager like so: 
```
public interface IPooledGetUnspentTransaction : IGetUnspentTransaction
{
}
```

noescape00:
>I think the 2 interfaces here are not necessary

Agreed and I think we should get rid of them since they are more like a hack to be honest. In general I don't like the way we break architecture sometimes by defining the interface in the base feature and implementing it somewhere else.

>INetworkDifficulty is only used there

It's used in FN controller's GetInfo() (so is the 2nd guy). 

I think those things should be improved but they are outside of the scope of this PR. It's more like a design decision of how we deal with dependencies or do we keep RPC in one feature or do the same as we do with API - different features have their controllers.

bokobza:
ah yeah didn't see that.
Ok leave it for now, we can rediscuss that after the current more pressing work.
Are all the tests passing?

noescape00:
>Are all the tests passing, apart from the ones that were already failing?

Yes. Also synced the chains- all fine for me.

bokobza:
ok i'll push it.

PR #1477, opened 2018-06-05
Coinstaske age tests

noescape00:
Tests that check that UTXOs with correct depth in the chain are eligible for staking.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
+++ b/src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
@@ -366,9 +366,12 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
         [Fact]
         public void CoinstakeAge_AfterActivation_Testnet()
         {
-            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, 450000, 450000 - 18));
+            int activationHeight = PosConsensusOptions.CoinstakeMinConfirmationActivationHeightTestnet;
+            int afterActivationHeight = activationHeight + 1000;
+
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, afterActivationHeight, afterActivationHeight - 18));
             
-            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisTest, 450000, 450000 - 17));
+            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisTest, afterActivationHeight, afterActivationHeight - 17));
         }
 
         /// <summary>This is a test of coinstake age softfork activation on testnet.</summary>
@@ -376,11 +379,13 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
         [Fact]
         public void CoinstakeAge_AtTheActivation_Testnet()
         {
-            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, 436000 - 2, 436000 - 10)); // mining block before activation
+            int activationHeight = PosConsensusOptions.CoinstakeMinConfirmationActivationHeightTestnet;
+
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, activationHeight - 2, activationHeight - 10)); // mining block before activation
             
-            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, 436000 - 1, 436000 - 19)); // mining activation block
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisTest, activationHeight - 1, activationHeight - 19)); // mining activation block
             
-            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisTest, 436000 - 1, 436000 - 18)); // mining activation block
+            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisTest, activationHeight - 1, activationHeight - 18)); // mining activation block
         }
 
         /// <summary>This is a test of coinstake age softfork activation on mainnet.</summary>
@@ -398,9 +403,12 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
         [Fact]
         public void CoinstakeAge_AfterActivation_Mainnet()
         {
-            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisMain, 950000, 950000 - 498));
+            int activationHeight = PosConsensusOptions.CoinstakeMinConfirmationActivationHeightMainnet;
+            int afterActivationHeight = activationHeight + 1000;
+
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisMain, afterActivationHeight, afterActivationHeight - 498));
             
-            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisMain, 950000, 950000 - 497));
+            Assert.False(this.WasUtxoSelectedForStaking(Network.StratisMain, afterActivationHeight, afterActivationHeight - 497));
         }
 
         /// <summary>This is a test of coinstake age softfork activation on mainnet.</summary>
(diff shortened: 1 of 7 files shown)


src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -351,6 +352,100 @@ public void GetNetworkWeight_UsesLast73Blocks_CalculatedNetworkWeightUsingLatest
 
             Assert.Equal(4607763.9659653762, weight);
         }
+        
+        [Fact]
+        public void CoinstakeAge_BeforeActivation_Testnet()
```
majikandy: These test names don't currently describe the behaviour being proven

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -351,6 +352,100 @@ public void GetNetworkWeight_UsesLast73Blocks_CalculatedNetworkWeightUsingLatest
...
+        {
+            bool suitable = this.WasUtxoSelectedForStaking(Network.StratisTest, 1000, 1000 - 8); // utxo depth is 9, mining block at 10
+            Assert.True(suitable);
+
+            bool suitable2 = this.WasUtxoSelectedForStaking(Network.StratisTest, 1000, 1000 - 7); // utxo depth is 8, mining block at 9
```
majikandy: 2? seriously? 😃

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -351,6 +352,100 @@ public void GetNetworkWeight_UsesLast73Blocks_CalculatedNetworkWeightUsingLatest
...
+            Assert.True(suitable);
+
+            bool suitable2 = this.WasUtxoSelectedForStaking(Network.StratisMain, 950000, 950000 - 497);
+            Assert.False(suitable2);
+        }
+
+        [Fact]
+        public void CoinstakeAge_AtTheActivation_Mainnet()
+        {
+            bool suitable1 = this.WasUtxoSelectedForStaking(Network.StratisMain, 940000 - 2, 940000 - 50); // mining block before activation
```
majikandy: magic numbers and comments make this difficult to understand what it is actually checking. Can you give this more meaning by giving all the magic numbers some meaning with variable names or constants.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
noescape00: Behavior is basically described in the assertions. Each test check if at some heights coins were suitable if it's expected and not suitable if it's not expected

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
noescape00: Yeah. Sounded better than moving it to a separated test

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -351,6 +352,100 @@ public void GetNetworkWeight_UsesLast73Blocks_CalculatedNetworkWeightUsingLatest
...
+        private bool WasUtxoSelectedForStaking(Network network, int chainTipHeight, int utxoHeight)
+        {
+            this.network = network;
+            this.network.Consensus.Options = new PosConsensusOptions();
+            this.chain = GenerateChainWithBlockTimeAndHeight(2, this.network, 60, 0x1df88f6f);
+
+            PosMinting miner = this.InitializePosMinting();
+
+            ChainedHeader chainTip = this.chain.Tip;
+            ReflectionExtensions.SetPrivatePropertyValue(chainTip, "Height", chainTipHeight);
```
majikandy: can this be called as a proper extension method or not?
chainTip.SetPrivatePropertyValue("Height", `chainTipHeight`);

Also. despite the fact I mentioned using reflection before when you needed to do something in a test that you didn't want in the proper code. I would say it is still a code smell and suggests an flaw in the design of the component (code smell implying, it is not always necessarily bad, it just needs considering....is this a bad design when this is needed?)

src/Stratis.Bitcoin.Features.Miner/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Stratis.Bitcoin.Features.Miner.Tests")]
```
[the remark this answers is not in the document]
noescape00: Makes sense, fixed.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
majikandy: It is good practice to put it in the test name, since you will see this in test runners like resharper. Otherwise we could just name test methods as test1(), test2(), test3().  The test name is your chance to provide some useful information to help locate tests when you are changing behaviours later and immediate information when a test fails.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
majikandy: Is it just testing the either side of the activation point?
I'd inline it to the Assert.True/False and not have the variable at all then possibly.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
noescape00: >can this be called as a proper extension method or not?
>chainTip.SetPrivatePropertyValue("Height", chainTipHeight);

Fair point. Converted `SetPrivatePropertyValue` to a proper extension method. 

>despite the fact I mentioned using reflection before when you needed to do something in a test >that you didn't want in the proper code. I would say it is still a code smell and suggests an flaw >in the design of the component (code smell implying, it is not always necessarily bad, it just >needs considering....is this a bad design when this is needed?)

I doubt that in order to write a test it's a good idea to change the design of a chained header. 

There is a way to do it alternatively- generate a concurrent chain with 950 000 headers. But do we really want to generate such a chain every test just to avoid reflection?

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
noescape00: Fair enough. Added comments.

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
noescape00: Ok, fixed

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -351,6 +352,96 @@ public void GetNetworkWeight_UsesLast73Blocks_CalculatedNetworkWeightUsingLatest
...
+        }
+
+        /// <summary>This is a test of coinstake age softfork activation on mainnet.</summary>
+        /// <remarks><see cref="PosConsensusOptions.GetStakeMinConfirmations"/></remarks>
+        [Fact]
+        public void CoinstakeAge_AtTheActivation_Mainnet()
+        {
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisMain, 940000 - 2, 940000 - 50)); // mining block before activation
+            
+            Assert.True(this.WasUtxoSelectedForStaking(Network.StratisMain, 940000 - 1, 940000 - 499)); // mining activation block
```
Aprogiena: All the big numbers here should be constants

noescape00: Fixed

Issue #1321, opened 2018-04-23
[RF] Create chained header tree component


dangershony:
Create chained header tree component and define Max interest gap. 
Create basic methods of this component and at unit tests.
- Chained header tree is a tree of all chains that we can see.
- Consider resolving issue 1156 during the design.

noescape00:
ChainedHeaderTree design:

## Private data structures:
- PeerTipsByHash: dictionary[hash - List of peer ids]
- ChainHeadersByHash: dictionary[hash - chainedBlock]

## Methods:

### CT changes (advances or reorgs):
Check every peer.tip against max reorg (using PeerTipsByHash and finding forkpoint(CT, tip)). If violated - discard tip, do CHB.PendingTip=null & CHB.TrySync, but dont disconnect or ban (be carefull in situation when we check peer.tip and at the same time new headers are presented)


### HeadersPresented
	1) If headers[0].PrevHash is not in the ChainHeadersByHash - return (Can't connect)
	2) Find using ChainHeadersByHash if some of the headers are already created and validated
	   For new headers- start creating and appending new ChainHeaders one by one. Validate each (if its not possible due to insufficency - return). First new header should be checked against max reorg (by finding forkpoint(CT, tip)) - If violated - discard tip; return
	   When new ChainHeader is created it should be added to previous.Next list
	3) Add tip to PeerTipsByHash (claim new tip)
	4) call RemoveChainClaim(peer.Id, peer.OldTip)
	5) Check work(tip) against work(CT) and mark tip's chain as DataRequired if it's better (report to BM if new data required)
	
	returns:
	* Max reorg violated (caller should disconnect, ban) - Exception
	* Can't connect (caller should request new headers) - Exception
	* New data marked DataRequested (caller should pass it to the block puller)	&&& how many headers were consumed (those that were not are insufficient)

### Peer disconnect
	RemoveChainClaim(peer.Id, peer.Tip)

### private RemoveChainClaim (peerId, peerTip):
	currentHeader = peerTip;

	while(true)
		nextExists = currentHeader.Next.Lenght != 0
		listOfPeersClaimingThisHeader = PeerTipsByHash[currentHeader]

		isEmpty = false;

		if (listOfPeersClaimingThisHeader != null)
			if (peerId == ANY_PEER)
				listOfPeersClaimingThisHeader.Clean()
			else
				listOfPeersClaimingThisHeader.Remove(peerId)

[...]

PR #1481, opened 2018-06-06
Tests for chained header tree

thefazzer:
**Fulfils the following tests supporting [RF] Create chained header tree component**
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1321
2) Supply headers that we already have, make sure no new ChainedHeaders were created
3) Supply some headers and after that supply some more headers and make sure that PeerTipsByHash is updated (at the total amount of items is the same)
6) Make sure checkpoints are off- supply some headers, CHT should return ToDownload array of the same size as the amount of headers.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -13,6 +13,11 @@ namespace Stratis.Bitcoin.Tests.Consensus
 {
     public class ChainedHeaderTreeTest
     {
+        private readonly int PEER_DEFAULT = -1;
+        private readonly int PEER_ONE = 1;
+        private readonly int PEER_TWO = 2;
+        private readonly int PEER_THREE = 3;
+
         public class TestContext
         {
             public Network Network = Network.RegTest;
@@ -149,6 +154,97 @@ namespace Stratis.Bitcoin.Tests.Consensus
             Assert.True(testContext.NoDownloadRequested(connectedHeadersOld));
             Assert.Equal(listOfNewHeaders.Last(), connectedHeadersNew.DownloadTo.Header);
             Assert.Equal(listOfNewHeaders.First(), connectedHeadersNew.DownloadFrom.Header);
+        }
+
+        // Supply headers that we already have, make sure no new ChainedHeaders were created.
+        [Fact]
+        public void ConnectHeaders_SupplyExistingHeaders_DontCreateNewChainHeaders()
+        {
+            var testContext = new TestContext();
+            ChainedHeaderTree cht = testContext.CreateChainedHeaderTree();
+            ChainedHeader chainTip = testContext.ExtendAChain(5);
+            Assert.Empty(cht.GetChainedHeadersByHash());
+
+            cht.Initialize(chainTip, true);
+
+            Assert.Equal(6, cht.GetChainedHeadersByHash().Keys.Count);
+
+            var beforeKeys = new List<uint256>(cht.GetChainedHeadersByHash().Keys);
+
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 5);
+            cht.ConnectNewHeaders(1, listOfExistingHeaders);
+            var afterKeys = new List<uint256>(cht.GetChainedHeadersByHash().Keys);
+
+            // check chainedHeadersByHash map before and after adding duplicate headers
+            Assert.Equal(beforeKeys.Count, afterKeys.Count);
+            Assert.True(beforeKeys.All(afterKeys.Contains));
+        }
+
+        // Supply some headers and after that supply some more headers
+        // and make sure that PeerTipsByHash is updated (the total amount of items is the same).
+        [Fact]
+        public void ConnectHeaders_SupplyHeadersThenSupplyMore_PeerTipsByHashShouldBeUpdated()
+        {
+            TestContext testContext = new TestContext();
+            ChainedHeaderTree cht = testContext.CreateChainedHeaderTree();
+
+            var chainTip = testContext.ExtendAChain(10);
+            cht.Initialize(chainTip, true);
+            chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
+            ChainedHeader newChainTip = testContext.ExtendAChain(10, chainTip);
+
+            var listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 10);
+            var listOfNewHeaders = testContext.ChainedHeaderToList(newChainTip, 10);
+
+            testContext.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
+            chainTip.BlockValidationState = ValidationState.FullyValidated;
+
+            cht.ConnectNewHeaders(1, listOfExistingHeaders);
+
+            var peerIdsByTipHashBefore = new Dictionary<uint256, HashSet<int>>(cht.GetPeerIdsByTipHash());
+            var chainedHeaderTreeHashesBefore = cht.GetChainedHeadersByHash();
+
+            cht.ConnectNewHeaders(1, listOfNewHeaders);
+
+            var peerIdsByTipHashAfter = cht.GetPeerIdsByTipHash();
+            var chainedHeaderTreeHashesAfter = cht.GetChainedHeadersByHash();
+
+            // Same number of entries as we are reassigning hash
+            Assert.True(chainedHeaderTreeHashesBefore.Count == chainedHeaderTreeHashesAfter.Count );
+
+            // Peer's tip hash has changed
+            Assert.True(peerIdsByTipHashBefore.FirstOrDefault(x => x.Value.Contains(1)).Key !=
+                        peerIdsByTipHashAfter.FirstOrDefault(x => x.Value.Contains(1)).Key);
+        }
+
+        // Make sure checkpoints are off - supply some headers,
+        // CHT should return ToDownload array of the same size as the amount of headers.
+        [Fact]
+        public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeaders()
+        {
+            // Setup
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            var chainTip = ctx.ExtendAChain(10);
+            cht.Initialize(chainTip, true);
+            chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
+            ctx.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
+            ctx.ConsensusSettings.UseCheckpoints = false;
+            ChainedHeader newChainTip = ctx.ExtendAChain(10, chainTip);
+            List<BlockHeader> listOfNewHeaders = ctx.ChainedHeaderToList(newChainTip, 10);
+
+            // Checkpoints are off
+            Assert.False(ctx.ConsensusSettings.UseCheckpoints);
+
+            // Supply some headers
+            var connectedNewHeaders = cht.ConnectNewHeaders(PEER_ONE, listOfNewHeaders);
+
+            var chainedHeaderFrom = connectedNewHeaders.DownloadFrom;
+            var chainedHeaderTo = connectedNewHeaders.DownloadTo;
+            int headersToDownloadCount = chainedHeaderTo.Height - chainedHeaderFrom.Height + 1; // Inclusive
+
+            // ToDownload array of the same size as the amount of headers
+            Assert.Equal(headersToDownloadCount, listOfNewHeaders.Count);
         }
     }
 }
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -132,20 +132,20 @@ namespace Stratis.Bitcoin.Tests.Consensus
         public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
         {
             var testContext = new TestContext();
-            var chainedHeaderTree = testContext.CreateChainedHeaderTree();
+            ChainedHeaderTree chainedHeaderTree = testContext.CreateChainedHeaderTree();
 
             ChainedHeader chainTip = testContext.ExtendAChain(10);
             chainedHeaderTree.Initialize(chainTip, true); // initialize the tree with 10 headers
             chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
             ChainedHeader newChainTip = testContext.ExtendAChain(10, chainTip); // create 10 more headers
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
             Assert.Equal(listOfNewHeaders.Last(), connectedHeadersNew.DownloadTo.Header);
             Assert.Equal(listOfNewHeaders.First(), connectedHeadersNew.DownloadFrom.Header);
         }
+
+        // Supply headers that we already have, make sure no new ChainedHeaders were created.
```
dangershony: For now for simplicity please set the issue test case number on the summery (we can delete later)

Aprogiena:
Test number 2 has already being implemented.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -13,6 +13,11 @@ namespace Stratis.Bitcoin.Tests.Consensus
 {
     public class ChainedHeaderTreeTest
     {
+        private readonly int PEER_DEFAULT = -1;
+        private readonly int PEER_ONE = 1;
+        private readonly int PEER_TWO = 2;
+        private readonly int PEER_THREE = 3;
```
Aprogiena: Code style

Aprogiena: Also why not const?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: Use XML comments for commenting on methods.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -13,6 +13,11 @@ namespace Stratis.Bitcoin.Tests.Consensus
 {
     public class ChainedHeaderTreeTest
     {
+        private readonly int PEER_DEFAULT = -1;
```
Aprogiena: I don't see this guy being used anywhere. If it is supposed to be ID of local node, use the constant that is already defined in CHT itself.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            Assert.Empty(cht.GetChainedHeadersByHash());
+
+            cht.Initialize(chainTip, true);
+
+            Assert.Equal(6, cht.GetChainedHeadersByHash().Keys.Count);
+
+            var beforeKeys = new List<uint256>(cht.GetChainedHeadersByHash().Keys);
+
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 5);
+            cht.ConnectNewHeaders(1, listOfExistingHeaders);
```
Aprogiena: Why would you define those constants for peer identifiers if you don't use them then?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: Also feel free to write those descriptions as the descriptions are just quickly hacked and don't have production quality.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+
+        // Supply some headers and after that supply some more headers
+        // and make sure that PeerTipsByHash is updated (the total amount of items is the same).
+        [Fact]
+        public void ConnectHeaders_SupplyHeadersThenSupplyMore_PeerTipsByHashShouldBeUpdated()
+        {
+            TestContext testContext = new TestContext();
+            ChainedHeaderTree cht = testContext.CreateChainedHeaderTree();
+
+            var chainTip = testContext.ExtendAChain(10);
```
Aprogiena: It would be much easier to read if the code style was respected.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            cht.Initialize(chainTip, true);
+            chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
+            ChainedHeader newChainTip = testContext.ExtendAChain(10, chainTip);
+
+            var listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 10);
+            var listOfNewHeaders = testContext.ChainedHeaderToList(newChainTip, 10);
+
+            testContext.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
+            chainTip.BlockValidationState = ValidationState.FullyValidated;
```
Aprogiena: Is it not already set to fully validated state by CHT initialisation?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            var peerIdsByTipHashBefore = new Dictionary<uint256, HashSet<int>>(cht.GetPeerIdsByTipHash());
+            var chainedHeaderTreeHashesBefore = cht.GetChainedHeadersByHash();
+
+            cht.ConnectNewHeaders(1, listOfNewHeaders);
+
+            var peerIdsByTipHashAfter = cht.GetPeerIdsByTipHash();
+            var chainedHeaderTreeHashesAfter = cht.GetChainedHeadersByHash();
+
+            // Same number of entries as we are reassigning hash
+            Assert.True(chainedHeaderTreeHashesBefore.Count == chainedHeaderTreeHashesAfter.Count );
```
Aprogiena: Extra space

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+
+            // Peer's tip hash has changed
+            Assert.True(peerIdsByTipHashBefore.FirstOrDefault(x => x.Value.Contains(1)).Key !=
```
Aprogiena: The original description of the test it's little bit outdated, we now have 2 structures that we want to check here - peerIdsByTipHash and peerTipsByPeerId. There should be change in both. In the 1st structure you already check it – the value of the peer identifier should move from one key to another key. In the 2nd structure the value should be changed for the peer.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+        // CHT should return ToDownload array of the same size as the amount of headers.
+        [Fact]
+        public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeaders()
+        {
+            // Setup
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            var chainTip = ctx.ExtendAChain(10);
+            cht.Initialize(chainTip, true);
+            chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
```
Aprogiena: This again should be already set by the initialisation.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            ctx.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
```
Aprogiena: Do we really need to make this setup? Is it used anywhere?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            Assert.False(ctx.ConsensusSettings.UseCheckpoints);
+
+            // Supply some headers
+            var connectedNewHeaders = cht.ConnectNewHeaders(PEER_ONE, listOfNewHeaders);
+
+            var chainedHeaderFrom = connectedNewHeaders.DownloadFrom;
+            var chainedHeaderTo = connectedNewHeaders.DownloadTo;
+            int headersToDownloadCount = chainedHeaderTo.Height - chainedHeaderFrom.Height + 1; // Inclusive
+
+            // ToDownload array of the same size as the amount of headers
```
Aprogiena: Little bit outdated to description again, that is why I said it might be good thing to rewrite it. No longer we use to download array, but as you correctly recognised we only transfer lower and upper bounds. So the in-line comments should reflect the reality.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -150,5 +155,96 @@ public void ConnectHeaders_NewAndExistingHeaders_ShouldCreateNewHeaders()
...
+            // Setup
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            var chainTip = ctx.ExtendAChain(10);
+            cht.Initialize(chainTip, true);
+            chainTip.BlockDataAvailability = BlockDataAvailabilityState.BlockAvailable;
+            ctx.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
+            ctx.ConsensusSettings.UseCheckpoints = false;
+            ChainedHeader newChainTip = ctx.ExtendAChain(10, chainTip);
+            List<BlockHeader> listOfNewHeaders = ctx.ChainedHeaderToList(newChainTip, 10);
```
Aprogiena: Not only here, general comment, you always supply the full chain from the 1st to the last header. Although this is fine in terms of test correctness, the test would actually cover little more if you for example use value 8 instead of 10 here so that we simulate the peer to provide not the full chain but only a part of it. Elsewhere you then could 1st provide just 2 or 3 blocks in the 1st run and then just supply another 5 or maybe 15 blocks so that we don't have every test case repeating the same scenario in terms of which chains are supplied in the relation to how CHT was initialised.

So in general, the otherwise would be, add some variability.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Agreed - removed these constants as they should be in CHT & we should have collective agreement on them

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: As above - removed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Updated to XML comment style, descriptions corrected.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Yes I need both the CHT and the context (for assigning and checking checkpoints)

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: I'm not sure which pattern or code construct replaces DownloadTo?  Happy to make these changes

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: (not)using var :-)

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: i meant specifically this ConsensusTip mock, where is it used?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: You have it correctly in the code, just the comment

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Corrected these

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Nowhere.  I've removed all instances.  You are correct - I think this was required by earlier code but is now redundant.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: Not really, you only done half where the type was not needed and you remove the type. You also need to do the other half where the type is required for example on line 168 and alike.

thefazzer: Have amended as follows:  
Used var where return object type specified explicitly or implied by method call or return variable name (I've made names more descriptive where necessary).  For other cases where return is type is unknown or ambiguous have used explicit typing

thefazzer: Reverted as per discussion.  Explicit typing used and var only where Type Name on RHS of =

PR #1483, opened 2018-06-06
Node stops attempting to connect

noescape00:
This is a bugfix for a bug reported by @maciejzaleski 

The problem is that at some point all the peers in peers.json have maxed out connectionAttempts count and the selector always returns nothing. 

This can happen when node was working without access to the internet for some time or if node was running for a very long time. 

I've added a quick fix that will reset attempts count in case there are no peers to try and all peers from the list that are not banned have their connection attempts count maxed out. 
After resetting the node will recover if any of the peers are still alive.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/P2P/PeerSelectorTests.cs
+++ b/src/Stratis.Bitcoin.Tests/P2P/PeerSelectorTests.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Net;
 using Stratis.Bitcoin.Configuration;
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -194,12 +194,25 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
                 this.logger.LogTrace("[RETURN_ATTEMPTED_HC_FAILED]");
                 return attempted;
             }
+
```
fassadlr: Ivan can you please move this out to its own method and write a unit test for it? The `SelectPreferredPeers` method is already getting too bloated ;)

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -194,12 +194,25 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
                 this.logger.LogTrace("[RETURN_ATTEMPTED_HC_FAILED]");
                 return attempted;
             }
+            
+            int attemptedReachedThresholdCount = this.peerAddresses.Values.Count(p => p.ConnectionAttempts == PeerAddress.AttemptThreshold && !this.IsBanned(p));
```
fassadlr: Please add an extension method for `peerAddresses` so that you can go `this.peerAddresses.NotBanned()` and set it to a field in the method. t=That way we cache to result and only work with that result set for the scope of the method.

noescape00: Good idea, fixed

src/Stratis.Bitcoin/P2P/PeerSelector.cs
noescape00: I didn't find a nice way to move this piece out into a method without breaking the consistency. 
After creating the extension method this piece became smaller so I think it's alright to keep it inside the `SelectPreferredPeers`

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -195,11 +198,24 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
                 return attempted;
             }
 
-            // If all the selection criteria failed to return a set of peers,
-            // then let the caller try again.
-            else
-                this.logger.LogTrace("[RETURN_NO_PEERS]");
+            List<PeerAddress> notBanned = this.NotBanned().ToList();
```
fassadlr: @noescape00 this is much better yes but I would really prefer this in a method so that we can unit test it as it is quite an important bit of code ;)

You can split this logic to a property or method called something like `HasAllPeersReachedConnectionThreshold`:

```
int attemptedReachedThresholdCount = notBanned.Count(p => p.ConnectionAttempts == PeerAddress.AttemptThreshold);
bool areAllPeersReachedThreshold = attemptedReachedThresholdCount == notBanned.Count;
```

and then, if so, create a method called `ResetConnectionAttemptsOnAllPeers()` which does:

```
this.logger.LogTrace("Resetting attempts for {0} addresses.", attemptedReachedThresholdCount);

// Reset attempts for all the peers since we've ran out of options.
foreach (PeerAddress peer in notBanned)
     peer.ResetAttempts();

this.logger.LogTrace("(-)[RESET_ATTEMPTS]");
return notBanned;
```

So after it will look like this:

```
if (HasAllPeersReachedConnectionThreshold)
    ResetConnectionAttemptsOnAllPeers();
```

fassadlr: This way its very easy to unit test.

noescape00: Ok, makes sense. 

Fixed, added unit tests.

src/Stratis.Bitcoin.Tests/P2P/PeerSelectorTests.cs
```diff
@@ -751,6 +750,49 @@ public void SelectPeersForDiscovery_WhenPeerAddressesContainsOwnIPEndoint_DoesNo
             }
         }
 
+        [Fact]
+        public void PeerSelector_HasAllPeersReachedConnectionThreshold()
```
fassadlr: @noescape00 Excellent couple of tests!

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -197,28 +203,44 @@ private IEnumerable<PeerAddress> SelectPreferredPeers()
...
-
-                // Reset attempts for all the peers since we've ran out of options.
-                foreach (PeerAddress peer in notBanned)
-                    peer.ResetAttempts();
-                
-                this.logger.LogTrace("(-)[RESET_ATTEMPTS]");
-                return notBanned;
-            }
+            
+            if (this.HasAllPeersReachedConnectionThreshold())
```
fassadlr: @noescape00 after your changes, on first call to selected peers, it'll just reset the peers but still return the empty collection, then on second call it'll select peers as per normal. I think this is actually better? If you agree I'll approve 👍

noescape00: If you think it's better- let's keep it 🙂

noescape00: Otherwise I can quickly change it

fassadlr: I think lets keep it like this, otherwise we are returning all the peers which isn't necessarily correct 👍

PR #1484, opened 2018-06-07
Preparing for Consensus manager

dangershony:
This PR is big but its mostly cosmetics (moving the rule context to the base project) and removing unnecessary properties.


src/Stratis.Bitcoin.Features.Consensus/RuleContextExtention.cs
```diff
@@ -0,0 +1,35 @@
...
+        public BlockStake BlockStake { get; set; }
+
+        public Money TotalCoinStakeValueIn { get; set; }
+
+        public uint256 HashProofOfStake { get; set; }
+
+        public uint256 TargetProofOfStake { get; set; }
+    }
+
+    public static class RuleContextExtention
```
Aprogiena: Why extension?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinbaseHeightRule.cs
```diff
@@ -17,7 +18,7 @@ public class CoinbaseHeightRule : ConsensusRule
         /// <exception cref="ConsensusErrors.BadCoinbaseHeight">Thrown if coinbase doesn't start with serialized block height.</exception>
         public override Task RunAsync(RuleContext context)
         {
-            int nHeight = context.BestBlock?.Height + 1 ?? 0;
+            int nHeight = context?.PreviousHeight + 1 ?? 0;
```
Aprogiena: It doesn't make sense if context is null.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/TransactionLocktimeActivationRule.cs
```diff
@@ -19,12 +20,12 @@ public class TransactionLocktimeActivationRule : ConsensusRule
         public override Task RunAsync(RuleContext context)
         {
             DeploymentFlags deploymentFlags = context.Flags;
-            int nHeight = context.BestBlock?.Height + 1 ?? 0;
+            int nHeight = context?.PreviousHeight + 1 ?? 0;
```
Aprogiena: It doesn't make sense if context is null.

src/Stratis.Bitcoin/Consensus/BlockValidationContext.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Net;
+using NBitcoin;
+using Stratis.Bitcoin.Configuration.Settings;
+using Stratis.Bitcoin.Consensus.Rules;
+
+namespace Stratis.Bitcoin.Consensus
+{
+    /// <summary>
+    /// Information about a block that is required for its validation.
+    /// It is used when a new block is downloaded or mined.
```
Aprogiena: No longer this is information only about a block.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -0,0 +1,70 @@
...
+        public ChainedHeader Tip { get; private set; }
+
+        private Dictionary<uint256, List<Action>> blocksRequested;
+
+        private AsyncQueue<(ChainedHeader headerFrom, ChainedHeader headerTo)> toDownloadQueue;
+
+        public ConsensusManager(Network network, ILoggerFactory loggerFactory, IChainedHeaderTree chainedHeaderTree, IChainState chainState, ConsensusSettings consensusSettings, ConcurrentChain concurrentChain, ILookaheadBlockPuller lookaheadBlockPuller)
+        {
+            this.network = network;
+            this.chainedHeaderTree = chainedHeaderTree;
```
Aprogiena: CHT is private instance of consensus manager. It is not a component running on itself it is not injected with dependency injection.

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PowCoinViewRuleTests.cs
```diff
@@ -61,7 +62,7 @@ private void AndSomeUnspentOutputs()
         {
             this.coinView = new UnspentOutputSet();
             this.coinView.SetCoins(new UnspentOutputs[0]);
-            this.ruleContext.Set = this.coinView;
+            this.ruleContext.SetItem( this.coinView);
```
Aprogiena: Extra space

src/Stratis.Bitcoin.Features.Consensus/RuleContextExtention.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public static void SetStake(this RuleContext ruleContext)
+        {
+            ruleContext.SetItem(new PosRuleContext
+            {
+                BlockStake = new BlockStake(ruleContext.Item<BlockValidationContext>().Block)
```
Aprogiena: Everywhere else you have context.BlockValidationContext, not Item

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
```diff
@@ -67,25 +31,49 @@ public class RuleContext
...
 
-        public bool IsPoS
+        public RuleContext()
         {
-            get { return this.Stake != null; }
+            this.objectsBag = new Dictionary<Type, object>();
         }
 
-        public RuleContext()
+        public T Item<T>() where T : class
```
Aprogiena: How is this better than normal inheritance?

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
```diff
@@ -67,25 +31,49 @@ public class RuleContext
         public bool SkipValidation { get; set; }
 
         /// <summary>The current tip of the chain that has been validated.</summary>
-        public ChainedHeader ConsensusTip { get; set; }
+        public ChainedHeader PreviousChainedHeader { get; set; }
```
Aprogiena: It is kind of confusing when something is current And its name is previous. Same with previous height below. Also comment previous height.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Yes I am aware, however then we will find it problematic to test CM on its own, we could hack a bit and allow to set CHT as an internal property.

Also to keep the implementation internal I had to add an interface

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
dangershony: Yeah I am not sure which is better, I decide on this approach as if we use inheritance each rule would have to cast the context (which is ok I guess) but then the engine will need to know what cobtext to creat.

Possibly inheritance is also an option.

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
dangershony: BestChainedheader
Is better?

Aprogiena: Yes, but even the old name of consensus tip was fine.

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
Aprogiena: My vote is on inheritance. For me, this is not nice.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: Are you sure we will have problems with testing? Because I don't think there will be any problems with CM to normally initialise its CHT instance and use it the full power even during tests, because CHT is quite independent of anything else.

src/Stratis.Bitcoin.Features.Consensus/RuleContextExtention.cs
dangershony: yeah good point I will remove this class

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: If it is a problem ... I cant say right now, testing CHT in all the tests of CM perhaps not a bad idea might be significantly harder.
Perhaps we try first as a hard instance and if we find mocking  is better we just move to an injected interface.

src/Stratis.Bitcoin.Features.Consensus/RuleContextExtention.cs
dangershony: I will fix it back

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: Yes, let's do it like that. Let's assume no problems and only solve them when we are hit back.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -151,6 +149,12 @@ protected void UpdateUTXOSet(RuleContext context, Transaction transaction)
             this.Logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Check whether the transaction is part of the protocol (Coinbase or Coinstake)
```
Aprogiena: Missing `.`

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -151,6 +149,12 @@ protected void UpdateUTXOSet(RuleContext context, Transaction transaction)
...
+        /// </summary>
+        /// <param name="transaction">The transaction to check.</param>
+        protected abstract bool IsProtocolCoin(Transaction transaction);
```
Aprogiena: IsProtocolTransaction instead?

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
dangershony: My argument against inheritance is that this is more flexible, if a new rule needs to set something in the bag that will be used by a later rule this becomes possible with inheritance this is much harder.

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
dangershony: done

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
dangershony: Yeah sure

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
Aprogiena: I don't see such a problem and I also don't consider this a cleaner solution. For example it has a problem if you want to add to thinks of the same type. I would certainly go for inheritance, but if you are unconvinced, please ask more people so that we have more opinions.

src/Stratis.Bitcoin/Consensus/ValidationContext.cs
```diff
@@ -7,10 +7,10 @@
 namespace Stratis.Bitcoin.Consensus
 {
     /// <summary>
-    /// Information about a block that is required for its validation.
+    /// A context that is used by consensus that is required for validation, in case validation failed the <see cref="Error"/> property will be set.
```
Aprogiena: probably need to use ValidationContext.Error in see

src/Stratis.Bitcoin.Features.Consensus/RuleContextOverrides.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            this.BlockStake = blockStake;
+        }
+
+        public PosRuleContext(ValidationContext validationContext, NBitcoin.Consensus consensus, ChainedHeader consensusTip) 
+            : base(validationContext, consensus, consensusTip)
+        {
+        }
+
+        public BlockStake BlockStake { get; set; }
```
noescape00: Please comment those values

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -51,7 +50,100 @@ public interface IChainedHeaderValidator
     /// This class is not thread safe and it the role of the component that uses this class to prevent race conditions.
     /// </para>
     /// </remarks>
-    internal sealed class ChainedHeaderTree
+    public interface IChainedHeaderTree
```
noescape00: Why moving it to the interface?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
noescape00: If we would like to mock CHT we can always just set private property in the test using the reflection

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
```
noescape00: redundant using

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Yeah still not complete, in discussion.

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
```diff
@@ -23,41 +21,20 @@ public class RuleContext
...
+        /// <summary>Indicate the block was created by our node.</summary>
+        public bool MinedBlock { get; set; }
 
         /// <summary>Whether to skip block validation for this block due to either a checkpoint or assumevalid hash set.</summary>
         public bool SkipValidation { get; set; }
 
         /// <summary>The current tip of the chain that has been validated.</summary>
         public ChainedHeader ConsensusTip { get; set; }
 
+        /// <summary>The height of the previous block.</summary>
```
Aprogiena: But is it not actually the height of the consensus tip above?

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
```diff
@@ -70,12 +47,7 @@ public RuleContext(ValidationContext validationContext, NBitcoin.Consensus conse
             this.ConsensusTip = consensusTip;
             this.PreviousHeight = consensusTip.Height;
```
Aprogiena: Like here the naming is weird.

noescape00:
There are merge conflicts.

src/Stratis.Bitcoin/Consensus/Rules/RuleContext.cs
dangershony: Its sort of the same thing, but I can change I dont mind actually

src/Stratis.Bitcoin.Features.Consensus/RuleContextOverrides.cs
dangershony: yes will do

src/Stratis.Bitcoin/Consensus/Rules/IConsensusRules.cs
```diff
@@ -69,6 +66,27 @@ public interface IConsensusRules
...
+        /// </summary>
+        void Initialize();
+
+        /// <summary>
+        /// Execute the logic in the current rule.
+        /// If the validation of the rule fails a <see cref="ConsensusErrorException"/> will throw.
+        /// </summary>
+        /// <param name="context">The context that has all info that needs to be validated.</param>
+        /// <returns>The execution task.</returns>
+        Task RunAsync(RuleContext context);
```
Aprogiena: We don't use returns tag if the method returns plain task.

dangershony: Please done review the last push I might rule it back.

Aprogiena: No problem

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -28,41 +30,55 @@ public class ConsensusManager
...
             this.chainState = chainState;
             this.consensusSettings = consensusSettings;
             this.concurrentChain = concurrentChain;
             this.lookaheadBlockPuller = lookaheadBlockPuller;
+            this.consensusRules = consensusRules;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            this.chainedHeaderTree = new ChainedHeaderTree(network, loggerFactory, chainedHeaderValidator, checkpoints, chainState, consensusSettings);
+
+            this.blocksRequested = new Dictionary<uint256, List<Action>>();
```
Aprogiena: This will not be action, it will be a delegate, but you probably just use it as a placeholder right?

dangershony: Yes when I need it I will work on this 👍

Aprogiena:
I have been following this 1 and it is probably okay, no tests or sync

noescape00:
I'll rerun the tests and sync the chains. If all fine- will push. 
@dangershony said it's ready.

PR #1487, opened 2018-06-07
Moved a few properties from ConsensOptions implementations to Consensus

bokobza:
Moved 
* CoinbaseMaturity
* PremineReward
* PremineHeight
* ProofOfWorkReward
from ConsensusOptions to Consensus.
This way these properties are more accessible. We need this if we want to create new networks with custom mining settings. For example POW network with 50M coins, no reward.

I have no idea why this test is failing: https://github.com/stratisproject/StratisBitcoinFullNode/compare/master...bokobza:consensus-options-move?expand=1#diff-b3695281628304a7bdb952a6e293ceb8R179
If someone has any idea, I'll happily take it.

All the other unit tests run ok, the integration ones are broken.

### DIFF
--- a/src/NBitcoin/Consensus.cs
+++ b/src/NBitcoin/Consensus.cs
@@ -52,6 +52,16 @@ namespace NBitcoin
         /// </summary>
         public Money ProofOfWorkReward { get; set; }
 
+        /// <summary>
+        /// The reward that goes to the miner when a block is mined using proof-of-stake.
+        /// </summary>
+        public Money ProofOfStakeReward { get; set; }
+
+        /// <summary>
+        /// Maximal length of reorganization that the node is willing to accept, or 0 to disable long reorganization protection.
+        /// </summary>
+        public uint MaxReorgLength { get; set; }
+
         public ConsensusOptions Options { get; set; }
 
         public class BuriedDeploymentsArray
--- a/src/NBitcoin/Consensus.cs
+++ b/src/NBitcoin/Consensus.cs
@@ -30,6 +30,26 @@ namespace NBitcoin
         {
         }
 
+        /// <summary>
+        /// How many blocks should be on top of the block with coinbase transaction until it's outputs are considered spendable.
+        /// </summary>
+        public long CoinbaseMaturity { get; set; }
+
+        /// <summary>
+        /// Amount of coins mined when a new network is bootstrapped.
+        /// </summary>
+        public Money PremineReward { get; set; }
+
+        /// <summary>
+        /// The height of the block in which the pre-mined coins should be,
+        /// </summary>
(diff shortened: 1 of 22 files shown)


src/NBitcoin/Consensus.cs
```diff
@@ -30,6 +30,26 @@ public class ConsensusOptions
         {
         }
 
+        /// <summary>
+        /// How many blocks should be on top of the block with coinbase transaction until it's outputs are considered spendable.
```
dangershony: >How many blocks should be on top of a coinbase transaction until it's outputs are considered spendable

src/NBitcoin/Consensus.cs
```diff
@@ -30,6 +30,26 @@ public class ConsensusOptions
...
+        /// </summary>
+        public long CoinbaseMaturity { get; set; }
+
+        /// <summary>
+        /// Amount of coins mined when a new network is bootstrapped.
```
dangershony: maybe add to the comment that zero means disabled? (same for the other guy)

src/NBitcoin/Consensus.cs
```diff
@@ -30,6 +30,26 @@ public class ConsensusOptions
...
+
+        /// <summary>
+        /// The height of the block in which the pre-mined coins should be,
+        /// </summary>
+        public long PremineHeight { get; set; }
+
+        /// <summary>
+        /// The reward that goes to the miner when a block is mined using proof-of-work.
+        /// </summary>
+        public Money ProofOfWorkReward { get; set; }
```
dangershony: Should we also bring here POS reward?

src/Stratis.Bitcoin.Features.Consensus/ConsensusOptions.cs
```diff
@@ -27,13 +21,8 @@ public class PosConsensusOptions : PowConsensusOptions
         /// </summary>
         public PosConsensusOptions()
         {
-            this.MaxMoney = long.MaxValue;
-            this.CoinbaseMaturity = 50;
-
-            this.ProofOfWorkReward = Money.Coins(4);
+            this.MaxMoney = long.MaxValue;
```
dangershony: I think this.MaxReorgLength can also go in consensus its not necessarily a pos thing.
Well it is but it could potentially be used on a pos netwrok

src/NBitcoin/Consensus.cs
bokobza: thanks

src/NBitcoin/Consensus.cs
bokobza: sure

dangershony:
I  have synced networks for 10 minutes looks good

PR #1490, opened 2018-06-08
Made the passphrase optional in ExtKey generation fro mnemonic

bokobza:
The passphrase can very well be optional when creating a mnemonic.
Also changed the names of these methods as they were not appropriate.


monsieurleberre:
the tests which are failing should have now by fixed by the mighty @noescape00 , I trust you guys ;)

PR #1492, opened 2018-06-08
Adding a block size to the Block property

dangershony:


src/NBitcoin/Block.cs
```diff
@@ -220,6 +220,9 @@ public partial class Block : IBitcoinSerializable
 
         private BlockHeader header;
 
+        /// <summary>The size of the block in bytes, the block must be serialized for this property to be set.</summary>
+        public long? BlockSize { get; protected set; }
```
quantumagi: The `BlockSize` is unqualified and public - i.e. it is visible to any code to make assumptions as to the `TransactionOptions` and `Network` the size is for.

src/NBitcoin/Block.cs
```diff
@@ -264,6 +267,23 @@ public virtual void ReadWrite(BitcoinStream stream)
         {
             stream.ReadWrite(ref this.header);
             stream.ReadWrite(ref this.transactions);
+
+            if (stream.Serializing)
+                this.BlockSize = stream.Counter.WrittenBytes;
+        }
+
+        public long GetBlockSize(Network network, TransactionOptions transactionOptions = TransactionOptions.None)
```
quantumagi: If this is called twice with different `TransactionOptions` that may affect the block size it will erroneously return the wrong size.

dangershony: Yes correct but if options is specified it will likely stay as a block with options.

src/NBitcoin/Block.cs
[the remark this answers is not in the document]
dangershony: hmmm the benefit of this guy is to get the size without needing to specify Network (it will already be specified in the BitcoinStream).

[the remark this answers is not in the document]
dangershony: We could actually remove the method `GetBlockSize` and let the caller deserialize in order to set that property.

dangershony: I didn't understand the comment.

quantumagi: E.g. Recording the BlockSizeNetwork and BlockSizeTransactionOptions would allow to check what the recorded BlockSize contains - i.e. qualify it. Not that I am suggesting that would be the best or safest approach.

PR #1495, opened 2018-06-08
Codestyle fixes

noescape00:
Fixed based on the output of the static code analysis (redundant usings, vars) 

No functionality changes


bokobza:
this is really nice, but, can we do it more slowly?
Please start with projects that no one is working on, so that the other devs don't have to do too many merging.
I hope this has been done automatically. If it hasn't not, I'll review it as is.

noescape00:
Closing since it makes sense to split this PR into several small ones that cause less disruption

PR #1497, opened 2018-06-10
Utilize BlockSize property in the block store queue

noescape00:
Minor blockstorequeue optimization which allows us to avoid serialization of every block in order to get block size.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
@@ -35,7 +35,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         internal const int BatchThresholdSizeBytes = 5 * 1000 * 1000;
 
         /// <summary>The current batch size in bytes.</summary>
-        private int currentBatchSizeBytes;
+        private long currentBatchSizeBytes;
 
         /// <summary>The highest stored block in the repository.</summary>
         private ChainedHeader storeTip;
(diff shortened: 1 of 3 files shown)


Aprogiena:
Looks good, NTNS

noescape00:
>NTNS?

Not tested, not synced I guess

>Are the integration tests passing for you @noescape00 ?

One moment, let me check that. I only synced the chains.

Aprogiena:
> Not tested, not synced I guess

yes

noescape00:
@bokobza fixed the unit tests. Integration tests are failing but not because of this PR.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -258,7 +258,7 @@ public void AddToPending(BlockPair blockPair)
 
                     batch.Add(item);
 
-                    this.currentBatchSizeBytes += item.Block.GetSerializedSize();
+                    this.currentBatchSizeBytes += item.Block.BlockSize.Value;
```
dangershony: you could potentiall do a 
```
this.currentBatchSizeBytes += tem.Block.BlockSize ?? item.Block.GetSerializedSize()
```

noescape00: It's not possible for a blocksize to be null here- in case it was downloaded it was deserialized. If it was mined- we checked it against the max size rule that serialized it.

PR #1498, opened 2018-06-10
Added UnconsumedBlocksDataBytes

noescape00:


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -52,6 +50,16 @@ public interface IChainedHeaderValidator
     /// </remarks>
     public interface IChainedHeaderTree
     {
+        /// <summary>
+        /// Gets size of unconsumed blocks data in bytes.
```
Aprogiena: It is not method, so probably it does not "get" the size but it is the size.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -358,6 +370,38 @@ public List<int> ConsensusTipChanged(ChainedHeader newConsensusTip)
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(newConsensusTip), newConsensusTip);
 
+            ChainedHeader oldConsensusTip = this.GetConsensusTip();
+            ChainedHeader fork = newConsensusTip.FindFork(oldConsensusTip);
+            ChainedHeader currentHeader = newConsensusTip;
+
```
Aprogiena: It might be nice to have all three guys logged – the old consensus tip, the new one, and the fork.

noescape00: Fixed

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Fixed

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -358,6 +370,40 @@ public List<int> ConsensusTipChanged(ChainedHeader newConsensusTip)
...
+            this.logger.LogTrace("Old consensus tip: '{0}', new consensus tip: '{1}', fork point: '{2}'.", oldConsensusTip, newConsensusTip, fork);
+
+            // Consider blocks that became a part of our best chain as consumed.
+            while (currentHeader != fork)
+            {
+                this.UnconsumedBlocksDataBytes -= currentHeader.Block.BlockSize.Value;
+                this.logger.LogTrace("Size of unconsumed block data is decreased by {0}, new value is {1}.", currentHeader.Block.BlockSize.Value, this.UnconsumedBlocksDataBytes);
```
dangershony: Cant `Block` be null between the two tips?

noescape00: No. We can't delete block data ahead of our consensus tip. Only when we switch to it we call this method and later in the code we call CleanOldBlockDataFromMemory that might delete that block data. 

So at the point of usage here we have a guarantee that the block data is not null.

PR #1501, opened 2018-06-10
Modify delay in inventory sends in mempool

mikedennis:
This PR is related to issues discussed in #1250 

It includes the following changes to mempool behavior:
* Removal of the delay implemented in CanSend (there is already a delay in asyncloop for SendTrickleAsync)
* Change to delay in asyncloop to be 5 seconds instead of 10 seconds as per bitcoin core.


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -104,29 +104,12 @@ public class MempoolBehavior : NetworkPeerBehavior
         /// <summary>Time of last memory pool request in unix time.</summary>
         public long LastMempoolReq { get; private set; }
 
-        /// <summary>Time of next inventory send in unix time.</summary>
-        public long NextInvSend { get; set; }
-
         /// <summary>Whether memory pool is in state where it is ready to send it's inventory.</summary>
```
noescape00: This comment no longer holds

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -104,29 +104,12 @@ public class MempoolBehavior : NetworkPeerBehavior
...
-                if (this.NextInvSend < this.mempoolManager.DateTimeProvider.GetTime())
-                {
-                    sendTrickle = true;
-                    // Use half the delay for outbound peers, as there is less privacy concern for them.
-                    this.NextInvSend = this.mempoolManager.DateTimeProvider.GetTime() + 10;
-                    // TODO: PoissonNextSend(nNow, InventoryBroadcastInterval >> !pto->fInbound);
-                }
-
-                return sendTrickle;
+                return (this.AttachedPeer?.PeerVersion?.Relay ?? false);
```
noescape00: Maybe remove the property completely? 

CanSend is called from 2 places- when we call SendTrickleAsync - here just modify the linq expression to avoid selecting peers with CanRelay ==false, and in 2nd place just remove it since it can't happen (and even if CanRelay==false peer asks us for our mempool we don't mind sending it).

dangershony: I am fine with that but we still need to honor `Relay` somehow.
And it might be we need to also add this flag in mempool reply code

src/Stratis.Bitcoin.Features.MemoryPool/MempoolSignaled.cs
```diff
@@ -128,8 +128,8 @@ public void Start()
                     await behaviour.SendTrickleAsync().ConfigureAwait(false);
             },
             this.nodeLifetime.ApplicationStopping,
-            repeatEvery: TimeSpans.TenSeconds,
-            startAfter: TimeSpans.TenSeconds);
+            repeatEvery: TimeSpans.FiveSeconds,
+            startAfter: TimeSpans.FiveSeconds);
```
dangershony: I think the start after should still be 10 seconds (even more) so we give the node enough time to start

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
noescape00: I looked at the usages of CanSend and basically it's needed in one place- when we send trickle (and there we have peer in our hands so we can select peers with Relay == true).

In the `SendMempoolPayloadAsync` we don't really need it. Actually quite the opposite- we are disallowing light wallets to be able to ask if their tx was included in the mempool (currently we don't do that because of the privacy issues but I can see why someone would like to do that).

dangershony: yeah we can indeed filter at the trickle method.

dangershony:
@mikedennis please fix conflicts

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -534,7 +505,7 @@ public void RelayTransaction(uint256 hash)
         {
             this.logger.LogTrace("()");
 
-            if (!this.CanSend)
+            if (!this.AttachedPeer?.PeerVersion?.Relay ?? true)
```
noescape00: Instead just don't select this peer for sending the Trickle:

```
                var behaviours = peers.Select(s => s.Behavior<MempoolBehavior>());
                foreach (var behaviour in behaviours)
                    await behaviour.SendTrickleAsync().ConfigureAwait(false);
```

in MempoolSignaled.


do there:

```
var behaviours = peers.Where(x => x.PeerVersion != null ? x.PeerVersion.Relay : true).Select(x => x.Behavior<MempoolBehavior>());
```

PR #1502, opened 2018-06-10
Codestyle cleanup 2

noescape00:
Vars for stratis features, reordering usings in alphabetic order


Aprogiena:
Labelling with medium because of the large number of files.

PR #1504, opened 2018-06-11
fix rpc methods tests

bokobza:
Only 4 tests are failing due to a null reference exception.

image


bokobza:
Thanks @noescape00

dangershony:
Ok thanks I will look at the other tests

PR #1505, opened 2018-06-11
Fix tests

dangershony:
image


dangershony:
Including NBitcoin tests
image

PR #1508, opened 2018-06-11
Codestyle cleanup 5

noescape00:
`.this` qualifiers and redundant namespaces when calling a method (DoSmth() instead of Lib1.SmthUtils.DoSmth()) for NBitcoin


src/NBitcoin.Tests/ChainTests.cs
```diff
@@ -187,7 +187,7 @@ public void CanFindFork()
 
         private void AssertFork(ConcurrentChain chain, ConcurrentChain chain2, ChainedHeader expectedFork)
         {
-            ChainedHeader fork = this.FindFork(chain, chain2);
+            ChainedHeader fork = FindFork(chain, chain2);
```
bokobza: hmm I wonder why this. has been removed from here and other places.

noescape00: Hm, that's really odd. 

I've saw resharper chocking on a file a two but executing the rule vise versa is something I didn't see. 

I've fixed  those places manually.

bokobza:
I just restarted the Mac on Travis build, the testSigHashes test fails sometimes.

noescape00:
> the testSigHashes test fails sometimes.

Yes, it is failing when ran with other tests for a month or so.

noescape00:
I reran the tests on my local machine- all passes. 
Let's also wait for the CI just in case.

PR #1509, opened 2018-06-11
Moved MaxMoney to Consensus

bokobza:


bokobza:
image

src/NBitcoin/Networks/StratisMain.cs
```diff
@@ -60,6 +60,7 @@ public StratisMain()
             this.Consensus.ProofOfWorkReward = Money.Coins(4);
             this.Consensus.ProofOfStakeReward = Money.COIN;
             this.Consensus.MaxReorgLength = 500;
+            this.Consensus.MaxMoney = long.MaxValue;
```
noescape00: I'm not sure if we should max out here. 
For BTC we have 21 mil since it's the max supply that ca exist. 
Maybe for stratis we can just use a reasonably high value like 150 mil? 

Reason behind MaxMoney is to quickly check if the tx inputs are invalid so setting it to long.MaxValue basically will make this check redundant for stratis network.

noescape00: Other than that I'm fine with this PR.

bokobza: I agree that long.Max is big, but I just wanted tom move it without changing the functionality.

noescape00: Ok, in that case cool, let's merge it.

bokobza:
Thanks @noescape00

PR #1510, opened 2018-06-12
Adde coinTicker to Network

bokobza:


bokobza:
Not really part of the protocol but can be useful in many instances.

PR #13, opened 2017-02-08
BlockPulling

nopara73:
I am working on BlockPulling for an other project, I'd like to keep my code in sync with you if possible.  
  
- Remove redundant qualifier
- Remove unnecessary whitespaces
- Use nameof expression
- Remove redundant ()
- Use explicit private modifiers
- Removed unused directives

PR #1512, opened 2018-06-12
Added new test to cover issue #13. Presenting chains with different chain work

fenix2222:
Create 2 chains Chain A and Chain B, where Chain A has more chain work than B. Connect both chains to chain header tree. Consensus tip should be set to chain A. Now extend / update Chain B to make it have more chain work. Attempt to connect B again. Consensus tip should be set to chain B.

See https://github.com/stratisproject/StratisBitcoinFullNode/issues/1321#issuecomment-389155904

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -65,11 +65,10 @@ namespace Stratis.Bitcoin.Tests.Consensus
                                         ? previousHeader.Header.Bits 
                                         : this.ChangeDifficulty(previousHeader, difficultyAdjustmentDivisor);
                     header.Nonce = (uint)new Random().Next(1, 100000);
-                    var newHeader = new ChainedHeader(header, header.GetHash(), previousHeader)
-                    {
-                        Block = Block.Parse(TestBlockHex, Network.StratisTest)
-                    };
-
+                    var newHeader = new ChainedHeader(header, header.GetHash(), previousHeader);
+                    Block block = this.Network.Consensus.ConsensusFactory.CreateBlock();
+                    block.GetSerializedSize();
+                    newHeader.Block = block;
                     previousHeader = newHeader;
                 }
 
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -33,10 +33,10 @@ namespace Stratis.Bitcoin.Tests.Consensus
                 return this.ChainedHeaderTree;
             }
 
-            internal Target ChangeDifficulty(ChainedHeader header, int difficultyAdjustmentScore)
+            internal Target ChangeDifficulty(ChainedHeader header, int difficultyAdjustmentDivisor)
             {
                 BigInteger newTarget = header.Header.Bits.ToBigInteger();
-                newTarget = newTarget.Divide(BigInteger.ValueOf(difficultyAdjustmentScore)); 
+                newTarget = newTarget.Divide(BigInteger.ValueOf(difficultyAdjustmentDivisor)); 
                 return new Target(newTarget);
             }
 
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -256,12 +256,12 @@ namespace Stratis.Bitcoin.Tests.Consensus
             List<BlockHeader> listOfChainABlockHeaders = ctx.ChainedHeaderToList(chainATip, commonChainSize + chainAExtension);
             List<BlockHeader> listOfChainBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, commonChainSize + chainBExtension);
 
-            // Chain A is presented by peer 1. DownloadTo should be chain A tip
+            // Chain A is presented by peer 1. DownloadTo should be Chain A tip
             ConnectNewHeadersResult connectNewHeadersResult = cht.ConnectNewHeaders(1, listOfChainABlockHeaders);
             ChainedHeader chainedHeaderTo = connectNewHeadersResult.DownloadTo;
             chainedHeaderTo.HashBlock.Should().Be(chainATip.HashBlock);
 
-            // Chain B is presented by peer 2. DownloadTo should be not set as chain
+            // Chain B is presented by peer 2. DownloadTo should be not set, as Chain
             // B has less chain work
             connectNewHeadersResult = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
             chainedHeaderTo = connectNewHeadersResult.DownloadTo;
@@ -272,7 +272,7 @@ namespace Stratis.Bitcoin.Tests.Consensus
             chainBTip = ctx.ExtendAChain(chainBAdditionalBLocks, chainBTip); // ie. (h1=h2=h3=h4)=b5=b6=b7=b8=b9=b10
             listOfChainBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, commonChainSize + chainBExtension + chainBAdditionalBLocks);
 
-            // Chain B is presented by peer 2 again. DownloadTo should be now be chain B tip
+            // Chain B is presented by peer 2 again. DownloadTo should now be Chain B tip
             // as B has more chain work than chain A
             connectNewHeadersResult = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
             chainedHeaderTo = connectNewHeadersResult.DownloadTo;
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +231,52 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            ChainedHeader chainBTip = ctx.ExtendAChain(chainBExtension, commonChainTip); // ie. (h1=h2=h3=h4)=b5=b6
+            List<BlockHeader> listOfChainABlockHeaders = ctx.ChainedHeaderToList(chainATip, commonChainSize + chainAExtension);
+            List<BlockHeader> listOfChainBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, commonChainSize + chainBExtension);
+
+            // Chain A is presented by peer 1. DownloadTo should be Chain A tip
+            ConnectNewHeadersResult connectNewHeadersResult = cht.ConnectNewHeaders(1, listOfChainABlockHeaders);
+            ChainedHeader chainedHeaderTo = connectNewHeadersResult.DownloadTo;
+            chainedHeaderTo.HashBlock.Should().Be(chainATip.HashBlock);
+
+            // Chain B is presented by peer 2. DownloadTo should be not set, as Chain
```
Aprogiena: Please finish all your comments sentences with dots not just some.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +231,52 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            // B has less chain work
+            connectNewHeadersResult = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
+            chainedHeaderTo = connectNewHeadersResult.DownloadTo;
+            chainedHeaderTo.Should().BeNull();
+
+            // Add more chain work and blocks into chain B
+            const int chainBAdditionalBLocks = 4;
```
Aprogiena: L should be lowercase

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +231,52 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            chainBTip = ctx.ExtendAChain(chainBAdditionalBLocks, chainBTip); // ie. (h1=h2=h3=h4)=b5=b6=b7=b8=b9=b10
+            listOfChainBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, commonChainSize + chainBExtension + chainBAdditionalBLocks);
```
Aprogiena: I think what we wanted to do here is only to present the new blocks as it would happen normally with the peer.You are presenting the whole chain instead which is not very realistic. It wouldn't be harmful, just intention was to present the new headers only.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +231,52 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // Chain B is presented by peer 2 again. DownloadTo should now be Chain B tip
+            // as B has more chain work than chain A
+            connectNewHeadersResult = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
+            chainedHeaderTo = connectNewHeadersResult.DownloadTo;
+            chainedHeaderTo.HashBlock.Should().Be(chainBTip.HashBlock);
```
Aprogiena: Also please check download from, especially after we no longer will be presenting the full chain.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -30,16 +33,30 @@ internal ChainedHeaderTree CreateChainedHeaderTree()
...
+            internal Target ChangeDifficulty(ChainedHeader header, int difficultyAdjustmentDivisor)
             {
+                BigInteger newTarget = header.Header.Bits.ToBigInteger();
+                newTarget = newTarget.Divide(BigInteger.ValueOf(difficultyAdjustmentDivisor)); 
+                return new Target(newTarget);
+            }
+
+            public ChainedHeader ExtendAChain(int count, ChainedHeader chainedHeader = null, int difficultyAdjustmentDivisor = 1)
+            {
+                if (difficultyAdjustmentDivisor == 0) throw new ArgumentException("Divisor cannot be 0");
```
Aprogiena: Code style – after return or throw their should be empty line

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -30,16 +33,30 @@ internal ChainedHeaderTree CreateChainedHeaderTree()
...
 
                 for (int i = 0; i < count; i++)
                 {
                     BlockHeader header = this.Network.Consensus.ConsensusFactory.CreateBlockHeader();
                     header.HashPrevBlock = previousHeader.HashBlock;
-                    header.Bits = previousHeader.Header.Bits - 1000; // just increase difficulty.
+                    header.Bits = difficultyAdjustmentDivisor == 1 
+                                        ? previousHeader.Header.Bits 
+                                        : this.ChangeDifficulty(previousHeader, difficultyAdjustmentDivisor);
+                    header.Nonce = (uint)new Random().Next(1, 100000);
```
Aprogiena: Once in awhile the test will fail just by chance of generating the same nonce. It would be better to keep a global variable did you increment with interlocked

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -30,16 +33,30 @@ internal ChainedHeaderTree CreateChainedHeaderTree()
...
                     var newHeader = new ChainedHeader(header, header.GetHash(), previousHeader);
+                    Block block = this.Network.Consensus.ConsensusFactory.CreateBlock();
+                    block.GetSerializedSize();
+                    newHeader.Block = block;
```
Aprogiena: Is the block needed for something here? It is invalid anyway and does not match the header

fenix2222: Yes, as per slack conversation, I am getting NullReferenceException when calling this `cht.ConsensusTipChanged(chainATip);`. I need to set fake block

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +247,63 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            ChainedHeader chainBTip = ctx.ExtendAChain(chainBExtension, commonChainTip); // ie. (h1=h2=h3=h4)=b5=b6
+            List<BlockHeader> listOfChainABlockHeaders = ctx.ChainedHeaderToList(chainATip, commonChainSize + chainAExtension);
+            List<BlockHeader> listOfChainBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, commonChainSize + chainBExtension);
+
+            // Chain A is presented by peer 1. DownloadTo should be chain A tip.
+            ConnectNewHeadersResult connectNewHeadersResult = cht.ConnectNewHeaders(1, listOfChainABlockHeaders);
+            ChainedHeader chainedHeaderTo = connectNewHeadersResult.DownloadTo;
+            chainedHeaderTo.HashBlock.Should().Be(chainATip.HashBlock);
+
+            // Set chain A tip as a consensus tip
```
Aprogiena: .

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
fenix2222: better to keep a global variable `did you increment with interlocked`. Can you elaborate, I am not understanding this.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -43,16 +46,18 @@ internal Target ChangeDifficulty(ChainedHeader header, int difficultyAdjustmentD
...
                 for (int i = 0; i < count; i++)
                 {
+                    Interlocked.Increment(ref nonceValue);
                     BlockHeader header = this.Network.Consensus.ConsensusFactory.CreateBlockHeader();
                     header.HashPrevBlock = previousHeader.HashBlock;
                     header.Bits = difficultyAdjustmentDivisor == 1 
                                         ? previousHeader.Header.Bits 
                                         : this.ChangeDifficulty(previousHeader, difficultyAdjustmentDivisor);
-                    header.Nonce = (uint)new Random().Next(1, 100000);
+                    header.Nonce = (uint) nonceValue;
```
Aprogiena: Use the return value from interlocked to prevent race conditions. Also after bracket here is no space

PR #1513, opened 2018-06-12
Chained Header Tree issues 5, 8, 9 initial check in (#1321)

thefazzer:
**Chained Header Tree issues 5, 8, 9 initial check in (#1321)**


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
             Assert.True(peerTipsByPeerIdBefore.Values.Count == peerTipsByPeerIdAfter.Values.Count);
         }
 
+        /// <summary>
```
majikandy: I don't think this is a good idea to use Summaries here to describe the scenario of the test.

This can be embedded into the test itself with better naming of the method (so that it shows in test runners like resharper)  but also in the code itself with method names.

Disconnecting the scenario from the code means that any developer needing to change this in the future has to read the scenario and try to correlate it with the blocks of code and this cognitive load is very high.

If we are writing tests just to sit there as regression and never come back to them then sure it just doesn;t matter - since we know that software almost always changes then lets make the blocks mean something.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+        /// Check that next pointers were created correctly.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_HeadersFromTwoPeersWithFork_ShouldCreateBlocksForNewHeaders()
+        {
+            // Setup
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            ChainedHeader chainTip = ctx.ExtendAChain(10);
+            cht.Initialize(chainTip, true);
```
majikandy: please expand variable name

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            ctx.ChainStateMock.Setup(s => s.ConsensusTip).Returns(chainTip);
+
+            // Default peer is on tip
+            Assert.True(cht.GetPeerTipsByPeerId()[ChainedHeaderTree.LocalPeerId].Equals(chainTip.HashBlock));
+
+            // Peer 1 forks two blocks back
+            BlockHeader forkedBlock = ctx.GetNewBlock(chainTip.Previous.Previous).Header;
```
majikandy: forkedBlock -> forkedBlockHeader

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+            cht.ConnectNewHeaders(1, new List<BlockHeader>() { forkedBlock });
+
+            uint256 peerOneTipHash = cht.GetPeerTipsByPeerId()[1];
+            uint256 hashOfForkedBlock = forkedBlock.GetHash();
+
+            ChainedHeader chainedHeaderOfForkedBlock = cht.GetChainedHeadersByHash()[hashOfForkedBlock];
+            ChainedHeader fork = chainTip.FindFork(chainedHeaderOfForkedBlock);
```
majikandy: FindFork returns a ChainedHeader?   Doesn't sounds like you would logically think that

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+            fork.Should().NotBeNull();
+
+            peerOneTipHash.Should().Be(hashOfForkedBlock, "Peer should be on the fork but wasn't");
+
+            // Next pointers were created correctly
+            Assert.Equal(2, fork.Next.Count);   // main chain + 1 branch
```
majikandy: can this be represented without the comment, by giving the magic number 2 that meaning

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +282,82 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            List<BlockHeader> listOfExistingHeaders = ctx.ChainedHeaderToList(chainTip, 10);
+            cht.ConnectNewHeaders(2, listOfExistingHeaders);
+            ChainedHeader peerTwoTip = cht.GetChainedHeaderByPeerId(2);
+
+            // Peer B / Peer 2 is on the best chain.
+            peerTwoTip.HashBlock.Should().BeEquivalentTo(chainTip.HashBlock);
+            peerTwoTip.ChainWork.Should().BeEquivalentTo(chainTip.ChainWork);
+
+            // After that peer C / Peer 3 presents an alternative chain with a forkpoint below B's Tip.
```
majikandy: I don't understand what "After that" means here

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +282,82 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+        }
+
+        /// <summary>
+        /// Issue 9 @ Create chained header tree component #1321
+        /// Check that everything is always consumed.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_MultiplePeers_CheckEverythingIsConsumed()
+        {
+            var ctx = new TestContext();
```
majikandy: quite a lot of repetition of blocks of code that look very similar.  Consider DRY and use methods

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +282,82 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            ChainedHeader chainTip = ctx.ExtendAChain(11);
+            cht.Initialize(chainTip, true);
+            List<BlockHeader> listOfNewHeaders = ctx.ChainedHeaderToList(chainTip, 11);
+
+            List<BlockHeader> peer1Headers = listOfNewHeaders.GetRange(0, 6);
+            ConnectNewHeadersResult connectedNewHeadersPeer1 = cht.ConnectNewHeaders(1, peer1Headers);
+
+            // 5 unprocessed headers
+            Assert.True(chainTip.Height - connectedNewHeadersPeer1.Consumed.Height == 5);
```
majikandy: use a comparison to 5 rather than assert true - it will give a better output on failure - perhaps 
```
var unprocessedHeaderCount = chainTip.Height - connectedNewHeadersPeer1.Consumed.Height
unprocessedHeaderCount.ShouldBe(5)
//OR Assert.Equal(5, unprocessedHeaderCount);
```

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+        /// <summary>
+        /// Issue 5 @ Create chained header tree component #1321
+        /// 2 peers should supply headers - one of the peers creates a fork.
+        /// Make sure that ChainedBlock is not created more than once for the same header.
+        /// Check that next pointers were created correctly.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_HeadersFromTwoPeersWithFork_ShouldCreateBlocksForNewHeaders()
+        {
+            // Setup
```
majikandy: //Setup 
is not expressive enough. and doesn't add anything here

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
```
majikandy: `var chainedHeaderTree = testContext.CreateChainedHeaderTree();`  would be easier to read.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: Some of scenarios are too complex to put in a test name, so I'd keep the desriptions there.

Also it helps during the review to compare the intention vs implementation.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: I think it's in the method desription- it returns last common block between 2 branches

majikandy: I see, interesting...  FindForkHeader more appropriate? or happy with FindFork since it couldn't be anything other than a header?

noescape00: >or happy with FindFork since it couldn't be anything other than a header?

yeah, could be only header

noescape00:
>Overall these tests are quite hard to follow and a lot of knowledge is embedded in the heads of developers who understand how the Class works

Well, yes, kinda. It's a huge class which deals with consensus code. Not sure if there is any other way except for learning the class and then checking the tests. 

Ofc we can always cover every test with comments that will explain every detail bat that would be too much and we will have to copypaste a lot of it between the tests.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+            cht.ConnectNewHeaders(1, new List<BlockHeader>() { forkedBlock });
+
+            uint256 peerOneTipHash = cht.GetPeerTipsByPeerId()[1];
+            uint256 hashOfForkedBlock = forkedBlock.GetHash();
+
+            ChainedHeader chainedHeaderOfForkedBlock = cht.GetChainedHeadersByHash()[hashOfForkedBlock];
+            ChainedHeader fork = chainTip.FindFork(chainedHeaderOfForkedBlock);
+
+            fork.Should().NotBeNull();
```
noescape00: Please use normal assert

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+            peerOneTipHash.Should().Be(hashOfForkedBlock, "Peer should be on the fork but wasn't");
```
noescape00: Please use Assert.Equal

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -196,6 +208,46 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
         }
 
+        /// <summary>
+        /// Issue 5 @ Create chained header tree component #1321
+        /// 2 peers should supply headers - one of the peers creates a fork.
+        /// Make sure that ChainedBlock is not created more than once for the same header.
+        /// Check that next pointers were created correctly.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_HeadersFromTwoPeersWithFork_ShouldCreateBlocksForNewHeaders()
```
noescape00: This test doesn't test what should be tested. 

Original test description is: 

```
5) 3 peers should supply headers so one of the peers is on a fork. Make sure that ChainedBlock is not created more than once for the same header. Check that next pointers were created correctly
```

So you have 3 peers that present you this: 
```
1) 1a-2a-3a-4a-5a-6a-7a-8a
2) 1a-2a-3a-4a-5a-6a-7a-8a
3) 1a-2a-3a-4a-5b-6b-7b-8b
```

After that CHT should contain 12 headers:

```
1a-2a-3a-4a-5a-6a-7a-8a
           -5b-6b-7b-8b
```

and each of them should have 1 .Next pointer except for 8a and 8b which contain none and 4a which contains 2

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: ok

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: I don't agree with that.  Fluent asserts have a very nice syntax that flows very nicely.

Assert.Equal is error prone and makes you think too much.    Fork should not be null. couldn't be any clearer could it ?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: Assert.Equal in xunit lacks the 2nd parameter for a custom message on failure, so I think Fluent assertions wins here.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: Fluent stuff hides the logic behind the comment so people read the intent, not what the thing does. 
Imho it's just adds another level ob abstraction and before starting to read the test I first will need to read all the methods with nice names that encapsulate it. 

This line from the next comment is a good example:

```
peerOneTipHash.Should().Be(hashOfForkedBlock, "Peer should be on the fork but wasn't");
```

It reads good but what `Should` and `Be` methods do? So I first will go and press F12 on should, read it, go back, go to Be, read it, go back, read the line again. 
Assert.Equal(val1, val2) is much easier to read.

majikandy:
>Ofc we can always cover every test with comments that will explain every detail bat that would be too much and we will have to copypaste a lot of it between the tests.

Comments on tests are wasteful, so definitely don't do more unless it really warrants it. The test code is the opportunity to get expressive, this is a prime candidate for that - it is a real opportunity.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: No, you won't - you just read and accept :)  `Peer one tip hash should be hash of forked block`  -fails if it isn't

There is no requirement to go to the implementation of that - but feel free to check it once. It is a well known nuget package and used by millions of projects.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: So we can create an overload of it.

Also `Be` usage here is wrong. 2n param's comment: 

```
        //   because:
        //     A formatted phrase as is supported by System.String.Format(System.String,System.Object[])
        //     explaining why the assertion is needed. If the phrase does not start with the
        //     word because, it is prepended automatically.
```

but it's treated as a failure message

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: you find `Assert equal expected actual`  is easier to read than `actual should be expected` ?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: I agree that it should have started with a lowercase p as it answers the question "because....."

majikandy: no need for the "but wasn't"

noescape00:
>Comments on tests are wasteful

If they express the intent of the test and the test is fairly difficult they are never wasteful

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: I find `Assert equal expected actual` and  `actual should be expected` equal easy to read.

But I find 

`Assert.Equal(var1,var2)` easier to read than `var1.Should().Be(var2)`

majikandy:
```
But I find

Assert.Equal(var1,var2) easier to read than var1.Should().Be(var2)
```
No you don't - because you got var1 and var 2 the wrong way round in Assert.Equal - which is the exact reason why a fluent syntax is more logical and less error prone.

majikandy: I suggest that you might jus be conditioned to reading Assert a lot, which is fine, but don't rule out fluent assertions as a bad thing just because you have seen Assert.Equal a lot until now.

majikandy:
>If they express the intent of the test and the test is fairly difficult they are never wasteful
Comments on code is bad enough, but comments on tests is a next level of waste, considering the test is what represents what the code must and must not do.  It is a maintenance overhead for the duration of the codebase.  The test itself is the intent.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -230,5 +282,82 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+        /// <summary>
+        /// Issue 8 @ Create chained header tree component #1321
+        /// We have a chain and peer B is on the best chain.
+        /// After that peer C presents an alternative chain with a forkpoint below B's Tip.
+        /// After that peer A prolongs C's chain (but sends all headers from the fork point) with a few valid and one invalid.
+        /// Make sure that C's chain is not removed - only headers after that.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_MultiplePeersWithForks_CorrectTip()
```
noescape00: This test don't test what is supposed to be tested.

Test's description:
```
We have a chain and peer B is on the best chain. After that peer C presents an alternative chain with a forkpoint below B.Tip. After that peer A prolongs C's chain (but he sends all headers from the fork point) with few valid and one invalid. 
	Make sure that C's chain is not removed, only headers after that. 
```


The setup is: 

```
b1=b2=b3=b4=b5
     =c3=c4=c5=a6=a7=a8
```

After we find out that A sent invalid header a8 CHT should disconnect it and it should remove a6,a7,a8.

Instead here you are prolonging the chain normally without sending an invalid header.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: Test Names should represent the high level - that's fine. The scenario itself can be documented inside with meaningful method names and or nicer test domain model / extensions - this would also help with review,

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: I base those preferences on several tries to read current integration tests in BDD style. Way harder for me to see the whole picture that way. 

So it's about personal preference- some find reading black letters on white paper better, some like white letters on black background.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: Please instead assert that the Consumed is equal to the last header that was previously presented
Same for the next assertion

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: For me personally `var1.Should().Be(var2)` is quite unreadable. It is incorrectly combining English language with C#. It reads wrong because it suggests reference to quality instead of value equality. It is absolutely illogical from my point of view.

Aprogiena:
Could we please not discuss personal preferences on PRs? I already asked for that before. If there is no policy, leave the choice to the one who wrote the code regardless of what you think about the style being used.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: I prefer `Should` instead of `FluentAssertions`  but fluent assertions was already used in the solution in the DNS project so suggesting to bring in a different `should` extensions library would have been rejected by the code style police. Hence a happy medium of using something that is already there.

majikandy: ha - this isn't even my code :)

majikandy:
>Could we please not discuss personal preferences on PRs? I already asked for that before. 

I will always push for what I think is right for the company/code.  Can we please add a policy to say that people cannot say what people can and can't discuss on PRs.

At the end of the day, if we get better code - everyone wins.  

>If there is no policy, leave the choice to the one who wrote the code regardless of what you think about the style being used.

I do, however, agree with this as a sensible middle ground.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
monsieurleberre: Just saying that I really like FluentAssertions over Assert, Assert is too symetrical, please don't deprive us of that ! :)

noescape00: >Just saying that I really like FluentAssertions over Assert, Assert is too symetrical, please don't deprive us of that ! :)

No one does. BDD is allowed for the integration tests. For the rest of the code- normal codestyle is required

noescape00:
@fassadlr please add fixes based on review (only those that touch logic. Codestyle fixes are up to you) & resolve merge conflicts.

noescape00:
@thefazzer  please add fixes based on review (only those that touch logic. Codestyle fixes are up to you) & resolve merge conflicts.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
majikandy: BDD is not a code style. To many developers, bdd is normal, not two things that are mutually exclusive. BDD is behaviour driven development - ie driving out the design of the code through specifying behaviours.  You would never say.... we don’t do tdd style tests, we do normal ones so you should never say we don’t do BDD style tests we do normal.   If there is a policy in place that says you aren’t allowed to write expressively worded tests with methods that describe the exact behaviour under test, without using fluent assertions, please show us that policy.

And in reference to the above, Should is unrelated to whether you use TDD, BDD or Test-After.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -217,6 +229,67 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 7);
+            chainedHeaderTree.ConnectNewHeaders(2, listOfExistingHeaders);
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().Count == 8);
+
+            // PEER 3: 1a - 2a - 3a - 4a - 5b - 6b - 7b - 8b
+            BlockHeader forkedBlockHeader = testContext.GetNewBlock(chainTip.Previous.Previous.Previous.Previous).Header;
+            chainedHeaderTree.ConnectNewHeaders(3, new List<BlockHeader>() { forkedBlockHeader });
+            ChainedHeader peerThreeChainedHeader = chainedHeaderTree.GetChainedHeaderByPeerId(3);
+            ChainedHeader peerThreeTip = testContext.ExtendAChain(3, peerThreeChainedHeader);
+            List<BlockHeader> listOfNewAndExistingHeaders = testContext.ChainedHeaderToList(peerThreeTip, 4);
```
Aprogiena: Let's have this list include some of the blocks that are common for both chains

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -217,6 +229,67 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+                // Each should have 1 Next pointer
+                if (height < 3 || (height > 3 && height < 7)) Assert.True(blockGroup.All(x => x.Next.Count == 1));
+
+                // Except for 8a and 8b which contain none
+                if (height == 7) Assert.True(blockGroup.All(x => x.Next.Count == 0));
+
+                // And 4a which contains 2
+                if (height == 3) Assert.True(blockGroup.All(x => x.Next.Count == 2));
+            }
```
Aprogiena: It seems correct, but the whole construction is unnecessarily complex. Why don't we just start at the tip of one chain and go back using previous, then take the tip of the other chain and do the same. We already have the fork point in a variable.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -217,6 +229,67 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            // PEER 1: 1a - 2a - 3a - 4a - 5a - 6a - 7a - 8a
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().Count == 8);
+
+            // PEER 2: 1a - 2a - 3a - 4a - 5a - 6a - 7a - 8a
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 7);
+            chainedHeaderTree.ConnectNewHeaders(2, listOfExistingHeaders);
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().Count == 8);
+
+            // PEER 3: 1a - 2a - 3a - 4a - 5b - 6b - 7b - 8b
+            BlockHeader forkedBlockHeader = testContext.GetNewBlock(chainTip.Previous.Previous.Previous.Previous).Header;
```
Aprogiena: This does not seem right. Why don't you get it is block header from `listOfExistingHeaders`?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+        /// Check that everything is always consumed.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_MultiplePeers_CheckEverythingIsConsumed()
+        {
+            var testContext = new TestContext();
+            ChainedHeaderTree chainedHeaderTree = testContext.CreateChainedHeaderTree();
+            ChainedHeader chainTip = testContext.ExtendAChain(11);
+            chainedHeaderTree.Initialize(chainTip, true);
+            List < BlockHeader > listOfNewHeaders = testContext.ChainedHeaderToList(chainTip, 11);
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // 5 unprocessed headers
+            Assert.True(chainTip.Height - connectedNewHeadersPeer1.Consumed.Height == 5);
+
+            List<BlockHeader> peer2Headers = listOfNewHeaders.GetRange(6, 5);
+            ConnectNewHeadersResult connectedNewHeadersPeer2 = chainedHeaderTree.ConnectNewHeaders(2, peer2Headers);
+
+            // All headers consumed
+            Assert.True(connectedNewHeadersPeer2.Consumed.Height == chainTip.Height);
+        }
```
Aprogiena: So currently we have two scenarios here, first one is that we submit a first part of the chain and we check that it is consumed. Then we add something more to this chain and check that it is consumed.

Now we should add couple more scenarios:

– We should now submit such a list of headers that presents some of those that are already in the tree as well as some that are not in the tree yet.
– Then we should submit list of headers that present some of those that are already in the tree and then the rest of the headers form the new fork the tree.
– Then we should submit list of headers in which nothing is already in the tree but it forms a fork
– Finally, we should submit headers that are all already in the tree.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 10);
+            chainedHeaderTree.ConnectNewHeaders(2, listOfExistingHeaders);
+            ChainedHeader peerTwoTip = chainedHeaderTree.GetChainedHeaderByPeerId(2);
+
+            // Peer B / Peer 2 is on the best chain.
+            peerTwoTip.HashBlock.Should().BeEquivalentTo(chainTip.HashBlock);
+            peerTwoTip.ChainWork.Should().BeEquivalentTo(chainTip.ChainWork);
+
+            // Peer C / Peer 3 presents an alternative chain with a forkpoint below B's Tip.
+            BlockHeader forkedBlock = testContext.GetNewBlock(chainTip.Previous.Previous).Header;
```
Aprogiena: Here again I don't think you need this get new block method. You can simply find your fork point in the list from above.

Aprogiena: The new simply extend the chain at that point with new headers.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            chainedHeaderTree.ConnectNewHeaders(3, new List<BlockHeader>() { forkedBlock });
+            ChainedHeader peerThreeTip = chainedHeaderTree.GetChainedHeaderByPeerId(3);
+
+            ChainedHeader fork = chainTip.FindFork(peerThreeTip);
+            fork.Height.Should().BeLessThan(peerTwoTip.Height);
+            
+            // Peer A / PEER_ONE prolongs C's chain.
+            const int numberOfBlocksToExtend = 7;
+            ChainedHeader peerOneTip = testContext.ExtendAChain(numberOfBlocksToExtend, peerThreeTip);
+            List < BlockHeader > listOfPeerOnesHeaders = testContext.ChainedHeaderToList(peerOneTip, numberOfBlocksToExtend + 2 /* (forked 2 back) */);
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            
+            // Sends all headers from the fork point
+            fork.HashBlock.Should().BeEquivalentTo(listOfPeerOnesHeaders[0].GetHash());
+
+            // With few valid and one invalid.
+            foreach (var header in listOfPeerOnesHeaders.GetRange(3, 6))
+            {
+                testContext.ChainStateMock.Setup(x => x.MarkBlockInvalid(header.GetHash(), null));
+            }
```
Aprogiena: 1) only for the last block in that list
2) you need to mock this.chainedHeaderValidator.ValidateHeader to throw exception for that particular block. ChainStateMock would not work here

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -252,6 +324,83 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            chainedHeaderTree.ConnectNewHeaders(1, listOfPeerOnesHeaders);
+
+            // Peer A / PEER_ONE's fork should now have the longest chain
+            peerOneTip.Height.Should().BeGreaterThan(peerThreeTip.Height);
```
Aprogiena: Here we actually want to check that the whole chain presented by peer A is not part of the tree

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: A compromise here might be to allow comments during test development and then a cleanup at the end of task.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Changed accordingly

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Changed for this approach - now reverse iteration from chain tips

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: GetNewBlock removed entirely

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: All these scenarios now present

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: See above

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: I understand this, I think.  Should be tested.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +294,63 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            ChainedHeader peerThreeTip = testContext.ExtendAChain(4, forkedBlockHeader);
+            List<BlockHeader> listOfNewAndExistingHeaders = testContext.ChainedHeaderToList(peerThreeTip, 6);   // includes common blocks
+            chainedHeaderTree.ConnectNewHeaders(3, listOfNewAndExistingHeaders);
+
+            // CHT should contain 12 headers
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().Count == 12);
+
+            var chainedHeadersByHash = chainedHeaderTree.GetChainedHeadersByHash();
+
+            Assert.True(chainTip.Height==7);
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +294,63 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+                {
+                    Assert.True(chainedHeadersByHashAtHeight.All(x => x.Value.Next.Count == 1));
+                }
+
+                // Except for 8a and 8b which contain none
+                if (height == 7) Assert.True(chainedHeadersByHashAtHeight.All(x => x.Value.Next.Count == 0));
+
+                // And 4a which contains 2
+                if (height == 3) Assert.True(chainedHeadersByHashAtHeight.All(x => x.Value.Next.Count == 2));
+            }
```
Aprogiena: This is still not what we were looking for. The main problem of this approach is that you only check the number of next pointers, but you don't check whether they are correctly set to the next header. This is why I suggested you should start at the tip and use previous pointers to traverse the chain down to the Genesis. You just have two tips of two different chains so that would mean just doing two loops.

I do not insist on the traversal using previous pointers, it just seems to me that it would be much easier code to read because of very simple logic that does not mix the two chains together. However, checking the next pointers properly should be done regardless of which way to traverse you choose.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            chainedHeaderTree.ConnectNewHeaders(3, listOfNewHeadersFromPeerThree);
+            
+            ChainedHeader fork = chainTip.FindFork(peerThreeTip);
+            fork.Height.Should().BeLessThan(peerTwoTip.Height);
+
+            // Peer 1 prolongs Peer 3's chain with one invalid block.
+            // c3 = c4 = c5 = |a6| = a7 = a8
+            const int numberOfBlocksToExtend = 3;
+            ChainedHeader peerOneTip = testContext.ExtendAChain(numberOfBlocksToExtend, peerThreeTip);
+            List <BlockHeader> listOfPeerOnesHeaders = testContext.ChainedHeaderToList(peerOneTip, numberOfBlocksToExtend);
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            
+            int depthOfInvalidHeader = 2;
+            BlockHeader invalidBlockHeader = listOfPeerOnesHeaders[depthOfInvalidHeader];
+            testContext.ChainStateMock.Setup(x => x.MarkBlockInvalid(invalidBlockHeader.GetHash(), null));
```
Aprogiena: This mock is not needed, I think

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            fork.Height.Should().BeLessThan(peerTwoTip.Height);
+
+            // Peer 1 prolongs Peer 3's chain with one invalid block.
+            // c3 = c4 = c5 = |a6| = a7 = a8
+            const int numberOfBlocksToExtend = 3;
+            ChainedHeader peerOneTip = testContext.ExtendAChain(numberOfBlocksToExtend, peerThreeTip);
+            List <BlockHeader> listOfPeerOnesHeaders = testContext.ChainedHeaderToList(peerOneTip, numberOfBlocksToExtend);
+            
+            int depthOfInvalidHeader = 2;
+            BlockHeader invalidBlockHeader = listOfPeerOnesHeaders[depthOfInvalidHeader];
```
Aprogiena: Can we make it in a way that the invalid block is a7? This would be nice because we would test that blocks before as well as after the invalid block were not added to the tree.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            testContext.ChainStateMock.Setup(x => x.MarkBlockInvalid(invalidBlockHeader.GetHash(), null));
+            testContext.ChainedHeaderValidatorMock.Setup(x => 
+                x.ValidateHeader(It.Is<ChainedHeader>(y => y.HashBlock == invalidBlockHeader.GetHash()))).Throws(new ConnectHeaderException());
```
Aprogiena: We should not use this exception here because that means something else. The validator's exceptions are not created yet, so create some special exception that you can recognise and that is not used anywhere else

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            Assert.Throws<ConnectHeaderException>(() => chainedHeaderTree.ConnectNewHeaders(1, listOfPeerOnesHeaders));
+
+            // Whole chain presented by peer A is not part of the tree (can't extend beyond invalid header).
+            int chainHeaderTreeCountChange = oldChainHeaderTreeCount - chainedHeaderTree.GetChainedHeadersByHash().Count;
+
+            // The two headers beyond the invalid header are removed from CHT
+            Assert.True(chainHeaderTreeCountChange == depthOfInvalidHeader + 1);
+
+            // C's chain remains at same height after A presents the invalid header
+            Assert.True(chainedHeaderTree.GetChainedHeaderByPeerId(2).Height == peerThreeTip.Height);
```
Aprogiena: Let's also check that the tip claimed by the third peer is in the chain headers by hash structure

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            int oldChainHeaderTreeCount = chainedHeaderTree.GetChainedHeadersByHash().Count;
+
+            Assert.Throws<ConnectHeaderException>(() => chainedHeaderTree.ConnectNewHeaders(1, listOfPeerOnesHeaders));
+
+            // Whole chain presented by peer A is not part of the tree (can't extend beyond invalid header).
+            int chainHeaderTreeCountChange = oldChainHeaderTreeCount - chainedHeaderTree.GetChainedHeadersByHash().Count;
+
+            // The two headers beyond the invalid header are removed from CHT
+            Assert.True(chainHeaderTreeCountChange == depthOfInvalidHeader + 1);
```
Aprogiena: Shouldn't the old chain had the tree count be just equal to the new count

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            // 1a - 2a - 3a - 4a - 5a - 6a - 7a - 8a
+            TestContext testContext = new TestContextBuilder().Build();
+            ChainedHeaderTree chainedHeaderTree = testContext.ChainedHeaderTree;
+            ChainedHeader chainTip = testContext.ExtendAChain(7);
+            chainedHeaderTree.Initialize(chainTip, true);
+
+            // Chain A is extended by Peer 1:
+            // 1a - 2a - 3a - 4a - 5a - 6a - 7a - 8a - 9a - 10a - 11a - 12a - 13a - 14a - 15a
+            ChainedHeader peer1ChainTip = testContext.ExtendAChain(6, chainTip);
+            List<BlockHeader> listOfPeer1NewHeaders = testContext.ChainedHeaderToList(peer1ChainTip, 6);
```
Aprogiena: The comment above says that seven new guys will be added, but you have six

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            List<BlockHeader> peer1NewHeaders = listOfPeer1NewHeaders.GetRange(0, 4);
+
+            // 4 headers are processed 2 are unprocessed.
+            ConnectNewHeadersResult connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, peer1NewHeaders);
+            Assert.True(peer1ChainTip.Height - connectedNewHeadersResult.Consumed.Height == 2);
```
Aprogiena: Why not assert equal?

Also we do not need to present the just a part of the chain, the only thing we need to test here is that when we presented the prolongation it is consumed in full. Therefore we could just test that the consumed value equals to `peer1ChainTip`

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // Remaining headers are presented.
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, listOfPeer1NewHeaders.GetRange(4, 2));
+            
+            // All headers are now consumed.
+            Assert.True(connectedNewHeadersResult.Consumed.Height == peer1ChainTip.Height);
```
Aprogiena: Don't compare heights, compare hashes

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            List<BlockHeader> listOfSomeExistingHeaders = testContext.ChainedHeaderToList(peer1ChainTip, 14);
+            listOfSomeExistingHeaders = listOfSomeExistingHeaders.GetRange(10, 4);
+
+            // As well as some that are not in it yet:
+            // 15a - 16a
+            ChainedHeader peer2ChainTip = testContext.ExtendAChain(2, peer1ChainTip); 
+            List<BlockHeader> listOfNewHeadersFromPeer2 = testContext.ChainedHeaderToList(peer2ChainTip, 2);
+
+            List<BlockHeader> listOfNewAndExistingHeaders = listOfSomeExistingHeaders;
+            listOfNewAndExistingHeaders.AddRange(listOfNewHeadersFromPeer2);
```
Aprogiena: Why not simply
```
ChainedHeader peer2ChainTip = testContext.ExtendAChain(2, peer1ChainTip); 
List<BlockHeader> listOfNewAndExistingHeaders = testContext.ChainedHeaderToList(peer2ChainTip, 6);

```

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            // Peer 2 presents a list of headers that are already in the tree:
+            // 13a - 14a - 15a - 16a
+            peer2ChainTip = chainedHeaderTree.GetChainedHeaderByPeerId(2);
+            listOfSomeExistingHeaders = testContext.ChainedHeaderToList(peer2ChainTip, 16);
+            listOfNewAndExistingHeaders = listOfSomeExistingHeaders.GetRange(12, 4);
+
+            // And the rest of the headers from a new fork to the tree.
+            // 13a - 14a - 15a - 16a
+            //     - 14b - 15b - 16b
+            var peer2Fork = testContext.ExtendAChain(3, peer2ChainTip.Previous.Previous.Previous);
```
Aprogiena: unvar

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +384,157 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            listOfNewAndExistingHeaders.AddRange(testContext.ChainedHeaderToList(peer2Fork, 3));
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(2, listOfNewAndExistingHeaders);
```
Aprogiena: you can't mix two chains in one like that
but again, why not simply
```
ChainedHeader peer3Tip = testContext.ExtendAChain(5, peer2ChainTip.Previous.Previous.Previous);
List<BlockHeader> listOfPeer3Headers = testContext.ChainedHeaderToList(peer3Tip, 5);
connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(3, listOfPeer3Headers);
```

src/Stratis.Bitcoin.Tests/Consensus/ConsensusTestExtensions.cs
```diff
@@ -21,5 +21,15 @@ internal static class ConsensusTestExtensions
         {
             return chainedHeaderTree.GetMemberValue("chainedHeadersByHash") as Dictionary<uint256, ChainedHeader>;
         }
+
+        public static ChainedHeader GetChainedHeaderByPeerId(this ChainedHeaderTree chainedHeaderTree, int peer)
+        {
+            return chainedHeaderTree.GetChainedHeadersByHash()[chainedHeaderTree.GetPeerTipsByPeerId()[peer]];
+        }
+
+        public static ChainedHeader GetChainedHeadersByHeight(this ChainedHeaderTree chainedHeaderTree, int peer)
```
Aprogiena: not used? also does not do what it says

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Rewritten accordingly.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Correct.  Removed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Replaced with custom exception

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Correct - now tests for this.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Comment corrected

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: I agree this is clearer

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Agreed - used this more simple approach

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done

src/Stratis.Bitcoin.Tests/Consensus/ConsensusTestExtensions.cs
thefazzer: Removed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +301,82 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+                ChainedHeader chainPointer = chainedHeadersByHash[chainedHeaderTree.GetPeerTipsByPeerId()[peer]];
+
+                // Start at the tip and use previous pointers to traverse the chain down to the Genesis.
+                while (chainPointer.Height > 0)
+                {
+                    // Checking the next pointers.
+                    Assert.Contains(chainPointer, chainPointer.Previous.Next);
+
+                    if ((!nextPointersByHeightMap.ContainsKey(chainPointer.Height)))
```
Aprogiena: To many brackets

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +301,82 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+            chainedHeaderTree.ConnectNewHeaders(3, listOfNewAndExistingHeaders);
+
+            // CHT should contain 12 headers
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().Count == 12);
+
+            var chainedHeadersByHash = chainedHeaderTree.GetChainedHeadersByHash();
+
+            foreach (int peer in new[] {2, 3})
+            {
+                var nextPointersByHeightMap = new Dictionary<int, HashSet<ChainedHeader>>();
```
Aprogiena: Instead of Hash set you would be better using list, because the tree was incorrect and contained the same next twice, with Hash set you would not recognise that

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +301,82 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+
+                // Except for 8a and 8b which contain none.
+                Assert.True(nextPointersByHeightMap.Where(x => x.Key == 7 ).All(y => y.Value.Count == 0));
+
+                // And 4a which has 2.
+                Assert.True(nextPointersByHeightMap.Where(x => x.Key == 3).All(y => y.Value.Count == 2));
+            }
+
+            // Two blocks at each height above the fork.
+            Assert.True(chainedHeadersByHash.GroupBy(x => x.Value.Height).Where(x => x.Key > 4).All( y => y.ToList().Count == 2));
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -294,6 +301,82 @@ public void ConnectHeaders_SupplyHeadersThenSupplyMore_Both_Tip_PeerId_Maps_Shou
...
+                    }
+
+                    chainPointer = chainPointer.Previous;
+                }
+
+                // Each should have 1 Next pointer.
+                Assert.True(nextPointersByHeightMap.Where(x => x.Key < 3 || (x.Key > 3 && x.Key < 7)).All(y => y.Value.Count == 1));
+
+                // Except for 8a and 8b which contain none.
+                Assert.True(nextPointersByHeightMap.Where(x => x.Key == 7 ).All(y => y.Value.Count == 0));
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            chainedHeaderTree.Initialize(chainTip, true);
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, 5);
+
+            chainedHeaderTree.ConnectNewHeaders(2, listOfExistingHeaders);
+            ChainedHeader peerTwoTip = chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(2);
+
+            // Peer 2 is on the best chain.
+            // b1 = b2 = b3 = b4 = b5
+            peerTwoTip.HashBlock.Should().BeEquivalentTo(chainTip.HashBlock);
+            peerTwoTip.ChainWork.Should().BeEquivalentTo(chainTip.ChainWork);
```
Aprogiena: If hashes are equal, you no longer need to check anything else it will be equal as well

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            const int numberOfBlocksToExtend = 3;
+            ChainedHeader peerOneTip = testContext.ExtendAChain(numberOfBlocksToExtend, peerThreeTip);
+            List<BlockHeader> listOfPeerOnesHeaders = testContext.ChainedHeaderToList(peerOneTip, numberOfBlocksToExtend);
+            
+            int depthOfInvalidHeader = 1;
+            BlockHeader invalidBlockHeader = listOfPeerOnesHeaders[depthOfInvalidHeader];
+            testContext.ChainedHeaderValidatorMock.Setup(x => 
+                x.ValidateHeader(It.Is<ChainedHeader>(y => y.HashBlock == invalidBlockHeader.GetHash()))).Throws(new InvalidHeaderTestException());
+
+            Assert.Throws<InvalidHeaderTestException>(() => chainedHeaderTree.ConnectNewHeaders(1, listOfPeerOnesHeaders));
```
Aprogiena: Here we wanted to present couple more headers – for example including C4 and C5

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            // The header before, the invalid header and the header beyond is removed from CHT (a6 = a7 = a8)
+            Assert.False(chainedHeaderTree.GetChainedHeadersByHash().ContainsKey(listOfPeerOnesHeaders[0].GetHash()));
+            Assert.False(chainedHeaderTree.GetChainedHeadersByHash().ContainsKey(listOfPeerOnesHeaders[1].GetHash()));
+            Assert.False(chainedHeaderTree.GetChainedHeadersByHash().ContainsKey(listOfPeerOnesHeaders[2].GetHash()));
+
+            Assert.Equal(oldChainHeaderTreeCount, chainedHeaderTree.GetChainedHeadersByHash().Count);
+
+            // tip claimed by the third peer is in the chain headers by hash structure
+            uint256 tipClaimedByThirdPeer = chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(3).HashBlock;
+            Assert.True(chainedHeaderTree.GetChainedHeadersByHash().ContainsKey(tipClaimedByThirdPeer));
```
Aprogiena: The way you implemented GetPeerTipChainedHeaderByPeerId makes it impossible for this assertion to be false. You need to get hash of C6 in different way, for example take hash of peerThreeTip

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // C's chain remains at same height after A presents the invalid header
+            Assert.True(chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(2).Height == peerThreeTip.Height);
```
Aprogiena: why 2 in GetPeerTipChainedHeaderByPeerId(2)?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            // All headers are now consumed.
+            Assert.Equal(connectedNewHeadersResult.Consumed.HashBlock, peer1ChainTip.HashBlock);
+
+            // Peer 2 presents headers that that are already in the tree: 11a - 12a - 13a - 14a
+            // As well as some that are not in it yet: 15a - 16a
+            ChainedHeader peer2ChainTip = testContext.ExtendAChain(2, peer1ChainTip); 
+            List<BlockHeader> listOfNewAndExistingHeaders = testContext.ChainedHeaderToList(peer2ChainTip, 6);
+
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(2, listOfNewAndExistingHeaders);
+            Assert.True(connectedNewHeadersResult.Consumed.Height == peer2ChainTip.Height);
```
Aprogiena: It is better to compare hashes than Heights

Also use assert equal

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // Peer 3 presents a list of headers that are already in the tree
+            // and the rest of the headers from a new fork to the tree.
+            // 13a - 14a - 15a - 16a
+            //     - 14b - 15b - 16b
+            ChainedHeader peer3ChainTip = testContext.ExtendAChain(2, peer2ChainTip.Previous.Previous.Previous);
+            List<BlockHeader> listOfPeer3Headers = testContext.ChainedHeaderToList(peer3ChainTip, 5);
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(3, listOfPeer3Headers);
+
+            Assert.True(connectedNewHeadersResult.Consumed.Height == peer3ChainTip.Height);
```
Aprogiena: As above

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(2, listOfNewAndExistingHeaders);
+            Assert.True(connectedNewHeadersResult.Consumed.Height == peer2ChainTip.Height);
+
+            // Peer 3 presents a list of headers that are already in the tree
+            // and the rest of the headers from a new fork to the tree.
+            // 13a - 14a - 15a - 16a
+            //     - 14b - 15b - 16b
+            ChainedHeader peer3ChainTip = testContext.ExtendAChain(2, peer2ChainTip.Previous.Previous.Previous);
+            List<BlockHeader> listOfPeer3Headers = testContext.ChainedHeaderToList(peer3ChainTip, 5);
```
Aprogiena: Here the comment does not correspond with the code is your extending the chain only by two and also you're taking five blocks while in comment you have extension of three and you're taking four

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            Assert.True(connectedNewHeadersResult.Consumed.Height == peer3ChainTip.Height);
+            
+            List<BlockHeader> peerOneChainedHeaderList = testContext.ChainedHeaderToList(peer1ChainTip, 7);
+            List<BlockHeader> peerTwoChainedHeaderList = testContext.ChainedHeaderToList(peer2ChainTip, 7);
+            
+            // Submit headers that are all already in the tree:
+            // Peer 3 supplies all headers from Peer 1.
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(3, peerOneChainedHeaderList);
+            Assert.True(connectedNewHeadersResult.Consumed.Height == chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(3).Height);
```
Aprogiena: As above, especially if there are more branches, the height is not unique

Also below it is the same.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +410,148 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // Peer 3 supplies all headers from Peer 2.
+            connectedNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(3, peerTwoChainedHeaderList);
+            Assert.True(connectedNewHeadersResult.Consumed.Height == chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(3).Height);
+
+            // Submit a list of headers in which nothing is already in the tree.
+            testContext = new TestContextBuilder().Build();
```
Aprogiena: We don't need to construct new tree for this

Aprogiena: Just add peer four and do the same thing and again down below compared hashes not heights

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Removed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Changed to List<>.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Now compare only #'s.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: C4 & C5 included, purpose commented and asserts added.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Changed to comparison with peerThreeTip (I see your point these structures are always consistent).

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Typo.  Fixed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed and changed all the comments so they follow the h1=h2=h3 format where these refer to heights

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Fixed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Done.

thefazzer:
@Aprogiena @noescape00  All changes made.

Aprogiena:
@noescape00 Looks good to me, run the tests and merge if you're happy.

PR #1514, opened 2018-06-12
CHT safety check

noescape00:
To prevent a possible race condition

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -357,6 +357,12 @@ namespace Stratis.Bitcoin.Consensus
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeader), chainedHeader);
 
+            if (!this.chainedHeadersByHash.ContainsKey(chainedHeader.HashBlock))
+            {
+                this.logger.LogTrace("(-)[NOT_FOUND]");
+                return new List<int>();
+            }
+
             List<int> peersToBan = this.RemoveSubtree(chainedHeader);
 
             this.RemoveUnclaimedBranch(chainedHeader.Previous);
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -357,6 +363,12 @@ public List<int> PartialOrFullValidationFailed(ChainedHeader chainedHeader)
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeader), chainedHeader);
 
+            if (!this.chainedHeadersByHash.ContainsKey(chainedHeader.HashBlock))
+            {
+                this.logger.LogTrace("(-)[NOT_FOUND]");
+                return new List<int>();
```
dangershony: We could return null here.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -300,7 +300,13 @@ public List<ChainedHeader> PartialValidationSucceeded(ChainedHeader chainedHeade
 
             if (!this.chainedHeadersByHash.ContainsKey(chainedHeader.HashBlock))
             {
-                this.logger.LogTrace("(-)[HEADER_NOT_FOUND]");
+                this.logger.LogTrace("(-)[HEADER_NOT_FOUND]:null");
+                return null;
+            }
+
+            if (chainedHeader.Block == null)
```
dangershony: This can happen?
In what cases?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Yes we could but in the other method we return an empty list. And imho it's more logically correct to return the empty list because method claim to return list of peers that should be banned. If it's empty- noone should be banned. 

Also since it's always not null the caller's logic will be cleaner.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Really odd race condition. Very unlikely to happen.

noescape00: We will add comments for those new guys to explain why can it happen

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: OK sure because this is a public method I accept this argument, if it was private I will probably got for null.

PR #1516, opened 2018-06-12
added ApiPort to the CoreNode

bokobza:


src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -39,5 +40,10 @@ public static Money GetFee(this CoreNode node, TransactionBuildContext transacti
         {
             return node.FullNode.WalletTransactionHandler().EstimateFee(transactionBuildContext);
         }
+
+        public static int GetApiPort(this CoreNode coreNode)
+        {
+            return coreNode.FullNode.NodeService<ApiSettings>()?.ApiPort ?? -1;
```
dangershony: This might be called form a qt node?

PR #1517, opened 2018-06-12
Consensus manager part 2

dangershony:


src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,33 +1,50 @@
...
 using Stratis.Bitcoin.Consensus.Rules;
+using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.Utilities;
+using Stratis.Bitcoin.Utilities.Extensions;
 
 namespace Stratis.Bitcoin.Consensus
 {
+    public interface IBlockPuller
+    {
+        void NewTipClaimed();
```
noescape00: This one should have peer Id and ChainedHeader as params.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,33 +1,50 @@
...
+
+        long AverageBlockSize { get; }
+
+        void RequestNewData(BlockDownloadRequest downloadRequest);
```
noescape00: Please rename to RequestBlocksDownload

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +106,242 @@ public class ConsensusManager
...
+    /// <summary>
+    /// A request that holds information of blocks to download.
+    /// </summary>
+    public class BlockDownloadRequest
+    {
+        /// <summary>The list of block headers to download.</summary>
+        public List<uint256> BlocksToDownload { get; set; }
+
+        /// <summary>The delegate to send the downloaded blocks to.</summary>
+        public ProcessDownloadedBlockDelegate DownloadedBlockDelegate { get; set; }
```
noescape00: Maybe rename type and variable name to OnBlockDownloadedCallback?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +106,242 @@ public class ConsensusManager
...
+                this.chainedHeaderTree.PeerDisconnected(peerId);
+                this.blockPuller.PeerDisconnected(peerId);
+                this.ProcessDownloadQueueLocked();
+            }
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>
+        /// A callback that is triggered when a block was downloaded.
```
noescape00: This callback is called only when CM itself requested this block. It's not called when a block requested by other component is downloaed

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +106,242 @@ public class ConsensusManager
...
+        /// <param name="blockHashes">The block hashes to download, the block hashes must be in a consecutive order.</param>
+        /// <param name="processesBlock">The callback that will be called for each downloaded block.</param>
+        /// <returns></returns>
+        public async Task GetOrDownloadBlocksAsync(List<uint256> blockHashes, ProcessDownloadedBlockDelegate processesBlock)
+        {
+            this.logger.LogTrace("({0}.{1}:{2})", nameof(blockHashes), nameof(blockHashes.Count), blockHashes.Count);
+
+            // TODO: verify the blockHashes are consecutive
+
+            List<uint256> downloadJob = new List<uint256>();
```
noescape00: Some logic is missing in this method:
```
hashesToDownload.Add(hash) //Check consequtiveness here. If for some reason newly added pair.ChainedHeader.Prev != DownloadBlocks.Last - Call DownloadBlocks on the list we have and recreate the list
							// So we always have only consecutive headers in the list.
```

So the point here is if we are asked for hashes

a1=a2=a3=a4=a5=a9=a10=a12=a13=b2=b3=b4=b5

we should generate 3 download jobs:
1) a1=a2=a3=a4=a5
2) a9=a10=a12=a13
3) b2=b3=b4=b5

because block puller requires download job headers to be always consecutive.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -112,6 +112,11 @@ public Task<uint256> GetTrxBlockIdAsync(uint256 trxid)
             return this.blockRepository.GetTrxBlockIdAsync(trxid);
         }
 
+        public Task<Block> GetBlockAsync(uint256 blockHash)
```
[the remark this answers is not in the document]
dangershony: hmm yeah why not will move it

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -28,6 +30,8 @@ public interface IChainedHeaderValidator
         /// <param name="block">The block that is going to be validated.</param>
         /// <param name="chainedHeader">The chained header of the block that will be validated.</param>
         void VerifyBlockIntegrity(Block block, ChainedHeader chainedHeader);
+
+        void StartPartialValidation(BlockPair blockPair, Action<BlockPair, bool> onPartialValidationCompletedCallback);
```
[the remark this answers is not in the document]
dangershony: At the moment I am going for a single validator interface, so would actualyl rename it ti IChainValidator (or IBlockValidator) if we find we need the break the logic we can just do it.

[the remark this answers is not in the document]
noescape00: Ok, sounds good

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -250,6 +260,21 @@ public void Initialize(ChainedHeader consensusTip, bool blockStoreEnabled)
...
 
+        // <inheritdoc />
+        public BlockPair GetBlockPair(uint256 blockHash)
+        {
+            this.logger.LogTrace("({0}:{1})", nameof(blockHash), blockHash);
+
+            BlockPair blockPair = null;
+            if (this.chainedHeadersByHash.TryGetValue(blockHash, out ChainedHeader chainedHeader))
+            {
+                blockPair = new BlockPair(chainedHeader.Block, chainedHeader);
```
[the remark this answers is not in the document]
dangershony: ah good spot, I think I will make the setters public.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: ok sure

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: yes I missed that one thanks
Fixed

src/Stratis.Bitcoin/Consensus/BlockValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <param name="block">The block that is going to be validated.</param>
+        /// <param name="chainedHeader">The chained header of the block that will be validated.</param>
+        void VerifyBlockIntegrity(Block block, ChainedHeader chainedHeader);
+
+        /// <summary>
+        /// Partial validation of a block, this will not changes any state in the consensus store when validating a block.
+        /// </summary>
+        /// <param name="chainedHeaderBlock">The block to validate.</param>
+        /// <param name="onPartialValidationCompletedCallback">A callback that is called when validation is complete.</param>
+        void StartPartialValidation(ChainedHeaderBlock chainedHeaderBlock, Action<ChainedHeaderBlock, bool> onPartialValidationCompletedCallback);
```
Aprogiena: This is how you asked about why not action but delegate. Here we don't know what those arguments in the action are, where in delegate we would now.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -148,13 +129,19 @@ public interface IChainedHeaderTree
         /// <param name="block">The block.</param>
         /// <returns>Newly created and connected chained header for the specified block.</returns>
         ChainedHeader CreateChainedHeaderWithBlock(Block block);
+
+        /// <summary>
+        /// Get the block and its chained header if it exists.
+        /// </summary>
+        /// <returns>The block and its chained header.</returns>
+        ChainedHeaderBlock GetChainedHeaderBlock(uint256 blockHash);
```
Aprogiena: We should specify when the return value is null and also when the block part of the result is null.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -250,6 +237,21 @@ public void Initialize(ChainedHeader consensusTip, bool blockStoreEnabled)
             this.logger.LogTrace("(-)");
         }
 
+        // <inheritdoc />
+        public ChainedHeaderBlock GetChainedHeaderBlock(uint256 blockHash)
+        {
+            this.logger.LogTrace("({0}:{1})", nameof(blockHash), blockHash);
```
Aprogiena: '

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -250,6 +237,21 @@ public void Initialize(ChainedHeader consensusTip, bool blockStoreEnabled)
...
+
+            ChainedHeaderBlock chainedHeaderBlock = null;
+            if (this.chainedHeadersByHash.TryGetValue(blockHash, out ChainedHeader chainedHeader))
+            {
+                chainedHeaderBlock = new ChainedHeaderBlock(chainedHeader.Block, chainedHeader);
+            }
+
+            this.logger.LogTrace("(-):'{0}'", chainedHeaderBlock);
```
Aprogiena: We don't have overload for to string here, we need to go with the chained header part.

src/Stratis.Bitcoin/Consensus/BlockValidator.cs
dangershony: ah yes I will change

dangershony: I wonder if the bool value needs to be some sort to of error message (with an error code)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: We have now

src/Stratis.Bitcoin/Primitives/ChainedHeaderBlock.cs
```diff
@@ -15,17 +15,22 @@ public sealed class BlockPair
         public ChainedHeader ChainedHeader { get; private set; }
 
         /// <summary>
-        /// Creates instance of <see cref="BlockPair" />.
+        /// Creates instance of <see cref="ChainedHeaderBlock" />.
         /// </summary>
-        /// <param name="block">The block.</param>
+        /// <param name="block">The block, the block can be <c>null</c>.</param>
```
Aprogiena: remove second the block, just can be null is enough or start new sentence

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: oh, my mistake, okay then

src/NBitcoin/ChainedHeader.cs
```diff
@@ -536,6 +536,28 @@ public ChainedHeader FindFork(ChainedHeader block)
             return highChain;
         }
 
+        /// <summary>
+        /// Convert the current and given chained headers  on the same chain to a list of consecutive items. 
+        /// </summary>
+        /// <param name="stopHeader">The header at the bottom.</param>
+        /// <returns>List of consecutive items.</returns>
+        /// <exception cref="InvalidOperationException">Then the items are not on the same chain.</exception>
+        public IEnumerable<ChainedHeader> ToConsecutiveList(ChainedHeader stopHeader)
```
Aprogiena: I don't think this method is very helpful. Seems to just adding overhead when you want to go down the chain. Its name also does not match what it does and the summary also does not match what it does.

src/Stratis.Bitcoin/Consensus/BlockValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <param name="chainedHeaderBlock">The block to validate.</param>
+        /// <param name="onValidationCompletedCallback">A callback that is called when validation is complete.</param>
+        void StartPartialValidation(ChainedHeaderBlock chainedHeaderBlock, OnValidationCompletedCallback onValidationCompletedCallback);
+    }
+
+    /// <summary>
+    /// A callback that is invoked when <see cref="IBlockValidator.StartPartialValidation"/> completes to validate a block.
+    /// </summary>
+    /// <param name="chainedHeaderBlock">The block and its chained header.</param>
+    /// <param name="success">An indicator weather validation succeeded.</param>
```
Aprogiena: Whether

src/Stratis.Bitcoin/Consensus/BlockValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Partial validation of a block, this will not changes any state in the consensus store when validating a block.
+        /// </summary>
+        /// <param name="chainedHeaderBlock">The block to validate.</param>
+        /// <param name="onValidationCompletedCallback">A callback that is called when validation is complete.</param>
+        void StartPartialValidation(ChainedHeaderBlock chainedHeaderBlock, OnValidationCompletedCallback onValidationCompletedCallback);
+    }
+
+    /// <summary>
+    /// A callback that is invoked when <see cref="IBlockValidator.StartPartialValidation"/> completes to validate a block.
```
Aprogiena: Completes validation of a block

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -148,13 +129,20 @@ public interface IChainedHeaderTree
         /// <param name="block">The block.</param>
         /// <returns>Newly created and connected chained header for the specified block.</returns>
         ChainedHeader CreateChainedHeaderWithBlock(Block block);
+
+        /// <summary>
+        /// Get the block and its chained header if it exists.
+        /// If the header is not in the tree returns <c>null</c>, the <see cref="ChainedHeaderBlock.Block"/> may also be null.
```
Aprogiena: This rather belongs to returns not to the summary

Also use see for the second null

Finally, I would at the information that the block is null when the block data has not yet been downloaded or if the block data has been persisted to the database and removed from the memory.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1097,5 +1100,17 @@ public override string ToString()
         {
             return $"{nameof(this.DownloadFrom)}='{this.DownloadFrom}',{nameof(this.DownloadTo)}='{this.DownloadTo}',{nameof(this.Consumed)}='{this.Consumed}'";
         }
+
+        /// <summary>
+        /// Convert the <see cref="DownloadFrom"/> and <see cref="DownloadTo"/> to a list of consecutive hashes where the fist item in the list is the hash of the download from.
```
Aprogiena: First

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
             ICheckpoints checkpoints, 
             ConsensusSettings consensusSettings, 
-            ConcurrentChain concurrentChain, 
-            ILookaheadBlockPuller lookaheadBlockPuller,
+            IBlockPuller blockPuller,
             IConsensusRules consensusRules,
-            IFinalizedBlockHeight finalizedBlockHeight)
+            IFinalizedBlockHeight finalizedBlockHeight,
+            IConnectionManager connectionManager,
+            IBlockStore blockStore = null)
```
Aprogiena: This could be a nice topic to discuss in the big class comment – how does consensus manager operates if the block store is presented and how does it operate when it is not presented.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
-            this.blocksRequested = new Dictionary<uint256, List<Action>>();
-            this.toDownloadQueue = new AsyncQueue<(ChainedHeader headerFrom, ChainedHeader headerTo)>();
+            this.blocksRequested = new Dictionary<uint256, List<OnBlockDownloadedCallback>>();
+            this.toDownloadQueue = new Queue<BlockDownloadRequest>();
         }
 
-        public async Task InitializeAsync()
+        /// <summary>
+        /// Set the tip of <see cref="ConsensusManager"/> to be the header that is stored to disk,
+        /// if the given <see cref="consensusTip"/> is not in store then rewind store till a tip is found.
```
Aprogiena: paramref instead of see

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
+        /// If <see cref="IChainState.BlockStoreTip"/> is not null (block store is enabled) ensure the store and consensus manager are on the same tip,
+        /// if consensus manager is ahead then reorg it to be the same height as store.
```
Aprogiena: Basically you repeated what you wrote in the first sentence. However, you are missing to describe what happens if the blocks store is not enabled.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
                 utxoHash = await this.consensusRules.RewindAsync().ConfigureAwait(false);
             }
 
-            this.concurrentChain.SetTip(this.Tip);
+            lock (this.treeLock)
```
Aprogiena: We discussed this, lock is not needed here.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
+        public async Task InitializeAsync(ChainedHeader consensusTip)
         {
-            this.logger.LogTrace("()");
+            this.logger.LogTrace("({0}:'{1}')", nameof(consensusTip), consensusTip);
+
+            // TODO: consensus store
+            // We should consider creating a consensus store class that will internally contain
+            //  coinview and it will abstract the methods `RewindAsync()` `GetBlockHashAsync()` 
 
             uint256 utxoHash = await this.consensusRules.GetBlockHashAsync().ConfigureAwait(false);
```
Aprogiena: I agree that loading block hash from consensus rules looks weird.

Aprogiena: And could we please find a better name for this variable. How about coin view tip hash?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
-            this.toDownloadQueue = new AsyncQueue<(ChainedHeader headerFrom, ChainedHeader headerTo)>();
+            this.blocksRequested = new Dictionary<uint256, List<OnBlockDownloadedCallback>>();
+            this.toDownloadQueue = new Queue<BlockDownloadRequest>();
         }
 
-        public async Task InitializeAsync()
+        /// <summary>
+        /// Set the tip of <see cref="ConsensusManager"/> to be the header that is stored to disk,
+        /// if the given <see cref="consensusTip"/> is not in store then rewind store till a tip is found.
+        /// If <see cref="IChainState.BlockStoreTip"/> is not null (block store is enabled) ensure the store and consensus manager are on the same tip,
```
Aprogiena: Would it not be better if instead of chain state, we directly looked into the value that we receive in constructor?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
+            // We should consider creating a consensus store class that will internally contain
+            //  coinview and it will abstract the methods `RewindAsync()` `GetBlockHashAsync()` 
 
             uint256 utxoHash = await this.consensusRules.GetBlockHashAsync().ConfigureAwait(false);
             bool blockStoreDisabled = this.chainState.BlockStoreTip == null;
 
             while (true)
             {
-                this.Tip = this.concurrentChain.GetBlock(utxoHash);
+                this.Tip = consensusTip.FindAncestorOrSelf(utxoHash);
```
Aprogiena: This is not correct. If coin view database is ahead of the chain repository, this will go through the whole blockchain one by one before it returns null. We need to find another way how to deal with this. Easiest solution seems to be to store height together with the hash in the coin view database.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,70 +1,113 @@
...
+            this.logger.LogTrace("({0}:'{1}')", nameof(consensusTip), consensusTip);
+
+            // TODO: consensus store
+            // We should consider creating a consensus store class that will internally contain
+            //  coinview and it will abstract the methods `RewindAsync()` `GetBlockHashAsync()` 
 
             uint256 utxoHash = await this.consensusRules.GetBlockHashAsync().ConfigureAwait(false);
             bool blockStoreDisabled = this.chainState.BlockStoreTip == null;
 
             while (true)
```
Aprogiena: Some kind of limit on this loop should be defined – possibly one block store rewound to Genesis

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        public ChainedHeader HeadersPresented(int peerId, List<BlockHeader> headers)
+        {
+            this.logger.LogTrace("({0}:{1},{2}.{3}:{4})", nameof(peerId), peerId, nameof(headers), nameof(headers.Count), headers.Count);
+
+            ConnectNewHeadersResult newHeaders = null;
+
+            lock (this.treeLock)
+            {
+                newHeaders = this.chainedHeaderTree.ConnectNewHeaders(peerId, headers);
+                this.blockPuller.NewTipClaimed(peerId, newHeaders.Consumed); // TODO: should this be inside the lock?
```
Aprogiena: Yes, it is just the name of the lock that is slightly misleading. We need to prevent race conditions with peer disconnections.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>
+        /// A list of headers are presented from a given peer,
+        /// we'll attempt to connect the headers to the tree and if new headers are found they will be queued for download.
+        /// </summary>
+        /// <param name="peerId">The peer that is providing the headers.</param>
+        /// <param name="headers">The list of new headers.</param>
+        /// <returns>The last chained header that is connected to the tree.</returns>
```
Aprogiena: When null?

Aprogiena: Actually we cannot have null, the code below is misleading.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+
+            ConnectNewHeadersResult newHeaders = null;
+
+            lock (this.treeLock)
+            {
+                newHeaders = this.chainedHeaderTree.ConnectNewHeaders(peerId, headers);
+                this.blockPuller.NewTipClaimed(peerId, newHeaders.Consumed); // TODO: should this be inside the lock?
+            }
+
+            if (newHeaders?.DownloadTo != null)
```
Aprogiena: It cannot be null, so no need for the ? Here

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            {
+                this.DownloadBlocks(newHeaders.ToHashList(), this.ProcessDownloadedBlock);
+            }
+
+            this.logger.LogTrace("(-):'{0}'", newHeaders?.Consumed);
```
Aprogiena: Cannot be null – it can only throw in case of problem.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            return newHeaders?.Consumed;
+        }
+
+        /// <summary>
+        /// A peer was disconnected, clean all its claimed headers that are above our consensus tip.
+        /// Processes any remaining blocks to download.
+        /// </summary>
+        /// <param name="peerId">The peer that is providing the headers.</param>
```
Aprogiena: This is copy and paste

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            {
+                this.DownloadBlocks(newHeaders.ToHashList(), this.ProcessDownloadedBlock);
+            }
+
+            this.logger.LogTrace("(-):'{0}'", newHeaders?.Consumed);
+            return newHeaders?.Consumed;
+        }
+
+        /// <summary>
+        /// A peer was disconnected, clean all its claimed headers that are above our consensus tip.
```
Aprogiena: On this level, we rather just inform the depending components and only one of those components is actually cleaning the headers, but that is its internal thing I guess

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        }
+
+        private void OnPartialValidationSucceeded(ChainedHeaderBlock chainedHeaderBlock)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeaderBlock), chainedHeaderBlock);
+
+            bool reorged = false;
+            List<ChainedHeader> nextHeadersToValidate = null;
+
+            lock (this.reorgLock)
```
Aprogiena: This must be async lock.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                    List<ChainedHeader> headersToConnect =  peer.Behavior<ChainHeadersBehavior>().ConsensusAdvanced(this.Tip);
+
+                    if (headersToConnect != null)
+                    {
+                        List<BlockHeader> headers = headersToConnect.Select(ch => ch.Header).ToList();
+
+                        lock (this.treeLock)
+                        {
+                            ConnectNewHeadersResult connectNewHeaders = this.chainedHeaderTree.ConnectNewHeaders(peer.Connection.Id, headers);
+                            blocksToDownload.Add(connectNewHeaders);
```
Aprogiena: I think here we want to add to the list only if download to is not null

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                {
+                    this.DownloadBlocks(newHeaders.ToHashList(), this.ProcessDownloadedBlock);
+                }
+            }
+
+            // Start validating all next blocks that come after the current block,
+            // all headers in this list have the blocks present in the header.
+            foreach (ChainedHeader chainedHeader in nextHeadersToValidate)
+            {
+                this.blockValidator.StartPartialValidation( new ChainedHeaderBlock(chainedHeader.Block, chainedHeader), this.OnPartialValidationCompletedCallback);
```
Aprogiena: Extra space

But maybe just create variable for the new object

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            }
 
-            this.lookaheadBlockPuller.SetLocation(this.Tip);
+            this.logger.LogTrace("(-)");
+        }
 
-            this.chainedHeaderTree.Initialize(this.Tip, !blockStoreDisabled);
+        private bool FullyValidateAndReorgLocked()
+        {
+            return false;
```
Aprogiena: Damn, that will not work

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                        switch (cex)
+                        {
+                            case BlockDownloadedForMissingChainedHeaderException bcex:
+                                return;
+
+                            // TODO: catch validation exceptions.
+                        }
+                    }
+                }
+            }
```
Aprogiena: Definitely we want to log if block was not delivered here

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                {
+                    try
+                    {
+                        chainedHeader = this.chainedHeaderTree.FindHeaderAndVerifyBlockIntegrity(block);
+                    }
+                    catch (ConsensusException cex)
+                    {
+                        switch (cex)
+                        {
+                            case BlockDownloadedForMissingChainedHeaderException bcex:
```
Aprogiena: Why not just catch it instead of switching? Also please are the logs everywhere

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            }
+
+            List<OnBlockDownloadedCallback> listOfCallbacks = null;
+
+            lock (this.blockRequestedLock)
+            {
+                if (this.blocksRequested.TryGetValue(blockHash, out listOfCallbacks))
+                {
+                    this.blocksRequested.Remove(blockHash);
+                }
```
Aprogiena: Here it would also be nice to log that no callback was found

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            if (listOfCallbacks != null)
+            {
+                foreach (var blockDownloadedCallback in listOfCallbacks)
+                {
+                    ChainedHeaderBlock chainedHeaderBlock = null;
+
+                    if (block != null)
+                    {
+                        chainedHeaderBlock = new ChainedHeaderBlock(block, chainedHeader);
+                    }
```
Aprogiena: I would go for little more compact way and write it on one line, but it is fine

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+
+                    blockDownloadedCallback(chainedHeaderBlock);
```
Aprogiena: Also hash should be passed

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// If the tree has too many unconsumed blocks or the download will wait, the amount of blocks to downloaded depends on the avg value in <see cref="IBlockPuller.AverageBlockSize"/>.
+        /// </summary>
+        /// <remarks>
+        /// Requests that have too many blocks will be split in batches.
+        /// </remarks>
+        private void ProcessDownloadQueueLocked()
+        {
+            this.logger.LogTrace("()");
+
+            const long MaxUnconsumedBlocksDataBytes = 1048576 * 10; // 10 mb
```
Aprogiena: Make it class constant with comment, But also 10 MB seems to be very small number, for bitcoin it could mean just six blocks being downloaded in parallel

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+
+            while (this.toDownloadQueue.Count > 0)
+            {
+                BlockDownloadRequest request = this.toDownloadQueue.Peek();
+
+                long freeMb = MaxUnconsumedBlocksDataBytes - this.chainedHeaderTree.UnconsumedBlocksDataBytes;
```
Aprogiena: Rename this one, it is not in megabytes

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                this.logger.LogTrace("There is {0} MB available do download is.", freeMb);
+
+                if (freeMb <= 32)
```
Aprogiena: This was supposed to be in megabytes, but create a constant in bytes and document the constant

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+            this.logger.LogTrace("()");
+
+            const long MaxUnconsumedBlocksDataBytes = 1048576 * 10; // 10 mb
+
+            while (this.toDownloadQueue.Count > 0)
+            {
+                BlockDownloadRequest request = this.toDownloadQueue.Peek();
+
+                long freeMb = MaxUnconsumedBlocksDataBytes - this.chainedHeaderTree.UnconsumedBlocksDataBytes;
+                this.logger.LogTrace("There is {0} MB available do download is.", freeMb);
```
Aprogiena: It is fine to have the log in megabytes, but make sure you are comparing apples with apples.

Also write the sentence again, does not make sense.

noescape00:
In CHT we have a method:

```
public List<uint256> ToHashList()
{
    List<uint256> blockHashes = this.DownloadTo.ToConsecutiveList(this.DownloadFrom).Select(header => header.HashBlock).ToList();
    blockHashes.Reverse();
    return blockHashes;
}
```

which is quite inefficient. Approach that uses LINQ on average 8 times slower than normal one (5 test runs, amount of headers is 2000)

Here is the code I used to test the performance:

```
[Fact]
public void TestSpeed()
{
    ConcurrentChain chain = this.CreateChain(2000);

    ChainedHeader from = chain.GetBlock(0);
    ChainedHeader to = chain.Tip;

    Stopwatch watchARRAY = Stopwatch.StartNew();
    IEnumerable<uint256> hashesARRAY = this.GetHashesARRAY(from, to);
    watchARRAY.Stop();

    Stopwatch watchLINQ = Stopwatch.StartNew();
    IEnumerable<uint256> hashesLINQ = this.GetHashesLINQ(from, to);
    watchLINQ.Stop();


    var linqTime = watchLINQ.Elapsed.TotalMilliseconds;
    var arrayTime = watchARRAY.Elapsed.TotalMilliseconds;
}

public IEnumerable<uint256> GetHashesLINQ(ChainedHeader from, ChainedHeader to)
{
    List<uint256> blockHashes = to.ToConsecutiveList(from).Select(header => header.HashBlock).ToList();
    blockHashes.Reverse();

    return blockHashes;
}

public IEnumerable<uint256> GetHashesARRAY(ChainedHeader from, ChainedHeader to)
{
    uint256[] hashes = new uint256[to.Height - from.Height];

    int index = hashes.Length - 1;

    ChainedHeader currentHeader = to;

    while (index >= 0)
    {
        hashes[index] = currentHeader.HashBlock;

        index--;
        currentHeader = currentHeader.Previous;
    }

    return hashes;
}

private ConcurrentChain CreateChain(int blocksCount)
{
    var chain = new ConcurrentChain(Network.StratisMain);

    for (int i = 0; i < blocksCount; i++)
    {
        var header = new BlockHeader()
        {
            Nonce = RandomUtils.GetUInt32(),
            HashPrevBlock = chain.Tip.HashBlock,
            Bits = Target.Difficulty1
        };

        var chainedHeader = new ChainedHeader(header, header.GetHash(), chain.Tip);

        chain.SetTip(chainedHeader);
    }

    return chain;
}
```

PS
Also the method is not inclusive so
`Convert the <see cref="DownloadFrom"/> and <see cref="DownloadTo"/> to a list of consecutive hashes where the fist item in the list is the hash of the download from.`
comment is incorrect.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+
+                if (freeMb <= 32)
+                    return;
+
+                long blocksToAsk = freeMb / this.blockPuller.AverageBlockSize;
+                this.logger.LogTrace("The slot of available blocks to ask is {0}.", blocksToAsk);
```
Aprogiena: With {0} average block size, we have {1} download slots available.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+                        OnBlockDownloadedCallback = request.OnBlockDownloadedCallback,
+                        BlocksToDownload = new List<uint256>()
+                    };
+
+                    for (int i = 0; i < blocksToAsk; i++)
+                    {
+                        uint256 hash = request.BlocksToDownload[0];
+                        request.BlocksToDownload.Remove(hash);
+                        newRequest.BlocksToDownload.Add(hash);
+                    }
```
Aprogiena: Please don't use quadratic algorithms

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        }
+
+        /// <summary>
+        /// Processes items in <see cref="toDownloadQueue"/> and ask the block puller for blocks to download.
+        /// If the tree has too many unconsumed blocks or the download will wait, the amount of blocks to downloaded depends on the avg value in <see cref="IBlockPuller.AverageBlockSize"/>.
+        /// </summary>
+        /// <remarks>
+        /// Requests that have too many blocks will be split in batches.
+        /// </remarks>
+        private void ProcessDownloadQueueLocked()
```
Aprogiena: It seems we have a problem in design with this method, so maybe just don't fix it yet until we discussed further.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+
+                    blockDownloadedCallback(chainedHeaderBlock);
+                }
+            }
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>
+        /// A method that to requests blocks, any block that is found is sent to the callback <see cref="onBlockDownloadedCallback"/>.
```
Aprogiena: Rephrase, doesn't make sense

Aprogiena: See => paramref

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// If the block is not part of the tree (on a known the chain) the a <c>null</c> value will be return in the callback.
```
Aprogiena: The a

return in => passed to

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// If the block is available in memory or on disk it will be immediately return the block on the callback.
```
Aprogiena: return the block on => passed to

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// The block hashes must be in a consecutive order.
```
Aprogiena: No longer required

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// </summary>
+        /// <param name="blockHashes">The block hashes to download, the block hashes must be in a consecutive order.</param>
```
Aprogiena: Consecutive nest is no longer required here

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -74,13 +117,388 @@ public class ConsensusManager
...
+        /// <param name="onBlockDownloadedCallback">The callback that will be called for each downloaded block.</param>
+        /// <returns></returns>
+        public async Task GetOrDownloadBlocksAsync(List<uint256> blockHashes, OnBlockDownloadedCallback onBlockDownloadedCallback)
+        {
+            this.logger.LogTrace("({0}.{1}:{2})", nameof(blockHashes), nameof(blockHashes.Count), blockHashes.Count);
+
+            List<uint256> downloadRequests = new List<uint256>();
+            ChainedHeaderBlock previousBlock = null;
+
+            for (int i = blockHashes.Count - 1; i >= 0; i--)
```
Aprogiena: erm why reverse?

src/NBitcoin/ChainedHeader.cs
dangershony: Either you do it here or in `ConnectNewHeadersResult` I dont mind, but it maybe useful to have it on the chain header.
If comment is incorrect then we can just fix the comment

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: I can add the null to the return as well but normally the summery is the first thing you read (and often the only thing users read)

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: I will put this in a remarks then

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1106,7 +1106,9 @@ public override string ToString()
         /// </summary>
         public uint256[] ToHashList()
         {
-            var hashes = new uint256[this.DownloadTo.Height - this.DownloadFrom.Height];
+            int itemsCount = this.DownloadTo.Height == this.DownloadFrom.Height ? 1 : this.DownloadTo.Height - this.DownloadFrom.Height + 1;
```
Aprogiena: Why the ternary operator here? If the guys are equal, the result will be one

src/NBitcoin/ChainedHeader.cs
Aprogiena: If you want to keep it, then yes, fix the name and fix the comments. But I'm questioning whether we need such a method at all.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: yeah but no harm either, its better practice to actually lock every entry to the dictionary.
I could remove and put a disable warring tag, but prefer not to.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Well I dont want to mention coinview in anyway its a store implementation detail,
I am thinking of creating a ConsensusStore class that internally will have coinview this will look much better.
But we can do it after CM is finished.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Yes it would, makes much more sense good call.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: Please use the sections properly.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: Why it would be better practice? And what warning are you talking about? Who gives you a warning?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: If you don't want to call it coin view, try something else, but better name should be suggested. How about persisted consensus tip hash?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: No because rewind rill move `utxoHash` one step back `consensusTip` wont change so it will eventually be above  `utxoHash`

dangershony: There is a bug here however `this.Tip` can never be null, I will need to either create a new method on chained header or check against consensusTip.

dangershony: I think its ok new, wait for next push.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Maybe use the max reorg flag?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: ok so basically puller locks are also in CM thats fine, we just need to make sure `NewTipClaimed` is a small and fast method.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: This is from the tree
```
            if (connectNewHeadersResult == null)
                connectNewHeadersResult = new ConnectNewHeadersResult() { Consumed = latestNewHeader };
```

dangershony: Ah you mean `newHeaders` cant be null yeah good call

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: yeah it will come when I implement the reorg method.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Well spotted

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: yeah I will

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: You must have faith master apro

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: How about bitcoin?

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: Not really, block puller has its own locks but here we need to synchronise these two components together

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -132,9 +132,12 @@ public interface IChainedHeaderTree
...
         /// <summary>
         /// Get the block and its chained header if it exists.
-        /// If the header is not in the tree returns <c>null</c>, the <see cref="ChainedHeaderBlock.Block"/> may also be null.
+        /// If the header is not in the tree <see cref="ChainedHeaderBlock"/> will be <c>null</c>, the <see cref="ChainedHeaderBlock.Block"/> may also be null.
         /// </summary>
-        /// <returns>The block and its chained header.</returns>
+        /// <remarks>
+        /// The block can be null when the block data has not yet been downloaded or if the block data has been persisted to the database and removed from the memory.
+        /// </remarks>
+        /// <returns>The block and its chained header (both the block and chained header can be null).</returns>
```
Aprogiena: Actually I don't think that chained header can be null. Either it is the whole result that is null, or the chained header has to value.

In any case, use `<c>` for null here and in the remarks and above in the summary

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1102,19 +1105,19 @@ public override string ToString()
         }
 
         /// <summary>
-        /// Convert the <see cref="DownloadFrom"/> and <see cref="DownloadTo"/> to a list of consecutive hashes where the fist item in the list is the hash of the download from.
+        /// Convert the <see cref="DownloadFrom"/> and <see cref="DownloadTo"/> to a list of consecutive hashes where the first item in the list is the hash of <see cref="DownloadFrom"/>.
```
Aprogiena: No longer it gives faster list of consecutive hashes. Newly it gives us a list of consecutive chained headers

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: This one still holds

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1102,19 +1105,19 @@ public override string ToString()
...
+        public ChainedHeader[] ToHashArray()
         {
             int itemsCount = this.DownloadTo.Height == this.DownloadFrom.Height ? 1 : this.DownloadTo.Height - this.DownloadFrom.Height + 1;
 
-            var hashes = new uint256[itemsCount];
+            var hashes = new ChainedHeader[itemsCount];
 
             ChainedHeader currentHeader = this.DownloadTo;
 
             for (int i = hashes.Length - 1; i >= 0; --i)
```
Aprogiena: I would use `i--` for consistency with other loops in the code.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -44,7 +46,7 @@ public class ConsensusManager
         /// <summary>The current tip of the chain that has been validated.</summary>
         public ChainedHeader Tip { get; private set; }
 
-        private readonly Dictionary<uint256, List<OnBlockDownloadedCallback>> blocksRequested;
+        private readonly Dictionary<uint256, List<OnBlockDownloadedCallback>> callbacksByBlocksRequesteHash;
```
Aprogiena: Requested

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -83,43 +85,44 @@ public class ConsensusManager
...
-        /// if the given <see cref="consensusTip"/> is not in store then rewind store till a tip is found.
-        /// If <see cref="IChainState.BlockStoreTip"/> is not null (block store is enabled) ensure the store and consensus manager are on the same tip,
-        /// if consensus manager is ahead then reorg it to be the same height as store.
+        /// Set the tip of <see cref="ConsensusManager"/>, if the given <paramref name="chainTip"/> is not similar to <see cref="Tip"/>
+        /// then rewind consensus until a common header is found.
         /// </summary>
-        /// <param name="consensusTip">The consensus tip that is attempted to be set.</param>
-        public async Task InitializeAsync(ChainedHeader consensusTip)
+        /// <remarks>
+        /// If <see cref="blockStore"/> is not null (block store is enabled) then all block headers in <see cref="IChainedHeaderTree"/> will be marked as blocks are available.
```
Aprogiena: `<c>`

Aprogiena: => `marked as their block data is available.`

and maybe not `IChainedHeaderTree` but our instance `chainedHeaderTree`

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -83,43 +85,44 @@ public class ConsensusManager
...
+        /// If store is disabled the <see cref="ConsensusManager"/> wont be able to serve blocks from disk, instead all block requests that are not in memory will be sent to the <see cref="IBlockPuller"/>.
```
Aprogiena: Won't

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -83,43 +85,44 @@ public class ConsensusManager
...
-            bool blockStoreDisabled = this.chainState.BlockStoreTip == null;
+            uint256 consensusTipHash = await this.consensusRules.GetBlockHashAsync().ConfigureAwait(false);
 
             while (true)
             {
-                this.Tip = consensusTip.FindAncestorOrSelf(utxoHash);
+                this.Tip = chainTip.FindAncestorOrSelf(consensusTipHash);
 
-                if ((this.Tip != null) && (blockStoreDisabled || (this.chainState.BlockStoreTip.Height >= this.Tip.Height)))
+                if (this.Tip.HashBlock == consensusTipHash)
```
Aprogiena: I don't think this change is correct.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: One more? Is in the log line

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -259,22 +262,27 @@ private void OnPartialValidationSucceeded(ChainedHeaderBlock chainedHeaderBlock)
...
+                                blocksToDownload.Add(connectNewHeaders);
+                            }
                         }
                     }
                 }
 
                 foreach (ConnectNewHeadersResult newHeaders in blocksToDownload)
                 {
-                    this.DownloadBlocks(newHeaders.ToHashList(), this.ProcessDownloadedBlock);
+                    this.DownloadBlocks(newHeaders.ToHashArray().ToList(), this.ProcessDownloadedBlock);
```
Aprogiena: How about changing the signature of download blocks to accept an array so that we don't need to convert to list?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: hmm yeah I will fix

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Ah yea its the ChainedHeaderBlock that can be null.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Is it not actually wrong, this will decrement i before it enters the loop

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: skype?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: No, it does exactly the same as postfix version, the only problem with it is inconsistency with other code.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -293,32 +293,63 @@ private bool FullyValidateAndReorgLocked()
...
+                        if (blockAlreadyAsked)
+                        {
+                            previousHeader = null;
+                            continue;
+                        }
                     }
 
-                    listOfCallbacks.Add(onBlockDownloadedCallback);
+                    previousHeader = chainedHeader;
+                    request.BlocksToDownload.Add(chainedHeader);
```
Aprogiena: How about the last batch?

How about two consecutive blocks that have been asked already?

src/Stratis.Bitcoin.Tests/Consensus/ConsensusManagerTest.cs
```diff
@@ -0,0 +1,102 @@
...
+                this.ConsensusManager = new ConsensusManager(this.Network, new ExtendedLoggerFactory(), this.ChainStateMock.Object, 
+                    this.ChainedHeaderValidatorMock.Object, this.CheckpointsMock.Object, this.ConsensusSettings, this.BlockPullerMock.Object,
+                    this.ConsensusRulesMock.Object, this.FinalizedBlockMock.Object, this.ConnectionManagerMock.Object );
+
+                this.ChainedHeaderTree = this.ConsensusManager.GetMemberValue("chainedHeaderTree") as ChainedHeaderTree;
+
+                return this.ConsensusManager;
+            }
+
+            public ChainedHeader ExtendAChain(int count, ChainedHeader chainedHeader = null)
```
Aprogiena: wtf? This code already exist elsewhere but it is better, why would you create a copy of that is just worse? Why not use the code from CHT tests?

dangershony: I think its not pushed yet, but yeah probably we need some shared test context, I will let the testers do that.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: Ah yeah I will fix

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,86 +1,596 @@
...
+            }
+
+            this.DownloadBlocks(downloadRequests.ToArray(), this.ProcessDownloadedBlock);
 
             this.logger.LogTrace("(-)");
         }
+
+        /// <summary>
+        /// Processes items in <see cref="toDownloadQueue"/> and ask the block puller for blocks to download.
+        /// If the tree has too many unconsumed blocks or the download will wait, the amount of blocks to downloaded depends on the avg value in <see cref="IBlockPuller.AverageBlockSize"/>.
```
Aprogiena: Rephrase

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,86 +1,596 @@
...
+                    var newRequest = new BlockDownloadRequest()
+                    {
+                        BlocksToDownload = new List<ChainedHeader>()
+                    };
+
+                    for (int i = 0; i < blocksToAsk; i++)
+                    {
+                        ChainedHeader chainedHeader = request.BlocksToDownload[0];
+                        request.BlocksToDownload.Remove(chainedHeader);
+                        newRequest.BlocksToDownload.Add(chainedHeader);
```
Aprogiena: Fix this

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -1,86 +1,596 @@
...
+                this.blockPuller.RequestNewData(request);
+
+                foreach (ChainedHeader chainedHeader in request.BlocksToDownload)
+                    this.expectedBlockSizes.Add(chainedHeader.HashBlock, avgSize);
+
+                this.expectedBlockDataBytes += request.BlocksToDownload.Count * avgSize;
+            }
+
+            this.logger.LogTrace("(-)");
+        }
```
Aprogiena: More logs into this method

PR #1519, opened 2018-06-12
Cleanup 7

noescape00:
Fixed brackets. 

This is the last automatic one.


Aprogiena:
Why do you have built failed CI?

noescape00:
>Why do you have built failed CI?

Because CI fails from time to time. 
The tests do pass.

PR #1521, opened 2018-06-13
[RF][SmartContracts] Abstract the AddToBlock method

fassadlr:
In SmartContracts we add transactions to the block differently than how we do pow/pos blocks. We first need to check whether or not the transaction to be added contains SC information and if it does, we execute it to validate and calculate fees etc.

Currently we have to make the `AddToBlock` method in `BlockDefinition` virtual so that we can override its behavior. This needs to be changed to an abstract method so that each block definition implements its own logic to add a transaction to a block.

The logic in the current `AddToBlock` has been split into 3 methods:

1> Add tx to block.
2> Update block stats.
3> Update block fee.


fassadlr:
All unit and integration tests pass. Busy syncing...

noescape00:
Shouldn't it also update merkle tree root?

dangershony:
Look good to me I just need to run tests.

fassadlr:
I confirm that all chains has synced and staking still works as per normal.

dangershony:
Ok all tests passing as well

PR #1526, opened 2018-06-14
Chain header tree test to cover issue 14

fenix2222:
Issue 14 @ Chain exists with checkpoints enabled. There are 2 checkpoints. Peer presents a chain that covers first checkpoint with a prolongation that does not match the 2nd checkpoint. Exception should be thrown.

See https://github.com/stratisproject/StratisBitcoinFullNode/issues/1321#issuecomment-389155904.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -354,12 +354,17 @@ namespace Stratis.Bitcoin.Tests.Consensus
             extendedChainTip = extendedChainTip.Previous;
             extendedChainTip = ctx.ExtendAChain(newChainExtension, extendedChainTip); 
             List<BlockHeader> listOfNewChainHeaders = ctx.ChainedHeaderToList(extendedChainTip, extendedChainTip.Height);
-            
-            // New chain is presented by peer 1.
+
+            // First 5 blocks are presented by peer 1.
+            // DownloadTo should be set to a checkpoint 1. 
+            ConnectNewHeadersResult result = cht.ConnectNewHeaders(1, listOfNewChainHeaders.Take(5).ToList());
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -309,5 +309,60 @@ public void PresentDifferentChains_AlternativeChainWithMoreChainWorkShouldAlways
...
+            extendedChainTip = extendedChainTip.Previous; // walk back to block 6
+            extendedChainTip = extendedChainTip.Previous;
+            extendedChainTip = ctx.ExtendAChain(newChainExtension, extendedChainTip); 
+            List<BlockHeader> listOfNewChainHeaders = ctx.ChainedHeaderToList(extendedChainTip, extendedChainTip.Height);
+            
+            // New chain is presented by peer 1.
+            // InvalidHeaderException should be thrown.
+            Action connectAction = () =>
+            {
+                cht.ConnectNewHeaders(1, listOfNewChainHeaders);
```
Aprogiena: We wanted the peer to 1st present such a chain that covers first checkpoint only. So in this case it would be for example up to block number five. There should be no problem with such presentation. And then the peer will try to prolong using blocks 6 to 10 which will violate the checkpoint.

So the test is correct except that we just want to split the presentation of headers into two separated steps so that we can distinguish between the first attempt which is correct and should be successful from the other attempt.

This should be only minor modification to the code. Otherwise it looks good.

fenix2222: Once I present first part, do I need to set consensus tip after that before presenting second part or not?

Aprogiena: No, you don't need to do anything with consensus tip.

Aprogiena:
It looks good to me, let's wait for review from @noescape00

PR #1527, opened 2018-06-15
Remove SetNull() from BlockHeader and make the constructor internal

fassadlr:
This PR was to remove the `SetNull()` method in `BlockHeader` as we override this in smart contracts. A better implementation would be to set all these values in the constructor.

As part of this PR I had to fix a lot of tests as we now correctly create the block header via the consensus factory everywhere.

All unit and integration tests pass. Busy syncing....

### DIFF
--- a/src/NBitcoin/Block.cs
+++ b/src/NBitcoin/Block.cs
@@ -30,14 +30,13 @@ namespace NBitcoin
         public Block()
         {
             this.header = new BlockHeader();
-            this.SetNull();
+            this.transactions.Clear();
+            this.BlockSize = null;
         }
 
         [Obsolete("Should use Block.Load outside of ConsensusFactories")]
-        internal Block(BlockHeader blockHeader)
+        internal Block(BlockHeader blockHeader) : this()
         {
-            this.header = new BlockHeader();
-            this.SetNull();
             this.header = blockHeader;
         }
 
@@ -73,18 +72,13 @@ namespace NBitcoin
             }
         }
 
-        private void SetNull()
-        {
-            this.header.SetNull();
-            this.transactions.Clear();
-            this.BlockSize = null;
-        }
-
         public BlockHeader Header => this.header;
 
+        /// <summary>
+        /// A block's hash is it's header's hash.
+        /// </summary>
         public uint256 GetHash()
         {
-            // A Block's hash is it's header's hash.
             return this.header.GetHash();
         }
 
--- a/src/NBitcoin/ConsensusFactory.cs
+++ b/src/NBitcoin/ConsensusFactory.cs
@@ -99,25 +99,22 @@ namespace NBitcoin
         /// <typeparam name="T">The generic type to resolve.</typeparam>
         /// <param name="result">If the type is known it will be initialized.</param>
         /// <returns><c>true</c> if it is known.</returns>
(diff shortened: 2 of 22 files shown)


Aprogiena:
Raising the review require level to high because the PR touches the very lowest level structures and is nontrivial.

fassadlr:
I confirm that TestNet has synced and that staking still works as expected.

fassadlr:
All the chains synced 👍

src/NBitcoin/ConsensusFactory.cs
```diff
@@ -99,25 +99,22 @@ private bool IsAssignable<T>(TypeInfo type, ConcurrentDictionary<Type, bool> cac
         /// <typeparam name="T">The generic type to resolve.</typeparam>
         /// <param name="result">If the type is known it will be initialized.</param>
         /// <returns><c>true</c> if it is known.</returns>
-        public virtual bool TryCreateNew<T>(out T result) where T : IBitcoinSerializable
+        public virtual T TryCreateNew<T>() where T : IBitcoinSerializable
```
dangershony: Why this change I wonder?

src/NBitcoin/Block.cs
```diff
@@ -30,14 +30,13 @@ public MerkleNode GetMerkleRoot()
         public Block()
         {
             this.header = new BlockHeader();
-            this.SetNull();
+            this.transactions.Clear();
```
dangershony: This should probably also have the `Obsolete` attribute

src/NBitcoin/ConsensusFactory.cs
fassadlr: Hi Dan! OK so when I made `BlockHeader`'s constructor internal, the `Clone` method didn't like it because it was looking for a public parameter-less constructor. This was due to the `new()` generic constraint (old code):

```public static T Clone<T>(this T serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null) where T : IBitcoinSerializable, new()```

Because of that change we now need to create the default instance via the `Activator`.

File: src/NBitcoin/Block.cs (hunk 2 above)
Comment (fassadlr) [reply src/NBitcoin/Block.cs]: OK will do!

Comment (fassadlr) [reply src/NBitcoin/Block.cs]: Ah I see it has already ;)

File: src/NBitcoin/IBitcoinSerializable.cs [hunk 3]
```diff
@@ -72,14 +73,16 @@ public static void FromBytes(this IBitcoinSerializable serializable, byte[] byte
             serializable.ReadWrite(bms);
         }
 
-        public static T Clone<T>(this T serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null) where T : IBitcoinSerializable, new()
+        public static T Clone<T>(this T serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null) where T : IBitcoinSerializable
```
dangershony: I suggest creating a new method `FactoryClone<T>()` without a `new()`
And throw if `instance == null`

fassadlr:
After the changes, all the unit and integration tests pass 👍

PR #1528, opened 2018-06-15
Add possibility to start custom nodes with custom args

monsieurleberre:
This is needed for sidechains integration testing, where we need to pass known ports to localhost nodes to connect them together


bokobza:
Good stuff, thank you.

PR #1535, opened 2018-06-18
CHT Test: Two chains one covering checkpoint and assume valid and another only checkpoint

fenix2222:
Issue 16 @ Checkpoints are enabled. After the last checkpoint, there is an assumed valid. The chain that covers them all is presented - marked for download. After that chain that covers the last checkpoint but doesn't cover assume valid and is longer is presented - marked for download.
See #1321 (comment)

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -472,11 +472,65 @@ namespace Stratis.Bitcoin.Tests.Consensus
             result.DownloadTo.HashBlock.Should().Be(listOfChainABlockHeaders.Last().GetHash());
 
             // Chain B is presented by peer 2.
-            // DownloadFrom should be set to header 3. 
+            // DownloadFrom should be set to header 5. 
             // DownloadTo should be set to header 10. 
             result = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
-            result.DownloadFrom.HashBlock.Should().Be(listOfChainBBlockHeaders.Skip(2).First().GetHash());
+            result.DownloadFrom.HashBlock.Should().Be(listOfChainBBlockHeaders[checkpointHeight].GetHash());
             result.DownloadTo.HashBlock.Should().Be(listOfChainBBlockHeaders.Last().GetHash());
+        }
+
+        /// <summary>
+        /// Issue 16 @ Checkpoints are enabled. After the last checkpoint, there is an assumed valid. The chain that covers
+        /// the last checkpoint but doesn't cover assume valid is presented - marked for download.
+        /// </summary>
+        [Fact]
+        public void ChainHasOneCheckPointAndAssumeValid_ChainsWithCheckpointButMissedAssumeValidIsPresented_BothChainsAreMarkedForDownload()
+        {
+            // Chain header tree setup.
+            const int initialChainSize = 2;
+            const int extensionChainSize = 2;
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            ChainedHeader initialChainTip = ctx.ExtendAChain(initialChainSize); // i.e. h1=h2
+            ctx.ConsensusSettings.UseCheckpoints = true;
+            cht.Initialize(initialChainTip, true);
+
+            // Extend chain with 2 more headers.
+            initialChainTip = ctx.ExtendAChain(extensionChainSize, initialChainTip); // i.e. h1=h2=h3=h4
+            List<BlockHeader> listOfCurrentChainHeaders = ctx.ChainedHeaderToList(initialChainTip, initialChainSize + extensionChainSize);
+
+            // Setup a known checkpoint at header 4.
+            // Example: h1=h2=h3=(h4).
+            const int checkpointHeight = 4;
+            var checkpoint = new CheckpointInfo(listOfCurrentChainHeaders.Last().GetHash());
+            ctx.CheckpointsMock
+               .Setup(c => c.GetCheckpoint(checkpointHeight))
+               .Returns(checkpoint);
+            ctx.CheckpointsMock
+               .Setup(c => c.GetCheckpoint(It.IsNotIn(checkpointHeight)))
+               .Returns((CheckpointInfo)null);
+            ctx.CheckpointsMock
+                .Setup(c => c.GetLastCheckpointHeight())
+                .Returns(checkpointHeight);
+
+            // Extend chain and add "Assume valid" at block 6.
+            // Example: h1=h2=h3=(h4)=h5=[h6].
+            const int chainExtension = 2;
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -412,5 +412,67 @@ public void ChainHasAssumeValidHeaderAndMarkedForDownloadWhenPresented_SecondCha
...
+            var ctx = new TestContext();
+            ChainedHeaderTree cht = ctx.CreateChainedHeaderTree();
+            ChainedHeader initialChainTip = ctx.ExtendAChain(initialChainSize); // ie. h1=h2=h3=h4
+            cht.Initialize(initialChainTip, true);
+            ctx.ConsensusSettings.UseCheckpoints = true;
+            List<BlockHeader> listOfCurrentChainHeaders = ctx.ChainedHeaderToList(initialChainTip, initialChainSize);
+
+            // Setup a known checkpoint at header 4.
+            // Example: h1=h2=h3=(h4).
+            const int checkpointHeight = 4;
```
Aprogiena: In this test, we want the checkpoint to be presented by a peer, so it should not be part of the initial chain. So if our initial chain has four blocks the checkpoint should be for example at block six and assumed valid block could be at block eight.

Then we expect the first peer to present headers that include the checkpointed block six as well as the assumed valid block eight and goes further up to let's say block 10.

The second peer should present headers that include let's say the whole chain up to block seven including the checkpoint and then block eight would be different so that it does not match the assumed valid block, And it could go farther to let's say block number nine.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -412,5 +412,71 @@ public void ChainHasAssumeValidHeaderAndMarkedForDownloadWhenPresented_SecondCha
...
+            ctx.CheckpointsMock
+                .Setup(c => c.GetLastCheckpointHeight())
+                .Returns(checkpointHeight);
+
+            // Extend chain and add "Assume valid" at block 6.
+            // Example: h1=h2=h3=(h4)=h5=[h6].
+            const int chainExtension = 2;
+            ChainedHeader extendedChainTip = ctx.ExtendAChain(chainExtension, initialChainTip);
+            listOfCurrentChainHeaders = ctx.ChainedHeaderToList(extendedChainTip, extendedChainTip.Height);
+            ctx.ConsensusSettings.BlockAssumedValid = listOfCurrentChainHeaders.Last().GetHash();
```
Aprogiena: Wouldn't this be equal to extendedChainTip hash?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -412,5 +467,107 @@ public void ChainHasAssumeValidHeaderAndMarkedForDownloadWhenPresented_SecondCha
...
+            ctx.SetupCheckpoints(checkpoint);
+
+            // Extend chain and add "Assume valid" at block 6.
+            // Example: h1=h2=h3=(h4)=h5=[h6].
+            const int chainExtension = 2;
+            ChainedHeader extendedChainTip = ctx.ExtendAChain(chainExtension, initialChainTip);
+            ctx.ConsensusSettings.BlockAssumedValid = extendedChainTip.HashBlock;
+
+            // Setup two alternative chains A and B. Chain A covers the last checkpoint (4) and "assume valid" (6).
+            // Chain B only covers the last checkpoint (4), but is longer that chain A.
```
Aprogiena: I would remove "but is longer" part from the comment because it is not relevant to what we want to test. It could be shorter or the same length and it would still be usable to test we want.

fenix2222: That's what initial specs said. I can remove it if it is irrelevant. "Checkpoints are enabled. After last checkpoint there is assume valid. Chain that covers them all is presented- marked for download. After that chain that covers last checkpoint but doesnt cover assume valid and `is longer` is presented- marked for download." https://github.com/stratisproject/StratisBitcoinFullNode/issues/1321#issuecomment-389155904

Aprogiena: I think it is little bit of misunderstanding. I think what we wanted to say is that the alternative chain is longer than the height of assumed valid block, which means that is a block at that height and it is different and also there is some other block after that. The longer part was not meant to be in comparison to the first chain, but to the assumed valid height.

fenix2222: Right, ok. Updated.

Aprogiena:
This good, we just need to investigate the reported failures.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -416,5 +477,107 @@ public void ChainHasAssumeValidHeaderAndMarkedForDownloadWhenPresented_SecondCha
...
+            // Chain B is presented by peer 2.
+            // DownloadFrom should be set to header 5. 
+            // DownloadTo should be set to header 10. 
+            result = cht.ConnectNewHeaders(2, listOfChainBBlockHeaders);
+            result.DownloadFrom.HashBlock.Should().Be(listOfChainBBlockHeaders[checkpointHeight].GetHash());
+            result.DownloadTo.HashBlock.Should().Be(listOfChainBBlockHeaders.Last().GetHash());
+        }
+
+        /// <summary>
+        /// Issue 16 @ Checkpoints are enabled. After the last checkpoint, there is an assumed valid. The chain that covers
```
noescape00: This is test case 17, not 16

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -416,5 +477,107 @@ public void ChainHasAssumeValidHeaderAndMarkedForDownloadWhenPresented_SecondCha
...
+            var checkpoint = new CheckpointFixture(checkpointHeight, listOfCurrentChainHeaders.Last());
+            ctx.SetupCheckpoints(checkpoint);
+
+            // Extend chain and add "Assume valid" at block 6.
+            // Example: h1=h2=h3=(h4)=h5=[h6].
+            const int chainExtension = 2;
+            ChainedHeader extendedChainTip = ctx.ExtendAChain(chainExtension, initialChainTip);
+            ctx.ConsensusSettings.BlockAssumedValid = extendedChainTip.HashBlock;
+
+            // Setup new chain, which covers the last checkpoint (4), but misses "assumed vaid".
```
noescape00: `vaid` - typo

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: Ok, cool then

PR #1539, opened 2018-06-19
CHT bugfix

noescape00:


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -667,13 +667,14 @@ public ConnectNewHeadersResult ConnectNewHeaders(int networkPeerId, List<BlockHe
...
                 // to deal with one special header, which simplifies the implementation.
-                while (currentChainedHeader != earliestNewHeader)
+                while (currentChainedHeader != earliestNewHeader.Previous)
                 {
                     if (currentChainedHeader.HashBlock == this.consensusSettings.BlockAssumedValid)
                     {
                         this.logger.LogDebug("Chained header '{0}' represents an assumed valid block.", currentChainedHeader);
 
-                        connectNewHeadersResult = this.HandleAssumedValidHeader(currentChainedHeader, latestNewHeader, isBelowLastCheckpoint);
+                        bool assumeValidBelowLastCheckpoint = this.consensusSettings.UseCheckpoints && (currentChainedHeader.Height <= this.checkpoints.GetLastCheckpointHeight());
```
dangershony: Perhaps add the call GetLastCheckpointHeight() before the loop starts to avoid calling the method on every iteration?

fassadlr:
@noescape00 I'm gonna jump in here by asking why wasn't this caught by a unit test? If this a new scenario then we need to ensure it is covered by a unit test 👍

noescape00:
@fassadlr it was caught by @fenix2222 's unit test. 

Unit tests are being written after component is completed for this refactoring because every new component has like 40-50 different test scenarios.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: ah yes correct lets push

PR #1542, opened 2018-06-20
Logging: block store stats

noescape00:
Added block store stats to periodic logs and removed old redundant counter

untitled


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -86,6 +90,7 @@ public class BlockStoreQueue : IBlockStore, IDisposable
             this.storeSettings = storeSettings;
             this.chain = chain;
             this.blockRepository = blockRepository;
+            this.batch = new List<ChainedHeaderBlock>();
```
dangershony: Where do you add/remove to the list?

noescape00: On `DequeueBlocksContinuouslyAsync`

I just moved batch list creation from the method body to a class variable to get it's size in the logs.

PR #1543, opened 2018-06-21
Consensus manager part 3

dangershony:


src/NBitcoin/ChainedHeader.cs
```diff
@@ -578,6 +578,29 @@ public ChainedHeader GetAncestor(int ancestorHeight)
...
+
+            ChainedHeader currentHeader = this;
+
+            for (int i = hashes.Length - 1; i >= 0; i--)
+            {
+                hashes[i] = currentHeader;
+                currentHeader = currentHeader.Previous;
+            }
+
+            if(currentHeader != chainedHeader)
```
Aprogiena: Spacing

src/NBitcoin/ChainedHeader.cs
```diff
@@ -578,6 +578,29 @@ public ChainedHeader GetAncestor(int ancestorHeight)
             return walk;
         }
 
+        /// <summary>
+        /// Convert the current <see cref="ChainedHeader" /> and <paramref name="chainedHeader"/> to an array
+        /// of consecutive headers, both items are included in the array.
+        /// </summary>
+        /// <returns>Array of consecutive headers.</returns>
```
Aprogiena: The whole description is confusing as well as the way this method should be used. I would expect the method to accept the count parameter which would say how many headers you want to get. From the comment it is not obvious what the result is supposed to be. So I would highly recommend to change the signature of the method so that it accept the count parameter.

src/Stratis.Bitcoin.Features.Consensus/Rules/ConsensusRules.cs
```diff
@@ -46,9 +46,12 @@ public override Task<uint256> GetBlockHashAsync()
         }
 
         /// <inheritdoc />
-        public override Task<uint256> RewindAsync()
+        public override async Task<ConsensusStoreTransitionState> RewindAsync()
         {
-            return this.UtxoSet.Rewind();
+            return new ConsensusStoreTransitionState()
+            {
+                BlockHash = await this.UtxoSet.Rewind()
```
Aprogiena: Configure await

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+
+            ChainedHeader[] newChain = newTip.ToChainedHeaderArray(fork);
+
+            ChainedHeader lastValidatedHeader = null;
+            ValidationContext validationContext = null;
+            List<int> badPeers = null;
+
+            foreach (ChainedHeader nextChainedHeader in newChain)
+            {
+                if (nextChainedHeader.BlockValidationState != ValidationState.PartiallyValidated)
```
Aprogiena: It can be fully validated. I'm not sure if we need such a check. But if you want the sanity check, just log it and throw

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+                {
+                    throw new Exception(); // TODO: what do we do.
+                }
+
+                ChainedHeaderBlock nextChainedHeaderBlock = this.chainedHeaderTree.GetChainedHeaderBlock(nextChainedHeader.HashBlock);
+
+                validationContext = new ValidationContext() { Block = nextChainedHeaderBlock.Block };
```
Aprogiena: The block here can be null, because we can jump back more than we keep in the memory. So we want to load the block from the block store if it was not in the memory and if we have the block store. For light wallet, we will try to make sure that we will never jump back more than the fork, which will ensure the availability of the blocks in the memory on the new forked branch. We cannot make such an ultimate guarantee for bitcoin, but in practice it will never happen. So we would just need the code that recognises that it happened and the up with the consensus tip before the fork.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+
+                // Call the validation engine.
+                await this.consensusRules.AcceptBlockAsync(validationContext, this.Tip);
```
Aprogiena: Configure await

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+
+                if (validationContext.Error != null)
```
Aprogiena: Invert condition

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+                if (badPeers == null)
+                    badPeers = new List<int>();
+
+                badPeers.AddRange(reorgViolated);
+
+                if (this.network.Consensus.MaxReorgLength != 0)
+                {
+                    int newFinalizedHeight = lastValidatedHeader.Height - (int)this.network.Consensus.MaxReorgLength;
+
+                    await this.finalizedBlockHeight.SaveFinalizedBlockHeightAsync(newFinalizedHeight);
```
Aprogiena: Config await

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
Aprogiena: Actually, I have not very happy with that so we're need to change the design a little bit

src/NBitcoin/ChainedHeader.cs
dangershony: Ah you mean instead of passing a chainedHeader pass a count?

Aprogiena: Yes

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -265,7 +268,12 @@ private void ProcessDownloadedBlock(ChainedHeaderBlock chainedHeaderBlock)
                 }
 
                 if (reorgRequired)
-                    reorged = await this.FullyValidateAndReorgLockedAsync(chainedHeaderBlock).ConfigureAwait(false);
+                {
+
+                   ValidationContext validationContext  = await this.FullyValidateAndReorgLockedAsync(chainedHeaderBlock).ConfigureAwait(false);
```
Aprogiena: Spacing and empty line

src/NBitcoin/ChainedHeader.cs
dangershony: Then I wont be able to do this
```
            if(currentHeader != chainedHeader)
                throw new NotSupportedException("Header must be on the same chain.");
```

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -319,8 +322,121 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
...
+                // New header is behind current tip this is a bug
+                throw new Exception(); // TODO: what do we do.
+            }
+
+            // If the new block is not on the current chain as our current consensus tip
+            // then rewind consensus tip to the common fork (or earlier because rewind might jump a few blocks back)
+            if (fork != oldTip)
+            {
+                ChainedHeader current = oldTip;
+                consensusStoreTransitionQueue = new Queue<ConsensusStoreTransitionState>();
```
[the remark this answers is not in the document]
dangershony: To keep the order, no preference but queue seems ok.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: This is no a valid comment anymore I guess, can you delete it if true

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
dangershony: ops...

src/NBitcoin/ChainedHeader.cs
Aprogiena: Yes and it won't be missed.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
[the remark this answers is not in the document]
Aprogiena: The list also preserves the order

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -379,8 +382,9 @@ private async Task<(ValidationContext ValidationContext, bool ConsensusTipChaneg
                 // In that case we try to load the blocks from store, if store is not present
                 // we mark the blocks for download and do not proceed.
 
-                if (oldTip.Height - current.Height > 1)
+                if (current != fork)
```
dangershony: yes correct this looks better.
But now the line bellow ` fork = current;` can go inside the `if`

noescape00:
Some tests are failing for the rules. @dangershony can you please merge the latest master in and check if it's still a problem? It shouldn't be.

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -161,30 +164,37 @@ public class ConsensusManager : IDisposable
         /// A list of headers are presented from a given peer,
         /// we'll attempt to connect the headers to the tree and if new headers are found they will be queued for download.
         /// </summary>
-        /// <param name="networkPeerId">The peer that is providing the headers.</param>
+        /// <param name="peer">The peer that providing the headers.</param>
         /// <param name="headers">The list of new headers.</param>
-        /// <returns>The last chained header that is connected to the tree.</returns>
+        /// <param name="triggerDownload">Specifies if the download should be scheduled for interesting blocks.</param>
```
dangershony: Interesting blocks will be determined by the tree no?

noescape00: Yes

dangershony: This is the complexity we had to get because of BP?

Aprogiena: Yes

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -227,123 +260,128 @@ private void ProcessDownloadedBlock(ChainedHeaderBlock chainedHeaderBlock)
...
-                        var peer = this.connectionManager.FindNodeById(peerId);
-                        ConnectionManagerSettings connectionSettings = this.connectionManager.ConnectionSettings;
-                        this.peerBanning.BanPeer(peer.PeerEndPoint, connectionSettings.BanTimeSeconds);
-                    }
-
                     lock (this.peerLock)
                     {
+                        foreach (int peerId in connectBlocksResult.PeersToBan)
+                        {
+                            if (this.peersByPeerId.TryGetValue(peerId, out INetworkPeer peer))
```
dangershony: I think we don't necessarily need `INetworkPeer` in the dictionary this will be looked up in `peerBanning.BanAndDisconnectPeer`

src/Stratis.Bitcoin/Consensus/ConsensusManger.cs
```diff
@@ -390,158 +417,249 @@ private void NotifyChainedHeaderBehaviorsOnReorg()
             ChainedHeader currentTip = fork;
 
             // If the new block is not on the current chain as our current consensus tip
-            // then rewind consensus tip to the common fork (or earlier because rewind might jump a few blocks back)
+            // then rewind consensus tip to the common fork (or earlier because rewind might jump a few blocks back).
             bool isExtension = fork == oldTip;
-
+            
             if (!isExtension)
+                currentTip = await this.RewindToForkPointOrBelowAsync(fork, oldTip).ConfigureAwait(false);
```
dangershony: Nice I like

PR #1544, opened 2018-06-21
Cleanup network usage in NBitcoin #1

fassadlr:
We need to start removing Network.Main references from the NBitcoin project as we eventually want to have all the networks int heir own project.


src/NBitcoin/ConsensusFactory.cs
```diff
@@ -39,6 +39,13 @@ public class ConsensusFactory
         /// </summary>
         private readonly TypeInfo transactionType = typeof(Transaction).GetTypeInfo();
 
+        private readonly Network network;
```
dangershony: In this case, please remove this field.
`public Consensus Consensus { get; set; }`

src/NBitcoin/BlockStake.cs
```diff
@@ -34,9 +34,9 @@ public BlockStake()
         {
         }
 
-        public BlockStake(byte[] bytes)
+        public BlockStake(Network network, byte[] bytes)
```
dangershony: We should probably not do that, ther eshould be a Load() method instead.

src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -10,23 +10,23 @@ public interface IBitcoinSerializable
 
     public static class BitcoinSerializableExtensions
     {
-        public static void ReadWrite(this IBitcoinSerializable serializable, Stream stream, bool serializing, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null)
+        public static void ReadWrite(this IBitcoinSerializable serializable, Stream stream, bool serializing, ConsensusFactory consensusFactory, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION)
```
dangershony: I think maybe perhaps we should create a  default `ConsensusFactory` that will just return null (or throw when one of the types it expect is asked) for every type.
then this method can be called with no `ConsensusFactory` and still work.

src/NBitcoin/ConsensusFactory.cs
fassadlr: `Consensus` still gets used in the `Coins` class to check the `IsProofOfStake` property. Maybe we can look at removing this or find a work around in the next PR?

src/NBitcoin/BlockStake.cs
fassadlr: OK yeah I have created a `BlockStake.Load` that takes in either `byte[]` and `Block`.

src/NBitcoin/IBitcoinSerializable.cs
fassadlr: Hmmm I'm really not a fan of these methods that can be called with null defaults. I think in the solution this is something that we should be moving away from. I would much rather create another `ReadWrite` where we don't pass in a `ConsensusFactory`, should the need arise. This way we are more explicit as to the intent of the method.

src/NBitcoin/BlockStake.cs
bokobza: why Load?

fassadlr: @bokobza This is to keep the pattern consistent throughout. We load blocks and block headers like this. Also, it lessens the footprint that the `Network` class has in NBitcoin. We can rename it to something else?

src/NBitcoin/IBitcoinSerializable.cs
dangershony: But there are many tyoes that use the serializer apart for the (Header,Block,Transaction) those dont need the factory.

src/NBitcoin/BlockStake.cs
```diff
@@ -166,6 +180,11 @@ public PosTransaction(byte[] bytes) : this()
     /// </summary>
     public class PosConsensusFactory : ConsensusFactory
     {
+        public PosConsensusFactory(Network network)
```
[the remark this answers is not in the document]
fassadlr: Will remove 👍

src/NBitcoin/IBitcoinSerializable.cs
fassadlr: Yes I understand that :) but I think that the part you are referring to is here:

https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/NBitcoin/BitcoinStream.cs#L194

So here, if we are serializing/deserializing an object that is NOT Header,Block,Transaction we will just use the Activator?

Why would all the tests pass and this branch be able to sync if this was an issue? (Just asking)

:)

fassadlr:
@dangershony staking still works as per normal as well as all the integration tests pass 👍

src/NBitcoin/BIP32/BitcoinExtKey.cs
```diff
@@ -67,7 +67,7 @@ public ExtKey ExtKey
                 if(this._Key == null)
                 {
                     this._Key = new ExtKey();
-                    this._Key.ReadWrite(this.vchData);
+                    this._Key.ReadWrite(this.vchData, network: this.Network);
```
dangershony: After the change we made with DefaultConsensusFactory we don't need network in this code

src/NBitcoin/Consensus.cs
```diff
@@ -101,7 +101,7 @@ public BIP9DeploymentsArray()
             }
         }
 
-        public Consensus()
+        public Consensus(Network network)
```
dangershony: I think you can remove Network form the constructor

src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -55,9 +68,9 @@ public static int GetSerializedSize(this IBitcoinSerializable serializable, Prot
             return GetSerializedSize(serializable, version, SerializationType.Disk);
         }
 
-        public static void ReadWrite(this IBitcoinSerializable serializable, byte[] bytes, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null)
+        public static void ReadWrite(this IBitcoinSerializable serializable, byte[] bytes, Network network, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION)
```
dangershony: You probably want to create another method that does not need network (and instead use DefaultConsensuFactory).

src/NBitcoin/Script.cs
```diff
@@ -409,14 +409,14 @@ public Script(IEnumerable<byte> data)
             this._Script = data.ToArray();
         }
 
-        public Script(byte[] data, bool compressed)
+        public Script(Network network, byte[] data, bool compressed)
```
dangershony: Script will never get any of the types that are in ConsensusFactory so we dont need to pass network in this case)

Use the override without network when calling `compressor.ReadWrite`

src/NBitcoin/DefaultConsensusfactory.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace NBitcoin
+{
+    /// <summary>
+    /// A default object factory to create instances that is not block, block header or transaction.
+    /// </summary>
+    public sealed class DefaultConsensusFactory : ConsensusFactory
```
dangershony: Super duper this makes our serializer much easier to use

src/NBitcoin/BIP32/ExtPubKey.cs
```diff
@@ -72,11 +72,11 @@ internal ExtPubKey()
         {
         }
 
-        public ExtPubKey(byte[] bytes)
+        public ExtPubKey(Network network, byte[] bytes)
```
dangershony: I think we can remove this from the constructor `ExtPubKey` is not one of the types in consensus factory.

PR #1547, opened 2018-06-21
Optional parameter never used optionally

majikandy:
Swapped the parameter order to the Create HDAddress path as it had an optional parameter that was never used optionally. Therefore optional changed to explicit.

should be tagged very-low


bokobza:
Excellent, thanks mate.

PR #1549, opened 2018-06-21
Added endpoint to show smart contract logs

majikandy:


src/Stratis.Bitcoin/Controllers/DashboardController.cs
```diff
@@ -9,24 +12,49 @@ namespace Stratis.Bitcoin.Controllers
...
+        /// <returns>text/html content</returns>
+        [HttpGet]
+        [Route("SmartContracts/{numberOfLogEntriesToShow?}")] 
+        public IActionResult SmartContractLogs(int numberOfLogEntriesToShow = 30)
+        {
+            string logPath = Path.Combine(this.nodeSettings.DataDir, @"Logs\smartcontracts.txt");
+
+            if (!System.IO.File.Exists(logPath))
+            {
+                return this.Content($"There is no log file at: {logPath}. An nlog.config is needed in the daemon directory.");
```
ferdeen: An nlog.config **_file_**...

bokobza:
Is it ok that this is in the full node? @fassadlr

bokobza:
Also, I don't think this is the right place to put this endpoint.
Each feature will have its own UI, so dumping a smart contract related thing in the Stratis.Bitcoin is probably not the right thing to do. @majikandy @fassadlr

fassadlr:
@bokobza this only a merge to our sc-alpha branch ;) but you are 100% correct the controller should go in the Smart Contract Feature. @majikandy can you please move it there and see that it works?

PR #1552, opened 2018-06-22
Dashboard feature

dippyeggs:
Here lie a possible implementation of dashboard feature as specified at SDF.

The dashboard is it's own feature which serve files from `wwwroot` which enable developer to create own front-end app in whatever language/framework they want.

Specification for this bounty not clear as to if new feature is needed, or whether dashboard should leverage API feature.  Discussion in PR was conflicting against original specification in SDF.

Implementation similar to API implementation and share some code already used here.

NOTE: Its probably all wrong, and not a good solution, but who know perhaps it is worth something.

Dashboard feature rely on APIFeature, so should be able to use REST endpoints to build dashboard when all running.

Dippyeggs all around?


dangershony:
Ah I see so the design is that we have a wwwroot folder that has the website files.
And when calling the endpoint dashboard/home the wwwroot folder is served as a website.

I suggest the following, instead of looking in the current node location we look in the current node data folder.

The structure can be as following.

The dashboard endpoint will become `/loadplugin/[plugin_name]`
This will look in `datafolder/plugins/[plugin_name]`
Is the plugin is found it will be served to the caller.
This will allow devs to just deploy plugins to the plugins (wwwroot) folder

Would that work @dippyeggs ?

dippyeggs:
Hello Dan, thanks for your time and reply.

Of course I think it will work.

We can have `NodeSettings` stored at runtime, then set static route of assets to `NodeSettings.DataDir `folder rather than feature folder..

We can then modify the controller method to take parameter of plugin name.  Also we can use razor engine to parse a `.cshtml` file rather than `.html` which will point the assets tin `plugins` inside `wwwroot` which will now live inside `NodeSettings.DataDir`.

Which means we can call `/loadplugin/dashboard`.

This will load assets inside `NodeSettings.DataDir\plugins\dashboard`  and we serve the application.

How does this sound?

dangershony:
Yes exactly what I think of, then you just drop (even a compiled razor once we upgrade to 2.1)  site as a plugin in to the flder

By the way I experimented with blazor and manged to load it form the node self host server, see this branch:
https://github.com/dangershony/StratisBitcoinFullNode/tree/blazor

rjmacarthy:
Thanks Dan. Right ok will have a look then when I get time. Perhaps we can look into blazor but from github page I can see it's experimental. It would be more beneficial though long term if they move to release.

dippyeggs:
Thanks dan. Yes will look into it. Blazor looks like experimental project but may be good going forward. Btw this is @rjmacarthy but use this account for some other purposes just let you know for discord etc.

dippyeggs:
You can now pass name of plugin folder to controller method. It will look in data dir for folder of parameter and render the page. Had to opt for handlebars.net not razor in order to render the correct assets path inside index.html file. Thoughts? I looked into blazor but would need to spend considerable amount of time which I don't have at the moment to get working.

dangershony:
Excellent @herbepau will check it out as well.

src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,28 @@
+﻿
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.IO;
```
fassadlr: I don't think that all these usings are required? Can you download and install Productivity Power Tools for Visual Studio? There is a setting where you can remove and sort usings on save, pretty cool!

src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,28 @@
...
+using HandlebarsDotNet;
+
+namespace Stratis.Bitcoin.Features.Dashboard.Controllers
+{
+    public class DashboardController : Controller
+    {
+        [HttpGet("/_plugins/{pluginName}")]
+        public ContentResult Index(string pluginName)
+        {
+
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,28 @@
...
+            System.Func<object, string> result = Handlebars.Compile(content);
+            var html = result(new { pluginName });
+            return new ContentResult
+            {
+                ContentType = "text/html",
+                StatusCode = (int)HttpStatusCode.OK,
+                Content = html
+            };
+        }
+
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/DashboardApiSettings.cs
```diff
@@ -0,0 +1,156 @@
...
+                this.KeepaliveTimer = new Timer
+                {
+                    AutoReset = false,
+                    Interval = keepAlive * 1000
+                };
+            }
+
+            this.callback?.Invoke(this);
+        }
+
```
fassadlr: Please remove the extra line.

src/Stratis.Bitcoin.Features.Dashboard/DashboardApiSettings.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// Get root directory
+        /// </summary>
+        /// <returns>The feature root directory.</returns>
+        public static string GetRootDirectory()
+        {
+            var path = Directory.GetParent(Directory.GetCurrentDirectory()).FullName + "\\Stratis.Bitcoin.Features.Dashboard";
+            if (Directory.Exists(path))
+            {
+                return path;
+            } else
```
fassadlr: `Else` should be on a new line here.

src/Stratis.Bitcoin.Features.Dashboard/DashboardFeature.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.Extensions.Logging;
+using Stratis.Bitcoin.Builder;
+using Stratis.Bitcoin.Builder.Feature;
+using Stratis.Bitcoin.Features.Api;
+
+namespace Stratis.Bitcoin.Features.Dashboard
+{
+    public class DashboardFeature : FullNodeFeature
+    {
+
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/DashboardFeature.cs
```diff
@@ -0,0 +1,85 @@
...
+                .UseKestrel()
+                .UseIISIntegration()
+                .UseUrls(apiUri.ToString())
+                .UseStartup<Startup>()
+                .Build();
+
+            this.host.Start();
+
+            this.logger.LogInformation("Stratis Dashboard Initialized on http://localhost:" + this.apiSettings.ApiPort);
+
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/DashboardFeature.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            if (this.host != null)
+            {
+                this.logger.LogInformation("Dashboard stopping on URL '{0}'.", this.apiSettings.ApiUri);
+                this.host.StopAsync(TimeSpan.FromSeconds(ApiStopTimeoutSeconds)).Wait();
+                this.host = null;
+            }
+        }
+
```
fassadlr: Please remove the empty line.

src/Stratis.Bitcoin.Features.Dashboard/Startup.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
```
fassadlr: Please cleanup unused `usings`.

src/Stratis.Bitcoin.Features.Dashboard/DashboardApiSettings.cs
```diff
@@ -0,0 +1,156 @@
+﻿using System;
```
fassadlr: Please cleanup unused `usings` and sort them.

bokobza:
> The dashboard endpoint will become /loadplugin/[plugin_name]
This will look in datafolder/plugins/[plugin_name]
Is the plugin is found it will be served to the caller.

I think all of these should be done automatically when the node starts and the dashboard can stay at /dashboard.
I don't see a viable solution where the user, be it a person or the app needs to call  /loadplugin/[plugin_name] every time.

dangershony:
I think this PR is obsolete now as we have the stratis apps being introduced.

@dippyeggs please find me for a pm on discord.

noescape00:
Closing it since it's obsolete.

PR #1553, opened 2018-06-23
Don't process tx when in block only mode

mikedennis:
This is continuation of refactoring for issue #1250 

Add BTC core logic for receiving transactions that skips processing them when in blocks only mode. Relevant BTC logic is here https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L2173-#L2181

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -293,10 +293,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
                 return;
             }
 
-            bool blocksOnly = !this.connectionManager.ConnectionSettings.RelayTxes;
-            // Allow whitelisted peers to send data other than blocks in blocks only mode if whitelistrelay is true
-            if (peer.Behavior<ConnectionManagerBehavior>().Whitelisted && this.mempoolManager.mempoolSettings.WhiteListRelay)
-                blocksOnly = false;
+            bool blocksOnly = this.IsBlocksOnly(peer);
 
             //uint32_t nFetchFlags = GetFetchFlags(pfrom, chainActive.Tip(), chainparams.GetConsensus());
 
@@ -367,6 +364,24 @@ namespace Stratis.Bitcoin.Features.MemoryPool
         }
 
         /// <summary>
+        /// Whether peer should receive only blocks.
+        /// </summary>
+        /// <param name="peer">Peer to check.</param>
+        /// <returns>Whether peer should receive only blocks.</returns>
+        private bool IsBlocksOnly(INetworkPeer peer)
+        {
+            this.logger.LogTrace("({0}:'{1}'", nameof(peer), peer.RemoteSocketEndpoint);
+            bool isBlocksOnly = !this.connectionManager.ConnectionSettings.RelayTxes;
+
+            // Allow whitelisted peers to send data other than blocks in blocks only mode if whitelistrelay is true
+            if (peer.Behavior<ConnectionManagerBehavior>().Whitelisted && this.mempoolManager.mempoolSettings.WhiteListRelay)
+                isBlocksOnly = false;
+
+            this.logger.LogTrace("(-):{0}", isBlocksOnly);
+            return isBlocksOnly;
+        }
+
+        /// <summary>
         /// Processing of the transaction payload message from peer.
         /// Adds transaction from the transaction payload to the memory pool.
         /// </summary>
@@ -375,6 +390,15 @@ namespace Stratis.Bitcoin.Features.MemoryPool
         private async Task ProcessTxPayloadAsync(INetworkPeer peer, TxPayload transactionPayload)
         {
             this.logger.LogTrace("({0}:'{1}',{2}.{3}:{4})", nameof(peer), peer.RemoteSocketEndpoint, nameof(transactionPayload), nameof(transactionPayload.Obj), transactionPayload?.Obj?.GetHash());
+
+            // Stop processing the transaction early if we are in blocks only mode.
+            if (this.IsBlocksOnly(peer))
+            {
+                this.logger.LogInformation("Transaction sent in violation of protocol, peer:'{0}'.", peer.RemoteSocketEndpoint);
+                this.logger.LogTrace("(-)[BLOCKSONLY]");
+                return;
+            }
+
             Transaction trx = transactionPayload.Obj;
             uint256 trxHash = trx.GetHash();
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -366,6 +363,24 @@ public override object Clone()
...
+        /// </summary>
+        /// <param name="peer">Peer to check.</param>
+        /// <returns>Whether peer should receive only blocks.</returns>
+        private bool IsBlocksOnly(INetworkPeer peer)
+        {
+            this.logger.LogTrace("({0}:'{1}'", nameof(peer), peer.RemoteSocketEndpoint);
+            bool isBlocksOnly = !this.connectionManager.ConnectionSettings.RelayTxes;
+
+            // Allow whitelisted peers to send data other than blocks in blocks only mode if whitelistrelay is true
+            if (peer.Behavior<ConnectionManagerBehavior>().Whitelisted && this.mempoolManager.mempoolSettings.WhiteListRelay)
```
dangershony: This can be improved, when calling the method AttachCore you could find this behaviour and set a flag on the current behaviour, then we wont need to look it up every time a trx is incoming.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -307,6 +304,20 @@ public void RemoveForBlock(List<Transaction> transactionsToRemove)
...
         }
 
+        /// <summary>
+        /// Adds transaction hash to recent rejects.
+        /// </summary>
+        /// <param name="orphanHash">Hash to add.</param>
+        public void AddToRecentRejects(uint256 orphanHash)
+        {
+            this.logger.LogTrace("{0}:'{1}'", nameof(orphanHash), orphanHash);
+            lock (this.lockObject)
```
dangershony: Put a space before the lock (for readability)

noescape00:
Do we even need to have MempoolBehavior in those cases? Maybe just don't add it to template behaviors in case `RelayTxes == false && WhiteListRelay == false`

That way we won't do useless work.

dangershony:
But how will you know which are WhiteListRelay?

noescape00:
>But how will you know which are WhiteListRelay?

WhiteListRelay is not a per-peer variable, it's per mempool. So in case it's false we won't relay to any of the whitelisted peers.

mikedennis:
What about the other mempool behavior messages? There are inventory and other messages.

noescape00:
>What about the other mempool behavior messages?

Not sure. That requires investigation.

mikedennis:
Took a quick look - For inventory message for example: In btc core (and in our implementation) if it's blocksonly it still adds the txs to the known inventory even if it doesn't relay it onto the attached peer. Not sure I understand reasoning why it just doesn't ignore it completely but that seems to be the current logic.

noescape00:
>Took a quick look - For inventory message for example: In btc core (and in our implementation) if it's blocksonly it still adds the txs to the known inventory even if it doesn't relay it onto the attached peer. Not sure I understand reasoning why it just doesn't ignore it completely but that seems to be the current logic.

Hmm, weird. The only reason for this logic that I can see is probably checking if tx was accepted to the mempool (might be useful for SPV clients that can't validate their own txes). 

So it looks like we can't just disable the behavior, ok.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -366,6 +371,21 @@ public override object Clone()
...
+        /// <summary>
+        /// Whether peer should receive only blocks.
+        /// </summary>
+        /// <param name="peer">Peer to check.</param>
+        /// <returns>Whether peer should receive only blocks.</returns>
+        private bool IsBlocksOnly(INetworkPeer peer)
+        {
+            this.logger.LogTrace("({0}:'{1}'", nameof(peer), peer.RemoteSocketEndpoint);
+
+            bool isBlocksOnly = !this.connectionManager.ConnectionSettings.RelayTxes && !this.isPeerWhitelistedForRelay;
```
dangershony: This is not suppose to change so why not do it as a parameter as well similar to `isPeerWhitelistedForRelay`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -406,11 +435,34 @@ public override object Clone()
...
+                // if they were already in the mempool or rejected from it due
+                // to policy, allowing the node to function as a gateway for
+                // nodes hidden behind it.
+                //
+                // Never relay transactions that we would assign a non-zero DoS
+                // score for, as we expect peers to do the same with us in that
+                // case.
+                if (this.isPeerWhitelistedForRelay)
+                {
+                    if (!state.IsInvalid)
```
dangershony: Can state be valid and we come to this point?
Ah this is in case `this.orphans.AlreadyHaveAsync` already has the trx?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -292,20 +307,16 @@ public override object Clone()
...
+            
             //uint32_t nFetchFlags = GetFetchFlags(pfrom, chainActive.Tip(), chainparams.GetConsensus());
 
             var send = new GetDataPayload();
             foreach (InventoryVector inv in invPayload.Inventory.Where(inv => inv.Type.HasFlag(InventoryType.MSG_TX)))
             {
                 //inv.type |= nFetchFlags;
 
-                if (blocksOnly)
+                // TODO: This is incorrect, in blocks only mode we should just add to known inventory but not relay
```
mikedennis: This blocksOnly logic is different than BTC core. Will fix up in my next PR.

noescape00:
Yeah, looks good. Didn't run the tests and synced the chains yet. Will do.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolOrphans.cs
```diff
@@ -307,6 +304,22 @@ public void RemoveForBlock(List<Transaction> transactionsToRemove)
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>
+        /// Adds transaction hash to recent rejects.
+        /// </summary>
+        /// <param name="orphanHash">Hash to add.</param>
+        public void AddToRecentRejects(uint256 orphanHash)
+        {
+            this.logger.LogTrace("{0}:'{1}'", nameof(orphanHash), orphanHash);
```
noescape00: "{0}:'{1}'" => "({0}:'{1}')"

noescape00:
Tests are passing, chains synced.

PR #1556, opened 2018-06-25
CHT tests issues #7, #10

thefazzer:
**Issues 7 & 10** (#1321)
(7)  _We have a chain, someone presents an invalid header. After that our chain's last block shouldn't change and it shouldn't have .Next and it throws_
(10) _When checkpoints are enabled and the only checkpoint is at block X, when we present headers before that- none marked for download. When we first present checkpointed header and some after- all previous are also marked for download & those that are up to the last checkpoint are marked as assumevalid_

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -176,6 +176,13 @@ namespace Stratis.Bitcoin.Tests.Consensus
             }
         }
 
+        public class InvalidHeaderTestException : ConsensusException
+        {
+            public InvalidHeaderTestException() : base()
+            {
+            }
+        }
+
         [Fact]
         public void ConnectHeaders_HeadersCantConnect_ShouldFail()
         {
@@ -354,6 +361,50 @@ namespace Stratis.Bitcoin.Tests.Consensus
 
             // Last block shouldn't have a Next.
             Assert.Empty(peerTipAfterInvalidHeaderPresented.Next);
+        }
+
+        /// <summary>
+        /// Issue 10 @ Create chained header tree component #1321
+        /// When checkpoints are enabled and the only checkpoint is at block X,
+        /// when we present headers before that none are marked for download.
+        /// When we first present checkpointed header and some after
+        /// all previous are also marked for download & those that are up
+        /// to the last checkpoint are marked as assumevalid.
+        /// </summary>
+        [Fact]
+        public void PresentChain_CheckpointsEnabled_MarkToDownloadWhenCheckpointPresented()
+        {
+            const int initialChainSize = 2;
+            const int currentChainExtension = 7;
+
+            TestContext testContext = new TestContextBuilder().WithInitialChain(initialChainSize).UseCheckpoints().Build();
+            ChainedHeaderTree chainedHeaderTree = testContext.ChainedHeaderTree;
+            ChainedHeader initialChainTip = testContext.InitialChainTip;
+
+            ChainedHeader extendedChainTip = testContext.ExtendAChain(currentChainExtension, initialChainTip);
+            List<BlockHeader> listOfCurrentChainHeaders = testContext.ChainedHeaderToList(extendedChainTip, initialChainSize + currentChainExtension);
+            
+            // Checkpoints are enabled and the only checkpoint is at block 6
+            const int firstCheckpointHeight = 6;
+            var checkpoint = new CheckpointFixture(firstCheckpointHeight, listOfCurrentChainHeaders[firstCheckpointHeight - 1]);
+            testContext.SetupCheckpoints(checkpoint);
+            
+            // When we present headers before that none are marked for download
+            int numberOfHeadersBeforeCheckpoint = firstCheckpointHeight - initialChainSize;
+            var listOfHeadersBeforeCheckpoint = listOfCurrentChainHeaders.GetRange(initialChainSize, numberOfHeadersBeforeCheckpoint - 1);
+            ConnectNewHeadersResult connectNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, listOfHeadersBeforeCheckpoint.ToList());
+            connectNewHeadersResult.DownloadFrom.Should().Be(null);
+            connectNewHeadersResult.DownloadTo.Should().Be(null);
+
+            // Headers up to the last checkpoint are marked as assumevalid.
+            testContext.ConsensusSettings.BlockAssumedValid = listOfCurrentChainHeaders[firstCheckpointHeight - 1].GetHash();
+
+            // When we present the checkpointed header and those beyond all previous are a marked for download
+            List<BlockHeader> listOfHeadersIncludingAndAfterCheckpoint = listOfCurrentChainHeaders.TakeLast(4).ToList();
+            connectNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, listOfHeadersIncludingAndAfterCheckpoint.ToList());
+            
+            connectNewHeadersResult.DownloadFrom.HashBlock.Should().Be(listOfHeadersBeforeCheckpoint.First().GetHash());
+            connectNewHeadersResult.DownloadTo.HashBlock.Should().Be(checkpoint.Header.GetHash());
         }
 
         /// <summary>
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -11,6 +11,7 @@ using Stratis.Bitcoin.Configuration;
 using Stratis.Bitcoin.Configuration.Logging;
 using Stratis.Bitcoin.Configuration.Settings;
 using Stratis.Bitcoin.Consensus;
+using Stratis.Bitcoin.Tests.Common;
 using Stratis.Bitcoin.Utilities;
 using Xunit;
 
@@ -428,6 +429,14 @@ namespace Stratis.Bitcoin.Tests.Consensus
             while (chainedHeader.Height > initialChainSize)
             {
                 chainedHeader.BlockValidationState.Should().Be(ValidationState.AssumedValid);
+                chainedHeader = chainedHeader.Previous;
+            }
+
+            // Checking from the checkpoint forward to the end of the chain.
+            chainedHeader = chainedHeaderTree.GetPeerTipChainedHeaderByPeerId(1);
+            while (chainedHeader.Height > checkpoint.Height)
+            {
+                chainedHeader.BlockValidationState.Should().Be(ValidationState.HeaderValidated);
                 chainedHeader = chainedHeader.Previous;
             }
         }
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -421,46 +421,6 @@ namespace Stratis.Bitcoin.Tests.Consensus
         }
 
         /// <summary>
-        /// Issue 11 @ Create chained header tree component #1321
-        /// When checkpoints are enabled and the first is at block X,
-        /// when we present headers before that- none marked for download.
-        /// When we first present checkpointed header and some after- only
-        /// headers before the first checkpoint(including it) are marked for download
-        /// </summary>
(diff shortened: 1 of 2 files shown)


Aprogiena:
Number seven looks good, number 10 we have explained on slack how to check for the block header validation state

thefazzer:
@Aprogiena - Have made AssumedValid changes as discussed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            TestContext testContext = new TestContextBuilder().WithInitialChain(initialChainSize).UseCheckpoints().Build();
+            ChainedHeaderTree chainedHeaderTree = testContext.ChainedHeaderTree;
+            ChainedHeader initialChainTip = testContext.InitialChainTip;
+
+            ChainedHeader extendedChainTip = testContext.ExtendAChain(currentChainExtension, initialChainTip);
+            List<BlockHeader> listOfCurrentChainHeaders =
+                testContext.ChainedHeaderToList(extendedChainTip, initialChainSize + currentChainExtension);
+
+            // Checkpoints are enabled and the only checkpoint is at block 20
```
Aprogiena: Use dot at the end of the sentence, not only here.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            const int checkpointHeight = 20;
+            var checkpoint = new CheckpointFixture(checkpointHeight, listOfCurrentChainHeaders[checkpointHeight - 1]);
+            testContext.SetupCheckpoints(checkpoint);
+
+            // When we present headers before that none are marked for download
+            int numberOfHeadersBeforeCheckpoint = checkpointHeight - initialChainSize;
+            var listOfHeadersBeforeCheckpoint = 
+                listOfCurrentChainHeaders.GetRange(initialChainSize, numberOfHeadersBeforeCheckpoint - 1);
```
Aprogiena: Why -1 here?

But feel free to include some blocks of the initial chain.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            ConnectNewHeadersResult connectNewHeadersResult = 
+                chainedHeaderTree.ConnectNewHeaders(1, listOfHeadersBeforeCheckpoint.ToList());
+            connectNewHeadersResult.DownloadFrom.Should().Be(null);
+            connectNewHeadersResult.DownloadTo.Should().Be(null);
+
+            // Headers up to the last checkpoint are marked as assumevalid.
+            testContext.ConsensusSettings.BlockAssumedValid = listOfCurrentChainHeaders.Last().GetHash();
```
Aprogiena: This is incorrect, you should not touch the settings at all here.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: The -1 here looks suspicious but we actually want to present something that ends before checkpoint and even have some gap between the last header be present at first and the checkpoint. So we could even go with -5.

Aprogiena: It would really help if there was such a comment with the diagram as we have it in other tests so it would be easy to see what was your intention.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: Also this point we actually want to check that all the headers have the foundation state of header only – except for those initial chain which are by definition fully validated

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            // When we present the checkpointed header and those beyond all previous are a marked for download
+            List<BlockHeader> listOfHeadersIncludingAndAfterCheckpoint = listOfCurrentChainHeaders.TakeLast(6).ToList();
```
Aprogiena: Here I would expect the number six be calculated based on what we presented before.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            connectNewHeadersResult =
+                chainedHeaderTree.ConnectNewHeaders(1, listOfHeadersIncludingAndAfterCheckpoint.ToList());
```
Aprogiena: To list is not necessary here.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,132 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            connectNewHeadersResult.DownloadFrom.HashBlock.Should().Be(listOfHeadersBeforeCheckpoint.First().GetHash());
+            connectNewHeadersResult.DownloadTo.HashBlock.Should().Be(listOfHeadersIncludingAndAfterCheckpoint.Last().GetHash());
+
+            const ValidationState expectedState = ValidationState.AssumedValid;
+
+            ChainedHeader consumed = connectNewHeadersResult.Consumed;
+            while (consumed.Height > initialChainSize)
```
Aprogiena: Here it seems that you are expecting all headers to be assumed valid, but that is not the case, only the headers up to the checkpoint.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: I'm commenting just because you updated this. This should not be here at all. In this test we are not setting the assume valid block hash.

Aprogiena:
Number 10 still needs some polishing and fixing

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: All sentences now end with.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Used -5 beneath checkpoint.  Comments added and hopefully some clarity.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: No longer hard coding any of these.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Occurrences of redundant ToList() removed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: Restricted this test to the relevant range.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,104 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+        /// We have a chain and someone presents an invalid header.
+        /// After that our chain's last block shouldn't change, it shouldn't have a valid .Next
+        /// and it should throw an exception.
+        /// </summary>
+        [Fact]
+        public void ConnectHeaders_SupplyInvalidHeader_ExistingChainTipShouldNotChange()
+        {
+            TestContext testContext = new TestContextBuilder().WithInitialChain(5).UseCheckpoints(false).Build();
+            ChainedHeaderTree chainedHeaderTree = testContext.ChainedHeaderTree;
+            uint256 peerTipHashBeforeInvalidBlockPresented = testContext.InitialChainTip.HashBlock;
```
noescape00: probably name it `consensusTip` and get ChainedHeader and not the hash there

Also
`peerTipAfterInvalidHeaderPresented` => `consensusTipAfterInvalidHeaderPresented`

and instead of 
`Assert.Equal(peerTipHashBeforeInvalidBlockPresented, peerTipAfterInvalidHeaderPresented.HashBlock);`

put there those 2 headers- their hashes will be checked to be equal (ChainedHeader has override on ==)

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,104 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+
+            connectNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, unconsumedHeaders);
+
+            connectNewHeadersResult.DownloadFrom.HashBlock.Should().Be(listOfHeadersBeforeCheckpoint.First().GetHash());
+            connectNewHeadersResult.DownloadTo.HashBlock.Should().Be(unconsumedHeaders.Last().GetHash());
+            
+            ChainedHeader chainedHeader = chainedHeaderTree.GetChainedHeadersByHash()
+                .SingleOrDefault(x => (x.Value.HashBlock == checkpoint.Header.GetHash())).Value;
+
+            // Checking from the checkpoint back to the initialised chain.
```
noescape00: `initialised` - typo

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: @noescape00 All done.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
thefazzer: @noescape00 - corrected.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -327,6 +334,104 @@ public void ConnectHeaders_SupplyHeaders_ToDownloadArraySizeSameAsNumberOfHeader
...
+            
+            ChainedHeader chainedHeader = chainedHeaderTree.GetChainedHeadersByHash()
+                .SingleOrDefault(x => (x.Value.HashBlock == checkpoint.Header.GetHash())).Value;
+
+            // Checking from the checkpoint back to the initialized chain.
+            while (chainedHeader.Height > initialChainSize)
+            {
+                chainedHeader.BlockValidationState.Should().Be(ValidationState.AssumedValid);
+                chainedHeader = chainedHeader.Previous;
+            }
```
Aprogiena: The tests look good now, but in this test number 10 I would even add a check, here and, that would verify that headers after the check point are in the right, validation state, which would mean header validated only. This is just trivial extension and would make the test stronger.

thefazzer: @Aprogiena Done

PR #1559, opened 2018-06-25
CHT test: BlockDataDownload for null pointer - issue 23

fenix2222:
Block data received (BlockDataDownloaded is called) for already FV block with null pointer. Nothing should be chained, pointer shouldn't be assigned. Should return false

See #1321 (comment)


Aprogiena:
Looks okay to me

PR #1560, opened 2018-06-25
CHT bugfix

noescape00:


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -325,6 +325,23 @@ public List<ChainedHeader> PartialValidationSucceeded(ChainedHeader chainedHeade
...
+                this.logger.LogTrace("(-)[PREV_BLOCK_NOT_VALIDATED]:null");
+                return null;
+            }
+
+            // Same scenario as above except for prev block was validated which triggered next partial validation to be started.
+            if ((chainedHeader.BlockValidationState == ValidationState.PartiallyValidated) || 
+                (chainedHeader.BlockValidationState == ValidationState.FullyValidated))
+            {
+                this.logger.LogTrace("(-)[ALREADY_VALIDATED]:null");
+                return null;
```
dangershony: In this case we should return an empty `headersToValidate` no?

noescape00: no, why making an additional allocation?

PR #1562, opened 2018-06-25
CHT test: BlockDataDownload for valid blocks - issue 24

fenix2222:
BlockDataDownloaded called for some blocks. Make sure CH.Block is not null and for the first block true is returned and false for others

See #1321 (comment)

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -126,6 +126,13 @@ namespace Stratis.Bitcoin.Tests.Consensus
                     .Returns(checkpoints.OrderBy(h => h.Height).Last().Height);
             }
 
+            public Block CreateBlock()
+            {
+                Block block = this.Network.Consensus.ConsensusFactory.CreateBlock();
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -692,5 +699,44 @@ public void BlockDataDownloadedIsCalled_ForFvBlockWithNullPointer_ResultShouldBe
...
         }
+
+        /// <summary>
+        /// Issue 24 @ BlockDataDownloaded called for some blocks. Make sure CH.Block is not null and for the
+        /// first block true is returned and false for others.
+        /// </summary>
+        [Fact]
+        public void BlockDataDownloadedIsCalled_ForValidBlocksAfterFv_ResultShouldBeTrueForTHeFirstAndFalseForTheRest()
+        {
+            // Chain header tree setup. Initial chain has 4 headers with no blocks.
```
Aprogiena: In this test we don't have any requirement of not having blocks on the initial chain, therefore I would just remove it from the comment as well as from the code as it is not the default behaviour.

Aprogiena:
Just tiny detail, otherwise the test looks good.

PR #1564, opened 2018-06-26
Gap limit integration tests of current behaviour

majikandy:
2 tests.

1.  Funds outside the gap limit not visible

2.  Generate addresses in advance, then funds at that same address are visible

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/Wallet/CustomisableAddressGapLimitBeyondBIP44.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Wallet/CustomisableAddressGapLimitBeyondBIP44.cs
@@ -9,9 +9,13 @@ namespace Stratis.Bitcoin.IntegrationTests.Wallet
         public void Coins_beyond_gap_limit_not_visble_in_balance()
         {
             Given(a_default_gap_limit_of_20);
-            And(a_wallet_with_funds_at_index_21);
+            And(a_wallet_with_funds_at_index_20);
             When(getting_wallet_balance);
             Then(the_balance_is_zero);
+
+            Given(_21_new_addresses_are_requested);
+            When(getting_wallet_balance);
+            Then(the_balance_is_no_longer_zero);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin.IntegrationTests/Wallet/CustomisableAddressGapLimitBeyondBIP44.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Xunit;
+// ReSharper disable once InconsistentNaming
+
+namespace Stratis.Bitcoin.IntegrationTests.Wallet
+{
+    public partial class Customisable_address_gap_limit_beyond_BIP44
```
bokobza: These tests are not really about the customizable gap limit but rather about 1. testing the gap and 2. transactions being found when the number of addresses goes beyond the gap limit.
If you intend to add the gap limit testing tests here, then the name is fine.

majikandy: Yes good point, I was thinking of the future - but you are right, right now it isn't - I will change the name

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletAddressGenerationAndFundsVisibility_Steps.cs
```diff
@@ -0,0 +1,130 @@
...
+            this.sendingStratisBitcoinNode.FullNode.Network.Consensus.CoinbaseMaturity = coinbaseMaturity;
+            this.receivingStratisBitcoinNode.FullNode.Network.Consensus.CoinbaseMaturity = coinbaseMaturity;
+
+            this.sharedSteps.MineBlocks(coinbaseMaturity + 1, this.sendingStratisBitcoinNode, AccountZero, SendingWalletName, WalletPassword);
+        }
+
+        private void a_wallet_with_funds_at_index_20_which_is_beyond_gap_limit()
+        {
+            ExtPubKey xPublicKey = this.GetExtendedPublicKey(ReceivingNodeName);
+            var recipientAddressBeyondGapLimit = xPublicKey.Derive(new KeyPath($"0/{20}")).PubKey.GetAddress(Network.RegTest);
```
YakupIpek: String interpolation is not needed in here.

majikandy: good point, I'll correct that

PR #1567, opened 2018-06-26
CHT test: Issue 25 26

fenix2222:
Consensus and headers are at block 5. Block 6 is presented and PV is successful. Call PartialOrFullValidationFailed and after make sure that there is a single US_CONSTANTs on CT.

CT is at block 5. Call PartialValidationSucceeded with header 6. ConsensusTipChanged on block 6. Make sure US_CONSTANT moved to 6.

See #1321 (comment)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -825,27 +824,91 @@ public void BlockDataDownloadedIsCalled_ForValidBlocksAfterFv_ResultShouldBeTrue
...
+
+            // Present header by peer with id 1 and then call PartialOrFullValidationFailed on it.
+            const int peerId = 1;
+            cht.ConnectNewHeaders(peerId, listOfCurrentChainHeaders);
+            List<int> validationResult = cht.PartialOrFullValidationFailed(chainTip);
+
+            // Validation result should contain 1 result pointing to CT.
+            validationResult.Should().HaveCount(1);
+            validationResult.First().Should().Be(peerId);
+        }
```
Aprogiena: In this test, we first need to call partial validation succeeded and then we want to call partial of validation failed on the same block header. Then we need to check the internal structures looking for the local PID. It must be found only once on header five and must not be present on six.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -825,27 +824,91 @@ public void BlockDataDownloadedIsCalled_ForValidBlocksAfterFv_ResultShouldBeTrue
...
+
+            // Present header by peer with id 1 and then call PartialValidationSucceeded on it.
+            const int peerId = 1;
+            cht.ConnectNewHeaders(peerId, listOfCurrentChainHeaders);
+            cht.PartialValidationSucceeded(chainTip, out bool reorgRequired);
+            reorgRequired.Should().BeTrue();
+
+            // Call ConsensusTipChanged on chaintip at header 6.
+            cht.ConsensusTipChanged(chainTip);
+            cht.GetPeerTipsByPeerId().Should().ContainKey(1);
```
Aprogiena: Similarly as above, we need to check which claims are present on the five and which claims are present on header six. The local PID is supposed to move from 5 to 6. So the end it should not be five anymore it should be on six.

Aprogiena: see peerIdsByTipHash

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
Aprogiena: see peerIdsByTipHash

fenix2222: This is not what test description was saying. Ok I will change it.

Aprogiena: The descriptions are sometimes cryptic– This is why I said that the descriptions put into the summary should not be just a copy and paste, because it might not make good sense. The description is not clear, it is better to ask for explanation before starting implementation. Here the description says that PV was successful, meaning that partial validation succeeded was supposed to be called. The us constant in the description refers to the local PID.

Aprogiena: So maybe I get it wrong, let's wait for @noescape00 two confirm the correct approach.

noescape00: Ok, the logic is a bit incomplete. 

First we should present the headers (ConnectNewHeaders), after that mark them as PV (call PartialValidationSucceeded) and after that fail full validation (call PartialOrFullValidationFailed) and ensure that there is only one header claimed by LocalPeerId and it is equal to old tip before reorg was tried.

noescape00: So logically 2 things are missing in this test- marking the header as PV and checking the inner structures to ensure that LocalPeerId is where it should be.

Aprogiena:
Looks good to me. The only thing I don't like is the local definition of local PID, which should be obtained from CHT, or at least defined as class wide constant.

Aprogiena:
@noescape00 Please make the last checks and then we can go, looks good to me

PR #1569, opened 2018-06-27
Reflection extensions

noescape00:
Added 2 more reflection extensions

### DIFF
--- a/src/Stratis.Bitcoin.Tests.Common/ReflectionExtensions.cs
+++ b/src/Stratis.Bitcoin.Tests.Common/ReflectionExtensions.cs
@@ -30,6 +30,18 @@ namespace Stratis.Bitcoin.Tests.Common
             throw new Exception();
         }
 
+        /// <summary>Gets private constant member of specified type.</summary>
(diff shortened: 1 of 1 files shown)


dangershony:
@noescape00 there are actually calls to not use reflection anymore in tests, we are reconsidering this approach.

noescape00:
@dangershony ok. So in case we decide to remove reflection and use internals we also should remove reflection helpers and rewrite tests that already use reflection. 
I don't mind if we use reflection or internals- both approaches are ugly, one makes the tests uglier, another makes the production code.

@Aprogiena @fenix2222 @thefazzer

Let's discuss it here.

Aprogiena:
I am quite tired. Looks like we do something, then someone suggests that it is not the way to do, so we agree on doing something else, then we stick to it and do it like that, and then someone new comes to the code and says that how we agreed on doing things is not good and we get stuck again.

So I don't care what you choose, just choose something and make it clear that it will not change in years so that we can rely on that.

dangershony:
I will let @fenix2222 comment on this but in the mean time lets push.

PR #1571, opened 2018-06-28
Assign all Consensus settings directly in inherited networks

rowandh:
Moves the assignment of all Consensus settings into inherited networks BitcoinTest, BitcoinRegTest, StratisTest, and StratisRegTest.

Reasoning:
* All settings are assigned in one place. Much easier to reason about than having to dive into multiple layers of inheritance
* These fields will not change, so it's no problem to have duplication
* Allows us to make `Consensus` immutable in the future

Testing:
✅ NBitcoin Network tests pass
✅ Sync and validate StratisTest (100k blocks)
✅ Sync and validate BitcoinTest (100k blocks)

### DIFF
--- a/src/NBitcoin/Networks/BitcoinRegTest.cs
+++ b/src/NBitcoin/Networks/BitcoinRegTest.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using NBitcoin.DataEncoders;
 using NBitcoin.Protocol;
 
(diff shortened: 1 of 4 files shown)


dangershony:
Hmm this change I am not sure about.
It makes a test and regtest simplified knowing only what are the differences with main.

rowandh:
That is one advantage, and I made some attempt in the assignments to demonstrate which properties came from where.

We could do something like this instead:
`consensus.SubsidyHalvingInterval = StratisMain.Consensus.SubsidyHalvingInterval;`

This makes it more obvious what the differences are, but still confines creating the Consensus object to one location, rather than spreading it over an inheritance hierarchy.

The end goal with this is to make Consensus immutable, but to get there we have to separate everything first.

bokobza:
> consensus.SubsidyHalvingInterval = StratisMain.Consensus.SubsidyHalvingInterval;

This, definitely not, because then StratisMain will have to get instantiated and we actually moved away from that when I started spreading out the Networks.

I'm actually happy with this change @dangershony, it's the obvious continuation to the work that started before on Networks.

bokobza:
@rowandh not sure there is any point in keeping the inheritance `BitcoinTest : BitcoinMain`.
I guess we can just move to `Bitcpoinest : Network`.

rowandh:
@bokobza I didn't look into any other inheritance yet, wanted to keep this PR as small as possible. I can do that in another PR though.

bokobza:
@dangershony this is good, you no like it?

dangershony:
To be honest no I like it how it is now using inheritance, it almost make sense as the test/reg networks will always be a by-product of there mainnet.
But then lets talk about the reason to this change immutable consensus, @rowandh can you describe your plan for Consensus? if this changes makes sense towards that then I am fine with it.

rowandh:
`Consensus` should not change for the lifetime of the application, so by making it immutable we prevent it from being in invalid states. It also makes it harder for a dev to inadvertently reassign some consensus property and break something.

To get there, we need to remove property setters from `Consensus` and instead inject values via the constructor (or use a builder). And to get *there* we need to remove the inheritance, so that we can assign all values in the same place.

How often do we need to know which Consensus properties are inherited in the test networks? If it's often enough and will be beneficial, then we can define a static `Consensus` field on each `Network` and explicitly display inheritance by referencing other `Network.Consensus` objects when creating new ones.

dangershony:
I'd like to have more opinions here @bokobza @fassadlr @noescape00

fassadlr:
@dangershony I am in favor of this change and have never really liked the inheritance structure of the different networks. I don't think we should view test networks as extensions or continuations of main networks but rather that they are in fact completely new networks in their own right.

noescape00:
I think inheritance is a good thing here and test networks should inherit from main and override only parameters that are different for testnet. Testnet is only there to imitate mainnet and derive only where it's convenient in order to increase speed of testing so it shouldn't be treated as a standalone network. 

However I don't really have a strong opinion about this change and I'm ok if this PR goes in.

bokobza:
So I'm torn.
* On one hand I like having all the properties laid out independently because you can in one glance see what these are.

* On the other hand I like the fact that Test is inherited from Main because you can clearly see what are the differences between the two. 
Normally, Test should be exactly the same as Main, just like test servers have identical setups to production servers.
Here, the inheritance says: "same as Main, with just these few changes".

For Consensus, I think it could have independent values, not inherited, because the values in consensus vary wildly between networks. But then if networks are inherited but consensus isn't then we're in a new world of confusion.

(However, I don't want StratisMain to inherit from BitcoinMain etc.)

I guess a way to mitigate the two approaches is to have 
* all networks inheriting from Network (Main, Test etc) rather than from each other
* a Constants.cs class containing all the values
so we can have:
```
this.Name = "StratisTest";
this.Magic = 0x5223570;
this.MaxTipAge = Constants.MaxTipAge;
this.MinTxFee = Constants.MinTxFee;
this.FallbackFee = Constants.FallbackFee;
this.MinRelayTxFee = Constants.MinRelayTxFee;
```

Thoughts?

fassadlr:
> all networks inheriting from Network (Main, Test etc) rather than from each other

I would go one step further and create a `StratisNetwork` and `BitcoinNetwork` class and then create `StratisTestNetwork` and `BitcoinTestNetwork` underneath that. That way shared properties can be in the Stratis or Bitcoin base class and a user cannot accidentally create a test Stratis network under Bitcoin.

> a Constants.cs class containing all the values

👍

rowandh:
I've just pushed an example to StratisRegTest which copies inherited properties directly from the base network. For non-value Types I instantiate a new instance of the Type.

This way we demonstrate where inheritance occurs between networks but still get to assign Consensus properties directly.

If this approach looks good for StratisRegTest, I will apply it to the other networks.

dangershony:
It looks better @rowandh 
I did have some think about it I feel Network should be entirely inherited (so doing same as Network with Consensus)

Your plan is to try to introduce a builder, it feels a builder is not the right approach as this implies you can build the network differently at a later stage, what if we used a new class StratisConsensus that inherits from Consensus (Consensus become abstract) and sets items in the constructor (then StraitsConsensusTest will call base constructor and override values.

Also the plan is to move ConsensusFactory up a level form Consensus to Network, then also add a ConsensusRules container.

WTY

bokobza:
Adding another consideration to the mix:
Let's take the case where someone creates a new network and doesn't want to use the Stratis.Networks project (that we'll create and that contains bitcoin + stratis networks).
The only thing they'll have is the Network class (from the FN) + the examples in Stratis.Networks. In this case, @rowandh's first jab at it is the right one to go with.

rowandh:
To add to @bokobza's suggestion, people may also want to create networks/consensus from a config file. In this case it makes sense to be able to pass the configuration options into an object the constructor or via a builder.

bokobza:
Re. Network from a config file.
Yeah I have a document that outlines the steps required to do that: config file that we run through an app that 1. mines the genesis 2. outputs a set of files for Main, Test etc.

dangershony:
We did consider config, my reservations about using config is that creating a Network is a bit of a specialised processes, and once created one will likely not change much in the network so there is less point in using config but rather a complied class would make more sense (Jeremy did work on using .cs file templates.)

bokobza:
@rowandh Let's say we now have consensus initialised entirely in the network to which it belongs (which I like), how would you make it immutable?

rowandh:
@bokobza first make all properties on `Consensus` readonly. Then we can create `Consensus` one of two ways:

`new Consensus(propertyA, propertyB, ...)`

or

```
var builder = new ConsensusBuilder();
builder.PropertyA = ...;
builder.PropertyB = ...;

Consensus consensus = builder.Build();
```

I favour the second way because there are quite a lot of properties that would need to be set in the constructor.

bokobza:
So a builder 😭 
I have nothing against builders per se but here I think it's quite an overhead.
How about doing like in Network: an abstract Consensus class with protected properties?
This way once it initialised in the derived class in the constructor and then it's immutable and not given to change anywhere else.

Then we'd have this:
```
public class StratisMain : Network
{
    public StratisMain()
    {
	  // network properties initialised here
	  // this.Name = "StratisMain";
	  // this.Magic = magic;

      this.Consensus = new StratisMainConsensus();
	}
}

public class StratisMainConsensus : Consensus
{
    public StratisMainConsensus()
    {
        // consensus properties initialised here
        this.SubsidyHalvingInterval = 210000;
        this.MajorityEnforceBlockUpgrade = 750;
    }   
}
```

rowandh:
@bokobza sounds like a reasonable suggestion. And for the inheritance - keep it, or redefine all properties?

bokobza:
> keep it, or redefine all properties

I prefer to redefine all the properties because when Network classes will be generated using our special "create your own network" tool, then it will be easier to produce.
But I'm not fussed either way.
@dangershony may think differently.

Like you though, I'm keen to reach consensus on this one.

dangershony:
If @rowandh you are fine with Jeremy suggestion where Conesnus will be abstract then thats super I good I also like that approach so now we can advance.
My preference is to keep inheritence for test/reg networks, but not going to insist as its a minor detail now @fassadlr? you tip the scale?

rowandh:
If we're happy with this approach, I will revert the last commit, and then we can merge this in as the first step in the process. I will then create a separate PR to create inherited networks and set the properties there.

fassadlr:
@dangershony @rowandh What is the approach that we are taking now, inheritance?

dangershony:
@fassadlr yes inheritance on Network and Consensus minor details though I think @rowandh already decided on assignment.

I'll approve this in case you guys agree.

fassadlr:
@rowandh seeing that we now have SC code in our master, can you also, as part of this PR, amend the SC networks?

rowandh:
Good idea @fassadlr 👍

rowandh:
@fassadlr This has been done in #1715

PR #1572, opened 2018-06-28
Cleanup network usage in NBitcoin #2

fassadlr:
All instances where we now create `Block`, `BlockHeader` and `Transaction` has to be created by passing in the network's `ConsensusFactory` all other types does not require it.

**This can be reviewed by commit.**

- All unit and integration tests pass.
- Staking working as per normal.
- Synced STRAT testnet
- Synced BTC testnet up until block 100k

### DIFF
--- a/src/NBitcoin.Tests/pos_transaction_tests.cs
+++ b/src/NBitcoin.Tests/pos_transaction_tests.cs
@@ -170,7 +170,7 @@ namespace NBitcoin.Tests
                 ScriptSig = scriptPubKey
             });
             tx.AddOutput(new TxOut("21", key.PubKey.Hash));
-            Transaction clone = tx.Clone(network: Network.StratisMain);
+            Transaction clone = tx.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             tx.Sign(Network.StratisMain, key, false);
             AssertCorrectlySigned(tx, scriptPubKey);
             clone.Sign(Network.StratisMain, key, true);
@@ -1201,7 +1201,7 @@ namespace NBitcoin.Tests
 
         private Transaction AssertClone(Transaction before)
         {
-            Transaction after = before.Clone(network: Network.StratisMain);
+            Transaction after = before.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             Transaction after2 = null;
 
             var ms = new MemoryStream();
@@ -1652,7 +1652,7 @@ namespace NBitcoin.Tests
 
             spendTransaction.Inputs[0].ScriptSig = redeem; //The redeem should be in the scriptSig before signing
 
-            Transaction partiallySigned = spendTransaction.Clone(network: Network.StratisMain);
+            Transaction partiallySigned = spendTransaction.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             //... Now I can partially sign it using one private key:
 
             partiallySigned.Sign(Network.StratisMain, privKeys[0], true);
@@ -1668,13 +1668,13 @@ namespace NBitcoin.Tests
             AssertCorrectlySigned(gistTransaction, fundingTransaction.Outputs[0].ScriptPubKey, this.allowHighS); //One sig in the hard code tx is high
 
             //Can sign out of order
-            partiallySigned = spendTransaction.Clone(network: Network.StratisMain);
+            partiallySigned = spendTransaction.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             partiallySigned.Sign(Network.StratisMain, privKeys[2], true);
             partiallySigned.Sign(Network.StratisMain, privKeys[0], true);
             AssertCorrectlySigned(partiallySigned, fundingTransaction.Outputs[0].ScriptPubKey);
 
             //Can sign multiple inputs
-            partiallySigned = spendTransaction.Clone(network: Network.StratisMain);
+            partiallySigned = spendTransaction.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             partiallySigned.Inputs.Add(new TxIn()
             {
                 PrevOut = new OutPoint(fundingTransaction.GetHash(), 1),
@@ -1759,7 +1759,7 @@ namespace NBitcoin.Tests
         {
             Transaction tx = Network.StratisMain.Consensus.ConsensusFactory.CreateTransaction();
             tx.LockTime = new LockTime(4);
-            Transaction clone = tx.Clone(network: Network.StratisMain);
+            Transaction clone = tx.Clone(Network.StratisMain.Consensus.ConsensusFactory);
             Assert.Equal(tx.LockTime, clone.LockTime);
 
             Assert.Equal("Height : 0", new LockTime().ToString());
@@ -2287,7 +2287,7 @@ namespace NBitcoin.Tests
 
                             if (flag == SigHash.None)
                             {
-                                Transaction clone = result.Clone(network: Network.StratisMain);
+                                Transaction clone = result.Clone(Network.StratisMain.Consensus.ConsensusFactory);
                                 foreach (TxIn input in clone.Inputs)
                                 {
                                     if (input.PrevOut != signedCoin.Outpoint)
@@ -2703,7 +2703,7 @@ namespace NBitcoin.Tests
             outputm.Outputs[0].Value = Money.Satoshis(1);
             outputm.Outputs[0].ScriptPubKey = outscript;
 
-            output = outputm.Clone(network: Network.StratisMain);
+            output = outputm.Clone(Network.StratisMain.Consensus.ConsensusFactory);
 
             Assert.True(output.Inputs.Count == 1);
             Assert.True(output.Inputs[0].ToBytes().SequenceEqual(outputm.Inputs[0].ToBytes()));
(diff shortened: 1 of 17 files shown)


src/NBitcoin/Transaction.cs
```diff
@@ -1190,7 +1190,7 @@ public static Transaction Load(string hex, Network network, ProtocolVersion vers
                 throw new ArgumentNullException(nameof(network));
 
             Transaction transaction = network.Consensus.ConsensusFactory.CreateTransaction();
-            transaction.FromBytes(Encoders.Hex.DecodeData(hex), version, network);
+            transaction.FromBytes(Encoders.Hex.DecodeData(hex), version);
```
bokobza: Are you sure you don't need the network or consensusFaactory when you Load a transaction?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -95,7 +94,7 @@ private void txn_mempool_conflict_error_occurs()
         private void receiving_node_attempts_to_double_spend_mempool_doesnotaccept()
         {
             var unusedAddress = this.stratisReceiver.FullNode.WalletManager().GetUnusedAddress(new WalletAccountReference("mywallet", "account 0"));
-            var transactionCloned = this.transaction.Clone();
+            var transactionCloned = this.transaction.Clone(this.stratisReceiver.FullNode.Network.Consensus.ConsensusFactory);
```
bokobza: ah here you need a consensusFactory for a transaction it seems.

src/NBitcoin.Tests/pos_transaction_tests.cs
```diff
@@ -170,7 +170,7 @@ public void CanSignTransaction()
                 ScriptSig = scriptPubKey
             });
             tx.AddOutput(new TxOut("21", key.PubKey.Hash));
-            Transaction clone = tx.Clone(network: Network.StratisMain);
+            Transaction clone = tx.Clone(Network.StratisMain.Consensus.ConsensusFactory);
```
bokobza: For all these tests, I think you can probably declare `Network network = Network.StratisMain;` somewhere and then use that, instead of repeating `Network.StratisMain.Consensus.ConsensusFactory` everywhere.

bokobza:
There are a few places where Network was used before and now it's not (nor consensusFactory). And vice-versa.
Can you please review to make sure there is no change in functionality?

src/NBitcoin/Transaction.cs
fassadlr: @bokobza To put it another way, if we create the Transaction object outside of the Load method we need to pass in the network as we need to create it via the network's consensus factory. I.e. with the Clone method we call Load, which requires the network.

That said, I'm going to look at removing the clone method as its confusing as to which method to call to create\clone a transaction object.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
fassadlr: Yeah see comment above.

src/NBitcoin.Tests/pos_transaction_tests.cs
fassadlr: OK I'll fix these 👍

fassadlr:
@bokobza the only reason we actually have the Network class in bitcoin stream etc is to check the IsProofOfStake setting on Consensus (As well as to create the object). I feel that this needs to be looked at properly.

src/NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -160,7 +160,7 @@ public void ReadWrite(BitcoinStream stream)
                 // coinbase height
                 stream.ReadWriteAsVarInt(ref this.nHeight);
 
-                if (stream.ConsensusFactory.Consensus.IsProofOfStake)
+                if (stream.ConsensusFactory.Consensus != null && stream.ConsensusFactory.Consensus.IsProofOfStake)
```
dangershony: Ah interesting this should throw DefaultConsensusFactory should throw is a stream is trying to access `DefaultConsensusFactory.Consensus` is my opinion.

src/NBitcoin/CachedNoSqlRepository.cs
```diff
@@ -89,37 +84,23 @@ protected override Task PutBytesBatch(IEnumerable<Tuple<string, byte[]>> enumera
...
+                    using (this.@lock.LockWrite())
                     {
                         this._Table.AddOrReplace(key, raw.Data);
                     }
                 }
             }
             return result;
         }
-
-        public void Flush()
```
dangershony: Is it not used?

src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -86,43 +85,62 @@ public static int GetSerializedSize(this IBitcoinSerializable serializable, Prot
...
+            instance.FromBytes(serializable.ToBytes(version), version);
 
             return instance;
         }
 
-        public static byte[] ToBytes(this IBitcoinSerializable serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null)
+        public static T Clone<T>(this T serializable, ConsensusFactory consensusFactory, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION) where T : IBitcoinSerializable, new()
         {
-            network = network ?? Network.Main;
+            T instance = consensusFactory.TryCreateNew<T>();
```
dangershony: You don't need to create the instance this will happen inside the BitcoinStream.
`consensusFactory.TryCreateNew<T>() ` should only be called in the `BitcoinStream` unless we find a good reason not to.

src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -86,43 +85,62 @@ public static int GetSerializedSize(this IBitcoinSerializable serializable, Prot
...
 
+        public static byte[] ToBytes(this IBitcoinSerializable serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION)
+        {
             using (var ms = new MemoryStream())
             {
                 var bms = new BitcoinStream(ms, true)
                 {
                     ProtocolVersion = version,
-                    ConsensusFactory = network.Consensus.ConsensusFactory
+                    ConsensusFactory = new DefaultConsensusFactory()
```
dangershony: I suggest doing this `ConsensusFactory = new DefaultConsensusFactory()` inside `BitcoinStream` constructor

src/NBitcoin/CachedNoSqlRepository.cs
fassadlr: Yes this is not used 👍

src/NBitcoin/IBitcoinSerializable.cs
fassadlr: Actually @dangershony I'm going to remove this `Clone` method as its ony used when we clone the genesis block. We can use `Block.Load` for that now.

src/NBitcoin/IBitcoinSerializable.cs
fassadlr: OK will do!

src/NBitcoin/BitcoinCore/Coins.cs
fassadlr: OK yeah I've removed this null check 👍 . Please see latest commit

src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -86,57 +82,70 @@ public static int GetSerializedSize(this IBitcoinSerializable serializable, Prot
...
+            serializable.ReadWrite(bitcoinStream);
         }
 
-        public static T Clone<T>(this T serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION, Network network = null) where T : IBitcoinSerializable, new()
+        public static T Clone<T>(this T serializable, ProtocolVersion version = ProtocolVersion.PROTOCOL_VERSION) where T : IBitcoinSerializable, new()
         {
-            network = network ?? Network.Main;
-
-            T instance = network.Consensus.ConsensusFactory.TryCreateNew<T>();
+            T instance = new DefaultConsensusFactory().TryCreateNew<T>();
```
dangershony: ok perfect yes

bokobza:
I'll give this the once-over.

src/NBitcoin/Transaction.cs
bokobza: That's true, having more methods to do the same thing is very confusing.
Still, to clone a method, calling Load and having to serialize the object to bytes is not very intuitive.
For a next PR, how about these:
```
/// Creates a transaction by cloning the transaction passed as parameter.
public Transaction(Transaction transaction)
{}

/// Creates a transaction by cloning the transaction passed as parameter.
public Transaction(string hex, Network network)
```

So you could do `Transaction trx = new Transaction(oldTransaction);`

bokobza:
Can you please merge from master and then wait for tests? appveyor is failing for some reason.
I'll be happy to merge afterwards.

src/NBitcoin/Transaction.cs
fassadlr: @bokobza Absolutely! 👍

PR #1575, opened 2018-06-29
CHT test: Race condition test - issue 29

fenix2222:
Remove ChainHeader manually from a tree and call PartialValidationSucceeded, make sure it returns nothing

See #1321 (comment)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1142,5 +1139,30 @@ public void ConsensusAndHeadersAreAtBlock5_Block6Presented_PartialValidationSucc
...
+            ChainedHeaderTree cht = ctx.ChainedHeaderTree;
+            ChainedHeader chainTip = ctx.InitialChainTip;
+
+            // Manually remove chained header.
+            Dictionary<uint256, ChainedHeader> chainedHeaders = cht.GetChainedHeadersByHash();
+            chainedHeaders.Remove(chainTip.HashBlock);
+
+            // PartialValidationSucceeded should return null.
+            List<ChainedHeader> headersToValidate = cht.PartialValidationSucceeded(chainTip, out bool reorgRequired);
+            headersToValidate.Should().BeNull();
```
Aprogiena: I don't really agree with Ivan here. I think this test should be performed differently. I think we should have peer presents at least two headers. Those headers will be connected the tree. Then we save the first such connected block to variable X.  Then we simulate block downloaded for BOTH blocks and then we call disconnect on the peer, which will remove its chain from the tree. Then we call partial validation succeeded on X. 

@noescape00 I don't think the test otherwise makes sense.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1281,45 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            
+            // Peer 1 presents these 2 headers.
+            const int peer1Id = 1;
+            List<BlockHeader> listOfHeaders = ctx.ChainedHeaderToList(chainTip, chainExtension);
+            cht.ConnectNewHeaders(peer1Id, listOfHeaders);
+
+            // Download both blocks.
+            ChainedHeader firstPresentedHeader = chainTip.Previous;
+            cht.BlockDataDownloaded(chainTip, chainTip.Block);
+            cht.BlockDataDownloaded(chainTip.Previous, firstPresentedHeader.Block);
```
Aprogiena: Either use the variable or use .previous consistently.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1281,45 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
                 }
             }
         }
+
+        /// <summary>
+        /// Issue 29 @ Peer presents at least two headers. Those headers will be connected the tree. 
+        /// Then we save the first such connected block to variable X and simulate block downloaded for BOTH block.
```
Aprogiena: Both block`s`

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1281,45 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+        /// Disconnect of the peer is then flollowed, which removes its chain from the tree.
```
Aprogiena: `Then the peer is disconnected` instead of the first part of the sentence with typo

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1281,45 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            // Download both blocks.
+            ChainedHeader firstPresentedHeader = chainTip.Previous;
+            cht.BlockDataDownloaded(chainTip, chainTip.Block);
+            cht.BlockDataDownloaded(chainTip.Previous, firstPresentedHeader.Block);
+
+            // Disconnect peer 1.
+            cht.PeerDisconnected(peer1Id);
+
+            // Attempt to call PartialValidationSucceeded on a saved bock.
+            List<ChainedHeader> headersToValidate = cht.PartialValidationSucceeded(firstPresentedHeader, out bool reorgRequired);
```
Aprogiena: Soon to be merged the change that renames reorganisation required to full validation required, please Rename here. Also please assert that it is false.

Aprogiena:
Looks good to me, the only thing is the same as I described in the other PR for test number 30 – the comment about supplying the block data we have already in the local chain. However, it is probably not necessary, so implemented only if you consider it better yourself.

fenix2222:
@Aprogiena Yeah, me being lazy. I have updated it as it makes more sense.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1326,5 +1322,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
                 }
             }
         }
+
+        /// <summary>
+        /// Issue 29 @ Peer presents at least two headers. Those headers will be connected to the tree. 
+        /// Then we save the first such connected block to variable X and simulate block downloaded for both blocks.
+        /// Then the peer is disconnected, which removes its chain from the tree. 
+        /// Partial validation succeeded is then called on X.
+        /// </summary>
```
noescape00: Please add info on what is being tested

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1326,5 +1322,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            // Peer 1 presents these 2 headers.
+            const int peer1Id = 1;
+            List<BlockHeader> listOfHeaders = ctx.ChainedHeaderToList(chainTip, chainExtension);
+            ConnectNewHeadersResult connectionResult = cht.ConnectNewHeaders(peer1Id, listOfHeaders);
+            ChainedHeader consumedHeader = connectionResult.Consumed;
+
+            // Download both blocks.
+            ChainedHeader firstPresentedHeader = consumedHeader.Previous;
+            cht.BlockDataDownloaded(consumedHeader, chainTip.Previous.Block);
+            cht.BlockDataDownloaded(firstPresentedHeader, chainTip.Block);
```
Aprogiena: Thanks for the update, yes, it makes more sense to me as well.

However, you have swapped the blocks.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1326,5 +1322,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            
+            // Peer 1 presents these 2 headers.
+            const int peer1Id = 1;
+            List<BlockHeader> listOfHeaders = ctx.ChainedHeaderToList(chainTip, chainExtension);
+            ConnectNewHeadersResult connectionResult = cht.ConnectNewHeaders(peer1Id, listOfHeaders);
+            ChainedHeader consumedHeader = connectionResult.Consumed;
+
+            // Download both blocks.
+            ChainedHeader firstPresentedHeader = consumedHeader.Previous;
+            cht.BlockDataDownloaded(consumedHeader, chainTip.Previous.Block);
```
noescape00: consumedHeader here is 7, but you take block from chainTip.Previous which is 6. 
On the next line 
```
cht.BlockDataDownloaded(firstPresentedHeader, chainTip.Block);
```
firstPresentedHeader is 6, chainTip is 7.


It doesn't affect the test because min validation (that checks if block hash == chainedheader.Hash) is invoked on CM level, but still would be nice to make it logically correct.

fenix2222: fixed

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
fenix2222: Fixed

PR #1576, opened 2018-06-29
Gap limit configurable

majikandy:
Gap limit is now configurable via Wallet Settings

PR level lowish

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Moq;
 using NBitcoin;
+using NBitcoin.Protocol;
 using Newtonsoft.Json;
 using Stratis.Bitcoin.Configuration;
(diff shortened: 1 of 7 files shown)


noescape00:
If we have a wallet with a gap of default 20 addresses gap and on one machine we change the gap to 30 and use address number 25, then run old copy of the wallet on a node which was configured to have gap of 30 as well we will not generate new addresses on node startup and will not find the tx. 

So maybe not only take configured gap in count when creating / recovering or finding new UTXO for a wallet but also check if appropriate amount of addresses was generated on a wallet startup. If not- generate new addresses.

bokobza:
So the unusedAddressBuffer has moved to the WalletSettings, but how do you set it when running your node?
Ideally you'd have a cmd line setting.

majikandy:
Command line setting done.

Still needs @noescape00 's suggestion of taking the value into account on wallet startup and not just create

bokobza:
> Still needs @noescape00 's suggestion of taking the value into account on wallet startup and not just create.

I think we can probably leave this for another PR.
The way to do it will probably to, at the time the wallets are loaded, count how many addresses we have, compare that to the gap setting, and then create a few more addresses if necessary.
Maybe take the logic that does the buffer calculation from `TransactionFoundInternal` and create a method from it.

PR #1577, opened 2018-06-29
[Bug] Null reference in BitcoinSerializableJsonConverter

fassadlr:
Raised by PistolPete in discord

### DIFF
--- a/src/NBitcoin/IBitcoinSerializable.cs
+++ b/src/NBitcoin/IBitcoinSerializable.cs
@@ -42,17 +42,13 @@ namespace NBitcoin
             ReadWrite(serializable, new MemoryStream(bytes), false, consensusFactory, version);
         }
 
-        private static BitcoinStream CreateBitcoinStream()
+        public static int GetSerializedSize(this IBitcoinSerializable serializable, ProtocolVersion version, SerializationType serializationType)
         {
-            return new BitcoinStream(Stream.Null, true)
+            var bitcoinStream = new BitcoinStream(Stream.Null, true)
             {
                 ConsensusFactory = new DefaultConsensusFactory(),
             };
-        }
 
-        public static int GetSerializedSize(this IBitcoinSerializable serializable, ProtocolVersion version, SerializationType serializationType)
-        {
-            BitcoinStream bitcoinStream = CreateBitcoinStream();
             bitcoinStream.Type = serializationType;
             bitcoinStream.ReadWrite(serializable);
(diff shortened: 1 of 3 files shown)


src/NBitcoin/IBitcoinSerializable.cs
```diff
@@ -42,20 +42,28 @@ public static void ReadWrite(this IBitcoinSerializable serializable, byte[] byte
             ReadWrite(serializable, new MemoryStream(bytes), false, consensusFactory, version);
         }
 
+        private static BitcoinStream CreateBitcoinStream()
+        {
+            return new BitcoinStream(Stream.Null, true)
+            {
+                ConsensusFactory = new DefaultConsensusFactory(),
```
dangershony: Possibly I would set that in the constructor

bokobza: ah, just raised an issue with that problem.

bokobza:
@fassadlr is everything tested and syncing smoothly?

src/NBitcoin/IBitcoinSerializable.cs
fassadlr: @dangershony can we look at maybe doing that across the board after this fix has gone in?

fassadlr:
@bokobza I have synced stratis testnet and tested with PistolPetes wallet to ensure that the node starts up correctly. Staking also works as per usual. I'm busy syncing the other chainz.

src/NBitcoin/IBitcoinSerializable.cs
bokobza: I think he just means that yo udon't need to have a new static method to instantiate the BitcoinStream object. Just bring the `new BitcoinStream(...` inline to from where it's called.

bokobza: ah no he's saying that you should set up the `DefaultConsensusFactory` in the ctor.
What I'm saying is that you don't need an extra method :-)

bokobza:
Also look at this issue I raised yesterday, it's related.
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1582

This happens when you hav a wallet in your folder, so that's something to be tested as well.

bokobza:
>  tested with PistolPete's wallet 

Cool, missed that.
so just the CreateBitcoinStream method then.

Aprogiena:
I missing some kind of explanation of the bug and the fix. Where does DefaultConsensusFactory come from and how is that we don't need to get it from the Network?

PR #1578, opened 2018-06-29
Check BlockDataAvailabilityState is set correctly (#1321)

thefazzer:
**Check BlockDataAvailabilityState is set correctly (#1321)**

_Additional PR to check BlockDataAvailabilityState is set correctly.  In tests where we return blocks that are requested for download BlockDataAvailabilityState.BlockRequired should be set for those headers._

Attention: @noescape00  @Aprogiena


thefazzer:
@Aprogiena - added extension method & coverage

src/Stratis.Bitcoin.Tests/Consensus/ConsensusTestExtensions.cs
```diff
@@ -26,5 +26,17 @@ public static ChainedHeader GetPeerTipChainedHeaderByPeerId(this ChainedHeaderTr
...
+        {
+            if (connectNewHeadersResult.DownloadFrom == null || connectNewHeadersResult.DownloadTo == null) return false;
+            ChainedHeader chainedHeader = connectNewHeadersResult.DownloadTo;
+            while (chainedHeader.Height > connectNewHeadersResult.DownloadFrom.Height)
+            {
+                if (chainedHeader.BlockDataAvailability != blockDataAvailabilityState) return false;
+                chainedHeader = chainedHeader.Previous;
+            }
+            return true;
+        }
```
Aprogiena: Almost good here, let's just polish it with code style compliance.

That would mean brackets on line 32, empty line after line 32, empty line after line 36 and also after line 38.

Also the wild condition should be greater or equal – imagine download from equal download to.

src/Stratis.Bitcoin.Tests/Consensus/ConsensusTestExtensions.cs
```diff
@@ -40,5 +40,24 @@ public static ChainedHeader[] ToArray(this ChainedHeader chainedHeader, int head
...
+            while (chainedHeader.Height >= connectNewHeadersResult.DownloadFrom.Height)
+            {
+                if (chainedHeader.BlockDataAvailability != blockDataAvailabilityState)
+                {
+                    return false;
+                }
+                chainedHeader = chainedHeader.Previous;
+            }
+            return true;
+        }
```
Aprogiena: i was thinking more like this:

```
        public static bool HaveBlockDataAvailabilityStateOf(this ConnectNewHeadersResult connectNewHeadersResult, BlockDataAvailabilityState blockDataAvailabilityState)
        {
            if ((connectNewHeadersResult.DownloadFrom == null) || (connectNewHeadersResult.DownloadTo == null))
            {
                return false;
            }

            ChainedHeader chainedHeader = connectNewHeadersResult.DownloadTo;
            while (chainedHeader.Height >= connectNewHeadersResult.DownloadFrom.Height)
            {
                if (chainedHeader.BlockDataAvailability != blockDataAvailabilityState)
                {
                    return false;
                }

                chainedHeader = chainedHeader.Previous;
            }

            return true;
        }
```

that is in compliance with the code style. Another option would be
```
        public static bool HaveBlockDataAvailabilityStateOf(this ConnectNewHeadersResult connectNewHeadersResult, BlockDataAvailabilityState blockDataAvailabilityState)
        {
            if ((connectNewHeadersResult.DownloadFrom == null) || (connectNewHeadersResult.DownloadTo == null))
                return false;

            ChainedHeader chainedHeader = connectNewHeadersResult.DownloadTo;
            while (chainedHeader.Height >= connectNewHeadersResult.DownloadFrom.Height)
            {
                if (chainedHeader.BlockDataAvailability != blockDataAvailabilityState)
                    return false;
   
                chainedHeader = chainedHeader.Previous;
            }

            return true;
        }
```

thefazzer: All good - used the former.

Aprogiena:
Looks good now

PR #1587, opened 2018-07-02
CHT test: Partial download of the header - issue 30

fenix2222:
Chain is 10 long. All are Data Available. Call PartialValidationSucceeded on 1st header. 1st header is marked PV. Make sure rest of them are returned for partial validation.

See #1321 (comment)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1284,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
                 }
             }
         }
+
+        /// <summary>
+        /// Issue 30 @ Chain is 10 headers long. All headers are "Data Available". Call PartialValidationSucceeded on
+        /// the first header. First header is marked Partially Validated (PV). Make sure the rest of them are returned for
```
Aprogiena: As we mentioned and you implemented below, all the do next header, the second one, should be returned for partial validation.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1284,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            ConnectNewHeadersResult connectionResult = cht.ConnectNewHeaders(1, listOfExtendedHeaders);
+            chainTip = connectionResult.Consumed;
+
+            // Download all header blocks and call PartialValidationSucceeded on h3.
+            ChainedHeader[] headerArray = chainTip.ToArray(extensionSize);
+            ChainedHeader firstHeader = headerArray[0];
+            ChainedHeader secondHeader = headerArray[1];
+            foreach (ChainedHeader chainedHeader in headerArray)
+            {
+                cht.BlockDataDownloaded(chainedHeader, ctx.CreateBlock());
```
Aprogiena: Here we could use the corresponding block data from the chain we have outside of the tree. It is not necessary, but the extension the routine is actually creating the block data, so it makes sense to supply blocks from the local chain to the tree. It doesn't really change anything for the test, so I'm not sure if it is better or not, it just makes better sense.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1284,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            }
+
+            List<ChainedHeader> listOfHeaders = cht.PartialValidationSucceeded(firstHeader, out bool reorgRequired);
```
Aprogiena: reorgRequired renamed to fullValidationRequired, please use that from now on

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1284,5 +1284,47 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+
+            // FIrst header validation state should be "PartiallyValidated" and next header returned.
```
Aprogiena: F`i`rst

Aprogiena:
Looks good to me

PR #1590, opened 2018-07-02
Wallet to use address buffer on startup

majikandy:
low/medium maybe?

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
@@ -1111,9 +1111,9 @@ namespace Stratis.Bitcoin.Features.Wallet
         private IEnumerable<HdAddress> AddAddressesToMaintainBuffer(HdAccount account, bool isChange)
         {
             HdAddress lastUsedAddress = account.GetLastUsedAddress(isChange);
-            int lastUsedAddressIndex = lastUsedAddress?.Index ?? 0;
+            int lastUsedAddressIndex = lastUsedAddress?.Index ?? -1;
             int addressesCount = isChange ? account.InternalAddresses.Count() : account.ExternalAddresses.Count();
-            int emptyAddressesCount = addressesCount - lastUsedAddressIndex; // should this be -1 ??
+            int emptyAddressesCount = addressesCount - lastUsedAddressIndex - 1;
             int addressesToAdd = this.walletSettings.UnusedAddressesBuffer - emptyAddressesCount;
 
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1106,6 +1108,16 @@ private void TransactionFoundInternal(Script script)
             this.logger.LogTrace("()");
         }
 
+        private IEnumerable<HdAddress> AddAddressesToMaintainBuffer(HdAccount account, bool isChange)
+        {
+            HdAddress lastUsedAddress = account.GetLastUsedAddress(isChange);
+            int lastUsedAddressIndex = lastUsedAddress?.Index ?? 0;
+            int addressesCount = isChange ? account.InternalAddresses.Count() : account.ExternalAddresses.Count();
+            int emptyAddressesCount = addressesCount - lastUsedAddressIndex; // should this be -1 ??
```
bokobza: So the Index is zero-based and it's the index from the address' HD path.
If you have a wallet with 20 addresses and a transaction is found for the 7th address (Index = 6), then there should be 13 addresses created, which is 20 - 6 - 1.

majikandy: Wouldn't this 
so, when this is called from the Start method, with a fresh wallet, lastUsedAddress returns null - so I set this  to 0 in that case as you can see on line 1114 - 

Either I set lastUsedAddress to -1 which makes the maths above work.

I think I favour the -1

<pre>
  HdAddress lastUsedAddress = account.GetLastUsedAddress(isChange);
  int lastUsedAddressIndex = lastUsedAddress?.Index ?? <b>-1</b>;
  int addressesCount = isChange ? account.InternalAddresses.Count() : account.ExternalAddresses.Count();
  int emptyAddressesCount = addressesCount - lastUsedAddressIndex <b>- 1</b>; 
</pre>


Or I get it more explicit with something like this

```
int addressesToAdd = 0;

            if (lastUsedAddress != null)
            {
                int addressesCount = isChange ? account.InternalAddresses.Count() : account.ExternalAddresses.Count();
                int emptyAddressesCount = addressesCount - lastUsedAddress.Index - 1;
                addressesToAdd = this.walletSettings.UnusedAddressesBuffer - emptyAddressesCount;
            }
            else
            {
                addressesToAdd = this.walletSettings.UnusedAddressesBuffer;
            }
```

bokobza: 👍 👍

PR #1591, opened 2018-07-02
CHT test: Connecting out of order header - issue 36

fenix2222:
The list of headers is presented where the 1st half of them can be connected but then there is header which is not consecutive – its previous hash is not hash of the previous header in the list.
The 1st nonconsecutive header should be header that we saw before, so that it actually connects but it is out of order.

See #1321 (comment)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1326,5 +1326,45 @@ public void ChainWithMaxReorgPlusExtraHeadersIsCalled_AnotherChainIsPresented_Co
...
+            };
+
+            // Exception is thrown and no new headers are connected.
+            ChainedHeader[] allHeaders = chainTip.ToArray(initialChainSize + extensionChainSize + 1);
+            connectHeadersAction.Should().Throw<ArgumentException>();
+            Dictionary<uint256, ChainedHeader> currentHeaders = cht.GetChainedHeadersByHash();
+            currentHeaders.Should().HaveCount(initialChainSize + 1); // initial chain size + genesis.
+            currentHeaders.Should().ContainKey(allHeaders[0].HashBlock);
+            currentHeaders.Should().ContainKey(allHeaders[1].HashBlock);
+            currentHeaders.Should().ContainKey(allHeaders[2].HashBlock);
```
Aprogiena: Looks good, I would just part check that H2.next is empty

fenix2222: Added

fenix2222: Done

Aprogiena:
Looks good to me

PR #1594, opened 2018-07-03
CHT test: Present chains with headers prior to checkpoint - issue 39

fenix2222:
Single checkpoint is at 1000. Max reorg is 10. We receive a chain of 100 headers from peer1. Nothing is marked for download. Peer2 presents a chain which forks at 50 and goes to 150. Nothing is marked for download but the chain is accepted.

See #1321 (comment)


Aprogiena:
There is a big problem with this test, need to discuss it first

noescape00:
Finalized height setup is missing here (we didn't update the tests after changing max reorg to finalized height), so the correct setup would be: Initialize CT with 20. Max reorg 10. Setup finalized height to 10. 
Everything else should be the same.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1481,6 +1481,70 @@ public void ListWithOnlyHalfConnectableHeadersPresented_TheFirstNonconsecutiveHe
...
+            // Peer 1 presents a chain A.
+            // Chain accepted but nothing marked for download.
+            const int peer1Id = 1;
+            ConnectNewHeadersResult connectionResult = cht.ConnectNewHeaders(peer1Id, listOfChainAHeaders);
+            connectionResult.DownloadFrom.Should().BeNull();
+            connectionResult.DownloadTo.Should().BeNull();
+            connectionResult.Consumed.HashBlock.Should().Be(chainATip.HashBlock);
+
+            // Setup chain BB that has 150 headers and is based on the previous 30 header extension, i.e. fork point at 50.
+            // Example: h1=h2=..=h50=b51=b52=..=b170.
```
Aprogiena: The last year should be b150.

Line above just single B.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1481,6 +1481,70 @@ public void ListWithOnlyHalfConnectableHeadersPresented_TheFirstNonconsecutiveHe
...
+            ChainedHeader chainATip = ctx.ExtendAChain(chainAExtensionSize, chainTip);
+            List<BlockHeader> listOfChainAHeaders =
+                ctx.ChainedHeaderToList(chainATip, extensionSize + chainAExtensionSize);
+
+            // Peer 1 presents a chain A.
+            // Chain accepted but nothing marked for download.
+            const int peer1Id = 1;
+            ConnectNewHeadersResult connectionResult = cht.ConnectNewHeaders(peer1Id, listOfChainAHeaders);
+            connectionResult.DownloadFrom.Should().BeNull();
+            connectionResult.DownloadTo.Should().BeNull();
```
Aprogiena: As mentioned in different test, if you test download from and download to, please also include the check of block availability using the extension method we have for that.

Same also couple of lines below.

Aprogiena:
Almost good, just mind the polishing needed – assertion of block data availability state.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
fenix2222: Done

Aprogiena:
Good now.

Aprogiena:
oh assert missing

Aprogiena:
Now good.

PR #1598, opened 2018-07-03
Fix Build

monsieurleberre:


bokobza:
This problem:
`Consensus\ChainedHeaderTreeTest.cs(1365,53): error CS0029: Cannot implicitly convert type 'System.Collections.Generic.List<Stratis.Bitcoin.Primitives.ChainedHeaderBlock>' to 'System.Collections.Generic.List<NBitcoin.ChainedHeader>' `

monsieurleberre:
something else fixed it, I am still confused, but now it seems to work for everybody

PR #1599, opened 2018-07-03
Clean up Transaction usage in NBitcoin

fassadlr:
We have far too many ways to create the transaction object, whereas it should only be done via the network's consensus factory, This PR cleans up the creation of the Transaction class in all areas.


bokobza:
This: `this.network.Consensus.ConsensusFactory.CreateTransaction();` will be impossible to find for someone trying to use our code.

Creating a transaction should be done using the Transaction ctor, even if you need to pass the ConsensusFactory to it as a parameter.
Only my point of view of course.

fassadlr:
@bokobza The thing is that all `Transaction`, `Block` and `BlockHeader` objects MUST be created via the consensus factory as eventually the design will be that the network will be responsible for providing those objects to whatever requests it. We shouldn't know what `Transaction` type (PosTransaction, PowTransaction etc) we are creating, that is  consensus factory's responsibility.

I agree its not yet optimal but its a big step in the right direction. At least now all instances of Transaction is only created in one place and in one way. The next step will be to make `Transaction` abstract so that calls to its ctor will not allowed. We can look at moving ConsensusFactory to Network so that you can go:

```Transaction tx = this.network.ConsensusFactory.CreateTransaction();```

we can even do:

```Transaction tx = this.network.CreateTransaction();```

via an internal call to the consensus factory's create transaction method. But this is just semantics.

PS Its really not that difficult for the dev to go Transaction and see that its marked obsolete ;) and when we make its constructor protected you wont be able to call it anyway.

**Comment by bokobza** [comment]
Even if they must be constructed via the consensusFactory, they're still objects in their own right and it shouldn't be the consensusFactory's job to instantiate these objects.
when you create an object, you typically do either `Type objectName = new Type...` or there is some kind of static method *on the Type itself* that helps you do it. Not a fan of the second way but it's sometimes necessary.
If you *can* instantiate an object using `new`, then there is no reason to do otherwise.

Also, what's wrong with `Transaction trx = new Transaction(network);`?

**Comment by fassadlr** [comment]
@bokobza I think this is rather a discussion as to where we want to end up. If we are going to architect the node in such a away that Network will be responsible for creating its consensus objects then this is the way to go. 

Imagine the scenario where another company wants to create their own network. They will install the `Stratis.Bitcoin.Networks` (to be created) and `Stratis.Bitcoin` Nuget packages and because `Transaction` will be abstract by then, they will need to implement their own version of `ConsensusFactory` which will return their own implementation of `Transaction` for example `ApexTransaction'.

So from a mining point of view, when we create blocks, `PowMining` (for example) will then ask Network to create a block and because we have injected the `MiningFeature` with our own network and consensus factory implementations, the miner will just call `network.CreateTransaction()` and receive `ApexTransaction` .

A second scenario to consider is `PosTransaction`. We can't just go `new PosTransaction()` as the consensus factory needs to return us a `PosTransaction` object. Thats the whole point of network being responsible to create its own set of consensus objects.

This implementation ensures modularity further down the line and doesn't tie the node into a specific implementation of consensus objects.

**Comment by fassadlr** [comment]
@bokobza I have now implemented `this.network.CreateTransaction()`. You can see the change in network. Also, I have made `CreateTransaction()` on `ConsensusFactory` internal, you have to call it via Network now 👍

File: src/NBitcoin/Network.cs [hunk 1]
```diff
@@ -239,16 +239,10 @@ public byte[] MagicBytes
         /// </code>
         /// </example>
         /// <returns>A genesis block.</returns>
-        public static Block MineGenesisBlock(ConsensusFactory consensusFactory, string coinbaseText, Target target, Money genesisReward, int version = 1)
+        public static Block MineGenesisBlock(Network network, string coinbaseText, Target target, Money genesisReward, int version = 1)
```
bokobza: Does this method really need to change?

src/NBitcoin/Networks/BitcoinMain.cs
```diff
@@ -132,12 +132,12 @@ public BitcoinMain()
         /// <summary> The default name used for the Bitcoin configuration file. </summary>
         public const string BitcoinDefaultConfigFilename = "bitcoin.conf";
 
-        protected static Block CreateBitcoinGenesisBlock(ConsensusFactory consensusFactory, uint nTime, uint nNonce, uint nBits, int nVersion, Money genesisReward)
+        protected static Block CreateBitcoinGenesisBlock(Network network, uint nTime, uint nNonce, uint nBits, int nVersion, Money genesisReward)
```
bokobza: Same comment as above.

src/NBitcoin/Network.cs
```diff
@@ -239,16 +239,10 @@ public byte[] MagicBytes
...
         {
-            if (target == null)
-                throw new ArgumentException($"Parameter '{nameof(target)}' cannot be null. Example use: new Target(new uint256(\"0000ffff00000000000000000000000000000000000000000000000000000000\"))");
-
-            if (consensusFactory == null)
-            {
-                throw new ArgumentException($"Parameter '{nameof(consensusFactory)}' cannot be null. Use 'new ConsensusFactory()' for Bitcoin-like proof-of-work blockchains" +
-                                            "and 'new PosConsensusFactory()' for Stratis-like proof-of-stake blockchains.");
-            }
+            if (network == null)
```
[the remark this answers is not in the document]
bokobza: In fact this methos needs to be changed back because when it is used, we don't have a Network yet.
All the user can pass is `new PosConsensusFactory()` or similar.

src/NBitcoin/QBitNinjaTransactionRepository.cs
```diff
@@ -6,61 +6,50 @@ namespace NBitcoin
 {
     public class QBitNinjaTransactionRepository : ITransactionRepository
     {
-        private readonly Uri _BaseUri;
-        public Uri BaseUri
```
bokobza: Are you sure the public `get` for BaseUri is not used anywhere? (might not be, just asking)

src/Stratis.Bitcoin.Features.WatchOnlyWallet.Tests/WatchOnlyWalletManagerTest.cs
```diff
@@ -181,10 +188,10 @@ public void Given_AWatchedAddress_ThenCanCalculateComplexRelativeBalance()
                 "0100000001dd19f894733ab950ae8e772ef781fe299f80f4e813ee4cb6ac067fedea3a052c000000006b483045022100d4d6570c054bcbafa8178e9074f2be32b078e0761537bfe4fb7ad1eea949ab66022022e31d6f43d9418b6dc02147aefb380a4730386364bd406365e790826400570b012102a31bf228b4508abe3f0aa5f91c53732dfd49233c2fef3c5cfa68f2aa12cc71b4ffffffff029e90bf00000000001976a914128240d302a4aadcdd08d241b54fa4ef11acb21388ace0c810000000000017a914188d767b139ef64ce4efa091c2957c6137fbe0ce8700000000"
             };
 
-            var block = new Block();
+            var block = this.network.Consensus.ConsensusFactory.CreateBlock();
```
bokobza: Any plans to move this like with the CreateTransaction method?

src/NBitcoin/Network.cs
fassadlr: It actually appears that we only use this method in tests? Can I just maybe delete it? If not I'll revert some of the changes 👍

src/NBitcoin/Networks/BitcoinMain.cs
fassadlr: Done 👍

src/NBitcoin/QBitNinjaTransactionRepository.cs
fassadlr: Hmmm you are right, it be used in other projects, I'll make this public again, just in case.

src/Stratis.Bitcoin.Features.WatchOnlyWallet.Tests/WatchOnlyWalletManagerTest.cs
fassadlr: For sure! We can do after this PR ;)

src/NBitcoin/Network.cs
bokobza: no it's used for people to generate an actual Genesis block!
so it'll be used by federated sidechains, or anyone wanting to create a genesis block in order to bootstrap a network.

src/NBitcoin/Network.cs
```diff
@@ -282,9 +276,9 @@ public static Block MineGenesisBlock(ConsensusFactory consensusFactory, string c
                 Value = genesisReward,
             });
 
-            Block genesis = consensusFactory.CreateBlock();
+            Block genesis = network.Consensus.ConsensusFactory.CreateBlock();
             genesis.Header.BlockTime = time;
-            genesis.Header.Bits = target;
+            genesis.Header.Bits = target ?? throw new ArgumentException($"Parameter '{nameof(target)}' cannot be null. Example use: new Target(new uint256(\"0000ffff00000000000000000000000000000000000000000000000000000000\"))");
```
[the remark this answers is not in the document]
fassadlr: Done 👍

src/NBitcoin/Network.cs
```diff
@@ -202,7 +202,7 @@ public byte[] MagicBytes
         /// Mines a new genesis block, to use with a new network.
         /// Typically, 3 such genesis blocks need to be created when bootstrapping a new coin: for Main, Test and Reg networks.
         /// </summary>
-        /// <param name="consensusFactory">
+        /// <param name="network">
```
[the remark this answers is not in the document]
fassadlr: This has been reverted 👍

src/NBitcoin/Network.cs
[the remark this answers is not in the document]
fassadlr: This has been reverted 👍

src/NBitcoin/Network.cs
fassadlr: OK will change back

src/NBitcoin/Networks/StratisMain.cs
```diff
@@ -141,17 +141,17 @@ public StratisMain()
             this.GenesisVersion = 1;
             this.GenesisReward = Money.Zero;
 
-            this.Genesis = CreateStratisGenesisBlock(this.Consensus.ConsensusFactory, this.GenesisTime, this.GenesisNonce, this.GenesisBits, this.GenesisVersion, this.GenesisReward);
+            this.Genesis = CreateStratisGenesisBlock(this, this.GenesisTime, this.GenesisNonce, this.GenesisBits, this.GenesisVersion, this.GenesisReward);
```
bokobza: One last thing, can you also revert these back please?

fassadlr: Done 👍

src/NBitcoin/ConsensusFactory.cs
```diff
@@ -166,10 +167,30 @@ public virtual BlockHeader CreateBlockHeader()
...
-        public virtual Transaction CreateTransaction()
+        internal virtual Transaction CreateTransaction()
         {
             return new Transaction();
         }
+
+        /// <summary>
+        /// Create a <see cref="Transaction"/> instance from a hex string representation.
+        /// </summary>
+        internal virtual Transaction CreateTransaction(string hex)
```
dangershony: This must be public especially to be used by other networks.

PR #1603, opened 2018-07-04
CHT finalized height bugfix

noescape00:
Check if fork-point is below finalized block, not the same.


noescape00:
@Aprogiena 2nd approve required in order to merge

PR #1608, opened 2018-07-04
CHT Issue 19 & 6 Common test cases

kogot:


Aprogiena:
When something is related to the CHT, please always highlight me and @noescape00 inside the PR description, so that we receive email about it.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,191 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+        /// Basic test for tests 18 to 20, 28.
+        /// 1/6 There is no branch which is claimed by no one (PeerTipsByHash have tip of every branch), none of list should be empty.
+        /// 2/6 ChainHeadersByHash contains only reachable headers.
+        /// 3/6 ChainHeader claims LocalPeer
+        /// 4/6 Each connected peer has exactly  1 entry in PeertipsByHash
+        /// 5/6 CH.Next[i].Prev == CH for every header and every .Next
+        /// 6/6 PeerTipsByPeerId should reflect PeerIdsByHash(except local marker).
+        /// </summary>
+        /// <param name="cht"></param>
+        private void CheckForbasicCases(ChainedHeaderTree cht, IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders)
```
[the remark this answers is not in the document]
Aprogiena: -> CheckChainedHeaderTreeConsistency

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,191 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+        ///  Initial setup for tests 18-20, 28
+        /// Chain header tree setup. Initial chain has 4 headers.
+        ///  SetUp:
+        ///                        =8d=9d=10d
+        ///                   6a=7a=8a=9a
+        /// GENESIS=1=2=3=4=5=
+        ///                   6b=7b=8b=9b
+        ///             3c=4c=5c
+        ///  </summary>
+        ///  <param name="initialChainSize"></param>
```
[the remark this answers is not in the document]
kogot: tried to make it look like a branches. So it's would be easier to read.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,191 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            Dictionary<uint256, ChainedHeader> chainedHeadersWithPeerK = cht.GetChainedHeadersByHash();
+
+            cht.PeerDisconnected(5);
+
+            Dictionary<uint256, ChainedHeader> chainedHeadersWithoutPeerK = cht.GetChainedHeadersByHash();
+
+            chainedHeadersWithoutPeerK.Should().BeEquivalentTo(chainedHeadersWithPeerK);
+
+        }
+
```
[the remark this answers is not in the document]
kogot: ok.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,70 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+            {
+                ChainedHeader chainedHeaderFrom = connectionPeerResult.DownloadFrom;
+                ChainedHeader currentConsumedCh = chainedHeaderFrom;
+                while (currentConsumedCh != null)
+                {
+                    cht.BlockDataDownloaded(currentConsumedCh, chainTip.GetAncestor(currentConsumedCh.Height).Block);
+                    cht.PartialValidationSucceeded(currentConsumedCh, out bool fullValidationRequired);
+                    cht.ConsensusTipChanged(currentConsumedCh);
+
+                    currentConsumedCh = currentConsumedCh.Next.Any() ? currentConsumedCh.Next[0] : null;
```
Aprogiena: It is not clear what this method is supposed to do, but using the next pointer like this is dangerous, what if there multiple pointers? Why would you go with the first one?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1966,6 +2030,122 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+        {
+            const int initialChainSize = 5;
+            TestContext ctx = new TestContextBuilder().WithInitialChain(initialChainSize).UseCheckpoints().Build();
+            ChainedHeaderTree cht = ctx.ChainedHeaderTree;
+            ChainedHeader initialChainTip = ctx.InitialChainTip;
+
+            IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders = ctx.SetupPeersForTest(initialChainSize, ctx, cht, initialChainTip, out ChainedHeader chainATip, out ChainedHeader chainDTip);
+            CheckChainedHeaderTreeConsistency(cht, listOfAllUniqueBlockHeaders);
+
+            //Additional SetUp for current test
```
Aprogiena: Use space after slashes, everywhere. Use "." At the end of the sentence, everywhere.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1966,6 +2030,122 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            ConnectNewHeadersResult connectionPeerKResult = cht.ConnectNewHeaders(5, peerKBlockHeaders);
+            ctx.ClaimPeerChain(cht, chainKTip, connectionPeerKResult);
+
+            Dictionary<uint256, ChainedHeader> chainedHeadersWithPeerK = cht.GetChainedHeadersByHash();
+
+            cht.PeerDisconnected(5);
+
+            Dictionary<uint256, ChainedHeader> chainedHeadersWithoutPeerK = cht.GetChainedHeadersByHash();
+
+            chainedHeadersWithoutPeerK.Should().BeEquivalentTo(chainedHeadersWithPeerK);
```
Aprogiena: Is this not just comparing the references?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
kogot: ok

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
kogot: no, it's not just reference.
From description for this method:
 Summary:
     Asserts that two dictionaries are equivalent.
// Remarks:
        //     The values within the dictionaries are equivalent when both object graphs have
        //     equally named properties with the same value, irrespective of the type of those
        //     objects. Two properties are also equal if one type can be converted to another
        //     and the result is equal. The type of the values in the dictionaries are ignored
        //     as long as both dictionaries contain the same keys and the values for each key
        //     are structurally equivalent. Notice that actual behavior is determined by the
        //     global defaults managed by the FluentAssertions.AssertionOptions class.

Aprogiena: Okay, thank you.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,69 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+            ///             3c=4c=5c
+            /// </summary>
+            /// <param name="initialChainSize"></param>
+            /// <param name="ctx">TestContext</param>
+            /// <param name="cht">ChainHeaderTree</param>
+            /// <param name="initialChainTip">Initial chain tip</param>
+            /// <param name="chainATip">out Chaind A tip</param>
+            /// <param name="chainDTip">out Chaind D tip</param>
+            /// <returns>List of BlockHeaders</returns>
+            internal IEnumerable<BlockHeader> SetupPeersForTest(int initialChainSize, TestContext ctx, ChainedHeaderTree cht, ChainedHeader initialChainTip, out ChainedHeader chainATip, out ChainedHeader chainDTip)
```
noescape00: No need to pass TestContext here because we are inside the TextContext already.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,69 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+                List<BlockHeader> peerDBlockHeaders = ctx.ChainedHeaderToList(chainDTip, chainDTip.Height);
+
+                List<BlockHeader> listOfAllBlockHeaders = new List<BlockHeader>(initialchainHeaders);
+                listOfAllBlockHeaders.AddRange(peerABlockHeaders);
+                listOfAllBlockHeaders.AddRange(peerBBlockHeaders);
+                listOfAllBlockHeaders.AddRange(peerCBlockHeaders);
+                listOfAllBlockHeaders.AddRange(peerDBlockHeaders);
+                IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders = listOfAllBlockHeaders.Distinct();
+
+                ConnectNewHeadersResult connectionPeerAResult = cht.ConnectNewHeaders(1, peerABlockHeaders);
```
noescape00: `ConnectNewHeadersResult connectionPeerAResult = ` is redundant here and on the following lines

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,69 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+                ChainedHeader chainBTip = ctx.ExtendAChain(peerBExtension, initialChainTip); // i.e. (h1=h2=h3=h4=h5)=6b=7b=8b=9b
+                ChainedHeader chainCTip = ctx.ExtendAChain(peerCExtension, initialChainTip.GetAncestor(2)); // i.e. (h1=h2)=3c=4c=5c
+                chainDTip = ctx.ExtendAChain(peerDExtension, chainATip.GetAncestor(7)); // i.e. ((h1=h2=h3=h4=h5)=6a=7a)=8d=9d=10d
+
+                List<BlockHeader> peerABlockHeaders = ctx.ChainedHeaderToList(chainATip, chainATip.Height);
+                List<BlockHeader> peerBBlockHeaders = ctx.ChainedHeaderToList(chainBTip, chainBTip.Height);
+                List<BlockHeader> peerCBlockHeaders = ctx.ChainedHeaderToList(chainCTip, chainCTip.Height);
+                List<BlockHeader> peerDBlockHeaders = ctx.ChainedHeaderToList(chainDTip, chainDTip.Height);
+
+                List<BlockHeader> listOfAllBlockHeaders = new List<BlockHeader>(initialchainHeaders);
```
noescape00: var

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: First peer ID is 0, not one.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1994,6 +2057,123 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+        [Fact]
+        public void NewPeerClaimsHeadSecondTime_PeerDisconnected_ChainShouldNotChange()
+        {
+            const int initialChainSize = 5;
+            TestContext ctx = new TestContextBuilder().WithInitialChain(initialChainSize).UseCheckpoints().Build();
+            ChainedHeaderTree cht = ctx.ChainedHeaderTree;
+            ChainedHeader initialChainTip = ctx.InitialChainTip;
+
+            IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders = ctx.SetupPeersForTest(initialChainSize, ctx, cht, initialChainTip, out ChainedHeader chainATip, out ChainedHeader chainDTip);
+            CheckChainedHeaderTreeConsistency(cht, listOfAllUniqueBlockHeaders);
```
noescape00: This should be called after the test scenario, not after initialization. 

Same will be for other tests that use same setup- we want to make sure that after the logic executed in the test inner structures are in sync.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1994,6 +2057,123 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+
+            // Additional SetUp for current test.
+            ChainedHeader chainKTip = ctx.ExtendAChain(0,chainDTip); //peer K has exactly the same chain as peer D.
```
noescape00: Why not just use the same header? `ChainedHeader chainKTip = chainDTip;`

ExtendAChain with first argument 0 will return you the input parameter which is chainDTip in current case. So there is no need to add this complication.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,69 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+
+                ConnectNewHeadersResult connectionPeerAResult = cht.ConnectNewHeaders(1, peerABlockHeaders);
+                ConnectNewHeadersResult connectionPeerBResult = cht.ConnectNewHeaders(2, peerBBlockHeaders);
+                ConnectNewHeadersResult connectionPeerCResult = cht.ConnectNewHeaders(3, peerCBlockHeaders);
+                ConnectNewHeadersResult connectionPeerDResult = cht.ConnectNewHeaders(4, peerDBlockHeaders);
+
+                return listOfAllUniqueBlockHeaders;
+            }
+
+            internal void ClaimPeerChain(ChainedHeaderTree cht, ChainedHeader chainTip, ChainedHeader consumedHeader, int extensionSize)
```
noescape00: It doesn't claim the chain, instead it switches current consensus tip to the tip of the chain block-by-block. 
So maybe rename to `SwitchToChain`

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
noescape00: The method itself is not comparing the references. However `chainedHeadersWithoutPeerK` is the same dictionary as `chainedHeadersWithPeerK` so the logic is not correct. 
To avoid comparing this guy against itself use 
`var chainedHeadersWithPeerK = new Dictionary<uint256, ChainedHeader>(cht.GetChainedHeadersByHash());`

instead of 
`Dictionary<uint256, ChainedHeader> chainedHeadersWithPeerK = cht.GetChainedHeadersByHash();`

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -2100,32 +2100,26 @@ public void NewPeerClaimsHeadSecondTime_PeerDisconnected_ChainShouldNotChange()
...
-                {3, 0},
-                {4, 0}
-            };
+            Dictionary<int, int> peerEntryDictionary = new Dictionary<int, int>();
+            Dictionary<int, int> peerHDictionary = new Dictionary<int, int>();
 
             Dictionary<uint256, HashSet<int>> tipsDictionary = cht.GetPeerIdsByTipHash();
             Dictionary<int, uint256> peerIdsDictionary = cht.GetPeerTipsByPeerId();
 
+            peerHDictionary =new Dictionary<int, int>();
```
Aprogiena: Spacing here and below.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -214,29 +214,28 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
             ///             3c=4c=5c
             /// </summary>
             /// <param name="initialChainSize"></param>
-            /// <param name="ctx">TestContext</param>
             /// <param name="cht">ChainHeaderTree</param>
             /// <param name="initialChainTip">Initial chain tip</param>
             /// <param name="chainATip">out Chaind A tip</param>
```
noescape00: typos & dot's are missing

.gitignore
```diff
@@ -252,3 +252,4 @@ ModelManifest.xml
 /src/Stratis.Bitcoin.Tests/Stratis.Bitcoin.Tests/TestData
 /src/Stratis.StratisD/Logs
 /src/Stratis.BitcoinD/Logs
+src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs.orig
```
noescape00: Please remove and check this out:

https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,68 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+            ///                        =8d=9d=10d
+            ///                   6a=7a=8a=9a
+            /// GENESIS=1=2=3=4=5=
+            ///                   6b=7b=8b=9b
+            ///             3c=4c=5c
+            /// </summary>
+            /// <param name="initialChainSize"></param>
+            /// <param name="cht">ChainHeaderTree</param>
+            /// <param name="initialChainTip">Initial chain tip</param>
+            /// <param name="chainATip">out Chaind A tip</param>
```
noescape00: typos and missing dots

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,118 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+
+            // Additional SetUp for current test.
+            ChainedHeader chainKTip = chainDTip; //peer K has exactly the same chain as peer D.
+            List<BlockHeader> peerKBlockHeaders = ctx.ChainedHeaderToList(chainKTip, chainKTip.Height);
+
+            // Claiming Peer K chain.
+            ConnectNewHeadersResult connectionPeerKResult = cht.ConnectNewHeaders(4, peerKBlockHeaders);
+            ctx.SwitchToChain(cht, chainKTip, connectionPeerKResult.Consumed, 5);
+
+            Dictionary<uint256, ChainedHeader> chainedHeadersWithPeerK = new Dictionary<uint256, ChainedHeader>(cht.GetChainedHeadersByHash());
```
noescape00: var

noescape00:
Looks good to me. Just a few minor fixes left.

.gitignore
kogot: oopps sorry...missed that one

noescape00:
@Aprogiena merge if ure fine with it

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+
+            IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders = ctx.SetupPeersForTest(initialChainSize, cht, initialChainTip, out ChainedHeader chainATip, out ChainedHeader chainDTip);
+
+            // Additional SetUp for current test.
+            ChainedHeader chainKTip = chainDTip; //peer K has exactly the same chain as peer D.
+            List<BlockHeader> peerKBlockHeaders = ctx.ChainedHeaderToList(chainKTip, chainKTip.Height);
+
+            // Claiming Peer K chain.
+            ConnectNewHeadersResult connectionPeerKResult = cht.ConnectNewHeaders(4, peerKBlockHeaders);
+            ctx.SwitchToChain(cht, chainKTip, connectionPeerKResult.Consumed, 5);
```
Aprogiena: This shouldn't be here. The test scenario only need the new peer to present the chain but not to advance the consensus to its tip. So this line should be deleted.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+        /// Basic test for tests 18 to 20, 28.
+        /// 1/6 There is no branch which is claimed by no one (PeerTipsByHash have tip of every branch), none of list should be empty.
+        /// 2/6 ChainHeadersByHash contains only reachable headers.
+        /// 3/6 ChainHeader claims LocalPeer.
+        /// 4/6 Each connected peer has exactly 1 entry in PeertipsByHash.
+        /// 5/6 CH.Next[i].Prev == CH for every header and every .Next.
+        /// 6/6 PeerTipsByPeerId should reflect PeerIdsByHash(except local marker).
+        /// </summary>
+        /// <param name="cht">ChainHeaderTree.</param>
+        /// <param name="listOfAllUniqueBlockHeaders">List of All unique blockheader in all connected chains.</param>
```
Aprogiena: I don't think this should be presented as an argument. It is easy to get this list from the tree itself.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+            bool eachPeerOneEntry = true;
+            bool reflecttipHash = true;
+            bool chbhHasOnlyReachebleHeaders = true;
+            bool isCorrectChainSequence = true;
+            bool hasLocalPeer = false;
+
+            var peerEntryDictionary = new Dictionary<int, int>();
+            var peerHDictionary = new Dictionary<int, int>();
+
+            Dictionary<uint256, HashSet<int>> tipsDictionary = cht.GetPeerIdsByTipHash();
```
Aprogiena: It would be nice to have in-line comments here so that it is clear which code is supposed to check which requirement. Or you can create a method for each requirement.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+            }
+
+            foreach (KeyValuePair<uint256, HashSet<int>> tips in tipsDictionary)
+            {
+                if (tips.Value.Count == 0) noUnclaimedBraches = false;
+
+                foreach (int peerId in tips.Value)
+                {
+                    // Ignore local.
+                    if (peerId != -1)
```
Aprogiena: There is a constant for the local PID.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+            peerHDictionary = new Dictionary<int, int>();
+            for (int i = 0; i < peerIdsDictionary.Count - 1; i++)
+            {
+                peerHDictionary.Add(i, 0);
+                peerEntryDictionary.Add(i, 0);
+            }
+
+            foreach (KeyValuePair<uint256, HashSet<int>> tips in tipsDictionary)
+            {
+                if (tips.Value.Count == 0) noUnclaimedBraches = false;
```
Aprogiena: Besides this condition, which is correct, and unclaimed branch could also exist if there is unregistered tip. To recognise that, we just take all chain headers from CHT and find all those that have empty next pointers. These selected headers must all be within the tips dictionary.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+        /// 6/6 PeerTipsByPeerId should reflect PeerIdsByHash(except local marker).
+        /// </summary>
+        /// <param name="cht">ChainHeaderTree.</param>
+        /// <param name="listOfAllUniqueBlockHeaders">List of All unique blockheader in all connected chains.</param>
+        private void CheckChainedHeaderTreeConsistency(ChainedHeaderTree cht, IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders)
+        {
+            bool noUnclaimedBraches = true;
+            bool eachPeerOneEntry = true;
+            bool reflecttipHash = true;
+            bool chbhHasOnlyReachebleHeaders = true;
```
Aprogiena: Spelling, but also you don't need these Boolean variables here or maybe at all if you just assert immediately when you send them to false

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+            chainedHeadersWithoutPeerK.Should().BeEquivalentTo(chainedHeadersWithPeerK);
+
+            this.CheckChainedHeaderTreeConsistency(cht, listOfAllUniqueBlockHeaders);
+        }
+
+        /// <summary>
+        /// Basic test for tests 18 to 20, 28.
+        /// 1/6 There is no branch which is claimed by no one (PeerTipsByHash have tip of every branch), none of list should be empty.
+        /// 2/6 ChainHeadersByHash contains only reachable headers.
+        /// 3/6 ChainHeader claims LocalPeer.
```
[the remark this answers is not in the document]
Aprogiena: The consensus tip can be obtained from peerTipsByPeerId (or by calling GetConsensusTip()).
To check this condition, you should pass expected CT to this method and compare it with the hash in CHT.peerTipsByPeerId[LocalPeerId] (or GetConsensusTip result). Expected CT depends on the test scenario.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+        /// 4/6 Each connected peer has exactly 1 entry in PeertipsByHash.
```
Aprogiena: Should be `in peerIdsByTipHash`

Aprogiena: Probably the list of expected connected peers should also be provided in the argument.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+        {
+            bool noUnclaimedBraches = true;
+            bool eachPeerOneEntry = true;
+            bool reflecttipHash = true;
+            bool chbhHasOnlyReachebleHeaders = true;
+            bool isCorrectChainSequence = true;
+            bool hasLocalPeer = false;
+
+            var peerEntryDictionary = new Dictionary<int, int>();
+            var peerHDictionary = new Dictionary<int, int>();
```
Aprogiena: No idea from the variable name about what it is.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1992,6 +2054,117 @@ public void ChainTHasCheckpointAt1000_MaxReorgIs10_TwoChainsPriorTo1000Presented
...
+
+            Dictionary<uint256, HashSet<int>> tipsDictionary = cht.GetPeerIdsByTipHash();
+            Dictionary<int, uint256> peerIdsDictionary = cht.GetPeerTipsByPeerId();
+
+            peerHDictionary = new Dictionary<int, int>();
```
Aprogiena: We just created a dictionary five lines ago

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,68 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+            ///                        =8d=9d=10d
+            ///                   6a=7a=8a=9a
+            /// GENESIS=1=2=3=4=5=
+            ///                   6b=7b=8b=9b
+            ///             3c=4c=5c
+            /// </summary>
+            /// <param name="initialChainSize">Initial chain size.</param>
+            /// <param name="cht">ChainHeaderTree.</param>
+            /// <param name="initialChainTip">Initial chain tip.</param>
+            /// <param name="chainATip">out Chain A tip.</param>
```
Aprogiena: I don't think we need these out parameters at all, we do have extension methods providing those headers already. I think it is called GetPeerTipChainedHeaderByPeerId.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -202,6 +202,68 @@ public bool NoDownloadRequested(ConnectNewHeadersResult connectNewHeadersResult)
...
+                listOfAllBlockHeaders.AddRange(peerCBlockHeaders);
+                listOfAllBlockHeaders.AddRange(peerDBlockHeaders);
+                IEnumerable<BlockHeader> listOfAllUniqueBlockHeaders = listOfAllBlockHeaders.Distinct();
+
+                cht.ConnectNewHeaders(0, peerABlockHeaders);
+                cht.ConnectNewHeaders(1, peerBBlockHeaders);
+                cht.ConnectNewHeaders(2, peerCBlockHeaders);
+                cht.ConnectNewHeaders(3, peerDBlockHeaders);
+
+                return listOfAllUniqueBlockHeaders;
```
Aprogiena: We shouldn't need to return this.

kogot: @Aprogiena based on your comments, maybe you just rewrite whole thing yourself? cause every single thing is not correct for you.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
kogot: yes but it's internal, so cannot be used in instance.
 internal const int LocalPeerId = -1;

Aprogiena: can - https://github.com/kogot/StratisBitcoinFullNode/blob/cb66f8717a703b05499597c100f7f1dfa64f88dd/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs#L305

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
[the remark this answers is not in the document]
kogot: Can't use none of them, cause of protection level.

kogot:
Thanks guys. Gonna celebrate it : )

PR #1613, opened 2018-07-04
Update Dockerfile with 2.1 sdk

r3lik:
No longer builds with 2.0:

`The current .NET SDK does not support targeting .NET Core 2.1.  Either target .NET Core 2.0 or lower, or use a version of the .NET SDK that supports .NET Core 2.1.`


bokobza:
Thanks for doing this @r3lik !
Can you please also do the other ones in the folder?
Thanks 😄

dangershony:
Hi @r3lik please do the other one in another PR

PR #1615, opened 2018-07-05
CHT test: Calling  PartialValidationSucceeded for various scenarios - issue 51, 52 and 53

fenix2222:
Call PartialValidationSucceeded on a chained header which doesn't have the block data but it is in the tree, make sure 
PartialValidationSucceeded returns null and ReorgRequired is false.

Call PartialValidationSucceeded on a chained header which prev header has validation state as HeaderOnly, make sure PartialValidationSucceeded 
returns null and ReorgRequired is false.

Call PartialValidationSucceeded on a chained header which validation state is PartiallyValid, make sure PartialValidationSucceeded returns 
null and ReorgRequired is false.

See #1321 (comment)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,87 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            // and full validation is not required.
+            List<ChainedHeaderBlock> headers = cht.PartialValidationSucceeded(chainTip, out bool fullValidationRequired);
+            headers.Should().BeNull();
+            fullValidationRequired.Should().BeFalse();
+        }
+
+        /// <summary>
+        /// Issue 52 @ Call PartialValidationSucceeded on a chained header which prev header has validation state as HeaderOnly,
+        /// make sure PartialValidationSucceeded returns null and ReorgRequired is false.
+        /// </summary>
```
Aprogiena: You original description assumed additional conditions that were not included in it. Here we need the chained header to be downloaded – i.e. the block pointer to be not null.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,87 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+        [Fact]
+        public void CallingPartialValidationSucceeded_FoHeaderWithPreviousHeaderWithValidationStateHeaderOnly_NothingReturnedAndFullValidationIsNotRequired()
+        {
+            // Chain header tree setup. Initial chain has 5 headers with no blocks.
+            // Example: h1=h2=h3=h4=h5.
+            const int initialChainSize = 5;
+            TestContext ctx = new TestContextBuilder()
+                .WithInitialChain(initialChainSize, assignBlocks: false)
```
Aprogiena: Here we don't want false.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,87 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+        /// make sure PartialValidationSucceeded returns null and ReorgRequired is false.
+        /// </summary>
+        [Fact]
+        public void CallingPartialValidationSucceeded_FoHeaderWithValidationStateAsPartiallyValidated_NothingReturnedAndFullValidationIsNotRequired()
+        {
+            // Chain header tree setup. Initial chain has 5 headers with no blocks.
+            // Example: h1=h2=h3=h4=h5.
+            const int initialChainSize = 5;
+            TestContext ctx = new TestContextBuilder()
+                .WithInitialChain(initialChainSize, assignBlocks: false)
```
Aprogiena: Same here, we want the blocks to be downloaded, and we want to have partially validated predecessor, and we also want to have partially validated the chained header we are testing. Basically you want to call partially twice on the tested header, when all other conditions are normal – previous headers are validated and all guys are downloaded.

Aprogiena: Tests 51 to 53 are supposed to incrementally add new conditions for the partial validation. so the first one we have is correct, the second basically fixes the condition on which the first failed and same for the third, it fixes both conditions on which the first one and the second one failed.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,110 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            bool fullValidationRequired;
+            for (int i = 0; i < headerArray.Length; i++)
+            {
+                cht.BlockDataDownloaded(headerArray[i], originalHeaderArray[i].Block);
+                cht.PartialValidationSucceeded(headerArray[i], out fullValidationRequired);
+            }
+
+            // Call partial validation on h9 again and make sure nothing is returned
+            // and full validation is not required.
+            List<ChainedHeaderBlock> headers = cht.PartialValidationSucceeded(chainTip, out fullValidationRequired);
```
Aprogiena: this should be called on `consumed`, not `chainTip`

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,110 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            ChainedHeader[] originalHeaderArray = chainTip.ToArray(extensionSize);
+            ChainedHeader[] headerArray = consumed.ToArray(extensionSize);
+            for (int i = 0; i < headerArray.Length; i++)
+            {
+                cht.BlockDataDownloaded(headerArray[i], originalHeaderArray[i].Block);
+            }
+            
+            // Call partial validation on h9 (h8 has validation state as HeaderOnly) and make sure nothing is returned
+            // and full validation is not required.
+            List<ChainedHeaderBlock> headers = cht.PartialValidationSucceeded(chainTip, out bool fullValidationRequired);
```
Aprogiena: Partial validation should be called on the header from tree, Not chain tip, which is not in it.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -1652,5 +1652,110 @@ public void ChainWith10DataAvailableHeaders_PartialValidationSucceededOnTheFirst
...
+            const int initialChainSize = 5;
+            TestContext ctx = new TestContextBuilder()
+                .WithInitialChain(initialChainSize, assignBlocks: false)
+                .Build();
+            ChainedHeaderTree cht = ctx.ChainedHeaderTree;
+            ChainedHeader chainTip = ctx.InitialChainTip;
+            
+            // Call partial validation on h4 and make sure nothing is returned
+            // and full validation is not required.
+            List<ChainedHeaderBlock> headers = cht.PartialValidationSucceeded(chainTip, out bool fullValidationRequired);
```
Aprogiena: Actually, this test is also not right. We need to present the headers and then call this one of the headers that there consumed, because the initial chain is fully validated by definition already.

PR #1616, opened 2018-07-05
Cleanup network usage in NBitcoin #4

fassadlr:
Removing more usages of Network.Main and unused usages of network.


src/NBitcoin/RPC/RestClient.cs
```diff
@@ -247,9 +219,9 @@ private WebRequest BuildHttpRequest(string resource, RestResponseFormat format,
 
             HttpWebRequest request = WebRequest.CreateHttp(uriBuilder.Uri);
             request.Method = "GET";
-#if !NETCORE
+            #if !NETCORE
```
bokobza: Prepocessor directives are usually written with no margin on the left.
But it doesn't matter, I plan to remove those soon.

PR #1618, opened 2018-07-05
Moved Json converters to the Stratis.Bitcoin project

bokobza:
Moving these because they're sitting at the top of NBitcoin but can easily be moved to Stratis.Bitcoin.

### DIFF
--- a/src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
+++ b/src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
@@ -8,112 +8,93 @@ namespace Stratis.Bitcoin.Utilities.JsonConverters
 {
     public class CoinJsonConverter : JsonConverter
     {
-        public class CoinJson
-        {
-            public CoinJson()
-            {
-
-            }
-            public CoinJson(ICoin coin, Network network)
-            {
-                if (network == null)
-                    network = Network.Main;
-                this.TransactionId = coin.Outpoint.Hash;
-                this.Index = coin.Outpoint.N;
-                this.ScriptPubKey = coin.TxOut.ScriptPubKey;
-                if (coin is ScriptCoin)
-                {
-                    this.RedeemScript = ((ScriptCoin)coin).Redeem;
-                }
-                if(coin is Coin)
-                {
-                    this.Value = ((Coin)coin).Amount;
-                }
-                if (coin is ColoredCoin)
-                {
-                    var cc = (ColoredCoin)coin;
-                    this.AssetId = cc.AssetId.GetWif(network);
-                    this.Quantity = cc.Amount.Quantity;
-                    this.Value = cc.Bearer.Amount;
-                    var scc = cc.Bearer as ScriptCoin;
-                    if (scc != null)
-                    {
-                        this.RedeemScript = scc.Redeem;
-                    }
-                }
-            }
-            public ICoin ToCoin()
-            {
-                Coin coin = this.RedeemScript == null ? new Coin(new OutPoint(this.TransactionId, this.Index), new TxOut(this.Value, this.ScriptPubKey)) : new ScriptCoin(new OutPoint(this.TransactionId, this.Index), new TxOut(this.Value, this.ScriptPubKey), this.RedeemScript);
-                if (this.AssetId != null)
-                    return coin.ToColoredCoin(new AssetMoney(this.AssetId, this.Quantity));
-                return coin;
-            }
-
-            public uint256 TransactionId
-            {
-                get;
-                set;
-            }
-            public uint Index
-            {
-                get;
-                set;
-            }
-            public Money Value
-            {
-                get;
-                set;
-            }
-
-            public Script ScriptPubKey
-            {
-                get;
-                set;
-            }
-
-            public Script RedeemScript
-            {
-                get;
-                set;
-            }
-            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
-            public BitcoinAssetId AssetId
-            {
-                get;
-                set;
-            }
-            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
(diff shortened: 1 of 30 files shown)


src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Stratis.Bitcoin.Utilities.JsonConverters
+{
+    /// <summary>
+    /// Converter used to convert an object implementing <see cref="ICoin"/> to and from JSON.
+    /// </summary>
+    /// <seealso cref="Newtonsoft.Json.JsonConverter" />
+    public class CoinJsonConverter : JsonConverter
+    {
+        public Network Network { get; set; }
```
fassadlr: Lets make this readonly and private?

src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
```diff
@@ -0,0 +1,101 @@
...
+        public BitcoinAssetId AssetId { get; set; }
+
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public long Quantity { get; set; }
+
+        public CoinJson()
+        {
+        }
+
+        public CoinJson(ICoin coin, Network network)
```
fassadlr: Network isn't being used here.

src/Stratis.Bitcoin/Utilities/JsonConverters/JsonObjectException.cs
```diff
@@ -1,25 +1,20 @@
...
         {
             this.Path = reader.Path;
         }
 
-        public string Path
-        {
-            get;
-            private set;
-        }
+        public string Path { get; }
```
fassadlr: Looks like this can also be read only?

src/Stratis.Bitcoin/Utilities/JsonConverters/NetworkJsonConverter.cs
```diff
@@ -1,48 +1,65 @@
...
                 return Network.Main;
+
             if (network.Equals("TestNet", StringComparison.OrdinalIgnoreCase) || network.Equals("test", StringComparison.OrdinalIgnoreCase))
                 return Network.TestNet;
-            if(network.Equals("RegTest", StringComparison.OrdinalIgnoreCase) || network.Equals("reg", StringComparison.OrdinalIgnoreCase))
+
+            if (network.Equals("RegTest", StringComparison.OrdinalIgnoreCase) || network.Equals("reg", StringComparison.OrdinalIgnoreCase))
                 return Network.RegTest;
+
             Network net = NetworksContainer.GetNetwork(network);
```
fassadlr: Might as well rename to `network` here too,

src/Stratis.Bitcoin/Utilities/JsonConverters/Serializer.cs
```diff
@@ -21,19 +25,12 @@ public static void RegisterFrontConverters(JsonSerializerSettings settings, Netw
...
+            settings.Converters.Add(new BitcoinStringJsonConverter(network));
 
-        public static T ToObject<T>(string data)
-        {
-            return ToObject<T>(data, null);
+            settings.ContractResolver = new CamelCasePropertyNamesContractResolver();
         }
 
-        public static T ToObject<T>(string data, Network network)
+        public static T ToObject<T>(string data, Network network = null)
```
fassadlr: Ah mate can we remove these null defaults? I really want us to move away from these 👍

src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
bokobza: yep

src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
bokobza: it's used here, line 83: `this.AssetId = cc.AssetId.GetWif(network);`

src/Stratis.Bitcoin/Utilities/JsonConverters/JsonObjectException.cs
bokobza: yep

src/Stratis.Bitcoin/Utilities/JsonConverters/NetworkJsonConverter.cs
bokobza: there is already a network object in the class. So I've renamed `network` to `networkName` and `net` to `network`.

src/Stratis.Bitcoin/Utilities/JsonConverters/Serializer.cs
bokobza: We don't assign a default Network if it's null and having a null Network is valid as well.
Do you still want me to make it non-null?

fassadlr: Hmm if there's no need to make it null then I would say leave it as non-null? If the need arises we can have overloads.

src/Stratis.Bitcoin/Utilities/JsonConverters/CoinJsonConverter.cs
fassadlr: OK cool!

src/Stratis.Bitcoin/Utilities/JsonConverters/Serializer.cs
bokobza: I think it's ok to leave it like this.
In Stratis.Bitcoin.Features.Api.Startup we call the Serializer.RegisterFrontConverters method but we don't have a Network.

PR #1621, opened 2018-07-05
Update Dockerfile

r3lik:
No longer builds with 2.0:

`The current .NET SDK does not support targeting .NET Core 2.1.  Either target .NET Core 2.0 or lower, or use a version of the .NET SDK that supports .NET Core 2.1.`


r3lik:
haha my bad

PR #1624, opened 2018-07-06
Cht tests fazzdev #1321 45

thefazzer:
**Cht tests fazzdev #1321 45**

- Node has a reorg of 5 blocks.
- Make sure that headers for the blocks that were reorged away.
- Have no block pointers and block data availability == headers only.

Preferred: @Aprogiena @noescape00

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
@@ -2129,10 +2129,19 @@ namespace Stratis.Bitcoin.Tests.Consensus
             ChainedHeaderTree chainedHeaderTree = testContext.ChainedHeaderTree;
             ChainedHeader chainTip = testContext.InitialChainTip;
 
+            // First peer presents headers from the original chain.
+            List<BlockHeader> listOfExistingHeaders = testContext.ChainedHeaderToList(chainTip, initialChainSize);
+            ConnectNewHeadersResult connectNewHeadersResult = chainedHeaderTree.ConnectNewHeaders(1, listOfExistingHeaders);
+
+            // None are marked for download.
+            connectNewHeadersResult.DownloadFrom.Should().Be(null);
+            connectNewHeadersResult.DownloadTo.Should().Be(null);
+
+            // Second peer presents fork.
             ChainedHeader forkedBlockHeader = chainTip.GetAncestor(depthOfReorg);
             ChainedHeader newTip = testContext.ExtendAChain(extensionSize, forkedBlockHeader);
             List<BlockHeader> listOfExtendedHeaders = testContext.ChainedHeaderToList(newTip, heightOfFork + extensionSize);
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -52,7 +53,8 @@ internal TestContextBuilder WithInitialChain(int initialChainSize, bool assignBl
                 if (initialChainSize < 0)
                     throw new ArgumentOutOfRangeException(nameof(initialChainSize), "Size cannot be less than 0.");
 
-                this.testContext.InitialChainTip = this.testContext.ExtendAChain(initialChainSize, assignBlocks: assignBlocks);
+                this.testContext.InitialChainTip =
```
Aprogiena: Could you please configure your visual studio so that you do not enforce such strict limits on the number of characters in the line?

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -129,31 +131,33 @@ public void SetupCheckpoints(params CheckpointFixture[] checkpoints)
 
                 this.CheckpointsMock
                     .Setup(c => c.GetCheckpoint(It.IsNotIn(checkpoints.Select(h => h.Height))))
-                    .Returns((CheckpointInfo)null);
+                    .Returns((CheckpointInfo) null);
```
Aprogiena: Also this should be configured because it is not consistent style with what we have elsewhere.

Aprogiena:
Could you please revert all the code style changes and leave just that code of the test? Also please make changes to your configuration of code style in visual studio.

thefazzer:
@Aprogiena reverted all style changes, re-added test, disabled line wrapping behavior in my dev env

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -2112,6 +2112,48 @@ public void PresentHeaders_ChainHeaderTreeAdvances_UnconsumedBlocksDataBytes_Equ
...
+            while (chainHeader.Height > forkedBlockHeader.Height)
+            {
+                // Header has no block pointer.
+                Assert.Null(chainHeader.Block);
+
+                // Header block data availability == headers only.
+                Assert.Equal(BlockDataAvailabilityState.HeaderOnly, chainHeader.BlockDataAvailability);
+                chainHeader = chainHeader.Previous;
+            }
+        }
```
Aprogiena: Looks quite good, I would just like to have a second peer to claim the original chain so that it is not removed from the tree when we fork away.So all I'm asking for is to have a second peer to present headers for the original chain before we present the fork.

thefazzer: @Aprogiena Made changes - have I understood them correctly?

Aprogiena:
Good now

PR #1630, opened 2018-07-07
CM and CHT minor optimizations 

noescape00:


Aprogiena:
Okay, basically we just introduced temporarily the concurrent chain so that we can set its tip, which is needed for the activation without removing concurrent chain from the code base.

Issue #1625, opened 2018-07-06
Unable to mint PoS coins throguh staking


einewton:
Referring to the test case Staking_wallet_can_mint_new_coins this method does 6 things. The test runs successful, however it is invalid.

When reading through the 6 things this test case runs, first it will create a genesis block via PoW, with a PoS node will stake the pre-mined coins, during the stake process new coins **should** show up in your wallet, but this is not the case.

The newly generated coins are done through the PoW process, not the PoS process. The easiest way to see this is setting the StratisRegTest network reward to zero like this:
this.Consensus.ProofOfWorkReward = Money.Zero;

Setting this to Zero, the PoW node will not create any new coins, only the PoS staking process will.

Then fix the test on "pos_node_wallet_has_earned_coins_through_staking" like so:

Referring to ProofOfStakeSteps.cs on line 80: https://github.com/stratisproject/StratisBitcoinFullNode/blob/c82b0200b4bfd59b2efd6748c5d0810ad94f1b48/src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs#L80

`return staked >= Money.COIN * 98000000;`

needs to be

`return staked > Money.COIN * 98000000;`

Then you can see there are no new coins minted through staking.

Could this be related to issue #1113 ?

fassadlr:
Hi @einewton, yes the test's assert was incorrect. However, there was other work that needed to be done on block creation to ensure that the test passes correctly. A PR for this will be submitted today. Thanks for pointing this out 👍 

(This relates to #1614 as well)

bokobza:
@fassadlr please close this if possible.

fassadlr:
Fixed with #1634

PR #1634, opened 2018-07-09
[Mining] Add the ability to create proof of work blocks on a proof of stake network.

### LINKED ISSUE #1614: Incorrect proof of work
Setting up a local network of two nodes, I had two nodes both mining, the first node was online first and mined 2 blocks, then I turned on the second node and it mined 1 block before it synced up with the other node. Then after a short while, the second node started complaining about "incorrect proof of work" (the node finds a new block).

```
Consensus error exception occurred in miner loop: Stratis.Bitcoin.Consensus.ConsensusErrorException: incorrect proof of work

at Stratis.Bitcoin.Consensus.ConsensusError.Throw() in C:\...\src\Stratis.Bitcoin\Consensus\ConsensusErrorException.cs:line 44

at Stratis.Bitcoin.Features.Consensus.Rules.CommonRules.BlockHeaderPowContextualRule.RunAsync(RuleContext context) in C:\...\src\Stratis.Bitcoin.Features.Consensus\Rules\CommonRules\BlockHeaderPowContextualRule.cs:line 33
```

The code at the line:

```
// Check proof of work.
if (header.Bits != context.NextWorkRequired)
```

Here is a zip of the data folder. This is built using the "main" Stratis network, but I have modified the node to not use checkpoints, no DNS seed and only my LAN nodes. The only modification to stratis.conf, is mine=1 and the address.

[StratisMain.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2164378/StratisMain.zip)

I have attempted a few times to use this data folder when running StratisD from Visual Studio (Debugging), but then the problem does not occur. Only when I run from cmd, with "dotnet stratis.stratisd.dll".

I did a run with debug level on trace, and here is the log output from a quick startup and close. Search for "Validating new block" to find the interesting parts.

[node.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2164421/node.zip)

fassadlr:
This PR fixes #1614 and #1625

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/BlockProvider.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/BlockProvider.cs
@@ -8,17 +8,25 @@ namespace Stratis.Bitcoin.Features.Miner
     /// <inheritdoc/>
     public sealed class BlockProvider : IBlockProvider
     {
+        private readonly Network network;
+
         /// <summary>Defines how proof of work blocks are built.</summary>
         private readonly PowBlockDefinition powBlockDefinition;
 
         /// <summary>Defines how proof of stake blocks are built.</summary>
         private readonly PosBlockDefinition posBlockDefinition;
 
+        /// <summary>Defines how proof of work blocks are built on a Proof-of-Stake network.</summary>
+        private readonly PosPowBlockDefinition posPowBlockDefinition;
+
         /// <param name="definitions">A list of block definitions that the builder can utilize.</param>
-        public BlockProvider(IEnumerable<BlockDefinition> definitions)
+        public BlockProvider(Network network, IEnumerable<BlockDefinition> definitions)
         {
+            this.network = network;
+
             this.powBlockDefinition = definitions.OfType<PowBlockDefinition>().FirstOrDefault();
             this.posBlockDefinition = definitions.OfType<PosBlockDefinition>().FirstOrDefault();
+            this.posPowBlockDefinition = definitions.OfType<PosPowBlockDefinition>().FirstOrDefault();
         }
 
         /// <inheritdoc/>
@@ -30,6 +38,9 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <inheritdoc/>
         public BlockTemplate BuildPowBlock(ChainedHeader chainTip, Script script)
         {
+            if (this.network.Consensus.IsProofOfStake)
+                return this.posPowBlockDefinition.Build(chainTip, script);
+
             return this.powBlockDefinition.Build(chainTip, script);
         }
     }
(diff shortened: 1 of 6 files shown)


fassadlr:
I have successfully sync all the chainz.

fassadlr:
Staking works as per normal 👍

src/Stratis.Bitcoin.Features.Miner/BlockProvider.cs
```diff
@@ -30,6 +38,9 @@ public BlockTemplate BuildPosBlock(ChainedHeader chainTip, Script script)
         /// <inheritdoc/>
         public BlockTemplate BuildPowBlock(ChainedHeader chainTip, Script script)
         {
+            if (this.network.Consensus.IsProofOfStake)
```
dangershony: Ah you made the change here, I was planing to do it in the PowMiner.cs but I guess here is just as fine.

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -168,110 +168,181 @@ public void StopMining()
...
-                    // where it is possible to generate multiple blocks within one second.
-                    if (blockTemplate.Block.Header.Time <= chainTip.Header.Time)
-                        continue;
-                }
+                if (!BuildBlock(context))
+                    continue;
 
-                nExtraNonce = this.IncrementExtraNonce(blockTemplate.Block, chainTip, nExtraNonce);
-                Block block = blockTemplate.Block;
+                if (!MineBlock(context))
```
dangershony: Reminder to add `this.` on all 3 mtheods

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -168,110 +168,181 @@ public void StopMining()
...
+
+            var chainedHeader = new ChainedHeader(context.BlockTemplate.Block.Header, context.BlockTemplate.Block.GetHash(), context.ChainTip);
+            if (chainedHeader.ChainWork <= context.ChainTip.ChainWork)
+                return false;
+
+            return true;
+        }
 
-                this.logger.LogInformation("Mined new {0} block: '{1}'.", BlockStake.IsProofOfStake(blockValidationContext.Block) ? "POS" : "POW", blockValidationContext.ChainedHeader);
+        private void ValidateMinedBlockWithConsensus(MineBlockContext context)
```
dangershony: I would suggest to merge this two methods `ValidateMinedBlockWithConsensus` `CheckValidationContext`
And call it `ValidateAndconnectBlock()`

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
fassadlr: 👍

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
fassadlr: 💯

PR #1638, opened 2018-07-09
CHB optimization

noescape00:
Slightly optimized CHB. This is the 1st PR to get CHB ready for the CHT and CM activation. 

Tests pass, networks synced

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -66,8 +66,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// </remarks>
         private uint256 getBlocksBatchLastItemHash;
 
-        /// <summary>The last block sent to the peer.</summary>
-        private ChainedHeader lastBlockSent;
+        /// <summary>Chained header of the last header sent to the peer.</summary>
+        private ChainedHeader lastHeaderSent;
 
         public BlockStoreBehavior(
             ConcurrentChain chain,
@@ -189,8 +189,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
             ChainedHeader highestHeader = peerTip;
 
-            if ((highestHeader == null) || ((this.lastBlockSent != null) && (highestHeader.Height < this.lastBlockSent.Height)))
-                highestHeader = this.lastBlockSent;
+            if ((highestHeader == null) || ((this.lastHeaderSent != null) && (highestHeader.Height < this.lastHeaderSent.Height)))
+                highestHeader = this.lastHeaderSent;
 
             this.logger.LogTrace("(-):'{0}'", highestHeader);
             return highestHeader;
@@ -293,7 +293,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 if (highestHeader?.Height < lastAddedChainedHeader.Height)
                 {
                     this.logger.LogTrace("Setting peer's last block sent to '{0}'.", lastAddedChainedHeader);
-                    this.lastBlockSent = lastAddedChainedHeader;
+                    this.lastHeaderSent = lastAddedChainedHeader;
 
                     // Set last item of the batch (unless we are announcing the tip), which is then used
                     // when the peer sends us "getdata" message. When we detect "getdata" message for this block,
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -176,6 +179,23 @@ protected override void DetachCore()
             this.logger.LogTrace("(-)");
         }
 
+        /// <summary>Gets the highest header between the header of the last block that was sent to a peer and the expected peer's tip.</summary>
+        private ChainedHeader GetHighestHeader(INetworkPeer peer)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(peer), peer.RemoteSocketEndpoint);
+
+            var chainBehavior = peer.Behavior<ChainHeadersBehavior>();
```
dangershony: I tend to set `chainBehavior` as a field on the behaviour and set it on `AttachCore`

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -33,62 +30,23 @@ public class ChainHeadersBehavior : NetworkPeerBehavior
...
-                return this.Chain.GetBlock(tip.HashBlock) ?? tip;
-            }
-        }
-
-        /// <summary><c>true</c> to respond to "getheaders" messages, <c>false</c> to ignore it.</summary>
-        public bool CanRespondToGetHeaders { get; set; }
+        /// <remarks>
+        /// The announced tip is accepted if it seems to be valid. Validation is only done on headers and so the announced tip may refer to invalid block.
+        /// </remarks>
+        public ChainedHeader ExpectedTip { get; private set; }
```
dangershony: Should this not be `PeerTip`?

noescape00: That's expected peer tip. 
We were arguing with apro on what name suits best, ended up with ExpectedTip. 
I can rename to ExpectedPeerTip if you'd like.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
```diff
@@ -373,47 +313,29 @@ protected override void DetachCore()
...
 
             // If we made any advancement or the sync is enforced by 'doTrySync'- continue syncing.
-            if (doTrySync || (this.pendingTip == null) || (pendingTipBefore == null) || (pendingTipBefore.HashBlock != this.pendingTip.HashBlock))
+            if (doTrySync || (this.ExpectedTip == null) || (pendingTipBefore == null) || (pendingTipBefore.HashBlock != this.ExpectedTip.HashBlock))
                 await this.TrySyncAsync().ConfigureAwait(false);
 
             this.logger.LogTrace("(-)");
         }
 
-        public void SetPendingTip(ChainedHeader newTip)
```
dangershony: Is this method not used?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
noescape00: Yeah. This whole behavior requires a refactoring, it's a big mess. 

Will fix this one.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
dangershony: Ok if there was an argument then I will just stay out of it leave as is.

noescape00: Not really an argument, just didn't know what would be the name. 

ExpectedPeerTip sounds good too, will fix.

src/Stratis.Bitcoin/Base/ChainHeadersBehavior.cs
noescape00: No. It used to be when BS behavior was using pending tip. Now we have 2 tips (one represents the last block that was sent to the peer and another represents last header received from the peer). 
BTC Core do that so we decided to copy that logic.

noescape00:
@dangershony applied fixes.

noescape00:
Synced the networks 2nd time just in case.

PR #1641, opened 2018-07-09
Moved Rpc stuff from NBitcoin to Stratis.Bitcoin.Features.RPC 

bokobza:
The goal of this change is to move all the RPC-related classes from NBitcoin to where they should be located: the RPC feature.
The move was pretty seamless and now the whole RPC layer lives inside the feature.

**Please note**: I haven't made any effort to modify the moved files to conform to our code policy. I believe a few of these methods and classes could be removed and the overall architecture could be improved, so I didn't want to start refactoring unnecessarily. 
This is something we should definitely look into doing at another time.

⚽️ unit tests
⚽️ integration tests
⚽️ 4 networks sync


bokobza:
thanks bruh

PR #1644, opened 2018-07-10
Matching peers : PeerVersion can be null

monsieurleberre:
Hopefully this is exactly the same except that we now check that PeerVersion is not null and has an address
I was getting exceptions when it was

### DIFF
--- a/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerCollection.cs
+++ b/src/Stratis.Bitcoin/P2P/Peer/NetworkPeerCollection.cs
@@ -103,16 +103,16 @@ namespace Stratis.Bitcoin.P2P.Peer
 
         private static bool Match(IPAddress ip, int? port, INetworkPeer peer)
         {
-            if (port.HasValue)
-            {
-                return ((peer.State == NetworkPeerState.Connected || peer.State == NetworkPeerState.HandShaked) && peer.RemoteSocketAddress.Equals(ip) && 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerCollection.cs
```diff
@@ -103,16 +103,16 @@ public INetworkPeer FindByEndpoint(IPEndPoint endpoint)
...
-                        peer.RemoteSocketAddress.Equals(ip)) || peer.PeerVersion.AddressFrom.Address.Equals(ip);
-            }
+            bool isConnectedOrHandShaked = (peer.State == NetworkPeerState.Connected || peer.State == NetworkPeerState.HandShaked);
+
+            bool isAddressMatching = peer.RemoteSocketAddress.Equals(ip) 
+                                     && (!port.HasValue || port == peer.RemoteSocketPort);
+
+            bool isPeerVersionAddressMatching = peer.PeerVersion?.AddressFrom != null
+                                                && peer.PeerVersion.AddressFrom.Address.Equals(ip)
+                                                && (!port.HasValue || port == peer.PeerVersion.AddressFrom.Port);
```
dangershony: Does this get disabled if its false?
` && (!port.HasValue || port == peer.PeerVersion.AddressFrom.Port)`

monsieurleberre: as discussed you can read it as "if port has no value, or if part has a value and that value matches" :)

dangershony:
Networks synced looks good.

PR #1645, opened 2018-07-10
remove node should send message synchronously

monsieurleberre:
This looks like a mistake to me, especially considering that the next method is the Async version of this one.
If that was on purpose, please explain and I will happily change the PR


bokobza:
LGTM

PR #1647, opened 2018-07-10
peers in the list are not necessarily connected

monsieurleberre:
so if the peers are in the list but disconnected this return the wrong answer

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
@@ -50,8 +50,7 @@ namespace Stratis.Bitcoin.IntegrationTests.Common
 
         public static bool IsNodeConnected(CoreNode node)
         {
-            if (node.FullNode.ConnectionManager.ConnectedPeers.Any()) return true;
-            return false;
+            return node.FullNode.ConnectionManager.ConnectedPeers.Any(p => p.IsConnected);
         }
 
         /// <summary>


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -50,8 +50,7 @@ public static void TriggerSync(CoreNode node)
 
         public static bool IsNodeConnected(CoreNode node)
         {
-            if (node.FullNode.ConnectionManager.ConnectedPeers.Any()) return true;
-            return false;
+            return node.FullNode.ConnectionManager.ConnectedPeers.Any(p => p.IsConnected);
```
dangershony: What about HandShaked? or is it covered in IsConnected?
If yes then push.

bokobza: The question is, where this IsNodeConnected method is used, do we care about the node having some peers connected to it, or do we want to see that they're actually connected.

dangershony: I am not sure too be honest, perhaps only check that there was a connection, but then again if connection failed something went wrong

monsieurleberre: As far as I have noticed, you can have Peers in the connectedPeers that are marked disconnected but stay in the list. That gives us timeouts when we check that a not is not connected in some integration tests (ex: reorg tests)

monsieurleberre: image
Tadaaaa :)

dangershony: >you can have Peers in the connectedPeers that are marked disconnected but stay in the list.

Do they stay in the list just a little longer and get removed or they don't get removed at all?
This is crucial if bad peers just stay in the list we have a bug, else it was just an expected race condition.

Issue #1555, opened 2018-06-25
Resolve todo issues in mempool behavior 


mikedennis:
Work based upon bitcoin core implementation

mikedennis:
@noescape00 @dangershony I notice that there is logic in btc core related to setting min tx fee on peers. Do you guys know do we have that feature in our peer implementation?

dangershony:
Its min relay tx fee no? (meaning how much fee required to relay a trx)
I think we have it and its configurable.

noescape00:
Yeah, we have it. And it's lower than min fee to include it to the block.

mikedennis:
```ProcessInvAsync()``` logic around blocksOnly mode is incorrect. It currently is just outputting a warning and still sending the the tx, but what btc core is doing is it should just add it to known inventory but not relay it to peer.

**Comment by mikedennis** [comment]
A few other todo's that need to be discussed for "INV" (``` ProcessInvAsync() ```)message:
* Misbehaving logic (basically peer banscore). I don't think we have banscore implemented at all in our p2p? Is this something we want to implement?
* Btc process blocks in inventory in addition to tx (we currently only process tx's). It looks like when btc receives a block in inventory that it doesn't already have it asks the peer for it's headers. Is this something we should be doing as well? 
```
if (inv.type == MSG_BLOCK) {
                UpdateBlockAvailability(pfrom->GetId(), inv.hash);
                if (!fAlreadyHave && !fImporting && !fReindex && !mapBlocksInFlight.count(inv.hash)) {
                    // We used to request the full block here, but since headers-announcements are now the
                    // primary method of announcement on the network, and since, in the case that a node
                    // fell back to inv we probably have a reorg which we should get the headers for first,
                    // we now only provide a getheaders response here. When we receive the headers, we will
                    // then ask for the blocks we need.
                    connman->PushMessage(pfrom, msgMaker.Make(NetMsgType::GETHEADERS, chainActive.GetLocator(pindexBestHeader), inv.hash));
                    LogPrint(BCLog::NET, "getheaders (%d) %s to peer=%d\n", pindexBestHeader->nHeight, inv.hash.ToString(), pfrom->GetId());
                }
}
```
* There is some logic around combining fetch flags from the peer to the inventory type 
``` 
uint32_t nFetchFlags = GetFetchFlags(pfrom, chainActive.Tip(), chainparams.GetConsensus()); 
inv.type |= nFetchFlags;
``` 
[...]

PR #1649, opened 2018-07-10
Minor logic fix in MempoolBehavior for blockonly mode

mikedennis:
Match BTC core logic for inv message when in blockonly mode. Add to known inventory but don't relay. Related to issue #1555


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -315,26 +315,17 @@ public override object Clone()
...
-            }
-
-            // add to known inventory
-            lock (this.lockObject)
-            {
-                foreach (InventoryVector inventoryVector in send.Inventory)
-                {
-                    this.filterInventoryKnown.Add(inventoryVector.Hash);                    
-                }
+                if (this.isBlocksOnlyMode)
```
dangershony: Please add braces, I feel it will be more readable especially when using `else`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -315,25 +315,22 @@ public override object Clone()
...
+                {
+                    this.filterInventoryKnown.Add(inv.Hash);
+                }
 
-                if (await this.orphans.AlreadyHaveAsync(inv.Hash))
+                if (this.isBlocksOnlyMode)
+                {
+                    this.logger.LogInformation("Transaction ID '{0}' inventory sent in violation of protocol peer '{1}'.", inv.Hash, peer.RemoteSocketEndpoint);
+                }
+                else if (await this.orphans.AlreadyHaveAsync(inv.Hash))
```
Aprogiena: Configure await

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -315,25 +315,22 @@ public override object Clone()
             {
                 //inv.type |= nFetchFlags;
 
-                // TODO: This is incorrect, in blocks only mode we should just add to known inventory but not relay
-                if (this.isBlocksOnlyMode)                 
-                    this.logger.LogInformation("Transaction ID '{0}' inventory sent in violation of protocol peer '{1}'.", inv.Hash, peer.RemoteSocketEndpoint);
+                lock (this.lockObject)
```
Aprogiena: I think in general, it is better to grab the lock only once as it was in the original code instead of grabbing it for each item and then release it again.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -315,25 +315,22 @@ public override object Clone()
...
                 }
-
-                send.Inventory.Add(inv);
-            }
-
-            // add to known inventory
-            lock (this.lockObject)
-            {
-                foreach (InventoryVector inventoryVector in send.Inventory)
+                else
```
Aprogiena: The original version which continue seems little more readable to me, prevents you to have else if statements. Not sure if others feel the same.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -307,39 +307,42 @@ public override object Clone()
...
-                    this.logger.LogDebug("Transaction ID '{0}' already in orphans, skipped.", inv.Hash);
-                }
-                else
+            if (!this.isBlocksOnlyMode && peer.IsConnected)
+            {
+                var send = new GetDataPayload();
+                foreach (var inv in inventoryTxs)
                 {
-                    send.Inventory.Add(inv);
+                    if (await this.orphans.AlreadyHaveAsync(inv.Hash).ConfigureAwait(false))
```
Aprogiena: Previously this was executed without the peer needed to be connected.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -307,34 +307,35 @@ public override object Clone()
...
+            var inventoryTxs = invPayload.Inventory.Where(inv => inv.Type.HasFlag(InventoryType.MSG_TX));
 
-                // TODO: This is incorrect, in blocks only mode we should just add to known inventory but not relay
-                if (this.isBlocksOnlyMode)                 
-                    this.logger.LogInformation("Transaction ID '{0}' inventory sent in violation of protocol peer '{1}'.", inv.Hash, peer.RemoteSocketEndpoint);
+            lock (this.lockObject)
+            {
+                foreach (var inv in inventoryTxs)
+                {
+                    this.filterInventoryKnown.Add(inv.Hash);
```
Aprogiena: Is it intentional that potentially different items are now added to the filter than before?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -307,34 +307,35 @@ public override object Clone()
...
                 }
 
-                send.Inventory.Add(inv);
-            }
-
-            // add to known inventory
-            lock (this.lockObject)
-            {
-                foreach (InventoryVector inventoryVector in send.Inventory)
+                if (this.isBlocksOnlyMode)
```
Aprogiena: Previously this was the first condition in the for loop and it probably makes sense because you just want to log all transactions and do nothing otherwise. You will always execute the orphans and at the end you will just log it anyway. Is there any side effect of the orphan call?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
mikedennis: BTC core adds all items to the known inventory filter not just the ones that are sent like our previous logic did.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
mikedennis: There is some rejects logic inside that AlreadyHaveAsync() which I'm not sure we want to skip.

Aprogiena: The argument here is that unless the orphan function has any side effects, which means it can only decide whether to continue not, then the order of these conditions here should be reversed

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
Aprogiena: I'm not familiar with that, so if that is the case then of course this is fine and intentional

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
mikedennis: I do agree with that in theory the inv is being sent out of protocol so should be ignored but this is the part of the orphan function I'm concerned about not executing:
```
 if (this.chain.Tip.HashBlock != this.hashRecentRejectsChainTip)
 {
      // If the chain tip has changed previously rejected transactions
      // might be now valid, e.g. due to a nLockTime'd tx becoming valid,
      // or a double-spend. Reset the rejects filter and give those
      // txs a second chance.
      this.logger.LogTrace("Executing task to clear rejected transactions.");
      this.hashRecentRejectsChainTip = this.chain.Tip.HashBlock;
      this.recentRejects.Clear();
}
```
In BTC core they also always execute this orphan functionality. Not sure if this is intentional or just a side effect of the way they wrote it.

PR #1650, opened 2018-07-11
Cleanup network usage in NBitcoin #7

fassadlr:
This should be the last part before we start moving the network classes.


bokobza:
LGTM

PR #1652, opened 2018-07-12
Wallet xpub key support

majikandy:
Recover wallet and Load Wallet and protection from AddAccount using current endpoint.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
+using FluentAssertions;
 using Microsoft.Extensions.Logging;
 using Moq;
 using NBitcoin;
@@ -164,7 +165,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             string passphrase = "this is my magic passphrase";
 
             // create the wallet
-            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet", passphrase);
+            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet", passphrase: passphrase);
 
             // assert it has saved it to disk and has been created correctly.
             var expectedWallet = JsonConvert.DeserializeObject<Wallet>(File.ReadAllText(dataFolder.WalletPath + "/mywallet.wallet.json"));
@@ -273,7 +274,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             var mnemonicList = new Mnemonic(Wordlist.French, WordCount.Eighteen);
 
             // create the wallet
-            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet", mnemonicList: mnemonicList.ToString());
+            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet");
 
             Assert.Equal(mnemonic.DeriveSeed(), mnemonicList.DeriveSeed());
         }
@@ -285,7 +286,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
 
             var walletManager = this.CreateWalletManager(dataFolder, "-walletaddressbuffer=100");
 
-            walletManager.CreateWallet("test", "mywallet", new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
+            walletManager.CreateWallet("test", "mywallet", passphrase: new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
             HdAccount hdAccount = walletManager.Wallets.Single().AccountsRoot.Single().Accounts.Single();
 
@@ -3161,7 +3162,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             DataFolder dataFolder = CreateDataFolder(this);
 
             WalletManager walletManager = this.CreateWalletManager(dataFolder);
-            walletManager.CreateWallet("test", "mywallet", new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
+            walletManager.CreateWallet("test", "mywallet", passphrase: new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
             // Default of 20 addresses becuause walletaddressbuffer not set
             HdAccount hdAccount = walletManager.Wallets.Single().AccountsRoot.Single().Accounts.Single();
@@ -3179,20 +3180,22 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         }
 
         [Fact]
-        public void Xpubkey_only_wallet_can_load_without_private_key()
+        public void RecoverViaXpubkey_can_recover_wallet_without_mnemonic()
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            Wallet wallet = this.walletFixture.GenerateBlankWallet("testWallet", "password");
-            wallet.IsExtPubKeyWallet = true;
-            wallet.EncryptedSeed = null;
-            wallet.ChainCode = null;
-            
-            File.WriteAllText(Path.Combine(dataFolder.WalletPath, "testWallet.wallet.json"), JsonConvert.SerializeObject(wallet, Formatting.Indented, new ByteArrayConverter()));
-
+            const string stratisAccount0ExtPubKey = "xpub661MyMwAqRbcEgnsMFfhjdrwR52TgicebTrbnttywb9zn3orkrzn6MHJrgBmKrd7MNtS6LAim44a6V2gizt3jYVPHGYq1MzAN849WEyoedJ";
             var walletManager = this.CreateWalletManager(dataFolder);
+            walletManager.RecoverWalletViaExtPubKey("testWallet", stratisAccount0ExtPubKey, 1, DateTime.Now.AddHours(-2));
+            
+            var wallet = walletManager.LoadWallet("password", "testWallet");
 
-            walletManager.LoadWallet("password", "testWallet");
+            wallet.IsExtPubKeyWallet.Should().BeTrue();
+            wallet.EncryptedSeed.Should().BeNull();
+            wallet.ChainCode.Should().BeNull();
+
+            wallet.AccountsRoot.SelectMany(x => x.Accounts).Single().ExtendedPubKey
+                .Should().Be(stratisAccount0ExtPubKey);
         }
 
         private WalletManager CreateWalletManager(DataFolder dataFolder, params string[] cmdLineArgs)
@@ -3210,7 +3213,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
                 dataFolder, new Mock<IWalletFeePolicy>().Object, new Mock<IAsyncLoopFactory>().Object, new NodeLifetime(), DateTimeProvider.Default);
 
             // create the wallet
-            Mnemonic mnemonic = walletManager.CreateWallet(password, walletName, passphrase);
+            Mnemonic mnemonic = walletManager.CreateWallet(password, walletName, passphrase: passphrase);
             Wallet wallet = walletManager.Wallets.ElementAt(0);
 
             File.Delete(dataFolder.WalletPath + $"/{walletName}.wallet.json");
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -274,7 +274,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             var mnemonicList = new Mnemonic(Wordlist.French, WordCount.Eighteen);
 
             // create the wallet
-            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet");
+            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet", mnemonicList: mnemonicList.ToString());
 
             Assert.Equal(mnemonic.DeriveSeed(), mnemonicList.DeriveSeed());
         }
@@ -284,7 +284,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            var walletManager = this.CreateWalletManager(dataFolder, "-walletaddressbuffer=100");
+            var walletManager = this.CreateWalletManager(dataFolder, Network.Main, "-walletaddressbuffer=100");
 
             walletManager.CreateWallet("test", "mywallet", passphrase: new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
@@ -3161,7 +3161,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            WalletManager walletManager = this.CreateWalletManager(dataFolder);
+            WalletManager walletManager = this.CreateWalletManager(dataFolder, Network.Main);
             walletManager.CreateWallet("test", "mywallet", passphrase: new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
             // Default of 20 addresses becuause walletaddressbuffer not set
@@ -3170,7 +3170,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             Assert.Equal(20, hdAccount.InternalAddresses.Count);
             
             // Restart with walletaddressbuffer set
-            walletManager = this.CreateWalletManager(dataFolder, "-walletaddressbuffer=30");
+            walletManager = this.CreateWalletManager(dataFolder, Network.Main, "-walletaddressbuffer=30");
             walletManager.Start();
 
             // Addresses populated to fill the buffer set
@@ -3180,13 +3180,13 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         }
 
         [Fact]
-        public void RecoverViaXpubkey_can_recover_wallet_without_mnemonic()
+        public void Recover_via_xpubkey_can_recover_wallet_without_mnemonic()
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
             const string stratisAccount0ExtPubKey = "xpub661MyMwAqRbcEgnsMFfhjdrwR52TgicebTrbnttywb9zn3orkrzn6MHJrgBmKrd7MNtS6LAim44a6V2gizt3jYVPHGYq1MzAN849WEyoedJ";
-            var walletManager = this.CreateWalletManager(dataFolder);
-            walletManager.RecoverWalletViaExtPubKey("testWallet", stratisAccount0ExtPubKey, 1, DateTime.Now.AddHours(-2));
+            var walletManager = this.CreateWalletManager(dataFolder, Network.Main);
+            walletManager.RecoverWallet("testWallet", ExtPubKey.Parse(stratisAccount0ExtPubKey), 0, DateTime.Now.AddHours(-2));
             
             var wallet = walletManager.LoadWallet("password", "testWallet");
 
@@ -3198,12 +3198,55 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
                 .Should().Be(stratisAccount0ExtPubKey);
         }
 
-        private WalletManager CreateWalletManager(DataFolder dataFolder, params string[] cmdLineArgs)
+        [Fact]
+        public void AddNewAccount_via_xpubkey_prevents_adding_an_account_as_an_existing_account_index()
         {
-            var nodeSettings = new NodeSettings(Network.RegTest, ProtocolVersion.PROTOCOL_VERSION, "StratisBitcoin", cmdLineArgs);
+            DataFolder dataFolder = CreateDataFolder(this);
+
+            const string stratisAccount0ExtPubKey = "xpub661MyMwAqRbcEgnsMFfhjdrwR52TgicebTrbnttywb9zn3orkrzn6MHJrgBmKrd7MNtS6LAim44a6V2gizt3jYVPHGYq1MzAN849WEyoedJ";
+            const string stratisAccount1ExtPubKey = "xpub6DGguHV1FQFPvZ5Xu7VfeENyiySv4R2bdd6VtvwxWGVTVNnHUmphMNgTRkLe8j2JdAv332ogZcyhqSuz1yUPnN4trJ49cFQXmEhwNQHUqk1";
+            var walletManager = this.CreateWalletManager(dataFolder, Network.StratisMain);
+            var wallet = walletManager.RecoverWallet("wallet1", ExtPubKey.Parse(stratisAccount0ExtPubKey), 0, DateTime.Now.AddHours(-2));
+
+            try
+            {
+                wallet.AddNewAccount(CoinType.Stratis, ExtPubKey.Parse(stratisAccount1ExtPubKey), 0, DateTime.Now.AddHours(-2));
+
+                Assert.True(false, "should have thrown exception but didn't.");
+            }
+            catch (WalletException e)
+            {
+                Assert.Equal("There is already an account in this wallet with index: " + 0, e.Message);
+            }
+        }
+
+        [Fact]
+        public void AddNewAccount_via_xpubkey_prevents_adding_the_same_xpub_key_as_different_account()
+        {
+            DataFolder dataFolder = CreateDataFolder(this);
+
+            const string stratisAccount0ExtPubKey = "xpub661MyMwAqRbcEgnsMFfhjdrwR52TgicebTrbnttywb9zn3orkrzn6MHJrgBmKrd7MNtS6LAim44a6V2gizt3jYVPHGYq1MzAN849WEyoedJ";
+            var walletManager = this.CreateWalletManager(dataFolder, Network.StratisMain);
+            var wallet = walletManager.RecoverWallet("wallet1", ExtPubKey.Parse(stratisAccount0ExtPubKey), 0, DateTime.Now.AddHours(-2));
+
+            try
+            {
+                wallet.AddNewAccount(CoinType.Stratis, ExtPubKey.Parse(stratisAccount0ExtPubKey), 1, DateTime.Now.AddHours(-2));
+
+                Assert.True(false, "should have thrown exception but didn't.");
+            }
+            catch (WalletException e)
+            {
+                Assert.Equal("There is already an account in this wallet with this xpubkey: " + stratisAccount0ExtPubKey, e.Message);
+            }
+        }
+
+        private WalletManager CreateWalletManager(DataFolder dataFolder, Network network, params string[] cmdLineArgs)
+        {
+            var nodeSettings = new NodeSettings(Network.RegTest, ProtocolVersion.PROTOCOL_VERSION, network.Name, cmdLineArgs);
             var walletSettings = new WalletSettings(nodeSettings);
 
-            return new WalletManager(this.LoggerFactory.Object, Network.Main, new ConcurrentChain(Network.Main),
+            return new WalletManager(this.LoggerFactory.Object, network, new ConcurrentChain(network),
                 NodeSettings.Default(), walletSettings, dataFolder, new Mock<IWalletFeePolicy>().Object, new Mock<IAsyncLoopFactory>().Object, new NodeLifetime(), DateTimeProvider.Default);
         }
 
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -3184,7 +3184,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             DataFolder dataFolder = CreateDataFolder(this);
 
             Wallet wallet = this.walletFixture.GenerateBlankWallet("testWallet", "password");
-            wallet.IsExtPubKeyWallet = "true";
+            wallet.IsExtPubKeyWallet = true;
             wallet.EncryptedSeed = null;
             wallet.ChainCode = null;
             
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -3178,6 +3178,24 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             Assert.Equal(30, hdAccount.InternalAddresses.Count);
         }
 
+        [Fact]
+        public void Xpubkey_only_wallet_can_load_without_private_key()
+        {
+            DataFolder dataFolder = CreateDataFolder(this);
+
+            Wallet wallet = this.walletFixture.GenerateBlankWallet("testWallet", "password");
+            wallet.IsExtPubKeyWallet = "true";
+            wallet.EncryptedSeed = null;
+            wallet.ChainCode = null;
+            
+            File.WriteAllText(Path.Combine(dataFolder.WalletPath, "testWallet.wallet.json"), JsonConvert.SerializeObject(wallet, Formatting.Indented, new ByteArrayConverter()));
+
+            var walletManager = new WalletManager(this.LoggerFactory.Object, Network.StratisMain, new Mock<ConcurrentChain>().Object, NodeSettings.Default(), new Mock<WalletSettings>().Object,
+                dataFolder, new Mock<IWalletFeePolicy>().Object, new Mock<IAsyncLoopFactory>().Object, new NodeLifetime(), DateTimeProvider.Default);
+
+            walletManager.LoadWallet("password", "testWallet");
+        }
+
         private WalletManager GetWalletManagerWithCustomConfigParam(DataFolder dataFolder, params string[] cmdLineArgs)
         {
             var nodeSettings = new NodeSettings(Network.RegTest, ProtocolVersion.PROTOCOL_VERSION, "StratisBitcoin", cmdLineArgs);
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
@@ -283,7 +283,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            var walletManager = GetWalletManagerWithCustomConfigParam(dataFolder, "-walletaddressbuffer=100");
+            var walletManager = this.CreateWalletManager(dataFolder, "-walletaddressbuffer=100");
 
             walletManager.CreateWallet("test", "mywallet", new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
@@ -3160,7 +3160,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            WalletManager walletManager = this.GetWalletManagerWithCustomConfigParam(dataFolder);
+            WalletManager walletManager = this.CreateWalletManager(dataFolder);
             walletManager.CreateWallet("test", "mywallet", new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
 
             // Default of 20 addresses becuause walletaddressbuffer not set
@@ -3169,7 +3169,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             Assert.Equal(20, hdAccount.InternalAddresses.Count);
             
             // Restart with walletaddressbuffer set
-            walletManager = this.GetWalletManagerWithCustomConfigParam(dataFolder, "-walletaddressbuffer=30");
+            walletManager = this.CreateWalletManager(dataFolder, "-walletaddressbuffer=30");
             walletManager.Start();
 
             // Addresses populated to fill the buffer set
@@ -3190,13 +3190,12 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             
             File.WriteAllText(Path.Combine(dataFolder.WalletPath, "testWallet.wallet.json"), JsonConvert.SerializeObject(wallet, Formatting.Indented, new ByteArrayConverter()));
 
-            var walletManager = new WalletManager(this.LoggerFactory.Object, Network.StratisMain, new Mock<ConcurrentChain>().Object, NodeSettings.Default(), new Mock<WalletSettings>().Object,
-                dataFolder, new Mock<IWalletFeePolicy>().Object, new Mock<IAsyncLoopFactory>().Object, new NodeLifetime(), DateTimeProvider.Default);
+            var walletManager = this.CreateWalletManager(dataFolder);
 
             walletManager.LoadWallet("password", "testWallet");
         }
 
-        private WalletManager GetWalletManagerWithCustomConfigParam(DataFolder dataFolder, params string[] cmdLineArgs)
+        private WalletManager CreateWalletManager(DataFolder dataFolder, params string[] cmdLineArgs)
         {
             var nodeSettings = new NodeSettings(Network.RegTest, ProtocolVersion.PROTOCOL_VERSION, "StratisBitcoin", cmdLineArgs);
             var walletSettings = new WalletSettings(nodeSettings);
--- a/src/Stratis.Bitcoin.Features.Wallet/CannotAddAccountToXpubKeyWalletException.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/CannotAddAccountToXpubKeyWalletException.cs
@@ -1,0 +1,11 @@
+﻿using System;
+
+namespace Stratis.Bitcoin.Features.Wallet
+{
+    public class CannotAddAccountToXpubKeyWalletException : Exception
+    {
+        public CannotAddAccountToXpubKeyWalletException(string message) : base(message)
+        {
+        }
+    }
+}
--- a/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
@@ -9,6 +9,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
+using NBitcoin.DataEncoders;
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.Features.Wallet.Broadcasting;
 using Stratis.Bitcoin.Features.Wallet.Helpers;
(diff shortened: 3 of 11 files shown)


src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
             }
         }
 
+        /// <summary>
+        /// Recovers a wallet using only the extended public key.
+        /// </summary>
+        /// <param name="request">The object containing the parameters used to recover a wallet.</param>
+        /// <returns></returns>
```
noescape00: redundant returns tag

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+            if (!this.ModelState.IsValid)
+            {
+                return BuildErrorResponse(this.ModelState);
+            }
+
+            try
+            {
+                string accountExtPubKey = this.SanitiseExtPubKey(request.ExtPubKey);
+
+                Wallet wallet = this.walletManager.RecoverWallet(request.Name, ExtPubKey.Parse(accountExtPubKey), request.AccountIndex, request.CreationDate);
```
noescape00: `Wallet wallet =` is redundant

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+
+                // start syncing the wallet from the creation date
+                this.walletSyncManager.SyncFromDate(request.CreationDate);
+
+                return this.Ok();
+            }
+            catch (WalletException e)
+            {
+                // indicates that this wallet already exists
```
noescape00: Capital letter and the dot in the end. Same on 292

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.NotFound, "Wallet not found.", e.ToString());
+            }
+            catch (Exception e)
+            {
+                this.logger.LogError("Exception occurred: {0}", e.ToString());
+                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, e.Message, e.ToString());
+            }
+        }
+
+        private string SanitiseExtPubKey(string requestExtPubKey)
```
noescape00: Please add summary

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+        {
+            if (this.network.IsBitcoin())
+                return requestExtPubKey;
+
+            byte[] stratisVersionBytes = network.GetVersionBytes(Base58Type.EXT_PUBLIC_KEY, true);
```
noescape00: Missing `.this`

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+        /// <summary>
+        /// Recovers a wallet using only the extended public key.
+        /// </summary>
+        /// <param name="request">The object containing the parameters used to recover a wallet.</param>
+        /// <returns></returns>
+        [Route("recover-via-extpubkey")]
+        [HttpPost]
+        public IActionResult RecoverViaExtPubKey([FromBody]WalletExtPubRecoveryRequest request)
+        {
+            Guard.NotNull(request, nameof(request));
```
noescape00: Missing entry and exit logs

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+            catch (Exception e)
+            {
+                this.logger.LogError("Exception occurred: {0}", e.ToString());
+                return ErrorHelpers.BuildErrorResponse(HttpStatusCode.BadRequest, e.Message, e.ToString());
+            }
+        }
+
+        private string SanitiseExtPubKey(string requestExtPubKey)
+        {
+            if (this.network.IsBitcoin())
```
noescape00: Missing entry and exit logs

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+                return requestExtPubKey;
+
+            byte[] stratisVersionBytes = network.GetVersionBytes(Base58Type.EXT_PUBLIC_KEY, true); 
+            byte[] ledgerVersionByte = new byte[] { (0x04), (0x88), (0xC2), (0x1E) };
+
+            byte[] decodedExtPubKey = Encoders.Base58Check.DecodeData(requestExtPubKey);
```
noescape00: Please add comments to the method body. The code is not trivial. 

Is it ledger & stratis specific? Will it work with non-stratis or non-ledger?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -253,6 +254,76 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+            byte[] extPubKeyVersionBytes = decodedExtPubKey.Take(4).ToArray();
+
+            if (extPubKeyVersionBytes.SequenceEqual(stratisVersionBytes))
+            {
+                return requestExtPubKey;
+            }
+            else if (extPubKeyVersionBytes.SequenceEqual(ledgerVersionByte))
+            {
+                byte[] vchData = decodedExtPubKey.Skip(4).ToArray();
```
noescape00: What is vchData and why 4 bytes are skipped? Please add comments.

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletManager.cs
```diff
@@ -66,6 +66,16 @@ public interface IWalletManager
         /// <returns>The recovered wallet.</returns>
         Wallet RecoverWallet(string password, string name, string mnemonic, DateTime creationTime, string passphrase = null);
 
+        /// <summary>
+        /// Recovers a wallet using extended public key and account number.
+        /// </summary>
+        /// <param name="name">The name of the wallet.</param>
+        /// <param name="extPubKey"></param>
```
noescape00: Empty tag

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -27,16 +27,22 @@ public Wallet()
         [JsonProperty(PropertyName = "name")]
         public string Name { get; set; }
 
+        /// <summary>
+        /// Flag indicating if it is a watch only wallet.
```
noescape00: But we already have a feature which is a watch only wallet. Should we remove this feature if it's logic is now here?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -180,6 +186,25 @@ public HdAccount AddNewAccount(string password, CoinType coinType, DateTimeOffse
             return accountRoot.AddNewAccount(password, this.EncryptedSeed, this.ChainCode, this.Network, accountCreationTime);
         }
 
+        /// <summary>
```
noescape00: summary is copy pasted from the method above. Parameter name `accountNumber` doesn't have matching xmldoc.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -380,6 +405,50 @@ public HdAccount AddNewAccount(string password, string encryptedSeed, byte[] cha
 
             return newAccount;
         }
+
+        /// <summary>
```
noescape00: Copy-pasted summary

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -380,6 +405,50 @@ public HdAccount AddNewAccount(string password, string encryptedSeed, byte[] cha
...
+        /// According to BIP44, an account at index (i) can only be created when the account at index (i - 1) contains transactions.
+        /// <seealso cref="https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki"/></remarks>
+        /// <param name="accountExtPubKey">The extended public key for the account.</param>
+        /// <param name="accountIndex">The zero-based account index.</param>
+        /// <param name="network">The network for which this account will be created.</param>
+        /// <param name="accountCreationTime">Creation time of the account to be created.</param>
+        /// <returns>A new HD account.</returns>
+        public HdAccount AddNewAccount(ExtPubKey accountExtPubKey, int accountIndex, Network network, DateTimeOffset accountCreationTime)
+        {
+            List<HdAccount> accounts = this.Accounts.ToList();
```
noescape00: ToList conversion is not needed here

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: cheers

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: @bokobza what does vch mean?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: I will try to find the answers to this - and I agree, the code needs a bit of improving for readability

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
mikedennis: I think it's hungarian for vector of characters (character array).

noescape00: @mikedennis is right. vch is hungarian notation for vector of char. But here instead of char or uchar you use bytes so the name is misleading.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: some other Post methods have it too, but I removed.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: same as method above but fixed here

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: Moving to its own class to capture the logic better

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: NA when moved

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: ok

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: Other methods in there don't have it, so do we really want to do it for these ones ?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: ok cool, I will correct it

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: No (discussed privately)

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: nice one

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: meaning rephrase to be more specific?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
noescape00: Nice 🙂

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: done

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
noescape00: Yep. Those methods are old and probably created before @Aprogiena introduced logging into the project. So we should start at some point here I guess.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -346,6 +347,42 @@ public Wallet RecoverWallet(string password, string name, string mnemonic, DateT
             return wallet;
         }
 
+        public Wallet RecoverWallet(string name, ExtPubKey extPubKey, int accountIndex, DateTime creationTime)
```
majikandy: missing summary, whoops

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+using NBitcoin;
+using NBitcoin.DataEncoders;
+
+namespace Stratis.Bitcoin.Features.Wallet.Controllers
+{
+    /// <summary>
+    /// Stratis X has a version prefix starting xq5h. Full node has a version prefix xpub for extended public key. 
+    /// 3rd parties like Ledger haven't update yet to the full node version prefix.
+    /// This class provides a way to convert from the legacy format to the current format.
+    /// <remarks>This class can be removed when the 3rd parties update their software.</remarks>
```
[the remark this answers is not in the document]
majikandy: Cheers

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public class LegacyExtPubKeyConverter
+    { 
+        /// <summary>
+        /// Converts a legacy Stratis format into a current Stratis format Base58 extended public key.
+        /// </summary>
+        /// <param name="extPubKey">The extended public key that may or may not need converting.</param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
majikandy: ok

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Stratis.Bitcoin.Features.Wallet.Controllers
+{
+    /// <summary>
+    /// Stratis X has a version prefix starting xq5h. Full node has a version prefix xpub for extended public key. 
+    /// 3rd parties like Ledger haven't update yet to the full node version prefix.
+    /// This class provides a way to convert from the legacy format to the current format.
+    /// </summary>
+    /// <remarks>This class can be removed when the 3rd parties update their software.</remarks>
+    public class LegacyExtPubKeyConverter
```
bokobza: 👍

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -72,6 +72,27 @@ public class WalletRecoveryRequest : RequestModel
...
+    public class WalletExtPubRecoveryRequest : RequestModel
+    {
+        [Required(ErrorMessage = "An extended public key is required.")]
+        public string ExtPubKey { get; set; }
+
+        [Required(ErrorMessage = "An account number is required. E.g. 0.")]
+        public int AccountIndex { get; set; }
+
+        public string FolderPath { get; set; }
+
```
bokobza: Please remove the extra lines here (2x).

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -380,6 +405,47 @@ public HdAccount AddNewAccount(string password, string encryptedSeed, byte[] cha
...
+        /// <param name="accountExtPubKey">The extended public key for the account.</param>
+        /// <param name="accountIndex">The zero-based account index.</param>
+        /// <param name="network">The network for which this account will be created.</param>
+        /// <param name="accountCreationTime">Creation time of the account to be created.</param>
+        /// <returns>A new HD account.</returns>
+        public HdAccount AddNewAccount(ExtPubKey accountExtPubKey, int accountIndex, Network network, DateTimeOffset accountCreationTime)
+        {
+            ICollection<HdAccount> accounts = this.Accounts;
+
+            if (accounts.Any(x => x.HdPath.Split('/').Last() == accountIndex + "'"))
```
bokobza: Maybe you could just check `if(accounts.Any(acc => acc.Index == accountIndex))`?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -432,7 +498,7 @@ public HdAccount()
         /// The list of external addresses, typically used for receiving money.
         /// </summary>
         [JsonProperty(PropertyName = "externalAddresses")]
-        public ICollection<HdAddress> ExternalAddresses { get; set; }
+        public ICollection<HdAddress> ExternalAddresses { get ; set; }
```
bokobza: extra space added by mistake after `get`.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -355,6 +392,11 @@ public HdAccount GetUnusedAccount(string walletName, string password)
 
             Wallet wallet = this.GetWalletByName(walletName);
 
+            if (wallet.IsExtPubKeyWallet)
+            {
+                throw new CannotAddAccountToXpubKeyWalletException("Use recover-via-extpubkey instead.");
```
bokobza: Excellent.
Maybe just throw a `WalletException` here (ref my previous comment).

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
```diff
@@ -286,9 +287,9 @@ public void CreateWalletWithWalletSetting100UnusedAddressBufferCreates100Address
         {
             DataFolder dataFolder = CreateDataFolder(this);
 
-            var walletManager = GetWalletManagerWithCustomConfigParam(dataFolder, "-walletaddressbuffer=100");
+            var walletManager = this.CreateWalletManager(dataFolder, Network.Main, "-walletaddressbuffer=100");
 
-            walletManager.CreateWallet("test", "mywallet", new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
+            walletManager.CreateWallet("test", "mywallet", passphrase: new Mnemonic(Wordlist.English, WordCount.Eighteen).ToString());
```
bokobza: Here and in a few other places, the mnemonic is passed as the passphrase..

src/Stratis.Bitcoin.Features.Wallet/CannotAddAccountToXpubKeyWalletException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Stratis.Bitcoin.Features.Wallet
+{
+    public class CannotAddAccountToXpubKeyWalletException : Exception
+    {
+        public CannotAddAccountToXpubKeyWalletException(string message) : base(message)
```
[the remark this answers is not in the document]
majikandy: Well, I used this specific one in order to catch it explicitly and then throw the wallet exception from the controller

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
majikandy: as per above, I wouldn't want to catch and then interrogate the message when building the error response - what do you think? Seems to be logical to throw the explicit as it is caught and the walletexception is the one that goes to the client caller

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
majikandy: I need to double check these as it happened accidentally when I did a resharper refactor

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: yes :) much better

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
majikandy: It does suggest here that it would be better if the CreateWallet on the WalletManager accepted a Mnemonic type rather than a string for the mnemonic

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
```diff
@@ -274,12 +274,12 @@ public void CreateWalletWithMnemonicListCreatesWalletUsingMnemonicList()
 
             string password = "test";
 
-            var mnemonicList = new Mnemonic(Wordlist.French, WordCount.Eighteen);
+            var mnemonic = new Mnemonic(Wordlist.French, WordCount.Eighteen);
 
             // create the wallet
-            Mnemonic mnemonic = walletManager.CreateWallet(password, "mywallet", mnemonicList: mnemonicList.ToString());
+            var returnedMnemonic = walletManager.CreateWallet(password, "mywallet", null, mnemonic);
```
bokobza: Can you please change var to Mnemonic here? 
It's already not very intuitive that a CreateWallet method shall return a mnemonic.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: ah it's because you want to return a HTTP (403) Forbidden. Fair enough.

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
```diff
@@ -3178,12 +3178,74 @@ public void Start_takes_account_of_address_buffer_even_for_existing_wallets()
...
+
+            const string stratisAccount0ExtPubKey = "xpub661MyMwAqRbcEgnsMFfhjdrwR52TgicebTrbnttywb9zn3orkrzn6MHJrgBmKrd7MNtS6LAim44a6V2gizt3jYVPHGYq1MzAN849WEyoedJ";
+            var walletManager = this.CreateWalletManager(dataFolder, Network.StratisMain);
+            var wallet = walletManager.RecoverWallet("wallet1", ExtPubKey.Parse(stratisAccount0ExtPubKey), 0, DateTime.Now.AddHours(-2));
+
+            try
+            {
+                wallet.AddNewAccount(CoinType.Stratis, ExtPubKey.Parse(stratisAccount0ExtPubKey), 1, DateTime.Now.AddHours(-2));
+
+                Assert.True(false, "should have thrown exception but didn't.");
```
noescape00: Use Assert.Throws()

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+    { 
+        /// <summary>
+        /// Converts a legacy Stratis format into a current Stratis format Base58 extended public key.
+        /// </summary>
+        /// <param name="extPubKey">The extended public key that may or may not need converting.</param>
+        /// <returns>The same extended public key if version bytes is already the Stratis full node one,
+        /// or The corrected extended public key if the version bytes was the Stratis X one.</returns>
+        public static string ConvertIfInLegacyStratisFormat(string extPubKey, Network network)
```
noescape00: no xml tag for the network

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
noescape00: Still missing entry and exit logs

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -180,6 +186,25 @@ public HdAccount AddNewAccount(string password, CoinType coinType, DateTimeOffse
...
+        /// The name given to the account is of the form "account (i)" by default, where (i) is an incremental index starting at 0.
+        /// According to BIP44, an account at index (i) can only be created when the account at index (i - 1) contains at least one transaction.
+        /// </remarks>
+        /// <seealso cref="https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki"/>
+        /// <param name="extPubKey">The extended public key for the wallet<see cref="EncryptedSeed"/>.</param>
+        /// <param name="coinType">The type of coin this account is for.</param>
+        /// <param name="accountIndex">Zero-based index of the account to add.</param>
+        /// <param name="accountCreationTime">Creation time of the account to be created.</param>
+        /// <returns>A new HD account.</returns>
+        public HdAccount AddNewAccount(CoinType coinType, ExtPubKey extPubKey, int accountIndex, DateTimeOffset accountCreationTime)
```
noescape00: This method has copypasted xmldoc from the method above. Use inheritdoc + custom param tags

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -345,6 +346,42 @@ public Wallet RecoverWallet(string password, string name, string mnemonic, DateT
...
+            this.logger.LogTrace("(-)");
+            return wallet;
+        }
+
+        /// <inheritdoc />
+        public Wallet RecoverWallet(string name, ExtPubKey extPubKey, int accountIndex, DateTime creationTime)
+        {
+            Guard.NotEmpty(name, nameof(name));
+            Guard.NotNull(extPubKey, nameof(extPubKey));
+            this.logger.LogTrace("({0}:'{1}')", nameof(name), name);
```
noescape00: Entry log should log all input parameters

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1230,6 +1273,34 @@ private Wallet GenerateWalletFile(string name, string encryptedSeed, byte[] chai
             return walletFile;
         }
 
+        private Wallet GenerateWalletFileXpub(string name, DateTimeOffset? creationTime = null)
+        {
+            Guard.NotEmpty(name, nameof(name));
+            this.logger.LogTrace("({0}:'{1}')", nameof(name), name);
```
noescape00: entry log is incomplete

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -276,6 +347,13 @@ private void the_block_should_contain_the_transaction()
                 .Should().Be(this.transaction.GetHash().ToString());
         }
 
+        private void it_is_rejected_and_user_is_told_to_restore_instead()
+        {
+            this.postResponse.StatusCode.Should().Be(StatusCodes.Status403Forbidden);
+
+            //TODO: this.postResponseText.Should().Be("Use recover - via - extpubkey instead.");
```
noescape00: do not leave uncommented pieces of code.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: ok I added this time

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletManagerTest.cs
majikandy: I don't think you can check the message within the exception very easily with that. I can use Fluent Assertions to do it but the rest of the file wasn't using it so I avoided

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
majikandy: ok,

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
majikandy: you mean commented, and I agree.

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Stratis.Bitcoin.Features.Wallet.Controllers
+{
+    /// <summary>
+    /// Stratis X has a version prefix starting xq5h. Full node has a version prefix xpub for extended public key. 
+    /// 3rd parties like Ledger haven't update yet to the full node version prefix.
+    /// This class provides a way to convert from the legacy format to the current format.
+    /// </summary>
+    /// <remarks>This class can be removed when the 3rd parties update their software.</remarks>
+    public class LegacyExtPubKeyConverter
```
noescape00: What's the reason for having this class to be static?

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            if (IsLegacyStratisExtpubKey(extPubKeyBytes))
+            {
+                return ReplaceBase58Prefix(extPubKeyBytes, network.GetVersionBytes(Base58Type.EXT_PUBLIC_KEY, true));
+            }
+
+            throw new FormatException($"ExtPubKey {extPubKey} could not be parsed.");
+        }
+
+        private static bool IsStratisExtPubKey(byte[] extPubKey, byte[] stratisVersionBytes)
```
noescape00: Please add comments and logs to methods

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            byte[] version = extPubKey.Take(4).ToArray();
+            return version.SequenceEqual(stratisVersionBytes);
+        }
+
+        private static bool IsLegacyStratisExtpubKey(byte[] extPubKey)
+        {
+            var legacyStratisVersionBytes = new byte[] { (0x04), (0x88), (0xC2), (0x1E) };
```
noescape00: to constant

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                return ReplaceBase58Prefix(extPubKeyBytes, network.GetVersionBytes(Base58Type.EXT_PUBLIC_KEY, true));
+            }
+
+            throw new FormatException($"ExtPubKey {extPubKey} could not be parsed.");
+        }
+
+        private static bool IsStratisExtPubKey(byte[] extPubKey, byte[] stratisVersionBytes)
+        {
+            byte[] version = extPubKey.Take(4).ToArray();
```
noescape00: 4 to constant

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+    public class LegacyExtPubKeyConverter
+    {
+        /// <summary>
+        /// Converts a legacy Stratis format into a current Stratis format Base58 extended public key.
+        /// </summary>
+        /// <param name="extPubKey">The extended public key that may or may not need converting.</param>
+        /// <param name="network">The network to get the version bytes for.</param>
+        /// <returns>The same extended public key if version bytes is already the Stratis full node one,
+        /// or The corrected extended public key if the version bytes was the Stratis X one.</returns>
+        public static string ConvertIfInLegacyStratisFormat(string extPubKey, Network network)
```
noescape00: Please add a unit test

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <remarks>This class can be removed when the 3rd parties update their software.</remarks>
+    public class LegacyExtPubKeyConverter
+    {
+        /// <summary>
+        /// Converts a legacy Stratis format into a current Stratis format Base58 extended public key.
+        /// </summary>
+        /// <param name="extPubKey">The extended public key that may or may not need converting.</param>
+        /// <param name="network">The network to get the version bytes for.</param>
+        /// <returns>The same extended public key if version bytes is already the Stratis full node one,
+        /// or The corrected extended public key if the version bytes was the Stratis X one.</returns>
```
noescape00: The => the

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+        public static string ConvertIfInLegacyStratisFormat(string extPubKey, Network network)
+        {
+            byte[] stratisVersionBytes = network.GetVersionBytes(Base58Type.EXT_PUBLIC_KEY, true);
```
noescape00: What if the network is not stratis? Add to param xmldoc that only stratis network is allowed or add a check that the network is stratis indeed.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -251,6 +253,63 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+        [Route("recover-via-extpubkey")]
+        [HttpPost]
+        public IActionResult RecoverViaExtPubKey([FromBody]WalletExtPubRecoveryRequest request)
+        {
+            Guard.NotNull(request, nameof(request));
+            this.logger.LogTrace("({0}.{1}:'{2}')", nameof(request), nameof(request.Name), request.Name);
+
+            if (!this.ModelState.IsValid)
+            {
+                return BuildErrorResponse(this.ModelState);
```
noescape00: Exit logs should be with every return. Otherwise they don't make any sense

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -380,6 +405,47 @@ public HdAccount AddNewAccount(string password, string encryptedSeed, byte[] cha
...
+        /// Adds an account to the current account root using extended public key.
+        /// </summary>
+        /// <param name="accountExtPubKey">The extended public key for the account.</param>
+        /// <param name="accountIndex">The zero-based account index.</param>
+        /// <param name="network">The network for which this account will be created.</param>
+        /// <param name="accountCreationTime">Creation time of the account to be created.</param>
+        /// <returns>A new HD account.</returns>
+        public HdAccount AddNewAccount(ExtPubKey accountExtPubKey, int accountIndex, Network network, DateTimeOffset accountCreationTime)
+        {
+            ICollection<HdAccount> accounts = this.Accounts;
```
noescape00: I can see a race condition here. 

So if we look at how it's handled on 379- there is a safeguard against a race condition when this method is executed twice in parallel. 

In your case you'll be getting an exception: collection was modified during the enumaration from time to time. 

If you have time: please introduce a lock to the wallet. If not- do a quickfix like on 379

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -345,6 +346,43 @@ public Wallet RecoverWallet(string password, string name, string mnemonic, DateT
...
+            return wallet;
+        }
+
+        /// <inheritdoc />
+        public Wallet RecoverWallet(string name, ExtPubKey extPubKey, int accountIndex, DateTime creationTime)
+        {
+            Guard.NotEmpty(name, nameof(name));
+            Guard.NotNull(extPubKey, nameof(extPubKey));
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}',{4}:'{5}')", nameof(name), name, nameof(extPubKey), extPubKey, nameof(accountIndex), accountIndex);
+
```
noescape00: Redundant like

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: It is covered within the Wallet Manager tests, there is no need for explicit unit test

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: oh cool, good spot - let me look at it

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: yep ok

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: This is a private method with a specific name to express intention, adding comments would be redundant.

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
bokobza: This is called only when the network is Stratis:
```
string accountExtPubKey = this.network.IsBitcoin()
                        ? request.ExtPubKey
                        : LegacyExtPubKeyConverter.ConvertIfInLegacyStratisFormat(request.ExtPubKey, this.network);
```

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
bokobza: No need for that, we plan to remove this method asap.

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
bokobza: No need for that, we plan to remove this method asap.

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: The method is static because it was just manipulating data, it could just as easily be a string extension

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: we could put that protection inside the static method too if we want to be safe, but it wouldn't really matter if called with a bitcoin one would it? as it doesn't match anyway

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
majikandy: I'm not familiar with how customparam works - is there an example somewhere?

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
noescape00: Ah, ok. In case this logic will be removed anyway- ignore comments on stuff inside this class, no need to waste time on smth that will be removed soon anyway

bokobza:
For the lock, look at how we use the wallet lock in WalletManager.GetUnusedAccount()

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -251,6 +253,65 @@ public IActionResult Recover([FromBody]WalletRecoveryRequest request)
...
+        [Route("recover-via-extpubkey")]
+        [HttpPost]
+        public IActionResult RecoverViaExtPubKey([FromBody]WalletExtPubRecoveryRequest request)
+        {
+            Guard.NotNull(request, nameof(request));
+            this.logger.LogTrace("({0}.{1}:'{2}')", nameof(request), nameof(request.Name), request.Name);
+
+            if (!this.ModelState.IsValid)
+            {
+                this.logger.LogTrace("(-)");
```
noescape00: Inside the same method the exit logs should be different to be able to distinguish execution path based on the exit log

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -390,15 +390,15 @@ public Wallet RecoverWallet(string name, ExtPubKey extPubKey, int accountIndex,
             return wallet;
         }
 
-        private HdAccount AddNewAccount(ExtPubKey extPubKey, int accountIndex, Wallet wallet)
+        private HdAccount AddNewAccountLock(ExtPubKey extPubKey, int accountIndex, Wallet wallet)
```
noescape00: AddNewAccountLock => AddNewAccountLocked

majikandy: I just copied the pattern of the other ones in the file which had Lock

noescape00: Ok, didn't see them. 

Please change other ones as well to have Locked in the end (that's the standard naming policy for methods that should be locked and it is used everywhere in the codebase. Looks like this class got inconsistent)

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
majikandy: ok

src/Stratis.Bitcoin.Features.Wallet/Controllers/LegacyExtPubKeyConverter.cs
majikandy: ok

majikandy:
I would say this is ready now to merge ?

PR #1653, opened 2018-07-13
Consensus manager behavior (CHB refactoring)

noescape00:
Reimplemented CHB and connected it to CM.


src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+        public async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync(ChainedHeader newTip)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(newTip), newTip);
+
+            ConnectNewHeadersResult result = null;
+            bool syncRequired = false;
+
+            using (await this.asyncLock.LockAsync().ConfigureAwait(false))
+            {
+                if (this.cachedHeaders.Count != 0)
```
fassadlr: There are too many nested ifs here. I would return straight here by checking this.cachedHeaders.Any():

```
if (!this.cachedHeaders.Any())
return;
```

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    try
+                    {
+                        result = await this.PresentHeadersLockedAsync(this.cachedHeaders, false).ConfigureAwait(false);
+
+                        if (result != null)
```
fassadlr: The same here:

```
 if (result != null)
    return;
```

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
fassadlr: Also, why bother opening a `using` statement when the cached header count is 0? `syncRequired` will be false anyway.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                this.logger.LogTrace("(-)[IGNORE_ON_IBD]");
+                return;
+            }
+
+            var headers = new HeadersPayload();
+            ChainedHeader consensusTip = this.consensusManager.Tip;
+
+            ChainedHeader fork = this.chain.FindFork(getHeadersPayload.BlockLocators);
+            if (fork != null)
```
fassadlr: Too many nested `if`s here, it reads badly. I would separate out into methods

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+        /// of our best chain's tip, we update our view of the best chain to that tip.
+        /// </para>
+        /// </remarks>
+        private async Task ProcessHeadersAsync(INetworkPeer peer, HeadersPayload headersPayload)
+        {
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(peer), peer.RemoteSocketEndpoint, nameof(headersPayload), headersPayload);
+
+            List<BlockHeader> headers = headersPayload.Headers;
+
+            if (headers.Count == 0)
```
fassadlr: `!headers.Any()`

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                this.logger.LogTrace("Headers payload with no headers was received. Assuming we're synced with the peer.");
+                this.logger.LogTrace("(-)[NO_HEADERS]");
+                return;
+            }
+
+            // Check headers for consecutiveness.
+            for (int i = 1; i < headers.Count; ++i)
+            {
+                if (headers[i].HashPrevBlock != headers[i - 1].GetHash())
```
fassadlr: Cant you rewrite this into a `foreach` and include the condition into the `where(x=>x...)`?

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+
+            if (headers.Count == 0)
+            {
+                this.logger.LogTrace("Headers payload with no headers was received. Assuming we're synced with the peer.");
+                this.logger.LogTrace("(-)[NO_HEADERS]");
+                return;
+            }
+
+            // Check headers for consecutiveness.
+            for (int i = 1; i < headers.Count; ++i)
```
fassadlr: Separate method

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+                if (headers[i].HashPrevBlock != headers[i - 1].GetHash())
+                {
+                    this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, "Peer presented nonconsecutive headers.");
+
+                    this.logger.LogTrace("(-)[PEER_BANNED]");
+                    return;
+                }
+            }
+
+            using (await this.asyncLock.LockAsync().ConfigureAwait(false))
```
fassadlr: Separate method.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                if (this.cachedHeaders.Count > CacheSyncHeadersThreshold)
+                {
+                    // Ignore this message because cache is full.
+                    return;
+                }
+
+                // If queue is not empty- add to queue instead of calling CM.
+                if (this.cachedHeaders.Count != 0)
```
fassadlr: .Any()

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    this.cachedHeaders.AddRange(headers);
+
+                    this.logger.LogTrace("{0} headers were added to cache.", headers.Count);
+                }
+                else
```
fassadlr: No need for else here, can just return.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    ConnectNewHeadersResult result = await this.PresentHeadersLockedAsync(headers).ConfigureAwait(false);
+
+                    if (result.Consumed == null)
+                    {
+                        this.cachedHeaders.AddRange(headers);
+                        this.logger.LogTrace("None of {0} headers were consumed, all were added to cache.", headers.Count);
+                    }
+                    else
```
fassadlr: Same here, no need for else

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,434 @@
...
+                    {
+                        this.ExpectedPeerTip = result.Consumed;
+
+                        if (result.Consumed.Header != headers.Last())
```
fassadlr: Nested `if'`s, I feel this can be written better.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: >Also, why bother opening a using statement when the cached header count is 0? 

Because `cachedHeaders` should be accessed under the lock

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Used to be like that, @Aprogiena insisted on doing it the way it is right now. I don't mind.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: LINQ is bad 🙂 
I'm trying to avoid it as much as possible. 

Here it will do more work that it's supposed to so `.Count == 0` is cheaper

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Here we have only BlockHeader instances so there is no way to access prev header inside `where(x=> x...)`

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: That will save only 2 lines here but will spawn 10+ lines for a separated method.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: LINQ is bad

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Fixed

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Fixed

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Fixed

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
dangershony: >LINQ is bad 
No LINQ is not bad generally, but I understand that in consensus code we want to avoid LINQ as a rule.

noescape00: I'll rephrase: LINQ is not bad, it's a gun with which you can shoot yourself in the leg. If you use it appropriately- it's good. If you are using it everywhere where it can be used- it's highly less efficient in comparison to if you'd write it yourself without linq.

noescape00:
In case this will be approved today: don't merge. 
Me and @Aprogiena will go over this tmr, we didn't finish today.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Here with Any it's not the worst case, but for example: 

https://referencesource.microsoft.com/#system.core/system/linq/Enumerable.cs

```
public static bool Any<TSource>(this IEnumerable<TSource> source) {
    if (source == null) throw Error.ArgumentNull("source");
    using (IEnumerator<TSource> e = source.GetEnumerator()) {
        if (e.MoveNext()) return true;
    }
    return false;
}
```

This guy will check if source is null. But I don't need it- I already know it's not null. Also it will create a disposable object, do smth with it and dispose it later. Kinda too much comparing to checking if count is 0.

Aprogiena: > !headers.Any()

no - https://stackoverflow.com/questions/5741617/listt-any-or-count 
https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0

+ LINQ is bad

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
Aprogiena: Probably he meant `result == null`?

noescape00: I can change it in order to return if the result is null. @Aprogiena are you ok with this?

I remember that's the place where I used to have it but then we changed it earlier today.

Aprogiena: The original you had wasn't `== null`, I would be okay with `== null`.

noescape00: Fixed

noescape00:
PR is finished, we've applied the last changes today.

@fassadlr please review the changes and let me know if you're ok with them.

Aprogiena:
Looks good to me, no testing though

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,441 @@
...
+        /// <remarks>
+        /// "getheaders" message is sent by the peer in response to "headers" message until an empty array is received.
+        /// <para>
+        /// This payload notifies peers of our current best validated height, which is held by consensus tip.
+        /// </para>
+        /// <para>
+        /// If the peer is behind/equal to our best height an empty array is sent back.
+        /// </para>
+        /// </remarks>
+        private async Task ProcessGetHeadersAsync(INetworkPeer peer, GetHeadersPayload getHeadersPayload)
```
fassadlr: @noescape00 thanks for making the changes thus far, it looks a lot better. However, this method along with `ProcessHeadersAsync` violates SRP (Single Responsilibity Principle). To me, this method does 3 things;

1. Check to see that we are in IBD
2. Construct a list of headers to send
3. Send headers to the peer

This in itself warrants 3 separate  methods to handle individual logic as SRP states that a class or method should only ever be responsible for one thing and one thing only.

Its either that or the consumer of this method (`OnMessageReceived)` should do the IBD check upfront before calling the method. But the piece of logic that creates the set of headers should definitly be separate piece of logic that can be tested on its own.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,441 @@
...
+        /// <remarks>
+        /// "headers" message is sent in response to "getheaders" message or it is solicited
+        /// by the peer when a new block is validated (unless in IBD).
+        /// <para>
+        /// When we receive "headers" message from the peer, we can adjust our knowledge
+        /// of the peer's view of the chain. We update its pending tip, which represents
+        /// the tip of the best chain we think the peer has.
+        /// </para>
+        /// </remarks>
+        private async Task ProcessHeadersAsync(INetworkPeer peer, HeadersPayload headersPayload)
```
fassadlr: @noescape00  as per the above comment I can't approve this method as it violates SRP. We always need to keep in mind how much a particular  piece of ocde is doing and in this case I can see:

1. If check to see if header count is 0 (can be moved to the caller)
2. If check to see if header count is less than MaxItemsPerHeadersMessage (can be moved to the caller)
3. Check to see if headers are in order (separate method, which should be unit tested)
4. Then a lock that does more checks and eventually, given the right scenario, resyncs. (separate method).

All of these needs to either be in separate methods or decisive logic (ifs) moved to the caller.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
Aprogiena: The IBD check does not count because it is a guard.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
Aprogiena: The guards totally belong here, not to the caller.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: Fixed

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
noescape00: First part of the method consists of guards, it's just there are a lot of them but they totally belong here. 

And after that we have a method with a single responsibility so it doesn't violate SRP

noescape00:
> I would suggest that you have a quick read up on basic SOLID principles, especially SRP.

I know them, I've read all the nice books like clean code, books about SOLID, SRP, band of 4 and stuff like that. I just not always agree. Those are guidelines that shouldn't be blindly followed.

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
fassadlr: @Aprogiena good point on the Guard! @noescape00 good job, this method looks way better now 👍  Getting there!

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,461 @@
...
+            // Ignoring "getheaders" from peers because node is in initial block download unless the peer is whitelisted.
+            // We don't want to reveal our position in IBD which can be used by attacker. Also we don't won't to deliver peers any blocks
+            // because that will slow down our own syncing process.
+            if (this.initialBlockDownloadState.IsInitialBlockDownload() && !peer.Behavior<ConnectionManagerBehavior>().Whitelisted)
+            {
+                this.logger.LogTrace("(-)[IGNORE_ON_IBD]");
+                return;
+            }
+
+            HeadersPayload headersPayload = this.ConstructHeadersPayload(getHeadersPayload.BlockLocator, getHeadersPayload.HashStop);
```
fassadlr: 💯

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,461 @@
...
+            }
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <summary>Constructs the headers from locator to consensus tip.</summary>
+        /// <param name="locator">Block locator.</param>
+        /// <param name="hashStop">Hash of the block after which constructing headers payload should stop.</param>
+        /// <returns><see cref="HeadersPayload"/> with headers from locator towards consensus tip or <c>null</c> in case locator was invalid.</returns>
+        private HeadersPayload ConstructHeadersPayload(BlockLocator locator, uint256 hashStop)
```
fassadlr: NICE

src/Stratis.Bitcoin/Base/ConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,461 @@
...
+            {
+                this.logger.LogTrace("Headers payload with {0} headers was received. Protocol violation. Banning the peer.", headers.Count);
+                this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, "Protocol violation.");
+
+                this.logger.LogTrace("(-)[TOO_MANY_HEADERS]");
+                return;
+            }
+
+            // Check headers for consecutiveness.
+            for (int i = 1; i < headers.Count; i++)
```
fassadlr: This should still go in a separate method as it should be unit tested. Looks like a rule.

fassadlr:
```
if(!CheckHeadersForConsecutiveness())
    return;
```

fassadlr: Fix this and we are golden 👍

fassadlr: How would moving the logic to a separate method violate single responsibility?

Hmm you do however raise an interesting point re banning and disconnection... You can actually move both those `if` that potentially ban and disconnect the peer to method:

```
if(CheckIfPeerShouldBeBanned(headers))
    return;
```

noescape00: Good idea. Moved checks that may load to peer banning to a separated method.

Aprogiena:
Looks good, happy with that

noescape00:
@fassadlr added some more logging so the last commit resetted the approval. Can you reapprove please?

fassadlr:
@noescape00 just waiting for AppVeyor to build and then we merge :)

dangershony:
Hi I wanted to review 😞 
No matter I will do it now.

noescape00:
@dangershony write here if you find anything. I'll implement the fixes in the next PR which will implement unit tests for CMB

PR #1654, opened 2018-07-13
Fix compiler warnings and new Block() usage

fassadlr:
This should be a pretty easy PR to review.

We are now down to 25 compiler warnings.

### DIFF
--- a/src/NBitcoin.Tests/ChainTests.cs
+++ b/src/NBitcoin.Tests/ChainTests.cs
@@ -405,7 +405,7 @@ namespace NBitcoin.Tests
         [Trait("UnitTest", "UnitTest")]
         public void CanForkSidePartialChain()
         {
-            Block genesis = TestUtils.CreateFakeBlock();
+            Block genesis = TestUtils.CreateFakeBlock(this.network);
             var side = new ConcurrentChain(this.network, new ChainedHeader(genesis.Header, genesis.GetHash(), 0));
             var main = new ConcurrentChain(this.network, new ChainedHeader(genesis.Header, genesis.GetHash(), 0));
             this.AppendBlock(side, main);
@@ -488,10 +488,11 @@ namespace NBitcoin.Tests
             ChainedHeader block = mainTip.GetAncestor(branchLength - 1);
             for (int i = 0; i < branchLength; i++)
             {
-                Block newBlock = TestUtils.CreateFakeBlock();
+                Block newBlock = TestUtils.CreateFakeBlock(this.network);
                 newBlock.Header.HashPrevBlock = block.Header.GetHash();
                 block = new ChainedHeader(newBlock.Header, newBlock.Header.GetHash(), block);
             }
+
             ChainedHeader branchTip = block;
 
             // Test 100 random starting points for locators.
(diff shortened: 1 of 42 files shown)


src/NBitcoin.Tests/ChainTests.cs
```diff
@@ -554,7 +555,7 @@ public ChainedHeader AppendBlock(ChainedHeader previous, params ConcurrentChain[
             uint nonce = RandomUtils.GetUInt32();
             foreach (ConcurrentChain chain in chains)
             {
-                Block block = TestUtils.CreateFakeBlock(this.network.CreateTransaction());
+                Block block = TestUtils.CreateFakeBlock(this.network, this.network.CreateTransaction());
```
bokobza: you can just call `TestUtils.CreateFakeBlock(this.network);` here.

src/NBitcoin.Tests/TestUtils.cs
```diff
@@ -32,27 +20,28 @@ internal static byte[] ParseHex(string data)
...
             block.AddTransaction(tx);
             block.UpdateMerkleRoot();
             return block;
         }
 
-        public static Block CreateFakeBlock()
+        public static Block CreateFakeBlock(Network network)
         {
-            Block block = CreateFakeBlock(new Transaction());
+            Block block = CreateFakeBlock(network, network.CreateTransaction());
```
bokobza: you can just call `CreateFakeBlock(this.network);` here.

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -85,8 +84,8 @@ public void PeerConnectorAddNode_CanAlwaysStart()
             var nodeSettings = new NodeSettings();
 
             var connectionSettings = new ConnectionManagerSettings(nodeSettings);
-
-            var connector = new PeerConnectorAddNode(this.asyncLoopFactory, DateTimeProvider.Default, this.extendedLoggerFactory, this.network, this.networkPeerFactory, this.nodeLifetime, nodeSettings, connectionSettings, peerAddressManager, new SelfEndpointTracker());
+            var networkPeerFactory = new Mock<INetworkPeerFactory>();
```
bokobza: How come this changed?

src/NBitcoin.Tests/ChainTests.cs
fassadlr: Fixed :)

src/NBitcoin.Tests/TestUtils.cs
fassadlr: Not with this one, as you will end up recursively calling the same method? There is a `public static Block CreateFakeBlock(Network network, Transaction tx)` above.

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
fassadlr: There was a global this.networkPeerFactory field in the class that wasn't used, so I implemented them in each test.

src/NBitcoin.Tests/TestUtils.cs
bokobza: ah yes lol, what was I thinking.

PR #1656, opened 2018-07-13
Role back a commit

dangershony:
@monsieurleberre please make sure you run integration tests next time.


monsieurleberre:
I have another one pending that fixes the issue coming with this

dangershony:
Ok but this needs to be rolled back first to fix the tests.

dangershony:
@Aprogiena thanks, this was pushed and tests are failing the fix is to change the test, but we need to look at that more in depth.

PR #1658, opened 2018-07-16
[Integration Tests] Remove GenerateStratis

fassadlr:
This method is now an extension on the CoreNode as there are valid scenarios where want to inject transactions into block generation.


src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
```diff
@@ -69,8 +69,8 @@ public void WalletCanReceiveAndSendCorrectly()
                 Assert.Null(stratisReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet("mywallet").First().Transaction.BlockHeight);
 
                 // generate two new blocks do the trx is confirmed
-                stratisSender.GenerateStratis(1, new List<Transaction>(new[] { stratisSender.FullNode.Network.CreateTransaction(trx.ToBytes()) }));
-                stratisSender.GenerateStratis(1);
+                stratisSender.GenerateBlocks(1, new List<Transaction>(new[] { stratisSender.FullNode.Network.CreateTransaction(trx.ToBytes()) }));
```
bokobza: what's the difference between the two methods?
If I want to write a test, which one should I use?

fassadlr: @bokobza this is a tricky one, there are 6 tests (I think) that use this from a valid perspective. But they are from 2 test classes... One option would be to make those methods internal to the class so that other devs can't use them? We could properly comment them maybe?

dangershony: `GenerateStratisWithMiner` uses the actual mining feature `GenerateBlocks` mines using test code that's why I wanted to get rid of it.

bokobza: It's a bit strange because the two methods do almost the same thing, but in completely different ways.

bokobza: Even if we prevent the devs to use this, it's not great.
Maybe a solution would be to throw trxs in the mempool first and then call `GenerateStratisWithMiner`.
If there is nothing we can do then at least rename GenerateBlocks (maybe back to GenerateStratis) and add some comments on it.

fassadlr:
> Maybe a solution would be to throw trxs in the mempool first and then call GenerateStratisWithMiner

This might be a great solution! Let me investigate 👍

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
fassadlr: >   Maybe a solution would be to throw trxs in the mempool first and then call GenerateStratisWithMiner

This might be a great solution! Let me investigate 👍

bokobza: @dangershony yeah sure but these are called next to each other, it's a bit confusing.

bokobza: @fassadlr and it's more integration than constructing blocks manually. 🐱‍👓

dangershony: Ah yeah I did not realise its a rename, indeed I agree we should delete this method all together and chuck trx in mempool

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -45,5 +48,98 @@ public static int GetApiPort(this CoreNode coreNode)
         {
             return coreNode.FullNode.NodeService<ApiSettings>()?.ApiPort ?? -1;
         }
+
+        public static uint256[] GenerateBlocks(this CoreNode coreNode, int blocksToGenerate)
+        {
+            return GenerateBlocks(coreNode, blocksToGenerate, new List<Transaction>());
+        }
+
+        public static uint256[] GenerateBlocks(this CoreNode coreNode, int blocksToGenerate, List<Transaction> transactions)
```
dangershony: As suggested push trx to mempool and call `GenerateWithMiner()`

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
fassadlr: This is actually a tricky one as we need to somehow inject a fake block with double spend transactions into the puller so that consensus can reject it. If we add a tx to the mempool and try and a double spend it, it breaks (correctly) So the test's assert then needs to be changed OR we move that method that creates the double spend blocks into the class.

fassadlr: Look at `TxMempoolBlockDoublespend`

fassadlr:
@bokobza @dangershony Actually, I just noticed that we have 2 integration tests that test double spending already:

`SendingTransactionWithDoubleSpend`

So I recommend we remove that unit test?

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
dangershony: As agreed keep this method for special cases and all others use the node miner.

bokobza:
@fassadlr Are the tests testing double spend in the same way?

PR #1659, opened 2018-07-16
do not AnnounceBlocks on null behaviors, Fix race condition expectation in MiningNodeWithOneConnectionAlwaysSynced, restore coinbase maturity after altering it in test

monsieurleberre:
restore coinbaseMaturity upon leaving the test

image


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -226,7 +226,8 @@ protected override void OnNextCore(Block block)
             }
 
             // Announce the blocks to each of the peers.
-            IEnumerable<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>());
+            IEnumerable<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>())
```
noescape00: Can the behavior be null here?

noescape00: Those null behaviors- are they added from the peer discovery? If so- don't add connectionManagerBehavior during peer creation in discovery

monsieurleberre: OK I'll check in more details, the issue I was seeing was that upon Disposing of the node, there was still some batches left to send, and then this causes an exception as the behavior is being disposed so it ends up in a strange state.
I think this change is correct anyway, but we could do try and avoid these situation passing and handling correctly a cancellation token...
I'll check in more details what you were advising, thanks

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -226,7 +226,8 @@ protected override void OnNextCore(Block block)
...
+                .Where(b => b != null).ToList();
```
noescape00: ToList is not needed here

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletAddressGenerationAndFundsVisibility_Steps.cs
```diff
@@ -29,6 +29,7 @@ public partial class Wallet_address_generation_and_funds_visibility : BddSpecifi
         private CoreNode sendingStratisBitcoinNode;
         private CoreNode receivingStratisBitcoinNode;
         private long walletBalance;
+        private long previousCoinBaseMaturity;
```
noescape00: Not sure if changes in this file belong to this pr. Why are they here?

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
```diff
@@ -801,6 +801,9 @@ public void Disconnect(string reason, Exception exception = null)
 
             if (this.IsConnected) this.SetStateAsync(NetworkPeerState.Disconnecting).GetAwaiter().GetResult();
 
+            this.Behaviors?.ToList().ForEach(b => b?.Detach());
```
noescape00: Why? No need to detach here.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -226,9 +226,10 @@ protected override void OnNextCore(Block block)
             }
 
             // Announce the blocks to each of the peers.
-            IEnumerable<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>());
+            List<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>())
+                .Where(b => b != null).ToList();
```
noescape00: ToList is not needed here

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
rowandh: `ToList()` should also use null propagation

rowandh: `ToList()` should also use null propagation

Aprogiena: +1 to @monsieurleberre for the courage, but -2 for the Linq usage and most importantly -99 for corrupting the code with this change.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
monsieurleberre: To List here is to prevent multiple enumeration : one with "Count", and one with foreach :)

src/Stratis.Bitcoin/P2P/Peer/NetworkPeer.cs
monsieurleberre: +110 to boost my damaged ego, that was not my idea, and it is getting reverted ;)

monsieurleberre: true! I'll remove that line anyway, thanks

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -226,9 +226,10 @@ protected override void OnNextCore(Block block)
             }
 
             // Announce the blocks to each of the peers.
-            IEnumerable<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>());
+            List<BlockStoreBehavior> behaviours = peers.Select(s => s.Behavior<BlockStoreBehavior>())
```
noescape00: Select will throw if peers collection is modified during the foreach which is used in the select
https://github.com/Microsoft/referencesource/blob/master/System.Core/System/Linq/Enumerable.cs

You can rewrite this to avoid using linq and avoid having a problem like mentioned above

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletAddressGenerationAndFundsVisibility_Steps.cs
```diff
@@ -38,6 +39,9 @@ protected override void BeforeTest()
 
         protected override void AfterTest()
         {
+            this.sendingStratisBitcoinNode.FullNode.Network.Consensus.CoinbaseMaturity = this.previousCoinBaseMaturity;
```
noescape00: Changes in WalletAddressGenerationAndFundsVisibility_Steps.cs do not belong to this PR

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
monsieurleberre: I think it is OK: Linq uses the GetEnumerator() function of the Enumerable, which is for ConcurrentHashSet the following
image

monsieurleberre:
image

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletAddressGenerationAndFundsVisibility_Steps.cs
monsieurleberre: OK, you are right, however I am just trying to make easy to understand PRs and not waste too much time in the process. This is an important fix too because not restoring CoinBaseMaturity leads to the value being changed for all subsequent tests, and test failures. It is easy to understand this very local change...

I should rename the PR to the branch it is based on : integration-test-small-fix :)

noescape00: Ah, so that's fixing integration tests that are currently failing on master?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
noescape00: Ah, yes, peer collection uses ConcurrentHashSet inside. Then ok, it should be good.

noescape00:
I'm fine with the PR, please sync chains and run the tests. 

I'll approve as soon as I do same as well.

monsieurleberre:
OK as far as I am concerned, it looks like it is working, let mw know what you see on your side, or just merge :) Thanks !

noescape00:
Chains synced, failed to pass integration tests. Do they pass for you?

monsieurleberre:
Yes they do, only the 4 in the screenshot don't pass...

Issue #1216, opened 2018-03-15
[Peer Discovery] Add filtering based on IP ranges


dangershony:
Previously nodes would not try to connect to peers that are within the same IP range we need to introduce this functionality again.

PR #1660, opened 2018-07-16
Peer Discovery: Add filtering based on IP ranges #1216

thefazzer:
**[Peer Discovery] Add filtering based on IP ranges (#1216)**

- Nodes should not try to connect to peers that are within the same IP range
- Add a cmd line argument to disable it
- Make this disabled for localhost nodes by default

Preferred: @bokobza @Aprogiena


src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -122,6 +122,7 @@ public ConnectionManagerSettings(NodeSettings nodeSettings)
             this.BurstModeTargetConnections = config.GetOrDefault("burstModeTargetConnections", 1, this.logger);
             this.SyncTimeEnabled = config.GetOrDefault<bool>("synctime", true, this.logger);
             this.RelayTxes = !config.GetOrDefault("blocksonly", DefaultBlocksOnly, this.logger);
+            this.IpRangeFiltering = config.GetOrDefault<bool?>("IpRangeFiltering", null, this.logger);
```
bokobza: You could default to `true` and make the IpRangeFiltering a `bool` rather than a `bool?` .

Also, can you please add the cmd line arg to `BuildDefaultConfigurationFile` and `PrintHelp`?

thefazzer: You could default to true and make the IpRangeFiltering a bool rather than a bool?

bool? exists as default value is ambiguous
This is because IP range filtering has two default behaviours:
If peer is local host: false unless set explicitly to true.
if peer is remote host: true unless set to false.

Aprogiena: I agree with Jeremy, this should be true by default. Local networks should never be considered

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -218,5 +219,8 @@ public static void PrintHelp(Network network)
 
         /// <summary><c>true</c> to enable bandwidth saving setting to send and received confirmed blocks only.</summary>
         public bool RelayTxes { get; set; }
+
+        /// <summary>Disable functionality whereby nodes disallow connection to peers that are within the same IP range to prevent sybil attacks.</summary>
+        public bool? IpRangeFiltering { get; internal set; }
```
Aprogiena: You should add remarks section explaining the exact ranges for each version of IP protocol

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -70,8 +70,41 @@ public override void OnStartConnect()
 
                 PeerAddress peerAddress = this.peerAddressManager.FindPeer(ipEndpoint);
                 if (peerAddress != null && !this.IsPeerConnected(peerAddress.Endpoint))
-                    await ConnectAsync(peerAddress).ConfigureAwait(false);
+                {
+                    // Nodes disallow connection to peers in same range to prevent sybil attacks.     
+                    if (ipEndpoint.Address.IsLocal())
+                    {
+                        // Local peer: filtering is disabled unless explicitly set to true.
```
Aprogiena: No need for this behaviour, just always ignore for local addresses

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -70,8 +70,41 @@ public override void OnStartConnect()
...
+                        if (this.ConnectionSettings.IpRangeFiltering == true)
+                        {
+                            if (this.PeerIsPartOfExistingGroupOfConnectedEndpoints(peerAddress))
+                            {
+                                continue;
+                            }
+                        }
+                        await this.ConnectAsync(peerAddress).ConfigureAwait(false);
```
Aprogiena: Code style and also don't repeat below the same code. Just greater variable weather to connect or not based on whether the address is local or not and whether it is in the same group.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -70,8 +70,41 @@ public override void OnStartConnect()
...
+                    }
+                }
             }
         }
+
+        private bool PeerIsPartOfExistingGroupOfConnectedEndpoints(PeerAddress peerAddress)
+        {
+            IPEndPoint[] connectedEndpoints = this.ConnectorPeers.Select(x => x.PeerEndPoint).ToArray();
+            byte[] GetGroup(IPEndPoint a) => a.Address.GetGroup();
+            return connectedEndpoints.Any(a => GetGroup(a).SequenceEqual(GetGroup(peerAddress.Endpoint.MapToIpv6())));
```
Aprogiena: Use more local variables, use less Linq and don't define the local getgroup method.

Also at comments to the method.

And you probably don't want to use this.ConnectorPeers but this.connectedPeers, or is it something intentional?

In any case, we probably don't need to construct the list of endpoints every time for each peer in this method.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -70,8 +70,41 @@ public override void OnStartConnect()
 
                 PeerAddress peerAddress = this.peerAddressManager.FindPeer(ipEndpoint);
                 if (peerAddress != null && !this.IsPeerConnected(peerAddress.Endpoint))
-                    await ConnectAsync(peerAddress).ConfigureAwait(false);
+                {
+                    // Nodes disallow connection to peers in same range to prevent sybil attacks.     
+                    if (ipEndpoint.Address.IsLocal())
```
fassadlr: @thefazzer can we also please move all of this logic to a separate method? And we need to clean up the nested `if`s please :)

Effectively, we want to go:

```
if(!PeerIsFilteredFromRange(peerAddress)
    continue;

await this.ConnectAsync(peerAddress).ConfigureAwait(false);
```

fassadlr:
@Aprogiena correct me if Im wrong but should the IP range filtering be on the peer connector discovery? I.e. we should be applying it to discovered peers and always connect to peers specified in -connect or -addnode args?

Or should we apply range filtering to all connections.

Just thought of this.

Aprogiena:
Ideally, we would have it on all connection attempts. In that case the protection would be strongest. So if it is reasonably achievable we should aim for that. If there is some problem with that then we can be okay with just having it on non-discovery connections.

fassadlr:
Thanks @Aprogiena,  it shouldnt be too much an issue. @thefazzer and I looked at this yesterday 👍

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Done.

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Have commented method PeerIsPartOfExistingGroup but do I misunderstand the level of detail?  Groups for Type C IP addresses just compare the first XXX.XXX

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: Done.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: This has been refactored completely.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: RE: Use more local variables, use less Linq and don't define the local getgroup method.
[DONE]
RE: this.ConnectorPeers
[connectedPeers is inaccessible an doesn't reflect the connected peers immediately like ConnectorPeers]

RE: we probably don't need to construct the list of endpoints every time for each peer in this method
[This now happens only once outside the loop]

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: Just noting we pair programmed this and new approach has a lot more clarity.

thefazzer:
All tests are for IP version 6 of format "::ffff:000.000.0.0"
IPV4 are converted to IPV6 before processing

Aprogiena:
I know, it is just the process differently under the hood and we don't really have tests for that layer so I was hoping to at least have it tested on this layer.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +279,19 @@ public void StartConnectAsync()
             this.logger.LogTrace("(-)");
         }
 
+        private bool PeerIsPartOfExistingGroup(PeerAddress peerAddress)
+        {
+            // Same group if first 16 bits of the network part of IP are equal ie. 192.168.
```
fassadlr: Fazz can you add this comment in the method's summary?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -79,6 +81,9 @@ public abstract class PeerConnector : IPeerConnector
 
         /// <summary>The network the node is running on.</summary>
         private Network network;
+        
+        /// <summary>Peers connected to this peer.</summary>
+        protected IPEndPoint[] connectedEndpoints;
```
fassadlr: Is this still needed? We have a property/field called `ConnectorPeers`?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -230,6 +235,10 @@ public void StartConnectAsync()
                 return;
             }
 
+            // Connect if local, ip range filtering disabled or ip range filtering enabled and peer in a different group. 
+            if (!peerAddress.Endpoint.Address.IsLocal() && this.ConnectionSettings.IpRangeFiltering && this.PeerIsPartOfExistingGroup(peerAddress))
+                return;
```
fassadlr: Fazz, before we return here might want to log this:

```
this.logger.LogTrace("(-) {0}:{1}", nameof(peerAddress.Endpoint), peerAddress.Endpoint, nameof(peerAddress.Endpoint.Address.IsLocal(), peerAddress.Endpoint.Address.IsLocal())...
```

@Aprogiena ?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -162,6 +163,8 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#agentprefix=<string>");
             builder.AppendLine($"#Enable bandwidth saving setting to send and received confirmed blocks only. Defaults to { (DefaultBlocksOnly ? 1 : 0) }.");
             builder.AppendLine($"#blocksonly={ (DefaultBlocksOnly ? 1 : 0) }");
+            builder.AppendLine($"#bantime=<number>");
+            builder.AppendLine($"#Disallow connection to peers in same IP range. Default true.");
```
Aprogiena: You sure about this line?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -185,6 +188,7 @@ public static void PrintHelp(Network network)
             builder.AppendLine($"-synctime=<0 or 1>        Sync with peers. Default 1.");
             builder.AppendLine($"-agentprefix=<string>     An optional prefix for the node's user agent that will be shared with peers in the version handshake.");
             builder.AppendLine($"-blocksonly=<0 or 1>      Enable bandwidth saving setting to send and received confirmed blocks only. Defaults to { DefaultBlocksOnly }.");
+            builder.AppendLine($"-iprangefiltering=<true or false> Disallow connection to peers in same IP range.  Default is true for remote hosts.");
```
Aprogiena: Doublespace

src/Stratis.Bitcoin/P2P/PeerConnector.cs
Aprogiena: i'd go with just
```this.logger.LogTrace("(-)[RANGE_FILTERED]")```

File: src/Stratis.Bitcoin/P2P/PeerConnector.cs [hunk 12]
```diff
@@ -270,6 +279,19 @@ public void StartConnectAsync()
...
+        {
+            // Same group if first 16 bits of the network part of IP are equal ie. 192.168.
+            byte[] peerAddressGroup = peerAddress.Endpoint.MapToIpv6().Address.GetGroup();
+
+            var endpoints = new List<byte[]>();
+            foreach (IPEndPoint endPoint in this.connectedEndpoints)
+            {
+                endpoints.Add(endPoint.MapToIpv6().Address.GetGroup());
+            }
+            return endpoints.Any(x => x.SequenceEqual(peerAddressGroup));
```
Aprogiena: put empty line before return line

src/Stratis.Bitcoin/P2P/PeerConnector.cs
Aprogiena: Yes but it is not very accurate description better to delete it

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -63,14 +63,21 @@ public override void OnStartConnect()
         /// </summary>
         public override async Task OnConnectAsync()
         {
+            base.connectedEndpoints = this.ConnectorPeers.Select(x => x.PeerEndPoint).ToArray();
```
Aprogiena: this is very weird, why would you do that here?

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -210,6 +210,136 @@ public void PeerConnectorDiscovery_ConnectsTo_DiscoveredPeers()
...
+            Assert.True(connectToThisPeer);
+            
+            // Peer from same group.
+            IPAddress addressFromNodeInSameGroup = IPAddress.Parse("::ffff:192.168.0.2");
+            var endpointOfNodeInSameGroup = new IPEndPoint(addressFromNodeInSameGroup, 80);
+            connectToThisPeer = ConnectToPeer(peerAddressManager, networkPeerFactoryExisting, connectionManagerSettingsExisting, peerConnector, endpointOfNodeInSameGroup);
+            Assert.False(connectToThisPeer);
+
+            // Peer from different group.
+            IPAddress addressFromNodeInDifferentGroup = IPAddress.Parse("::ffff:192.168.0.1");
```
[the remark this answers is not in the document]
thefazzer: I'm confused about this.  We need to use non local addresses to test that part of the functionality, surely?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Removed this line while I can confirm the purpose of this code block.

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Corrected.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: You are correct I can replace this with ConnectorPeers.  Have removed it.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Done.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Agreed - its not a good comment.  I've removed it.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Done.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: Removed connectedEndpoints entirely as per @fassadlr comment.

thefazzer:
@fassadlr @Aprogiena Apologies for checking in failing tests here but I'm having trouble debugging the error being thrown by ipv4 addresses.  await this.networkPeerFactory.CreateConnectedNetworkPeerAsync returns null in these cases and I've tried a number of approaches to step into the function, run it synchronously etc. but without success

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -230,6 +232,13 @@ public void StartConnectAsync()
                 return;
             }
 
+            // Connect if local, ip range filtering disabled or ip range filtering enabled and peer in a different group. 
+            if (!peerAddress.Endpoint.Address.IsLocal() && this.ConnectionSettings.IpRangeFiltering && this.PeerIsPartOfExistingGroup(peerAddress))
```
Aprogiena: I believe here it should not be `IsLocal`, but `IsRoutable(false)`

Aprogiena: The reason for that is that IP addresses such as 192.168.0.1 are local network addresses and IsLocal only considers localhost - 127.0.0.x

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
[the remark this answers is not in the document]
Aprogiena: I would expect some tests to work with normal IP version 6 addresses, not only IP version 4 addresses in IP version 6 format. You are only working with IP version 4 addresses everywhere. Moreover, mostly you use local network addresses, so it feels like the test cases are very limited. There are different rules for IP version 6 addresses so if you just use IP version 4 addresses, even if they are converted to IP version 6 format, you will certainly not test how does it behave for a IP version six addresses. Similarly, if you mostly use local network addresses, it will not test nonlocal network addresses scenarios.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
```diff
@@ -69,8 +69,13 @@ public override void OnStartConnect()
                     return;
 
                 PeerAddress peerAddress = this.peerAddressManager.FindPeer(ipEndpoint);
-                if (peerAddress != null && !this.IsPeerConnected(peerAddress.Endpoint))
-                    await ConnectAsync(peerAddress).ConfigureAwait(false);
+                if (peerAddress == null)
```
fassadlr: @thefazzer actually, can you completely revert the changes to this file? Then the PR is smaller 👍

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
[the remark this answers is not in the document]
thefazzer: Have changed to use a mix of ipv4 and ipv6.  If these need changing or extending pls comment

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Made this substitution.  Logic therefore ignores filtering behaviour for routable IPs.

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: Not sure I understand the purpose?

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -210,6 +210,141 @@ public void PeerConnectorDiscovery_ConnectsTo_DiscoveredPeers()
...
+            bool connectToThisPeer = ConnectToPeer(peerAddressManager, networkPeerFactoryExisting, connectionManagerSettingsExisting, peerConnector, endpointPeer1);
+            Assert.True(connectToThisPeer);
+            
+            // Peer from same group.
+            IPAddress addressInSameNetworkGroupPeer2 = IPAddress.Parse("::ffff:213.86.221.36"); // ipv4
+            var endpointPeer2 = new IPEndPoint(addressInSameNetworkGroupPeer2, 80);
+            connectToThisPeer = ConnectToPeer(peerAddressManager, networkPeerFactoryExisting, connectionManagerSettingsExisting, peerConnector, endpointPeer2);
+            Assert.False(connectToThisPeer);
+
+            // Peer from same group: (213.86.221.37)
```
Aprogiena: Wrong comment?

Also why this peer should fail to connect?

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -210,6 +210,141 @@ public void PeerConnectorDiscovery_ConnectsTo_DiscoveredPeers()
...
+
+            // Peer 2 has different network group.
+            IPAddress addressInDifferentNetworkGroupPeer2 = IPAddress.Parse("99be:f5c5:adc2:525c:f6d7:7b30:5336:5a0f");  // ipv6
+            var endpointPeer2 = new IPEndPoint(addressInDifferentNetworkGroupPeer2, 80);
+            peerAddressManager.AddPeer(endpointPeer2, IPAddress.Loopback);
+            connectedToThisPeer = ConnectToPeer(peerAddressManager, networkPeerFactoryExisting, connectionManagerSettingsExisting, peerConnector, endpointPeer2);
+            Assert.True(connectedToThisPeer);  // Different network group: connects.
+
+            // Peer 3 in same network group as Peer 2.
+            IPAddress addressInSameNetworkGroupPeer3 = IPAddress.Parse("99be:f5c5:adc2:525c:db45:d36e:ce01:a394"); // ipv6
```
Aprogiena: Use these exact two peers also in the test range filtering is enabled

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -162,6 +163,7 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#agentprefix=<string>");
             builder.AppendLine($"#Enable bandwidth saving setting to send and received confirmed blocks only. Defaults to { (DefaultBlocksOnly ? 1 : 0) }.");
             builder.AppendLine($"#blocksonly={ (DefaultBlocksOnly ? 1 : 0) }");
+            builder.AppendLine($"#bantime=<number>");
         }
```
Aprogiena: can we add iprangefiltering?

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -162,6 +163,7 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#agentprefix=<string>");
             builder.AppendLine($"#Enable bandwidth saving setting to send and received confirmed blocks only. Defaults to { (DefaultBlocksOnly ? 1 : 0) }.");
             builder.AppendLine($"#blocksonly={ (DefaultBlocksOnly ? 1 : 0) }");
+            builder.AppendLine($"#bantime=<number>");
```
Aprogiena: It seems like every line here has an explanation line before it, so we should have an explanation line before this one as well

fassadlr:
@thefazzer I'm happy with the changes I requested. Over to @Aprogiena !

src/Stratis.Bitcoin/P2P/PeerConnectorConnect.cs
thefazzer: If I've done this correctly PeerConnectorConnect should now show no changes.

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
thefazzer: Have corrected and standardized tests so they use same data, same positive and negative tests and same mix of ipv4 and ipv6.

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
thefazzer: As above ^^^

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Done.

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: As above ^^^

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +279,23 @@ public void StartConnectAsync()
...
+        private bool PeerIsPartOfExistingGroup(PeerAddress peerAddress)
+        {
+            if (this.ConnectorPeers == null)
+                return false;
+
+            byte[] peerAddressGroup = peerAddress.Endpoint.MapToIpv6().Address.GetGroup();
+
+            var endpoints = new List<byte[]>();
+
+            foreach (INetworkPeer endPoint in this.ConnectorPeers)
```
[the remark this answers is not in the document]
thefazzer: Agreed - I want to update the collection of connected peers that is managed by the ConnectionManager. However after making this change test is failings as the ConnectionManager object is not updated after ConnectorPeers is.   In the tests will I have to trigger ConnectionManager.OnStateChangedAsync?  How does the connection mgr know about changes in the connector?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +279,23 @@ public void StartConnectAsync()
...
+            {
+                endpoints.Add(endPoint.PeerEndPoint.MapToIpv6().Address.GetGroup());
```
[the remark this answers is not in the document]
thefazzer: Yes I see the point here - have refactored.

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -164,6 +164,8 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#Enable bandwidth saving setting to send and received confirmed blocks only. Defaults to { (DefaultBlocksOnly ? 1 : 0) }.");
             builder.AppendLine($"#blocksonly={ (DefaultBlocksOnly ? 1 : 0) }");
             builder.AppendLine($"#bantime=<number>");
+            builder.AppendLine($"#Disallow connection to peers in same IP range. Default is true for remote hosts.");
+            builder.AppendLine($"#iprangefiltering=<true or false>");
```
Aprogiena: is it really true/false or is it 1/0?

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -281,19 +281,18 @@ public void StartConnectAsync()
...
+            foreach (INetworkPeer endPoint in this.connectedPeers)
             {
-                endpoints.Add(endPoint.PeerEndPoint.MapToIpv6().Address.GetGroup());
+                byte[] endPointGroup = endPoint.PeerEndPoint.MapToIpv6().Address.GetGroup();
+                if (endPointGroup.SequenceEqual(peerAddressGroup))
+                    return true;
             }
-
-            return endpoints.Any(x => x.SequenceEqual(peerAddressGroup));
+            return false;
```
Aprogiena: empty line above this

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
thefazzer: Correcting this to to 1,0 to be consistent with other settings.  Also changed PrintHelp for consistency.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Fixed.

Aprogiena:
What is the status of the tests? Do all tests including intergration tests pass?

thefazzer:
@Aprogiena Tests fixed.  Needed reflection as ConnectionManager is out of scope and _connectedPeers is immutable and private in the base class

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -274,5 +436,20 @@ private IConnectionManager CreateConnectionManager(NodeSettings nodeSettings, Co
 
             return connectionManager;
         }
+
+        private static void AddMissingPeersToConnectionManager(PeerConnectorConnectNode peerConnector)
```
fassadlr: @thefazzer we really need to stop using reflection in unit tests. If we are forced to its either bad class design or we aren't exposing the relevant properties.

In this case you will see that the PeerConnector gets initialized with the connection manager's `connectedPeers` collection. So in the test you can add peers to a connection manager instance and call initialize on the peer connector. This will then set the connected peers in the connector class.

thefazzer:
@fassdlr Would love to remove reflection, but I need to access ConnectionManager or connectedPeers somehow.  Am I allowed to refactor the PeerConnector code to do this?

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -367,6 +367,7 @@ private static bool ConnectToPeer(PeerAddressManager peerAddressManager, Mock<IN
                     It.IsAny<NetworkPeerConnectionParameters>(), It.IsAny<NetworkPeerDisposer>())).Returns(Task.FromResult(networkPeer.Object));
             connectionManagerSettingsExisting.AddNode.Add(endpointNode);
             connectionManagerSettingsExisting.Connect.Add(endpointNode);
+            AddMissingPeersToConnectionManager(peerConnector);
```
Aprogiena: There should be no need for this, just check whether you want to return true or false from this method and if true, just add it to the peer collection

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
Aprogiena: > we really need to stop using reflection in unit tests

I don't think this is correct – it has been agreed while ago that it is the preferable method of doing things

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -435,5 +436,20 @@ private IConnectionManager CreateConnectionManager(NodeSettings nodeSettings, Co
 
             return connectionManager;
         }
+
+        private static void AddMissingPeersToConnectionManager(PeerConnectorConnectNode peerConnector)
+        {
+            Type type = peerConnector.GetType().BaseType;
+            System.Reflection.FieldInfo f = ((System.Reflection.FieldInfo[])((System.Reflection.TypeInfo)type).DeclaredFields)[3];
```
Aprogiena: Though this is another level that is definitely wrong

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
fassadlr: It was agreed in one instance but then blew out of proportion. Like I said, it should really really be last resort but then again I can't see how something can't be refactored or redesigned so that it can be better unit tested.

fassadlr:
@thefazzer please see my comment explaining how this can be achieved 👍

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -210,6 +210,168 @@ public void PeerConnectorDiscovery_ConnectsTo_DiscoveredPeers()
...
+            networkPeer.SetupGet(np => np.PeerEndPoint).Returns(new IPEndPoint(endpointNode.Address, endpointNode.Port));
+            networkPeer.SetupGet(np => np.RemoteSocketAddress).Returns(endpointNode.Address);
+            networkPeer.SetupGet(np => np.RemoteSocketPort).Returns(endpointNode.Port);
+            networkPeer.SetupGet(np => np.State).Returns(NetworkPeerState.HandShaked);
+            networkPeerFactoryExisting.Setup(npf =>
+                npf.CreateConnectedNetworkPeerAsync(It.Is<IPEndPoint>(x => Equals(x, endpointNode)), 
+                    It.IsAny<NetworkPeerConnectionParameters>(), It.IsAny<NetworkPeerDisposer>())).Returns(Task.FromResult(networkPeer.Object));
+            connectionManagerSettingsExisting.AddNode.Add(endpointNode);
+            connectionManagerSettingsExisting.Connect.Add(endpointNode);
+            AddMissingPeersToConnectionManager(peerConnector);
```
fassadlr: @thefazzer here you can just use mock a new instance of ConnectionManager and to its collection of ConnectedPeers and the initialize the PeerConnector with it :)

fassadlr: This will eliminate the need for reflection etcetc

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
Aprogiena: It just feels weird that half of us thinks that we agreed that we should use it and yet have does not. We should discuss this on slack or elsewhere separately

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
thefazzer: I agree - but the problem is "just add it to the peer collection"
because the collection is immutable and private         
private IReadOnlyNetworkPeerCollection connectedPeers;

fassadlr: @fenix2222 go look at ConnectionManager there is a `AddConnectedPeer` method on it ;)

Aprogiena: @fassadlr @fenix2222 != @thefazzer , otherwise you are perfectly right ;)

fassadlr: @Aprogiena hahaha tag fail 👎

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
thefazzer: Done.

src/Stratis.Bitcoin.Tests/P2P/PeerConnectorTests.cs
```diff
@@ -436,20 +427,34 @@ private IConnectionManager CreateConnectionManager(NodeSettings nodeSettings, Co
 
             return connectionManager;
         }
-
-        private static void AddMissingPeersToConnectionManager(PeerConnectorConnectNode peerConnector)
+        private bool ConnectToPeer(PeerAddressManager peerAddressManager, Mock<INetworkPeerFactory> networkPeerFactoryExisting, ConnectionManagerSettings connectionManagerSettingsExisting, PeerConnectorConnectNode peerConnector, IPEndPoint endpointNode, Mock<IConnectionManager> mockConnectionManager)
```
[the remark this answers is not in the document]
thefazzer: Fixed.

PR #1662, opened 2018-07-17
[Integration Tests] Fix GenerateBlockManually

fassadlr:


src/Stratis.Bitcoin.IntegrationTests/SmartContracts/SmartContractWalletTests.cs
```diff
@@ -89,7 +88,8 @@ public void SendAndReceiveCorrectly()
                 Assert.Null(scReceiver.FullNode.WalletManager().GetSpendableTransactionsInWallet(WalletName).First().Transaction.BlockHeight);
 
                 // generate two new blocks do the trx is confirmed
-                scSender.GenerateBlockManually(new List<Transaction>(new[] { scSender.FullNode.Network.CreateTransaction(trx.ToBytes()) }));
+                scSender.AddToStratisMempool(scSender.FullNode.Network.CreateTransaction(trx.ToBytes()));
```
bokobza: Is this to follow up on what we talked about in the previous PR? (add to mempool then mine)

fassadlr: Hi Jeremy, yes that is correct 👍

fassadlr: I was able to change some of the tests to use this.

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -311,7 +311,6 @@ public void MiningNodeWithOneConnectionAlwaysSynced()
                 Thread.Sleep(1);
 
             // Make sure that miner did not advance yet but connector did.
-            Assert.NotEqual(miner.FullNode.Chain.Tip.HashBlock, networkNode1.FullNode.Chain.Tip.HashBlock);
```
bokobza: I agree that these look a bit flaky:
- go fast and the miner will be equal to networkNode1
- go slow and it won't be equal

At the same time, a slow or fast computer will mean slow/fast propagation but also slow/fast pass through the assertions.

I think this needs to be redesigned to either wait for nodes to be in sync or only connect nodes after one of them has moved forward.

Issue #1391, opened 2018-05-06
Don't allow block signature validation rule skip


noescape00:
Block signature rule shouldn't be skippable (because a bogus signature might be supplied and then when some peer asks us for this block we will provide a block with a wrong signature and we will be banned by such a peer).

Created by @Aprogiena

dangershony:
Ah yeah thats because the block signature is not part of the header hash we have to check it every time.

fenix2222:
@bokobza @dangershony Can this be closed now as https://github.com/stratisproject/StratisBitcoinFullNode/pull/1663 was merged?

PR #1663, opened 2018-07-17
Made block signature validation rule non-skippable

fenix2222:
As requested in issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/1391

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
@@ -12,7 +12,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
     /// <summary>
     /// A rule that will validate the signature of a PoS block.
     /// </summary>
-    [PartialValidationRule(CanSkipValidation = true)]
+    [PartialValidationRule(CanSkipValidation = false)]
     [IntegrityValidationRule]
     public class PosBlockSignatureRule : StakeStoreConsensusRule
     {


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRule.cs
```diff
@@ -12,7 +12,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
     /// <summary>
     /// A rule that will validate the signature of a PoS block.
     /// </summary>
-    [PartialValidationRule(CanSkipValidation = true)]
+    [PartialValidationRule(CanSkipValidation = false)]
```
dangershony: Its false by default so you can just remove the setter, but this is fine also.

dangershony: Ah should this not be IntegrityValidator attribute? @Aprogiena

Aprogiena: Integrity validator should be block signature and Merkle tree root and mutation, so yes it should be

Aprogiena: It just means that another fix is required and Dan will probably take care of that.

dangershony: Yeah it means it will be executed in a different part of the process of downloading a block.

Issue #1420, opened 2018-05-18
Exception in Periodic peer flush


bokobza:
While syncing bitcoin Mainnet IBD, I encountered this:

```

[2018-05-17 21:32:41.5325 13] INFO: Stratis.Bitcoin.Features.Wallet.WalletManager.<Start>b__28_0 Wallets saved to file at 05/17/2018 21:32:41.
[2018-05-17 21:32:41.5325 15] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__75_0 ======Node stats====== 05/17/2018 21:32:41 agent StratisBitcoin:1.1.2
Headers.Height:      523156   Headers.Hash:     000000000000000000282956c72c31f8245af7a3ddc00c001ff7e590d2f66d49
Consensus.Height:    220659   Consensus.Hash:   00000000000000d5f9a2c3297c8e0ced836569f928d5e0c592f86e7a87683b87
BlockStore.Height:   220633   BlockStore.Hash:  00000000000002bda08f9acd807812b84441d0e7ea139ef74cad69e9024b4240
Wallet.Height:       No Wallet

======Mempool======
MempoolSize: 0    DynamicSize: 0 kb   OrphanSize: 0   

======Connection======
Peer:[::ffff:62.109.20.99]:8333,         connected:outbound,        height:523156,        agent:/Satoshi:0.15.1/
Peer:[::ffff:199.182.129.26]:8333,       connected:outbound,        height:523156,        agent:/Satoshi:0.16.0/
Peer:[::ffff:113.252.193.10]:8333,       connected:outbound,        height:523156,        agent:/Satoshi:0.15.1/
Peer:[::ffff:91.106.194.97]:8333,        connected:outbound,        height:523156,        agent:/Satoshi:0.16.0/
Peer:[::ffff:176.9.109.123]:8333,        connected:outbound,        height:523156,        agent:/Satoshi:0.16.0/
Peer:[::ffff:204.9.50.25]:8333,          connected:outbound,        height:523156,        agent:/Satoshi:0.15.1/
Peer:[::ffff:67.205.142.50]:8333,        connected:outbound,        height:523156,        agent:/Satoshi:0.16.0/
Peer:[::ffff:67.149.65.252]:8333,        connected:outbound,        height:523156,        agent:/Satoshi:0.15.1/


[...]

PR #1664, opened 2018-07-17
Fixed sorting issue with concurrent dictionary

fenix2222:
As per issue #1420

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
@@ -158,7 +158,13 @@ namespace Stratis.Bitcoin.P2P
                 return;
 
             var fileStorage = new FileStorage<List<PeerAddress>>(this.PeerFilePath.AddressManagerFilePath);
-            fileStorage.SaveToFile(this.peers.OrderByDescending(p => p.Value.LastConnectionSuccess).Select(p => p.Value).ToList(), PeerFileName);
+            fileStorage.SaveToFile(
+                this.peers
+                    .ToList()
+                    .OrderByDescending(p => p.Value.LastConnectionSuccess)
+                    .Select(p => p.Value)
+                    .ToList(), 
+                PeerFileName);
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


dangershony:
I don't think it matters so much the time of the snapshot, we need to prevent concurrency bugs.
But ok use `ToArray()` or if you ant to use to `ToList()` then enumerate the first item as so
```
this.peers.Skip(0).ToList()
```

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -158,7 +158,13 @@ public void SavePeers()
                 return;
 
             var fileStorage = new FileStorage<List<PeerAddress>>(this.PeerFilePath.AddressManagerFilePath);
-            fileStorage.SaveToFile(this.peers.OrderByDescending(p => p.Value.LastConnectionSuccess).Select(p => p.Value).ToList(), PeerFileName);
+            fileStorage.SaveToFile(
+                this.peers
+                    .ToList()
+                    .OrderByDescending(p => p.Value.LastConnectionSuccess)
+                    .Select(p => p.Value)
+                    .ToList(),
```
dangershony: You probably don't need to enumerate twice so remove the second `ToList()`

dangershony: Perhaps also for readability take the snapshot first in a new param?

fenix2222: Actually looking at the implementation, `ToArray` is definitely safer. `GetEnumerator` doesn't guarantee correct snapshot and you do need second `ToList` as this is the type of the parameter SaveToFile expects

fenix2222: Found a much better alternative. We can use `.Values` collection which has a lock and removes the need to use `Select(v => v.Value)`

fenix2222: @dangershony can we merge now?

fenix2222:
@dangershony @bokobza Can one of you merge?

PR #1666, opened 2018-07-17
[SC] Merge master into sc-dev

fassadlr:


rowandh:
To save us all time, can we instead merge sc-dev into master at the frequency we discussed, and afterwards merge master back into sc-dev?

We're not so dependent on the FN that we need the latest changes in our dev branch immediately.

codingupastorm:
To add to that, I think we should permit merges from master -> sc-dev to be pushed freely, rather than opening PRs.

fassadlr:
Thanks guys! The only reason I had to do this merge from master was because of a merge conflict ;)

PR #1678, opened 2018-07-18
Update to commenting policy

dangershony:
To stop long discussions about commenting this will relax a bit the commenting policy.

### DIFF
--- a/Documentation/commenting-policy.md
+++ b/Documentation/commenting-policy.md
@@ -39,8 +39,7 @@ We have two types of rules. Hard rules are those that are required and your code
 them. Soft rules are more like guidelines - try to use them where possible. The reviewer can ask you to update your code 
 if it violates a soft rule and it can be fixed easily.
 
- * **XMLDOC** - Use triple slash (XML documentation) comments for classes, interfaces, methods, parameters, fields, properties, structures, enums, etc. These comments are required - even for trivial code. *(HARD)*
-   * **Exceptions** - Constructors, method parameters, private methods, private properties, private fields. *(SOFT)*
+ * **XMLDOC** - Use triple slash (XML documentation) comments for classes, interfaces, methods, parameters, fields, properties, structures, enums, etc. These comments are optional for: constructors, method parameters, private methods, private properties, private fields. *(SOFT)* These comments are required for everything else - even for trivial code. *(HARD)*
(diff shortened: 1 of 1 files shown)


bokobza:
ACK

Aprogiena:
LOL

fassadlr:
I approve @dangershony !

Aprogiena:
Dude, I disagree with the idea, but you know that. The LOL here is that you completely mess it up, so at least do it properly:

```
* **XMLDOC** - Use triple slash (XML documentation) comments for classes, interfaces, methods, parameters, fields, properties, structures, enums, etc. These comments are optional for: constructors, method parameters, private methods, private properties, private fields. *(SOFT)* These comments are required for everything else - even for trivial code. *(HARD)*
   * **SUMMARY-1** - First paragraph in `<summary>` must be short, no more than 3 lines. If you find you need more than 2 lines, maybe you are not writing the summary anymore and you are already explaining details. *(HARD)*
```

instead of what you created there

noescape00:
I strongly disagree with allowing not having comments on all private methods, private properties, private fields. 
I'd go with allowing not commenting on those if those are instances of classes that have summary comments.

dangershony:
>I'd go with allowing not commenting on those if those are instances of classes that have summary comments.
This is inferred as you wont have a private class, so all classes must have summaries.

monsieurleberre:
Although I am a bit late to the party, I agree with being more relaxed, I've done it my whole life and it is awesome ❤️

majikandy:
A slight tweak to this would be good....

> These comments are required for everything else - even for trivial code. *(HARD)*

I don't agree on any policy that suggests writing comments on trivial code.  The reason being, this is another way of saying "write redundant comments" which is definitely not a good idea when there is an opportunity for the code to communicate itself.

I would suggest that this line of text also be removed from the policy.

Aprogiena:
@majikandy disagree, there are good reasons, so nothing like "definitely not a good idea", but just vote on slack for deleting the policies

codingupastorm:
I think a high degree of flexibility is the way to get strong devs to thrive.

When a PR comes in, if both the developer and the reviewers believe that what they're committing is strong, correct, and reflects well on themselves and the team, it should be merged.

So +1 to a loosening of the rules, not deleting the rules altogether.

dangershony:
Andy trivial for you is not trivial for others, perhaps text should be "even if may seem trivial"

PR #1681, opened 2018-07-19
[SC] Precursor to creating contracts

codingupastorm:
Sets up the API, validation rules and the AccountState in preparation for storing contracts. To declare what contract to deploy we use the `ToDeploy` attribute. 

Building on top of this soon:
* Implement the Create method on SmartContract
* If there is just 1 contract in the assembly, deploy that, don't need ToDeploy.
* Compiling more than 1 contract via Sct  / use of namespaces


src/Stratis.SmartContracts.Executor.Reflection/ReflectionVirtualMachine.cs
```diff
@@ -53,7 +53,7 @@ public class ReflectionVirtualMachine : ISmartContractVirtualMachine
 
             // Decompile the contract execution code and validate it.
             SmartContractDecompilation decompilation = SmartContractDecompiler.GetModuleDefinition(callData.ContractExecutionCode);
-
+            decompilation.ContractType = decompilation.ModuleDefinition.Types.FirstOrDefault(x => x.CustomAttributes.Any(y => y.AttributeType.Name == typeof(ToDeployAttribute).Name));
```
rowandh: Move this to SmartContractDecompiler.GetModuleDefinition

codingupastorm:
Thanks for the review @rowandh , made some significant changes (all from that one line comment you suggested), so you will need to re-review, just the last commit. 

All tests passing!

src/Stratis.SmartContracts.Executor.Reflection/SmartContractGasInjector.cs
```diff
@@ -41,19 +41,19 @@ public static class SmartContractGasInjector
             OpCodes.Callvirt
         };
 
-        public static TypeDefinition GetContractType(ModuleDefinition moduleDefinition)
-        {
-            return moduleDefinition.Types.FirstOrDefault(x => x.FullName != "<Module>");
-        }
+        //public static TypeDefinition GetContractType(ModuleDefinition moduleDefinition)
```
fassadlr: Remove commented code 👍

src/Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "commandLineArgs": "-txindex"
```
fassadlr: Revert :)

src/Stratis.SmartContracts.Core/State/ContractStateRepository.cs
```diff
@@ -100,6 +100,19 @@ public byte[] GetStorageValue(uint160 addr, byte[] key)
             return accountState == null ? null : this.storageCache.Get(addr.ToBytes()).Get(key);
         }
 
+        public string GetContractType(uint160 addr)
+        {
+            AccountState accountState = this.GetAccountState(addr);
+            return accountState != null ? accountState.TypeName : "";
```
fassadlr: "" to string.Empty.

```return accountState != null ? accountState.TypeName : string.Empty;```

rowandh: I have to ask if this is worth holding up a PR for? There is no difference between string.Empty and "" - they both refer to the same string literal. You can check this with `ReferenceEquals("", string.Empty);`.

In which case, this is a coding standards issue, and are we *really* at this level of finicky?

codingupastorm: I changed it for now but if that is actually written somewhere I'm really against it and will never enforce it in PRs I'm reviewing. 

More rules kills the coding flair yo!

PR #1593, opened 2018-07-03
Upgrade to dotnet 2.1

bokobza:
* Swashbuckle.AspNetCore 1.0.0 -> 2.5.0 No breaking changes
* Microsoft.NET.Test.Sdk 15.5.0 -> 15.7.2 No breaking changes
* FluentAssertions 5.2.0 -> 5.4.1 No breaking changes
* Moq 4.8.2 -> 4.8.3 
  Breaking Change: Add ISetupSequentialResult<TResult>.Returns method overload that support delegate for deferred results (@snrnats, #594)
  This change requires the API consumer to specify the type in ISetupSequentialResult<TResult>.Returns when returning null. Otherwise, it is ambiguous whether the type returned is a Type or a Func<Type>.
  https://github.com/moq/moq4/blob/master/CHANGELOG.md

* System.Reactive 3.1.1 -> 4.0.0
* Newtonsoft.Json 10.0.3 -> 11.0.2
* dotnet core 2.0 -> 2.1

✅ Unit tests
✅ Integration tests
✅ 4 blockchains synced

All tests passing apart from 5 consensus-related ones related to another piece of work.
 
image

PR #1682, opened 2018-07-19
[SC] Create Smart Contract Wallet / Clean up sc code in wallet classes.

fassadlr:
@dangershony @bokobza can you review the wallet changes?


fassadlr:
@codingupastorm @rowandh you guys can review the whole PR please :)

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -34,16 +34,16 @@ public class WalletController : Controller
         private readonly CoinType coinType;
 
         /// <summary>Specification of the network the node runs on - regtest/testnet/mainnet.</summary>
-        private readonly Network network;
+        public readonly Network Network;
```
bokobza: How come a lot of readonly properties have become public?

src/Stratis.Bitcoin.Features.Wallet/WalletFeature.cs
```diff
@@ -193,4 +193,4 @@ public static IFullNodeBuilder UseWallet(this IFullNodeBuilder fullNodeBuilder)
             return fullNodeBuilder;
         }
     }
-}
+}
```
bokobza: Can you please turn off the thing that removes the extra line at the end of files?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
fassadlr: It is because SmartContractWalletController now inherits from WalletController :)

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/SmartContractWalletController.cs
```diff
@@ -0,0 +1,104 @@
...
+            IConnectionManager connectionManager,
+            Network network, ConcurrentChain chain,
+            IBroadcasterManager broadcasterManager,
+            IDateTimeProvider dateTimeProvider) :
+            base(loggerFactory, walletManager, walletTransactionHandler, walletSyncManager, connectionManager, network, chain, broadcasterManager, dateTimeProvider)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+        }
+
+        public override IActionResult SendTransaction([FromBody] SendTransactionRequest request)
```
bokobza: No need to decorate the endpoint with 
```
[Route("send-transaction")]
[HttpPost]
```
here?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
dangershony: Then use protected in that case.

dangershony: Ok as discussed undo all WalletController stuff and only delete SC related code.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -82,7 +82,7 @@ private string GetRPCAuth()
 
         public void NotInIBD()
         {
-            ((InitialBlockDownloadStateMock) this.FullNode.NodeService<IInitialBlockDownloadState>()).SetIsInitialBlockDownload(false, DateTime.UtcNow.AddMinutes(5));
+            ((InitialBlockDownloadStateMock)this.FullNode.NodeService<IInitialBlockDownloadState>()).SetIsInitialBlockDownload(false, DateTime.UtcNow.AddMinutes(5));
```
dangershony: You are just bloating you PR with this kind of things

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisSmartContractNode.cs
```diff
@@ -25,11 +24,11 @@ public override void BuildNode()
                 .UseNodeSettings(settings)
                 .UseBlockStore()
                 .UseMempool()
-                .UseWallet()
                 .AddRPC()
                     .AddSmartContracts()
                     .UseSmartContractConsensus()
                     .UseSmartContractMining()
+                    .UseSmartContractWallet()
```
dangershony: Nice

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
bokobza: Yeah please @fassadlr turn this auto-cleaning off, or do a PR that does this across the whole solution. ✨✨

src/Stratis.Bitcoin/Utilities/ModelStateErrors/ModelStateErrors.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Stratis.Bitcoin.Utilities.JsonErrors;
+
+namespace Stratis.Bitcoin.Utilities.ModelStateErrors
+{
+    public static  class ModelStateErrors
```
bokobza: I think this is better suited to the Stratis.Bitcoin.Features.Api, since this will only ever be used when the Api is present.

fassadlr:
All unit and integration tests pass.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -192,37 +188,18 @@ public Money EstimateFee(TransactionBuildContext context)
...
-            };
-
             this.AddRecipients(context);
             this.AddOpReturnOutput(context);
             this.AddCoins(context);
             this.AddSecrets(context);
-            // Also added for smart contract demo
-            if (!context.ChangeAddressSet)
-            {
+            if (context.ChangeAddress == null)
```
bokobza: I think we didn't have this check before.

src/Stratis.Bitcoin/Utilities/ModelStateErrors/ModelStateErrors.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Stratis.Bitcoin.Utilities.ModelStateErrors
+{
+    public static class ModelStateErrors
+    {
+        /// <summary>
+        /// Builds an <see cref="IActionResult"/> containing errors contained in the <see cref="ControllerBase.ModelState"/>.
+        /// </summary>
+        /// <returns>A result containing the errors.</returns>
+        public static IActionResult BuildErrorResponse(ModelStateDictionary modelState)
```
bokobza: I think this belongs to the Api feature because it will only ever be used when the feature is present.

bokobza: ah no because the Stratis.Bitcoin project also has controller where this is used.
Please ignore.

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/SmartContractWalletExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Linq;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Wallet
```
codingupastorm: We could get rid of this and  Stratis.Bitcoin.Features.SmartContracts.Wallet could reference Stratis.SmartContracts.Core. 

We have to work out the .NET Standard 2.0 vs .NET Core 2.1 rules to do this though. Do we have any rules for which projects are which? It looks like the rule is that all Features should be .NET Standard 2.0, but our SmartContract feature is Core 2.1. This is because SmartContracts.Core and Executor.Reflection are also Core 2.1. Think we just bootstrapped all the way. 

@rowandh @fassadlr let's discuss!

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
codingupastorm: If I recall correctly, you're right the FN didn't, but you also couldn't select which address to send your change to by setting a ChangeAddress in the TransactionBuildContext that gets passed into BuildTransaction. 

I think it's a little confusing that the TransactionBuildContext that I build and pass into BuildTransaction is then continually modified by the WalletTransactionHandler + Builder.

An alternative approach would be to have a kind of `TransactionBuildOptions` object that is passed into the TransactionBuilder. TransactionBuildContext would only be used internally. I'd really like to set this up and it would help both FN and SC a lot I think @dangershony what do you reckon? Can I work on this?

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/SmartContractWalletExtensions.cs
rowandh: The FN was updated to .NET Core 2.1 in #1593. We could bring our solutions in line with this (I had a PR to do it in VSTS).

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -416,11 +392,11 @@ public class TransactionBuildContext
         /// <param name="recipients">The target recipients to send coins to.</param>
         /// <param name="walletPassword">The password that protects the wallet in <see cref="accountReference"/></param>
         /// <param name="opReturnData">Optional transaction data <see cref="OpReturnData"/></param>
-        public TransactionBuildContext(WalletAccountReference accountReference, List<Recipient> recipients,
-            string walletPassword = "", string opReturnData = null)
+        public TransactionBuildContext(Network network, WalletAccountReference accountReference, List<Recipient> recipients, string walletPassword = "", string opReturnData = null)
         {
             Guard.NotNull(recipients, nameof(recipients));
 
+            this.TransactionBuilder = new TransactionBuilder(network);
```
rowandh: This feels like a smell. We pollute our code with `Network` but it's not an explicit dependency of `TransactionBuildContext`.

fassadlr: It is an explicit dependency on `TransactionBuilder`, which must (currently) be set on `TransactionBuildContext`.

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/SmartContractWalletTransactionHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+                GroupByScriptPubKey = false
+            };
+
+            context.TransactionBuilder.DustPrevention = false;
+            context.TransactionBuilder.StandardTransactionPolicy = this.TransactionPolicy;
+
+            if (context.ChangeAddress != null)
+                context.TransactionBuilder.SetChange(context.ChangeAddress.ScriptPubKey);
+
+            base.InitializeTransactionBuilder(context);
```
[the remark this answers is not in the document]
fassadlr: Hmmm I did initially thought of being explicit here. OK will implement as suggested 👍

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
rowandh: Exactly, but given that TransactionBuilder needs to change so often it's likely not even a true class-level dependency.

I don't think this is a blocker to merging though.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
fassadlr: @codingupastorm just so that we can move this along, I would suggest that we get this merged and you can work on this in a separate PR? BTW I agree with your suggestion 👍

fassadlr:
@codingupastorm @bokobza @bokobza I was thinking maybe we should combine or move the current API methods in `SmartContractController` to `SmartContractWalletController`? It seems a bit superfluous to have two cobtrollers when the current controller is actually wallet methods?

codingupastorm:
@fassadlr I think there's a semantic difference between the actual wallet calls and the smart contract queries. 

For example, retrieving storage, calling methods locally, checking contract code could all happen from someone without an identity / address (e.g. a block explorer). So maybe we keep them separate, although you could move the wallet functions for now.

There could definitely be a case made for renaming the 'SmartContractsController' though. It's about Local Smart Contract State Querying. Maybe there's something in there.

fassadlr:
@codingupastorm Can I perhaps suggested that I move those wallet methods over in a following PR? It appears to be a bit more work and I want have this PR merged. @bokobza can you please review and approve?

codingupastorm:
Definitely! No worries. Shall we ping @dangershony to take a look instead?

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -24,26 +24,26 @@ namespace Stratis.Bitcoin.Features.Wallet
...
         private readonly CoinType coinType;
 
         private readonly ILogger logger;
 
+        public Network Network { get; }
+
         private readonly MemoryCache privateKeyCache;
 
-        private readonly StandardTransactionPolicy transactionPolicy;
+        public readonly StandardTransactionPolicy TransactionPolicy;
```
dangershony: Protected

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -356,10 +332,9 @@ private void AddCoins(TransactionBuildContext context)
         /// <remarks>
         /// Add outputs to the <see cref="TransactionBuilder"/> based on the <see cref="Recipient"/> list.
         /// </remarks>
-        private void AddRecipients(TransactionBuildContext context)
+        public virtual void AddRecipients(TransactionBuildContext context)
```
dangershony: protected

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/SmartContractWalletController.cs
fassadlr: @bokobza I have fixed this :)

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
fassadlr: Ah this needs to be virtual as we override the behavior in SmartContractWalletTransactionHandler

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
dangershony: Also agree on a TransactionBuildOptions yeah explore this possibility

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
fassadlr: Fixed :)

dangershony:
@codingupastorm totally agree on the comment for `SmartContractsController` this is not related to wallet, next PR is fine.

PR #1687, opened 2018-07-20
[SC] Separate CreateData and CallData

rowandh:
Separates CreateData and CallData from ContractTxData


src/Stratis.SmartContracts.Executor.Reflection/ContractTxData.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.GasPrice = gasPrice;
+            this.GasLimit = gasLimit;
+            this.ContractAddress = address;
+            this.MethodName = method;
+            this.MethodParametersRaw = rawParameters;            
+            this.MethodParameters = methodParameters;
+            this.ContractExecutionCode = new byte[0];
+        }
+
+        public ContractTxData(int vmVersion, ulong gasPrice, Gas gasLimit, byte[] code, string rawParameters = "",
```
codingupastorm: These are for create and call yes? Let's document that

src/Stratis.SmartContracts.Executor.Reflection/CallData.cs
```diff
@@ -1,74 +1,23 @@
...
 {
-    public struct CallData
+    /// <summary>
+    /// Smart contract transaction data for a method call
+    /// </summary>
+    public class CallData : ICallData
     {
-        public CallData(int vmVersion, ulong gasPrice, Gas gasLimit, uint160 address,
-            string method, string rawParameters = "", object[] methodParameters = null)
+        public CallData(Gas gasLimit, uint160 address, string method, object[] methodParameters = null)
```
fassadlr: This might be trivial but can we name address here `contractAddress`? I always confused with what address we refer to as and when.

fassadlr: Also, please move the properties to the top of the constructor so that its inline with FN.

src/Stratis.SmartContracts.Executor.Reflection/CreateData.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class CreateData : ICreateData
+    {
+        public CreateData(Gas gasLimit, byte[] code, object[] methodParameters = null)
+        {
+            this.GasLimit = gasLimit;
+            this.ContractExecutionCode = code;
+            this.MethodParameters = methodParameters;
+        }
+
+        public object[] MethodParameters { get; }
```
fassadlr: Properties on top of constructor.

src/Stratis.SmartContracts.Executor.Reflection/ContractTxData.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.GasLimit = gasLimit;
+            this.ContractExecutionCode = code;
+            this.MethodName = "";
+            this.MethodParametersRaw = rawParameters;
+            this.MethodParameters = methodParameters;
+            this.ContractAddress = uint160.Zero;
+        }
+
+        /// <summary>The method name of the contract that will be executed.</summary>
+        public string MethodName { get; }
```
fassadlr: Properties before constructor.

fassadlr: Also this might be trivial but I prefer the properties ordered alphabetically?

src/Stratis.Bitcoin.Features.SmartContracts/ReflectionExecutor/Consensus/Rules/SmartContractFormatRule.cs
```diff
@@ -69,7 +69,7 @@ private void CheckTransaction(Transaction transaction, Money suppliedBudget)
 
             if (callDataDeserializationResult.IsFailure)
             {
-                new ConsensusError("invalid-calldata-format", "Invalid CallData format").Throw();
+                new ConsensusError("invalid-calldata-format", "Invalid ContractTxData format").Throw();
```
fassadlr: You can change this to:

```
new ConsensusError("invalid-calldata-format", string.Format("Invalid {0} format", typeof(ContractTxData).Name)).Throw();
```

PR #1688, opened 2018-07-20
-reindex command now functions as expected.

### LINKED ISSUE #1215: -reindex-chainstate not working
I was attempting to retrieve a transaction's details via `getrawtransaction`. It transpires that this needs a transaction index, similarly to stratisX. When I shut down the full node and enabled this via the txindex command line flag, I receive the following message on startup:

`
You need to rebuild the BlockStore database using -reindex-chainstate to change -txindex) ---> Stratis.Bitcoin.Features.BlockStore.BlockStoreException: You need to rebuild the BlockStore database using -reindex-chainstate to change -txindex
`

However, adding the mentioned flag to the command line does not seem to cause the message to go away.

codingupastorm:
Fixes #1215.

There is likely potential to reuse code from `OnInsertBlocks` and `OnInsertTransactions`, however for now I believe it makes more sense to have a completely separate method as the logic is slightly different and I don't believe we need to involve the PerformanceCounter in any way.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
@@ -461,6 +461,84 @@ namespace Stratis.Bitcoin.Features.BlockStore.Tests
             }
         }
 
+        [Fact]
+        public void ReIndexAsync_TxIndex_OffToOn()
+        {
+            string dir = CreateTestDir(this);
+            Block block = Network.Main.Consensus.ConsensusFactory.CreateBlock();
+            Transaction transaction = Network.Main.CreateTransaction();
+            block.Transactions.Add(transaction);
+
+            using (var engine = new DBreezeEngine(dir))
+            {
+                DBreeze.Transactions.Transaction dbreezeTransaction = engine.GetTransaction();
+                dbreezeTransaction.Insert<byte[], byte[]>("Block", block.GetHash().ToBytes(), block.ToBytes());
+                dbreezeTransaction.Commit();
+            }
+
+            using (IBlockRepository repository = this.SetupRepository(Network.Main, dir))
+            {
+                Task setIndexTask = repository.SetTxIndexAsync(true);
+                setIndexTask.Wait();
+
+                Task reindexTask = repository.ReIndexAsync();
+                reindexTask.Wait();
+            }
+
+            using (var engine = new DBreezeEngine(dir))
+            {
+                DBreeze.Transactions.Transaction dbreezeTransaction = engine.GetTransaction();
+                Dictionary<byte[], Block> blockDict = dbreezeTransaction.SelectDictionary<byte[], Block>("Block");
+                Dictionary<byte[], byte[]> transDict = dbreezeTransaction.SelectDictionary<byte[], byte[]>("Transaction");
+
+                Assert.Single(blockDict);
+                Assert.Equal(block.GetHash(), blockDict.FirstOrDefault().Value.GetHash());
+
+                Assert.Single(transDict);
+                KeyValuePair<byte[], byte[]> savedTransactionRow = transDict.FirstOrDefault();
+                Assert.Equal(transaction.GetHash().ToBytes(), savedTransactionRow.Key);
+                Assert.Equal(block.GetHash().ToBytes(), savedTransactionRow.Value);
+            }
+        }
+
+        [Fact]
+        public void ReIndexAsync_TxIndex_OnToOff()
+        {
+            string dir = CreateTestDir(this);
+            Block block = Network.Main.Consensus.ConsensusFactory.CreateBlock();
+            Transaction transaction = Network.Main.CreateTransaction();
+            block.Transactions.Add(transaction);
+
+            using (var engine = new DBreezeEngine(dir))
+            {
+                DBreeze.Transactions.Transaction dbreezeTransaction = engine.GetTransaction();
+                dbreezeTransaction.Insert<byte[], byte[]>("Block", block.GetHash().ToBytes(), block.ToBytes());
+                dbreezeTransaction.Insert<byte[], byte[]>("Transaction", transaction.GetHash().ToBytes(), block.GetHash().ToBytes());
+                dbreezeTransaction.Commit();
+            }
+
+            using (IBlockRepository repository = this.SetupRepository(Network.Main, dir))
+            {
+                Task setIndexTask = repository.SetTxIndexAsync(false);
+                setIndexTask.Wait();
+
+                Task reindexTask = repository.ReIndexAsync();
+                reindexTask.Wait();
+            }
+
+            using (var engine = new DBreezeEngine(dir))
+            {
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -89,6 +98,12 @@ public interface IBlockRepository : IDisposable
 
     public class BlockRepository : IBlockRepository
     {
+        private const string BlockTableName = "Block";
```
fassadlr: Nice

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -114,8 +114,8 @@ public class BlockStoreQueue : IBlockStore, IDisposable
 
             if (this.storeSettings.ReIndex)
             {
-                this.logger.LogTrace("(-)[REINDEX_NOT_SUPPORTED]");
-                throw new NotSupportedException("Re-indexing the block store in currently not supported.");
+                await this.blockRepository.SetTxIndexAsync(this.storeSettings.TxIndex).ConfigureAwait(false);
+                await this.blockRepository.ReIndexAsync().ConfigureAwait(false);
```
fassadlr: Should TxIndex perhaps be set back after the ReIndex? (Not sure, just wondering)

src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
```diff
@@ -461,6 +461,84 @@ public void DeleteAsyncRemovesBlocksAndTransactions()
...
+
+        [Fact]
+        public void ReIndexAsync_TxIndex_OnToOff()
+        {
+            string dir = CreateTestDir(this);
+            Block block = Network.Main.Consensus.ConsensusFactory.CreateBlock();
+            Transaction transaction = Network.Main.CreateTransaction();
+            block.Transactions.Add(transaction);
+
+            using (var engine = new DBreezeEngine(dir))
```
fassadlr: @codingupastorm can you perhaps explain in the test with comments how this is tested?

src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
```diff
@@ -461,6 +461,84 @@ public void DeleteAsyncRemovesBlocksAndTransactions()
...
+                Assert.Equal(transaction.GetHash().ToBytes(), savedTransactionRow.Key);
+                Assert.Equal(block.GetHash().ToBytes(), savedTransactionRow.Value);
+            }
+        }
+
+        [Fact]
+        public void ReIndexAsync_TxIndex_OnToOff()
+        {
+            string dir = CreateTestDir(this);
+            Block block = Network.Main.Consensus.ConsensusFactory.CreateBlock();
```
fassadlr: Please use `Network.Main.CreateBlock();`

src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockRepositoryTest.cs
```diff
@@ -461,6 +461,84 @@ public void DeleteAsyncRemovesBlocksAndTransactions()
             }
         }
 
+        [Fact]
+        public void ReIndexAsync_TxIndex_OffToOn()
+        {
+            string dir = CreateTestDir(this);
+            Block block = Network.Main.Consensus.ConsensusFactory.CreateBlock();
```
fassadlr: Please use Network.Main.CreateBlock();

dangershony:
Quick view looks good to me, apart for @fassadlr comments.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
codingupastorm: The expected functionality is that we're setting the flag indefinitely. So we either always index transactions, or always don't. So the flag should stay set :)

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -69,12 +69,21 @@ public interface IBlockRepository : IDisposable
         /// <param name="trxid">The transaction hash</param>
         Task<uint256> GetTrxBlockIdAsync(uint256 trxid);
 
+        /// <summary>
+        /// For every block in the database, store the transactions (or don't) according to whether TxIndex is set.
```
Aprogiena: Use `<see>` or English for TxIndex. I did you want to reference some existing object by its name or you want to just say some thing about transaction index.

Aprogiena: From the description I couldn't really understand what it is supposed to do.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -322,12 +337,53 @@ protected virtual void OnInsertTransactions(DBreeze.Transactions.Transaction dbr
...
+        {
+            this.logger.LogTrace("()");
+
+            Task task = Task.Run(() =>
+            {
+                this.logger.LogTrace("()");
+
+                using (DBreeze.Transactions.Transaction dbreezeTransaction = this.DBreeze.GetTransaction())
+                {
+                    if (this.TxIndex)
```
Aprogiena: I'm thinking of putting 
``` transaction.SynchronizeTables(BlockTableName, TransactionTableName);```

here, what do thing?

Comment (codingupastorm) [reply src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs]: Makes sense. This should only be called on node startup so it shouldn't be a problem, but of course we don't know what future developers will try to do, so I'll add it in.

File: src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs (hunk 6 above)
Comment (codingupastorm) [reply src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs]: Hopefully better.

File: src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs [hunk 8]
```diff
@@ -70,7 +70,8 @@ public interface IBlockRepository : IDisposable
         Task<uint256> GetTrxBlockIdAsync(uint256 trxid);
 
         /// <summary>
-        /// For every block in the database, store the transactions (or don't) according to whether TxIndex is set.
+        /// Iterate over every block in the database. 
+        /// If <see cref="TxIndex"/> is true, we store the block hash alongside the transaction hash in the transaction table, otherwise clear the transaction table.
```
Aprogiena: `<c>true</c>`

Aprogiena: however, there is no TxIndx on the interface, so it is weird to have it in the comment

Aprogiena: ah sorry, there is TxIndex

PR #1691, opened 2018-07-20
Introduce the API feature to BitcoinD.

### LINKED ISSUE #1685: Add Api support to BitcoinD
At the moment, for some unknown reason, bitcoinD isn't using the Api feature.

ferdeen:
Fix #1685.

This was required so I can test syncing wallets ( #1103) when running BitcoinD. 

Confirm Api support :

image

And in Swagger :

image


noescape00:
Does all the API methods support btc network? Can you please check if every method works appropriately and doesn't throw? 
I'd expect potential problems with methods like getstakinginfo/startstaking/stopstaking. Also those methods should not be present in the API in case it's running on BTC network.

src/Stratis.BitcoinD/Program.cs
```diff
@@ -29,6 +30,7 @@ public class Program
                     .AddMining()
                     .AddRPC()
                     .UseWallet()
+                    .UseApi()
```
dangershony: Could we also add apps in that case?

dangershony:
@noescape00 I am not sure you will have those methods as they are not registered in the BTC mining feature.

src/Stratis.BitcoinD/Program.cs
ferdeen: Yes will do.

ferdeen:
Tested the majority of the web api calls (@dangershony & @noescape00).

- GetRawMempool didn't return an empty value
- StartStaking failed 
- The rest were fine - I didn't fully test all the wallet api call, as the majority are working as normal.

@dangershony  For the failures do we remove the web api calls ?  I don't want to spend too much time on this as this card was included as a quick win and simple task.

Test results in the below image.

image

noescape00:
Ok. So there are staking related api methods on BTC network? 
Those should be disabled.

ferdeen:
I am unable to hide the MiningFeature web API calls by network.  However, I have returned early out of the calls when on the the Bitcoin network :

image

ferdeen:
Sorry hold off, can see some related tests are failing.  Let me look into this.  Cheers.

noescape00:
>I am unable to hide the MiningFeature web API calls by network. However, I have returned early out of the calls when on the the Bitcoin network :

Check this out: 
https://stackoverflow.com/questions/38991603/how-can-i-disable-some-apis-of-my-asp-net-application

We can introduce attributes like [StratisOnly] and use them on staking methods.

ferdeen:
Trying out the `ActionFilterAttribute`.  

Problem I'm facing is trying to pass `this.fullNode...` as a parameter to the attribute.  I need to know what the network is to determine whether to bypass the action.

image

For now can we leave it with the check within the method calls, as this is stopping me from completing a more important task - #1103 ?

noescape00:
> as this is stopping me from completing a more important task - #1103 ?

Why? 1103 has nothing to do with API

>For now can we leave it with the check within the method calls

I don't mind. Create an issue for that and assign it to yourself and let's for now have current approach.

ferdeen:
> Why? 1103 has nothing to do with API

To make sure the wallet syncs I was going to run @bokobza's https://gist.github.com/bokobza/8b25a5edeb36f1343372c2d4cc14438c , and need to create wallets for each environment.

Had a chat with @fassadlr and he said it's better to remove `AddMining()` from Stratis.BitcoinD.   And also to rename `MinerController` to `StakingController` as this is what the class is actually doing.

ferdeen:
Cool @fassadlr.  I'll pick this up in the morning.  Cheers.

src/Stratis.Bitcoin.Features.SmartContracts/Stratis.Bitcoin.Features.SmartContracts.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="CSharpFunctionalExtensions" Version="1.10.0" />
```
dangershony: Whats this?

ferdeen: This was missing from the Stratis.Bitcoin.Features.SmartContracts.csproj file and stopped me from building this project.

Here is the original file from the initial SC commit - https://github.com/stratisproject/StratisBitcoinFullNode/commit/bf7c94d0d9959a6be79910ff072d7e805f9fb01e#diff-e1a79fe58674a2072ea55f05485e7987R7.

noescape00:
>@noescape00 The idea of removing AddMining() didn't work.

The idea was to introduce an attribute that will disable API methods on networks that are not supposed to support them. No need to remove AddMining itself.

ferdeen:
@noescape00 Yes, agreed.  This was a suggestion to try. 

For now can we have this PR merged then I can look at the attribute at a later date.

Aprogiena:
> Why having a suboptimal solution now and fix it later instead of implementing a proper one right away?

In this case I would agree that it is probably worth exploring now. In general it may often be reasonable to go with suboptimal temporary solution.

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingRPCController.cs
```diff
@@ -0,0 +1,106 @@
...
+using Stratis.Bitcoin.Controllers;
+using Stratis.Bitcoin.Features.Miner.Interfaces;
+using Stratis.Bitcoin.Features.Miner.Models;
+using Stratis.Bitcoin.Features.Wallet.Interfaces;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Miner.Controllers
+{
+    /// <summary>
+    /// RPC controller for calls related to PoW mining and PoS minting.
```
fassadlr: `/// RPC controller for calls related to PoS minting.`

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingRPCController.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// </summary>
+    [Controller]
+    public class StakingRpcController : FeatureController
```
fassadlr: Shouldn't `route` be specified here?

`[Route("api/[controller]")]`

fassadlr:
PS fix conflicts 👍

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingRPCController.cs
fassadlr: Oh no it shouldn't, this is the RPC controller...

ferdeen:
What we have done here is tidy up the RPC and API controller around staking/mining.  This allowed us to set the correct controllers for Mining and Staking features.

During this process we realised that all API controllers are discoverable in Swagger.   

So starting StratisD or BitcoinD you will always see all the API controllers (Staking and Mining) if at least one API controller is included within a feature setup.  We will need to tackle this at a later date and focus on sprint tasks.

I'm sorting out the conflicts right now.

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingRPCController.cs
ferdeen: cool

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingRPCController.cs
ferdeen: The code uses Route for controllers that inherit from `AspNetCore.MVC.Controller` (the API type controllers).

For RPC type controllers these inherit from `FeatureController`.

fassadlr:
@ferdeen please confirm staking can be started via the api calls as per usual.

ferdeen:
**StratisD Staking Test Evidence.**

`GetStakingInfo()`

image

`StartStaking()`

image

`StopStaking()`

image

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingApiController.cs
```diff
@@ -63,9 +59,6 @@ public StakingApiController(IFullNode fullNode, ILoggerFactory loggerFactory, IW
         [HttpGet]
         public IActionResult GetStakingInfo()
         {
-            if (this.fullNode.Network.IsBitcoin())
```
fassadlr: Ah dude,  this need to be put back, your test is failing cause you just need to set the network on the mocked FullNode object.

ferdeen: This was the code that I put in originally to return out of the Staking API methods on the Bitcoin
network.  

I have reverted this change and we have refactored the controllers so the feature setup for BitcoinD and StratisD will use the relevant controllers..

TjadenFroyda:
MvcBuilderExtensions.cs from Features.Api is used to automatically find all of the api endpoints. 

Could the network specific mining/staking endpoints be filtered there?

ferdeen:
@TjadenFroyda - This looks like the right place - or maybe attributes that get checked in MvcBuilderExtensions.cs.  

I'll add your suggestion to #1747.  

Thanks for letting us know.

fassadlr:
@TjadenFroyda perhaps you can look into this for us?

TjadenFroyda:
PR #1842  submitted

PR #1693, opened 2018-07-20
Remove node removes endpoint everywhere

monsieurleberre:
This makes the remove node logic more reliable
The test in NodeSyncTest was often failing with a timeout because it keeps its address in the ConnectedPeers
I have added implementation of IDisposable in NodeSettings and move the disposal of LoggerFactory after the disposal of features which use it in FullNode.Dispose()


monsieurleberre:
Usual suspects ! MiningNodeWith ... is fixed in another pending PR
****
image

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -49,19 +51,21 @@ public void CanStratisSyncFromCore()
...
+                TestHelper.WaitLoop(() => stratisNodeRpcClient.GetBestBlockHash() == coreNodeRpcClient.GetBestBlockHash());
+                uint256 bestBlockHash = stratisNodeRpcClient.GetBestBlockHash();
                 Assert.Equal(tip.GetHash(), bestBlockHash);
 
-                //Now check if Core connect to stratis
-                stratisNode.CreateRPCClient().RemoveNode(coreNode.Endpoint);
-                TestHelper.WaitLoop(() => coreNode.CreateRPCClient().GetPeersInfo().Length == 0);
+                stratisNodeRpcClient.RemoveNode(coreNode.Endpoint);
+                TestHelper.WaitLoop(() => coreNodeRpcClient.GetPeersInfo()
+                    .All(pi => pi.Address.MapToIpv6().ToString() != coreNode.Endpoint.MapToIpv6().ToString()));
```
dangershony: I am not sure ts required to map to ip v6 all addresses should already be in that format

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -166,7 +166,11 @@ public sealed class BaseFeature : FullNodeFeature, INodeStats
             this.loggerFactory = loggerFactory;
             this.dbreezeSerializer = dbreezeSerializer;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
-            this.disposableResources = new List<IDisposable>();
+            this.disposableResources = new List<IDisposable>()
```
dangershony: I think `loggerFactory` components are already disposed elsewhere

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -421,7 +360,16 @@ public void RemoveNodeAddress(IPEndPoint ipEndpoint)
...
-            peer?.Disconnect("Requested by user");
+
+            if (peer != null)
+            {
+                peer.Disconnect("Requested by user");
+                this.RemoveConnectedPeer(peer, "Requested by user");
+            }
+
+            this.peerAddressManager.RemovePeer(ipEndpoint.MapToIpv6(), IPAddress.Loopback);
+            List<IPEndPoint> matchingAddNodes = this.ConnectionSettings.AddNode.Where(p => p.Match(ipEndpoint)).ToList();
```
dangershony: Why this funny linq you are enumerating several times.

Also do we need to to the same for -connet? (I am not sure just asking)
```
this.ConnectionSettings.Connect
```

src/Stratis.Bitcoin/Connection/IConnectionManager.cs
```diff
@@ -0,0 +1,76 @@
...
+
+    using NBitcoin;
+
+    using Stratis.Bitcoin.Configuration;
+    using Stratis.Bitcoin.Configuration.Settings;
+    using Stratis.Bitcoin.P2P;
+    using Stratis.Bitcoin.P2P.Peer;
+    using Stratis.Bitcoin.P2P.Protocol.Payloads;
+
+    public interface IConnectionManager : IDisposable
```
dangershony: If you moved this out then please add comments as well.

src/Stratis.Bitcoin/FullNode.cs
```diff
@@ -277,14 +277,15 @@ public void Dispose()
...
             this.bestChainSelector.Dispose();
-            this.loggerFactory.Dispose();
 
             foreach (IDisposable disposable in this.Resources)
                 disposable.Dispose();
 
             // Fire the NodeFeatureExecutor.Stop.
             this.fullNodeFeatureExecutor.Dispose();
 
+            this.loggerFactory.Dispose();
```
dangershony: Hmm this is odd, loggerFactory needs to be disposed last so I find it weird it was not already here(I remember putting it here some time ago)

In any case there is only one logger factory so not need to dispose of it several times, so remove it from the `this.Resources`

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -176,6 +103,16 @@ public void AddPeer(IPEndPoint endPoint, IPAddress source)
             this.peers.TryAdd(peerToAdd.Endpoint, peerToAdd);
         }
 
+        public void RemovePeer(IPEndPoint endPoint, IPAddress source)
```
dangershony: I find it odd that you need to add `IPAddress source`  this should be inside IPEndPoint no?

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -371,5 +371,10 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
 
             ConnectionManagerSettings.BuildDefaultConfigurationFile(builder, network);
         }
+
+        public void Dispose()
+        {
+            this.LoggerFactory?.Dispose();
```
Aprogiena: why `?`

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -371,5 +371,10 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
 
             ConnectionManagerSettings.BuildDefaultConfigurationFile(builder, network);
         }
+
+        public void Dispose()
```
Aprogiena: inheritdoc

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
Aprogiena: Also when you do this you should ensure that it is not this post elsewhere, which it is

src/Stratis.Bitcoin/FullNode.cs
Aprogiena: This should be replaced with disposal of node settings

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: Yes, it does not make sense.

src/Stratis.Bitcoin/P2P/IPeerAddressManager.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        /// <summary>
+        /// Add a set of peers to the <see cref="Peers"/> dictionary.
+        /// <para>
+        /// Only routable IP addresses will be added. <see cref="IpExtensions.IsRoutable(IPAddress, bool)"/>
+        /// </para>
+        /// </summary>
+        void AddPeers(IPEndPoint[] endPoints, IPAddress source);
+
+        void RemovePeer(IPEndPoint endPoint, IPAddress source);
```
Aprogiena: Comment

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: Also inheritdoc

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -176,6 +103,16 @@ public void AddPeer(IPEndPoint endPoint, IPAddress source)
...
+        {
+            if (!endPoint.Address.IsRoutable(true))
```
Aprogiena: Why do we have this condition?

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
```diff
@@ -176,6 +103,16 @@ public void AddPeer(IPEndPoint endPoint, IPAddress source)
...
+                return;
+
+            PeerAddress peer = PeerAddress.Create(endPoint, source);
+            this.peers.TryRemove(peer.Endpoint, out PeerAddress address);
```
Aprogiena: Why do you create peer address object when you only used endpoint from which you already had before

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -11,6 +11,8 @@
 
 namespace Stratis.Bitcoin.IntegrationTests
 {
+    using Stratis.Bitcoin.Utilities.Extensions;
```
Aprogiena: Code style

src/Stratis.Bitcoin/Connection/IConnectionManager.cs
monsieurleberre: I don't understand the logic behind your comment ?
If I separate the interface I have to comment it, but if it is in the same file as the class I don't ? Why ?

src/Stratis.Bitcoin/FullNode.cs
monsieurleberre: I don't think it is added to this.Resources, and yes agreed I should dispose of the nodeSettings instead

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
monsieurleberre: If endpoint was not Routable it has not been added to the peers list, so no need to go further right ?

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
monsieurleberre: because doing this I go through the same steps as the AddPeer method which ensure that the IP is IPV6

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
monsieurleberre: yes agreed it feels weird, but this is only to go through the same code path as for AddPeer, for symmetry

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
monsieurleberre: I think it is not needed, Connect seems to not be altered when we call AddNodeAddress, however AddNode is modified
image

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: ok, I don't think it is needed, but ok. For me this is something that one has to think about when changing add peer method and does not really solve any problem, so I would not use it

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: That seemed to be quite a bad argument. How about you just make sure yourself?

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: Okay, remove it

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
Aprogiena: Still the question holds - Why the question mark?

src/Stratis.Bitcoin/Connection/IConnectionManager.cs
```diff
@@ -68,7 +68,11 @@ public interface IConnectionManager : IDisposable
 
         /// <summary>Connection settings.</summary>
         ConnectionManagerSettings ConnectionSettings { get; }
-
+        
+        /// <summary>
+        /// Remove a peer to the address manager's collection as well as
```
Aprogiena: Remove from instead of to

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -368,8 +369,8 @@ public void RemoveNodeAddress(IPEndPoint ipEndpoint)
             }
 
             this.peerAddressManager.RemovePeer(ipEndpoint.MapToIpv6(), IPAddress.Loopback);
-            List<IPEndPoint> matchingAddNodes = this.ConnectionSettings.AddNode.Where(p => p.Match(ipEndpoint)).ToList();
-            matchingAddNodes.ForEach(m => this.ConnectionSettings.AddNode.Remove(m));
+            IEnumerable<IPEndPoint> matchingAddNodes = this.ConnectionSettings.AddNode.Where(p => p.Match(ipEndpoint));
+            foreach (IPEndPoint m in matchingAddNodes) this.ConnectionSettings.AddNode.Remove(m);
```
Aprogiena: I like this change, though you could put the body on the separated line as we always do with loops

src/Stratis.Bitcoin/P2P/IPeerAddressManager.cs
```diff
@@ -32,7 +32,13 @@ public interface IPeerAddressManager : IDisposable
         /// </para>
         /// </summary>
         void AddPeers(IPEndPoint[] endPoints, IPAddress source);
-
+        
+        /// <summary>
+        /// Remove a peer from the <see cref="Peers"/> dictionary.
+        /// <para>
+        /// Only routable IP addresses will be removed since only routable IP addressed can be added. <see cref="IpExtensions.IsRoutable(IPAddress, bool)"/>
```
Aprogiena: use `<seealso>` instead of `<see>` and put it OUTSIDE of summary section

monsieurleberre: Looking at the comments on the 2 methods above, this looks in line with the current way of doing things...
Hopefully not a blocker...

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
monsieurleberre: then you have to ensure that both methods rely on IPV6, which I haven't found a better way to enforce

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
monsieurleberre: OK Can I do that then to ensure it is not set to null somewhere else in the class ?
image

monsieurleberre:
image

src/Stratis.Bitcoin/P2P/IPeerAddressManager.cs
Aprogiena: correctly used on line 23, incorrectly on 31 and here

src/Stratis.Bitcoin/P2P/PeerAddressManager.cs
Aprogiena: there is an extension method for that, something like ensureipv6 or something

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
Aprogiena: i'll check if it can be null

Aprogiena: https://github.com/monsieurleberre/StratisBitcoinFullNode/blob/cfba559d0f75d5abf4d302a4087939b4664b034e/src/Stratis.Bitcoin/Configuration/NodeSettings.cs#L93 - seems to be the only constructor and it sets up the factory, so it can NOT be null, we can remove `?`

monsieurleberre:
still OK
image

monsieurleberre:
I created and linked that to this issue : https://github.com/stratisproject/StratisBitcoinFullNode/issues/1744

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
dangershony: I agree try to put bodies in a new line especially when its an operation.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -421,7 +361,16 @@ public void RemoveNodeAddress(IPEndPoint ipEndpoint)
...
+
+            if (peer != null)
+            {
+                peer.Disconnect("Requested by user");
+                this.RemoveConnectedPeer(peer, "Requested by user");
+            }
+
+            this.peerAddressManager.RemovePeer(ipEndpoint);
+            IEnumerable<IPEndPoint> matchingAddNodes = this.ConnectionSettings.AddNode.Where(p => p.Match(ipEndpoint));
+            foreach (IPEndPoint m in matchingAddNodes) this.ConnectionSettings.AddNode.Remove(m);
```
dangershony: New line! FFS

PR #1700, opened 2018-07-21
[SC] Compile multiple files within a directory. Option to compile specific namespace.

codingupastorm:
Our compiler util previously only allowed the compilation of a specific file or string.

This just updates the API so we can now compile an entire directory, with the option of selecting by namespace too. 

Seems likely that initially, we only allow a single namespace and when creating another contract it should also be in the same namespace. 

Note this doesn't update the Sct yet, just gives us the option to in the future though.


rowandh:
Looks good. I imagine in the future we'll want to define something to resolve locations of files.

PR #1701, opened 2018-07-21
Staking vulnerability

monsieurleberre:
This is just to avoid taking coins that are less than .1 Strat
Next PR will include a few tests, I'd rather get these signature changes merged in if possible
image
Unit and Integration tests all good apart from usual suspects, CanStratisSyncFromCore is fixed in another PR

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewStackTest.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewStackTest.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using NBitcoin;
 using Stratis.Bitcoin.Features.Consensus.CoinViews;
@@ -115,7 +116,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests.CoinViews
             {
             }
 
-            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
             {
                 throw new NotImplementedException();
             }
@@ -144,7 +145,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests.CoinViews
             public int OutputCount { get; }
             public CoinView Inner => this.inner;
 
-            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
             {
                 throw new NotImplementedException();
             }
@@ -173,7 +174,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests.CoinViews
             public int OutputCount { get; }
             public CoinView Inner => this.inner;
 
-            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
             {
                 throw new NotImplementedException();
             }
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
@@ -10,23 +11,23 @@ using Stratis.Bitcoin.Utilities;
 namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 {
     /// <summary>
-    /// Cache layer for coinview prevents too frequent updates of the data in the underlaying storage.
+    /// Cache layer for coinview prevents too frequent updates of the data in the underlying storage.
     /// </summary>
     public class CachedCoinView : CoinView, IBackedCoinView, IDisposable
     {
         /// <summary>
         /// Item of the coinview cache that holds information about the unspent outputs
-        /// as well as the status of the item in relation to the underlaying storage.
+        /// as well as the status of the item in relation to the underlying storage.
         /// </summary>
         private class CacheItem
         {
             /// <summary>Information about transaction's outputs. Spent outputs are nulled.</summary>
             public UnspentOutputs UnspentOutputs;
 
-            /// <summary><c>true</c> if the unspent output information is stored in the underlaying storage, <c>false</c> otherwise.</summary>
+            /// <summary><c>true</c> if the unspent output information is stored in the underlying storage, <c>false</c> otherwise.</summary>
             public bool ExistInInner;
 
-            /// <summary><c>true</c> if the information in the cache is different than the information in the underlaying storage.</summary>
+            /// <summary><c>true</c> if the information in the cache is different than the information in the underlying storage.</summary>
             public bool IsDirty;
 
             /// <summary>Original state of the transaction outputs before the change. This is used for rewinding to previous state.</summary>
@@ -113,7 +114,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// <summary>
         /// Initializes instance of the object based on DBreeze based coinview.
         /// </summary>
-        /// <param name="inner">Underlaying coinview with database storage.</param>
+        /// <param name="inner">Underlying coinview with database storage.</param>
         /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="loggerFactory">Factory to be used to create logger for the puller.</param>
         /// <param name="stakeChainStore">Storage of POS block information.</param>
@@ -127,7 +128,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// <summary>
         /// Initializes instance of the object based on memory based coinview.
         /// </summary>
-        /// <param name="inner">Underlaying coinview with memory based storage.</param>
+        /// <param name="inner">Underlying coinview with memory based storage.</param>
         /// <param name="dateTimeProvider">Provider of time functions.</param>
         /// <param name="loggerFactory">Factory to be used to create logger for the puller.</param>
         /// <param name="stakeChainStore">Storage of POS block information.</param>
@@ -161,7 +162,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         }
 
         /// <inheritdoc />
-        public override async Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+        public override async Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
         {
             Guard.NotNull(txIds, nameof(txIds));
             this.logger.LogTrace("({0}.{1}:{2})", nameof(txIds), nameof(txIds.Length), txIds.Length);
@@ -170,7 +171,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
             var outputs = new UnspentOutputs[txIds.Length];
             var miss = new List<int>();
             var missedTxIds = new List<uint256>();
-            using (await this.lockobj.LockAsync().ConfigureAwait(false))
+            using (await this.lockobj.LockAsync(ct).ConfigureAwait(false))
             {
                 for (int i = 0; i < txIds.Length; i++)
                 {
@@ -194,10 +195,10 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
                 this.PerformanceCounter.AddHitCount(txIds.Length - miss.Count);
             }
 
-            this.logger.LogTrace("{0} cache missed transaction needs to be loaded from underlaying coinview.", missedTxIds.Count);
-            FetchCoinsResponse fetchedCoins = await this.Inner.FetchCoinsAsync(missedTxIds.ToArray()).ConfigureAwait(false);
+            this.logger.LogTrace("{0} cache missed transaction needs to be loaded from underlying CoinView.", missedTxIds.Count);
+            FetchCoinsResponse fetchedCoins = await this.Inner.FetchCoinsAsync(missedTxIds.ToArray(), ct).ConfigureAwait(false);
 
-            using (await this.lockobj.LockAsync().ConfigureAwait(false))
+            using (await this.lockobj.LockAsync(ct).ConfigureAwait(false))
             {
                 uint256 innerblockHash = fetchedCoins.BlockHash;
                 if (this.blockHash == null)
@@ -234,7 +235,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         }
 
         /// <summary>
-        /// Finds all changed records in the cache and persists them to the underlaying coinview.
+        /// Finds all changed records in the cache and persists them to the underlying CoinView.
         /// </summary>
         /// <param name="force"><c>true</c> to enforce flush, <c>false</c> to flush only if <see cref="lastCacheFlushTime"/> is older than <see cref="CacheFlushTimeIntervalSeconds"/>.</param>
         /// <remarks>
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using NBitcoin;
 using Stratis.Bitcoin.Utilities;
@@ -14,9 +15,9 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// Retrieves the block hash of the current tip of the coinview.
         /// </summary>
         /// <returns>Block hash of the current tip of the coinview.</returns>
-        public async Task<uint256> GetBlockHashAsync()
+        public async Task<uint256> GetBlockHashAsync(CancellationToken ct = default(CancellationToken))
         {
-            FetchCoinsResponse response = await this.FetchCoinsAsync(new uint256[0]).ConfigureAwait(false);
+            FetchCoinsResponse response = await this.FetchCoinsAsync(new uint256[0], ct).ConfigureAwait(false);
             return response.BlockHash;
         }
 
@@ -50,7 +51,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         /// If the i-th item of <see cref="FetchCoinsResponse.UnspentOutputs"/> is <c>null</c>, it means that there are no unspent outputs in the given transaction.
         /// </para>
         /// </returns>
-        public abstract Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds);
+        public abstract Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken));
 
         /// <summary>
         /// Rewinds the coinview to the last saved state.
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using DBreeze;
 using DBreeze.DataTypes;
@@ -114,7 +115,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         }
 
         /// <inheritdoc />
-        public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+        public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
         {
             Task<FetchCoinsResponse> task = Task.Run(() =>
             {
@@ -146,7 +147,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 
                 this.logger.LogTrace("(-):*.{0}='{1}',*.{2}.{3}={4}", nameof(res.BlockHash), res.BlockHash, nameof(res.UnspentOutputs), nameof(res.UnspentOutputs.Length), res.UnspentOutputs.Length);
                 return res;
-            });
+            }, ct);
 
             return task;
         }
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -249,6 +249,12 @@ namespace Stratis.Bitcoin.Features.Miner
         /// <summary>Indicates that the stake flag is in progress.</summary>
         public const int StakeInProgress = 1;
 
+        /// <summary>
+        /// We don't take coins that are smaller that 0.1 as described in
+        /// <see cref="https://github.com/stratisproject/StratisBitcoinFullNode/issues/1180"/>
+        /// </summary>
+        public const long MinimumStakingCoinValue = 10 * Money.CENT;
+
         /// <summary>a flag that indicates if stake is on/off based on the <see cref="StakeInProgress"/> and <see cref="StakeNotInProgress"/> constants.</summary>
         private int stakeProgressFlag;
 
@@ -509,16 +515,8 @@ namespace Stratis.Bitcoin.Features.Miner
                     return;
                 }
 
-                IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
-                    .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
-                    .Where(t => t != null).ToList();
-
-                FetchCoinsResponse coinset = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray()).ConfigureAwait(false);
-
-                List<UtxoStakeDescription> utxoStakeDescriptions = GetUtxoStakeDescriptions(walletSecret, spendableTransactions, coinset);
-
-                this.logger.LogTrace("Wallet total staking balance is {0}.", new Money(utxoStakeDescriptions.Sum(d => d.TxOut.Value)));
-
+                List<UtxoStakeDescription> utxoStakeDescriptions = await GetUtxoStakeDescriptions(walletSecret, this.stakeCancellationTokenSource.Token);
+                
                 blockTemplate = blockTemplate ?? this.blockProvider.BuildPosBlock(chainTip, new Script());
                 var posBlock = (PosBlock)blockTemplate.Block;
                 
@@ -545,14 +543,17 @@ namespace Stratis.Bitcoin.Features.Miner
             }
         }
 
-        /// <summary>
-        /// We don't take coins that are smaller that 0.1 as described in
-        /// <see cref="https://github.com/stratisproject/StratisBitcoinFullNode/issues/1180"/>>
-        /// </summary>
-        private List<UtxoStakeDescription> GetUtxoStakeDescriptions(WalletSecret walletSecret,
-                IEnumerable<UnspentOutputReference> spendableTransactions, FetchCoinsResponse fetchedCoinSet)
+        private async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptions(WalletSecret walletSecret, CancellationToken ct)
         {
             var utxoStakeDescriptions = new List<UtxoStakeDescription>();
+            IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+                .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
+                .Where(t => t != null).ToList();
+
+            //todo 
+            FetchCoinsResponse fetchedCoinSet = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray(), ct).ConfigureAwait(false);
+            if (ct.IsCancellationRequested) return utxoStakeDescriptions;
+
             foreach (UnspentOutputReference outputReference in spendableTransactions)
             {
                 UnspentOutputs coinSet = fetchedCoinSet.UnspentOutputs
@@ -560,7 +561,7 @@ namespace Stratis.Bitcoin.Features.Miner
                 if (coinSet == null || outputReference.Transaction.Index >= coinSet.Outputs.Length) continue;
 
                 TxOut utxo = coinSet.Outputs[outputReference.Transaction.Index];
-                if (utxo == null || utxo.Value <= 10 * Money.CENT) continue;
+                if (utxo == null || utxo.Value <= MinimumStakingCoinValue) continue;
 
                 uint256 hashBlock = this.chain.GetBlock((int)coinSet.Height)?.HashBlock;
                 if (hashBlock == null) continue;
@@ -577,8 +578,11 @@ namespace Stratis.Bitcoin.Features.Miner
                 utxoStakeDescriptions.Add(utxoStakeDescription);
 
                 this.logger.LogTrace("UTXO '{0}' with value {1} might be available for staking.", utxoStakeDescription.OutPoint, utxo.Value);
+
+                if (ct.IsCancellationRequested) return utxoStakeDescriptions;
             }
 
+            this.logger.LogTrace("Wallet total staking balance is {0}.", new Money(utxoStakeDescriptions.Sum(d => d.TxOut.Value)));
             return utxoStakeDescriptions;
         }
 
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -509,65 +509,77 @@ namespace Stratis.Bitcoin.Features.Miner
                     return;
                 }
 
-                var utxoStakeDescriptions = new List<UtxoStakeDescription>();
-                IEnumerable<UnspentOutputReference> spendable = this.walletManager.GetSpendableTransactionsInWallet(walletSecret.WalletName, 1);
+                IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+                    .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
+                    .Where(t => t != null).ToList();
 
-                FetchCoinsResponse coinset = await this.coinView.FetchCoinsAsync(spendable.Select(t => t.Transaction.Id).ToArray()).ConfigureAwait(false);
+                FetchCoinsResponse coinset = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray()).ConfigureAwait(false);
 
-                long totalBalance = 0;
-                foreach (UnspentOutputReference infoTransaction in spendable)
-                {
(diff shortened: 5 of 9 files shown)


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -161,7 +162,7 @@ private CachedCoinView(IDateTimeProvider dateTimeProvider, ILoggerFactory logger
         }
 
         /// <inheritdoc />
-        public override async Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+        public override async Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
```
Aprogiena: We rather use long names like cancellationToken instead of ct, consider replacement, although it is not required

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -234,7 +235,7 @@ private CachedCoinView(IDateTimeProvider dateTimeProvider, ILoggerFactory logger
         }
 
         /// <summary>
-        /// Finds all changed records in the cache and persists them to the underlaying coinview.
+        /// Finds all changed records in the cache and persists them to the underlying CoinView.
```
Aprogiena: Why `CoinView` instead of `coinview`? If you want to reference an object, you're supposed to use `<see>`, but I don't think that was intended here

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -170,7 +171,7 @@ private CachedCoinView(IDateTimeProvider dateTimeProvider, ILoggerFactory logger
             var outputs = new UnspentOutputs[txIds.Length];
             var miss = new List<int>();
             var missedTxIds = new List<uint256>();
-            using (await this.lockobj.LockAsync().ConfigureAwait(false))
+            using (await this.lockobj.LockAsync(ct).ConfigureAwait(false))
```
Aprogiena: Why did you do that? Is there a deeper thought behind that or is it just random improvement? You do realise that you introduced exception to the method that is not handled anywhere?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -249,6 +249,12 @@ public class CoinstakeContext
         /// <summary>Indicates that the stake flag is in progress.</summary>
         public const int StakeInProgress = 1;
 
+        /// <summary>
+        /// We don't take coins that are smaller that 0.1 as described in
+        /// <see cref="https://github.com/stratisproject/StratisBitcoinFullNode/issues/1180"/>
```
Aprogiena: This should be in `<seealso>` section in the summary. Also `s`take.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
-
-                if (blockTemplate == null)
-                    blockTemplate = this.blockProvider.BuildPosBlock(chainTip, new Script());
-
-                if (!(blockTemplate.Block is PosBlock block))
-                {
-                    throw new InvalidCastException();
-                }
-
+                List<UtxoStakeDescription> utxoStakeDescriptions = await GetUtxoStakeDescriptions(walletSecret, this.stakeCancellationTokenSource.Token);
```
Aprogiena: Configure await

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+                               Address = outputReference.Address,
+                               HashBlock = hashBlock,
+                               UtxoSet = coinSet,
+                               Secret = walletSecret // Temporary.
+                           };
+                utxoStakeDescriptions.Add(utxoStakeDescription);
+
+                this.logger.LogTrace("UTXO '{0}' with value {1} might be available for staking.", utxoStakeDescription.OutPoint, utxo.Value);
+
+                if (ct.IsCancellationRequested) return utxoStakeDescriptions;
```
Aprogiena: not really needed?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
```diff
@@ -50,7 +51,7 @@ public abstract class CoinView
         /// If the i-th item of <see cref="FetchCoinsResponse.UnspentOutputs"/> is <c>null</c>, it means that there are no unspent outputs in the given transaction.
         /// </para>
         /// </returns>
-        public abstract Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds);
+        public abstract Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken));
```
Aprogiena: would be nice if you added parameter comment if you are adding parameter to a commented method

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
```diff
@@ -14,9 +15,9 @@ public abstract class CoinView
         /// Retrieves the block hash of the current tip of the coinview.
         /// </summary>
         /// <returns>Block hash of the current tip of the coinview.</returns>
-        public async Task<uint256> GetBlockHashAsync()
+        public async Task<uint256> GetBlockHashAsync(CancellationToken ct = default(CancellationToken))
```
Aprogiena: would be nice if you added parameter comment if you are adding parameter to a commented method

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
                 else
                 {
-                    this.logger.LogTrace("{0} failed, waiting {1} ms for next round...", nameof(this.StakeAndSignBlockAsync), this.minerSleep);
+                    this.logger.LogTrace("{0} failed to create POS block, waiting {1} ms for next round...", nameof(this.StakeAndSignBlockAsync), this.minerSleep);
                     await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
                 }
             }
         }
 
+        private async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptions(WalletSecret walletSecret, CancellationToken ct)
```
Aprogiena: good to separate, but would be nice if the logic was then commented as well

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+        {
+            var utxoStakeDescriptions = new List<UtxoStakeDescription>();
+            IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+                .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
+                .Where(t => t != null).ToList();
+
+            //todo 
+            FetchCoinsResponse fetchedCoinSet = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray(), ct).ConfigureAwait(false);
+            if (ct.IsCancellationRequested) return utxoStakeDescriptions;
```
Aprogiena: you are unlikely to hit here as the exception would be thrown on the previous line, so no need for this

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+
+            foreach (UnspentOutputReference outputReference in spendableTransactions)
+            {
+                UnspentOutputs coinSet = fetchedCoinSet.UnspentOutputs
+                    .FirstOrDefault(f => f.TransactionId == outputReference.Transaction.Id);
```
Aprogiena: looks like you have removed `?` here, are you sure it can't be null?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: though still ... another PR

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: also the original line was not that much long, you didn't need to split it

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
monsieurleberre: I generally go long names, so yes I will change it no big deal

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
monsieurleberre: yes I will change this one back agreed

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
monsieurleberre: Honestly I don't see any relevant comment I can put here :(

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
monsieurleberre: Honestly I don't see any relevant comment I can put here :(

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: OK will do, I don't understand what you mean by "Also s take" do you just want to replace the verb "take" with "stake" ?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: OK Good point

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: I don't agree with the "other PR", I think if we refrain from  doing minor refactor on the code just because is not *strictly* within the scope of the PR, then we can't afford to improve with small increments like I am trying to do here

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
         }
 
+        private async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptions(WalletSecret walletSecret, CancellationToken ct)
+        {
+            var utxoStakeDescriptions = new List<UtxoStakeDescription>();
+            IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+                .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
+                .Where(t => t != null).ToList();
+
+            //todo
```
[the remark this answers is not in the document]
monsieurleberre: yes it needs to be removed

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
                 }
             }
         }
 
+        private async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptions(WalletSecret walletSecret, CancellationToken ct)
+        {
+            var utxoStakeDescriptions = new List<UtxoStakeDescription>();
+            IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+                .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
+                .Where(t => t != null).ToList();
```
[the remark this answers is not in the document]
monsieurleberre: please clarify

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+            if (ct.IsCancellationRequested) return utxoStakeDescriptions;
+
+            foreach (UnspentOutputReference outputReference in spendableTransactions)
+            {
+                UnspentOutputs coinSet = fetchedCoinSet.UnspentOutputs
+                    .FirstOrDefault(f => f.TransactionId == outputReference.Transaction.Id);
+                if (coinSet == null || outputReference.Transaction.Index >= coinSet.Outputs.Length) continue;
+
+                TxOut utxo = coinSet.Outputs[outputReference.Transaction.Index];
+                if (utxo == null || utxo.Value <= MinimumStakingCoinValue) continue;
```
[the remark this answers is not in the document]
monsieurleberre: which one ? Honestly I don't know
- A single line statement block can go without braces, one line statements are allowed (without braces) if it makes readability better ? this one looks fine
- We add parentheses to group the parts of complex boolean expressions ? surely this is *not* complex for our high caliber developers 😂

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+
+                var utxoStakeDescription = new UtxoStakeDescription
+                           {
+                               TxOut = utxo,
+                               OutPoint = new OutPoint(coinSet.TransactionId, outputReference.Transaction.Index),
+                               Address = outputReference.Address,
+                               HashBlock = hashBlock,
+                               UtxoSet = coinSet,
+                               Secret = walletSecret // Temporary.
+                           };
```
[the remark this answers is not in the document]
monsieurleberre: same here, no clue

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +515,77 @@ public void StopStake()
...
+                utxoStakeDescriptions.Add(utxoStakeDescription);
+
+                this.logger.LogTrace("UTXO '{0}' with value {1} might be available for staking.", utxoStakeDescription.OutPoint, utxo.Value);
+
+                if (ct.IsCancellationRequested) return utxoStakeDescriptions;
+            }
+
+            this.logger.LogTrace("Wallet total staking balance is {0}.", new Money(utxoStakeDescriptions.Sum(d => d.TxOut.Value)));
+            return utxoStakeDescriptions;
```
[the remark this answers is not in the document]
monsieurleberre: what is wrong ? sorry I can't see

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: Oh F** yes I think it can actually be null you are right, just saw a way in DBreeze coinview implementation, thanks

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: just to exit a little quicker if we need to, I'm not too fussed honestly

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: Yes, please do everywhere, but as mentioned, I would still prefer this to be on separate PR and not on the PR that should fix a vulnerability

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
Aprogiena: try `Cancellation token to allow the caller to cancel the execution of the operation.`
or `Cancellation token to observe while waiting for the task to complete.`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
Aprogiena: See above

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: We don't take coins that are smaller that 0.1
->
We don't `s`take coins that are smaller tha`n` 0.1

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: The problem is that I'm here for staking vulnerability PR, instead and checking completely different functionality which maybe does not need my review. Those incremental improvement possibly can be checked by someone else, while the staking vulnerability I definitely want to check myself. Most of the comments I have towards this PR are completely irrelevant to what was supposed to be the primary objective of this PR. You can afford to make incremental improvements, but do it on separated PR. For example I quite like introduction of cancellation token, but totally hate it on this PR.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: You have changed this line in a way that I don't understand.

- You are using IEnumerable but at the end you are converting to list, so why not using the list?

- You have added condition for the elements not being null, but is it really needed? The previous version did not have that, right? So do you think it can be null somewhere and that the previous version was buggy?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: brackets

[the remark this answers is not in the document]
Aprogiena: It is not about the complexity, see the rule examples

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: Crazy indentation? Look elsewhere into the code or even to the previous version of this code

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I think it would only be needed if there is an expensive operation inside the loop, which I think it is not

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: Entry and exit logs are missing

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: if you want... OK

src/Stratis.Bitcoin.Features.Miner/Interfaces/IPosMinting.cs
```diff
@@ -29,7 +30,7 @@ public interface IPosMinting
         /// </para>
         /// </summary>
         /// <param name="walletSecret">Credentials to the wallet with which will be used for staking.</param>
-        Task GenerateBlocksAsync(PosMinting.WalletSecret walletSecret);
+        Task GenerateBlocksAsync(PosMinting.WalletSecret walletSecret, CancellationToken ct);
```
Aprogiena: Didn't we just agree that we will use longer names?

And again parameter comment

Aprogiena: But would you please stop expanding this PR?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
Aprogiena: why IList, you have List already

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -538,21 +540,20 @@ public void StopStake()
...
-            IEnumerable<UnspentOutputReference> spendableTransactions = this.walletManager
+            IList<UnspentOutputReference> spendableTransactions = this.walletManager
                 .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1)
                 .Where(t => t != null).ToList();
 
-            //todo 
-            FetchCoinsResponse fetchedCoinSet = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray(), ct).ConfigureAwait(false);
-            if (ct.IsCancellationRequested) return utxoStakeDescriptions;
+            FetchCoinsResponse fetchedCoinSet = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray(), cancellationToken).ConfigureAwait(false);
+            if (cancellationToken.IsCancellationRequested) return utxoStakeDescriptions;
```
Aprogiena: this should really NOT be needed

monsieurleberre: In my opinion, if you want to cancel the task, then why would you bother to go through the other spendableTransactions in the loop ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
monsieurleberre: Honestly I think this would just add noise, cancellationTokens are always passed in for that purpose, so unless we use it in an unconventional way, why explain ?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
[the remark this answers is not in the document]
monsieurleberre: OK you are right, it looks like the previous code rightly assumes non null values from "GetSpendableTransactions", I am OK with removing it, but I think if the code changed and a new implementation of the IWalletManager could return null in its output reference then we would be happy to have this cheap safety in place... It really doesn't shock me but I can remove it OK...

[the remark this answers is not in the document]
Aprogiena: Yes, please remove it

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -568,14 +569,14 @@ public void StopStake()
...
                 var utxoStakeDescription = new UtxoStakeDescription
-                           {
-                               TxOut = utxo,
-                               OutPoint = new OutPoint(coinSet.TransactionId, outputReference.Transaction.Index),
-                               Address = outputReference.Address,
-                               HashBlock = hashBlock,
-                               UtxoSet = coinSet,
-                               Secret = walletSecret // Temporary.
-                           };
+                   {
```
Aprogiena: Still not the right code style as is used elsewhere

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -584,6 +585,7 @@ public void StopStake()
             }
 
             this.logger.LogTrace("Wallet total staking balance is {0}.", new Money(utxoStakeDescriptions.Sum(d => d.TxOut.Value)));
+            this.logger.LogTrace("(-)");
```
Aprogiena: ` this.logger.LogTrace("(-):*.{0}={1}", nameof(utxoStakeDescriptions.Count), utxoStakeDescriptions.Count);`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
Aprogiena: This PR was supposed to be a change of one line. You changed 250 lines and now you don't like adding noise? This whole PR is noise. So at least if you add something, add it properly. Or just delete the whole PR, I don't care. We have https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/commenting-policy.md or if you like the new version - https://github.com/stratisproject/StratisBitcoinFullNode/pull/1678 - pick one you like. If you don't like it, open discussion elsewhere, don't discuss this on PR.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: You pass the token to FetchCoinsAsync, so if there is cancellation at any point before line 555, FetchCoinsAsync will throw exception. Therefore the chance of actually executing line 556 when the cancellation token is cancelled is terribly low. The line is not justified because it will be executed only in one of 10,000 attempts. And of course you're missing exit logs. But just delete it

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: hopefully not a blocker!

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -553,34 +553,39 @@ public void StopStake()
...
-            if (cancellationToken.IsCancellationRequested) return utxoStakeDescriptions;
+            if (cancellationToken.IsCancellationRequested)
+                return utxoStakeDescriptions;
 
             foreach (UnspentOutputReference outputReference in spendableTransactions)
             {
                 UnspentOutputs coinSet = fetchedCoinSet.UnspentOutputs
                     .FirstOrDefault(f => f?.TransactionId == outputReference.Transaction.Id);
-                if (coinSet == null || outputReference.Transaction.Index >= coinSet.Outputs.Length) continue;
+                if (coinSet == null || outputReference.Transaction.Index >= coinSet.Outputs.Length)
```
Aprogiena: Brackets here and below as mentioned earlier

Aprogiena:
There are still a few places which I asked to change and they then exchanged yet, so not yet for me.

src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewStackTest.cs
```diff
@@ -173,7 +174,7 @@ public BackedCoinView2(NonBackedCoinView inner, int outputCount = 0)
             public int OutputCount { get; }
             public CoinView Inner => this.inner;
 
-            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+            public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
```
fassadlr: Please rename `ct` to `cancellationToken`

src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewStackTest.cs
```diff
@@ -144,7 +145,7 @@ public BackedCoinView1(BackedCoinView2 inner, int outputCount = 0)
...
```
fassadlr: Please rename `ct` to `cancellationToken`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -114,7 +115,7 @@ public Task InitializeAsync()
         }
 
         /// <inheritdoc />
-        public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds)
+        public override Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, CancellationToken ct = default(CancellationToken))
```
fassadlr: Please rename `ct` to `cancellationToken`

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
dangershony: Please add the braces

fassadlr:
@Aprogiena the code policy states:

`We add parentheses to group the parts of complex boolean expressions.`

This is not a complex boolean operation:

`if (coinSet == null || outputReference.Transaction.Index >= coinSet.Outputs.Length)`

If you can't justify this as a complex boolean operation then I will dismiss your review and merge this PR cause we can't hold up work in this sprint.

fassadlr:
To add to this, the review policy is also that if 2 reviewers approve the changes we are allowed to to push.

fassadlr:
@Aprogiena It has also been agreed with @monsieurleberre to not create PRs like this again in the future.

Aprogiena:
The rule specifically mentions this example:

```
if (x.Value > 5 && y) // parentheses needed to avoid confusion
```

Which is less complex than the line I was commenting on. Therefore the code style requires the brackets there.

Aprogiena:
> To add to this, the review policy is also that if 2 reviewers approve the changes are allowed to be pushed.

Where is that written? Can you link please?

Aprogiena:
> If you can't justify this as a complex boolean operation then I will dismiss your review and merge this PR cause we can't hold up work in this sprint.

That would be very unfortunate.

Aprogiena:
Also https://github.com/stratisproject/StratisBitcoinFullNode/pull/1701#discussion_r205111616

fassadlr:
@Aprogiena, what myself and @dangershony is concerned about in this review is whether or not you can confirm that the code changes wont break staking.

monsieurleberre:
OK, I have seen what I needed to see; that a code review can be held for some a pair of brackets that don't actually improve code readability or code quality. I think this is wrong and I have said it enough. It is a shame that common sense doesn't weigh in sometimes, and I am convinced this is actually an important issue.
So I'll give you some brackets because they are less important than my point ❤️

Aprogiena:
> OK, I have seen what I needed to see; that a code review can be held for some a pair of brackets that don't actually improve code readability or code quality. I think this is wrong and I have said it enough. It is a shame that common sense doesn't weigh in sometimes, and I am convinced this is actually an important issue.

Any reason you can give me not to write the code properly and according to the requested code style the first time? Just give me one common sense reason why would you not do that instead of just trying to make drama.

Aprogiena:
@monsieurleberre you should read https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/codereview-policy.md

And it is perfectly fine if you don't agree with certain policy or if you don't understand why is it in place. It is not okay if you just ignore it if you don't understand it. So please don't do that.

If you don't agree with something, you are supposed to open a discussion about the topic, but you're not supposed to create a code that is against the policies. So please with all the policies when you create a code up until the point the policy is changed.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +517,82 @@ public void StopStake()
...
 
+        public async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptionsAsync(WalletSecret walletSecret, CancellationToken cancellationToken)
+        {
+            this.logger.LogTrace("()");
+            var utxoStakeDescriptions = new List<UtxoStakeDescription>();
+            List<UnspentOutputReference> spendableTransactions = this.walletManager
+                .GetSpendableTransactionsInWallet(walletSecret.WalletName, 1).ToList();
+
+            FetchCoinsResponse fetchedCoinSet = await this.coinView.FetchCoinsAsync(spendableTransactions.Select(t => t.Transaction.Id).ToArray(), cancellationToken).ConfigureAwait(false);
+            if (cancellationToken.IsCancellationRequested)
```
Aprogiena: remove this

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +517,82 @@ public void StopStake()
...
+                    HashBlock = hashBlock,
+                    UtxoSet = coinSet,
+                    Secret = walletSecret // Temporary.
+                };
+                utxoStakeDescriptions.Add(utxoStakeDescription);
+
+                this.logger.LogTrace("UTXO '{0}' with value {1} might be available for staking.", utxoStakeDescription.OutPoint, utxo.Value);
+
+                if (cancellationToken.IsCancellationRequested)
+                    return utxoStakeDescriptions;
```
Aprogiena: don't return just break and move to the top of foreach

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -509,67 +517,82 @@ public void StopStake()
...
                 {
-                    this.logger.LogTrace("{0} failed, waiting {1} ms for next round...", nameof(this.StakeAndSignBlockAsync), this.minerSleep);
-                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), this.stakeCancellationTokenSource.Token).ConfigureAwait(false);
+                    this.logger.LogTrace("{0} failed to create POS block, waiting {1} ms for next round...", nameof(this.StakeAndSignBlockAsync), this.minerSleep);
+                    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), cancellationToken).ConfigureAwait(false);
                 }
             }
         }
 
+        public async Task<List<UtxoStakeDescription>> GetUtxoStakeDescriptionsAsync(WalletSecret walletSecret, CancellationToken cancellationToken)
```
Aprogiena: make it internal?

monsieurleberre: then I have to friend the assembly testing it, not sure it is a good solution

Aprogiena:
@monsieurleberre Also not cool that you just ignored some of my other comments on this PR, but let's say I will make a compromise and be okay unblocking if you just fix those last three things that I have just pointed out again.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: OK I don't understand; if you know that the staking loop has been cancelled, why bother going on any further with it ? (that is assuming that by "break" you meant "continue")

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: Definitely should not be public. Elsewhere we either use internal or we use reflection.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: I put it there so that you can exit the function as soon as you know you won't care about its result, can you detail why you want it removed please ?

monsieurleberre:
It's a fair point too @Aprogiena, I should not be making drama, actually neither of us should.
I find the reviewing process extremely useful, but in my opinion, on many occasions it focuses on issue that are more in the realm of opinions than anything else. On these, I feel that we should let consensus emerge by itself, or simply tolerate other people's preferences.

About coding policy, yes I have read it twice or more, but sometimes I forget a rule or two genuinely. In the case of the brackets, I also genuinely thought the expression was way too simple to require any brackets.

I know it has been a painful review for you, I do apologise and I will try to keep the other ones more focused. Hopefully we'll then keep it about the code's quality. Thanks for the help!

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
monsieurleberre: OK

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: i mean break, not continue

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: As mentioned previously, there is almost 0 chance of executing this line and if you move the other check to the top of the loop you will basically have the same functionality without this code here

Aprogiena:
All the reviews are painful, that's why I love it. Thanks for understanding.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: actually, even break is too weak probably as it would just not prevent the caller to continue, therefore i'd suggest
1) move it to the top of the loop
2) make it throw with exit log:
```
if (cancellationToken.IsCancellationRequested)
{
    throw new OperationCancelledException();
    this.logger.LogTrace("(-)[CANCELLATION]");
}
```

note that i'm intentionally (cause of log) not using throw if cancelled on the token

monsieurleberre:
Thanks a lot guys ❤️

Issue #631, opened 2017-10-21
Bitcoin testnet, mempool rejecting witness txs


mikedennis:
I'm not sure if this is an issue with bitcoin testnet or with nbitcoin but when I'm running stratis.bitcoind on bitcoin testnet I'm getting lots of tx rejected to mempool because of this condition.

https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/ScriptEvaluationContext.cs#L586 The txs are have witness.PushCount == 0 so this condition is failing.

Is this a problem with testnet or this nbitcoin check?

dangershony:
@NicolasDorier what do you think?

mikedennis:
One other interesting thing is this sets the context to 'Fail' not 'Invalid' because it is launched from here https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs#L1107 - So we don't even get a log of it occuring because Fails are not logged because it ends up here https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs#L445 

There are a few Todo's along this path what do you think is the proper way of handling this @dangershony ?

NicolasDorier:
The problem likely come from the fact that stratis bitcoind is not asking segwit transactions to the network.

Stratis.bitcoind should ask for blocks only to nodes supporting segwit, and then use

```
invPayload.Inventory.Add(new InventoryVector(node.AddSupportedOptions(InventoryType.MSG_TX), hash));
```

Same things when asking for Blocks. I remember I did that initially,

NicolasDorier:
Issue is likely

```
https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs#L468
```

The witness flag is not added.

mikedennis:
Thanks Nicolas that makes sense. I will look into that. What are these  Program.Length == 20 txs, are they some sort of pre-segwit txs?

NicolasDorier:
It is a segwit scriptPubkey program version 0, p2wpkh as described by https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki

Because the stratis full node enforce segwit, it needs the segwit data for validating a segwit program.
Since the mempool was not asking for it, the transaction were rejected.

[...]

PR #1703, opened 2018-07-23
When sending getdata message for tx inventory add appropriate witness flags

mikedennis:
Related to issue #631 

I finally think I got to the bottom of this issue. We weren't adding the witness flags when requesting txs using "getdata" after receiving inventory. This is why the peers were sending us invalid txs. This fits with the reasoning that Nicolas mentioned in the original issue. I had missed this location when I had looked at this code previously.

I've been running the bitcoin testnet for a while and haven't seen any of these invalid txs come through yet so I'm hopefully this is the fix.

I will also try syncing from scratch.

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -325,7 +325,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
                     continue;
                 }
 
-                send.Inventory.Add(inv);
+                send.Inventory.Add(new InventoryVector(peer.AddSupportedOptions(InventoryType.MSG_TX), inv.Hash));
             }
 
             // add to known inventory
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -325,7 +325,7 @@ public override object Clone()
                     continue;
                 }
 
-                send.Inventory.Add(inv);
+                send.Inventory.Add(new InventoryVector(peer.AddSupportedOptions(InventoryType.MSG_TX), inv.Hash));
```
noescape00: Why not 

```
inv.Type = peer.AddSupportedOptions(InventoryType.MSG_TX);
send.Inventory.Add(inv);
```

to avoid creating a new instance of `InventoryVector`

mikedennis: I'm not sure I like that. I prefer to keep collections immutable especially with async stuff going on, but maybe I'm just paranoid. Should be noted InventoryVector is only a hash and flag.

noescape00: This collection is not used after that method so it's ok to modify items inside of it or even add/remove items from the collection. 

But ok, I don't mind.

PR #1704, opened 2018-07-23
 ConsensusOptions per Network

### LINKED ISSUE #1217: Default max block size & weight values that are used in PoSMining are not configurable and different comparing to consensus values
PowMining:34

```
/// <summary>Default for "-blockmaxsize", which controls the maximum size of block the mining code will create.</summary>
public const int DefaultBlockMaxSize = 750000;

/// <summary>
/// Default for "-blockmaxweight", which controls the maximum block weight the mining code can create.
/// Block is measured in weight units. Data which touches the UTXO (What addresses are involved in the transaction, how many coins are being transferred) costs
/// 4 weight units (WU) per byte. Witness data (signatures used to unlock existing coins so that they can be spent) costs 1 WU per byte.
/// <seealso cref="http://learnmeabitcoin.com/faq/segregated-witness"/>
/// </summary>
public const int DefaultBlockMaxWeight = 3000000;
```

Those values are used in the block assembler but the max allowed values by the consensus are 1000000 for the block size and 4000000 for the block weight. 
Also they are not configurable which goes against the comments for those values. 

So effectively without changing them it is not possible to produce blocks larger than 732 kb.

I suggest implementing the configuration of those values and changing default values to be the maximum allowed values by the consensus (those changes are tested and [block ](https://chainz.cryptoid.info/strat-test/block.dws?334566.htm) generated with them is perfectly valid).

codingupastorm:
This allows us to set current block size settings inside networks. e.g. inside StratisMain.

I planned to fix #1217 but in doing so we really need to make things a little more modular. When we're setting the max block size settings in the constructor for `BlockDefinition`, we actually need to know the maximum block size for the network we're on.

Moreover, when looking at the properties inside `PowConsensusOptions`, they're not specific to Pow at all:
```
            this.MaxBlockSerializedSize = 4000000;
            this.MaxBlockWeight = 4000000;
            this.WitnessScaleFactor = 4;
            this.SerializeTransactionNoWitness = 0x40000000;
            this.MaxStandardVersion = 2;
            this.MaxStandardTxWeight = 400000;
            this.MaxBlockBaseSize = 1000000;
            this.MaxBlockSigopsCost = 80000;
```

Correct me if I'm wrong but these are all settings related to blocks, and are necessary regardless whether we're PoW, PoS, PoA, etc etc.

They could really just be fields on the Consensus object itself but for now this is slightly simpler.

This PR makes the following changes:
* Move ConsensusOptions to be a standalone class in NBitcoin.
* Move all of the properties from PowConsensusOptions to ConsensusOptions
* Remove the hacky `Option<T>` method, and when retrieving `PoSConsensusOptions` use an `as` operator.

### DIFF
--- a/src/NBitcoin/Consensus.cs
+++ b/src/NBitcoin/Consensus.cs
@@ -24,13 +24,6 @@ namespace NBitcoin
     public class Consensus
     {
         /// <summary>
-        /// An extension to <see cref="Consensus"/> to enable additional options to the consensus data.
-        /// </summary>
-        public class ConsensusOptions
-        {
-        }
-
-        /// <summary>
         /// How many blocks should be on top of a coinbase transaction until its outputs are considered spendable.
         /// </summary>
         public long CoinbaseMaturity { get; set; }
--- a/src/NBitcoin/ConsensusOptions.cs
+++ b/src/NBitcoin/ConsensusOptions.cs
@@ -1,0 +1,51 @@
+﻿namespace NBitcoin
+{
+    /// <summary>
+    /// An extension to <see cref="Consensus"/> to enable additional options to the consensus data.
+    /// </summary>
+    public class ConsensusOptions
+    {
+        /// <summary>The maximum allowed size for a serialized block, in bytes (only for buffer size limits).</summary>
+        public int MaxBlockSerializedSize { get; set; }
+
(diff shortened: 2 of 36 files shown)


codingupastorm:
@fassadlr saw that you are adding Rules to Networks, thought it might be useful for you to see this too, it's in the same vein.

Issue #1397, opened 2018-05-08
PoSMinting race condition


noescape00:
There is a race condition in case you call `StopStake` and `Stake` at the same time: 

First Stake will create a new `stakeCancellationTokenSource ` and StopStake will dispose it so by the end of the Stake call staking won't start. Also this race condition covers the whole staking loop itself.

dangershony:
Is that an issue? Staking can then just be called again and will work properly.
Is stop staking was called it was intentional

noescape00:
I guess it is an issue in case in UI we will have a button Start staking/Stop staking. Calling stop and after that start might produce such a race condition so node stats will say that staking is on when it's off.

dangershony:
I am not sure you are right, the cancellation will be cancelled the loop will stop, and this code will still be called `this.rpcGetStakingInfoModel.Enabled = false;`   
So the UI will see staking is not on and user will try again.

Yeah it might crash the node if the cancellation is not initialized already.

But I agree we should put a regular lock around the code in the methods just to be safe.

Aprogiena:
This is a race condition, I appeal to argument by authority and conclude that I am right. LOL

Anyway, seriously, it can crash the node, I can show you on Skype.

dangershony:
Fair point a race condition as small as it is will bite back eventually I learned that the hard way.

PR #1712, opened 2018-07-23
1397 fix race condition

fenix2222:
Updated PosMinting logic to prevent race condition when calling "Stake" and "Stop" immediately.

Preferred reviewer @Aprogiena as we discussed this one

Related issue #1397


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -243,14 +252,11 @@ public class CoinstakeContext
         /// <summary>Loop in which the node attempts to generate new POS blocks by staking coins from its wallet.</summary>
         private IAsyncLoop stakingLoop;
 
-        /// <summary>Indicates that the stake flag is not in progress.</summary>
-        public const int StakeNotInProgress = -1;
+        /// <summary>a flag that indicates if stake is on/off based on the <see cref="CurrentState"/> enum.</summary>
```
Aprogiena: Capital letter to start the sentence with

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -243,14 +252,11 @@ public class CoinstakeContext
...
+        private int stakeStateFlag;
 
-        /// <summary>Indicates that the stake flag is in progress.</summary>
-        public const int StakeInProgress = 1;
-
-        /// <summary>a flag that indicates if stake is on/off based on the <see cref="StakeInProgress"/> and <see cref="StakeNotInProgress"/> constants.</summary>
-        private int stakeProgressFlag;
+        /// <summary>a flag that indicates if stake stopping is on/off based on the <see cref="CurrentState"/> enum.</summary>
```
Aprogiena: Capital letter to start the sentence with

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -381,54 +388,71 @@ public void Stake(WalletSecret walletSecret)
         {
             this.logger.LogTrace("()");
 
-            if (Interlocked.CompareExchange(ref this.stakeProgressFlag, StakeInProgress, StakeNotInProgress) == StakeInProgress)
+            if (Interlocked.CompareExchange(
+                    ref this.stakeStateFlag,
```
Aprogiena: Do you really prefer this way of calling methods with each argument separated line or is it just the IDE setting that is doing that for you?

In any case, if you really want to have it on multiple lines, which I think is terrible ugly and less readable, you should respect the indentation which is four spaces.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -381,54 +388,71 @@ public void Stake(WalletSecret walletSecret)
...
-                {
-                    // All consensus exceptions should be ignored. It means that the miner
-                    // run into problems while constructing block or verifying it
-                    // but it should not halted the staking operation.
-                    this.logger.LogDebug("Consensus error exception occurred in miner loop: {0}", cee.ToString());
-                    this.rpcGetStakingInfoModel.Errors = cee.Message;
-                }
-                catch
+            this.rpcGetStakingInfoModel.Enabled = true;
+            this.stakeCancellationTokenSource =
```
Aprogiena: Here I would argue that it definitely does not belong onto separated lines. Probably your IDE setting or resharper is set with quite a low number of characters on one line. Please consider week your settings.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -381,54 +388,71 @@ public void Stake(WalletSecret walletSecret)
...
-                }
+                    this.logger.LogTrace("()");
+
+                    try
+                    {
+                        await this.GenerateBlocksAsync(walletSecret).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                            // Application stopping, nothing to do as the loop will be stopped.
```
Aprogiena: Indentation fuck up here and below

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -176,6 +176,15 @@ public class CoinstakeContext
             public Key Key { get; set; }
         }
 
+        /// <summary>
+        /// Indicates the current state of the staking: in progress, stopped or idle.
```
Aprogiena: This is weird, the comment mentions three states, but the enumeration only has two states

Aprogiena: I personally think you should have three or four states here and only have one field, not two. The way you have it now will not work, just consider what would happen if you executed start – stop – start – stop – start

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -176,6 +176,15 @@ public class CoinstakeContext
...
+        /// </summary>
+        public enum CurrentState
```
noescape00: Why not just int with a comment that 0 is idle and 1 is executing? 

For example in network peer class we don't use enum and just int. 

Or we can add a class for atomic operations with bool. Example can be taken from here: https://stackoverflow.com/questions/6661055/using-interlocked-compareexchange-operation-on-a-bool-value/18027246

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -381,54 +388,71 @@ public void Stake(WalletSecret walletSecret)
...
+                            // but it should not halted the staking operation.
+                            this.logger.LogDebug("Miner exception occurred in miner loop: {0}", me.ToString());
+                        this.rpcGetStakingInfoModel.Errors = me.Message;
+                    }
+                    catch (ConsensusErrorException cee)
+                    {
+                            // All consensus exceptions should be ignored. It means that the miner
+                            // run into problems while constructing block or verifying it
+                            // but it should not halted the staking operation.
+                            this.logger.LogDebug(
```
noescape00: Please configure VS to use 120 chars in a line, not 80. Everyone have big monitors. 
Imho no need to split it that much. 

However it's purely a personal preference question so if you like it as it is- don't change.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: Yes I do. It makes it much easier to see what is passed into the method. There are too many brackets, commas and operators in one line to clearly see boundaries of function and comparison. "Ugly" is highly subjective and comes down to personal preference. I would prefer to leave it as it. I have fixed tabs.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: Fixed

noescape00:
I mean not using enum there. Like it's in network peer. 

Instead just using int or adding an atomic bool class. 

I don't mind the current approach, just not sure if always casting to int looks good.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: Fixed comment.

Not sure how we can get away with just one flag. Every time we run Interlocked.CompareExchange it sets the flag to one of the states, i.e. I cannot just check the state without setting the flag. Unless I am missing something.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: I am on the fence here. Enums are much better for readability of the code and also allow you to extend states in the future, while booleans are prone to double negative usage which makes it hard to read and buggy, i.e. `stakeNotInProgress` and `!stakeNotInProgress` ... But I do see your point with casting to `int` which is a bit verbose, it would be nice if Interlocked.CompareExchenge supported enums.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: Agree, fixed. It was autoformatting doing it.

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I actually like enum here

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I still see eight spaces.

I agree that it is subjective, that is why I don't require you to write it in the other way is there is no policy for that. Though I really want you to know that it is ugly ;-)

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: I think you are missing something. But let's start with a simple question – do you see why it cannot work as it is?

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
fenix2222: That’s correct. 2 indents. Logically Parameters should not be on the same line as method. So Interlocked should not be online with ref

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: it is still wrong everywhere

src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
Aprogiena: Though you should not be propagating your personal style at expense of creating code that is inconsistent with the rest of the code.

Aprogiena: OK, https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/coding-style.md first paragraph

fenix2222:
@Aprogiena that's fine, updated styling. It is pointless to argue about this here and pollute the PR.

majikandy:
I will run the tests

noescape00:
Please fix merge conflicts

fenix2222:
@noescape00 Done

majikandy:
All 4 networks sync fine

PR #1716, opened 2018-07-24
Move Network registrations to Networks class

rowandh:
This is a large diff because networks are referenced everywhere, but there's only two things that have changed:
* Move networks to a different namespace (NetworkDefinitions)
* Move static network registrations to a new class (Networks)

Importantly, this brings static network registrations out of `Network.cs`.


fassadlr:
@rowandh we are happy to merge but fix your conflicts 👍  Good change!

dangershony:
@rowandh this is priority as its easy and not to block you, so once you resolve conflicts ask us to push immediately, PR looks good.

fassadlr:
@rowandh i fixed this for you 👍 I confirm that all unit and integration tests pass.

rowandh:
@fassadlr thanks for fixing!

PR #1717, opened 2018-07-24
Remove Consensus from ConsensusFactory

rowandh:
`Consensus.ConsensusFactory` was only being used to check if the network is POS. 

This PR:
* Replaces the check for `IsProofOfStake` by looking at the type of `ConsensusFactory`
* Removes the `Consensus` property from `ConsensusFactory`

### DIFF
--- a/src/NBitcoin/BitcoinCore/Coins.cs
+++ b/src/NBitcoin/BitcoinCore/Coins.cs
@@ -160,7 +160,7 @@ namespace NBitcoin.BitcoinCore
                 // coinbase height
                 stream.ReadWriteAsVarInt(ref this.nHeight);
 
-                if (stream.ConsensusFactory.Consensus.IsProofOfStake)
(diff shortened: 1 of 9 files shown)


src/NBitcoin/BitcoinCore/Coins.cs
```diff
@@ -160,7 +160,7 @@ public void ReadWrite(BitcoinStream stream)
                 // coinbase height
                 stream.ReadWriteAsVarInt(ref this.nHeight);
 
-                if (stream.ConsensusFactory.Consensus.IsProofOfStake)
+                if (stream.ConsensusFactory is PosConsensusFactory)
```
fassadlr: The assumption here would be that if the network's consensus factory is POS then it is a POS network? @dangershony this change won't break anything (currently) but do you agree with this? What if we on a POS network create a Pow consensus factory? I'm just trying to think how someone can mess this up accidentally.

PR #1721, opened 2018-07-24
Fixes for comments and warnings

fenix2222:
Fixed warnings and comments

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
@@ -32,7 +32,7 @@ namespace Stratis.Bitcoin.Features.Miner
     /// <remarks>
     /// Staking is attempted only if the node is fully synchronized and connected to the network.
     /// If not it will wait till node is synced. Only transactions that were confirmed at least
-    /// <see cref="PosConsensusOptions.StakeMinConfirmations"/> blocks ago are eligible for staking.
+    /// <see cref="PosConsensusOptions.GetStakeMinConfirmations(int, Network)"/> blocks ago are eligible for staking.
     /// <para>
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Miner/PosMinting.cs
```diff
@@ -304,18 +299,18 @@ public class CoinstakeContext
...
-        private IInitialBlockDownloadState initialBlockDownloadState;
+        private readonly IInitialBlockDownloadState initialBlockDownloadState;
 
         /// <summary>State of time synchronization feature that stores collected data samples.</summary>
         private readonly ITimeSyncBehaviorState timeSyncBehaviorState;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PosMinting"/> class.
         /// </summary>
+        /// <param name="blockProvider">Block provider.</param>
```
monsieurleberre: @dangershony in the end, what have we decided ? are we going to update the commenting policy for cases like that where comments are just noise, or is it still a hotly debated topic ?

fenix2222: @dangershony well we either remove all parameter comments or keep
them up-to-date. Doesn’t make sense to have half-completed comments.

monsieurleberre: @fenix2222 please don't take my question as personal attack, it is just a side question :) I just find it sad to have to write this kind of things because of a rule. I don't mind if people do it, but I personally can't understand why I would be forced to do that.
To disable warnings: https://stackoverflow.com/questions/7982525/visual-studio-disabling-missing-xml-comment-warning

fenix2222:
@dangershony or @monsieurleberre Could you guys merge it? I don't have permissions to do it

PR #1724, opened 2018-07-24
CM- signal 1 by one during the connection and extend signaler to have block disconnected signal

noescape00:


src/Stratis.Bitcoin/Signals/Signals.cs
```diff
@@ -42,36 +55,49 @@ public class Signals : ISignals
...
+            this.blocksDisconnected = blockDisonnectedSignaler;
             this.transactions = transactionSignaler;
         }
 
         /// <summary>Signaler providing notifications about newly available blocks to its subscribers.</summary>
-        private ISignaler<Block> blocks { get; }
+        private ISignaler<Block> blocksConnected { get; }
+
+        /// <summary>Signaler providing notifications about blocks being disconnected to its subscribers.</summary>
+        private ISignaler<Block> blocksDisconnected { get; }
```
majikandy: @dangershony 's suggestion was that this should be ChainedHeader and not Blocks  - I have done the subscription based on Block for now but can change to ChainedHeader if you decide to take that approach.

Aprogiena: The intention is to replace with signalling of ChainHeaderBlock

noescape00: Yeah. Discussed that on skype- we will use ChainHeaderBlock. But after a few PRs that will change rewinding. Possibly after activation

src/Stratis.Bitcoin/Signals/Signals.cs
```diff
@@ -42,36 +55,49 @@ public class Signals : ISignals
...
         /// </summary>
-        /// <param name="blockSignaler">Signaler providing notifications about newly available blocks to its subscribers.</param>
+        /// <param name="blockConnectedSignaler">Signaler providing notifications about newly available blocks to its subscribers.</param>
+        /// <param name="blockDisonnectedSignaler">Signaler providing notifications about a block being disconnected to its subscribers.</param>
         /// <param name="transactionSignaler">Signaler providing notifications about newly available transactions to its subscribers.</param>
-        public Signals(ISignaler<Block> blockSignaler, ISignaler<Transaction> transactionSignaler)
+        public Signals(ISignaler<Block> blockConnectedSignaler, ISignaler<Block> blockDisonnectedSignaler, ISignaler<Transaction> transactionSignaler)
         {
-            Guard.NotNull(blockSignaler, nameof(blockSignaler));
+            Guard.NotNull(blockConnectedSignaler, nameof(blockConnectedSignaler));
```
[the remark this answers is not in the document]
Aprogiena: I don't think we would have use case for that. Can you think of any? If you would allow this to be not provided, you would like to specify that the parameter comment that it can be null, but I can't see any reason for that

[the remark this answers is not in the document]
noescape00: Ok, makes sense. Added a guard for disconnected signal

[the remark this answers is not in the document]
dangershony: Yes you will always initialise a signaller, it might not have any subscribers but you dont know that at this point

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -584,34 +584,41 @@ private void ProcessDownloadedBlock(ChainedHeaderBlock chainedHeaderBlock)
...
+                    int newFinalizedHeight = newTip.Height - (int)this.network.Consensus.MaxReorgLength;
+
+                    await this.finalizedBlockHeight.SaveFinalizedBlockHeightAsync(newFinalizedHeight).ConfigureAwait(false);
+                }
 
-            foreach (ChainedHeaderBlock validatedBlock in blocksToConnect)
-            {
                 // TODO: change signal to take ChainedHeaderBlock
-                this.signals.SignalBlock(validatedBlock.Block);
+                this.signals.SignalBlockConnected(blockToConnect.Block);
```
dangershony: This is a bit of a change from before, it means we will be signalling blocks while CM is in a special state, did we consider all the consequences?

noescape00: Yes, as long as callback methods don't call the CM itself in a way that it will try to take reorg lock it should be totally fine.

PR #1728, opened 2018-07-24
Split posminting

monsieurleberre:
This PR is for Francois eyes only, I would like to re-do it when my other PR called stake-vulnerability has gone in, unless someone else is OK to put it in


monsieurleberre:
image

Issue #1686, opened 2018-07-19
On ReOrg, transactions removed need to return to mempool


majikandy:
**In order to** maintain a complete chain
**As a** node
**I want** transactions missed on reorg to return to the mempool

**Acceptance Criteria**
- Transactions missing on reorg are in the mempool after the reorg completes     

NOTE: current behaviour in the node is ...
When a reorg happens, blocks in the discarded shorter chain are not returned to the mempool.

PR #1734, opened 2018-07-25
Missing transactions returned to mempool on reorg

majikandy:
This builds on https://github.com/stratisproject/StratisBitcoinFullNode/pull/1724 which should be merged before this.

This is the subscriber on the Disconnect Signal and fixes issue #1686 

After PR 1724 is merged, it still needs wiring up to push the disconnect event when rewinding the blocks...

therefore the Integration test:-

A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect()

will fail on the final line until that disconnect signal is fired for the blocks reorged away.


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -68,7 +68,7 @@ public class BlockStoreSignaled : SignalObserver<Block>
             this.initialBlockDownloadState = initialBlockDownloadState;
 
             this.blocksToAnnounce = new AsyncQueue<ChainedHeader>();
-            this.dequeueLoopTask = this.DequeueContinuouslyAsync();
+            this.dequeueLoopTask = this.DequeueContinuouslyAsync();
```
Aprogiena: Don't need those spaces, right?

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
+
+            var state = new MempoolValidationState(true);
+
+            await this.mempoolLock.WriteAsync(async () =>
+            {
+                foreach (Transaction transaction in block.Transactions)
+                {
+                    bool success = await this.mempoolValidator.AcceptToMemoryPool(state, transaction);
```
Aprogiena: We use configure await

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,60 @@
...
+                    this.logger.LogTrace("[ACCEPT_TO_MEMPOOL]:", success);
```
Aprogiena: This log does not make much sense, but you wanted to talk about that, right?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolFeature.cs
```diff
@@ -135,8 +143,7 @@ public override void Dispose()
                 }
             }
 
-            if (this.mempoolSignaled != null)
-                this.mempoolSignaled.Stop();
+            this.mempoolSignaled?.Stop();
```
Aprogiena: Is it even possible that we would have null here?

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public async Task AddBackToMempoolAsync(Block block)
+        {
+            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
+
+            var state = new MempoolValidationState(true);
+
+            await this.mempoolLock.WriteAsync(async () =>
+            {
+                foreach (Transaction transaction in block.Transactions)
```
Aprogiena: Are you sure, you can process the transactions in this order? I think it should be processed in reverse order

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    bool success = await this.mempoolValidator.AcceptToMemoryPool(state, transaction);
+                    this.logger.LogTrace("[ACCEPT_TO_MEMPOOL]:", success);
+                }
+            });
```
Aprogiena: and confawait

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
dangershony: No I think this is correct, trx at the bottom may be spending trx at the top and trx at the top should be spending trx in the chain.

Would it make sense to look at the possibility to push back to mempool without full trx validation

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,66 @@
...
+                this.logger.LogTrace("()");
+
+                foreach (Transaction transaction in block.Transactions)
+                {
+                    bool success = await this.mempoolValidator.AcceptToMemoryPool(state, transaction);
+
+                    if (!success)
+                        this.logger.LogDebug("Transaction with hash:{0} failed to go back into mempool on block disconnect.", transaction.GetHash());
+                    else
+                        this.logger.LogTrace("Transaction with hash:{0} accepted back to mempool.");
```
Aprogiena: this.logger.LogTrace("Transaction with hash '{0}' accepted back to mempool.", transaction.GetHash());

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -0,0 +1,66 @@
...
+            await this.mempoolLock.WriteAsync(async () =>
+            {
+                this.logger.LogTrace("()");
+
+                foreach (Transaction transaction in block.Transactions)
+                {
+                    bool success = await this.mempoolValidator.AcceptToMemoryPool(state, transaction);
+
+                    if (!success)
+                        this.logger.LogDebug("Transaction with hash:{0} failed to go back into mempool on block disconnect.", transaction.GetHash());
```
Aprogiena: this.logger.LogDebug("Transaction with hash '{0}' failed to go back into mempool on block disconnect.", transaction.GetHash());

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
Aprogiena: It turned out the question is trickier than it seems, I need to think about it little more

Aprogiena: As for the transaction validation, I'm thinking there could be problems

Aprogiena: I think you're right, it is because we do validation one by one and not as a batch, we need to go in this order and not in reverse order.

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
Aprogiena: Just for data type consistency

src/Stratis.Bitcoin.Features.MemoryPool/MempoolFeature.cs
Aprogiena: still open question :)

src/Stratis.Bitcoin.Features.MemoryPool/MempoolFeature.cs
```diff
@@ -128,22 +128,21 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
         /// <inheritdoc />
         public override void Dispose()
         {
-            if (this.mempoolManager != null)
+            if (this.mempoolManager == null) return;
```
Aprogiena: And that even happen?

fenix2222: I guess it won't even initialise if mempool manager is null. We should probably guard it in the constructor.

bokobza:
Is this finished? I can't see where the signal is sent.

noescape00:
> I can't see where the signal is sent.

It's not. Signal will be triggered by CM. CM will have this logic after the activation.

dangershony:
I will sync networks

dangershony:
Networks syncing

PR #1742, opened 2018-07-26
Begin making Consensus immutable

rowandh:
* Makes most properties of Consensus immutable. Some are used in tests and making them immutable is less of an easy task
* Uses a MutableConsensus object to set properties before 'freezing' in Consensus. I felt this was a better approach than having a constructor with 30 params. Future work on breaking Consensus into logical chunks may mean that constructor injection is possible so this might not last long.

Future work:
* Make remaining properties immutable
* Break Consensus down into logical components


src/NBitcoin/BIP9DeploymentsArray.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace NBitcoin
```
dangershony: If you make changes here can we look at moving all deployment code to Stratis.Bitcoin.Base.Deplyoments folder?

src/NBitcoin/NetworkDefinitions/BitcoinMain.cs
```diff
@@ -27,33 +27,51 @@ public BitcoinMain()
...
+            this.GenesisNonce = 2083236893;
+            this.GenesisBits = 0x1d00ffff;
+            this.GenesisVersion = 1;
+            this.GenesisReward = Money.Coins(50m);
+
+            Block genesisBlock = CreateBitcoinGenesisBlock(consensusFactory, this.GenesisTime, this.GenesisNonce, this.GenesisBits, this.GenesisVersion, this.GenesisReward);
+
+            this.Genesis = genesisBlock;
+
+            var consensus = new MutableConsensus();
```
dangershony: Interesting approach

src/NBitcoin/MutableConsensus.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using NBitcoin.BouncyCastle.Math;
+
+namespace NBitcoin
+{
+    /// <summary>
+    /// Used to create a Mutable version of the <see cref="Consensus"/> object during configuration
+    /// </summary>
+    public class MutableConsensus : IConsensus
```
dangershony: I am not sure the naming is correct but I like it none the less.

dangershony: I wonder if instead of using an interface we just use an Abstract consensus class with protected setters this will make it immutable and not need to have duplicated properties (on interface and two implementations)

dangershony:
Yes with the time difference we can do it today in London so Rowan just need to merge tomorrow

src/NBitcoin/MutableConsensus.cs
monsieurleberre: I think it is better to keep an interface that we can easily mock when doing tests

src/NBitcoin/Consensus.cs
```diff
@@ -3,133 +3,107 @@
...
-            this.ConsensusFactory = new ConsensusFactory();
+            this.CoinbaseMaturity = consensus.CoinbaseMaturity;
+            this.PremineReward = consensus.PremineReward;
+            this.PremineHeight = consensus.PremineHeight;
+            this.ProofOfWorkReward = consensus.ProofOfWorkReward;
+            this.ProofOfStakeReward = consensus.ProofOfStakeReward;
+            this.MaxReorgLength = consensus.MaxReorgLength;
+            this.MaxMoney = consensus.MaxMoney;
+            this.Options = consensus.Options;
+            this.BuriedDeployments[NBitcoin.BuriedDeployments.BIP34] =
```
noescape00: Please configure VS to disable splitting the line when >80 chars. Lines up to 120-160 lines are ok.

src/NBitcoin/IConsensus.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System;
+using NBitcoin.BouncyCastle.Math;
+
+namespace NBitcoin
+{
+    public interface IConsensus
+    {
+        /// <summary>
+        /// How many blocks should be on top of a coinbase transaction until its outputs are considered spendable.
```
noescape00: on top of a block that includes a coinbase transaction

src/NBitcoin/IConsensus.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// </summary>
+        long CoinbaseMaturity { get; set; }
+
+        /// <summary>
+        /// Amount of coins mined when a new network is bootstrapped.
+        /// Set to Money.Zero when there is no premine.
```
noescape00: `<see cref="Money.Zero"/>`

src/NBitcoin/MutableConsensus.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        public MutableConsensus()
+        {
+            this.BuriedDeployments = new BuriedDeploymentsArray();
+            this.BIP9Deployments = new BIP9DeploymentsArray();
+        }
+
+        public long CoinbaseMaturity { get; set; }
```
noescape00: empty lines between fields

src/NBitcoin/IConsensus.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// </summary>
+        long MaxMoney { get; }
+
+        ConsensusOptions Options { get; set; }
+
+        BuriedDeploymentsArray BuriedDeployments { get; }
+
+        BIP9DeploymentsArray BIP9Deployments { get; }
+
+        int SubsidyHalvingInterval { get; }
```
noescape00: Please comment all of those methods.

src/NBitcoin/MutableConsensus.cs
noescape00: Also use inheritdoc for them

src/NBitcoin/IConsensus.cs
```diff
@@ -0,0 +1,104 @@
...
+        uint256 HashGenesisBlock { get; }
+
+        /// <summary> The minimum amount of work the best chain should have. </summary>
+        uint256 MinimumChainWork { get; }
+
+        int MinerConfirmationWindow { get; set; }
+        int RuleChangeActivationThreshold { get; set; }
+
+        /// <summary>
+        /// Specify the BIP44 coin type for this network
```
noescape00: dot missing

src/NBitcoin/IConsensus.cs
rowandh: @noescape00 all of these docs are copied straight from the existing `Consensus` object. If they are missing it's because they were missing there. I'm not sure I'm the right person to be describing what these fields are used for.

src/NBitcoin/IConsensus.cs
fassadlr: @rowandh can you fix the comment here?

src/NBitcoin/IConsensus.cs
noescape00: Ok. 

It's just those values are being touched in several PRs for the last few months but they never get commented.

src/NBitcoin/Consensus.cs
```diff
@@ -3,133 +3,101 @@
...
+        /// <inheritdoc />
+        public uint MaxReorgLength { get; }
+
+        /// <inheritdoc />
+        public long MaxMoney { get; }
+
+        public ConsensusOptions Options { get; set; }
 
-        public Consensus()
+        public Consensus(IConsensus consensus)
```
noescape00: Some of the fields are above the constructor, some are below. 
Can you please move all of them to be before the constructor?

src/NBitcoin/Consensus.cs
```diff
@@ -3,133 +3,101 @@
...
+        /// <inheritdoc />
+        public Money ProofOfStakeReward { get; }
+
+        /// <inheritdoc />
+        public uint MaxReorgLength { get; }
+
+        /// <inheritdoc />
+        public long MaxMoney { get; }
+
+        public ConsensusOptions Options { get; set; }
```
noescape00: Will the setter be removed in another PR or it's left here intentionally?

src/Stratis.Bitcoin.Networks/BitcoinMain.cs
```diff
@@ -28,33 +28,51 @@ public BitcoinMain()
...
+            this.GenesisNonce = 2083236893;
+            this.GenesisBits = 0x1d00ffff;
+            this.GenesisVersion = 1;
+            this.GenesisReward = Money.Coins(50m);
+
+            Block genesisBlock = CreateBitcoinGenesisBlock(consensusFactory, this.GenesisTime, this.GenesisNonce, this.GenesisBits, this.GenesisVersion, this.GenesisReward);
+
+            this.Genesis = genesisBlock;
+
+            var consensus = new MutableConsensus();
```
noescape00: Can be changed in a way to use object initializer: 

```
var consensus = new MutableConsensus
{
    ConsensusFactory = consensusFactory,
    HashGenesisBlock = genesisBlock.GetHash(),
    SubsidyHalvingInterval = 210000,
    MajorityEnforceBlockUpgrade = 750,
    MajorityRejectBlockOutdated = 950,
    MajorityWindow = 1000,
    BuriedDeployments =
    {
        [BuriedDeployments.BIP34] = 227931,
        [BuriedDeployments.BIP65] = 388381,
        [BuriedDeployments.BIP66] = 363725
    },
    BIP34Hash = new uint256("0x000000000000024b89b42a942fe0d9fea3bb44ab7bd1b19115dd6a759c0808b8"),
    PowLimit = new Target(new uint256("00000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffff")),
    MinimumChainWork = new uint256("0x0000000000000000000000000000000000000000002cb971dd56d1c583c20f90"),
    PowTargetTimespan = TimeSpan.FromSeconds(14 * 24 * 60 * 60),
    PowTargetSpacing = TimeSpan.FromSeconds(10 * 60),
    PowAllowMinDifficultyBlocks = false,
    PowNoRetargeting = false,
    RuleChangeActivationThreshold = 1916,
    MinerConfirmationWindow = 2016,
    BIP9Deployments =
    {
        [BIP9Deployments.TestDummy] = new BIP9DeploymentsParameters(28, 1199145601, 1230767999),
        [BIP9Deployments.CSV] = new BIP9DeploymentsParameters(0, 1462060800, 1493596800),
        [BIP9Deployments.Segwit] = new BIP9DeploymentsParameters(1, 1479168000, 1510704000)
    },
    CoinType = 0,
    DefaultAssumeValid = new uint256("0x000000000000000000174f783cc20c1415f90c4d17c9a5bcd06ba67207c9bc80"),
    CoinbaseMaturity = 100,
    PremineReward = Money.Zero,
    ProofOfWorkReward = Money.Coins(50),
    ProofOfStakeReward = Money.Zero,
    MaxReorgLength = 0,
    MaxMoney = 21000000 * Money.COIN
};
```

But I guess it's personal preference so don't change if you consider it to be better as it is.

noescape00: Same stands for other networks.

src/NBitcoin/Consensus.cs
rowandh: Some setters are used in tests and making them immutable requires additional refactoring

noescape00: Ok, makes perfect sense.

bokobza:
I'm gonna have a look at this as well @rowandh @noescape00

rowandh:
@bokobza it's still possible to do something like that by implementing IConsensus. I chose not to because there's still the chance that Consensus gets broken down into smaller pieces later.

bokobza:
> there's still the chance that Consensus gets broken down into smaller pieces later.

but we don't have any real need for that yet, right?
I just want to avoid the overhead of these new classes if possible.

I was also thinking we can make as much as possible immutable and leave the maturity one mutable OR make everything immutable and change the value of CoinMaturity by force using reflection or whatever.

@dangershony RFC

dangershony:
I think my main requirement is that all the bootstrapping happens in one place the network constructor.

rowandh:
@dangershony I considered that but decided against it because the constructor would have 30 params. I don't mind it though, it's probably my preferred approach if having that many constructor params is acceptable.

rowandh:
@dangershony I have updated with your suggestions. Since I created this, another PR was merged to add the Rules to Consensus. This was not done in a way which is easy to make immutable so I have left it as is.

src/Stratis.Bitcoin.Networks/BitcoinMain.cs
```diff
@@ -110,15 +141,6 @@ public BitcoinMain()
...
-            // Create the genesis block.
-            this.GenesisTime = 1231006505;
-            this.GenesisNonce = 2083236893;
-            this.GenesisBits = 0x1d00ffff;
-            this.GenesisVersion = 1;
-            this.GenesisReward = Money.Coins(50m);
-
-            this.Genesis = CreateBitcoinGenesisBlock(this.Consensus.ConsensusFactory, this.GenesisTime, this.GenesisNonce, this.GenesisBits, this.GenesisVersion, this.GenesisReward);
-            this.Consensus.HashGenesisBlock = this.Genesis.GetHash();
             Assert(this.Consensus.HashGenesisBlock == uint256.Parse("0x000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"));
```
bokobza: I think the asserts should probably move next to the GenesisBlock

dangershony:
Great running networks.

PR #1750, opened 2018-07-26
Create Stratis.Bitcoin.Networks

fassadlr:


rowandh:
Hey @fassadlr really like what you've done here. Definitely much nicer when we separate out the known networks like that - it becomes obvious that they are mostly used for tests only!

Feedback:
* Solution doesn't compile as-is. `NetworksContainer` is referenced in a lot of places, but it's missing
* I'm finding it really difficult to review the PR due to the sheer number of files touched by these changes. Is it possible to separate into multiple PRs, or even multiple logical commits? Eg. Move networks to own solution -> Create known networks -> Remove networks registrations -> etc.

fassadlr:
image

All integration tests pass 👍

fassadlr:
I confirm that staking still works as per normal.

dangershony:
@rowandh I already started peer review this yesterday together with fasss so I will push it, already covered most of the code.

src/NBitcoin.Tests/util_tests.cs
```diff
@@ -7,24 +7,39 @@
...
         private static byte[] ParseHex_expected = new byte[]{
     0x04, 0x67, 0x8a, 0xfd, 0xb0, 0xfe, 0x55, 0x48, 0x27, 0x19, 0x67, 0xf1, 0xa6, 0x71, 0x30, 0xb7,
     0x10, 0x5c, 0xd6, 0xa8, 0x28, 0xe0, 0x39, 0x09, 0xa6, 0x79, 0x62, 0xe0, 0xea, 0x1f, 0x61, 0xde,
     0xb6, 0x49, 0xf6, 0xbc, 0x3f, 0x4c, 0xef, 0x38, 0xc4, 0xf3, 0x55, 0x04, 0xe5, 0x1e, 0xc1, 0x12,
     0xde, 0x5c, 0x38, 0x4d, 0xf7, 0xba, 0x0b, 0x8d, 0x57, 0x8a, 0x4c, 0x70, 0x2b, 0x6b, 0xf1, 0x1d,
     0x5f};
 
+        public Util_Tests()
+        {
+            NetworkRegistration.Clear();
```
dangershony: Why the need to clear NetworkRegistration?

src/Stratis.Bitcoin.Tests.Common/KnownNetworks.cs
```diff
@@ -0,0 +1,21 @@
+﻿using NBitcoin;
+using NBitcoin.Networks;
+using Stratis.Bitcoin.Networks;
+
+namespace Stratis.Bitcoin.Tests.Common
+{
+    public static class KnownNetworks
```
dangershony: I am wondering now if KnownNetworks should not be in the Networks project

src/NBitcoin.Tests/util_tests.cs
fassadlr: @dangershony I thought it a useful method should we wish to do however, the reason is a bit complex...

Prior to this change NetworkRegistration.GetNetworks yield returned the networks in this order:

```
public static IEnumerable<Network> GetNetworks()
{
    yield return NBitcoin.Networks.Main;
    yield return NBitcoin.Networks.TestNet;
    yield return NBitcoin.Networks.RegTest;
    ...
```

Because we can now potentially have Stratis network's in the collection AS WELL as the fact that the main networks (StratisMain and BitcoinMain) share the same `Bech32Encoder` set:

```
    var encoder = new Bech32Encoder("bc");
    this.Bech32Encoders = new Bech32Encoder[2];
    this.Bech32Encoders[(int)Bech32Type.WITNESS_PUBKEY_ADDRESS] = encoder;
    this.Bech32Encoders[(int)Bech32Type.WITNESS_SCRIPT_ADDRESS] = encoder;
```
 when we try and get the network type from the base address it sometime returns BitcoinMain and sometimes StratisMain. So I needed to clear the container before the test start.

src/Stratis.Bitcoin.Tests.Common/KnownNetworks.cs
fassadlr: Remember that `KnownNetworks` is just a wrapper for the `NetworkRegistration` class. `NetworkRegistration` manages the registered networks in its own internal collection called `registeredNetworks`. We can even look at deleting this over time.

dangershony: yeah also this will force us to develop in a network agnostic way so ok lets keep it out.

src/Stratis.BreezeD/Program.cs
```diff
@@ -29,7 +31,7 @@ public class Program
 
                 if (isStratis)
                 {
-                    Network network = isTestNet ? Networks.StratisTest : Networks.StratisMain;
+                    Network network = isTestNet ? NetworkRegistration.Register(new StratisTest()) : NetworkRegistration.Register(new StratisMain());
```
dangershony: I wonder if we should actually register a network when creating it, so inside the constructor of StratisTest() for example, syncing networks now.

dangershony:
All 4 networks synced beyond 100k blocks (stratis main to block 74k) with checkpoints disabled.

PR #1757, opened 2018-07-27
Split posminting class

monsieurleberre:
Split the big PosMinting class before more changes


fassadlr:
@monsieurleberre Please confirm all integration tests pass? 

I can confirm that staking works as per normal 👍

monsieurleberre:
Sorry, I confirm all tests are good and we checked staking still works yesterday together. I also checked the 4 deamons can sync no pb, thanks bru :)

PR #1759, opened 2018-07-28
GetAccount was throwing a RPCServerException which is not correct.

ferdeen:
I noticed when I created the `MiningApiController` based on the `MiningRPCController` it was still throwing RPC type exceptions, which is wrong.

This small PR is to throw a normal exception.

Related to work carried out in #1691.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/Controllers/MiningApiController.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/Controllers/MiningApiController.cs
@@ -6,8 +6,6 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Stratis.Bitcoin.Features.Miner.Interfaces;
 using Stratis.Bitcoin.Features.Miner.Models;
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Miner/Controllers/MiningApiController.cs
```diff
@@ -89,26 +90,35 @@ public IActionResult Generate([FromBody]MiningRequest request)
...
-   
+
         /// <summary>
         /// Finds first available wallet and its account.
         /// </summary>
         /// <returns>Reference to wallet account.</returns>
         private WalletAccountReference GetAccount()
         {
+            const string noWalletMessage = "No wallet found";
+            const string noAAccountMessage = "No account found on wallet";
```
fassadlr: `noAAccountMessage ` should be `noAccountMessage` ;)

ferdeen: Oops! Done.

Issue #1255, opened 2018-03-27
Wallet allows to spend not matured/recently staked coins


maciejzaleski:
Wallet will allow user to spend coins before they are mature.
Such transaction will not be accepted by accepted into the mempool.

dangershony:
Is it not enough that the mempool will reject it?
What I mean is the trx still in the wallet even that it got rejected?

noescape00:
@majikandy @dangershony wasn't it already fixed on master a week ago or so?

bokobza:
Closed with https://github.com/stratisproject/StratisBitcoinFullNode/commit/81b6d424e7d0c9658e46b337decc5dfe49a39f03

PR #1760, opened 2018-07-28
[Wallet] Prevent spending of immature coins

zeptin:
Fix for #1255 

Most tests are passing, I had to fix several dozen that were implicitly relying on the incorrect wallet logic.

There is still an issue with the SharedSteps.MineBlocks helper method not correctly taking fees into account.


src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -728,12 +728,29 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentC
 
                     if (confirmationCount >= confirmations)
                     {
-                        yield return new UnspentOutputReference
+                        if (includeImmature || (!(transactionData.IsCoinBase ?? false) && !(transactionData.IsCoinStake ?? false)))
```
Aprogiena: What does it mean if transaction data have null for Coinbase or coinstake flags?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -728,12 +728,29 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentC
...
+
+                            yield return new UnspentOutputReference
+                            {
+                                Account = this,
+                                Address = address,
+                                Transaction = transactionData
+                            };
+                        }
+
+                        if (!includeImmature && ((transactionData.IsCoinBase ?? false) || (transactionData.IsCoinStake ?? false)) && (confirmationCount >= consensus.CoinbaseMaturity))
```
Aprogiena: In any case it would be nice to separate the parts of the condition to variables and you could reuse that here and about

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -847,7 +864,13 @@ public class TransactionData
         public Money Amount { get; set; }
 
         /// <summary>
-        /// A value indicating whether this is a coin stake transaction or not.
+        /// A value indicating whether this is a coinbase transaction or not.
+        /// </summary>
+        [JsonProperty(PropertyName = "isCoinBase", NullValueHandling = NullValueHandling.Ignore)]
+        public bool? IsCoinBase { get; set; }
```
Aprogiena: When null?

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -784,7 +784,7 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactionsInAccount(Wal
                         $"Account '{walletAccountReference.AccountName}' in wallet '{walletAccountReference.WalletName}' not found.");
                 }
 
-                res = account.GetSpendableTransactions(this.chain.Tip.Height, confirmations).ToArray();
+                res = account.GetSpendableTransactions(this.chain.Tip.Height, consensus, confirmations, includeImmature: includeImmature).ToArray();
```
Aprogiena: Not sure if explicitly naming the argument here has any value

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -985,6 +985,7 @@ private void AddTransactionToWallet(Transaction transaction, TxOut utxo, int? bl
                 var newTransaction = new TransactionData
                 {
                     Amount = amount,
+                    IsCoinBase = transaction.IsCoinBase == false ? (bool?)null : true,
```
Aprogiena: Wouldn't it be better to use false instead of null?

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -15,17 +15,30 @@ namespace Stratis.Bitcoin.IntegrationTests
...
             int startBlock = node.FullNode.Chain.Height - numberOfBlocks + 1;
 
-            return Enumerable.Range(startBlock, numberOfBlocks)
-                .Sum(p => coinviewRule.GetProofOfWorkReward(p));
-        }
+            int maturity = (int)node.FullNode.Network.Consensus.CoinbaseMaturity;
 
+            if (considerMaturity)
+            {
+                if (numberOfBlocks == 1 && maturity > 1)
```
Aprogiena: Brackets

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -15,17 +15,30 @@ namespace Stratis.Bitcoin.IntegrationTests
...
+                    return 0;
 
+                return Enumerable.Range(startBlock,
+                        ((numberOfBlocks - maturity + 1) >= startBlock ? (numberOfBlocks - maturity + 1) : startBlock))
```
Aprogiena: Any explanation about the logic here?

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -74,14 +74,10 @@ public void PremineNodeStartsStaking()
...
         {
-            var network = this.nodes[this.PremineNode].FullNode.Network;
-            var premine = network.Consensus.ProofOfWorkReward + network.Consensus.PremineReward;
-            var mineToMaturity = network.Consensus.ProofOfWorkReward * 110;
-            var balanceShouldBe = premine + mineToMaturity;
             TestHelper.WaitLoop(() =>
             {
                 long staked = this.nodes[this.PremineNode].FullNode.WalletManager().GetSpendableTransactionsInWallet(this.PremineWallet).Sum(s => s.Transaction.Amount);
-                return staked > balanceShouldBe;
+                return staked > 0;
```
fassadlr: Woah dude you are essentially breaking the POS minting integration test that ensures that we can stake coins!? This isn't right at all.

src/Stratis.Bitcoin.IntegrationTests.Common/SharedSteps.cs
```diff
@@ -69,7 +69,10 @@ public void MinePremineBlocks(CoreNode node, string walletName, string walletAcc
             this.WaitForNodeToSync(node);
 
             IEnumerable<UnspentOutputReference> spendable = node.FullNode.WalletManager().GetSpendableTransactionsInWallet(walletName);
-            Money amountShouldBe = node.FullNode.Network.Consensus.PremineReward + node.FullNode.Network.Consensus.ProofOfWorkReward;
+
+            // The premine and other block rewards are not mature yet.
+            Money amountShouldBe = 0;
```
fassadlr: This isn't right as we need to ensure that the wallet has enough coins to stake with, otherwise the weight of the wallet will be too little to stake with.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
fassadlr: I also want to know how this can null? A transaction is either a coinbase tx or its not?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -728,12 +728,29 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentC
 
                     if (confirmationCount >= confirmations)
                     {
```
fassadlr: I would flip the condition here to just `continue` so that the nested `if` can be avoided.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -728,12 +728,29 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentC
...
-                        yield return new UnspentOutputReference
+                        if (includeImmature || (!(transactionData.IsCoinBase ?? false) && !(transactionData.IsCoinStake ?? false)))
                         {
-                            Account = this,
-                            Address = address,
-                            Transaction = transactionData
-                        };
+                            // This output can unconditionally be included in the results.
+
+                            yield return new UnspentOutputReference
```
fassadlr: I feel this whole new `if` can be written better so that we avoid code duplication here.

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -15,17 +15,30 @@ namespace Stratis.Bitcoin.IntegrationTests
...
             var coinviewRule = node.FullNode.NodeService<IConsensusRules>().GetRule<CoinViewRule>();
 
             int startBlock = node.FullNode.Chain.Height - numberOfBlocks + 1;
 
-            return Enumerable.Range(startBlock, numberOfBlocks)
-                .Sum(p => coinviewRule.GetProofOfWorkReward(p));
-        }
+            int maturity = (int)node.FullNode.Network.Consensus.CoinbaseMaturity;
 
+            if (considerMaturity)
```
fassadlr: You can combine this `if` and the one below?

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
fassadlr: Yes please write this a bit better so that its more readable 👍

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
zeptin: The IsCoinStake JSON property is nullable, so I am merely matching that with the introduction of IsCoinBase. This was presumably originally done so that the wallet JSON does not get numerous unnecessary boolean fields (as coinbase/stake transactions are the exception rather than the rule).

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
zeptin: This construct is used for the existing IsCoinStake, I am merely duplicating it.

src/Stratis.Bitcoin.IntegrationTests.Common/SharedSteps.cs
zeptin: The premine is not immediately spendable. Clearly this method only mines sufficient blocks to generate the coins, nowhere near enough for them to be spendable, the comment indicates. I am correcting a constant that was chosen under false assumptions.

zeptin: Perhaps a better way would be to optionally include the immature coins and leave the calculation of the expected balance unchanged. It feels incorrect using a method that gets 'spendable' transactions to do that, though, so maybe a different or new method should be used instead.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
zeptin: Ok, this was very much my bad; I misread what this method was actually doing. I have added `includeImmature` flags to the other GetSpendable* methods so that the test logic can be used essentially unchanged here.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
zeptin: Elsewhere I have explicitly named it because in 99% of cases you do not want to use anything other than `false` for it, so using it at all is unexpected and it should be visually apparent what parameter is being passed. However, since an argument with the same name is being passed in this case, it does not really give additional information. I will remove it.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
Aprogiena: It would be probably better to replace it for the Coinbase as well instead.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -74,10 +74,17 @@ public void PremineNodeStartsStaking()
 
         public void PremineNodeWalletHasEarnedCoinsThroughStaking()
         {
+            var network = this.nodes[this.PremineNode].FullNode.Network;
+            var premine = network.Consensus.ProofOfWorkReward + network.Consensus.PremineReward;
+            var mineToMaturity = network.Consensus.ProofOfWorkReward * 110;
+            var balanceShouldBe = premine + mineToMaturity;
+
+            var spendable = this.nodes[this.PremineNode].FullNode.WalletManager().GetSpendableTransactionsInWallet(this.PremineWallet, includeImmature: true);
```
Aprogiena: I know this is just as for which we don't have many requirements in terms of code style, so just saying that about five lines should not use var

bokobza:
To reiterate my question, why do we have `bool includeImmature = false`?
Is there a scenario where we actually want to be able to send immature coins? As opposed to it being forbidden. @zeptin

zeptin:
@bokobza Removing it is an option, but some tests would have to be rearchitected to take the changed behaviour into account. Fixing the tests wasn't really in the initial time estimate for the task, I suspect, so this may push it out past the end of the sprint. I agree that it is messy having the parameter, and there is no obvious use case for it in reality.

bokobza:
Fair enough. But having tests passing because something not allowed is now allowed is a bit of a problem. 
How many tests are failing because of that? In any case we'll have to fix them.

zeptin:
There are now 3 failing tests, the rest pass:

`A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect` - seems to fail sporadically with a mempool problem. Unclear if this is related to this PR, most likely not

`MockChain_AuctionTest` - failing locally due to regional settings of test machine, issue raised

`TestSegwit_MinedOnCore_ActivatedOn_StratisNode` - seems to randomly pass and fail, not related to this PR

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -710,7 +709,7 @@ public IEnumerable<HdAddress> CreateAddresses(Network network, int addressesQuan
         /// <param name="currentChainHeight">The current height of the chain. Used for calculating the number of confirmations a transaction has.</param>
         /// <param name="confirmations">The minimum number of confirmations required for transactions to be considered.</param>
         /// <returns>A collection of spendable outputs that belong to the given account.</returns>
-        public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentChainHeight, int confirmations = 0)
+        public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentChainHeight, NBitcoin.Consensus consensus, int confirmations = 0, bool includeImmature = false)
```
noescape00: Not sure if we need `includeImmature` parameter at all. Make make it always to be `false`? 

The method here is called `GetSpendableTransactions`, but the immature coins are never spendable. 

If we ever need to get those we might as well introduce a new method like `GetAllUnspentUTXOs` which will include immature coins.

zeptin: I agree, but correcting the behaviour of this method broke a lot of tests that were written under the old, incorrect behaviour. The `includeImmature` parameter is a temporary measure until these tests can be corrected (which I do not really see as being within the scope of the original bug fix).

noescape00: I think if the tests were testing the wrong logic and passing the tests were incorrect. 
We shouldn't adapt the code to allow incorrect behavior so the tests can pass.

bokobza:
The `bool includeImmature = false` parameter should be removed as we should never includeImmature coins.
In order to make the tests pass, you can maybe change the CoinbaseMaturity in Network to 0 where it's necessary because of time (and only because of time). 

Look how we do it in some tests:
https://github.com/stratisproject/StratisBitcoinFullNode/blob/0e232c7c8e7205c79caccc3984d6c7adb9f23fbd/src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletAddressGenerationAndFundsVisibility_Steps.cs#L61-L64

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -77,10 +78,14 @@ public void PremineNodeWalletHasEarnedCoinsThroughStaking()
             var premine = network.Consensus.ProofOfWorkReward + network.Consensus.PremineReward;
             var mineToMaturity = network.Consensus.ProofOfWorkReward * 110;
             var balanceShouldBe = premine + mineToMaturity;
+
+            Money balanceBefore = this.nodes[this.PremineNode].FullNode.WalletManager()
```
bokobza: is balanceShouldBe not relevant anymore?

src/Stratis.Bitcoin.Features.Miner.Tests/PosMintingTest.cs
```diff
@@ -212,7 +212,10 @@ public void StopStake_DisposesResources()
         public async Task GenerateBlocksAsync_does_not_use_small_coins()
         {
             var walletSecret = new WalletSecret(){WalletName = "wallet", WalletPassword = "password"};
-            var wallet = new Wallet.Wallet();
+            var wallet = new Wallet.Wallet()
```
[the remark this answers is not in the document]
zeptin: It crashes without it, so yes. How else would the maturity check know what network it is on?

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -710,7 +709,7 @@ public IEnumerable<HdAddress> CreateAddresses(Network network, int addressesQuan
         /// <param name="currentChainHeight">The current height of the chain. Used for calculating the number of confirmations a transaction has.</param>
         /// <param name="confirmations">The minimum number of confirmations required for transactions to be considered.</param>
         /// <returns>A collection of spendable outputs that belong to the given account.</returns>
-        public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentChainHeight, int confirmations = 0)
+        public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentChainHeight, NBitcoin.Consensus consensus, int confirmations = 0)
```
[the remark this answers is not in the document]
zeptin: This is in HdAccount, which does not have the network.

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSteps.cs
```diff
@@ -61,6 +64,9 @@ protected void a_sending_and_receiving_stratis_bitcoin_node_and_wallet()
 
             this.coinbaseMaturity = (int)this.sendingStratisBitcoinNode.FullNode
                 .Network.Consensus.CoinbaseMaturity;
+
+            this.sendingStratisBitcoinNode.FullNode.Network.Consensus.CoinbaseMaturity = 5L;
```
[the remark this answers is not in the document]
zeptin: I disagree, I am using the variable to hold the original value of the maturity setting and then overriding it for the rest of the test.

The variable could be renamed if it actually matters. I started doing that with tests modified later.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
noescape00: +1. 
It shouldn't be nullable.

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -15,18 +15,31 @@ namespace Stratis.Bitcoin.IntegrationTests
 {
     public static class CoreNodeExtensions
     {
-        public static Money GetProofOfWorkRewardForMinedBlocks(this CoreNode node, int numberOfBlocks)
+        public static Money GetProofOfWorkRewardForMinedBlocks(this CoreNode node, int numberOfBlocks, bool considerMaturity = false)
```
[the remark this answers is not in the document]
zeptin: There is at least one place I use it with the boolean set to true. So the tests were not originally written with a consistent set of assumptions about the spendability of coins.

In particular, it looks like it needs to be false for some miner tests that interrogate the quantity the method returns.

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -36,7 +49,7 @@ public static Money WalletBalance(this CoreNode node, string walletName)
             return node.FullNode.WalletManager().GetSpendableTransactionsInWallet(walletName).First().Transaction.BlockHeight;
         }
 
-        public static int WalletSpendableTransactionCount(this CoreNode node, string walletName)
+        public static int WalletSpendableTransactionCount(this CoreNode node, string walletName, bool includeImmature = false)
```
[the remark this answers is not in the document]
zeptin: Indeed, fixed.

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -15,18 +15,31 @@ namespace Stratis.Bitcoin.IntegrationTests
...
+                    .Sum(p => coinviewRule.GetProofOfWorkReward(p));
+            }
+            else
+            {
+                return Enumerable.Range(startBlock, numberOfBlocks)
+                    .Sum(p => coinviewRule.GetProofOfWorkReward(p));
+            }
+        }
+        
+        public static Money WalletBalance(this CoreNode node, string walletName, bool includeImmature = false)
```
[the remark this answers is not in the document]
zeptin: Fixed

src/Stratis.Bitcoin.IntegrationTests.Common/SharedSteps.cs
```diff
@@ -68,9 +71,12 @@ public void MinePremineBlocks(CoreNode node, string walletName, string walletAcc
 
             this.WaitForNodeToSync(node);
 
-            IEnumerable<UnspentOutputReference> spendable = node.FullNode.WalletManager().GetSpendableTransactionsInWallet(walletName);
+            // Since the premine will not be immediately spendable, the transactions have to be counted directly from the address.
```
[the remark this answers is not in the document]
zeptin: The premine is immature just after it is mined, the change checks that it has been mined but does not rely on `GetSpendableTransactionsInWallet` any more.

fassadlr:
@zeptin So I took the wallet only changes from your PR and ran the staking test to see why it would fail given your changes. It turns out that the integration test fails because the wallet doesn't have enough coins to stake with.

I dug a bit deeper into the logs and saw that after staking starts, 1 coin gets minted and directly after when we try and get the spendable balance to continue staking only 440 coins gets returned, which isnt nearly enough for the wallet to mint any new coins. The reason for this is your code change filters out the coinstake utxo after the 1 coin mint and because the node isn't connected to any other nodes the staked block doesn't mature (the node doesnt advance) and essentially the test gets stuck.

I think it should be an easy fix to add another node into the mix to advance the chain so that the stakeable coins matures and therefore can be staked.

You can see the logs here:

```
[2018-08-14 17:02:37.9858 3] INFO: Stratis.Bitcoin.Features.Miner.PowMining.OnBlockMined Mined new POW block: '112-5f1e2461686932dd56f2c60da4edc0b38003a27411ae50ac6516cfa3fd19d44f'.
[2018-08-14 17:02:37.9945 21] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0+<<StartAsync>b__0>d.MoveNext PosMining.Stake starting
[2018-08-14 17:02:39.1466 15] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting+<CreateCoinstakeAsync>d__43.MoveNext Node staking with 98000404.00000000 (0.00 % of the network weight 0.00000000), est. time to find new block is 00:00:00.
[2018-08-14 17:02:42.8083 16] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__75_0 ======Node stats====== 01/01/2018 02:01:20 agent node11666-StratisBitcoin:1.1.12
Headers.Height:      112      Headers.Hash:     5f1e2461686932dd56f2c60da4edc0b38003a27411ae50ac6516cfa3fd19d44f
BlockStore.Height:   112      BlockStore.Hash:  5f1e2461686932dd56f2c60da4edc0b38003a27411ae50ac6516cfa3fd19d44f
Consensus.Height:    112      Consensus.Hash:   5f1e2461686932dd56f2c60da4edc0b38003a27411ae50ac6516cfa3fd19d44f
Wallet.Height:       112      Wallet.Hash:      5f1e2461686932dd56f2c60da4edc0b38003a27411ae50ac6516cfa3fd19d44f

======BlockStore======
Pending Blocks: 0
Batch Size: 0 kb / 5000 kb

=======Mempool=======
MempoolSize: 0    DynamicSize: 0 kb   OrphanSize: 0   

======Wallets======
Wallet: preminewallet,       Confirmed balance: 98000444.00000000

======Connection======


[2018-08-14 17:02:43.7921 5] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting.CheckStake ==================================================================
[2018-08-14 17:02:43.8920 5] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting.CheckStake Found new POS block hash '9b01e630213dac54431234f863189c00cd0161f34a8684c111e80e5929e60644' at height 113.
[2018-08-14 17:02:43.8939 5] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting.CheckStake ==================================================================
[2018-08-14 17:02:43.8939 15] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting+<CreateCoinstakeAsync>d__43.MoveNext Node staking with 408.00000000 (0.00 % of the network weight 0.00000000), est. time to find new block is 00:00:00.
[2018-08-14 17:02:47.8821 20] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__75_0 ======Node stats====== 01/01/2018 02:02:25 agent node11666-StratisBitcoin:1.1.12
Headers.Height:      113      Headers.Hash:     9b01e630213dac54431234f863189c00cd0161f34a8684c111e80e5929e60644
BlockStore.Height:   113      BlockStore.Hash:  9b01e630213dac54431234f863189c00cd0161f34a8684c111e80e5929e60644
Consensus.Height:    113      Consensus.Hash:   9b01e630213dac54431234f863189c00cd0161f34a8684c111e80e5929e60644
Wallet.Height:       113      Wallet.Hash:      9b01e630213dac54431234f863189c00cd0161f34a8684c111e80e5929e60644

======BlockStore======
Pending Blocks: 0
Batch Size: 0 kb / 5000 kb

=======Mempool=======
MempoolSize: 0    DynamicSize: 0 kb   OrphanSize: 0   

======Wallets======
Wallet: preminewallet,       Confirmed balance: 444.00000000
```

zeptin:
@fassadlr Either way the test has to be changed from what it originally was. I don't see how adding another node is necessarily better than the change I made to that particular test (a change which is already passing).

Unless the logic of the test can be shown to be wrong, I believe it is testing what it is supposed to, that coins can be staked and result in new utxos in the wallet.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -257,7 +260,8 @@ private void RecoverViaExtPubKey(string walletName, string extPubKey, int accoun
             {
                 ExtPubKey = extPubKey,
                 AccountIndex = accountIndex,
-                Name = walletName
+                Name = walletName,
+                CreationDate = DateTime.Now
```
bokobza: why the change?

src/Stratis.Bitcoin.IntegrationTests/BlockStore/ProofOfWorkSpendingSteps.cs
```diff
@@ -23,11 +23,11 @@ public partial class ProofOfWorkSpendingSpecification
         private const string AccountName = "account 0";
         private CoreNode sendingStratisBitcoinNode;
         private CoreNode receivingStratisBitcoinNode;
-        private int coinbaseMaturity;
         private Exception caughtException;
         private Transaction lastTransaction;
         private SharedSteps sharedSteps;
         private NodeGroupBuilder nodeGroupBuilder;
+        private int coinbaseMaturity;
```
bokobza: why has this moved down?

src/Stratis.Bitcoin.IntegrationTests/BlockStore/RetrieveFromBlockStoreSteps.cs
```diff
@@ -62,6 +62,9 @@ private void a_pow_node_running()
             this.node = this.builder.CreateStratisPowNode(this.network);
             this.node.Start();
             this.node.NotInIBD();
+
+            this.maturity = 1;
```
bokobza: Can we set the maturity in `BeforeTest` or even in `private int maturity = 1;` so that it's more visible, or would that not work?

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -46,8 +46,10 @@ public void AddToMempool()
                 CoreNode stratisNodeSync = builder.CreateStratisPowNode(this.network);
                 builder.StartAll();
 
+                stratisNodeSync.FullNode.Network.Consensus.CoinbaseMaturity = 1L;
+
                 stratisNodeSync.SetDummyMinerSecret(new BitcoinSecret(new Key(), stratisNodeSync.FullNode.Network));
-                stratisNodeSync.GenerateStratisWithMiner(105); // coinbase maturity = 100
+                stratisNodeSync.GenerateStratisWithMiner(5);
```
bokobza: maybe create `int maturity = 5` and then do `stratisNodeSync.GenerateStratisWithMiner(maturity + 5);` so that the calculation is clear?

bokobza: could you not change the network maturity in the ctor? then it'll be available for the whole class.

bokobza:
@zeptin I've made a few comments. These are valid for all the tests so please apply where possible.
Also please revert the changes aimed at improving the speed of the tests. If the test works, it's fine to wait 100 more blocks. Otherwise it makes it hard to figure out what **needed** to be changed vs what is an improvement. We can do the improvements in a further PR.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -57,31 +62,50 @@ public void MineGenesisAndPremineBlocks()
...
             TestHelper.WaitLoop(() =>
             {
-                long staked = this.nodes[this.PremineNode].FullNode.WalletManager().GetSpendableTransactionsInWallet(this.PremineWallet).Sum(s => s.Transaction.Amount);
-                return staked > balanceShouldBe;
+                foreach (TransactionData transactionData in this.nodes[this.PremineNode].FullNode.WalletManager().Wallets
+                    .First()
+                    .GetAllTransactionsByCoinType((CoinType)this.nodes[this.PremineNode].FullNode.Network.Consensus
+                        .CoinType))
+                {
+                    if (!this.transactionsBeforeStaking.Contains(transactionData.Id))
```
fassadlr: @zeptin OK I understand now why this change is required. As a secondary measure, can we perhaps check the coinstake flag on the transactionData to verify that it IS in fact a coinstake tx? We can also check the value so that it matches the network's stake reward?

The whole point of this test is to ensure that we can stake and mint new coins, so just checking that there is a new tx might not be entirely sufficient?

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
zeptin: It fixes the failing `CreateExtPubKeyOnlyWallet_creates_wallet_with_extra_flag` test that no one else seemed to have bothered about yet.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
zeptin: Yeah, that seems sensible, I've added a check for that.

Checking the value might be overkill, but it isn't hard to do. It is however dependent on how many outputs the coinstake consists of (I think it is 8 by default, but this can be overridden and could change in future).

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
bokobza: You mean this test was failing before that change?

zeptin: As far as I am aware, yes. `AddNewAccount_for_xpub_only_wallet_informs_user_to_create_new_wallet_` as well.

bokobza: These 2 don't fail for me or for the others. The integration tests are not running smoothy and sometimes a rerun of the failed ones fixes the issue.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -350,8 +350,6 @@ public void MempoolSyncTransactions()
                 // generate blocks and wait for the downloader to pickup
                 stratisNodeSync.SetDummyMinerSecret(new BitcoinSecret(new Key(), stratisNodeSync.FullNode.Network));
                 stratisNodeSync.GenerateStratisWithMiner(105); // coinbase maturity = 100
-                // wait for block repo for block sync to work
```
bokobza: don't we need to check if the node is synced?

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -56,12 +56,15 @@ private void two_connected_nodes()
                     .Connect(NodeOne, NodeTwo)
                     .AndNoMoreConnections()
                 .Build();
+
+            this.CoinBaseMaturity = 1;
```
bokobza: Here you can do `KnownNetworks.RegTest.Consensus.CoinbaseMaturity = 1;` in the beforeTests method.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
zeptin: The test failed for me unless I removed that. I tried it repeatedly.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingTxWithDoubleSpendSpecification_Steps.cs
```diff
@@ -100,8 +100,7 @@ private void receiving_node_attempts_to_double_spend_mempool_doesnotaccept()
 
         private void trx_is_mined_into_a_block_and_removed_from_mempools()
         {
-            new SharedSteps().MineBlocks(1, this.stratisSender, "account 0", "mywallet", "123456", 16360L);
-
+            new SharedSteps().MineBlocks(1, this.stratisSender, "account 0", "mywallet", "123456");
```
bokobza: there is no more fees for the mine block?

zeptin: The test fails if I do not change this.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -57,31 +62,50 @@ public void MineGenesisAndPremineBlocks()
...
         {
-            this.nodes[this.PremineNode].GenerateStratisWithMiner(100);
+            this.nodes[this.PremineNode].GenerateStratisWithMiner(Convert.ToInt32(this.nodes[this.PremineNode].FullNode.Network.Consensus.CoinbaseMaturity));
             this.sharedSteps.WaitForNodeToSync(this.nodes[this.PremineNode]);
         }
 
         public void PremineNodeMinesTenBlocksMoreEnsuringTheyCanBeStaked()
         {
-            this.nodes[this.PremineNode].GenerateStratisWithMiner(Convert.ToInt32(this.nodes[this.PremineNode].FullNode.Network.Consensus.CoinbaseMaturity));
+            this.nodes[this.PremineNode].GenerateStratisWithMiner(10);
```
bokobza: should that not be "10 + maturity" instead of 10?

fassadlr: @bokobza `this.nodes[this.PremineNode].GenerateStratisWithMiner(10);` is correct as we want to advance the node 10 blocks so that they coins become "stake-able"

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
```diff
@@ -21,11 +21,17 @@ public static Money GetProofOfWorkRewardForMinedBlocks(this CoreNode node, int n
 
             int startBlock = node.FullNode.Chain.Height - numberOfBlocks + 1;
 
-            return Enumerable.Range(startBlock, numberOfBlocks)
+            int maturity = (int)node.FullNode.Network.Consensus.CoinbaseMaturity;
```
bokobza: The goal of this method is simply to return the reward, whether or not the rewards are spendable or not.
Are you sure this change is necessary? It shouldn't be. If it is then it's the tests using it that need looking into.

src/Stratis.Bitcoin.IntegrationTests/Wallet/WalletTests.cs
```diff
@@ -289,44 +286,39 @@ public void Given_TheNodeHadAReorg_And_WalletTipIsBehindConsensusTip_When_ANewBl
 
                 stratisSender.GenerateStratisWithMiner(10);
 
-                // wait for block repo for block sync to work
-                TestHelper.WaitLoop(() => TestHelper.IsNodeSynced(stratisSender));
-
-                //// sync all nodes
+                // Sync all nodes
```
bokobza: Is it not better to wait for one node to be fully synced, before making sure all the nodes are in sync with each other?

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
zeptin: I spent the last two weeks looking at the tests. This change is the least required to get everything to work.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -726,7 +728,25 @@ public IEnumerable<UnspentOutputReference> GetSpendableTransactions(int currentC
...
 
-                    if (confirmationCount >= confirmations)
+                    if (confirmationCount < confirmations)
+                        continue;
+
+                    bool isCoinBase = transactionData.IsCoinBase ?? false;
+                    bool isCoinStake = transactionData.IsCoinStake ?? false;
+
+                    // This output can unconditionally be included in the results.
+                    if (!isCoinBase && !isCoinStake)
```
bokobza: Maybe we can combine the 2 ifs if we format it like this:
```
if ((!isCoinBase && !isCoinStake) ||
	((isCoinBase || isCoinStake) && (confirmationCount >= network.Consensus.CoinbaseMaturity))
{
	yield return new UnspentOutputReference
	{
		Account = this,
		Address = address,
		Transaction = transactionData
	};
}
```

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
bokobza: Because the creationDate used is the default one: 01/01/0001 and doesn't fail.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
bokobza: A test should never fail if we wait a bit longer for a node to be synced.
There is only one test that we know always fail and it's "A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect".

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
bokobza: ok so I guess before we mined coins of quantity "maturity" and that was too much?

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
bokobza: > This change is the least required to get everything to work.

I don't understand how this is related.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
zeptin: It does default, but it still throws:

```
{System.ArgumentOutOfRangeException: The UTC time represented when the offset is applied must be between year 0 and 10,000.
Parameter name: offset
   at System.DateTimeOffset.ValidateDate(DateTime dateTime, TimeSpan offset)
   at System.DateTimeOffset..ctor(DateTime dateTime)
   at System.DateTimeOffset.op_Implicit(DateTime dateTime)
   at Stratis.Bitcoin.Features.Wallet.WalletManager.RecoverWallet(String name, ExtPubKey extPubKey, Int32 accountIndex, DateTime creationTime) in C:\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Wallet\WalletManager.cs:line 369
   at Stratis.Bitcoin.Features.Wallet.Controllers.WalletController.RecoverViaExtPubKey(WalletExtPubRecoveryRequest request) in C:\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Wallet\Controllers\WalletController.cs:line 285}
```

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
zeptin: The test does fail if it takes the node longer than the helper method's timeout to sync. Why it might be taking longer than that, I do not know.

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
zeptin: I am not sure how I am supposed to simultaneously get the tests to pass, and not change anything about them. This method gets used implicitly every time `SharedSteps.MineBlocks` is called, to ensure internal consistency. Its result gets compared to the increase in wallet balance, and any discrepancy causes the test to immediately fail. As the wallet balance is dependent on the underlying bug this PR fixes, it necessitates that this function be changed too. The alternative is to change every test that uses MineBlocks, which results in a lot more changed code & possible structural changes to the tests.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
zeptin: Done

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
zeptin: Done

src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/CoreNodeExtensions.cs
bokobza: This method is used in `GetProofOfWorkRewardForMinedBlocksTest` to check what's the miner reward after mining, but here you've changed the test there to match your change.
And you've added a test called `GetProofOfWorkRewardForMinedBlocksTestLowMaturity` that includes maturity in your calculation but it's not related to the miner reward.
@zeptin @fassadlr

PR #1762, opened 2018-07-29
Fix SmartContractGasInjector failing gas injection with recursive methods

### LINKED ISSUE #1697: SmartContractGasInjector fails with recursive methods.
The SmartContractGasInjector throws `System.InvalidOperationException: 'Operation is not valid due to the current state of the object.'` when injecting gas into a recursive method.

Simple, albeit contrived example:

```
    public void ThrowException()
    {
        ThrowException();
    }
```

We should have a test for a recursive example, like above, in GasInjectorTests.cs, and have a fix so this is resolved.

Current Bounty: 10 STRAT

nathanielcook:
Fixes #1697 

The test contract method is non-deterministic in terms of how many iterations it makes. Seems sufficient in terms of test coverage for this case, but let me know if you disagree.


codingupastorm:
First glance looks great @nathanielcook. Will have a more robust look tomorrow :) 

Could you do one thing - retarget this to sc-dev branch instead of master? Cheers

nathanielcook:
Looks like I need to retarget in Git. Will close this one and reopen

Issue #1194, opened 2018-03-09
[PeerAddressManager] Stop connecting to peers that connected but failed handshake within a certain time frame.


fassadlr:
We should throttle connections to peers that we've connected to but didnt manage to version handshake. 

I.e. if we connected to a peer X but failed version hand shake, that attempt should also be logged and checked when we try and connect to it again. After a certain amount of attempts we shouldn't try and connect to peer X within a certain time frame, possibly a day or so.

dangershony:
We need to extend the time between failed handshake attempts

fassadlr:
@thefazzer we will somehow need to inform the address manager of the fact that version handshake failed. For this I recommend we create a `VersionHandshakeException` which we can then catch when we try and connect. Please see `PeerConnector.ConnectAsync`, I would change this like so:

```
            try
            {
                ...
            }
            catch (OperationCanceledException)
            {
                ...
            }
            catch (VersionHandshakeException exception)
            {
                this.logger.LogTrace("The peer failed to handshake: {0}", exception.ToString());
                peer?.VersionHandshakeFailed();
                peer?.Disconnect("Error while connecting", exception);
            }
            catch (Exception exception)
            {
                ...
            }
```

We can then implement the relevant methods etc and throw the exception in `INetworkPeer.VersionHandshakeAsync`.

@Aprogiena @dangershony do you guys have a better suggestion?

Aprogiena:
```
        /// <exception cref="ProtocolException">Thrown when the peer rejected our "version" message.</exception>
        /// <exception cref="OperationCanceledException">Thrown during the shutdown or when the peer disconnects.</exception>
        Task VersionHandshakeAsync(NetworkPeerRequirement requirements, CancellationToken cancellationToken);
```

We already have these two exceptions and we already handled them both (protocol is caught with Exception class), so we only need to call the logic that will mark it down.

I think we should not try to reconnect to the peer anytime soon even after the first failure. It probably doesn't make sense to try to connect to someone for at least an hour if we just failed establish the connection properly – including handshake. Then of course, as usual, we should try again eventually and if we fail again for multiple times then to mark this node as completely useless and not to try again.

fassadlr:
@Aprogiena Oh yes OK I see now. However, I see that we also use ProtocolException for exceptions in `CoreNode`'s `SynchronizeChain` method:

```
            if (newTip.Height <= oldTip.Height)
[...]

PR #1767, opened 2018-07-30
 Don't connect to failed handshake peers #1194

thefazzer:
**Stop connecting to peers that connected but failed handshake within a certain time frame #1194**

- Throttle connections to peers that we've connected to but didn't manage to version handshake.
- Log attempts to version hand shake. 
- After a certain amount of attempts we shouldn't try and connect to peer X within a certain time frame.

Preferred: @fassadlr @Aprogiena


fassadlr:
@thefazzer I'll review this one 👍

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -272,6 +272,8 @@ public IEnumerable<PeerAddress> SelectPeersForGetAddrPayload(int peerCount)
             //peers we have not yet connected to and/or that are fresh.
             if (connectedAndHandshaked.Any())
             {
+                connectedAndHandshaked = this.GetPeersNotBannedForBadHandShakes(connectedAndHandshaked).ToList();
```
dangershony: Better name is probably `FilterBadHandshakedPeers`

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +270,14 @@ public void StartConnectAsync()
                     peer?.Disconnect("Connection timeout");
                 }
             }
+
+            catch (NBitcoin.Protocol.ProtocolException)
+            {
+                this.logger.LogDebug("Peer {0} handshake failed.", peerAddress.Endpoint);
```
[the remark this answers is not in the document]
Aprogiena: I think so, there are several ways how the handshake can fail in this one is triggered only when our version message is rejected by the other side, so it might be slightly better to say something like
```
this.logger.LogDebug("Handshake rejected by peer '{0}'.", peerAddress.Endpoint);
```

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +270,14 @@ public void StartConnectAsync()
                     peer?.Disconnect("Connection timeout");
                 }
             }
+
```
Aprogiena: No empty line here.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +270,14 @@ public void StartConnectAsync()
                     peer?.Disconnect("Connection timeout");
```
Aprogiena: This also counts as failed handshake

src/Stratis.Bitcoin/P2P/PeerConnector.cs
```diff
@@ -270,6 +270,14 @@ public void StartConnectAsync()
...
+            catch (NBitcoin.Protocol.ProtocolException)
+            {
+                this.logger.LogDebug("Peer {0} handshake failed.", peerAddress.Endpoint);
+                peerAddress.SetHandshakeAttempted(this.dateTimeProvider.GetUtcNow(), handshakeSuccessful: false);
+                peer?.Disconnect("Error while handshaking");
+            }
+
             catch (Exception exception)
             {
                 this.logger.LogTrace("Exception occurred while connecting: {0}", exception.ToString());
```
Aprogiena: And also this should count as failed handshake

src/Stratis.Bitcoin/P2P/PeerSelector.cs
Aprogiena: Agree

src/Stratis.Bitcoin/P2P/PeerSelector.cs
```diff
@@ -319,6 +321,12 @@ public IEnumerable<PeerAddress> Attempted()
         }
 
         /// <inheritdoc/>
+        public IEnumerable<PeerAddress> GetPeersNotBannedForBadHandShakes(IEnumerable<PeerAddress> peers)
+        {
+            return peers.Where(p => p.HandshakedAttempts < PeerAddress.AttemptHandshakeThreshold ||
```
Aprogiena: Brackets

Aprogiena:
> Should we also set handshake rejected for inbounds?
> In the method NetworkFactory.CreateConnectedNetworkPeerAsync

I think so, currently this only handles failures from NP.VersionHandshakeAsync, but when inbound connection is used, NP.RespondToHandShakeAsync is called instead. This one can also produce number of exceptions and I think it should be handled similarly, but not exactly in the same fashion. For example disconnection of the peer in outbound case should lead to failed handshake situation, while in case of inbound, the disconnection should not lead to failed handshake situation. ProtocolException, however, should lead to failed handshake situation in both cases.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
fassadlr: The filtering shouldnt happen here @thefazzer. It should happen in `SelectPreferredPeers`

fassadlr: `SelectPeersForGetAddrPayload` is when a peer asks another for it's addresses.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -70,6 +83,21 @@ public IPAddress Loopback
...
+        public DateTimeOffset? LastHandshakeAttempt { get; private set; }
+
+        /// <summary>
+        /// The last unsuccessful handshake attempt.
+        /// <para>
+        /// This is reset when a handshake is done.
+        /// </para>
+        /// </summary>
+        [JsonProperty(PropertyName = "lastHandshakeFailure", NullValueHandling = NullValueHandling.Ignore)]
+        public DateTimeOffset? LastHandshakeFailure { get; private set; }
```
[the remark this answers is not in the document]
dangershony: Ah yeah good point they are basically the same thing right?

src/Stratis.Bitcoin/P2P/PeerSelector.cs
thefazzer: Done.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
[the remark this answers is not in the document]
thefazzer: Done.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Removed.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Updated.

src/Stratis.Bitcoin/P2P/PeerConnector.cs
thefazzer: Updated.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
thefazzer: Added.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
```diff
@@ -226,7 +254,23 @@ internal void SetAttempted(DateTime peerAttemptedAt)
             this.LastConnectionSuccess = null;
             this.LastConnectionHandshake = null;
         }
+        
+        internal void SetHandshakeAttempted(DateTimeOffset peerAttemptedAt, bool handshakeSuccessful)
```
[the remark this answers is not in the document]
thefazzer: Refactored.

src/Stratis.Bitcoin/P2P/PeerAddress.cs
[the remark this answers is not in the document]
thefazzer: Removed.

src/Stratis.Bitcoin/P2P/PeerSelector.cs
thefazzer: Refactored.

dangershony:
I see what you mean @fassadlr an inbound peer that has bad version we will not necessarily have it in our local db so this means adding an entry and banning the peers, I feel perhaps its not that necessary so am happy we dont do it on this PR and perhaps open an issue to see if its necessary at all.

PR #1772, opened 2018-07-31
Load NLog.config from the DataFolder

### LINKED ISSUE #1336: NLog file to be picked from datafolder
More info here
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1307

rowandh:
Closes #1336.

Adds a method to set up the NLog configuration using the NLog.config in the DataDir. This occurs after we establish the path to the DataDir, but before the programmatic configuration takes place. 

The order of logger config precedence is: 
DataDir NLog.config > Regular [NLog.config path resolution](https://github.com/nlog/NLog/wiki/Configuration-file#configuration-file-locations) > Default empty NLogConfiguration. 
Programmatic rules are always loaded.

There's an existing chicken-and-egg problem where a Logger is initialized and used to log within `NodeSettings` before it has been configured, but can't be configured until after settings are read. Some calls made to logger before configuration do not actually go anywhere. We can move some existing logging to occur after `AddFilters` has been called, otherwise I don't see an easy fix for this.

Filename "NLog.config"is case-sensitive on Unix platforms other than OSX.

### DIFF
--- a/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
+++ b/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
@@ -28,7 +28,7 @@ namespace Stratis.Bitcoin.Configuration.Logging
         /// <summary>Provider of console logger.</summary>
         public ConsoleLoggerProvider ConsoleLoggerProvider { get; set; }
 
-        /// <summary>Loads an NLog.config file from the <see cref="DataFolder"/>, if it exists</summary>
+        /// <summary>Loads an NLog.config file from the <see cref="DataFolder"/>, if it exists.</summary>
         public void LoadNLogConfiguration(DataFolder dataFolder)
         {
             if (dataFolder == null)
--- a/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
+++ b/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
@@ -28,7 +28,7 @@ namespace Stratis.Bitcoin.Configuration.Logging
         /// <summary>Provider of console logger.</summary>
         public ConsoleLoggerProvider ConsoleLoggerProvider { get; set; }
 
-        /// <summary>Loads an NLog.config file from the <see cref="DataFolder"/>, if it exists.</summary>
+        /// <summary>Loads the NLog.config file from the <see cref="DataFolder"/>, if it exists.</summary>
         public void LoadNLogConfiguration(DataFolder dataFolder)
         {
             if (dataFolder == null)
--- a/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
+++ b/src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
@@ -7,6 +7,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Console;
 using NLog;
 using NLog.Config;
+using NLog.Extensions.Logging;
 using NLog.Targets;
 using NLog.Targets.Wrappers;
 using Stratis.Bitcoin.Configuration.Settings;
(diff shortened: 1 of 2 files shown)


bokobza:
@Aprogiena is probably the best person to review this.
@Aprogiena would you mind?

src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
```diff
@@ -19,11 +20,24 @@ namespace Stratis.Bitcoin.Configuration.Logging
...
         /// <summary>Provider of console logger.</summary>
         public ConsoleLoggerProvider ConsoleLoggerProvider { get; set; }
+
+        /// <summary>Loads an NLog.config file from the <see cref="DataFolder"/>, if it exists</summary>
+        public void LoadNLogConfiguration(DataFolder dataFolder)
+        {
+            if (dataFolder == null)
+                return;
+
+            string configPath = Path.Combine(dataFolder.RootPath, NLogConfigFileName);
```
monsieurleberre: Maybe inside a try catch ? unless we are OK with a bad config file or IO issue to mess things up ...

Aprogiena:
I would definitely want @noescape00 to actually test it. But yes, I will review the change in the code as well

Aprogiena:
@noescape00 Let me know when you get to this one

src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
```diff
@@ -19,11 +20,24 @@ namespace Stratis.Bitcoin.Configuration.Logging
...
     {
+        private const string NLogConfigFileName = "NLog.config";
+
         /// <summary>Configuration of console logger.</summary>
         public ConsoleLoggerSettings ConsoleSettings { get; set; }
 
         /// <summary>Provider of console logger.</summary>
         public ConsoleLoggerProvider ConsoleLoggerProvider { get; set; }
+
+        /// <summary>Loads an NLog.config file from the <see cref="DataFolder"/>, if it exists</summary>
```
Aprogiena: missing `.`

also i'd replace `NLog.config` with either `configuration` or `<see cref="NLogConfigFileName"/>`

src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
Aprogiena: We either would need to crash the node or to display some message to the user. The worst thing would be to silently get the exception and discard it.

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -174,6 +175,9 @@ public class NodeSettings : IDisposable
             // Set the data folder.
             this.DataFolder = new DataFolder(this.DataDir);
 
+            // Attempt to load NLog configuration from the DataFolder
```
Aprogiena: Missing the `.`

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -96,7 +96,8 @@ public class NodeSettings : IDisposable
             string agent = "StratisBitcoin", string[] args = null)
         {
             // Create the default logger factory and logger.
-            this.LoggerFactory = new ExtendedLoggerFactory();
+            var loggerFactory = new ExtendedLoggerFactory();
```
Aprogiena: I don't really get introduction of this local variable, but okay, but why is it not used everywhere then?

rowandh: LoggerFactory is `ILoggerFactory`, we need access to `ExendedLoggerFactory` to call the load method

Aprogiena:
So I have focused on checking that logging is enabled soon enough, which it is, so I have no problem with this, just couple of formal things and then I want to test with @noescape00

src/Stratis.Bitcoin/Configuration/Logging/LoggingConfiguration.cs
rowandh: `configuration` is not specific, and <see cref="NLogConfigFileName"/> is obscure. NLog.config is exactly the name of the file that it loads.

rowandh: I can inline the constant if you'd prefer?

Aprogiena: i'd probably prefer `logging configuration`, inlining constants is not a good idea

bokobza:
Thanks @Aprogiena

Aprogiena:
I now realised that this change in the code should be also reflected in the documentation: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/Documentation/using-logging.md#nlogconfig

rowandh:
@Aprogiena good suggestion, I will make another PR for that.

Aprogiena:
Cool

Issue #1689, opened 2018-07-20
TransactionBuildContext should only be used internally by the WalletTransactionHandler.


codingupastorm:
Currently `WalletTransactionHandler` uses `TransactionBuildContext` internally, but this same object is passed in to `BuildTransaction` as a kind of option. What ends up happening is you can set values on the context and pass it into `BuildTransaction` but they will be completely overwritten by the internal functionality of the `WalletTransactionHandler`.

There should really be a separate options object to pass into `BuildTransaction` that has fields like `ChangeAddress` which you can set. 

Would be beneficial for SC and the FN! 

@majikandy @dangershony can you guys add this to the board and give appropriate labels? <3

rowandh:
@codingupastorm while you're looking at this, can you take a look at the TransactionBuilder property and see if it makes sense to remove the property and instead add a method `Build(TransactionBuilder builder)` on TransactionBuildContext?

If it does we can add another issue, feel free to assign to me.

codingupastorm:
Hey @bokobza the PR is still open. If you're happy with the changes want to merge it? I made a last little refactor and dropped a comment on the PR too, it should help you out a bit!

Thanks

PR #1773, opened 2018-07-31
TransactionBuildOptions now a separate entity to TransactionBuildContext

codingupastorm:
TransactionBuildContext is now an internal class - used only for passing transaction data around inside the Handler. 

It's now very clear what is an option when building transactions (TransactionBuildOptions) and what is being used internally by the WalletTransactionHandler. 

Also updated many tests to check actual transaction fees rather than context fees.


majikandy:
fixes #1689

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+            this.Recipients = recipients;
+
+            // Set defaults for options
+            this.ChangeAddress = null;
+            this.FeeType = FeeType.Medium;
+            this.MinConfirmations = 1;
+            this.OpReturnData = null;
+            this.OverrideFeeRate = null;
+            this.SelectedInputs = new List<OutPoint>();
+            this.ShuffleOutputs = false;
```
majikandy: is this the opposite to what it was before when it being missing from the request and was defaulted to true for "better anonymity ?

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+        public bool ShuffleOutputs { get; set; }
+
+        /// <summary>
+        /// Set the transaction fee explicitly for this transaction.
+        /// </summary>
+        public Money TransactionFee { get; set; }
+
+        public TransactionBuildOptions(WalletAccountReference wallet, string password, List<Recipient> recipients)
+        {
+            Guard.NotNull(recipients, nameof(recipients));
```
majikandy: what calls this? would be better if this could never be null

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            // Set required fields
+            this.WalletAccountReference = wallet;
+            this.WalletPassword = password;
+            this.Recipients = recipients;
+
+            // Set defaults for options
```
majikandy: methodise and remove comment

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Set the transaction fee explicitly for this transaction.
+        /// </summary>
+        public Money TransactionFee { get; set; }
+
+        public TransactionBuildOptions(WalletAccountReference wallet, string password, List<Recipient> recipients)
+        {
+            Guard.NotNull(recipients, nameof(recipients));
+
+            // Set required fields
```
majikandy: comment can be removed when the other comment below changed to a method

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -705,28 +703,29 @@ public IActionResult BuildTransaction([FromBody] BuildTransactionRequest request
             try
             {
                 Script destination = BitcoinAddress.Create(request.DestinationAddress, this.network).ScriptPubKey;
-                var context = new TransactionBuildContext(
-                    this.network,
+
+                var txBuildOptions = new TransactionBuildOptions(
```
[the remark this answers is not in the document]
codingupastorm: If I made the rule it would roughly be:

Class/interface names must use Transaction. 
Fields / variables / properties may use tx or Transaction.

Some lines are too long with spelling out Transaction every time, and I would try to discourage use of txn.

Happy to amend here if you wish.

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// Whether to randomly order the outputs of the transaction to preserve privacy.
+        /// </summary>
+        public bool ShuffleOutputs { get; set; }
+
+        /// <summary>
+        /// Set the transaction fee explicitly for this transaction.
+        /// </summary>
+        public Money TransactionFee { get; set; }
+
+        public TransactionBuildOptions(WalletAccountReference wallet, string password, List<Recipient> recipients)
```
majikandy: wallet -> walletAccountReference

majikandy: we are currently supposed to have XML doc on public constructors - no need to repeat the parameters if it would just be a copy paste of the xml doc above on the properties

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
codingupastorm: Is public, copied over from BuildContext, so rather leave it for now. Lets remove guards in future

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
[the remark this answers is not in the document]
majikandy: Nah, don't amend, let's keep tx since it is used in the codebase a lot. Just food for thought.

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
majikandy: ok

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -247,11 +254,24 @@ protected void AddSecrets(TransactionBuildContext context)
...
         /// <summary>
-        /// Find the next available change address.
+        /// If a ChangeAddress was set in the options send change there, otherwise find the next available change address.
         /// </summary>
         /// <param name="context">The context associated with the current transaction being built.</param>
         protected void FindChangeAddress(TransactionBuildContext context)
         {
+            if (context.BuildOptions.ChangeAddress != null)
+            {
+                // Ensure the given ChangeAddress is from the same wallet.
```
majikandy: Methodise this protection instead of a comment

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -247,11 +254,24 @@ protected void AddSecrets(TransactionBuildContext context)
...
+                if (!wallet.ContainsAddress(context.BuildOptions.ChangeAddress))
+                {
+                    throw new WalletException("Change address given does not belong to this wallet");
+                }
+
+                context.TransactionBuilder.SetChange(context.BuildOptions.ChangeAddress.ScriptPubKey);
+                return;
+            }
+
             // Get an address to send the change to.
```
majikandy: comment would be redundant here if you methodise the two things within FindChangeAddress

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -380,146 +400,128 @@ protected void AddOpReturnOutput(TransactionBuildContext context)
...
-        /// The destination script.
-        /// </summary>
-        public Script ScriptPubKey { get; set; }
+            public TransactionBuildContext(Network network, TransactionBuildOptions options)
+            {
+                this.TransactionBuilder = new TransactionBuilder(network);
+                this.BuildOptions = options;
+
+                this.AccountReference = options.WalletAccountReference;
+                this.Recipients = options.Recipients;
```
majikandy: I see the guard is removed on this now, it is because it can't be null inside `options`, right?

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    /// <summary>
+    /// Options object to be passed to <see cref="WalletTransactionHandler"/>.
+    /// </summary>
+    public class TransactionBuildOptions
+    {
+        /// <summary>
+        /// Wallet account to use to build transaction. This is the account from which UTXOs will be selected. Required.
+        /// </summary>
+        public WalletAccountReference WalletAccountReference { get; set; }
```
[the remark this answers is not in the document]
codingupastorm: It could be - I'd rather either have the whole thing immutable or none of it. 

There could feasibly be situations where you want to build up an options object then try different wallet references afterwards. 

Maybe there is a specific BuildOptions setting that you keep because it makes a lot of sense for your individual application, and you wish to change only the wallet used.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -247,11 +254,24 @@ protected void AddSecrets(TransactionBuildContext context)
         }
 
         /// <summary>
-        /// Find the next available change address.
+        /// If a ChangeAddress was set in the options send change there, otherwise find the next available change address.
         /// </summary>
         /// <param name="context">The context associated with the current transaction being built.</param>
         protected void FindChangeAddress(TransactionBuildContext context)
         {
+            if (context.BuildOptions.ChangeAddress != null)
```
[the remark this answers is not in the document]
majikandy: The above is now redundant. 👍

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -130,6 +116,31 @@ public void FundTransaction(TransactionBuildContext context, Transaction transac
...
         }
 
+        public Money EstimateFee(TransactionBuildOptions options)
+        {
+            var context = new TransactionBuildContext(this.Network, options);
+            this.InitializeTransactionBuilder(context);
+            return context.TransactionFee;
+        }
+
+        private Transaction BuildTransaction(TransactionBuildContext context)
```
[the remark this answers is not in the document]
codingupastorm: This private method is called by both `FundTransaction` and `BuildTransaction`. I'd really like to get rid of this, but I think that's a refactor for a separate PR in the future.

For now will add a comment justifying this at least.

codingupastorm:
Sup @majikandy @bokobza I feel a bit like I'm not getting across what this is meant to achieve. This is obviously my fault, so I did just push an amendment to WalletTransactionHandler which hopefully helps a little, and I'll explain here:

Before this, BuildTransaction was very impure - you input a Context object, on which some of the properties you are able to use as options, and some you cannot use as options. After BuildTransaction, you will receive your Transaction object as requested, but through a bunch of side effects the Context is now in a completely different state to how it was input. 

If I want to create many transactions over and over, I have to create a brand new context every time. If I was to use the same one, it would break, or yield a completely different result (issues with change addresses, assign a different fee, etc etc.)

Given that this is one of the most likely tools to become 'public facing' this is not really okay, it's highly confusing and painful to consume. (It's also a tool that SC is going to be using a lot, there are many ways we may want to create a transaction).

This PR is the start of a pure approach. Options IN -> TransactionResult OUT.

Thanks for your comments so far, it is ready for review again.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -58,16 +58,39 @@ public class WalletTransactionHandler : IWalletTransactionHandler
             this.TransactionPolicy = transactionPolicy;
         }
 
+        /// <inheritdoc />
         public Transaction BuildTransaction(TransactionBuildOptions options)
         {
             var transactionBuildContext = new TransactionBuildContext(this.Network, options);
-            return BuildTransaction(transactionBuildContext);
+            return BuildTransactionInternal(transactionBuildContext);
```
majikandy: why the suffix of Internal?

src/Stratis.Bitcoin.Tests.Wallet.Common/WalletTestsHelpers.cs
```diff
@@ -113,6 +113,18 @@ public static (ChainedHeader ChainedHeader, Block Block) AppendBlock(Network net
             return (last, block);
         }
 
+        public static TransactionBuildOptions CreateOptions(WalletAccountReference accountReference, string password,
```
majikandy: I don't like the idea of dumping another method in this already massively bloated "Helper" class.

majikandy:
> Options IN -> TransactionResult OUT.

Sounds like a factory ?

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// </summary>
+        private void InitializeDefaults()
+        {
+            this.ChangeAddress = null;
+            this.FeeType = FeeType.Medium;
+            this.MinConfirmations = 1;
+            this.OpReturnData = null;
+            this.OverrideFeeRate = null;
+            this.SelectedInputs = new List<OutPoint>();
+            this.ShuffleOutputs = false;
```
zeptin: We seem to be defaulting this to `true` in `WalletController.BuildTransaction`. Is it defaulted to `false` for any particular reason here? Note that shuffling by default is generally preferable as it makes finding related wallet addresses more difficult when performing chain analysis.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -247,16 +280,37 @@ protected void AddSecrets(TransactionBuildContext context)
...
-        /// Find the next available change address.
+        /// If a ChangeAddress was set in the options, check it is from the wallet given and use that for the transaction.
+        /// Otherwise find the next available change address and use that for the transaction.
         /// </summary>
         /// <param name="context">The context associated with the current transaction being built.</param>
         protected void FindChangeAddress(TransactionBuildContext context)
         {
-            // Get an address to send the change to.
-            context.ChangeAddress = this.walletManager.GetUnusedChangeAddress(new WalletAccountReference(context.AccountReference.WalletName, context.AccountReference.AccountName));
+            if (context.BuildOptions.ChangeAddress != null)
```
bokobza: There is a change of functionality here.
Can we maybe add this in another PR?
I'm already not confident I can review this properly.

bokobza:
@codingupastorm this PR looks ok but it's really messing with my head in terms of me being confident I've taken everything into account.
If you want I can help you recreate it by splitting it in 2-3 PRs that you can review.
I can do this this evening UK time.
What do you think?

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -98,9 +107,17 @@ public void FundTransaction(TransactionBuildContext context, Transaction transac
...
+        /// a transaction that was built via <see cref="BuildTransactionInternal(TransactionBuildContext)"/>.
+        /// </summary>
+        /// <param name="transaction"></param>
+        /// <param name="newTransaction"></param>
+        /// <param name="context"></param>
+        private void UpdateTransactionFromBuiltTransaction(Transaction transaction, Transaction newTransaction, TransactionBuildContext context)
+        {
             if (context.ChangeAddress != null)
             {
                 // find the position of the change and move it over.
```
zeptin: If I remember correctly, this movement of the change address is only done so that the masternode registration protocol knows which address is the change. It essentially defeats the point of having shuffled outputs in the first place. The change address position could be an enum instead (First, Last, Random)? Or just an int for position, but I'm not sure if that's worth the possibility of error in the caller.

src/Stratis.Bitcoin.Features.Wallet/TransactionBuildOptions.cs
codingupastorm: It also defaulted to false in the previous implementation, with it being set true in the controller like you mentioned. 

At the moment this PR is just to recreate the same functionality, but I do agree that the defaults should be re-evaluated in the future

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -55,32 +53,43 @@ public class WalletTransactionHandler : IWalletTransactionHandler
...
 
-            if (context.Shuffle)
-                context.TransactionBuilder.Shuffle();
+        /// <inheritdoc />
+        public void FundTransaction(TransactionBuildOptions options, Transaction transaction)
+        {
+            var context = new TransactionBuildContext(this.Network, options);
 
-            context.Transaction = context.TransactionBuilder.BuildTransaction(context.Sign);
+            TransferTransactionDetailsToContext(transaction, context);
```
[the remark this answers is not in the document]
codingupastorm: I don't think methods need to be broken down only when they're used multiple times?

It's broken up logically and for readability - though they certainly could be used again. I imagine that we may want to add extra functionality to the building of transactions.

codingupastorm:
Yo I'm going to leave this with @bokobza to break down or implement differently, or even approve is he has a change of heart :)

I'm not pedantic about the implementation, I hope at least we can all agree that the current WalletTransactionHandler is not safe for public consumption yet

PR #1778, opened 2018-07-31
Fix default MinRelayTxFee

### LINKED ISSUE #1363: WalletTransactionHandler TransactionBuilder does not use network MinRelayTxFee
StandardTransactionPolicy has a hardcoded MinRelayTxFee of 5000 Satoshis. Mainnet has a MinRelayTxFee of 1000 Satoshis.

This affects these usages:
* TransactionBuilder.Verify (uses StandardTransactionPolicy.Check)
* TransactionBuilder.GetDust
* TransactionBuilder.Send (uses TransactionBuilder.GetDust)
* WalletTransactionHandler (uses TransactionBuilder.Verify and TransactionBuilder.Send)

Result: 
* Dust threshold is higher for transactions created using WalletTransactionHandler
* Transactions can fail verification in WalletTransactionHandler even though they might be using a fee rate higher than the network's fee rate

bokobza:
Closes #1363 

* Changed the MinRelayTxFee in StandardPolicy to be that set by the Network.
Previously, it was hardcoded to 5000 which was wrong for both Bitcoin And Stratis.
* The tests have also been updated and the accounting is correct.
* Some extra formatting

✔️ All tests pass but `A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect` whcih is supposed to fail.

image

✔️ The 4 networks are syncing nicely.


bokobza:
If they're below the relay fee, it won't get relayed, right? 
So if an attacker sends such transactions, the connected nodes receiving will discard them without relaying and there is little chance that it will block the network.
What do you think?

noescape00:
>If they're below the relay fee, it won't get relayed, right?

Correct

>So if an attacker sends such transactions, the connected nodes receiving will discard them without relaying and there is little chance that it will block the network.
What do you think?

That's correct. With how we do it now- it's all cool. 

But before we used to have relayFee < minFee. That came from bitcoin where there are some miners that are willing to mine non-standard txes. In our case it will just populate the mempool. With no miners who accept non-standard txes those txes will always hang in the mempool.

codingupastorm:
Looks good to me, no comments.

Verified that Bitcoin Core uses 1000 and StratisX 10_000 for MinRelayTxFee

https://github.com/bitcoin/bitcoin/blob/ad51e1372bc645fc8710da2bc9e0a9fc6c0f27bb/src/validation.h

https://github.com/stratisproject/stratisX/blob/master/src/main.h

Running tests now.

codingupastorm:
All tests pass but the `A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect` and the 2 that have been breaking in ApiSpecification...

bokobza:
Thanks @noescape00 and @codingupastorm !

Issue #1423, opened 2018-05-18
StratisDnsD Seeding Banned Peers


StratisIain:
Banned peers are written to the Masterfile.JSON resulting in banned peers continuing to be distributed to new nodes on the network.

dangershony:
This task should also fix  the issue of banning of white-listed nodes (a node that is about to be banned if its white-listed it should not be banned)

ferdeen:
Did a quick review of the code around this.  

It looks like the `DnsFeature `(which deals with seednodes and the whitelist), needs access to peer banning functionality, which is carried out in `Stratis.Bitcoin.Connection`.

PR #1798, opened 2018-08-01
Added coverage to ban and disconnect peer

majikandy:
Added coverage over existing method that was missing the coverage

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
@@ -29,11 +29,11 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests
             MinerScriptPubKey = new Key().ScriptPubKey;
         }
 
-        private async Task<(TestChainContext context, IPEndPoint peerEndPoint)> InitialiseContextAndPeerEndpointAsync()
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -66,6 +66,30 @@ private async Task<(TestChainContext context, IPEndPoint peerEndPoint)> Initiali
             Assert.True(context.PeerBanning.IsBanned(peerEndPoint));
         }
 
+        [Fact]
+        public async Task NodeIsSynced_PeerSendsAMutatedBlockAndPeerDisconnected_AndAddressIsNull_ThePeerGetsBanned_Async()
+        {
+            await this.NodeIsSynced_PeerSendsABadBlockAndPeerDisconnectedAndAddressIsNull_ThePeerGetsBanned_Async(
```
codingupastorm: This looks weird now but I'm guessing you may have more tests in future using the same method?

fassadlr:
@majikandy does this PR relate to #1423 ?

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
majikandy: Yeah it does, just copied the way another test was done in the file even though not varying the way the blocks were bad for this one, it was easier

majikandy:
Yes @fassadlr but doesn't close it

PR #1801, opened 2018-08-01
Wallet Sync Speedup

### LINKED ISSUE #1103: Wallet sync speedup

ferdeen:
Fixes #1103 

The change to allow the Wallet to move forward while syncing is based on the[ Producer Consumer DataFlow pattern](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-implement-a-producer-consumer-dataflow-pattern).  Which has been implemented locally in the `WalletSyncManager` class.

Tests impacted : 

image

Syncing against all for Daemons (Bitcoin, Stratis - MainNet and TestNet).

image

The syncing is currently running and I will be keeping an eye on any issues, but thought it would be best to get the PR out there for any comments on the solution - and hit the sprint deadline.

Cheers,
Ferdeen


monsieurleberre:
Starting the review now :)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreCache.cs
```diff
@@ -79,6 +88,44 @@ public virtual BlockStoreCachePerformanceCounter BlockStoreCachePerformanceCount
             return block;
         }
 
+        /// <inheritdoc />
+        public async Task<List<Block>> GetBlockAsync(List<uint256> blockids)
```
monsieurleberre: GetBlock*s*Async ?

ferdeen: done

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,158 +119,216 @@ public void Stop()
...
-                    this.walletTip = fork;
+                        this.walletManager.RemoveBlocks(fork);
+                        this.walletTip = fork;
 
-                    this.logger.LogTrace("Wallet tip set to '{0}'.", this.walletTip);
+                        this.logger.LogTrace("Wallet tip set to '{0}'.", this.walletTip);
+                    }
+
+                    // The new tip can be ahead or behind the wallet.
+                    // If the new tip is ahead we try to bring the wallet up to the new tip.
```
monsieurleberre: Was the change form > to < intended ?

ferdeen: changed to <=

ferdeen:
@noescape00 the current blockstorecache is working fine - the wallet height is advancing as normal.  

I am not too sure why you need to change it now.

When I made the change locally `IBlockStore.GetBlockAsync()` is not implemented - so I have undone the change.

https://github.com/stratisproject/StratisBitcoinFullNode/blob/16f602fb1c08082ae2d86ee6b2c7c8c79dbbf979/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs#L170-L174

Also this method is used in ConsensusManager - which if not implemented, doesn't seem right.

https://github.com/stratisproject/StratisBitcoinFullNode/blob/a238bc18bf97bfab93b8e936cd92dc8ed55d53f2/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs#L1043

noescape00:
>@noescape00 the current blockstorecache is working fine - the wallet height is advancing as normal.
>I am not too sure why you need to change it now.

It does. However after the activation we will have only 2 block providers- ConsensusManager and BlockStoreQueue. CHT will do the caching of the last 100 connected blocks. 

So any change to BlockStoreCache will become obsolete after activation is merged.

ferdeen:
@noescape00 Happy to do this.  However this is turning into something different.  Maybe we need a new card for this related to the activation work ?

@bokobza what do you think ?

ferdeen:
Had a chat with @noescape00 on Slack and the change is a one liner -

 `return this.blockRepository.GetAsync(blockHash);`

However, after adding this after a cycle the wallet doesn't advance.  I'll check the logging to see why this is the case.

Cheers

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: Hi @monsieurleberre,

When I test wallet syncing the logic changed from > to < by mistake.  

I have reverted the change and testing to make sure the wallet sync's in the correct manner.

Cheers,

Ferdeen

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -852,7 +852,7 @@ public void ProcessBlock(Block block, ChainedHeader chainedHeader)
                 if (chainedHeader.Height > current.Height)
                 {
                     this.logger.LogTrace("(-)[BLOCK_TOO_FAR]");
-                    throw new WalletException("block too far in the future has arrived to the wallet");
+                    return;
```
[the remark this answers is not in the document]
ferdeen: ok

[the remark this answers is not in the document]
ferdeen: Carry on as we now store any blocks that come in from consensus in the local concurrent queue.

ferdeen:
New PR (#1869) created that is based on today's master.

PR #1806, opened 2018-08-02
Move DifficultyAdjustmentInterval calculation

rowandh:
This PR moves the calculation of `DifficultyAdjustmentInterval` closer to its usage. `DifficultyAdjustmentInterval` is only used in one method of `ChainedHeader`.

Moving this also makes it possible for us to use `IConsensus` in place of `Consensus`. I will do this as part of a later PR.

### DIFF
--- a/src/NBitcoin/ChainedHeader.cs
+++ b/src/NBitcoin/ChainedHeader.cs
@@ -431,7 +431,7 @@ namespace NBitcoin
         /// Calculate the difficulty adjustment interval in blocks based on settings defined in <see cref="IConsensus"/>.
         /// </summary>
         /// <returns>The difficulty adjustment interval in blocks.</returns>
-        public long GetDifficultyAdjustmentInterval(IConsensus consensus)
+        private long GetDifficultyAdjustmentInterval(IConsensus consensus)
         {
             return (long)consensus.PowTargetTimespan.TotalSeconds / (long)consensus.PowTargetSpacing.TotalSeconds;
         }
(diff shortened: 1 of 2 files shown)


rowandh:
Chains sync, integration tests passing except for the usual suspects

Aprogiena:
The change itself looks well done, I am undecided about the motivation of this change, but not familiar with the code or original decisions there enough to raise any objections

dangershony:
@rowandh this is a step in the right direction.
We need to think about this a bit further perhaps?
For example I think POS uses a different also to calculate the difficulty adjustment  (in another part of the code) perhaps think about a way to unify adjustment calculation in one place?

An idea I had is use a special abstract Rule for that (rules will be available soon in IConsensus) so you could potentially do

```
consensus.Rules<CalculateWorkRule>().GetNextWorkRequired(chainedHeader)

```

Have a think about that.

src/NBitcoin/ChainedHeader.cs
```diff
@@ -425,6 +427,15 @@ public Target GetWorkRequired(Consensus consensus)
             return finalTarget;
         }
 
+        /// <summary>
+        /// Calculate the difficulty adjustment interval in blocks based on settings defined in <see cref="IConsensus"/>.
+        /// </summary>
+        /// <returns>The difficulty adjustment interval in blocks.</returns>
+        public long GetDifficultyAdjustmentInterval(IConsensus consensus)
```
bokobza: This can even be private

rowandh:
@dangershony agree, would make sense to unify the logic in the same place. I am not familiar enough with the codebase to know where else it would be calculated. Maybe something for a future PR?

My main motivation for this was to move it out of Consensus so that I can begin to make the remaining properties immutable. It's a longish path and this was the first step. Given how it touches core logic I felt it needed its own PR.

PR #1815, opened 2018-08-02
StyleCop Analyzers

bokobza:
So here it what I've got:
- `StyleCopAnalyzers` as a nuget in `Stratis.Bitcoin`
  - I originally wasn't too keen on adding a nuget for that, but apparently, static analyzers are a thing from the past while Roslyn analyzers are the way forward and that's one everone is doing.
- 2 rulesets:
  - one called `Stratis.ruleset` which is where we have the rules
  - one called `None.ruleset` where the rules are all disabled.
- `stylecop.json`
  - some of the rules are configurable. This is done in `stylecop.json`.

`Stratis.Bitcoin` is referencing `Stratis.ruleset` and had been partially fixed.
All the other projects are referencing `None.ruleset` and therefore don't show any warnings.

The goal is too either change a project to use `Stratis.ruleset` and fix all the warnings there OR to enforce a rule in `None.ruleset` and fix it for the whole solution.
I'm in favour of the second solution as it's fewer clicks.

When all is done and clean we can remove the `None.ruleset`.


src/Stratis.Bitcoin/Base/Deployments/ThresholdConditionCache.cs
```diff
@@ -9,18 +9,18 @@ namespace Stratis.Bitcoin.Base.Deployments
     public class ThresholdConditionCache
     {
         /** What block version to use for new blocks (pre versionbits) */
-        public static int VersionbitsLastOldBlockVersion = 4;
+        private static int VersionbitsLastOldBlockVersion = 4;
```
noescape00: Looks like public methods that are not used now made private. Not sure if that's correct. 
Some PRs for example introduce some public fields and use them only in follow up PRs.

src/Stratis.Bitcoin/Controllers/FeatureController.cs
```diff
@@ -19,17 +19,17 @@ public ActionDescription(string description)
 
     public abstract class FeatureController : Controller
     {
-        protected IFullNode FullNode;
+        protected IFullNode FullNode { get; set; }
```
noescape00: Do we want conversion to fields by default? If so: why?

noescape00:
Excepts for points mentioned I really like it.

src/Stratis.Bitcoin/Base/Deployments/ThresholdConditionCache.cs
bokobza: yeah I made those private because they're not used outside, or not used.
I appreciate that these might be used in the future so I left them there, just made them private.

src/Stratis.Bitcoin/Controllers/FeatureController.cs
bokobza: This is converted to a property.
Maybe I should convert this back to a field but make them camelCase (I think this is the better way).

PR #1825, opened 2018-08-02
Check if peer is banned and when refreshing masterfile

majikandy:
Fixes: #1423 

Don't include the node in masterlist when it is banned

Put the banning check in the refresh loop - which happens every 30 seconds.

Therefore there is upto a 30 second delay where a banned node would still be returned.

[And demonstration of NSubstitute in one test for @dangershony to show syntax difference to Moq]


majikandy:
I have tested this on.... 51.141.83.38   @StratisIain @dangershony  

ban_in_peers_json

whitelist_in_masterfile_json

not_returned_as_a_record

That server is currently running this PR branch.

src/Stratis.Bitcoin.Features.Dns/IFullNodeBuilderExtensions.cs
```diff
@@ -30,6 +31,7 @@ public static IFullNodeBuilder UseDns(this IFullNodeBuilder fullNodeBuilder)
                     services.AddSingleton<DnsSettings>();
                     services.AddSingleton<IUdpClient, DnsSeedUdpClient>();
                     services.AddSingleton<IWhitelistManager, WhitelistManager>();
+                    services.AddSingleton<IPeerBanning, PeerBanning>();
```
dangershony: You don't need to do this peer banning is available by default in base

majikandy: ok cool, I couldn't see it there

PR #1829, opened 2018-08-02
[Activation] Shutdown sequence and bugfixes

noescape00:
shutdown sequence fix


dangershony:
Is this ready for review?

noescape00:
Almost. 
Will need around 30 minutes more. Still didn't test it properly and didn't check everything we've discussed on a call

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -133,7 +134,8 @@ public sealed class BaseFeature : FullNodeFeature, INodeStats
             IConsensusManager consensusManager,
             IConsensusRuleEngine consensusRules,
             IPartialValidator partialValidator,
-            IBlockStore blockStore = null)
+            IBlockPuller blockPuller,
+            IBlockStore blockStore)
```
dangershony: Ivan please create an issue to move block store to Stratis.Bitcoin if this is going to become a mandatory component and we would just implement pruning instead of not using a feature we need to track this.

dangershony: @bokobza FYI

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -74,8 +75,7 @@ public interface IBlockPuller
         void ShowStats(StringBuilder statsBuilder);
     }
 
-    /// <inheritdoc cref="IBlockPuller"/>
```
dangershony: I noticed that ;)

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -82,6 +82,8 @@ public IReadOnlyNetworkPeerCollection ConnectedPeers
 
         private readonly IVersionProvider versionProvider;
 
+        private IConsensusManager consensusManager;
```
dangershony: So this is how we solve the dependency fine with me.

src/Stratis.Bitcoin/Base/BaseFeature.cs
noescape00: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1845

done

PR #1840, opened 2018-08-03
Add guidelines around rebasing PRs with conflicts

rowandh:
Proposal to favour rebasing vs. merging from parent branch when PRs become stale.

Reasons:
* Keeps unrelated changes out of PRs, making them easier to review
* History is cleaner/easier to reason about

Guideline taken from [corefx](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/contributing.md). Bitcoin also do a [similar thing](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md).

### DIFF
--- a/Documentation/contributing.md
+++ b/Documentation/contributing.md
@@ -70,6 +70,7 @@ DOs and DON'Ts
 * **DO** keep the discussions focused. When a new or related topic comes up
   it's often better to create new issue than to side track the discussion.
 * **DO** blog and tweet (or whatever) about your contributions, frequently!
+* **DO** favour fixing merge conflicts using `git rebase` over `git merge`.
 
 * **DO NOT** send PRs for style changes. 
 * **DON'T** surprise us with big pull requests. Instead, file an issue and start
@@ -77,7 +78,6 @@ DOs and DON'Ts
   of time.
 * **DON'T** commit code that you didn't write. If you find code that you think is a good fit to add to .NET Core, file an issue and start a discussion before proceeding.
 * **DON'T** submit PRs that alter licensing related files or headers. If you believe there's a problem with them, file an issue and we'll be happy to discuss it.
-* **DO NOT** fix merge conflicts using a merge commit. Prefer `git rebase`.
 
 Copyright
 =========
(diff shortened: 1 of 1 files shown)


Documentation/contributing.md
```diff
@@ -77,6 +77,7 @@ DOs and DON'Ts
   of time.
 * **DON'T** commit code that you didn't write. If you find code that you think is a good fit to add to .NET Core, file an issue and start a discussion before proceeding.
 * **DON'T** submit PRs that alter licensing related files or headers. If you believe there's a problem with them, file an issue and we'll be happy to discuss it.
+* **DO NOT** fix merge conflicts using a merge commit. Prefer `git rebase`.
```
dangershony: I am not sure about this @rowandh we attempted using rebase and always got problems, to approve this I would need you to do a small video meeting and demonstrate what you mean exactly.

rowandh: Happy to do that

majikandy: I found rebase worked well when on a project where we used branches on the main repo (ie all had commit rights) - When doing it via forks here on this project, I got myself into a mess a couple of times.  However, if you can show it is better and how to manage it, would be great to see a demo.

rowandh: Let's schedule a call when everyone's back and I'll run through how I've been doing this already.

As an example, on [this PR](https://github.com/stratisproject/StratisBitcoinFullNode/pull/1742) I had to rebase to master 3 times because it was in review for so long. But if you look at the history, you see no "Merge branch 'master' into ..." commits, and the diff maintains the original changeset.

noescape00:
https://medium.com/@fredrikmorken/why-you-should-stop-using-git-rebase-5552bee4fed1

bokobza:
> Also pretty much no one creates PRs in a fashion that makes it convenient to review commit by commit.

but it should be the aim, so that's it's easier for reviewers to review.

However, I don't think this is something we should enforce.
My advice would be for the devs to do "whatever works".
@rowandh I'm still keen to see how you use `git rebase`. I usually stay away from it (maybe because although I'm confident with my git fu, rebase has always eluded me).

rowandh:
@noescape00 that article makes some good points. I think they might be more relevant in the instance of merging large feature branches that many developers have worked on. I don't think they are very important in the context of rebasing smaller PRs. Reasons:
* The need to rebase a PR should not happen very often if we do not allow PRs to go stale
* If a PR is not easily readable (due to a 'merge master' introducing irrelevant changes) then a reviewer is more likely to miss something that may introduce a bug
* Merges can also resolve conflicts in a way that introduces bugs, which are be obscured in a 'merge' commit that are almost impossible to review correctly
* We squash-merge all PRs into master anyway, so tracing back the source of an introduced bug in is no different to now

bokobza:
I'm in favour of using `git rebase` where possible as it's cleaner, but I wouldn't want to enforce it. I had a chat with @rowandh and we agreed to recommend it without enforcing it.

Issue #1747, opened 2018-07-26
Add ActionFilters to web api calls to show or hide calls by network.


ferdeen:
In relation to  #1691. 

We were trying to stop the Staking type web api calls to be executed when running on the Bitcoin network.

Look into whether it's possible to do this via Actions filters.

So far, trying out the `ActionFilterAttribute`.  

Problem we are facing is trying to pass `this.fullNode.Network...` as a parameter to the attribute.   We know that primitive types can only be used with attributes, however we need a way to pass the current network into the attribute.

![image](https://user-images.githubusercontent.com/16000683/43204074-e92ad00c-9017-11e8-9409-40539a438181.png)

While working on #1691 we realised that all API controllers are discoverable in Swagger.   

So starting StratisD or BitcoinD you will always see all the API controllers (Staking and Mining) if at least one API controller is included within a feature setup.  We will need to tackle this at a later date and focus on sprint tasks.

@TjadenFroyda 

> MvcBuilderExtensions.cs from Features.Api is used to automatically find all of the api endpoints.
> 
> Could the network specific mining/staking endpoints be filtered there?

TjadenFroyda:
Replying on this issue vs the closed PR. Happy to give it a try.

Thinking out loud - this might work and will submit PR if successful:
I think I'll be able to get the network inside AddControllers with:

var network = services.GetService<Network>();

And then add filtering logic from there on line 23 with network.Consensus.IsProofOfStake (as opposed to hardcoding network names) to remove the respective controllers.

TjadenFroyda:
The above didn't work, but found something that did (see below). Wrote some integration tests to prove that calling staking methods from a pow node (and vice versa) fails appropriately. Will submit PR when done, waiting for ITs to finish running. 

Proof of Stake Consensus: Mining methods removed. (staking still present)
![image](https://user-images.githubusercontent.com/38709281/43561136-2ead4cc2-95db-11e8-8c9c-29aed9e0bfee.png)


Proof of Work Consensus: Staking methods removed (mining still present)
![image](https://user-images.githubusercontent.com/38709281/43561079-fc09888a-95da-11e8-88df-416a3692c099.png)

bokobza:
Solution implemented in https://github.com/stratisproject/StratisBitcoinFullNode/pull/1842

PR #1842, opened 2018-08-03
Hide staking API when POW and mining API when POS

TjadenFroyda:
Fix for #1747 by implementing swagger documentfilters. Please see screenshots attached to that issue for demonstration. 
Also includes:
- Staking and Mining API methods now throw bad request (400) exception when call is attempted from wrong consensus context. 
- Updated unit and integration tests, including methods demonstrating the above. 

1914 tests passing, 18 failing locally - all unrelated and all "System.UnauthorizedAccessException" 
@ferdeen @fassadlr


src/Stratis.Bitcoin.Features.Api/ProofOfStakeConsensusApiFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using Swashbuckle.AspNetCore.Swagger;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace Stratis.Bitcoin.Features.Api
+{
+    /// <summary>
+    /// Swagger Documentation filter for removing mining related methods from API documentation.
+    /// </summary>
+    class ProofOfStakeConsensusApiFilter : IDocumentFilter
```
fassadlr: Missing `public`.

src/Stratis.Bitcoin.Features.Api/ProofOfWorkConsensusApiFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using Swashbuckle.AspNetCore.Swagger;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace Stratis.Bitcoin.Features.Api
+{
+    /// <summary>
+    /// Swagger Documentation filter for removing staking related methods from API documentation.
+    /// </summary>
+    class ProofOfWorkConsensusApiFilter : IDocumentFilter
```
fassadlr: Missing `public`

src/Stratis.Bitcoin.Features.Api/ProofOfStakeConsensusApiFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public void Apply(SwaggerDocument swaggerDoc, DocumentFilterContext context)
+        {
+            var pathsToRemove = swaggerDoc.Paths
+                .Where(pathItem => pathItem.Key.Contains("/api/Mining/"))
```
fassadlr: It would be cool if we can somehow filter by the actual controller name instead of magic strings?

`.Where(pathItem => pathItem.Key.Contains(string.Format("/api/{0}/", typeof(MiningController).Name))`

You will need to somehow filter "controller" out of the name? Maybe `swaggerDoc.Paths` has the controller name it?

src/Stratis.Bitcoin.Features.Api/ProofOfWorkConsensusApiFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public void Apply(SwaggerDocument swaggerDoc, DocumentFilterContext context)
+        {
+            var pathsToRemove = swaggerDoc.Paths
+                .Where(pathItem => pathItem.Key.Contains("/api/Staking/"))
```
fassadlr: See comment above re filtering.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -28,21 +28,27 @@ public class MiningController : Controller
         /// <summary>Wallet manager.</summary>
         private readonly IWalletManager walletManager;
 
+        /// <summary>Full Node.</summary>
+        private readonly IFullNode fullNode;
```
fassadlr: Rather just inject `Network` ;)

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
```diff
@@ -177,5 +177,29 @@ public void Status_returns_status_info()
...
+        {
+            Given(a_proof_of_stake_node_with_api_enabled);
+            When(calling_getstakinginfo);
+            Then(staking_information_is_returned);
+        }
+
+        [Fact]
+        public void POW_node_calls_getstakinginfo_and_receives_error()
+        {
+            Given(a_pow_node_with_api_enabled);
```
majikandy: to be consistent with the other Givens,  this should be `a_proof_of_work_node....`  or the  `a_proof_of_stake_node...`  should be `a_pos_node...`    I favour the long form as it is easier for anyone new to the code.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
```diff
@@ -177,5 +177,29 @@ public void Status_returns_status_info()
             When(calling_status);
             Then(status_information_is_returned);
         }
+
```
majikandy: At the high level, these new tests read well and express intent very well. 👍

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -49,7 +49,6 @@ public partial class ApiSpecification : BddSpecification
 
         private HttpResponseMessage response;
         private string responseText;
-        private HttpResponseMessage postResponse;
```
majikandy: 👍

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -322,6 +321,17 @@ private void calling_getrawtransaction_verbose()
             this.send_api_get_request($"api/Node/getrawtransaction?trxid={this.transaction.GetHash().ToString()}&verbose=true");
         }
 
+        private void calling_getstakinginfo()
+        {
+            this.send_api_get_request("api/Staking/getstakinginfo");
```
majikandy: I wonder if we should make these a list of constants at the top of the file?

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -464,11 +474,25 @@ private void the_txout_is_returned()
             txOutResponse.Value.Should().Be(this.transferAmount);
         }
 
+        private void staking_information_is_returned()
```
majikandy: we need another test at some point for when staking is enabled and there are actual coins at stake (the test somewhere near the top of the Specification does startstaking but without any coins at stake) - Perhaps create an Issue to add the test if you agree ?

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -464,11 +474,25 @@ private void the_txout_is_returned()
...
+        {
+            var stakingInfoModel = JsonDataSerializer.Instance.Deserialize<GetStakingInfoModel>(this.responseText);
+            stakingInfoModel.Enabled.Should().Be(false);
+            stakingInfoModel.Staking.Should().Be(false);
+        }
+
+        private void a_bad_request_error_is_returned()
+        {
+            this.response.StatusCode.Should().Be(StatusCodes.Status400BadRequest);
```
majikandy: I think 405 might be more appropriate....

```
405 Method Not Allowed
A request method is not supported for the requested resource; for example, a GET request on a form that requires data to be presented via POST, or a PUT request on a read-only resource.
```

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
TjadenFroyda: Good idea. Done.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
TjadenFroyda: Haven't addressed this with this PR, but could consider in a future one.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
TjadenFroyda: Agree, updated.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
TjadenFroyda: Sounds good, will update accordingly.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
TjadenFroyda: so much easier that way ;)

TjadenFroyda:
@bokobza 
Re 1 - [Looks doable](https://github.com/domaindrivendev/Swashbuckle/issues/153#issuecomment-213342771), will make change accordingly. Plan on naming `[HideWhenProofOfStake]` and `[HideWhenProofOfWork]`. 
Re 2 - I initially tried adding logic to MvcBuilderExtensions to exclude specific controllers based on network (modified the for loop to exclude MiningController or StakingController depending on network), but that was ineffective at keeping those methods from appearing in the swagger documentation. The only way I was able to do it was by checking the network consensus type in startup.cs and then applying a custom IDocumentFilter.

TjadenFroyda:
Ahh, I think I found part of the problem with item 2 - AddControllersAsServices was slipping the excluded methods back into the api even after being excluded

TjadenFroyda:
So moving the logic to the MvcBuilderExtensions didn't work to remove it from swagger. Might be forced to use the logic in the api. Thoughts or other ideas?

src/Stratis.Bitcoin.Features.Api/ProofOfStakeConsensusApiFilter.cs
TjadenFroyda: I tried to do that here, but couldn't get it to work without pulling in a reference to the Mining Feature.

noescape00:
@TjadenFroyda really good job, ty!

Also please fix merge conflicts

TjadenFroyda:
@noescape00 Thanks! Merge conflicts fixed. 
Randomly applied one of the attributes to some of the other controllers and hiding the swagger docs worked as expected.

noescape00:
@majikandy @bokobza please check if requested changes were applied and approve\request more.

src/Stratis.Bitcoin/Controllers/ProofOfStakeAttribute.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Stratis.Bitcoin.Controllers
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    public class ProofOfStakeAttribute : Attribute
```
bokobza: Can you please add some documentation here to show where this and ProofOfWorkAttribute are intended to be used?

bokobza:
tbh I would have happily just returned a 405 if a non-allowed method was called.
There is quite a bit of plumbing necessary in order to removed the not allowed method from Swagger.
My main issue though is that the API now had some knowledge about pow Vs pos, where it really shouldn't.

The ultimate goal would be to have a custom attribute on the controller class (like `HideIfNotApplicable` or `ShowIfApplicable`) + a document filter that filters out all the classes that are not applicable. Then in the custom attribute code, you can check whether we're in pow or pos and what to do.

@TjadenFroyda Please check if this is possible, otherwise we'll see what to do, but I really don't want to introduce any logic to the Api project.

Another way to do it is to leave the methods there and add a comment so that in Swagger it'll show "This endpoint is not relevant for pow networks" or something like that.

TjadenFroyda:
@bokobza I wasn't able to figure out a way to dynamically apply a document filter to exclude specific controller endings other than what I have here. Ran into issues passing the network into the attribute at runtime. For this PR, I could delete the attribute classes, documentfilter classes, and swagger logic used to hide those paths from the documentation.  The 405's for not allowed methods are now built into each method call and would remain. The error message added would indicate that the method doesn't apply for that type of network as suggested. Should I make these changes?

bokobza:
@TjadenFroyda ok thanks you for doing that. 
We can look at this again in the future.

ferdeen:
@TjadenFroyda apololgies for not getting involved in this - must have missed the alert when you included my name at the start of this PR.

Good work!

PR #1843, opened 2018-08-03
Use same Network in NodeGroupBuilder when creating nodes

rowandh:
`NodeGroupBuilder` builds a group of nodes to presumably be used together, so it makes sense to use the same network for all the nodes it builds.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
@@ -86,7 +86,7 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
 
         public CoreNode CreateStratisPowNode(bool start = false)
         {
-            return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName()), start);
+            return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName(), KnownNetworks.RegTest), start);
         }
 
         public CoreNode CreateStratisPowNode(Network network, bool start = false)
@@ -136,7 +136,7 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
 
         public CoreNode CloneStratisNode(CoreNode cloneNode)
         {
-            var node = new CoreNode(new StratisBitcoinPowRunner(cloneNode.FullNode.Settings.DataFolder.RootPath), this, "bitcoin.conf");
+            var node = new CoreNode(new StratisBitcoinPowRunner(cloneNode.FullNode.Settings.DataFolder.RootPath, cloneNode.FullNode.Network), this, "bitcoin.conf");
             this.Nodes.Add(node);
             this.Nodes.Remove(cloneNode);
             return node;
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
@@ -84,10 +84,10 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
             return CreateNode(new BitcoinCoreRunner(this.GetNextDataFolderName(), bitcoinDPath), start: false, useCookieAuth: useCookieAuth);
         }
 
-        public CoreNode CreateStratisPowNode(bool start = false)
-        {
-            return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName(), KnownNetworks.RegTest), start);
-        }
+        //public CoreNode CreateStratisPowNode(Network network, bool start = false)
+        //{
+        //    return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName(), KnownNetworks.RegTest), start);
+        //}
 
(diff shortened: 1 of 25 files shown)


fassadlr:
@rowandh not sure if this change is required as we set the network in the constructor in each runner? The runner is pre-configured with a specific network. I.e. `StratisBitcoinPosRunner` will never be configured with `BitcoinMain`

rowandh:
@fassadlr assigning the network in the constructor is what I want to avoid. It makes it impossible to define tests that run using other networks with different parameters.

majikandy:
>NodeGroupBuilder builds a group of nodes to presumably be used together, so it makes sense to use the same network for all the nodes it builds.

It does make sense to have all the nodes on the same network.  So this was just a parameterising network?  Seems ok, why would you need something other than RegTest?  I'm guessing SmartContract has a different reg test or something ?

rowandh:
> Seems ok, why would you need something other than RegTest?
Some tests need a network with parameters that are different to the others, eg. coinbase maturity.

fassadlr:
@rowandh 

1> Is this change supposed to fix an integration test?
2>In order to keep things uniform... we either remove the parameter-less constructor in all the runners or we create a new runner which can take in a network. I prefer the latter.

Thoughts?

bokobza:
Also please let Lukasz know about changes to Integration tests as he's going to look at those in the next sprint.

rowandh:
@fassadlr 
1 - No
2 - The parameterless constructors in these runners are no longer used anyway. I'd rather just remove them.

fassadlr:
@rowandh ok yeah, lets remove all the parameter-less constructors 👍

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowRunner.cs
```diff
@@ -18,9 +19,15 @@ public StratisBitcoinPowRunner(string dataDir)
             this.Network = KnownNetworks.RegTest;
```
bokobza: Can `public StratisBitcoinPowRunner(string dataDir)` also be removed?

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPosRunner.cs
```diff
@@ -19,6 +20,12 @@ public StratisBitcoinPosRunner(string dataDir)
             this.Network = KnownNetworks.StratisRegTest;
```
bokobza: Can `public StratisBitcoinPosRunner(string dataDir)` also be removed?

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowRunner.cs
fassadlr: +1

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPosRunner.cs
fassadlr: +1

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowRunner.cs
rowandh: This has been done. The method that consumes it however has 51 usages, so I resorted to defaulting the network in that method - not ideal I know.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -86,10 +86,15 @@ public CoreNode CreateBitcoinCoreNode(string version = "0.13.1", bool useCookieA
 
         public CoreNode CreateStratisPowNode(bool start = false)
         {
-            return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName()), start);
+            return CreateNode(new StratisBitcoinPowRunner(this.GetNextDataFolderName(), KnownNetworks.RegTest), start);
```
fassadlr: @rowandh you know that beer you owe me? I'm calling in the favour ;)

rowandh: 🍺 Done

rowandh:
Integration tests passing except for:
* AddNewAccount_for_xpub_only_wallet_informs_user_to_create_new_wallet_
* CreateExtPubKeyOnlyWallet_creates_wallet_with_extra_flag
* A_cut_off_miner_advanced_ahead_of_network_causes_reorg_on_reconnect

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -84,12 +84,17 @@ public CoreNode CreateBitcoinCoreNode(string version = "0.13.1", bool useCookieA
             return CreateNode(new BitcoinCoreRunner(this.GetNextDataFolderName(), bitcoinDPath), start: false, useCookieAuth: useCookieAuth);
         }
 
-        public CoreNode CreateStratisPowNode(bool start = false)
+        //public CoreNode CreateStratisPowNode(Network network, bool start = false)
```
fassadlr: Just delete this green code ;)

rowandh:
Thanks @fassadlr!

PR #1669, opened 2018-07-17
[LightWallet] Fix for #1550

zeptin:
Ported Maciej's fix from BreezeHub SBFN branch

PR #1844, opened 2018-08-03
Wallet Recovery Fix

### LINKED ISSUE #1550: When restoring multiple wallets some transactions might be missing
**Steps**
- Start FullNode
- Restore WalletA (from block height 100)
- Restore WalletB (from block height 500)
- When the second wallet begins restoring the Wallet tip will jump to 500 hence some transactions might not be restored from WalletA

**Fix in LightWalletSyncManager.cs**
https://github.com/breezehub/StratisBitcoinFullNode/blob/breeze-beta/src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs

bokobza:
Fixes #1550 
Closes #1669 

The call for recovering a wallet would always trigger a sync, but recovering a wallet ahead of the current sync run would make the sync jump forward.

This was fixed by only triggering a sync if the current sync run needs to go back a number of blocks.

@maciejzaleski Can you please retest?
@zeptin

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
@@ -327,13 +327,13 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
                 Network = NetworkHelpers.GetNetwork("mainnet")
             };
 
-            var mockWalletWrapper = new Mock<IWalletManager>();
-            mockWalletWrapper.Setup(w => w.RecoverWallet(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime>(), null)).Returns(wallet);
+            var mockWalletManager = new Mock<IWalletManager>();
+            mockWalletManager.Setup(w => w.RecoverWallet(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime>(), null)).Returns(wallet);
 
             Mock<IWalletSyncManager> walletSyncManager = new Mock<IWalletSyncManager>();
             walletSyncManager.Setup(w => w.WalletTip).Returns(new ChainedHeader(this.Network.GetGenesis().Header, this.Network.GetGenesis().Header.GetHash(), 3));
 
-            var controller = new WalletController(this.LoggerFactory.Object, mockWalletWrapper.Object, new Mock<IWalletTransactionHandler>().Object, walletSyncManager.Object, It.IsAny<ConnectionManager>(), this.Network, this.chain, new Mock<IBroadcasterManager>().Object, DateTimeProvider.Default);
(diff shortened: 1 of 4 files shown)


majikandy:
Something similar needed in RecoverViaExtPubKey ?

bokobza:
Good point @majikandy, let me check.

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
```diff
@@ -346,6 +349,49 @@ public void RecoverWalletSuccessfullyReturnsWalletModel()
...
+            {
+                Name = "myWallet",
+                Network = NetworkHelpers.GetNetwork("mainnet")
+            };
+
+            // The chain is at height 100.
+            ConcurrentChain concurrentChain = WalletTestsHelpers.GenerateChainWithHeight(100, this.Network);
+            DateTime lastBlockDateTime = concurrentChain.Tip.Header.BlockTime.DateTime;
+
+            var mockWalletWrapper = new Mock<IWalletManager>();
```
majikandy: -> mockWalletManager

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
```diff
@@ -346,6 +349,49 @@ public void RecoverWalletSuccessfullyReturnsWalletModel()
...
+            mockWalletWrapper.Setup(w => w.RecoverWallet(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime>(), null)).Returns(wallet);
+
+            Mock<IWalletSyncManager> walletSyncManager = new Mock<IWalletSyncManager>();
```
majikandy: -> var

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletControllerTest.cs
```diff
@@ -346,6 +349,49 @@ public void RecoverWalletSuccessfullyReturnsWalletModel()
...
+            {
+                Name = "myWallet",
+                FolderPath = "",
+                Password = "",
+                Network = "MainNet",
+                Mnemonic = "mnemonic",
+                CreationDate = lastBlockDateTime
+            });
+
+            mockWalletWrapper.VerifyAll();
```
majikandy: Is this necessary? the test wouldn't work if the setups didn't work.

Also stacking up asserts in the arrange, to verify in the assert blurs the boundaries between arrange act and assert.   If an explicit Verify is needed (meaning there is no side effect to test other than knowing that method got called)  it can happen explicity in the Assert section (here)

bokobza: I guess calling VerifyAll is more for making sure everything has been verified.
We should look into making the difference between arrange and assert more distinct, but let's leave it for another time. I'd like to do it for all the tests at some point.

zeptin:
@bokobza I have tested your branch and it looks pretty solid. Shall I close 1669?

PR #1850, opened 2018-08-04
Fixes #1847 NodeSettings is initalized 3 times

### LINKED ISSUE #1847: NodeSettings is initialized 3time
Conditions to reproduce: 
clean build directory, clean databolder folder in appdata, open solution in VS and hit F5 (stratisD as default startup project). 
After that NodeSettings cpnstructor will be called 3 times

rowandh:
Looks like two unused variables were introduced that load NodeSettings defaults.

Fix #1847


bokobza:
I'm rebuilding AppVeyor

Issue #1396, opened 2018-05-08
Stabilize Integration Tests


bokobza:

dangershony:
This will also include find what tests can move out from the Integration tests talk to @majikandy @bokobza  or myself to discuss more.

monsieurleberre:
OK I'll give it a try

dangershony:
@maciejzaleski team will take over this task after TB release.

ArcSin2000X:
Running the suite of **intergration tests(158)** is providing better results when executed on bigger box(16CPU) when compared to lower spec. After running numerous test sessions I was able to isolate list of tests that consequently fail on big box test. Notably if you run most of the tests on the list as standalone test you are unlikely to find problems. The next step is to work through the list and fix tests  1 by 1
![image](https://user-images.githubusercontent.com/37782512/44218699-28f09800-a172-11e8-9a76-1e2aef2f0a02.png)

ArcSin2000X:
Zip with stack traces
[HTML_version_of_top_offenders.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2294238/HTML_version_of_top_offenders.zip)

dangershony:
Closing this as it will be traced by
https://github.com/stratisproject/StratisBitcoinFullNode/issues/2195

bokobza:
Closing this as we now have many issues opened to fix this.

PR #1860, opened 2018-08-06
Improvements toTestHelper.Wait loop, increased wait time on Segwit test

ArcSin2000X:
TestHelper.Wait loop is now accepting 3 additional arguments to control exceptions, sampling interval and cancellation token

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
@@ -5,18 +5,28 @@ using System.Threading;
 using Stratis.Bitcoin.Base;
 using Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers;
 using Stratis.Bitcoin.P2P.Peer;
+using Xunit;
 
 namespace Stratis.Bitcoin.IntegrationTests.Common
 {
     public class TestHelper
     {
-        public static void WaitLoop(Func<bool> act)
+        public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reason", int millisecondsTimeout = 50, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var cancel = new CancellationTokenSource(Debugger.IsAttached ? 15 * 60 * 1000 : 30 * 1000);
+            cancellationToken = cancellationToken == default(CancellationToken)
+                ? new CancellationTokenSource(Debugger.IsAttached ? 15 * 60 * 1000 : 30 * 1000).Token
+                : cancellationToken;
             while (!act())
             {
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.IntegrationTests/SegWitTests.cs
```diff
@@ -42,8 +43,8 @@ public void TestSegwit_MinedOnCore_ActivatedOn_StratisNode()
                 {
                     // generate 450 blocks, block 431 will be segwit activated.
                     coreRpc.Generate(450);
-
-                    TestHelper.WaitLoop(() => stratisNode.CreateRPCClient().GetBestBlockHash() == coreNode.CreateRPCClient().GetBestBlockHash());
+                    var cancellationToken = new CancellationTokenSource(TimeSpan.FromMinutes(2)).Token;
```
zeptin: Nice, I assume this explains why this test seems to fail randomly?

ArcSin2000X: yes, basically we have 2 different values for cancellation tokens, if you run it in debug it will get more time to check for the segwit block(with increased time you are going to pass), I think I will want to rewrite it not to relay on the wait loop to check if we got to block 450...

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -5,18 +5,28 @@
...
+                ? new CancellationTokenSource(Debugger.IsAttached ? 15 * 60 * 1000 : 30 * 1000).Token
+                : cancellationToken;
             while (!act())
             {
-                cancel.Token.ThrowIfCancellationRequested();
-                Thread.Sleep(50);
+                try
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
+                    Thread.Sleep(millisecondsTimeout);
```
zeptin: Can you elaborate a bit on what the benefit is of providing custom values for this?

ArcSin2000X: We gain ability to programatically control the sampling interval so that we can probe less or more frequently as well as we can increase the overall loop duration for processes that are likely to take longer then currently hardcoded 30 seconds. The thinking process behind this was that we want to maintain most checks to max 30[s] at 50[ms] interval, however we have ability to override defaults for edge cases. Other think to look into would be if the Thread being called with sleep method is appropriate here

bokobza:
Linking to https://github.com/stratisproject/StratisBitcoinFullNode/issues/1396

bokobza:
Good stuff.
@zeptin please approve if you're happy.

zeptin:
@bokobza Done, please merge

PR #1861, opened 2018-08-06
little reorg of the consensus manager file

monsieurleberre:
split of the file
use Action<ChainedHeaderBlock> instead of OnBlockDownloaded callback (I asked in the room and everybody was finding that more readable)

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
+++ b/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
@@ -4,9 +4,9 @@ using NBitcoin;
 namespace Stratis.Bitcoin.Consensus
 {
     /// <summary>
-    /// A request that holds information of blocks to download.
+    /// A request to the block puller that holds the chained headers of the blocks that are requested for download
     /// </summary>
-    public class BlockDownloadRequest
+    internal class BlockDownloadRequest
     {
         /// <summary>The list of block headers to download.</summary>
         public List<ChainedHeader> BlocksToDownload { get; set; }
--- a/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
+++ b/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
@@ -1,0 +1,14 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Consensus
+{
+    /// <summary>
+    /// A request that holds information of blocks to download.
+    /// </summary>
+    public class BlockDownloadRequest
+    {
+        /// <summary>The list of block headers to download.</summary>
+        public List<ChainedHeader> BlocksToDownload { get; set; }
+    }
+}
--- a/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
+++ b/src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
@@ -4,7 +4,7 @@ using NBitcoin;
 namespace Stratis.Bitcoin.Consensus
 {
     /// <summary>
-    /// A request to the block puller that holds the chained headers of the blocks that are requested for download
+    /// A request to the block puller that holds the chained headers of the blocks that are requested for download.
     /// </summary>
     internal class BlockDownloadRequest
     {
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -52,7 +52,7 @@ namespace Stratis.Bitcoin.Consensus
         /// <inheritdoc />
         public ChainedHeader Tip { get; private set; }
 
-        private readonly Dictionary<uint256, List<Action<ChainedHeaderBlock>>> callbacksByBlocksRequestedHash;
+        private readonly Dictionary<uint256, List<OnBlockDownloadedCallback>> callbacksByBlocksRequestedHash;
 
         /// <summary>Peers mapped by their ID.</summary>
         /// <remarks>This object has to be protected by <see cref="peerLock"/>.</remarks>
@@ -116,7 +116,7 @@ namespace Stratis.Bitcoin.Consensus
             this.expectedBlockDataBytes = 0;
             this.expectedBlockSizes = new Dictionary<uint256, long>();
 
-            this.callbacksByBlocksRequestedHash = new Dictionary<uint256, List<Action<ChainedHeaderBlock>>>();
+            this.callbacksByBlocksRequestedHash = new Dictionary<uint256, List<OnBlockDownloadedCallback>>();
             this.peersByPeerId = new Dictionary<int, INetworkPeer>();
             this.toDownloadQueue = new Queue<BlockDownloadRequest>();
             this.ibdState = ibdState;
@@ -197,13 +197,13 @@ namespace Stratis.Bitcoin.Consensus
         }
 
         /// <inheritdoc />
-        public void OnPeerDisconnected(int peerId)
+        public void PeerDisconnected(int peerId)
         {
             this.logger.LogTrace("({0}:{1})", nameof(peerId), peerId);
 
             lock (this.peerLock)
             {
-                this.OnPeerDisconnectedLocked(peerId);
+                this.PeerDisconnectedLocked(peerId);
             }
 
             this.logger.LogTrace("(-)");
@@ -216,7 +216,7 @@ namespace Stratis.Bitcoin.Consensus
         /// </summary>
         /// <remarks>Have to be locked by <see cref="peerLock"/>.</remarks>
         /// <param name="peerId">The peer that was disconnected.</param>
-        private void OnPeerDisconnectedLocked(int peerId)
+        private void PeerDisconnectedLocked(int peerId)
         {
             this.logger.LogTrace("({0}:{1})", nameof(peerId), peerId);
 
@@ -309,7 +309,7 @@ namespace Stratis.Bitcoin.Consensus
 
                 if (fullValidationRequired)
                 {
-                    connectBlocksResult = await this.TrySwitchTipAfterFullValidationLocked(chainedHeaderBlock).ConfigureAwait(false);
+                    connectBlocksResult = await this.FullyValidateLockedAsync(chainedHeaderBlock).ConfigureAwait(false);
                 }
             }
 
@@ -406,7 +406,7 @@ namespace Stratis.Bitcoin.Consensus
         /// </remarks>
         /// <param name="proposedNewTip">Tip of the chain that will become the tip of our consensus chain if full validation will succeed.</param>
         /// <returns>Validation related information.</returns>
-        private async Task<ConnectBlocksResult> TrySwitchTipAfterFullValidationLocked(ChainedHeaderBlock proposedNewTip)
+        private async Task<ConnectBlocksResult> FullyValidateLockedAsync(ChainedHeaderBlock proposedNewTip)
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(proposedNewTip), proposedNewTip);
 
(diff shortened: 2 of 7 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -99,7 +52,7 @@ public class ConsensusManager : IConsensusManager, IDisposable
         /// <inheritdoc />
         public ChainedHeader Tip { get; private set; }
 
-        private readonly Dictionary<uint256, List<OnBlockDownloadedCallback>> callbacksByBlocksRequestedHash;
+        private readonly Dictionary<uint256, List<Action<ChainedHeaderBlock>>> callbacksByBlocksRequestedHash;
```
noescape00: I don't think delegate should be replaced with an action since the delegate has a comment on it which improves readability.

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Consensus
+{
+    /// <summary>
+    /// A request that holds information of blocks to download.
+    /// </summary>
+    public class BlockDownloadRequest
```
noescape00: Why not internal?

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Consensus
+{
+    /// <summary>
+    /// A request that holds information of blocks to download.
```
noescape00: request to block puller

it holds chained headers of the blocks that are requested for download

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -237,20 +190,20 @@ public ConnectNewHeadersResult HeadersPresented(INetworkPeer peer, List<BlockHea
             }
 
             if (triggerDownload && (connectNewHeadersResult.DownloadTo != null))
-                this.DownloadBlocks(connectNewHeadersResult.ToArray(), this.ProcessDownloadedBlock);
+                this.DownloadBlocks(connectNewHeadersResult.ToArray(), this.OnBlockDownloaded);
```
noescape00: `OnBlockDownloaded`. Usually smth that strats with `On` is an event. In our case it's a normal method so I think old name was less missleading

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -237,20 +190,20 @@ public ConnectNewHeadersResult HeadersPresented(INetworkPeer peer, List<BlockHea
...
 
             this.logger.LogTrace("(-):'{0}'", connectNewHeadersResult);
             return connectNewHeadersResult;
         }
 
         /// <inheritdoc />
-        public void PeerDisconnected(int peerId)
+        public void OnPeerDisconnected(int peerId)
```
noescape00: Same as before. It's not an event.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -356,7 +309,7 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeaderBlock chainedH
 
                 if (fullValidationRequired)
                 {
-                    connectBlocksResult = await this.FullyValidateLockedAsync(chainedHeaderBlock).ConfigureAwait(false);
+                    connectBlocksResult = await this.TrySwitchTipAfterFullValidationLocked(chainedHeaderBlock).ConfigureAwait(false);
```
noescape00: Methods that start with `try` usually return `bool`. Here it's not the case. 
`Async` suffix is missing. 

I think old name was perfectly fine, no need to change it.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -756,7 +711,7 @@ private async Task<ConnectBlocksResult> ConnectBlockAsync(ChainedHeader chainTip
                 return failureResult;
             }
 
-            lock (this.peerLock)
+            lock (this.peerLock) //why do we lock here for instance ?
```
noescape00: Ask on slack, not in code, please.

And we need to lock here because all access to CHT should be protected by the `peerLock`.

src/Stratis.Bitcoin/Consensus/ValidationResults/PartialValidationResult.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Stratis.Bitcoin.Primitives;
```
noescape00: Why do we need to put small similar classes into several files? 

This one and ValidationResult can be in the same file.

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
monsieurleberre: Don't know, the original was public :) Agreed for internal

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
monsieurleberre: OK I can improve it as part of this PR too no pb

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
monsieurleberre: I disagree here : the comment was 
```    /// <summary>	
    /// A delegate that is used to send callbacks when a bock is downloaded from the of queued requests to downloading blocks.	
    /// </summary>	
    /// <param name="chainedHeaderBlock">The pair of the block and its chained header.</param>
```
The summary is basically explaining what a callback is more than anything, and the name of the dictionary hints at where the callbacks are called. The param simply describes the class ChainedHeaderBlock.
I am happy to comment the Dictionary if we can't agree that the comment is not that valuable

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
monsieurleberre: yes sure, I did planned to remove that

src/Stratis.Bitcoin/Consensus/ValidationResults/PartialValidationResult.cs
monsieurleberre: I think splitting into smaller files helps reducing the amount of information you have to keep in mind, if you then need this information you can have it on a separate window, or peek at it.
Also, when many contributors are involved it makes changes less likely to conflict for instance...

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
Aprogiena: If you don't like a comment, improve it, don't delete it

fassadlr: @monsieurleberre Im going to have to agree with @noescape00 on this one, can you please revert this change? We can look at this in following PRs. Please try and keep the changes minimal.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Agreed, ProcessDownloadedBlock was a callback not an event.

src/Stratis.Bitcoin/Consensus/ValidationResults/ValidationResult.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Stratis.Bitcoin.Consensus.ValidationResults
+{
+    public class ValidationResult
```
fassadlr: Can comment here:

```
    /// <summary>
    /// Feedback after a block was validated.
    /// </summary>
```

src/Stratis.Bitcoin/Consensus/ValidationResults/PartialValidationResult.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Stratis.Bitcoin.Consensus.ValidationResults
+{
+    public class PartialValidationResult : ValidationResult
```
fassadlr: Can comment this?

```
    /// <summary>
    /// Feedback specific to a partial block validation.
    /// </summary>
```

src/Stratis.Bitcoin/Consensus/ValidationResults/PartialValidationResult.cs
Aprogiena: Too much of separation actually makes it less readable. If it is something very closely related, it is more readable for me to have it in the same file.

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Consensus
+{
+    /// <summary>
+    /// A request to the block puller that holds the chained headers of the blocks that are requested for download
```
noescape00: `.` missing

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -284,7 +237,7 @@ private void PeerDisconnectedLocked(int peerId)
         /// <summary>
         /// A callback that is triggered when a block that <see cref="ConsensusManager"/> requested was downloaded.
         /// </summary>
-        private void ProcessDownloadedBlock(ChainedHeaderBlock chainedHeaderBlock)
+        private void OnBlockDownloaded(ChainedHeaderBlock chainedHeaderBlock)
```
noescape00: Same as for other ones: it's not a callback, please return the old name

src/Stratis.Bitcoin/Consensus/IConsensusManager.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// First we check if the block exists in chained header tree, then it check the block store and if it wasn't found there the block will be scheduled for download.
+        /// Given callback is called when the block is obtained. If obtaining the block fails the callback will be called with <c>null</c>.
+        /// </remarks>
+        /// <param name="blockHashes">The block hashes to download.</param>
+        /// <param name="onBlockDownloadedCallback">The callback that will be called for each downloaded block.</param>
+        Task GetOrDownloadBlocksAsync(List<uint256> blockHashes, OnBlockDownloadedCallback onBlockDownloadedCallback);
+    }
+
+    /// <summary>
+    /// A delegate that is used to send callbacks when a bock is downloaded from the of queued requests to downloading blocks.
```
Aprogiena: Bock -> block

src/Stratis.Bitcoin/Consensus/BlockDownloadRequest.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: This has been reverted.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: This has been reverted.

src/Stratis.Bitcoin/Consensus/IConsensusManager.cs
fassadlr: Fixed

fassadlr:
@noescape00 I have dismissed your review as @Aprogiena has approved and I have made the changes you requested on @monsieurleberre PR 👍

Aprogiena:
@fassadlr Unless totally necessary, I would recommend not to dismiss anyone's review. May or may not be okay in this case, probably will be, but that is not up to us to decide. There shouldn't be a reason to do such a thing under normal circumstances.

fassadlr:
@Aprogiena all of @noescape00's issues was addressed and it was holding up the queue. I need Matt's changes to avoid merge conflicts as I need to push my PR asap for you guys to review.

PR #1865, opened 2018-08-07
[SC] Addresses generated with nonce that increments with each create.

### LINKED ISSUE #1864: [SC] New contract addresses should be derived from the transaction id concatenated with a counter.
Subtask for #1833 

To create multiple contracts inside of a single transaction call, we need to be able to get the address from `Keccak(txid + counter)` where counter is incremented for each new contract inside this transaction. It's currently only `Keccak(txid)`.

codingupastorm:
Precursor to the rest of contract creation.

Closes #1864.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/SmartContracts/SmartContractWalletTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/SmartContracts/SmartContractWalletTests.cs
@@ -162,7 +162,7 @@ namespace Stratis.Bitcoin.IntegrationTests.SmartContracts
                 // Ensure that boths nodes has the contract
                 ContractStateRepositoryRoot senderState = scSender.FullNode.NodeService<ContractStateRepositoryRoot>();
                 ContractStateRepositoryRoot receiverState = scReceiver.FullNode.NodeService<ContractStateRepositoryRoot>();
-                uint160 tokenContractAddress = transferContractTransaction.GetNewContractAddress();
+                uint160 tokenContractAddress = transferContractTransaction.GetNewContractAddress(0); // nonce is 0 for user contract creation.
                 Assert.NotNull(senderState.GetCode(tokenContractAddress));
                 Assert.NotNull(receiverState.GetCode(tokenContractAddress));
                 scSender.FullNode.MempoolManager().Clear();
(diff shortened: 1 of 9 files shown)


codingupastorm:
So far all the affected tests are passing. Yet to run all.

src/Stratis.Bitcoin.IntegrationTests/SmartContracts/SmartContractWalletTests.cs
```diff
@@ -195,7 +195,7 @@ public void SendAndReceiveSmartContractTransactions()
                 // Ensure that boths nodes has the contract
                 senderState = scSender.FullNode.NodeService<ContractStateRepositoryRoot>();
                 receiverState = scReceiver.FullNode.NodeService<ContractStateRepositoryRoot>();
-                tokenContractAddress = transferContractTransaction.GetNewContractAddress();
+                tokenContractAddress = transferContractTransaction.GetNewContractAddress(0); // nonce is 0 for user contract creation.
```
fassadlr: @codingupastorm I agree with @majikandy on this one. Can you remove the comments here and move to the summary of the method? I.e. describe that the nonce is used for. I can just see the devs using this going forward will be definitely be interested in what the use of the nonce is here.

src/Stratis.SmartContracts.Core/NewContractAddressExtension.cs
```diff
@@ -10,14 +10,15 @@ public static class NewContractAddressExtension
         /// Get the address for a newly deployed contract.
         /// </summary>
         /// <param name="transaction"></param>
-        public static uint160 GetNewContractAddress(this Transaction transaction)
+        public static uint160 GetNewContractAddress(this Transaction transaction, ulong nonce)
```
fassadlr: Add a `param` here and document the use of the nonce.

src/Stratis.SmartContracts.Executor.Reflection/SmartContractExecutionContext.cs
```diff
@@ -53,6 +61,14 @@ public interface ITransactionContext
...
         /// </summary>
         uint160 To { get; }
+
+        /// <summary>
+        /// Get the next number to use as the nonce in the contract address calculation and then 
+        /// increment the number behind the scenes, so that the next contract address calculation
+        /// gets a new number. 
+        /// </summary>
+        /// <returns>The next nonce to use in the contract address formula.</returns>
+        ulong GetNonceAndIncrement();
```
ferdeen: Does `IncrementNonce()` sound better ?

rowandh: Agree with @ferdeen. The method really returns the incremented nonce.

codingupastorm: It returns the nonce, then increments.

rowandh: So it does, all good

ferdeen: If the calling code really needs to know that it will be incremented after then the method seems fine.  If it doesn't care then it could be called GetNonce(),  or even a Nonce Get property.

PR #1869, opened 2018-08-07
WalletSyncManager speedup

### LINKED ISSUE #1103: Wallet sync speedup

ferdeen:
Fixes #1103

The change to allow the Wallet to move forward while syncing is based on the[ Producer Consumer DataFlow pattern](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-implement-a-producer-consumer-dataflow-pattern).  Which has been implemented locally in the `WalletSyncManager` class.

- Include System.Threading.Tasks.Dataflow.

- Add consumer producer DataFlow pattern to `WalletSyncManager`.

- Use `IBlockStore` as `IBlockStoreCache` will be redundant after the core activation work.

- Use `asyncLoopFactory` on `WalletSyncManager.Start(`).  To process blocks from the `IBlockStore` and local concurrent queue - `blocksQueue`.

- Update related tests.


src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
...
-            this.walletManager.ProcessTransaction(transaction);
+            target.Post(block);
 
-            this.logger.LogTrace("(-)");
+            this.logger.LogDebug("(-)");
         }
 
         /// <inheritdoc />
-        public virtual void SyncFromDate(DateTime date)
+        private async Task ConsumerBlockAsync(ISourceBlock<Block> source, ConcurrentQueue<Block> queue)
```
bokobza: You should use `/// <inheritdoc />` when you don't want to rewrite the comments on a method implementing an interface member, which is not the case with these private methods.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
...
+            {
+                this.logger.LogError("Exception occurred: {0}", e.ToString());
+            }
 
             this.logger.LogTrace("(-)");
         }
 
         /// <inheritdoc />
-        public virtual void ProcessTransaction(Transaction transaction)
+        private void ProducerBlock(ITargetBlock<Block> target, Block block)
```
bokobza: `ProducerBlock` and `ConsumerBlockAsync` are 2 very small methods that are not used anywhere else but in `QueueBlock`. What's the advantage of splitting those in 3? I think the `QueueBlock` could include the other 2 and still be very readable.
In fact, `public void ProcessBlock(Block block)` only calls `QueueBlock` so it could be inlined there as well. 4 methods could be 1 without any loss of readability and a lot less logging :-)

Also the names `ProducerBlock` and `ConsumerBlockAsync` are not excessively descriptive.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
...
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public virtual void ProcessTransaction(Transaction transaction)
+        {
+            this.logger.LogTrace("()");
+
+            Guard.NotNull(transaction, nameof(transaction));
```
bokobza: `Guard` should be the first thing in the method.
You don't need `this.logger.LogTrace("()");` as `this.logger.LogTrace("({0}:'{1}')", nameof(transaction), transaction.GetHash());` acts as the entry log.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
...
+            int blockSyncStart = this.chain.GetHeightAtTime(date);
+            this.SyncFromHeight(blockSyncStart);
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public virtual void SyncFromHeight(int height)
+        {
+            this.logger.LogTrace("()");
```
bokobza: You don't need this line of log.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
...
+
+            this.walletManager.ProcessTransaction(transaction);
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public virtual void SyncFromDate(DateTime date)
+        {
+            this.logger.LogTrace("()");
```
bokobza: You don't need this line of log.

bokobza: Also can you please move SyncFromDate and SyncFromHeight back down to the end of the class? It's just that it'll be easier for us to compare the changes. Thanks 😗

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -96,157 +120,257 @@ public void Stop()
         }
 
         /// <inheritdoc />
-        public virtual void ProcessBlock(Block block)
+        public void ProcessBlock(Block block)
         {
-            Guard.NotNull(block, nameof(block));
+            this.logger.LogTrace("()");
```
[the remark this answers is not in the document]
ferdeen: Ok, cool.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: Ok, yes I realised this when I added debug logging to fix the wallet sync issue.  I should have removed these then.. thanks.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: They moved as I wanted to keep all private methods at the bottom.  They are back at the bottom of the class now.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

ferdeen:
Closing this one as I have implemented a version based on @Aprogiena's comments - new PR is #1890.

PR #1873, opened 2018-08-08
[SC] Abstract contract address generation

rowandh:
Introduces an abstraction `IAddressGenerator` for generating contract addresses.


rowandh:
All smart contracts tests passing

codingupastorm:
Makes sense but I liked the idea of having a method with the 0 automatically injected, as this is what we're using in 90% of cases.

PR #1875, opened 2018-08-08
[SC] Meter gas separately

### LINKED ISSUE #1754: Gas should be metered separately for internal contract calls.
Inside `InternalTransactionExecutor` when we make a Create or TransferFunds contract call, we should be creating a new GasMeter instance and injecting that into the virtual machine. 

Currently it always uses the meter from the initial contract call.

This way smart contract developers can declare how much gas they want to allow to be used in a internal call / create.

codingupastorm:
Closes #1754


src/Stratis.SmartContracts.Core/Exceptions/InsufficientGasException.cs
```diff
@@ -0,0 +1,13 @@
...
+    /// <summary>
+    /// Thrown when the remaining gas left for contract execution is less than the amount selected to allocate to an internal call. 
+    /// <para>
+    /// See <see cref="IInternalTransactionExecutor"/>.
+    /// </para>
+    /// </summary>
+    public sealed class InsufficientGasException : SmartContractException
+    {
+    }
+
```
fassadlr: Empty line

src/Stratis.SmartContracts.Core/Exceptions/InsufficientGasException.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Stratis.SmartContracts.Core.Exceptions
+{
+    /// <summary>
+    /// Thrown when the remaining gas left for contract execution is less than the amount selected to allocate to an internal call. 
+    /// <para>
+    /// See <see cref="IInternalTransactionExecutor"/>.
+    /// </para>
+    /// </summary>
+    public sealed class InsufficientGasException : SmartContractException
```
fassadlr: @codingupastorm seeing as this is only thrown from nested gas expenditure maybe a better name would be something like `InsufficientGasForNestedMeterException` ? Or can this exception be used elsewhere?

src/Stratis.SmartContracts.Executor.Reflection/InternalTransactionExecutor.cs
```diff
@@ -48,7 +50,8 @@ public ITransferResult TransferFunds(ISmartContractState smartContractState, Add
         {
             this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(addressTo), addressTo, nameof(amountToTransfer), amountToTransfer);
 
-            // TODO: The act of calling this should cost a lot of gas!
+            // TODO: Spend BaseFee here
```
fassadlr: Is this comment still relevant then?

codingupastorm: Yep! We need to spend the basefee as soon as we enter that part :)

src/Stratis.SmartContracts.Core/Exceptions/InsufficientGasException.cs
codingupastorm: It will also be used for Create calls from inside contracts, so I think that the name makes enough sense for now.

Lets change it in the future if we end up with more exceptions and this one isn't reused?

Issue #1737, opened 2018-07-25
SmartContractsTest and SmartContractsRegTest do not set MaxTimeOffsetSeconds


rowandh:
StratisMain uses `MaxTimeOffsetSeconds = 25 * 60`. We should consider a similar value.

PR #1879, opened 2018-08-09
[SC] PoS Smart Contract Network

codingupastorm:


codingupastorm:
@rowandh this should also close #1737

PR #1883, opened 2018-08-09
[SC] Closes #1766 Update smart contracts .NET versions to be same as FN

### LINKED ISSUE #1766: [SC] Smart contracts should use the same versions of .NET as the Full Node.
Currently all non-smartcontracts features are .NET Standard 2.0 whilst smart contracts ones are .NET Core 2.1. We should move all the features to .NET Standard 2.0 and ensure that the daemons are congruent with what the full node is doing too.

We will likely want to go live for SC with .NET Standard 2.0 on all libraries and .NET Core 2.1 as the node / daemon as it is an LTS release.

rowandh:
Updates everything to netstandard2.0 as used in FN.


fassadlr:
@rowandh any ideas why all the SC tests are failing? I'll as well 👍

rowandh:
This needs more work. I'll close it

PR #1889, opened 2018-08-09
Clean up TransactionBuilderContext to be easier for a user to use.

bokobza:
This PR was geared up to be something a bit different, in line with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1773 but in the end we opted for a less disruptive change.

* `Transaction Transaction` and `bool Sign` were removed from the TransactionBuilderContext object because the user shouldn't set them.
* The object originally passed to TransactionBuilderContext's ctor are now initialized via properties. This has the advantage to allow more flexibility as to where the inputs are taken from, and it also shows to the user that all the properties are settable and effectively on the same level.

Closes https://github.com/stratisproject/StratisBitcoinFullNode/pull/1773
Related to https://github.com/stratisproject/StratisBitcoinFullNode/issues/1689

@codingupastorm if you need changes with the changeAddress stuff, please do it as part of a SC issue.

🍒 All unit test + integration tests passing (all bar one).
🍒 Nodes are syncing.

image

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletTransactionHandlerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet.Tests/WalletTransactionHandlerTest.cs
@@ -292,17 +292,18 @@ namespace Stratis.Bitcoin.Features.Wallet.Tests
             };
 
             // create a trx with 3 outputs 50 + 50 + 49 = 149 BTC
-            var context = new TransactionBuildContext(this.Network, walletReference,
-                new[]
+            var context = new TransactionBuildContext(this.Network)
+            {
+                AccountReference = walletReference,
+                MinConfirmations = 0,
+                FeeType = FeeType.Low,
+                WalletPassword = "password",
+                Recipients = new[]
                 {
                     new Recipient { Amount = new Money(50, MoneyUnit.BTC), ScriptPubKey = destinationKeys1.PubKey.ScriptPubKey },
                     new Recipient { Amount = new Money(50, MoneyUnit.BTC), ScriptPubKey = destinationKeys2.PubKey.ScriptPubKey },
                     new Recipient { Amount = new Money(49, MoneyUnit.BTC), ScriptPubKey = destinationKeys3.PubKey.ScriptPubKey }
-                }
-                .ToList(), "password")
-            {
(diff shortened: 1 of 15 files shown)


bokobza:
This one is very easy when reviewed commit by commit.

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletTransactionHandlerTest.cs
```diff
@@ -318,10 +319,13 @@ public void FundTransaction_Given__a_wallet_has_enough_inputs__When__adding_inpu
...
             Transaction fundTransactionClone = this.Network.CreateTransaction(fundTransaction.ToBytes());
-            var fundContext = new TransactionBuildContext(this.Network, walletReference, new List<Recipient>(), "password")
+            var fundContext = new TransactionBuildContext(this.Network)
             {
+                AccountReference = walletReference,
                 MinConfirmations = 0,
-                FeeType = FeeType.Low
+                FeeType = FeeType.Low,
+                WalletPassword = "password",
+                Recipients = new List<Recipient>()
```
fassadlr: Can't this at least be in the constructor as `new List<Recipient>()`? Instead of having to define it here like this even if its an empty list?

bokobza: I moved them all out of the constructor so that the consumer doesn't wonder why this one is in the ctor and not the others.
So that they're all at the same level if that makes sense.

codingupastorm:
Logic looks good to me. 

I don't believe this closes #1773 though and I don't think this gets us any closer to safe public consumption of the `BuildTransaction` functionality, so I'll reopen the issue after this PR goes in.

bokobza:
I think it does bring us a bit closer, as Transaction and Sign are now removed and can't be set by the user.
Maybe in another PR we can make it clear that when you give a context to the TransactionBuilder methods, it comes back "enriched".

bokobza:
@codingupastorm @fassadlr if one of you can merge this, it'll be great.
@codingupastorm I've changed this issue to NOT close #1689 so we can improve on it.
However I'll remove the sp1 label from it as I don't think this is critical from production.

PR #1890, opened 2018-08-09
WalletSyncManager speedup with AsyncQueue<Block>

ferdeen:
Related to #1103.

This PR implements `AsyncQueue<Block>` and calls back to process a `Block`, in `OnProcessBlockAsync()`.

The aim of this change is to allow Consensus to advance, rather then wait for `WalletSyncManager` to process blocks - point 1 in #1103.

Unit tests passing:

image

Test evidence : 

image

Consensus is advancing, when a wallet that is processing a huge amount of transactions - at this wallet address [TNkjjYwkkeX9sfssCP5Qq4ta67d3meKS2k](https://chainz.cryptoid.info/strat-test/address.dws?TNkjjYwkkeX9sfssCP5Qq4ta67d3meKS2k.htm).


bokobza:
wow, so much simpler 😍 
How's the manual sync looking?
For info, @ferdeen has a wallet file in which some addresses contains thousands of transactions.
Previously, when trying to recover this wallet, some blocks would take long minutes to process and we could clearly see that the consensus was waiting in line for these blocks to move on.
With the new changes, we're expecting to see that on the same wallet, when the big blocks are being processed, the consensus continues to send blocks to the WalletSyncManager, filling the local queue as it does so.

ferdeen:
Consensus has advanced even though the wallet is still processing transactions - as mentioned by @bokobza above.

image

Aprogiena:
> With the new changes, we're expecting to see that on the same wallet, when the big blocks are being processed, the consensus continues to send blocks to the WalletSyncManager, filling the local queue as it does so.

though this may seem nice, it may also be dangerous if the wallet processes blocks slower than new blocks are added to it, you could easily go out of memory

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -95,8 +98,13 @@ public void Stop()
...
 
-        /// <inheritdoc />
-        public virtual void ProcessBlock(Block block)
+        /// <summary>Called when a block is added to the blocksQueue.
+        /// Depending on the WalletTip and incoming block height, this method will decide whether the block will be processed by the <see cref="WalletManager"/>.
+        /// </summary>
+        /// <param name="block">Block to be processed.</param>
+        /// <param name="cancellationtoken">The cancellation token.</param>
+        /// <returns><placeholder>A <see cref="Task"/> representing the asynchronous operation.</placeholder></returns>
+        private Task OnProcessBlockAsync(Block block, CancellationToken cancellationtoken)
```
bokobza: cancellationtoken => cancellationToken

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -248,5 +268,15 @@ public virtual void SyncFromHeight(int height)
 
             this.logger.LogTrace("(-)");
         }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            this.logger.LogTrace("()");
+
+            this.blocksQueue?.Dispose();
```
bokobza: I don't think blocksQueue can ever be null.

Aprogiena:
e.g. in the above pictures, it seems that we have 180k blocks in memory in the queue? that does not seem good if true

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletSyncManagerTest.cs
```diff
@@ -98,21 +102,23 @@ public void ProcessBlock_NewBlock_PreviousHashSameAsWalletTip_PassesBlockToManag
...
-                this.blockStoreCache.Object, this.storeSettings, this.nodeLifetime.Object);
+                this.blockStore.Object, this.storeSettings, this.nodeLifetime.Object);
             walletSyncManager.SetWalletTip(this.chain.GetBlock(3));
 
             Block blockToProcess = blocks[3];
             walletSyncManager.ProcessBlock(blockToProcess); //4th block in the list has same prevhash as which is loaded
 
-            uint256 expectedBlockHash = this.chain.GetBlock(4).Header.GetHash();
-            Assert.Equal(expectedBlockHash, walletSyncManager.WalletTip.Header.GetHash());
+            uint256 expectedBlockHash = this.ExpectedBlockHash(walletSyncManager, 4);
```
bokobza: Is this correct?

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletSyncManagerTest.cs
```diff
@@ -357,5 +363,35 @@ public void SetWalletTip(ChainedHeader tip)
...
+                    Thread.Sleep(millisecondsTimeout);
+                }
+                catch (OperationCanceledException e)
+                {
+                    Assert.False(true, $"{failureReason}{Environment.NewLine}{e.Message}");
+                }
+            }
+        }
+
+        private uint256 ExpectedBlockHash(IWalletSyncManager walletSyncManager, int blockHeight)
```
bokobza: Maybe change the name to `AssertTipBlockHash` or something.

bokobza:
> e.g. in the above pictures, it seems that we have 180k blocks in memory in the queue? that does not seem good if true

I guess that's correct.
Although this is an extreme wallet and in practice the queue will empty as it gets filled (slower but still).

And the queue can grow in size as much as is available so in practice I don't think it's a problem.

Now if we wanted to fix that, one option would be to limit the number of items in the queue. Say we limit it to 50K items and discard the rest. What will happen then is that when all 50K are processed and a new block is received, the syncManager will get into "I need to catch up" mode and will start getting blocks from the repository.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletSyncManagerTest.cs
ferdeen: ok

ferdeen:
>Now if we wanted to fix that, one option would be to limit the number of items in the queue. Say we limit it to 50K items and discard the rest. What will happen then is that when all 50K are processed and a new block is received, the syncManager will get into "I need to catch up" mode and will start getting blocks from the repository.

Or rather then wait for the queue to empty, have a threshold, so once below, say 50K then allow more blocks on the queue.

Aprogiena:
guys, 50k blocks on bitcoin is over 50 gigz, could you please try this on bitcoin network ?

Aprogiena:
i'd say you need not count limit, but size limit, something like 100MB and whatever fits, it fits. then you would actually need to block on signal and really block the consensus

src/Stratis.Bitcoin.Features.Wallet.Tests/WalletSyncManagerTest.cs
ferdeen: Yes, before adding this line the test would call `ProcessBlock()`, enqueue the block, then drop out the method - which is correct.  

However, this would invoke the callback `WalletSyncManager.OnProcessBlockAsync()` to determine if the block needs to be processed - and at the same time the test would start to Assert or Verify and fail.

To get around this the Assert on the expected block is wrapped in a `WaitLoop()` - which is now called `AssertTipBlockHash()`.  This does not return until the Func predicate returns true, or false after 30 seconds.

This is a better approach then using `Thread.Sleep()`.

bokobza:
> Or rather then wait for the queue to empty, have a threshold, so once below, say 50K then allow more blocks on the queue.

no because then you'll have a big gap in your queue and then you might as well discard them all.

@Aprogiena meh I can't do that on my phone no problem.
In all seriousness, we can look into limiting that.

Not sure how you'd check the physical size of the queue without popping the items.
Maybe looking at itemsMaxCount = (max queue size)/ (max block size)?

Aprogiena:
@bokobza we do know how big a block is actually, so we can know how big the queue is

also - what kind of info do we need? surely we don't need whole blocks, so why to have queue of blocks and not only info we need? how much does it cost to find out whether transaction is interesting?

bokobza:
yes but you can't get to the blocks without popping items out of the queue. 
2 ways to do this:
* either do an approximate calculation like I said above (knowing that it'll be suboptimal because blocks will always be <= (max block size) so you'll always stop enqueuing before the limit). 
* we calculate the size of items pushed to the queue and the size of items being dequeued.

ferdeen:
> guys, 50k blocks on bitcoin is over 50 gigz, could you please try this on bitcoin network ?

We don't have a Bitcoin wallet to mimic what we did above in Stratis TestNet.

I'll kick off a run on Bitcoin now to make sure it's advancing as expected.

bokobza:
> also - what kind of info do we need? surely we don't need whole blocks, so why to have queue of blocks and not only info we need? how much does it cost to find out whether transaction is interesting?

we need all the inputs, all the outputs, the hash of the block etc.
Yeah we might as well keep it as is.

So I think since this wallet is an extreme case, we either:
- merge as is, we'll improve later
- add a simple "max 500 blocks in the queue", knowing that this'll be 500MB max and we improve further later.

Thoughts?

Aprogiena:
Maybe we should make a call on Monday about this. Not available tomorrow that much, only early morning, so I'm not sure we can make one tomorrow

bokobza:
Can do mate.
@Aprogiena can you block the review please so we don't merge it by mistake?

ferdeen:
After dicussing this with @bokobza, we are going to limit the queue size as follows :

`const int MaxQueueSize = 500;`

In `WalletSyncManager.ProcessBlock(Block block)`

            if (blocksQueue.Count >= MaxQueueSize && blocksQueue.Count > 0)
                return

            this.blocksQueue.Enqueue(block);   

This way we will always let Consensus advance.

When the `blocksQueue.Count` is zero and the queue starts to fill up there will be a gap of missed blocks.  These blocks will be picked up - when it realises the gap  - then it will check `IBlockStore` to catch up.

Aprogiena:
That if condition is quite weird, why the second part of it?

This looks like a very temporary solution and it looks weird to me that we will even enqueue blocks when there is a gap, which I would probably not do. But it could somehow work, it is just weird

ferdeen:
> That if condition is quite weird, why the second part of it?

Adding the second part will not enqueue until the queue is back to zero.

> This looks like a very temporary solution and it looks weird to me that we will even enqueue blocks when there is a gap, which I would probably not do. But it could somehow work, it is just weird

The gap will be sorted out here -  https://github.com/ferdeen/StratisBitcoinFullNode/blob/33d92a6291d85d3f52c5469fce027a3ff027fa3e/src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs#L175-L202). 

> This looks like a very temporary solution 

What's your alternative solution ?

Aprogiena:
> Adding the second part will not enqueue until the queue is back to zero.

```        if (blocksQueue.Count >= MaxQueueSize && blocksQueue.Count > 0)```

How is this not equivalent to

```        if (blocksQueue.Count >= MaxQueueSize)```

For any value of the constant bigger than zero?

**Comment by Aprogiena** [comment]
I understand that the gap is being handled, but it seems useless to enqueue new blocks in the meanwhile

**Comment by Aprogiena** [comment]
> What's your alternative solution ?

Better solutions would require little more work, so I think this weird solution is okay for now, we just should know that it is temporarily

**Comment by bokobza** [comment]
I'll explain the reasoning with the use of scenarios.
This is an important bit, we need to get it right.

## scenario 1 - everything up to date:
**start state:**
- header tip is at 1000.
- consensus is at 1000 
- wallet is synced at height 1000.

**now:**
- blocks are arriving, they get queued, dequeued, processed => no problem.


## scenario 2 - consensus up to date, wallet behind:
**start state:**
- header tip is at 1000.
- consensus is at 1000 
- wallet is synced at height 100.

**now:** 
- block 1001 is received and queued
- then: it is dequeued and the syncManager will see that we're too much ahead and will go to the block store to catch up (the syncManager already knows how to do that)
- then: the queue will fill up slowly (1 block every minute) and the syncManager will wait to have caught up to process them
- then: even if the wallet already has blocks 1000-1200 and they are dequeued, the syncManager knows how to discard them.

## scenario 3 - consensus behind, wallet behind:
**start state:**
- header tip is at 1000.
- consensus is at 100 (need to download and save 900 blocks)
- wallet is synced at height 100.

**now:**
- blocks 101, 102, 103 etc are arriving, getting into the queue
- then: they get dequeued, but not fast enough, so the queue fills up so the last item in - then: the last item in the queue is 600 (or 595 if blocks 1001, 1002...1005 have arrived in the meantime).
- then: blocks 601, 602 etc arrive as well.

3 solutions:
1. we make blocks 601, 602 etc.. wait their turn => consensus is blocked again and all we've done is reduced the number of pending blocks by 500.
2. we discard them while the queue is full, but as soon as there is space, we push them to the queue => the problem with that is that if block 101 takes a loooong time to process, we'll be discarding block 601, 602 etc (because the queue is still full) and then when the queue gets more empty, blocks 610 will be the next one to be enqueued. So in the queue you'll have items 599, 600, 610, which is wrong.
3. we discard blocks 601, 602 etc... and we only allow more to get into the queue when the queue is empty. So now the items 100-600 get processed, blocks 601-699 gets discarded, and then when item 700 is received by the queue, the syncManager will get items from the block store and catch up everything that's missing. 

The key takeaway is that the syncManager already knows how to deal with blocks that are coming in **any** order.

The option we want to implement is 3.
The problem with that is that it's very difficult to predict and keep items in order when you need to discard some. Are the items in the block store? If I've processed all of them, will the next items in the queue be the right ones to process? etc...

Now mind you, the only scenario where this whole logic is applicable is scenario 3, when a use is in IBD and has a wallet present. Not likely to happen very often.

Happy to have a call on Monday to discuss this further.
@Aprogiena @ferdeen @dangershony 

PS: we can also look at preventing the user from creating a wallet while in IBD, which will make the wallet's logic so so much simpler. In Bitcoin I think they try to prevent you from moving forward until you're sync, don't they?

File: src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs [hunk 5]
```diff
@@ -32,24 +32,27 @@ public class WalletSyncManager : IWalletSyncManager
 
         public ChainedHeader WalletTip => this.walletTip;
 
+        /// <summary>Queue which contains blocks that should be processed by <see cref="WalletManager"/></summary>
```
Aprogiena: Don't forget to end your sentences with a `.`

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -95,8 +98,13 @@ public void Stop()
             this.logger.LogTrace("(-)");
         }
 
-        /// <inheritdoc />
-        public virtual void ProcessBlock(Block block)
+        /// <summary>Called when a block is added to the blocksQueue.
```
Aprogiena: Use `<see>` when you talk about an existing object like the block queue or wallet tip below.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -95,8 +98,13 @@ public void Stop()
...
+        /// Depending on the WalletTip and incoming block height, this method will decide whether the block will be processed by the <see cref="WalletManager"/>.
+        /// </summary>
+        /// <param name="block">Block to be processed.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns><placeholder>A <see cref="Task"/> representing the asynchronous operation.</placeholder></returns>
```
Aprogiena: We don't really use returns section for methods that the returns plain task

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: I've removed.

Maybe we need to remove this from Style Cop.  If there return value isn't documented there is a warning :

image

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -168,7 +176,7 @@ public virtual void ProcessBlock(Block block)
                         {
                             token.ThrowIfCancellationRequested();
 
-                            nextblock = this.blockStoreCache.GetBlockAsync(next.HashBlock).GetAwaiter().GetResult();
+                            nextblock = this.blockStore.GetBlockAsync(next.HashBlock).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: Yes, it should not produce warning in case there is no return value except for the task

ferdeen:
After a call with @bokobza and @Aprogiena, the interim solution is to go with option3.

> We discard blocks 601, 602 etc... and we only allow more to get into the queue when the queue is empty. So now the items 100-600 get processed, blocks 601-699 gets discarded, and then when item 700 is received by the queue, the syncManager will get items from the block store and catch up everything that's missing.

Longer term @Aprogiena mentioned we can ask ConsensusManager for a particular block in a callback (secondary source, primary source is via the CM signaller).  The callback will then return null (no block found), or the actual block.  

We can look at this in more detail when we tackle the rest of issue #1103.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -35,6 +35,12 @@ public class WalletSyncManager : IWalletSyncManager, IDisposable
         /// <summary>Queue which contains blocks that should be processed by <see cref="WalletManager"/>.</summary>
         private readonly AsyncQueue<Block> blocksQueue;
 
+        /// <summary>Limit blocksQueue size.</summary>
+        private const int MaxQueueSize = 1000;
```
Aprogiena: This is too much, this should be on the order of 100 or 200

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -35,6 +35,12 @@ public class WalletSyncManager : IWalletSyncManager, IDisposable
         /// <summary>Queue which contains blocks that should be processed by <see cref="WalletManager"/>.</summary>
         private readonly AsyncQueue<Block> blocksQueue;
 
+        /// <summary>Limit blocksQueue size.</summary>
```
Aprogiena: use `<see>`

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: When I ran 500, 1000, 2000 blocks - when I was trying to setup a unit test, the blocks were being processed faster than the queue building up.  The queue count would stay around 0-1.

Admittedly the blocks have minimal txn's so not a real use case.  

Shall we bring it back down to 500 ?

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: I'll get this right one day! Cool.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
Aprogiena: You need to think about the worst-case scenario when setting such limits. The worst-case scenario is the queue is full on bitcoin network when all blocks are very big. So let's assume you have 2 MB average block, then 100 limit here means 200 MB of memory that you need to have for this. I don't think it is reasonable to go to 500 or even thousand. I would go with just 100

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -95,13 +109,21 @@ public void Stop()
...
+        /// Depending on the <see cref="WalletTip"/> and incoming block height, this method will decide whether the <see cref="Block"/> will be processed by the <see cref="WalletManager"/>.
+        /// </summary>
+        /// <param name="block">Block to be processed.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        private async Task OnProcessBlockAsync(Block block, CancellationToken cancellationToken)
         {
             Guard.NotNull(block, nameof(block));
             this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
 
+            if (block.BlockSize != null)
```
Aprogiena: This should not be needed because there should be a guarantee that the block size is calculated properly every time. Only in case of some incomplete test setup, this can be null

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -214,6 +236,44 @@ public virtual void ProcessBlock(Block block)
...
+
+            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
+
+            if (!this.walletManager.ContainsWallets)
+            {
+                this.logger.LogTrace("(-)[NO_WALLET]");
+                return;
+            }
+
+            if (block.BlockSize != null)
```
Aprogiena: As mentioned above, this should not be needed

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -214,6 +236,44 @@ public virtual void ProcessBlock(Block block)
...
+                this.blocksQueueSize += block.BlockSize.Value;
```
Aprogiena: You must move this to line 272, you only want to increase when you actually enqueue after that. So down below if you want to enqueue, first increase the size, then enqueue. But we cannot increase the queue size for blocks that we don't enqueue

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
```diff
@@ -214,6 +236,44 @@ public virtual void ProcessBlock(Block block)
...
+                if (this.blocksQueueSize >= MaxQueueSize)
+                {
+                    this.maxQueueSizeReached = true;
+                    this.logger.LogTrace("(-)[REACHED_MAX_QUEUE_SIZE]");
+                    return;
+                }
+            }
+            else
+            {
+                // If queue is empty the reset the maxQueueSizeReached flag.
```
Aprogiena: the`n`

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: Ar yes, my mistake - will move down.

src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs
ferdeen: ok

bokobza:
Ok so I've run a bunch of manual tests where I'm syncing wallets when the blockchain is fully downloaded and when it's not, resyncing wallet etc.. and all seem to work as expected.

Issue #1877, opened 2018-08-08
[Activation] List of issues to resolve


noescape00:
Those are the issues that are needed to be done in order to finish work on activation (besides ones that are already created and marked with `Activation` label)

_I'll be constantly updating this list._

1. **[MERGED]** Add max limitation on amount of blocks asked from puller besides the mb size limitation
PR: #1878

2. **[MERGED]** BlockStoreBehavior - fix todos that are left for activation
PR: #1881

3. **[MERGED]** PR that saves tips of block store and FinalizedHash (save height and not only hash). Change initialization to use those values. If finalized height is above consensus save it to consensus tip height.
PR: #1882 (just save hash and height of block repository and finalized block)

4.  **[MERGED]** Remove redundant files related to BSC and BR performance counting. 
PR: #1891

5. **[DONE]** Check initialization sequence of all new components and node in general

6. Ensure that all validators throw and the exception is not swallowed (at least header validator swallows the exception)
Part 1: **[MERGED]** PR: #1918
Part 2:  **[MERGED]**  PR: #1930

7. **[FIXED]** Bug: consensus rewinds on startup (probably wrong initialization or smth with coinview)

8. Fix some failing unit tests
Part 1: **[MERGED]** PR: #1898
Part 2: **[MERGED]** PR: #1910

9. **[MERGED]** Remove fixes for the gap situation after coinview rewind data aggregation issue is resolved
PR: #1964

noescape00:
Fixed

PR #1891, opened 2018-08-09
[ACTIVATION] Redundant files removal

noescape00:
#1877


noescape00:
they are not used anywhere. 
also half of them belong to component which we disconnected for good (block store cache)

bokobza:
Every dev that wants to remove code gets a special place in my heart.

Aprogiena:
agree with most, but not sure about block repository one

noescape00:
>agree with most, but not sure about block repository one

1) When was last time anyone looked at those performance metrics? 
2) We have logs to investigate performance issues
3) DBreeze is never a bottleneck
4) This performance counter has interface which is overkill: `AddCacheHitCount(1);` and other methods always accept integer but never any value except for 1 is passed
5) There is `public virtual BlockStoreCachePerformanceCounter BlockStoreCachePerformanceCounterFactory()` do we really need to have it? Do we expect some component to inherit from block store and override performance counter?
6) AddRepositoryMissCount - this shouldn't even exist. If component asks for a block which is not on best chain from the repository then this component has a bug or there was a reorg just recently

Aprogiena:
LOL, seven reasons, okay I guess then

bokobza:
Alt Text

PR #1895, opened 2018-08-10
Spending staking balance #1294

### LINKED ISSUE #1294: spending staking balance doesn't reduce the staking weight even though staking stops for that stake
spending balance during staking doesn't reduce the staking weight.  

It does change the staking to false, even though enabled is true.

Eg a wallet with 3ish strat, started staking and then spent 3 strat.

http://localhost:37221/api/Miner/getstakinginfo
Response Body
{
  "enabled": true,
  "staking": false,
  "errors": null,
  "currentBlockSize": 1051,
  "currentBlockTx": 3,
  "pooledTx": 2,
  "difficulty": 2801938.94206241,
  "searchInterval": 16,
  "weight": 300000000,
  "netStakeWeight": 2416188529798498,
  "expectedTime": 484818631
}

kogot:
Closes #1294
If wallet has just 1 UTXO -> start Staking -> make transaction => Staking will be temporary stoped till new UTXO get enough depth(20/50 without "activation"). 
If within that time call GetStakingInfo() it will return old invalid values for "weight" & "ExpectedTime". That's what this PR will fix.

If stop Staking => rpcGetStakingInfoModel will be reseted.


bokobza:
@kogot can you please write Closes https://github.com/stratisproject/StratisBitcoinFullNode/issues/1294 at the top of the page in your first comment? This way the issue will be closed automatically when this PR is merged.
Thanks!

monsieurleberre:
Currently reviewing, I am happy to run the whole unit testing suite and try staking on that PR if people think it is needed :)

kogot:
@monsieurleberre I did many different tests in testnet, staking-spending-staking...
You're more than welcome to try staking and run all unit tests ;)

monsieurleberre:
Just tried the staking, and yes it does work :) but I think I shot myself in the foot by spending a big coin during the test and now I only have 10 strats mature enough to stake. I will wait for the 499,990 other strats to mature again and update you :) I'm sure it's all good though 👍

src/Stratis.Bitcoin.Features.Miner/Staking/PosMinting.cs
```diff
@@ -647,6 +647,8 @@ public async Task<bool> CreateCoinstakeAsync(List<UtxoStakeDescription> utxoStak
             if (!stakingUtxoDescriptions.Any())
             {
                 this.rpcGetStakingInfoModel.Staking = false;
+                this.rpcGetStakingInfoModel.Weight = balance;
+                this.rpcGetStakingInfoModel.ExpectedTime = balance != 0 ? StakeValidator.TargetSpacingSeconds * this.networkWeight / this.rpcGetStakingInfoModel.Weight : 0;
```
monsieurleberre: maybe here instead of dividing by this.rpcGetStakingInfoModel.Weight, I would divide by balance, just makes it a bit easier to follow (or compare this.rpcGetStakingInfoModel.Weight to 0 instead of balance). No big deal for me :)

kogot: yeah, make sense.

Aprogiena: Basically something like we do online 361

Aprogiena: And also it should probably be done on line 638

kogot: @Aprogiena won't it confuse? cause you didn't actually stop staking as a user, its just node waiting till your UTXOs get some depth because you just spent some coins.
And you were talking about "stop" staking. Yes, when staking is stopped, values will be reset to 0.

Aprogiena: The original argument in the issue was that it is confusing when it is not reset to 0 and the staking is not in progress. Basically we say that you will mine a block in about one hour (for example) in that information but at the same time you are not staking at all, so obviously you will not mine a block about one hour.

See the discussion on the issue. I argued that all values should be ignored if staking is not in progress and therefore does not matter which values are there. Dan was of an opinion that it doesn't matter and therefore it should be reset to 0, which I don't have a problem with, so this is what I was asking for here.

Aprogiena: In any case, whichever option we choose, it should be implemented in all three places, not only at one place

kogot: Okey mate, it's easy to me, I will do reset in 3 places

kogot:
@Aprogiena @monsieurleberre please check.

Aprogiena:
I'm okay with that. I would prefer not having the pause method at all because I don't see any benefit of having it and I would always call the stop one.

Aprogiena:
I'm just hoping that one day people will learn to write good comments, LOL

PR #1897, opened 2018-08-10
Https Api endpoint

monsieurleberre:
Currently not working for Mac and Linux - the certificate store is different and we will need a different strategy, unless we missed something from the .Net Core framework

User can provide the name of an exiting certificate to use from his Root store in appsettings
Otherwise, the default is to create a default certificate with a random password and add it to the root store if it doesn't already exist (this will give a warning dialog in Windows) otherwise, just re-use the default certificate from last run (no warnings, no change in user experience).


Aprogiena:
Why do you use store at all?

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
```diff
@@ -25,33 +25,58 @@ public TestBase(Network network)
...
+                if (deleteAttempts >= 50)
+                    throw new Exception(string.Format("The test folder: {0} could not be deleted.", directoryWithTestNamePrefix));
+
+            }
 
             Directory.CreateDirectory(dir);
             return dir;
         }
 
+        private static IEnumerable<DirectoryInfo> GetDirectoriesToDelete(string dir)
```
ferdeen: Can you invert the if statement so you can return early and `Enumerable.Empty<DirectoryInfo>()` is not always created.

Something like this :

```
        private static IEnumerable<DirectoryInfo> GetDirectoriesToDelete(string dir)
        {
            var dirInfo = new DirectoryInfo(dir);
            var parentDirInfo = dirInfo.Parent;

            if (parentDirInfo == null && !parentDirInfo.Exists)
                return Enumerable.Empty<DirectoryInfo>();

            string guidSuffixPattern = "_????????-????-????-????-????????????";
            string searchPattern = dirInfo.Name.Substring(0, dirInfo.Name.LastIndexOf("_")) + guidSuffixPattern;

            return  parentDirInfo.EnumerateDirectories(searchPattern);
        }
```

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -106,7 +108,8 @@ public ApiSpecification(ITestOutputHelper output) : base(output)
         protected override void BeforeTest()
         {
             this.sharedSteps = new SharedSteps();
-            this.httpClient = new HttpClient();
+            this.httpHandler = new HttpClientHandler() { ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => true };
```
ferdeen: This callback always returns true.  Is this work in progress ?

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -106,7 +108,8 @@ public ApiSpecification(ITestOutputHelper output) : base(output)
...
+            this.httpClient = new HttpClient(httpHandler);
```
ferdeen: `this.httpHandler`

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -106,7 +108,8 @@ public ApiSpecification(ITestOutputHelper output) : base(output)
         protected override void BeforeTest()
         {
             this.sharedSteps = new SharedSteps();
-            this.httpClient = new HttpClient();
```
ferdeen: Does this line look better like this ? I'm not too sure : 

```
this.httpHandler = new HttpClientHandler
{
    ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => true
};
```

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -107,6 +112,12 @@ public override void Dispose()
 
     public sealed class ApiFeatureOptions
```
ferdeen: This class name doesn't seem to match what its doing.  If it's going to include more API type options then fine, if it's only going to deal with Certification then maybe it needs to be renamed.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisPosApiRunner.cs
```diff
@@ -31,7 +32,7 @@ public override void BuildNode()
                             .UseMempool()
                             .AddPowPosMining()
                             .UseWallet()
-                            .UseApi()
+                            .UseApi(option => option.CertificateStoreName = StoreName.My)
```
ferdeen: Why is the property being set as an Action ? 

Would this read better :
```


public override void BuildNode()
{
    var settings = new NodeSettings(args: new string[] { "-conf=bitcoin.conf", "-datadir=" + this.DataFolder });
    var options = new ApiFeatureOptions(StoreName.My);

    this.FullNode = (FullNode)new FullNodeBuilder()
        .UseNodeSettings(settings)
        .UseBlockStore()
        .UsePowConsensus()
        .UseMempool()
        .AddMining()
        .UseWallet()
        .AddRPC()
        .UseApi(options)
        .MockIBD()
        .Build();
}

```

Then in ApiFeature : 

```
public static class ApiFeatureExtension
{
    public static IFullNodeBuilder UseApi(this IFullNodeBuilder fullNodeBuilder, ApiFeatureOptions options = null)
    {
        CertificateStore certificateStore = null;

        if (options != null)
            certificateStore = new CertificateStore(options.CertificateStoreName, StoreLocation.CurrentUser);

        fullNodeBuilder.ConfigureFeature(features =>
        {
            features
                .AddFeature<ApiFeature>()
                .FeatureServices(services =>
                {
                    services.AddSingleton(options);
                    services.AddSingleton<ApiSettings>();

                    if (certificateStore != null)
                        services.AddSingleton<ICertificateStore>(certificateStore);
                });
        });

        return fullNodeBuilder;
    }
}
```

monsieurleberre:
Let me remove the change that adds a Guid to the folder name

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
monsieurleberre: OK I am not too fussed about what the name is/was, I think this is a certificate that is only used for the ApiFeature so I am not shocked to have its location in the ApiFeatureOptions, do you have a better name ?

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisPosApiRunner.cs
monsieurleberre: I just didn't feel like changing the option or apifeature class too much
If everybody feels like it is bad I will change no pb, let me know

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
monsieurleberre: no this is to bypass the "invalid certificate" errors you receive by using a certificate that is not in the "Root" store.
for tests, we don't put the certificate in the "Root" store because you get a windows pop up warning when you do that for the first time

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
monsieurleberre: tru dat bro

monsieurleberre:
@Aprogiena so far, on windows, we have not found a workaround that gives a seamless user experience. It looked like the certificate needed to be in the Trusted Root Certification Authority folder, otherwise Chrome and other browsers get in panick mode.
This is definitely not ideal and very windows bound, I am looking at alternatives for Mac and Linux, and hopefully they can port to Windows too. We wanted to put that in PR now for people to let us know what they think in the meantime

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -107,6 +112,12 @@ public override void Dispose()
...
     {
+        public StoreName CertificateStoreName { get; set; }
+
+        public ApiFeatureOptions(StoreName storeName = StoreName.Root)
```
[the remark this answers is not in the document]
monsieurleberre: Then you get warnings in chrome that the certificate is not secure, do we want that ?
image
this one plus a big page saying where you have to go to "advanced" and choose to accept the certificate anyway...

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -120,6 +131,8 @@ public static IFullNodeBuilder UseApi(this IFullNodeBuilder fullNodeBuilder, Act
             var options = new ApiFeatureOptions();
             optionsAction?.Invoke(options);
 
+            var certificateStore = new CertificateStore(options.CertificateStoreName, StoreLocation.CurrentUser);
```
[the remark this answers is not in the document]
monsieurleberre: It looked good here to me as the action hapenning above needs to happen before, but let's try to find another way

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -112,9 +120,10 @@ public static void PrintHelp(Network network)
         {
             var builder = new StringBuilder();
 
-            builder.AppendLine($"-apiuri=<string>          URI to node's API interface. Defaults to '{ DefaultApiHost }'.");
-            builder.AppendLine($"-apiport=<0-65535>        Port of node's API interface. Defaults to { GetDefaultPort(network) }.");
-            builder.AppendLine($"-keepalive=<seconds>      Keep Alive interval (set in seconds). Default: 0 (no keep alive).");
+            builder.AppendLine($"-apiuri=<string>                     URI to node's API interface. Defaults to '{ DefaultApiHost }'.");
+            builder.AppendLine($"-apiport=<0-65535>                   Port of node's API interface. Defaults to { GetDefaultPort(network) }.");
+            builder.AppendLine($"-keepalive=<seconds>                 Keep Alive interval (set in seconds). Default: 0 (no keep alive).");
+            builder.AppendLine($"-certificatesubjectname=<string>     Subject name of the certificate to use for https traffic encryption. Defaults to { DefaultCertificateSubjectName }.");
```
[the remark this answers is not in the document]
monsieurleberre: As you like honestly I don't mind, I think it is not complicated and might be helping some people but...

Aprogiena:
> so far, on windows, we have not found a workaround that gives a seamless user experience. It looked like the certificate needed to be in the Trusted Root Certification Authority folder, otherwise Chrome and other browsers get in panick mode.

@monsieurleberre How is this envisioned to be used? Is this is API, why do we care about chrome? I would expect some API client to access the API and not a browser. So where does browser come in?

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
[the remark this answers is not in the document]
monsieurleberre: image

monsieurleberre:
@Aprogiena so currently I think people are relying on a fair bit on Swagger to do the calls, therefore they use the browsers. I think there is also a plan to have apps in browser at a later stage.
Then if I try to use only a local certificate (in "my" store) then in postman and probabl other stuff like fiddler, you need to specifically ask it to not do SSL validation, otherwise you get errors that say "couldn't get any response"

Aprogiena:
End-users are not going to use swagger, so I don't see a valid use case there. As for the applications in the browser, it still seems completely useless to use the store. HTTPS only makes sense if you try to connect to remote location. As such you would expect the client browser to be a different machine than the node, so installing a certificate into the node certificate store will not affect the client browser. I still can't see a valid use case why should we store the Certificate into the certificate store.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisPosApiRunner.cs
ferdeen: I was thinking it looks more consistent with how `NodeSettings` is setup.

Also if new functionality is introduced in `ApiFeatureOptions`, then the code would need to change to this anyway - maybe cross that bridge when it comes to it.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
ferdeen: Ar okay, thanks.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
ferdeen: Ar sorry I thought this was new.  If it exists then not to worry.

src/Stratis.Bitcoin.Features.Api/ApiFeature.cs
```diff
@@ -29,40 +30,44 @@ public sealed class ApiFeature : FullNodeFeature
...
-            this.webHost = Program.Initialize(this.fullNodeBuilder.Services, this.fullNode, this.apiSettings);
+            this.webHost = Program.Initialize(this.fullNodeBuilder.Services, this.fullNode, this.apiSettings, this.certificateStore);
+
+            if (this.apiSettings.KeepaliveTimer == null)
+                return;
 
             // Start the keepalive timer, if set.
             // If the timer expires, the node will shut down.
-            if (this.apiSettings.KeepaliveTimer != null)
+            this.apiSettings.KeepaliveTimer.Elapsed += (sender, args) =>
```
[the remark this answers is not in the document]
monsieurleberre: I don't understand your comment, do you think this changes the behaviour of the Initialize method ? to me this is just a little reduction in nesting.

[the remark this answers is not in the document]
Aprogiena: Yes, it is totally unrelated to this PR. It is just making a code look better per your personal preference. So someone else will come later and change it back because their preference will be different from yours. It makes nonsense, it doesn't belong to this PR, why don't you just focus on the issue?

Aprogiena:
Is this still in progress? Will you remove the password things?

monsieurleberre:
So the way this works now
- there is a flag "usehttps" that one can pass as startup argument, it is false by default and no certificate is then required
- if the usehttps flag is used, http:// doesn't work anymore and https:// is used instead

When Https is activated
- by default the ApiFeature will look for the certificate file at StratisNode/stratis/_Network_/StratisApi.pfx, if it doesn't exist it will create a new one and offer the option to secure it with a password. Users will then have to manually add it to the store of their OS if they want to remove browser warnings.
- alternatively, users can add an existing/trusted certificate file in the StratisNode/stratis/_Network_/ to use with https and pass in the argument "certificatefilename" if their certificate isn't the default StratisApi.pfx. A password will only be asked if needed.

I have tested on Windows and MacOs, it seems to work as expected

bokobza:
I think in the interest of simplicity, we should 
- not create a certificate for the user as it's not the node's business to do so (maybe a UI wrapping the node could do that in the future). If someone wants to use a certificate, they should create it and put it in the right folder and the store.
- not prompt for a password and not use a password. It works very well and it's cool but in practice a password can't help the user as an attacker can just use http and the protection given by a password is voided.

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -127,12 +146,16 @@ public static void PrintHelp(Network network)
...
+            builder.AppendLine($"#URI to node's API interface. Defaults to '{ DefaultApiHost }'.");
             builder.AppendLine($"#apiuri={ DefaultApiHost }");
-            builder.AppendLine($"#Port of node's API interface. Defaults to { GetDefaultPort(network) }");
+            builder.AppendLine($"#Port of node's API interface. Defaults to { GetDefaultPort(network) }.");
             builder.AppendLine($"#apiport={ GetDefaultPort(network) }");
-            builder.AppendLine($"#Keep Alive interval (set in seconds). Default: 0 (no keep alive)");
+            builder.AppendLine($"#Keep Alive interval (set in seconds). Default: 0 (no keep alive).");
             builder.AppendLine($"#keepalive=0");
+            builder.AppendLine($"#Use HTTPS protocol on the API. Default is { false }.");
+            builder.AppendLine($"#usehttps={DefaultCertificateFileName}");
```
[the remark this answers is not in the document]
monsieurleberre: absolutely!

src/Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisD": {
-      "commandName": "Project"
+      "commandName": "Project",
```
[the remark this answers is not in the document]
monsieurleberre: yes

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -71,7 +67,9 @@ public ApiSettings(NodeSettings nodeSettings)
             TextFileConfiguration config = nodeSettings.ConfigReader;
 
             this.UseHttps = config.GetOrDefault("usehttps", false);
-            this.HttpsCertificateFileName = config.GetOrDefault("certificatefilename", DefaultCertificateFileName);
+            this.HttpsCertificateFilePath = config.GetOrDefault("certificatefilepath", (string)null);
+            if(this.UseHttps && string.IsNullOrWhiteSpace(this.HttpsCertificateFilePath))
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -153,9 +151,9 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#Keep Alive interval (set in seconds). Default: 0 (no keep alive).");
             builder.AppendLine($"#keepalive=0");
             builder.AppendLine($"#Use HTTPS protocol on the API. Default is { false }.");
-            builder.AppendLine($"#usehttps={DefaultCertificateFileName}");
+            builder.AppendLine($"#usehttps={ false }");
```
Aprogiena: Maybe we don't need interpolated hardcoded value and just use the value as we do for example as we do on line 152

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -153,9 +151,9 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
...
             builder.AppendLine($"#Subject name of the certificate to use for https traffic encryption.");
-            builder.AppendLine($"#certificatefilename={DefaultCertificateFileName}");
+            builder.AppendLine($"#certificatefilepath=");
```
Aprogiena: Here we should totally describe what kind of Certificate do we expect. For example, which formats do we support – i.e. which extensions. We should also say that the certification should not be password protected

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -40,6 +40,12 @@ public class ApiSettings
         /// <summary>URI to node's API interface.</summary>
         public Timer KeepaliveTimer { get; private set; }
 
+        /// <summary>The Https certificate file name.</summary>
```
Aprogiena: Capital HTTPS, just as below

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
```diff
@@ -60,7 +66,16 @@ public ApiSettings(NodeSettings nodeSettings)
...
             TextFileConfiguration config = nodeSettings.ConfigReader;
 
-            string apiHost = config.GetOrDefault("apiuri", DefaultApiHost, this.logger);
+            this.UseHttps = config.GetOrDefault("usehttps", false);
+            this.HttpsCertificateFilePath = config.GetOrDefault("certificatefilepath", (string)null);
+            if(this.UseHttps && string.IsNullOrWhiteSpace(this.HttpsCertificateFilePath))
+                throw new ConfigurationException("The path to a certificate needs to be provided when using https. Please use the argument 'certificatefilepath' to provide it.");
+
+            var defaultApiHost = this.UseHttps 
+                                     ? DefaultApiHost.Replace(@"http://", @"https://")
```
Aprogiena: Just use normal indentation of four spaces

src/Stratis.Bitcoin.Features.Api/CertificateStore.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public bool TryGet(string filePath, out X509Certificate2 certificate)
+        {
+            certificate = null;
+
+            if (!File.Exists(filePath))
+                return false;
+
+            var fileInBytes = File.ReadAllBytes(filePath);
+            certificate = new X509Certificate2(fileInBytes);
```
Aprogiena: Both of these lines can throw in which case I would expect this method to return false. But you don't have any comments anywhere, so I have no idea what your intention was

src/Stratis.Bitcoin.Features.Api/Startup.cs
```diff
@@ -27,7 +27,7 @@ public Startup(IHostingEnvironment env)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
-            // Add service and create Policy to allow Cross-Origin Requests
+            // Save service and create Policy to allow Cross-Origin Requests
```
Aprogiena: It looks like that the old comment made better sense, what is the rationale?

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -9,18 +11,26 @@ namespace Stratis.Bitcoin.Features.Api
...
+            X509Certificate2 certificate = apiSettings.UseHttps ? GetHttpsCertificate(apiSettings, store) : null;
+
+            webHostBuilder = webHostBuilder ?? new WebHostBuilder();
+
+            webHostBuilder
+                .UseKestrel(options =>
+                    {
+                        if (!apiSettings.UseHttps) return;
+                        options.Listen(
+                            IPAddress.Loopback,
```
Aprogiena: Is this correct? Only loopback? I would actually expect all the normal interfaces the node is listening on because it actually doesn't make sense on the loopback, it only makes sense on the other interfaces. There are Bind addresses in the configuration – we should listen on those if they are provided. If they are not provided, we should listen on all interfaces

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -36,7 +46,8 @@ public static IWebHost Initialize(IEnumerable<ServiceDescriptor> services, FullN
                     foreach (ServiceDescriptor service in services)
                     {
                         object obj = fullNode.Services.ServiceProvider.GetService(service.ServiceType);
-                        if (obj != null && service.Lifetime == ServiceLifetime.Singleton && service.ImplementationInstance == null)
+                        if (obj != null && service.Lifetime == ServiceLifetime.Singleton &&
```
Aprogiena: Brackets

src/Stratis.Bitcoin.Features.Api/Startup.cs
monsieurleberre: Sorry this change was not intended, fat fingers replace

src/Stratis.Bitcoin.Features.Api/CertificateStore.cs
monsieurleberre: Intention is to throw, I initially had a try catch rethrow, but then thought it was useless

Aprogiena: But you throw here https://github.com/stratisproject/StratisBitcoinFullNode/pull/1897/files/74a135a28a70b125505334abfbd8dd0811ba3f45#diff-5257e1acdaea875a7a2d2c6169336535R77

Basically it is quite hard to use this method GetHttpsCertificate bcause you don't declare which exception can be thrown and it looks like you ony throw FileLoadException, but you actually can throw many other exceptions. By using try prefix it looks like it should not throw and it does.

src/Stratis.Bitcoin.Features.Api/Program.cs
monsieurleberre: Yes correct, that was definitely an issue
Looking in there I have also seen a lot of options that could make sense for us too:
options.Limits.MaxConcurrentConnections, options.Limits.MaxRequestBodySize
Might be worth exploring...

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -46,12 +57,24 @@ public static IWebHost Initialize(IEnumerable<ServiceDescriptor> services, FullN
...
-            host.Start();
+            var host = webHostBuilder.Build();
+                
 
+            host.Start();
+           
             return host;
         }
+
+        private static X509Certificate2 GetHttpsCertificate(ApiSettings apiSettings, ICertificateStore store)
```
bokobza: I think you should pass `apiSettings.HttpsCertificateFilePath` as a parameter instead of `apiSettings`, it'll make the method signature clearer.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowApiRunner.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using NBitcoin;
+using System.Security.Cryptography.X509Certificates;
```
bokobza: is this still needed?

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisPosApiRunner.cs
```diff
@@ -1,4 +1,5 @@
-﻿using NBitcoin;
+﻿using System.Security.Cryptography.X509Certificates;
```
bokobza: is this still needed?

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -9,18 +11,26 @@ namespace Stratis.Bitcoin.Features.Api
...
-                .UseKestrel()
+            X509Certificate2 certificate = apiSettings.UseHttps ? GetHttpsCertificate(apiSettings, store) : null;
+
+            webHostBuilder = webHostBuilder ?? new WebHostBuilder();
+
+            webHostBuilder
+                .UseKestrel(options =>
+                    {
+                        if (!apiSettings.UseHttps) return;
+                        options.Listen(
```
bokobza: please add a new line after `if (!apiSettings.UseHttps) return;`, I find it a bit too crammed like that, don't you?

src/Stratis.Bitcoin.Features.Api/Program.cs
bokobza: @monsieurleberre we don't have a maximum line length limit. I think up to 200 is acceptable.

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -9,18 +11,26 @@ namespace Stratis.Bitcoin.Features.Api
...
+                            IPAddress.Any,
```
Aprogiena: This is better, but it still is not respecting bind addresses

src/Stratis.Bitcoin.Features.Api/CertificateStore.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            try
+            {
+                var fileInBytes = File.ReadAllBytes(filePath);
+                certificate = new X509Certificate2(fileInBytes);
+                return true;
+            }
+            catch (Exception e)
+            {
+                this.logger.LogWarning(e, "Failed to read certificate at {0}", filePath);
```
Aprogiena: Better not to introduce logging if you are not fluent with it.We would just argue in many iterations how it should be properly done

src/Stratis.Bitcoin.Features.Api/Program.cs
monsieurleberre: @Aprogiena Thanks for feedback, wondering if you think this is better ?

```
            webHostBuilder
                .UseKestrel(options =>
                    {
                        if (!apiSettings.UseHttps) return;
                        Action<ListenOptions> configureListener = listenOptions => { listenOptions.UseHttps(certificate); };
                        var ipAddresses = Dns.GetHostAddresses(apiSettings.ApiUri.DnsSafeHost);
                        foreach (var ipAddress in ipAddresses)
                        {
                            options.Listen(ipAddress, apiSettings.ApiPort, listenOptions => listenOptions.UseHttps(certificate));
                        }
                    })
```

Or are you thinking of getting the "bind" addresses from the ConnectionManagerSettings, probably something like that
```
this.Listen.AddRange(config.GetAll("bind")
        .Select(c => new NodeServerEndpoint(c.ToIPEndPoint(port), false)));
```

 and then do the listen on them ?

Aprogiena: I'm not familiar with Kestrel, so I cannot advise on the exact implementation, but i meant `this.Listen` On connection manager settings. i don't understand `var ipAddresses = Dns.GetHostAddresses(apiSettings.ApiUri.DnsSafeHost);`  - Why would we need that here

src/Stratis.Bitcoin.Features.Api/Program.cs
```diff
@@ -9,18 +11,26 @@ namespace Stratis.Bitcoin.Features.Api
 {
     public class Program
     {
-        public static void Main(string[] args)
-        {
-        }
-
-        public static IWebHost Initialize(IEnumerable<ServiceDescriptor> services, FullNode fullNode, ApiSettings apiSettings)
+        public static IWebHost Initialize(IEnumerable<ServiceDescriptor> services, FullNode fullNode,
+            ApiSettings apiSettings, ICertificateStore store, IWebHostBuilder webHostBuilder = null)
```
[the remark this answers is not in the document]
monsieurleberre: I'll remove this, we wanted to be able to unit test that the webhost builder got configured properly (if https, which certificate, which endpoints to listen to, etc.), but as discussed, we won't

src/Stratis.Bitcoin.Features.Api/Program.cs
monsieurleberre: 👍

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowApiRunner.cs
monsieurleberre: no, thanks

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisPosApiRunner.cs
monsieurleberre: no, thanks

src/Stratis.Bitcoin.Features.Api/Program.cs
monsieurleberre: for me it is better in one line, this is really "read and forget", but I think most people are like you. Changed.

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
monsieurleberre: added a ref to the appropriate MSDN doc, and mentioned password are not supported, that should do it hopefully.

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
monsieurleberre: done

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
monsieurleberre: ok...

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
monsieurleberre: ok...

src/Stratis.Bitcoin.Features.Api/ApiSettings.cs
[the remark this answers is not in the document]
monsieurleberre: yes

monsieurleberre: yes

src/Stratis.Bitcoin.Api.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        private X509Certificate2 certificateRetrieved;
+
+        public ProgramTest()
+        {
+            this.apiSettings = new ApiSettings { UseHttps = true };
+            this.certificateToUse = new X509Certificate2();
+            this.certificateStore = Substitute.For<ICertificateStore>();
+            this.webHostBuilder = Substitute.For<IWebHostBuilder>();
+        }
+
```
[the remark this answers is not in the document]
monsieurleberre: done

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowApiRunner.cs
bokobza: anytime

monsieurleberre:
image

monsieurleberre:
testing on MacOs revealed this issue 
https://github.com/dotnet/corefx/issues/31328
should I put back my changes to allow users to input password and try again ?

PR #1899, opened 2018-08-10
[Aprogina] fixed some more tests

noescape00:
Apparently tests don't use ==. Assert.Equal uses default.Equal
So This one fixes around 5 tests


noescape00:
Reviewed and approved by @Aprogiena on skype call

Issue #960, opened 2017-12-20
[Peer] MyVersion.AddressFrom is not initialized properly


Aprogiena:
There is currently no mechanism to detect our external IP address. According to https://en.bitcoin.it/wiki/Protocol_documentation#version and StratisX code, version.addr_from should contain external IP address where the node can be connected. 

It is possible to get our own IP address from version message sent to us from another node. So if we don't have it set, we should look in version.addr_recv and set it.

Current it is the network peer server that holds the IP address, but it should not. Other component should care of it and network peer server should always ask about the latest value - in case it will be initialized later.

Aprogiena:
When we start, we should
- check if externalIp is set, if so, this should be the address regardless of anything else and it should not change - make immutable
- if externalIp is not set, we should look at bind addresses - each node server may have different address, but these can be local addresses. so we should initialize external addresses with those values but then either 1) check if they are local or not, if not make them immutable, if they are local, make them mutable 2) OR always make them mutable
- if we make outbound connection, we use the best value we have (externalIp -> any non local bind -> whatever)
- if we receive version from a peer and the external address is mutable, we should update it if it does not match

maybe we should also prevent some kind of attacks from other peers, so maybe we should only consider version information from peers that we connect to, not those that connect to us

thefazzer:
From @Aprogiena and @noescape00:

We think it should be implemented in SelfEndpointTracker
Interface should be extended.

There should be a property `MyExternalAddres` and it will be constant in case we initialized the node with ExternalIp set. In case it wasn't specified this property is null unless many peers send us the same address claiming it's ours (only outbound peers).
there will be score for the address if it wasnt set by external IP value. If we use first routeble bind address we set MyExternalEddress to that address and the score to 10. Whenever we receive message version message from peer that we've connected to (the outbound peer) we will call method on this self endpoint tracker and this method will set my external address to the address that peer sent us in the version message in case my external address was null .

If it wasn't null and was the same as value that was there we just increment the score by 1
If it was different we decrement the score by 1
if the new score is 0 we relace the old one with the new one with score 1

Issue #1816, opened 2018-08-02
[PeerAddressManager] Reset counter to 0 after Threshold expired


thefazzer:
- Handshake counter does not get reset to zero until a successful handshake occurs.
- Handshake counter should get reset after threshold time expires.

bokobza:
Fixed with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1827

PR #1900, opened 2018-08-10
MyVersion.AddressFrom is not initialized properly (#1816).

thefazzer:
- If external IP address supplied this overrides all.
- If external IP address not supplied take first routeable bind address and set score to 10.
- If none supplied or routableable take from version handshake.
- If it was the same as value that was there we just increment the score by 1.
- If it was different we decrement the score by 1.
- If the new score is 0 we replace the old one with the new one with score 1.

Preferred: @Aprogiena @noescape00


src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +33,48 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <summary>Updates external IP address and peer score of the node.</summary>
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, ConnectionManager connectionManager)
```
ferdeen: Do the parameters need guards ?

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +33,48 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+            IPEndPoint nodeServerEndpoint = connectionManager.ConnectionSettings.Listen?.FirstOrDefault(x => x.Endpoint.Address.IsRoutable(false))?.Endpoint;
+            if (nodeServerEndpoint != null)
+            {
+                this.MyExternalAddress = new KeyValuePair<IPEndPoint, int>(nodeServerEndpoint, 10);
+                connectionManager.ConnectionSettings.ExternalEndpoint = this.MyExternalAddress.Key;
+                return;
+            }
+
+            // If none supplied or routableable take from version handshake.
+            if (ipEndPoint == null)
```
ferdeen: Could this happen earlier in the method, so you can return out early ?  Or do you need to setup connectionManager.ConnectionSettings.ExternalEndpoint first ?

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +33,48 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+            // If external IP address not supplied take first routeable bind address and set score to 10.
+            IPEndPoint nodeServerEndpoint = connectionManager.ConnectionSettings.Listen?.FirstOrDefault(x => x.Endpoint.Address.IsRoutable(false))?.Endpoint;
+            if (nodeServerEndpoint != null)
+            {
+                this.MyExternalAddress = new KeyValuePair<IPEndPoint, int>(nodeServerEndpoint, 10);
+                connectionManager.ConnectionSettings.ExternalEndpoint = this.MyExternalAddress.Key;
+                return;
+            }
+
+            // If none supplied or routableable take from version handshake.
```
ferdeen: Are routableable (and routeable on line 47) actual words ?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,8 @@ public void Initialize()
 
             this.StartNodeServer();
 
+            this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(null, this);
```
ferdeen: Can the parameter you are supplying as null be an optional parameter in UpdateAndAssignMyExternalAddress ?

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -1,5 +1,13 @@
...
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
 using ConcurrentCollections;
+using NBitcoin;
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Configuration.Settings;
+using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Utilities.Extensions;
```
ferdeen: using Stratis.Bitcoin.Configuration;
using Stratis.Bitcoin.Configuration.Settings;
using Stratis.Bitcoin.Utilities.Extensions;

Directives are not used in this class.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
ferdeen: I am struggling to understand what this method is doing when the summary says it's updating external IP addresses and working out the Peer score of the node.

ferdeen: I can see two actions in this method - SetExternalAddress and SetExternalEndPoint.  Maybe move these into separate methods.  I'm not 100% sure if this is required, but something like this - (I haven't run this) ;

```
/// <summary>Updates external IP address and peer score of the node.</summary>
public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, ConnectionManager connectionManager)
{
    // If none supplied or able to route, take from version handshake.
    if (ipEndPoint == null)
        return;

    if (this.SetExternalEndpoint(connectionManager)) return;

    connectionManager.ConnectionSettings.ExternalEndpoint = this.SetExternalAddress(ipEndPoint).Key;
}

private KeyValuePair<IPEndPoint, int> SetExternalAddress(IPEndPoint ipEndPoint)
{
    // If it was the same as value that was there we just increment the score by 1.
    if (ipEndPoint.Equals(this.MyExternalAddress.Key))
    {
        this.MyExternalAddress =
            new KeyValuePair<IPEndPoint, int>(this.MyExternalAddress.Key, this.MyExternalAddress.Value + 1);
    }

    // If it was different we decrement the score by 1.
    this.MyExternalAddress =new KeyValuePair<IPEndPoint, int>(this.MyExternalAddress.Key, this.MyExternalAddress.Value - 1);

    // If the new score is 0 we replace the old one with the new one with score 1.
    return this.MyExternalAddress.Value <= 0 ? new KeyValuePair<IPEndPoint, int>(ipEndPoint, 1) : default(KeyValuePair<IPEndPoint, int>);
}

private bool SetExternalEndpoint(IConnectionManager connectionManager)
{
    // if external IP address supplied this overrides all.
    if (connectionManager.ConnectionSettings.ExternalIpFromNodeConfiguration != null)
    {
        connectionManager.ConnectionSettings.ExternalEndpoint =
            connectionManager.ConnectionSettings.ExternalIpFromNodeConfiguration;
        return true;
    }

    // If external IP address not supplied take first routeable bind address and set score to 10.
    IPEndPoint nodeServerEndpoint = connectionManager.ConnectionSettings.Listen
        ?.FirstOrDefault(x => x.Endpoint.Address.IsRoutable(false))?.Endpoint;

    if (nodeServerEndpoint == null) return false;

    this.MyExternalAddress = new KeyValuePair<IPEndPoint, int>(nodeServerEndpoint, 10);
    connectionManager.ConnectionSettings.ExternalEndpoint = this.MyExternalAddress.Key;

    return true;
}
```

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -408,6 +410,9 @@ public async Task<INetworkPeer> ConnectAsync(IPEndPoint ipEndpoint)
             {
                 this.peerAddressManager.PeerAttempted(ipEndpoint, this.dateTimeProvider.GetUtcNow());
                 await peer.VersionHandshakeAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
+
+                // Set external address in self endpoint tracker to peer in version payload.
+                this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(peer.PeerVersion.AddressFrom, this);
```
Aprogiena: This is not the right place, let's just have a call about improvements on this

Aprogiena:
Please contact me on this one when you want to make progress

thefazzer:
All updates as discussed.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +19,15 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
         private readonly ConcurrentHashSet<IPEndPoint> knownSelfEndpoints = new ConcurrentHashSet<IPEndPoint>();
 
+        /// <summary>External IP address of the node.</summary>
+        public IPEndPoint MyExternalAddress { get; set; }
+
+        /// <summary>Whether IP address of the node is final or can be updated.</summary>
+        public bool IsMyExternalAddressFinal { get; set; } = false;
+
+        /// <summary>Peer score of external IP address of the node.</summary>
+        public int MyExternalAddressPeerScore { get; set; }
```
mikedennis: Cool you fixed most of the issues I was going to comment on. I like this much better than the KeyValuePair.  Only small thing I noticed is I think ```MyExternalAddressPeerScore``` can be made private as I don't think it's referenced outside of this class.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,22 @@ public void Initialize()
 
             this.StartNodeServer();
 
+            // if external IP address supplied this overrides all.
```
Aprogiena: Capital letter please at the start of the sentence

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,22 @@ public void Initialize()
...
+            if (this.ConnectionSettings.ExternalEndpoint != null)
+            {
+                this.selfEndpointTracker.MyExternalAddress = this.ConnectionSettings.ExternalEndpoint;
+                this.selfEndpointTracker.IsMyExternalAddressFinal = true;
```
Aprogiena: This logic is correct, but I would actually use the update method even here instead of this direct modification of properties that belong to different components. You basically just need a new parameter in the update method saying that it is final.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -408,6 +424,9 @@ public async Task<INetworkPeer> ConnectAsync(IPEndPoint ipEndpoint)
             {
                 this.peerAddressManager.PeerAttempted(ipEndpoint, this.dateTimeProvider.GetUtcNow());
                 await peer.VersionHandshakeAsync(this.nodeLifetime.ApplicationStopping).ConfigureAwait(false);
+
+                // Set external address in self endpoint tracker to peer in version payload.
+                this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(peer.PeerVersion.AddressFrom, null);
```
Aprogiena: We actually said, that we want this to be hard of VersionHandshakeAsync

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,15 @@ namespace Stratis.Bitcoin.P2P
...
     public interface ISelfEndpointTracker
     {
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, int? ipEndPointPeerScore);
+
+        /// <summary>External IP address of the node.</summary>
+        IPEndPoint MyExternalAddress { get; set; }
+
+        /// <summary>Whether external IP address of the node is final or can be updated.</summary>
+        bool IsMyExternalAddressFinal { get; set; }
```
Aprogiena: This guy should not be an interface if we update everything through update method and the guy above should only have public getter

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,15 @@ namespace Stratis.Bitcoin.P2P
     /// </summary>
     public interface ISelfEndpointTracker
     {
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, int? ipEndPointPeerScore);
```
Aprogiena: We don't need the second argument nullable. It would be better if it has a default value of one. Also please add comments for parameters

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerConnectionParameters.cs
```diff
@@ -98,7 +98,7 @@ public VersionPayload CreateVersion(IPEndPoint peerAddress, Network network, Dat
                 Version = this.Version,
                 Timestamp = timeStamp,
                 AddressReceiver = peerAddress,
-                AddressFrom = this.AddressFrom ?? new IPEndPoint(IPAddress.Parse("0.0.0.0").MapToIPv6Ex(), network.DefaultPort),
+                AddressFrom = externalAddressEndPoint ?? new IPEndPoint(IPAddress.Parse("0.0.0.0").MapToIPv6Ex(), network.DefaultPort),
```
Aprogiena: We said that we will guarantee that it may not be null

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,22 @@ public void Initialize()
...
+                this.selfEndpointTracker.IsMyExternalAddressFinal = true;
+            }
+            else
+            {
+                // If external IP address not supplied take first routeable bind address and set score to 10.
+                IPEndPoint nodeServerEndpoint = this.ConnectionSettings.Listen?.FirstOrDefault(x => x.Endpoint.Address.IsRoutable(false))?.Endpoint;
+                if (nodeServerEndpoint != null)
+                {
+                    this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(nodeServerEndpoint, 10);
+                }
```
Aprogiena: We said we want else branch here which will initialise with endpoint `0.0.0.0:correctport`

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +19,15 @@ public class SelfEndpointTracker : ISelfEndpointTracker
         /// <summary>Hashset to hold the endpoints currently known to be itself.</summary>
         private readonly ConcurrentHashSet<IPEndPoint> knownSelfEndpoints = new ConcurrentHashSet<IPEndPoint>();
 
+        /// <summary>External IP address of the node.</summary>
```
Aprogiena: Let's also introduced a logger – so have a constructor that accepts logger factory. It also move all the initialisation to constructor

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +39,41 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <summary>Updates external IP address and peer score of the node.</summary>
```
Aprogiena: As it is on interface, you just use `<inheritdoc/>`

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +39,41 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, int? ipEndPointScore = null)
+        {
+            if (this.IsMyExternalAddressFinal)
+               return;
+
+            if (ipEndPoint == null)
+                return;
+
+            // Explicitly supplied a score for this endpoint.
+            if (ipEndPointScore.HasValue)
```
Aprogiena: I would remove this whole if block, we don't need that logic

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +39,41 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <summary>Updates external IP address and peer score of the node.</summary>
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, int? ipEndPointScore = null)
+        {
```
Aprogiena: And we just introduce logs into this method

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerConnectionParameters.cs
thefazzer: I didn't understand this and what correction I needed to make.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
thefazzer: Done but I didn't alter all the calling code with logmanagers as wanted to confirm before affecting a lot of files.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
thefazzer: Didn't understand this comment so will wait for a clarification.

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerConnectionParameters.cs
Aprogiena: We wanted to guarantee that it is never null, therefore we don't need `??` And the right side

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
Aprogiena: What I mean is we should remove these three lines because the we handle it elsewhere

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,25 @@ public void Initialize()
...
+            {
+                // If external IP address not supplied take first routeable bind address and set score to 10.
+                IPEndPoint nodeServerEndpoint = this.ConnectionSettings.Listen?.FirstOrDefault(x => x.Endpoint.Address.IsRoutable(false))?.Endpoint;
+                if (nodeServerEndpoint != null)
+                {
+                    this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(nodeServerEndpoint, false, 10);
+                }
+                else
+                {
+                    this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(new IPEndPoint(IPAddress.Parse("0.0.0.0").MapToIPv6Ex(), this.Network.DefaultPort), false);
```
Aprogiena: Now I believe we have here the same mistake as it was in the original code. We must used correct port, not the default one.

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,18 @@ namespace Stratis.Bitcoin.P2P
...
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        /// <param name="ipEndPoint">The endpoint to add.</param>
+        /// <param name="suppliedEndpointIsFinal">Whether the ipEndPoint supplied should be marked final on the endpoint tracker.</param>
+        /// <param name="ipEndPointPeerScore">Peer score of the ipEndPoint supplied.  Default value of 1.</param>
+        void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, bool suppliedEndpointIsFinal, int ipEndPointPeerScore = 0);
+
+        /// <summary>External IP address of the node.</summary>
+        IPEndPoint MyExternalAddress { get; }
+
+        /// <summary>Peer score of external IP address of the node.</summary>
```
Aprogiena: Now we can remove this one from the interface completely and it will become just a private property of the self endpoint tracker implementation

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,18 @@ namespace Stratis.Bitcoin.P2P
     /// </summary>
     public interface ISelfEndpointTracker
     {
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        /// <param name="ipEndPoint">The endpoint to add.</param>
+        /// <param name="suppliedEndpointIsFinal">Whether the ipEndPoint supplied should be marked final on the endpoint tracker.</param>
+        /// <param name="ipEndPointPeerScore">Peer score of the ipEndPoint supplied.  Default value of 1.</param>
+        void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, bool suppliedEndpointIsFinal, int ipEndPointPeerScore = 0);
```
Aprogiena: The default value should be one, not zero

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,18 @@ namespace Stratis.Bitcoin.P2P
     /// </summary>
     public interface ISelfEndpointTracker
     {
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        /// <param name="ipEndPoint">The endpoint to add.</param>
+        /// <param name="suppliedEndpointIsFinal">Whether the ipEndPoint supplied should be marked final on the endpoint tracker.</param>
+        /// <param name="ipEndPointPeerScore">Peer score of the ipEndPoint supplied.  Default value of 1.</param>
```
Aprogiena: use `<paramref>` and no double spacing

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
Aprogiena: yes

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +20,35 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
         private readonly ConcurrentHashSet<IPEndPoint> knownSelfEndpoints = new ConcurrentHashSet<IPEndPoint>();
 
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        /// <summary>External IP address of the node.</summary>
+        public IPEndPoint MyExternalAddress { get; set; }
+
+        /// <summary>Whether IP address of the node is final or can be updated.</summary>
+        public bool IsMyExternalAddressFinal { get; set; } = false;
```
Aprogiena: Make it is private field. Also remove the initialisation or move it to constructor

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
Aprogiena: This should also be private field

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +20,35 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
+
+        /// <summary>Peer score of external IP address of the node.</summary>
+        public int MyExternalAddressPeerScore { get; set; }
+
+        /// <summary>
+        /// Initializes an instance of the self endpoint tracker.
+        /// </summary>
+        public SelfEndpointTracker() : this(new LoggerFactory())
```
Aprogiena: You don't want this constructor. It would be only used by tests and for those they can call it with the correct log factory anyway

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +20,35 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
+        }
+
+        /// <summary>
+        /// Initializes an instance of the self endpoint tracker.
+        /// </summary>
+        /// <param name="loggerFactory">Factory for creating loggers.</param>
+        public SelfEndpointTracker(ILoggerFactory loggerFactory)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.logger.LogTrace("()");
```
Aprogiena: Remove this line

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <inheritdoc/>
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, bool suppliedEndpointIsFinal, int ipEndPointScore = 0)
+        {
+            if (ipEndPoint == null)
```
Aprogiena: entry log here:


```
this.logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(ipEndPoint), ipEndPoint, nameof(suppliedEndpointIsFinal), suppliedEndpointIsFinal, nameof(ipEndPointScore), ipEndPointScore);
```

Aprogiena: and remove this IF completely, we don't need it

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+            {
+                this.logger.LogTrace("(IP ENDPOINT IS NULL)");
+                return;
+            }
+
+            // override peer score if supplied
+            if (ipEndPointScore != 0)
```
Aprogiena: also remove this if, we don't want it

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <inheritdoc/>
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, bool suppliedEndpointIsFinal, int ipEndPointScore = 0)
```
Aprogiena: suppliedEndpointIsFinal -> suppliedEndPointIsFinal

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+                this.MyExternalAddressPeerScore = ipEndPointScore;
+                this.logger.LogTrace("(Peer score is non default for {0}.  Setting to {1}.)", ipEndPoint.ToString(), ipEndPointScore);
+                return;
+            }
+
+            if (suppliedEndpointIsFinal)
+            {
+                this.MyExternalAddress = ipEndPoint;
+                this.IsMyExternalAddressFinal = true;
+                this.logger.LogTrace("(Supplied endpoint {0} is final)", ipEndPoint.ToString());
```
Aprogiena: just exit log instead of this one
```
this.logger.LogTrace("(-)[FINAL]");
```

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+                return;
+            }
+
+            if (this.IsMyExternalAddressFinal)
+            {
+                this.logger.LogTrace("(Found an existing final endpoint {0})", this.MyExternalAddress.ToString());
```
Aprogiena: just exit log
```this.logger.LogTrace("(-)[EXISTING_FINAL]");```

File: src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs (hunk 29 above)
Comment (Aprogiena) [inline src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs]: just exit log
```this.logger.LogTrace("(-)[SUPPLIED_FINAL]");```

File: src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs [hunk 31]
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+                return;
+            }
+
+            // If it was the same as value that was there we just increment the score by 1.
+            if (ipEndPoint.Equals(this.MyExternalAddress))
+            {
+                this.MyExternalAddressPeerScore += 1;
+                this.logger.LogTrace("(Same endpoint {0} supplied.  Score incremented to {1})", ipEndPoint, this.MyExternalAddressPeerScore);
```
Aprogiena:
```
this.logger.LogTrace("Same endpoint '{0}' supplied. Score incremented to {1}.", ipEndPoint, this.MyExternalAddressPeerScore);
```

similarly below

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +60,59 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+            }
+
+            // If the new score is 0 we replace the old one with the new one with score 1.
+            if (this.MyExternalAddressPeerScore <= 0)
+            {
+                this.logger.LogTrace("(Score for old endpoint {0} is <= 0. Updating endpoint to {1} and resetting peer score to 1)", this.MyExternalAddress, ipEndPoint);
+                this.MyExternalAddress = ipEndPoint;
+                this.MyExternalAddressPeerScore = 1;
+            }
+        }
```
Aprogiena: exit log here

```this.logger.LogTrace("(-)");```

File: src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs [hunk 33]
```diff
@@ -11,6 +20,35 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
+
+        /// <summary>External IP address of the node.</summary>
+        public IPEndPoint MyExternalAddress { get; set; }
+
+        /// <summary>Whether IP address of the node is final or can be updated.</summary>
+        public bool IsMyExternalAddressFinal { get; set; } = false;
+
+        /// <summary>Peer score of external IP address of the node.</summary>
+        public int MyExternalAddressPeerScore { get; set; }
+
```
Aprogiena: You also need to introduce a lock object and protect the score as well as the external address and also IsMyExternalAddressFinal  With it.

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +20,35 @@ public class SelfEndpointTracker : ISelfEndpointTracker
         /// <summary>Hashset to hold the endpoints currently known to be itself.</summary>
         private readonly ConcurrentHashSet<IPEndPoint> knownSelfEndpoints = new ConcurrentHashSet<IPEndPoint>();
 
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        /// <summary>External IP address of the node.</summary>
+        public IPEndPoint MyExternalAddress { get; set; }
```
Aprogiena: Make the setter private

thefazzer:
-Synchronize whole UpdateAndAssignMyExternalAddress method 
-Remove redundant  if (ipEndPointScore != 0) block
-Fix logging syntax
-Fix connectionsettings port logic
-Fix test logic

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -7,6 +11,15 @@ namespace Stratis.Bitcoin.P2P
     /// </summary>
     public interface ISelfEndpointTracker
     {
+        /// <summary>Update external IP address and peer score of the node.</summary>
+        /// <paramref name="ipEndPoint">The endpoint to add.</paramref>
+        /// <paramref name="suppliedEndPointIsFinal">Whether the ipEndPoint supplied should be marked final on the endpoint tracker.</paramref>
+        /// <paramref name="ipEndPointPeerScore">Peer score of the ipEndPoint supplied. Default value of 1.</paramref>
```
Aprogiena: misunderstanding here

the parameters should be done with `<param>` as you see elsewhere, but if inside of that description you refer to DIFFERENT parameter, you use `<paramref>` so for example here:

`<param name="ipEndPointPeerScore">Peer score of the <paramref name="ipEndPoint"/> supplied. Default value of 1.</paramref>`

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -11,6 +14,32 @@ public class SelfEndpointTracker : ISelfEndpointTracker
...
+        /// <summary>Peer score of external IP address of the node.</summary>
+        internal int MyExternalAddressPeerScore { get; set; }
+
+        /// <summary>Whether IP address of the node is final or can be updated.</summary>
+        private bool IsMyExternalAddressFinal { get; set; }
+
+        /// <summary>Protects access to <see cref="MyExternalAddress"/>, <see cref="MyExternalAddressPeerScore"/> and <see cref="IsMyExternalAddressFinal"/>.</summary>
+        private readonly object lockObject;
+
+        /// <summary>External IP address of the node.</summary>
```
Aprogiena: `<inheritdoc />` instead here

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +51,51 @@ public bool IsSelf(IPEndPoint ipEndPoint)
         {
             return this.knownSelfEndpoints.Contains(ipEndPoint);
         }
+
+        /// <inheritdoc/>
+        public void UpdateAndAssignMyExternalAddress(IPEndPoint ipEndPoint, bool suppliedEndPointIsFinal, int ipEndPointScore = 0)
+        {
+            lock (this.lockObject)
+            {
+                this.logger.LogTrace("({0}:'{1}',{2}:{3},{4}:{5})", nameof(ipEndPoint), ipEndPoint, nameof(suppliedEndPointIsFinal), suppliedEndPointIsFinal, nameof(ipEndPointScore), ipEndPointScore);
```
Aprogiena: entry and exit log should go outside of the lock, otherwise good

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +51,51 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+
+                // If the new score is 0 we replace the old one with the new one with score 1.
+                if (this.MyExternalAddressPeerScore <= 0)
+                {
+                    this.logger.LogTrace("Score for old endpoint '{0}' is <= 0. Updating endpoint to '{1}' and resetting peer score to 1.", this.MyExternalAddress, ipEndPoint);
+                    this.MyExternalAddress = ipEndPoint;
+                    this.MyExternalAddressPeerScore = 1;
+                }
+
+                this.logger.LogTrace("(-)");
```
Aprogiena: this one should also go outside of the lock

src/Stratis.Bitcoin/P2P/SelfEndpointTracker.cs
```diff
@@ -22,5 +51,51 @@ public bool IsSelf(IPEndPoint ipEndPoint)
...
+                    this.logger.LogTrace("(-)[EXISTING_FINAL]");
+                    return;
+                }
+
+                // If it was the same as value that was there we just increment the score by 1.
+                if (ipEndPoint.Equals(this.MyExternalAddress))
+                {
+                    this.MyExternalAddressPeerScore += 1;
+                    this.logger.LogTrace("(-)[SUPPLIED_EXISTING]");
+                }
```
Aprogiena: you are missing return here

thefazzer:
All changes made.

Aprogiena:
@noescape00 I'm happy with this one, could you please check and approve?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -141,6 +141,25 @@ public void Initialize()
...
             this.StartNodeServer();
 
+            // If external IP address supplied this overrides all.
+            if (this.ConnectionSettings.ExternalEndpoint != null)
+            {
+                this.selfEndpointTracker.UpdateAndAssignMyExternalAddress(this.ConnectionSettings.ExternalEndpoint, true);
+            }
+            else
+            {
+                // If external IP address not supplied take first routeable bind address and set score to 10.
```
noescape00: typo. `routable`

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
```diff
@@ -1,4 +1,8 @@
-﻿using System.Net;
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
```
noescape00: all usings except for this are redundant

noescape00:
Looks good. Just a few minor changes.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
thefazzer: Fixed.

src/Stratis.Bitcoin/P2P/ISelfEndpointTracker.cs
thefazzer: Removed.

noescape00:
@Aprogiena

bokobza:
Closes https://github.com/stratisproject/StratisBitcoinFullNode/issues/960

PR #1903, opened 2018-08-12
[Activation] Merge master into activation

fassadlr:


noescape00:
Looks good except for there are now:

```
SmartContractPowConsensusRuleEngine.cs~upstream_master
SmartContractConsensusRuleEngine.cs~HEAD
```

according to diff.

PR #1905, opened 2018-08-13
[SC] Closes #1699 constructor gas injection

### LINKED ISSUE #1699: Initialized readonly field in a smart contract causes null reference.
When creating the following contract, a NullReferenceException occurs as the contract tries to spendgas in a method that occurs before the constructor is executed (I believe).

```
public class Auction : SmartContract
{
	private readonly string test = "Slugman's test: ";

	public Auction(ISmartContractState smartContractState, ulong durationBlocks)
		: base(smartContractState)
	{
	}
}
```

@rowandh may have more to say about this.

rowandh:


codingupastorm:
So I can understand this, our assumption is that the constructor will always be the first CALL  @rowandh  ?

rowandh:
Yes, the IL for constructor

PR #1906, opened 2018-08-13
[SC] Consending txs use Time of 'parent' transaction

codingupastorm:


src/Stratis.SmartContracts.Core/State/AccountAbstractionLayer/TransactionCondenser.cs
```diff
@@ -86,7 +86,7 @@ public Transaction CreateCondensingTransaction()
         private Transaction BuildTransaction()
         {
             Transaction tx = this.network.CreateTransaction();
-            tx.Time = (uint)this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
```
fassadlr: I'm pretty sure that the DTP in this class is no longer being used  now, @codingupastorm can you remove it?

codingupastorm: Donezo!

Issue #1862, opened 2018-08-06
Try to isolate responsibilities from the ConsensusManager


monsieurleberre:
Part of #1736

This class is currently doing a few too many things, as a first cut I see 2 groups responsibilities that lay themselves well for separation

First group is
- download and request blocks

Second group something like
- validate chains / headers
- select tip candidates
- rewind/reconnect chains

#### Acceptance Criteria:
This is done when we have extracted a class and its interface, of which the sole purpose is to handle the download of the blocks, from the ConsensusManager

fassadlr:
@monsieurleberre I'm reopening this as I feel that `ConsensusManager` can still be looked at some point to extract some logic as it is a massive class sharing too many responsibilities.

fassadlr:
This is a continuation on the discussion in PR #1911.

@noescape00 My idea with the visitor pattern was to isolate parts of CM so that we can individually test them. A good example would be `BlockMined`.

@dangershony mentioned that the biggest role of CM is to ensure that locks are taken when changes are made to the CHT. So this is very good use case. If my understanding is correct, we don't want anything else apart from CM to be responsible for making changes to CHT because CM manages locks around it.

So how do we isolate code in CM that does specific things to CHT so that we can test them in isolation? We dont want to have to start a whole new instance of CM every time we want to have test, for example, `BlockMined`.

The test cases you wrote for `BlockMined` proves my point that they arent actually that complicated but somehow it appears to be difficult to isolate the logic from CM.

This is why I envisioned a scenario where you have isolated pieces of logic that is allowed to make changes to CHT and ONLY via CM's locks. So essentially CM becomes a lock manager class of sorts. This is why I thought that a class called BlockMined "visits" CM, makes changes to CHT internally and exits.

@Aprogiena raised a valid point of not wanting to make CHT public in CM. I agree with this, so how do we solve this problem? How can an isolated piece of logic makes changes to CHT internally, without exposing CHT to the outside?

If we can solve this, we future proof CM to be pluggable with isolated pieces of code that can make changes CHT. I.e. if in the future we have another scenario where have to affect CHT via CM, we can just build another class that we can pass to CM, it executes, and voila, you have made changes to CHT

Lets think of a scenarios where this can be possible! 👍

dangershony:
[...]

PR #1911, opened 2018-08-13
[Activation] Separate concerns in CM part #1

fassadlr:
Currently, consensus manager has too many responsibilities making it very difficult to unit test and maintain reliably going forward. 

Please note that with this work there were NO logic changes.

@monsieurleberre and myself want to split off some of the responsibilities from CM so that they can be separately tested and maintained. We have a big issue with potentially many devs working on this code and we want to ensure that we can create boundaries around parts of CM that can "componitized".

As part of this the block downloading logic was pulled into `ConsensusBlockDownloader` and the shared block loading logic into `ConsensusBlockLoader`. 

The individual methods on IConsensusManager has been extracted to individual visitor classes that "visits" consensus, makes changes to CHT and returns. This is a very basic visitor pattern implementation.

This PR will be split in 2 as part 2 further improves on the modularity and greatly improves he ease of writing unit tests for CM.


noescape00:
I'll review it tomorrow, don't merge. 

PS
Didn't we agree that refactoring will be done after the component is ready, best case when we finish with the activation? 
Why spending time now without even talking to anyone who was originally working on the component or anyone who is working on activation?

noescape00:
@fassadlr the visitor pattern is the pattern that typically solves the problem of double dispatching. 

It's purpose is to add common functionality to a group of objects without having to derive from the same class and\or modify it.

Here is a good example of where this pattern is needed: https://www.codeproject.com/Articles/588882/TheplusVisitorplusPatternplusExplained

In case of this PR every visitor works ALWAYS with just one type. It would make sense if for example we would have smth like this for block downloaded visitor:
```
AcceptBlock(PoSBlock)
AcceptBlock(CompactPoSBlock)
AcceptBlock(PoWBlock)
AcceptBlock(CompactPoWBlock)
```
but that's not the case. 

The pattern is being missuses and it's not justified here. 

Since the whole PR is about using visitor pattern I think it's safe to discard it.

fassadlr:
@noescape00 thank you for your constructive comments on this PR 💯 Lets continue the discussion here #1862 .

PR #1917, opened 2018-08-14
[SC] ByteHelper in new project and allowed in smart contracts

codingupastorm:
ByteHelper package and its tests should really be in their own repo and hosted on a public NuGet - I think that should be a separate task and this can go through our normal workflow, unless you guys think it should go straight to a new repo.


src/Stratis.SmartContracts.ByteHelper/ByteConverter.cs
```diff
@@ -0,0 +1,163 @@
+﻿using System;
+using System.Text;
+
+namespace Stratis.SmartContracts.ByteHelper
```
rowandh: Could we perhaps pick a name without the word 'helper' in it? I think @majikandy might have some thoughts around this.

codingupastorm:
We have discussed endianness and will be coming back to this.

PR #1918, opened 2018-08-14
[Activation] rules engine refactoring

noescape00:
Things in this PR: 

1) Separate consensus rule into 2 rule types- async and sync rules. 
2) Separate list of rules into 4 sets: Header, Integrity, Partial, Full validation rules. First 2 are sync, last 2 are async. 
3) Fill those 4 groups in the registration instead of the old way (old way was to fill them into 1 list and then separate into 4 lists based on rule type which is provided by attribute). Attributes were removed. 
4) Remove attribute for skipping the validation to have 1 way of skipping the validation (before we had 2 ways- one is to start executing the rule and the relu would decide on it's own based on context.CanSkipValidation and 2nd is the attribute which would preven rule execution at all. Now we have only 1st way)


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BitcoinActivationRule.cs
```diff
@@ -1,17 +1,15 @@
...
 
 namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
 {
     /// <summary>
-    /// Check that a <see cref="BitcoinMain"/> network block has the correct version according to the defined active deployments.
+    /// Check that a <see cref="BitcoinMain" /> network block has the correct version according to the defined active deployments.
     /// </summary>
-    [HeaderValidationRule]
-    public class BitcoinActivationRule : ConsensusRule
+    /// <remarks>This is header validation rule.</remarks>
```
Aprogiena: Why not have four base classes so that you do not need this remarks comment.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CheckDifficultyPosRule.cs
```diff
@@ -1,20 +1,35 @@
...
     /// Calculate the difficulty for a POS block and check that it is correct.
     /// This rule is only activated after the POW epoch is finished according to the value in <see cref="Consensus.LastPOWBlock"/>.
     /// </summary>
-    [HeaderValidationRule(CanSkipValidation = true)]
-    public class CheckDifficultyPosRule : StakeStoreConsensusRule
+    /// <remarks>This is header validation rule.</remarks>
+    public class CheckDifficultyPosRule : SyncConsensusRule
     {
+        /// <summary>Allow access to the POS parent.</summary>
+        protected PosConsensusRuleEngine PosParent;
```
Aprogiena: Is it used?

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/SmartContractPosCoinviewRule.cs
```diff
@@ -14,7 +14,7 @@ namespace Stratis.Bitcoin.Features.SmartContracts.Consensus.Rules
     /// <summary>
     /// Proof of stake override for the coinview rules - BIP68, MaxSigOps and BlockReward checks.
     /// </summary>
-    [FullValidationRule]
+    /// <remarks>This is partial validation rule</remarks>
```
Aprogiena: Missing the . everywhere copy and paste

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -130,41 +115,32 @@ public virtual void Dispose()
...
-            Guard.Assert(this.Network.Consensus.Rules.Any());
-
-            foreach (ConsensusRule consensusRule in this.Network.Consensus.Rules)
-            {
-                consensusRule.Parent = this;
-                consensusRule.Logger = this.loggerFactory.CreateLogger(consensusRule.GetType().FullName);
-                consensusRule.Initialize();
-
-                this.consensusRules.Add(consensusRule.GetType().FullName, consensusRule);
+            this.headerValidationRules = this.Network.Consensus.HeaderValidationRules.Select(x => x as SyncConsensusRule).ToList();
```
Aprogiena: I don't understand this

src/Stratis.Bitcoin/Consensus/Rules/ConsensusRule.cs
```diff
@@ -24,25 +21,32 @@ public abstract class ConsensusRule : IConsensusRule
...
+    {
         /// <summary>
-        /// Execute the logic in the current rule in an async approach.
+        /// Execute the logic in the current rule.
         /// If the validation of the rule fails a <see cref="ConsensusErrorException"/> will be thrown.
         /// </summary>
         /// <param name="context">The context that has all info that needs to be validated.</param>
-        /// <returns>The execution task.</returns>
-        public virtual Task RunAsync(RuleContext context)
+        public virtual void Run(RuleContext context)
```
Aprogiena: Why not abstract

src/NBitcoin/Rules/IConsensusRule.cs
```diff
@@ -1,6 +1,14 @@
 ﻿namespace NBitcoin.Rules
 {
-    public interface IConsensusRule
+    public interface IBaseConsensusRule
+    {
+    }
+
+    public interface ISyncBaseConsensusRule : IBaseConsensusRule
```
ferdeen: Why do you need this.  If it's not Async this it's synchronous by default.

src/Stratis.Bitcoin/Consensus/Rules/IRuleRegistration.cs
```diff
@@ -1,22 +1,19 @@
...
-        /// The rules that will be registered with the rules engine.
-        /// </summary>
-        /// <returns>A list of rules.</returns>
-        /// <remarks>
-        /// It is important to note that there is high importance to the order the rules are registered
-        /// with the engine, this is important for rules with dependencies on other rules.
-        /// Rules are executed in the same order they are registered with the engine.
-        /// </remarks>
-        ICollection<IConsensusRule> GetRules();
+        void RegisterRules(IConsensus consensus);
```
Aprogiena: Please no uncommented methods on interfaces

src/NBitcoin/Consensus.cs
```diff
@@ -116,10 +125,12 @@ public class Consensus : IConsensus
...
             BigInteger proofOfStakeLimitV2,
-            Money proofOfStakeReward
-            )
+            Money proofOfStakeReward)
         {
-            this.Rules = new List<IConsensusRule>();
+            this.IntegrityValidationRules = new List<ISyncBaseConsensusRule>();
+            this.HeaderValidationRules = new List<ISyncBaseConsensusRule>();
+            this.PartialValidationRules = new List<IAsyncBaseConsensusRule>();
+            this.FullValidationRules = new List<IAsyncBaseConsensusRule>();
```
ferdeen: If the member is IAsync... then maybe suffix the member with Async.  Like this.FullValidationRulesAsync and this.PartialValidationRulesAsync.

src/NBitcoin/IConsensus.cs
```diff
@@ -104,9 +104,16 @@ public interface IConsensus
         /// </summary>
         ConsensusFactory ConsensusFactory { get; }
 
-        /// <summary>
-        /// Rules specific to the given network.
-        /// </summary>
-        ICollection<IConsensusRule> Rules { get; set; }
+        /// <summary>Group of rules that are used during block's header validation specific to the given network.</summary>
+        List<ISyncBaseConsensusRule> HeaderValidationRules { get; set; }
```
ferdeen: block not block's ?

src/NBitcoin/IConsensus.cs
```diff
@@ -104,9 +104,16 @@ public interface IConsensus
...
+
+        /// <summary>Group of rules that are used during block integrity validation specific to the given network.</summary>
+        List<ISyncBaseConsensusRule> IntegrityValidationRules { get; set; }
+
+        /// <summary>Group of rules that are used during partial block validation specific to the given network.</summary>
+        List<IAsyncBaseConsensusRule> PartialValidationRules { get; set; }
+
+        /// <summary>Group of rules that are used during full validation (connection of a new block) specific to the given network.</summary>
+        List<IAsyncBaseConsensusRule> FullValidationRules { get; set; }
     }
```
ferdeen: For any types in the comments  - like Network - use <see>

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PosBlockSignatureRuleTest.cs
```diff
@@ -223,7 +223,7 @@ public async Task RunAsync_ProofOfStakeBlock_ScriptKeyDoesNotPassBlockSignatureV
             this.ruleContext.ValidationContext.Block = block;
             Assert.True(BlockStake.IsProofOfStake(this.ruleContext.ValidationContext.Block));
 
-            ConsensusErrorException exception = await Assert.ThrowsAsync<ConsensusErrorException>(() => this.consensusRules.RegisterRule<PosBlockSignatureRule>().RunAsync(this.ruleContext));
+            ConsensusErrorException exception = Assert.Throws<ConsensusErrorException>(() => this.consensusRules.RegisterRule<PosBlockSignatureRule>().Run(this.ruleContext));
```
ferdeen: The above tests are asserting the same bad block consensus exception.   

Maybe move the same code into a function.

```
private void AssertBadBlockSignatureWithConsensusException()
{
    Assert.True(BlockStake.IsProofOfStake(this.ruleContext.ValidationContext.Block));

    ConsensusErrorException exception = Assert.Throws<ConsensusErrorException>(() => this.consensusRules.RegisterRule<PosBlockSignatureRule>().Run(this.ruleContext));

    Assert.Equal(ConsensusErrors.BadBlockSignature, exception.ConsensusError);
}
```

src/Stratis.Bitcoin/Consensus/Rules/ConsensusRule.cs
noescape00: Answered on slack. Copypasted here: 
we need typecasts because interface is empty
But we need to call run or run async
So we are doing this ugly thing

src/Stratis.Bitcoin/Consensus/Rules/IRuleRegistration.cs
noescape00: What do you mean? There is no code commented out in this class

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/BitcoinActivationRule.cs
noescape00: done

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/SmartContractPosCoinviewRule.cs
noescape00: fixed

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
noescape00: Answered on slack. Copypasted here: 
we need typecasts because interface is empty
But we need to call run or run async
So we are doing this ugly thing

src/NBitcoin/Consensus.cs
noescape00: @Aprogiena @dangershony I don't mind both ways. Should I rename?

src/NBitcoin/IConsensus.cs
noescape00: fixed

src/NBitcoin/IConsensus.cs
noescape00: here I'm not referencing network class. I mean different network in general- like stratis, bitcoin etc

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/PosBlockSignatureRuleTest.cs
noescape00: Totally don't mind. Just outside of the scope of this PR

src/Stratis.Bitcoin/Consensus/Rules/IRuleRegistration.cs
Aprogiena: Commented code, not commented out

src/NBitcoin/Rules/IConsensusRule.cs
```diff
@@ -4,11 +4,18 @@ public interface IBaseConsensusRule
     {
     }
 
-    public interface ISyncBaseConsensusRule : IBaseConsensusRule
+    public interface IHeaderValidationConsensusRule : IBaseConsensusRule
+    {
+    }
+    public interface IIntegrityValidationConsensusRule : IBaseConsensusRule
```
Aprogiena: Empty line

src/Stratis.Bitcoin/Consensus/Rules/ConsensusRule.cs
```diff
@@ -24,31 +23,41 @@ public virtual void Initialize()
     }
 
     /// <summary>An abstract rule for implementing consensus rules.</summary>
-    public abstract class SyncConsensusRule : ConsensusRuleBase, ISyncBaseConsensusRule
+    public abstract class SyncConsensusRule : ConsensusRuleBase
```
Aprogiena: And now, why not interface?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/SetActivationDeploymentsPartialValidationRule.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            // Calculate the consensus flags and check they are valid.
+            context.Flags = this.Parent.NodeDeployments.GetFlags(context.ValidationContext.ChainTipToExtend);
+
+            return Task.CompletedTask;
+        }
+    }
+
+    /// <summary>Set the <see cref="RuleContext.Flags"/> property that defines what deployments have been activated.</summary>
+    public class SetActivationDeploymentsFullValidationRule : FullValidationConsensusRule
```
Aprogiena: I don't think you need this duplication of the code

src/NBitcoin/Consensus.cs
Aprogiena: I don't care

src/Stratis.Bitcoin/Consensus/Rules/ConsensusRule.cs
noescape00: You mean why not inherit from IConsensusRuleBase? 
Because abstract class already has 3 members already which all rules need.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/SetActivationDeploymentsPartialValidationRule.cs
noescape00: We kind of do because this rule is both for PV and FV. )

Aprogiena:
I'm okay with this, I do have a lot of things that should be improved, but I can understand that it is mostly the old code that is causing the problems and therefore I'm okay to say that this can be postponed, so let's just have it

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -230,25 +206,16 @@ public async Task PartialValidationAsync(ValidationContext validationContext)
...
         {
             try
             {
                 using (new StopwatchDisposable(o => this.PerformanceCounter.AddBlockProcessingTime(o)))
                 {
                     ruleContext.SkipValidation = ruleContext.ValidationContext.ChainTipToExtend.IsAssumedValid;
 
-                    foreach (ConsensusRuleDescriptor ruleDescriptor in rules)
-                    {
-                        if (ruleContext.SkipValidation && ruleDescriptor.RuleAttribute.CanSkipValidation)
```
dangershony: A suggestion from Fassa was to still check the can skip validation but instead we put it as a property on the rule itself so no need for attributes.

src/NBitcoin/Rules/IConsensusRule.cs
```diff
@@ -1,6 +1,22 @@
 ﻿namespace NBitcoin.Rules
 {
-    public interface IConsensusRule
+    public interface IConsensusRuleBase
+    {
+    }
+
+    public interface IHeaderValidationConsensusRule : IConsensusRuleBase
```
dangershony: I dont think we need interfaces for each type of collection, just make the distinction by name.

src/NBitcoin/Consensus.cs
```diff
@@ -83,12 +83,21 @@ public class Consensus : IConsensus
         public ConsensusFactory ConsensusFactory { get; }
 
         /// <inheritdoc />
-        public ICollection<IConsensusRule> Rules { get; set; }
+        public List<IIntegrityValidationConsensusRule> IntegrityValidationRules { get; set; }
```
dangershony: Here for example each can be a list of IConsensusRuleBase

src/NBitcoin/Consensus.cs
```diff
@@ -83,12 +83,21 @@ public class Consensus : IConsensus
...
+
+        /// <inheritdoc />
+        public List<IHeaderValidationConsensusRule> HeaderValidationRules { get; set; }
```
dangershony: The plan was to create a new class called ConsensusRules (like ConsensusFactory) and put the lists in that class.
So you would call  network.Consensus.Rules.HeaderValidationRules

src/NBitcoin/Rules/IConsensusRule.cs
Aprogiena: I actually think this is a good one

PR #1922, opened 2018-08-15
[SC] Abstract contract assembly loading functionality

rowandh:
Defines an assembly loader which is responsible for loading a ContractAssembly from ContractByteCode.


rowandh:
SC integration tests pass

src/Stratis.Bitcoin.Features.SmartContracts.Tests/GasInjectorTests.cs
```diff
@@ -131,7 +134,7 @@ public void TestGasInjector()
             var internalTxExecutorFactory =
                 new InternalTransactionExecutorFactory(this.keyEncodingStrategy, this.loggerFactory, this.network);
 
-            var vm = new ReflectionVirtualMachine(this.validator, internalTxExecutorFactory, this.loggerFactory, this.network, this.addressGenerator);
+            var vm = new ReflectionVirtualMachine(this.validator, internalTxExecutorFactory, this.loggerFactory, this.network, this.addressGenerator, this.assemblyLoader);
```
codingupastorm: Not a blocker but would be cleaner to just have this.vm, set it once in constructor, seeing as it's being used in all the methods the same way.

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ReflectionVirtualMachineTests.cs
```diff
@@ -67,7 +70,7 @@ public void VM_ExecuteContract_WithoutParameters()
 
             var internalTxExecutorFactory =
                 new InternalTransactionExecutorFactory(this.keyEncodingStrategy, this.loggerFactory, this.network);
-            var vm = new ReflectionVirtualMachine(this.validator, internalTxExecutorFactory, this.loggerFactory, this.network, this.addressGenerator);
+            var vm = new ReflectionVirtualMachine(this.validator, internalTxExecutorFactory, this.loggerFactory, this.network, this.addressGenerator, this.assemblyLoader);
```
codingupastorm: Same here as previous comment.

src/Stratis.SmartContracts.Executor.Reflection/Loader/ContractAssemblyLoader.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            // Assembly.Load(byte[]) loads the assembly into a new anonymous AssemblyLoadContext
+            // TODO in the future, we will use a custom AssemblyLoadContext
+            try
+            {
+                Assembly assembly = Assembly.Load(bytes.Value);
+
+                return Result.Ok<IContractAssembly>(new ContractAssembly(assembly));
+            }
+            catch (BadImageFormatException e)
```
codingupastorm: Curious about this, I'm interested whether this exception handles all possible errors that we could have.
Given that though, gas would be injected before we reach here so it doesn't really affect us.

thinking out loud..

src/Stratis.SmartContracts.Executor.Reflection/Loader/ContractByteCode.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Stratis.SmartContracts.Executor.Reflection.Loader
+{
+    /// <summary>
+    /// Represents a contract's bytecode.
+    /// </summary>
+    public class ContractByteCode
```
codingupastorm: Wouldn't block, but seems silly to me.

Interested in discussion: @fassadlr @majikandy

codingupastorm:
lgtm - interested in that 1 discussion and the test changes can probably be their own smaller PR one of us can put together after.

src/Stratis.SmartContracts.Executor.Reflection/Loader/ContractAssemblyLoader.cs
rowandh: I believe this handles all the exceptions that we expect could be thrown from `Assembly.Load`.

src/Stratis.Bitcoin.Features.SmartContracts.Tests/GasInjectorTests.cs
rowandh: Agreed

Issue #1755, opened 2018-07-27
[SC] Reuse smart contract assemblies or unload to avoid memory leak.


codingupastorm:
Currently we load assemblies into the node application every time we are executing a smart contract. However, these assemblies aren't tracked, and we discard the reference to them after getting the type that we need. 

Instead, assemblies should be stored in some kind of in-memory dictionary that we can access, and decide whether we can access this stored assembly or need to create an assembly from scratch. 

This is particularly interesting given the gas injection - we will have to make a decision about how best to handle this. Currently the gas injection code is only added to any methods which we know could be called. However if we're loading the assembly into memory semi-permanently then we may need to inject gas into every method in the assembly.

codingupastorm:
@rowandh lets make this post-launch too?

codingupastorm:
This issue will be used to track the potential memory leak that occurs when loading many assemblies with Assembly.Load.

Solution should be to wait for .NET Core 3.0.

fassadlr:
@codingupastorm is this still applicable? If not, close, otherwise re-open in VSTS.

codingupastorm:
Replaced with https://stratisplatformuk.visualstudio.com/StratisBitcoinFullNode/_workitems/edit/4195

Issue #1752, opened 2018-07-27
Smart contract compilation should fail for assemblies without only 1 contract or without a Deploy attribute.


codingupastorm:
Now that we can compile multiple contracts, we need to ensure that the contract the developer wants to be deployed is always the one deployed. 

To make this happen, the Deploy attribute must be on one class, or there has to only be one class, or compilation should fail. 

This will avoid situations where the contract that is deployed is not what was intended, which would obviously be problematic.

codingupastorm:
We decided that the rule should be:

If there is only one contract - always deploy this contract (regardless of whether it has Deploy).

If there is more than one contract, one contract MUST have Deploy. No others may have Deploy.

If these rules are broken throw a validation error (both in Sct and on network)

PR #1923, opened 2018-08-15
[SC] Creating contracts implementation

codingupastorm:
The beginning of the implementation of contract creation inside contracts.

Seeing this as really just a starting point with which to start writing full specifications and adjusting this code. 

All smart contract tests and integration tests are passing.


src/Stratis.SmartContracts.Executor.Reflection/Compilation/SmartContractDecompiler.cs
```diff
@@ -35,6 +35,9 @@ public static SmartContractDecompilation GetModuleDefinition(byte[] bytes, IAsse
                 result.ContractType = developedTypes.FirstOrDefault(x => x.CustomAttributes.Any(y => y.AttributeType.Name == typeof(DeployAttribute).Name));
             }
 
+            // If contract type wasn't set yet, set the first.
```
codingupastorm: Still working out the exact right functionality here - whether we want to throw an exception or just work around it and deploy the first one. 

Will be solved in https://github.com/stratisproject/StratisBitcoinFullNode/issues/1752

src/Stratis.SmartContracts/CreateContract.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Stratis.SmartContracts
+{
+    public sealed class CreateContract
```
codingupastorm: Think this *and* TransferFundsToContract should be named better.

codingupastorm:
All Smartcontracts tests and integration tests passing

src/Stratis.SmartContracts/CreateContract.cs
rowandh: Can we remove this class and pass the params directly into `Create`? Removes an object allocation, simplifies whitelisting, keeps our API surface smaller, prevents confusion about what types can and cannot be instantiated, etc.

src/Stratis.SmartContracts.Executor.Reflection/InternalTransactionExecutor.cs
```diff
@@ -40,9 +41,59 @@ public sealed class InternalTransactionExecutor : IInternalTransactionExecutor
...
+            var balance = smartContractState.GetBalance();
+            if (balance < amountToTransfer)
+            {
+                this.logger.LogTrace("(-)[INSUFFICIENT_BALANCE]:{0}={1}", nameof(balance), balance);
+                throw new InsufficientBalanceException();
+            }
+
+            // Build objects for VM
+            byte[] contractCode = this.contractStateRepository.GetCode(smartContractState.Message.ContractAddress.ToUint160(this.network)); // TODO: Fix this when calling from constructor.
+
```
rowandh: We already have an instance of the Assembly with this Type on it. Rather than retrieving the code and creating another one, can we reuse the existing Assembly?

codingupastorm: The instance of the assembly we have doesn't necessarily have the correct methods injected with gas spending code. I think overhauling assembly caching should be a separate issue, believe we have one for it #1755

rowandh: Sounds good 👍

Issue #1810, opened 2018-08-02
Ensure rewind data is not aggregated


dangershony:
Ensure rewind data is not aggregated (1 block per rewind data) while still saving those in batches

Discussion here
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1768

dangershony:
The issue I have here is that in order to reorg we will need to have blocks available.
Can we guarantee that? what if we dont have blocks available we need to delete the entire store (or alternatively download form a peer).

noescape00:
More info on the issue: 

How it currently works: 
When new block is connected changes that his block has introduced are commited to the CachedCoinView. After some time those changes are flushed- a new rewind data instance is created. This instance might combine changes for many blocks. In IBD one rewind data might have 2000 blocks worth of changes.
When we try to rewind cosnensus we can't go 1 block back, only 1 rewind data back so it's possible that we will rewind 2000 blocks which is not intended and creates several problems for consensus manager. 


We need to implement a mechanism similar to pending storage in BSQ- 
create rewind data per block and put it to the pending storage (in the memory),
after pending storage reaches threshold we flush it to the database. We also should flush every ~30 seconds no matter the pending storage batch size and we should when we've reached the tip (synced with the best chain we are aware of).

Changes will mostly touch CachedCoinView. 
Also it's important to make sure we can rewind using the rewind data from the pending storage and not only from DBreeze coinview.

noescape00:
fixed

PR #1924, opened 2018-08-15
[Activation] Rewind data refactor

fenix2222:
This is a refactor of coin views to ensure rewind data can be aggregated per block as per https://github.com/stratisproject/StratisBitcoinFullNode/issues/1810 

Unit tests will be fixed in a separate PR to make this PR smaller


src/Stratis.Bitcoin.Features.MemoryPool.Tests/TestChainFactory.cs
```diff
@@ -151,7 +152,8 @@ public static async Task<ITestChainContext> CreateAsync(Network network, Script
             blockDefinition = new PowBlockDefinition(consensus, dateTimeProvider, loggerFactory, mempool, mempoolLock, minerSettings, network, consensusRules);
             newBlock = blockDefinition.Build(chain.Tip, scriptPubKey);
 
-            var mempoolValidator = new MempoolValidator(mempool, mempoolLock, dateTimeProvider, new MempoolSettings(nodeSettings), chain, cachedCoinView, loggerFactory, nodeSettings, consensusRules);
+            //var mempoolValidator = new MempoolValidator(mempool, mempoolLock, dateTimeProvider, new MempoolSettings(nodeSettings), chain, cachedCoinView, loggerFactory, nodeSettings, consensusRules);
```
ferdeen: Can we remove this comment ?

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
```diff
@@ -9,16 +10,17 @@
...
 using Stratis.Bitcoin.Features.Consensus;
-using Stratis.Bitcoin.Features.Consensus.CoinViews;
 using Stratis.Bitcoin.Features.Consensus.Rules.CommonRules;
 using Stratis.Bitcoin.IntegrationTests.Common;
 using Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers;
 using Stratis.Bitcoin.Tests.Common;
 using Stratis.Bitcoin.Utilities;
 using Xunit;
 using static NBitcoin.Transaction;
+using UInt256Comparer = Stratis.Bitcoin.Consensus.UInt256Comparer;
```
ferdeen: If you remove the namespace at line 15, you can also remove this line - so won't need to alias UInt256Comparer.

src/Stratis.Bitcoin.Features.MemoryPool.Tests/TestChainFactory.cs
fenix2222: Fixed

src/Stratis.Bitcoin.IntegrationTests/CoinViewTests.cs
fenix2222: Fixed

src/Stratis.Bitcoin/Utilities/RewindData.cs
```diff
@@ -60,6 +60,9 @@ public List<UnspentOutputs> OutputsToRestore
             }
         }
 
+        /// <summary>The size of the block in bytes, the block must be serialized for this property to be set.</summary>
+        public long? BlockSize { get; protected set; }
```
noescape00: no need to save it. 

I see you need it in cached coinview for batching. Instead create a class which will have 2 members- rewind data and block size (or maybe some more metadata if you need) and use that class with metadata in the async queue. 
But don't save block size to the DB, we don't need it

src/Stratis.Bitcoin/Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -34,17 +34,17 @@ public class DBreezeCoinView : ICoinView, IDisposable
...
         /// <summary>Performance counter to measure performance of the database insert and query operations.</summary>
-        public BackendPerformanceCounter PerformanceCounter
-        {
-            get { return this.performanceCounter; }
-        }
+        public BackendPerformanceCounter PerformanceCounter { get; }
 
         /// <summary>Provider of time functions.</summary>
-        protected readonly IDateTimeProvider dateTimeProvider;
+        protected readonly IDateTimeProvider DateTimeProvider;
```
noescape00: might be private

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -105,8 +103,8 @@ public static IFullNodeBuilder UsePowConsensus(this IFullNodeBuilder fullNodeBui
                 .FeatureServices(services =>
                 {
                     services.AddSingleton<ConsensusOptions, ConsensusOptions>();
-                    services.AddSingleton<DBreezeCoinView>();
-                    services.AddSingleton<ICoinView, CachedCoinView>();
+                    services.AddSingleton<ICachedCoinView, CachedCoinView>().AddSingleton<ICoinView>(provider => provider.GetService<ICachedCoinView>());
+                    services.AddSingleton<ICoinViewStorage, DBreezeCoinView>().AddSingleton(provider => (DBreezeCoinView)provider.GetService<ICoinViewStorage>());
```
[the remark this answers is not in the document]
fenix2222: There are places in code that use DBreezeCoinView directly (i.e. not the interface), so I am making sure if you inject an interface or the type, you will get the same singleton. There might be a different way to do it. I am mainly familiar with Autofac, Ninject and StructureMap. Not sure how to do it with MS DI container.

src/Stratis.Bitcoin.Features.Consensus/Rules/PowConsensusRuleEngine.cs
```diff
@@ -52,20 +58,7 @@ public override async Task<RewindState> RewindAsync()
...
-            var cache = this.UtxoSet as CachedCoinView;
-            if (cache != null)
-            {
-                this.logger.LogInformation("Flushing Cache CoinView...");
-                cache.FlushAsync().GetAwaiter().GetResult();
-                cache.Dispose();
-            }
-
-            ((DBreezeCoinView)((CachedCoinView)this.UtxoSet).Inner).Dispose();
+            return Task.CompletedTask;
```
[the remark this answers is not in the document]
fenix2222: We need it as it is part of the interface and `PosConsensusRuleEngine.cs` uses it

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
```diff
@@ -23,18 +22,13 @@ public class SaveCoinviewRule : UtxoStoreConsensusRule
...
             // Persist the changes to the coinview. This will likely only be stored in memory,
             // unless the coinview treashold is reached.
             this.Logger.LogTrace("Saving coinview changes.");
             var utxoRuleContext = context as UtxoRuleContext;
-            await this.PowParent.UtxoSet.SaveChangesAsync(utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), null, oldBlockHash, nextBlockHash).ConfigureAwait(false);
-
-            bool forceFlush = this.FlushRequired(context.ValidationContext.ChainTipToExtend);
-            if (this.PowParent.UtxoSet is CachedCoinView cachedCoinView)
-                await cachedCoinView.FlushAsync(forceFlush).ConfigureAwait(false);
+            await this.PowParent.UtxoSet.AddRewindDataAsync(utxoRuleContext?.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), currentBlock).ConfigureAwait(false);
```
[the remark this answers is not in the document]
fenix2222: Change the cast. If it is never null doesn't make sense to do `as` cast

src/Stratis.Bitcoin/Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -324,10 +302,10 @@ public Task<uint256> Rewind()
                 uint256 res = null;
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
-                    transaction.SynchronizeTables("BlockHash", "Coins", "Rewind");
-                    if (this.GetRewindIndex(transaction) == -1)
+                    transaction.SynchronizeTables(BlockHashKey, CoinsKey, RewindDataKey);
+                    if (!this.TryGetRewindIndex(transaction, out int _))
```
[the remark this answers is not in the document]
fenix2222: Fixed

src/Stratis.Bitcoin/Consensus/CoinViews/DBreezeCoinView.cs
fenix2222: Removed it in favour of a local variable.

src/Stratis.Bitcoin/Consensus/CoinViews/ICachedCoinView.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using NBitcoin;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Consensus.CoinViews
+{
+    /// <summary>
+    /// Database of UTXOs.
```
ferdeen: What does 'Database of UTXOs' mean in terms of this interface ICachedCoinView ?

fenix2222:
@dangershony 
> It is important, that coinview is not moved to Stratis.Bitcoin

Why? We discussed this with @noescape00 and decided to move it there...

src/Stratis.Bitcoin/Consensus/CoinViews/ICoinView.cs
```diff
@@ -1,45 +1,28 @@
...
+﻿using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using NBitcoin;
 using Stratis.Bitcoin.Utilities;
 
-namespace Stratis.Bitcoin.Features.Consensus.CoinViews
+namespace Stratis.Bitcoin.Consensus.CoinViews
 {
```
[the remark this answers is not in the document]
fenix2222: I haven't changed the summary, it was like this before.

src/Stratis.Bitcoin/Consensus/CoinViews/ICachedCoinView.cs
fenix2222: I haven't changed the summary, it was like this before. It only shows as new file because I moved it to another project.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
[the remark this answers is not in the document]
noescape00: >There are places in code that use DBreezeCoinView directly (i.e. not the interface - see ConsensusStats.cs)

It is used to access performance counter. If you move performance counter to the interface you will be able to use interface everywhere. 

Imho the component should be always accessed by the interface.

src/Stratis.Bitcoin/Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
             public TxOut[] OriginalOutputs;
         }
 
+        /// <summary>
+        /// Rewind data with block size that will be kept in the queue until it is persisted to a storage
```
noescape00: it's not block size. 
it kinda depends on the block size but rewind data has less stuff than block so it's always smaller

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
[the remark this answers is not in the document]
fenix2222: It is not using Dependency injection, it uses `this.dbreeze = stack.Find<DBreezeCoinView>()` this coinviewstack class. That means we should refactor it or maybe inject interfaces directly?

src/Stratis.Bitcoin/Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -58,106 +85,112 @@ private class CacheItem
...
+
+        /// <summary>Batch of rewind data items which should be saved in the database.</summary>
+        /// <remarks>Write access should be protected by <see cref="getBlockLock"/>.</remarks>
+        private readonly List<QueuedRewindData> rewindDataBatch;
+
+        /// <summary>Task that runs <see cref="DequeueRewindDataContinuouslyAsync"/>.</summary>
+        private Task dequeueLoopTask;
+
+        /// <summary>Protects the batch from being modifying while <see cref="GetTipHashAsync"/> method is using the batch.</summary>
+        private readonly object getBlockLock;
```
noescape00: lock is not needed here. 

In BSQ it's needed because we access batch when we get the block- if it's not in batch we get it from DB. 
In case of coinview we never have to access rewind data from the outside so the lock is not needed

src/Stratis.Bitcoin/Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -424,6 +419,141 @@ public async Task<uint256> Rewind()
         public void Dispose()
         {
             this.lockobj.Dispose();
+
```
noescape00: Can you please move Dispose() method to the end of the class?

src/Stratis.Bitcoin/Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -424,6 +419,141 @@ public async Task<uint256> Rewind()
...
+            // Let current batch saving task finish.
+            this.rewindDataQueue.Dispose();
+            this.dequeueLoopTask?.GetAwaiter().GetResult();
+        }
+
+        public void Initialize()
```
noescape00: please move to the top of the class (after constructor)

src/Stratis.Bitcoin/Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -340,23 +312,27 @@ private async Task EvictAsync()
...
-            Guard.NotNull(oldBlockHash, nameof(oldBlockHash));
-            Guard.NotNull(nextBlockHash, nameof(nextBlockHash));
+            Guard.NotNull(currentBlock, nameof(currentBlock));
             Guard.NotNull(unspentOutputs, nameof(unspentOutputs));
-            this.logger.LogTrace("({0}.Count():{1},{2}.Count():{3},{4}:'{5}',{6}:'{7}')", nameof(unspentOutputs), unspentOutputs.Count(), nameof(originalOutputs), originalOutputs?.Count(), nameof(oldBlockHash), oldBlockHash, nameof(nextBlockHash), nextBlockHash);
+
+            List<UnspentOutputs> unspentOutputsList = unspentOutputs.ToList();
+            this.logger.LogTrace("({0}.Count():{1},{2}:'{3}')", nameof(unspentOutputsList), unspentOutputsList.Count, nameof(currentBlock), currentBlock.HashBlock);
+
+            RewindData rewindData = new RewindData(currentBlock.HashBlock);
```
noescape00: => 

```
var rewindData = new RewindData(currentBlock.HashBlock);
```

src/Stratis.Bitcoin/Consensus/CoinViews/CoinViewStack.cs
```diff
@@ -1,10 +1,10 @@
 ﻿using System.Collections.Generic;
 using Stratis.Bitcoin.Utilities;
 
-namespace Stratis.Bitcoin.Features.Consensus.CoinViews
+namespace Stratis.Bitcoin.Consensus.CoinViews
 {
     /// <summary>
-    /// Stack of coinview layers. All classes in the stack have to be based on <see cref="CoinView"/> class
+    /// Stack of coinview layers. All classes in the stack have to be based on <see cref="ICoinView"/> interface
```
noescape00: CoinViewStack can be completely removed. It's used only in tests and in ConsensusStats.
Since we have both coinviews accessible via DI we don't need to use stack in consensus stats anymore

src/Stratis.Bitcoin/Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -201,33 +195,20 @@ private void SetBlockHash(DBreeze.Transactions.Transaction transaction, uint256
...
             this.blockHash = nextBlockHash;
-            transaction.Insert<byte[], uint256>("BlockHash", blockHashKey, nextBlockHash);
+            transaction.Insert<byte[], uint256>(BlockHashKey, blockHashKey, nextBlockHash);
 
             this.logger.LogTrace("(-)");
         }
 
         /// <inheritdoc />
-        public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash, uint256 nextBlockHash)
+        public Task PersistDataAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, List<RewindData> rewindDataCollection, uint256 oldBlockHash, uint256 nextBlockHash)
```
noescape00: `originalOutputs` is not used in the method anymore and can be removed

src/Stratis.Bitcoin/Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -201,33 +195,20 @@ private void SetBlockHash(DBreeze.Transactions.Transaction transaction, uint256
...
         {
-            this.logger.LogTrace("({0}.Count():{1},{2}.Count():{3},{4}:'{5}',{6}:'{7}')", nameof(unspentOutputs), unspentOutputs?.Count(), nameof(originalOutputs), originalOutputs?.Count(), nameof(oldBlockHash), oldBlockHash, nameof(nextBlockHash), nextBlockHash);
+            Guard.NotNull(unspentOutputs, nameof(unspentOutputs));
 
-            RewindData rewindData = originalOutputs != null ? new RewindData(oldBlockHash) : null;
-            int insertedEntities = 0;
+            List<UnspentOutputs> allUnspentOutputs = unspentOutputs.ToList();
+            this.logger.LogTrace("({0}.Count():{1},{2}.Count():{3},{4}.Count():'{5}',{6}:{7},{8}:{9})", nameof(unspentOutputs), allUnspentOutputs.Count, nameof(originalOutputs), originalOutputs?.Count(), nameof(rewindDataCollection), rewindDataCollection.Count, nameof(oldBlockHash), oldBlockHash, nameof(nextBlockHash), nextBlockHash);
```
noescape00: logging style violation

fenix2222:
@dangershony Refactored it back
@noescape00 all feedback addressed

src/Stratis.Bitcoin.Features.Consensus/CoinViews/ICachedCoinView.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Threading.Tasks;
+using NBitcoin;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.CoinViews
+{
+    /// <summary>
+    /// Database of UTXOs.
+    /// </summary>
+    public interface ICachedCoinView : ICoinView
```
ferdeen: With regards to naming, what about `ICoinViewRewindData` ?

src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewTest.cs
```diff
@@ -11,11 +10,17 @@ namespace Stratis.Bitcoin.Features.Consensus.Tests.CoinViews
...
         {
-            this.coinView = new TestCoinView();
+            var coinViewMock = new Mock<ICoinView>();
+            var fetchCoinResponse = new FetchCoinsResponse(new UnspentOutputs[0], new uint256(987263876253));
+            coinViewMock.Setup(c => c.FetchCoinsAsync(It.IsAny<uint256[]>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync(fetchCoinResponse);
+            coinViewMock.Setup(c => c.GetTipHashAsync(It.IsAny<CancellationToken>()))
+                .ReturnsAsync(fetchCoinResponse.BlockHash);
+            this.coinView = coinViewMock.Object;
         }
```
ferdeen: Maybe move this into the actual test, add some line spacing, and make uint256(987263876253) a variable ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
...
 
+        /// <summary>
+        /// Rewind data with its byte size that will be kept in the queue until it is persisted to a storage
+        /// </summary>
+        private class QueuedRewindData
+        {
+            /// <summary>Rewind data to be saved.</summary>
+            public RewindData RewindData { get; set; }
+
+            /// <summary>The size of the rewind data in bytes.</summary>
```
ferdeen: Does this need to be nullable ? If removed, set this value to zero on construction.  Then any calling code doesn't need to check for null.

ferdeen: Above set's can be made private.

noescape00: Then it will need to check for 0

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -201,33 +196,20 @@ private void SetBlockHash(DBreeze.Transactions.Transaction transaction, uint256
...
         /// <inheritdoc />
-        public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash, uint256 nextBlockHash)
+        public Task PersistDataAsync(IEnumerable<UnspentOutputs> unspentOutputs, List<RewindData> rewindDataCollection, uint256 oldBlockHash, uint256 nextBlockHash)
         {
-            this.logger.LogTrace("({0}.Count():{1},{2}.Count():{3},{4}:'{5}',{6}:'{7}')", nameof(unspentOutputs), unspentOutputs?.Count(), nameof(originalOutputs), originalOutputs?.Count(), nameof(oldBlockHash), oldBlockHash, nameof(nextBlockHash), nextBlockHash);
+            Guard.NotNull(unspentOutputs, nameof(unspentOutputs));
 
-            RewindData rewindData = originalOutputs != null ? new RewindData(oldBlockHash) : null;
-            int insertedEntities = 0;
+            List<UnspentOutputs> allUnspentOutputs = unspentOutputs.ToList();
```
ferdeen: Maybe unspentOutputList ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -250,40 +232,27 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
 
                         this.SetBlockHash(transaction, nextBlockHash);
 
-                        all.Sort(UnspentOutputsComparer.Instance);
-                        foreach (UnspentOutputs coin in all)
+                        allUnspentOutputs.Sort(UnspentOutputsComparer.Instance);
+                        foreach (UnspentOutputs coin in allUnspentOutputs)
```
ferdeen: Maybe `unspentOutput`, instead of `Coin` ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/InMemoryCoinView.cs
```diff
@@ -25,6 +27,9 @@ public class InMemoryCoinView : ICoinView
         /// <remarks>All access to this object has to be protected by <see cref="lockobj"/>.</remarks>
         private uint256 tipHash;
 
+        private CachePerformanceCounter performanceCounter;
+        private BackendPerformanceCounter performanceCounter1;
```
ferdeen: Could this have a better name ?

src/Stratis.Bitcoin.Features.Consensus/Rules/PowConsensusRuleEngine.cs
```diff
@@ -50,22 +56,24 @@ public override async Task<RewindState> RewindAsync()
         }
 
         /// <inheritdoc />
-        public override Task Initialize()
+        public override async Task Initialize()
         {
-            return ((DBreezeCoinView)((CachedCoinView)this.UtxoSet).Inner).InitializeAsync();
+            if (this.UtxoSet is IBackedCoinView backedCoinView)
+            {
+                await backedCoinView.CoinViewStorage.InitializeAsync();
```
ferdeen: `.ConfigureAwait(false);`

src/Stratis.Bitcoin.Features.MemoryPool/MemPoolCoinView.cs
```diff
@@ -51,14 +51,7 @@ public MempoolCoinView(ICoinView inner, ITxMempool memPool, SchedulerLock mempoo
...
         /// </summary>
-        public ICoinView Inner { get; }
-
-        /// <inheritdoc />
-        public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash,
-            uint256 nextBlockHash)
-        {
-            throw new NotImplementedException();
-        }
+        public ICoinView CoinView { get; }
```
ferdeen: Can be made private.

src/Stratis.Bitcoin.Features.MemoryPool/MemPoolCoinView.cs
```diff
@@ -213,5 +206,11 @@ public TxOut GetOutputFor(TxIn input)
         {
             return this.Set.GetOutputFor(input);
         }
+
+        public void Dispose()
+        {
+        }
+
+        public ICoinViewStorage CoinViewStorage { get; }
```
ferdeen: This is never used.

src/Stratis.Bitcoin.Features.MemoryPool/MemPoolCoinView.cs
```diff
@@ -213,5 +206,11 @@ public TxOut GetOutputFor(TxIn input)
         {
             return this.Set.GetOutputFor(input);
         }
+
+        public void Dispose()
+        {
+        }
```
ferdeen: Could any of the locking type objects get disposed in here ?

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/SmartContractSaveCoinviewRule.cs
```diff
@@ -13,14 +13,13 @@ public sealed class SmartContractSaveCoinviewRule : UtxoStoreConsensusRule
...
-            uint256 oldBlockHash = context.ValidationContext.ChainTipToExtend.Previous.HashBlock;
-            uint256 nextBlockHash = context.ValidationContext.ChainTipToExtend.HashBlock;
+            ChainedHeader currentBlock = context.ValidationContext.ChainTipToExtend;
 
             // Persist the changes to the coinview. This will likely only be stored in memory,
             // unless the coinview treashold is reached.
             this.Logger.LogTrace("Saving coinview changes.");
             var utxoRuleContext = context as UtxoRuleContext;
-            await this.PowParent.UtxoSet.SaveChangesAsync(utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), null, oldBlockHash, nextBlockHash).ConfigureAwait(false);
+            await this.PowParent.UtxoSet.AddRewindDataAsync(utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), currentBlock).ConfigureAwait(false);
```
ferdeen: `utxoRuleContext` possible NullReferenceException.

ferdeen:
In CoinViewTests CanHandleReorgs, CanRewind and TestCacheCoinView are failing, when I run locally on my machine. 

Please can you double check non your branch @fenix2222.

image

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
```diff
@@ -116,6 +115,7 @@ public class TestContext
             public Money HIGHERFEE = 4 * Money.COIN;
             public int baseheight;
             public CachedCoinView cachedCoinView;
+            public InMemoryCoinView inMemoryCoinView;
```
ferdeen: As this is public maybe name to `InMemoryCoinView`.

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
```diff
@@ -136,8 +136,11 @@ public async Task InitializeAsync()
                 this.network.Consensus.Rules = new FullNodeBuilderConsensusExtension.PowConsensusRulesRegistration().GetRules();
 
                 IDateTimeProvider dateTimeProvider = DateTimeProvider.Default;
+                var chainState = new ChainState(new InvalidBlockHashStore(dateTimeProvider));
 
-                this.cachedCoinView = new CachedCoinView(new InMemoryCoinView(this.chain.Tip.HashBlock), dateTimeProvider, new LoggerFactory());
+                var coinViewStorageMock = new Mock<ICoinViewStorage>();
+                this.cachedCoinView = new CachedCoinView(chainState, coinViewStorageMock.Object, dateTimeProvider, new LoggerFactory(), new NodeLifetime());
+                this.inMemoryCoinView = new InMemoryCoinView(newBlock.Block.GetHash());
```
ferdeen: Should be `this.newBlock`.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
             public TxOut[] OriginalOutputs;
         }
 
+        /// <summary>
+        /// Rewind data with its byte size that will be kept in the queue until it is persisted to a storage
```
Aprogiena: Missing `.`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
...
+        /// </summary>
+        private class QueuedRewindData
+        {
+            /// <summary>Rewind data to be saved.</summary>
+            public RewindData RewindData { get; set; }
+
+            /// <summary>The size of the rewind data in bytes.</summary>
+            public long? DataSize { get; set; }
+
+            public static QueuedRewindData Create(RewindData rewindData)
```
Aprogiena: Why not normal constructor?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
...
+            {
+                long size;
+                using (var ms = new MemoryStream())
+                {
+                    var bitcoinStream = new BitcoinStream(ms, true);
+                    bitcoinStream.ReadWrite(rewindData);
```
Aprogiena: This is not okay. We should decide whether we want the byte size or not. I think we could be fine without the actual byte size and instead just having the number of elements that the rewind data is holding.

Or, we can go this way and really measure exactly how much data we have, but in that case we need to reuse the data when we are saving to the database. This means that the queue rewind data should actually be just byte array which is already serialised, so that we do not serialise it again when we actually do the writing to the database.

As it is right now, we do have a massive overhead of extra serialisation which is just wasted. So please decide on either not having the exact size and just using the number of elements we are holding inside or reuse the serialised data.

One could argue that in the case of reusing the serialise data, we will have overhead when reorg happens, but is that is a very rare case, it is totally fine.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -34,6 +36,31 @@ private class CacheItem
...
+                    size = bitcoinStream.Serializing ? bitcoinStream.Counter.WrittenBytes : bitcoinStream.Counter.ReadBytes;
```
Aprogiena: You know you are writing here, so you don't really need this if

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -58,106 +85,112 @@ private class CacheItem
...
+        internal const int BatchThresholdSizeBytes = 5 * 1000 * 1000;
 
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
+        /// <summary>Queue which contains rewind data items that should be saved to the database.</summary>
+        private readonly AsyncQueue<QueuedRewindData> rewindDataQueue;
+
+        /// <summary>Batch of rewind data items which should be saved in the database.</summary>
+        /// <remarks>Write access should be protected by <see cref="getBlockLock"/>.</remarks>
```
Aprogiena: I don't see any such Getblocklock

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -58,106 +85,112 @@ private class CacheItem
...
+        ) : this(chainState, dateTimeProvider, loggerFactory, nodeLifetime)
         {
-            Guard.NotNull(inner, nameof(inner));
-            this.inner = inner;
+            Guard.NotNull(coinViewStorage, nameof(coinViewStorage));
+            this.chainState = chainState;
+            this.CoinViewStorage = coinViewStorage;
+            this.nodeLifetime = nodeLifetime;
+            this.rewindDataBatch = new List<QueuedRewindData>();
+            this.rewindDataQueue = new AsyncQueue<QueuedRewindData>();
```
Aprogiena: These two guys you already initialised in the other constructor

Aprogiena: And also the chain state and also the node lifetime

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -340,23 +312,27 @@ private async Task EvictAsync()
...
+                if ((this.blockHash != null) && (currentBlock.Previous.HashBlock != this.blockHash))
                 {
                     this.logger.LogTrace("(-)[BLOCKHASH_MISMATCH]");
                     throw new InvalidOperationException("Invalid oldBlockHash");
                 }
 
-                this.blockHash = nextBlockHash;
-                foreach (UnspentOutputs unspent in unspentOutputs)
+                this.blockHash = currentBlock.HashBlock;
+                unspentOutputsList.Sort(UnspentOutputsComparer.Instance);
```
Aprogiena: Why are we doing this?

src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinViewTest.cs
fenix2222: Test refactoring will be done in a separate PR as per my PR summary.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: This is just a DTO and not a domain entity and if it needs to be serialised we may end up having issues with private setters

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: Usually if I have a non-trivial constructor code (in this case serialization) I prefer using factory pattern as it gives you a much better flexibility. This is also consistent with MS guide: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/constructor and https://refactoring.guru/replace-constructor-with-factory-method. Adding a constructor might also force us to create a default constructor if serialization is needed. 

Having said all this, if we are getting away from serialization and byte size tracking, I have more than happy to move it to a constructor.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: I agree. It does feel a bit heavy. I think we can go with `number of transactions to remove + number of original ones`. It will not be very accurate but probably is good enough for a flush logic. The question now is, what is a sensible threshold? I have no idea: @Aprogiena, @ferdeen  or @noescape00?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: Sure, but most likely we will remove all this logic anyway as per above.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/ICachedCoinView.cs
fenix2222: Problem is it is not just a rewinddata.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: Yeah, not much improvement here and it looks like we are going to change this thing anyway

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -56,108 +70,109 @@ private class CacheItem
         };
 
         /// <summary>Default maximum number of transactions in the cache.</summary>
-        public const int CacheMaxItemsDefault = 100000;
+        public const int CacheMaxItemsDefault = 100_000;
+
+        /// <summary>The current batch size in bytes.</summary>
```
ferdeen: I think this is the old comment when the code was storing the batch size in bytes.  Now it's doing a count of items.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -393,37 +366,162 @@ public async Task<uint256> Rewind()
...
+                if (dequeueTask.Status == TaskStatus.RanToCompletion)
+                {
+                    QueuedRewindData item = dequeueTask.Result;
+
+                    // Set the dequeue task to null so it can be assigned on the next iteration.
+                    dequeueTask = null;
+
+                    this.rewindDataBatch.Add(item);
+
+                    this.currentCountOfAllRewindDataItems += item.ItemsCount;
```
ferdeen: In a recent PR I did @Aprogiena asked me to use `Interlocked.Add` when keeping count of an `AsyncQueue`.

https://github.com/stratisproject/StratisBitcoinFullNode/blob/39e7a38e70ef9cf1ee4b1cde1b0c7843736cf9a1/src/Stratis.Bitcoin.Features.Wallet/WalletSyncManager.cs#L268-L273

I'm not 100% sure you would need it here - @Aprogiena ?

Also we added trace logging around any change in queue size - which was useful.  Maybe you can add some around here as well.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -393,37 +366,162 @@ public async Task<uint256> Rewind()
...
+                    saveBatch = true;
+                }
+
+                if (saveBatch)
+                {
+                    if (this.rewindDataBatch.Count != 0)
+                    {
+                        await this.SaveBatchAsync().ConfigureAwait(false);
+
+                        this.rewindDataBatch.Clear();
```
ferdeen: There is an assumption the `this.SaveBatchAsync()` has worked.  If it fails or returns before saving the code is always clearing `this.rewindDataBatch`.

Is there any mileage is getting some sort of return value from `this.SaveBatchAsync()` and if all is well clear `this.rewindDataBatch`, and set `this.currentCountOfAllRewindDataItems = 0`. ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: Actually after reading what you linked, it looks like they give us plenty of arguments against this construction here, which seems unjustified. I don't really care that much, just looks odd to use it for such a trivial construction

src/Stratis.Bitcoin.Features.Consensus/CoinViews/InMemoryCoinView.cs
```diff
@@ -40,6 +45,59 @@ public Task<uint256> GetTipHashAsync(CancellationToken cancellationToken = defau
...
+                    {
+                        existing = unspent.Clone();
+                        this.unspents.Add(unspent.TransactionId, existing);
+                    }
+
+                    if (existing.IsPrunable)
+                        this.unspents.Remove(unspent.TransactionId);
+                }
+            }
+
```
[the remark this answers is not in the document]
fenix2222: Yeah, we might need to refactor this one. It is only used by tests. Ideally, it should be moved to some common tests project.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -352,7 +352,7 @@ public async Task AddRewindDataAsync(IList<UnspentOutputs> unspentOutputs, Chain
                     {
                         // We'll need to restore the original outputs.
                         UnspentOutputs clone = unspent.Clone();
-                        clone.Outputs = original.OriginalOutputs;
+                        clone.Outputs = original.UnspentOutputs.Outputs;
```
Aprogiena: Should this not be using toarray in order to have a clone?

noescape00:
@fenix2222 please pm me when suggested fixes are addressed. I'll do my 2nd round of review.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -34,17 +35,14 @@ public class DBreezeCoinView : ICoinView, IDisposable
...
+        private const string RewindDataKey = "RewindData";
 
-        /// <summary>Performance counter to measure performance of the database insert and query operations.</summary>
-        public BackendPerformanceCounter PerformanceCounter
-        {
-            get { return this.performanceCounter; }
-        }
+        private const string CoinsKey = "Coins";
+
+        private const string BlockHashKey = "BlockHash";
```
Aprogiena: These three guys, they are table names, not keys

src/Stratis.Bitcoin.Features.Consensus/CoinViews/ICoinViewStorage.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    public interface ICoinViewStorage : ICoinView
+    {
+        /// <summary>
+        /// Persists changes to the coinview storage.
+        /// <para>
+        /// This method is provided (in <paramref name="unspentOutputs"/> parameter) with information about all
+        /// transactions that are either new or were changed in the new block. It is also provided with information
+        /// in original outputs about the previous state of those transactions (if any),
+        /// which is used for <see cref="ICoinView.Rewind"/> operation.
```
Aprogiena: It looks like you changed the logic, so this comment no longer reflects the reality

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -250,40 +231,26 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
...
+                            this.logger.LogTrace("Outputs of transaction ID '{transactionId}' are {prunableState} and will be {action} to the database.", unspentOutput.TransactionId, unspentOutput.IsPrunable ? "PRUNABLE" : "NOT PRUNABLE", unspentOutput.IsPrunable ? "removed" : "inserted");
+                            if (unspentOutput.IsPrunable) transaction.RemoveKey(CoinsKey, unspentOutput.TransactionId.ToBytes(false));
+                            else transaction.Insert(CoinsKey, unspentOutput.TransactionId.ToBytes(false), unspentOutput.ToCoins());
+                        }
 
-                            if (originalOutputs != null)
+                        foreach (RewindData rewindData in rewindDataCollection)
+                        {
+                            int nextRewindIndex = 0;
+                            if (this.TryGetRewindIndex(transaction, out int currentRewindIndex))
```
Aprogiena: I thought we said that the index comes from the block height, so no longer we should have such a logic here

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -250,40 +231,26 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
...
                         this.SetBlockHash(transaction, nextBlockHash);
 
-                        all.Sort(UnspentOutputsComparer.Instance);
-                        foreach (UnspentOutputs coin in all)
+                        foreach (UnspentOutputs unspentOutput in unspentOutputs)
                         {
-                            this.logger.LogTrace("Outputs of transaction ID '{0}' are {1} and will be {2} to the database.", coin.TransactionId, coin.IsPrunable ? "PRUNABLE" : "NOT PRUNABLE", coin.IsPrunable ? "removed" : "inserted");
-                            if (coin.IsPrunable) transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
-                            else transaction.Insert("Coins", coin.TransactionId.ToBytes(false), coin.ToCoins());
+                            this.logger.LogTrace("Outputs of transaction ID '{transactionId}' are {prunableState} and will be {action} to the database.", unspentOutput.TransactionId, unspentOutput.IsPrunable ? "PRUNABLE" : "NOT PRUNABLE", unspentOutput.IsPrunable ? "removed" : "inserted");
```
Aprogiena: No interpolated strings in logs

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
ferdeen: I debugged the issue, and @fenix2222 sent the code fix over so I could test locally.  Happy to change, but would like confirmation from @fenix2222 first before changing.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -297,21 +264,27 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
         }
 
         /// <summary>
-        /// Obtains order number of the last saved rewind state in the database.
+        /// Attempts to obtain order number of the last saved rewind state in the database.
```
Aprogiena: Why do we have this considering the change in the logic?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/InMemoryCoinView.cs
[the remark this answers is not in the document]
Aprogiena: It is really weird that in memory coin view is implementing persist method which is not empty.

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
```diff
@@ -18,5 +21,40 @@ public static class Extensions
 
             return await Task.FromResult(true);
         }
+
+        public static long GetSizeInBytes(this UnspentOutputs unspentOutputs)
```
Aprogiena: This is a very unfortunate placement of such methods. Just imagine that serialisation methods would be implemented here with extensions. It completely disconnects from the object itself and that object is changed you will very likely change the serialisation method because you see it in the class in the same file, but I can guarantee you that you will never find this one so you will not change it.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
```diff
@@ -22,18 +21,14 @@ public class SaveCoinviewRule : UtxoStoreConsensusRule
...
-
-            bool forceFlush = this.FlushRequired(context.ValidationContext.ChainedHeaderToValidate);
-            if (this.PowParent.UtxoSet is CachedCoinView cachedCoinView)
-                await cachedCoinView.FlushAsync(forceFlush).ConfigureAwait(false);
+            var utxoRuleContext = (UtxoRuleContext)context;
+            List<UnspentOutputs> unspentOutputs = utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet)?.ToList();
+            await this.PowParent.UtxoSet.AddRewindDataAsync(unspentOutputs, currentBlock).ConfigureAwait(false);
         }
 
         /// <summary>
```
Aprogiena: So this method is not used anymore?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
```diff
@@ -22,18 +21,14 @@ public class SaveCoinviewRule : UtxoStoreConsensusRule
         /// <inheritdoc />
```
Aprogiena: The threshold above, not used any more?

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/SmartContractSaveCoinviewRule.cs
```diff
@@ -12,14 +15,14 @@ public sealed class SmartContractSaveCoinviewRule : UtxoStoreConsensusRule
...
+            ChainedHeader currentBlock = context.ValidationContext.ChainedHeaderToValidate;
 
             // Persist the changes to the coinview. This will likely only be stored in memory,
-            // unless the coinview treashold is reached.
+            // unless the coinview threshold is reached.
             this.Logger.LogTrace("Saving coinview changes.");
-            var utxoRuleContext = context as UtxoRuleContext;
-            await this.PowParent.UtxoSet.SaveChangesAsync(utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), null, oldBlockHash, nextBlockHash).ConfigureAwait(false);
+            var utxoRuleContext = (UtxoRuleContext)context;
+            List<UnspentOutputs> unspentOutputs = utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet)?.ToList();
```
Aprogiena: GetCoins - Is it possible for that to return null?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
```diff
@@ -22,18 +21,14 @@ public class SaveCoinviewRule : UtxoStoreConsensusRule
...
+            // unless the coinview threshold is reached.
             this.Logger.LogTrace("Saving coinview changes.");
-            var utxoRuleContext = context as UtxoRuleContext;
-            await this.PowParent.UtxoSet.SaveChangesAsync(utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet), null, oldBlockHash, nextBlockHash).ConfigureAwait(false);
-
-            bool forceFlush = this.FlushRequired(context.ValidationContext.ChainedHeaderToValidate);
-            if (this.PowParent.UtxoSet is CachedCoinView cachedCoinView)
-                await cachedCoinView.FlushAsync(forceFlush).ConfigureAwait(false);
+            var utxoRuleContext = (UtxoRuleContext)context;
+            List<UnspentOutputs> unspentOutputs = utxoRuleContext.UnspentOutputSet.GetCoins(this.PowParent.UtxoSet)?.ToList();
```
Aprogiena: can it be null?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
fenix2222: I am not sure what benefit we add by using block height instead of an index. Rewind does not accept block height as a parameter, so we don't know what block to rewind and will just rewind the last item from the storage. That would also mean that height will need to be added to `RewindData` object as it doesn't have it atm or possible infer it from `OutputsToRestore` if there are any.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
fenix2222: There are no interpolated strings here. Using NLog structured logging feature

src/Stratis.Bitcoin.Features.Consensus/Extensions.cs
fenix2222: Moved each one of those to a corresponding object

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
fenix2222: Correct. There is no more manual flushing. It is handled by internal task, same as BSQ

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/LoadCoinviewRule.cs
fenix2222: Yes, I will remove it

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -56,108 +71,109 @@ private class CacheItem
...
-        /// <seealso cref="lastCacheFlushTime"/>
-        public const int CacheFlushTimeIntervalSeconds = 60;
+        /// <summary>Maximum number of bytes the batch can hold until the rewind data items are stored to the disk.</summary>
+        internal const int BatchThresholdSizeBytes = 5 * 1_000 * 1_000;
 
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
+        /// <summary>Queue which contains rewind data items that should be saved to the database.</summary>
+        private readonly AsyncQueue<QueuedRewindData> rewindDataQueue;
```
dangershony: This is a bit alarming to me, why do we need an asyncqueue in the utxo store layer?

PR #1928, opened 2018-08-15
Integration test improvements

ArcSin2000X:
Folders for test data are now created with unique names to prevent problems with delete. Improvements to  TestHelper to preven locking main thread. Increased default time for WaitLoop freaquency


src/Stratis.Bitcoin.Tests.Common/TestBase.cs
```diff
@@ -25,31 +25,10 @@ public TestBase(Network network)
...
-                else
-                    break;
-            }
-
-            if (deleteAttempts >= 50)
-                throw new Exception(string.Format("The test folder: {0} could not be deleted.", dir));
-
-            Directory.CreateDirectory(dir);
-            return dir;
+            long ticks = DateTime.UtcNow.Ticks;
```
zeptin: This seems like it is going to get really messy; how do you suggest we ensure the created directories don't accumulate forever? On the CI I assume the folders for each test run are isolated from each other regardless, so this is mostly for local tests?

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -11,17 +12,23 @@ namespace Stratis.Bitcoin.IntegrationTests.Common
...
+        static async Task AsyncTaskDelay(int milisec)
+        {
+            await Task.Delay(milisec).ConfigureAwait(false);
+        }
+
+        public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reason", int millisecondsTimeout = 150, CancellationToken cancellationToken = default(CancellationToken))
         {
             cancellationToken = cancellationToken == default(CancellationToken)
-                ? new CancellationTokenSource(Debugger.IsAttached ? 15 * 60 * 1000 : 30 * 1000).Token
+                ? new CancellationTokenSource(Debugger.IsAttached ? 15 * 60 * 1000 : 40 * 1000).Token
```
zeptin: Can you give a bit of detail as to how you arrived at this particular choice of timeout? I also find that tests seem to be more reliable the longer this interval is, but perhaps the underlying tests should be corrected instead? Do we know which ones are particularly problematic?

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
```diff
@@ -25,31 +25,10 @@ public TestBase(Network network)
...
+            string uniqueDirName = dir + ticks.ToString();
```
bokobza: ticks is not very readable. It'd be hard to quickly see the results of the tests you ran a couple of days ago.
I prefer `$"{dir}-{DateTime.UtcNow.ToString("dd/MM/yyyyTHH:mm:ss.fff")}"` which will give you folders like _forldername_-15/08/2018T15:40:43.144

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -11,17 +12,23 @@ namespace Stratis.Bitcoin.IntegrationTests.Common
...
                 : cancellationToken;
+
             while (!act())
             {
                 try
                 {
                     cancellationToken.ThrowIfCancellationRequested();
-                    Thread.Sleep(millisecondsTimeout);
+                    AsyncTaskDelay(millisecondsTimeout).Wait();
```
bokobza: So I think the `Wait()` here defeats the purpose of having a `Task.Delay` as the task is waited on. 
Running the code, I saw no difference in speed or reliability.

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -11,17 +12,23 @@ namespace Stratis.Bitcoin.IntegrationTests.Common
 {
     public class TestHelper
     {
-        public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reason", int millisecondsTimeout = 50, CancellationToken cancellationToken = default(CancellationToken))
+        static async Task AsyncTaskDelay(int milisec)
```
ferdeen: `millisecs` ?

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
ArcSin2000X: 2 use cases:
1) you run tests on server after build, according to @bokobza you always start with empty dir, that way its not messy
2) you run locally, we will need to housekeep directories, I found a lot of problems when run locally and trying to delete dirs from C# this way you are guaranteed new folder every time

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
bokobza: https://stackoverflow.com/questions/34052381/thread-sleep2500-vs-task-delay2500-wait

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
zeptin: +1, but you may need to use underscores etc in lieu of characters that are invalid in folder names.

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
ArcSin2000X: we managed to isolate around 12 tests that cause problems if executed in integration test suite. I think given large peformance profile of the test its just giving extra 10 seconds for the loop to spin, additionally I decreased the frequency at which we invoke methods

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
ArcSin2000X: accept

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
ArcSin2000X: accept

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
ArcSin2000X: I stand corrected

codingupastorm:
Big fan of this. 

I think we should create an issue to address the underlying problems with tests failing in the current times - some of them are smart contract ones (sorry), but for an interim solution I like the idea of this.

src/Stratis.Bitcoin.Tests.Common/TestBase.cs
```diff
@@ -25,31 +25,9 @@ public TestBase(Network network)
 
         public static string AssureEmptyDir(string dir)
         {
-            int deleteAttempts = 0;
```
monsieurleberre: Would it be possible / desirable to clean folders from previous runs ? I can help you write the code if the is something we want to do 👍

PR #1933, opened 2018-08-16
[SC] Remove SmartContractConstructor

rowandh:
Removes `SmartContractConstructor`. It has been replaced by `Contract`.


src/Stratis.SmartContracts.Executor.Reflection/Stratis.SmartContracts.Executor.Reflection.csproj
```diff
@@ -13,4 +13,8 @@
     <ProjectReference Include="..\Stratis.SmartContracts.Core\Stratis.SmartContracts.Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
codingupastorm: Think this isn't meant to be here? Got a little folder with an x next to it in VS

Issue #1131, opened 2018-02-03
Restarting node has connection issues


mikedennis:
I stopped my node updated my source and restarted and after restarting it had problems connecting to nodes (see log entries below). I then stopped the node and restarted it and it then was able to connect to nodes.

```
14:37:57.2099 15] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0.<StartAsync>b__0 DiscoverPeersAsync starting
[2018-02-03 14:37:57.2264 1] INFO: Stratis.Bitcoin.Connection.ConnectionManager.StartNodeServer Node listening on:
0.0.0.0:26178

[2018-02-03 14:37:58.1857 16] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0.<StartAsync>b__0 PeerConnectorAddNode.ConnectAsync starting
[2018-02-03 14:37:58.3643 16] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0.<StartAsync>b__0 PeerConnectorDiscovery.ConnectAsync starting
[2018-02-03 14:38:01.1841 19] INFO: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0.<StartAsync>b__0 PeriodicLog starting
[2018-02-03 14:38:02.1974 20] INFO: Stratis.Bitcoin.Connection.ConnectionManagerBehavior.OnStateChangedAsync Peer '[::ffff:93.157.168.237]:62031' connected (inbound), agent 'StratisBitcoin:1.0.7', height 0
[2018-02-03 14:38:02.2420 20] INFO: Stratis.Bitcoin.Connection.ConnectionManagerBehavior.OnStateChangedAsync Peer '[::ffff:52.233.33.138]:51697' connected (inbound), agent 'StratisBitcoin:1.0.7', height 0
[2018-02-03 14:38:03.2031 21] INFO: Stratis.Bitcoin.Connection.ConnectionManagerBehavior.OnStateChangedAsync Peer '[::ffff:13.92.132.44]:26178' offline, reason: 'unknown'.
[2018-02-03 14:38:04.1845 6] INFO: Stratis.Bitcoin.Connection.ConnectionManagerBehavior.OnStateChangedAsync Peer '[::ffff:104.215.26.11]:26178' offline, reason: 'Unexpected exception while waiting for a message'.
[...]

PR #1946, opened 2018-08-17
Add delay & parallelise node connections

zeptin:
For #1131

### DIFF
--- a/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
+++ b/src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
@@ -1,4 +1,5 @@
-﻿using System.Net;
+﻿using System;
+using System.Net;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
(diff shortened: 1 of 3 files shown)


ferdeen:
Hi Kevin,

Could the `OnConnectAsync()` code logic move from both classes - `PeerConnectorAddNode` & `PeerConnectorConnect`, into a shared class as it's the same code ?

Cheers.

zeptin:
Hi @ferdeen, note that there are three classes that override the abstract `OnConnectAsync` method (the third being `PeerConnectorDiscovery`). The `PeerConnectorDiscovery` logic is quite different to the `-connect` and `-addnode` connectors, so it would probably just clutter up the inheritance trying to avoid this one instance of duplication.

Also note that the design of how `-connect` and `-addnode` get handled may further diverge in the future, in which case the method implementations would need to remain separate. But if there is general agreement that the duplication should be removed then I will do it.

fassadlr:
@zeptin the most immediately obvious way to stop high CPU usage in that `foreach` would be to disable burst connections for the add node and connect node peer connectors.

I would rather introduce a burst mode enabled property that is disabled for the add node and connect connectors (this will be set in each connector's `OnInitialize` override) and set the burst mode interval to the default 1 second on start up.

We shouldnt introduce extra logic in the OnConnect to check for last connections attempts etc. This should be (and is) the responsibility of the `PeerSelector` class. We already filter peer addresses by not selecting them if they failed connection less than a minute ago.

So the actual problem is the `foreach` constantly executing without a delay because of the burst mode interval which only gets set to 1 second after we've had one successful connection.

ferdeen:
Thanks @zeptin I understand now.  If they need to stay separate then ignore my comment.

Cheers.

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -61,15 +62,22 @@ public override void OnStartConnect()
...
+            await this.ConnectionSettings.AddNode.ForEachAsync(this.ConnectionSettings.MaxOutboundConnections, this.nodeLifetime.ApplicationStopping,
+                async (ipEndpoint, cancellation) =>
+                {
+                    if (!this.nodeLifetime.ApplicationStopping.IsCancellationRequested)
+                    {
+                        PeerAddress peerAddress = this.peerAddressManager.FindPeer(ipEndpoint);
+                        if (peerAddress != null && !this.IsPeerConnected(peerAddress.Endpoint))
+                        {
+                            // Introduce a delay between attempts in case ConnectAsync fails instantly without the usual timeout.
+                            if ((peerAddress.LastAttempt == null) || ((peerAddress.LastAttempt - DateTimeOffset.Now) >= this.defaultConnectionInterval))
```
Aprogiena: Date time provider instead of immediate value

src/Stratis.Bitcoin/P2P/PeerConnectorAddNode.cs
```diff
@@ -70,11 +76,7 @@ public override async Task OnConnectAsync()
                         PeerAddress peerAddress = this.peerAddressManager.FindPeer(ipEndpoint);
                         if (peerAddress != null && !this.IsPeerConnected(peerAddress.Endpoint))
                         {
-                            // Introduce a delay between attempts in case ConnectAsync fails instantly without the usual timeout.
```
fassadlr: Ah Zeptin can we revert so that we return from if cancellation was requested? So that we eliminate the nested if.  Same with the other connector.

PR #1947, opened 2018-08-17
[Activation] block store behavior fix

noescape00:


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -395,10 +395,14 @@ public async Task AnnounceBlocksAsync(List<ChainedHeader> blocksToAnnounce)
                 {
                     bool foundStartingHeader = false;
 
+                    // In case we don't have any information about peer's tip send him only last header and don't update best sent header.
+                    // We expect peer to answer with getheaders message.
```
dangershony: Brilliant this will trigger the sync any way right?

noescape00: Yes. Peer will get an unknow header and will call TrySync which will send getheaders msg.

PR #1949, opened 2018-08-17
[Activation] Minimal viable rewind data refactor

fenix2222:
This is a refactor of coin views to ensure rewind data can be aggregated per block as per #1810

Unit tests will be fixed in a separate PR to make this PR smaller

The approach is as follows:

1. Construct rewind data list in cache and keep creating rewind data per block.
2. Once cache is flushed we pass this list to DBreezeCoinView to be persisted.
3. When rewind is requested, check cache first and if it is available in cache, rewind it from there. Otherwise, attempt to rewind from the DBreezeCoinView.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
@@ -416,7 +416,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 
             using (await this.lockobj.LockAsync().ConfigureAwait(false))
             {
-                // Check if rewind data is available in local cache. If it is 
+                // Check if rewind data is available in local cache. If it is
                 // we can rewind and there is no need to check underlying storage.
                 if (this.cachedRewindDataList.Count > 0)
                 {
@@ -450,14 +450,22 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
             foreach (UnspentOutputs unspentToRestore in rewindData.OutputsToRestore)
             {
                 this.logger.LogTrace("Outputs of transaction ID '{0}' will be restored.", unspentToRestore.TransactionId);
+
                 if (this.unspents.TryGetValue(unspentToRestore.TransactionId, out CacheItem cacheItem))
                 {
                     cacheItem.UnspentOutputs = unspentToRestore;
+                    cacheItem.IsDirty = true;
                 }
                 else
                 {
                     this.logger.LogTrace("Outputs of transaction ID '{0}' not found in cache, inserting them.", unspentToRestore.TransactionId);
-                    cacheItem = new CacheItem { UnspentOutputs = unspentToRestore };
+
+                    cacheItem = new CacheItem
+                    {
+                        UnspentOutputs = unspentToRestore,
+                        IsDirty = true
+                    };
+
                     this.unspents.Add(unspentToRestore.TransactionId, cacheItem);
                 }
             }


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -85,6 +85,9 @@ private class CacheItem
         /// <summary>Coin view at one layer below this implementaiton.</summary>
         private readonly ICoinView inner;
 
+        /// <summary>Pending list of rewind data to be persisted to a persistent storage.</summary>
+        private readonly List<RewindData> cachedRewindDataList = new List<RewindData>();
```
Aprogiena: Code style – initialisation goes into constructor

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -85,6 +85,9 @@ private class CacheItem
         /// <summary>Coin view at one layer below this implementaiton.</summary>
         private readonly ICoinView inner;
 
+        /// <summary>Pending list of rewind data to be persisted to a persistent storage.</summary>
```
Aprogiena: Add a remarks section saying that access to this object has to be protected with lock object. See other fields on the class for similar remarks section.

Also expand lock objects comment to include this new contract

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -356,11 +360,17 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
+
                 foreach (UnspentOutputs unspent in unspentOutputs)
                 {
-                    CacheItem existing;
-                    if (this.unspents.TryGetValue(unspent.TransactionId, out existing))
+                    if (this.unspents.TryGetValue(unspent.TransactionId, out CacheItem existing))
                     {
+                        // We'll need to restore the original outputs.
+                        UnspentOutputs clone = unspent.Clone();
+                        clone.Outputs = existing.UnspentOutputs?.Outputs.ToArray() ?? Array.Empty<TxOut>();
```
Aprogiena: Is this correct? I'm not sure you should be using `existing` here is that seems to be an old version of the structure. Should you not use `unspent`? The previous line made a full copy of `unspent`and my question is why would we want to change anything about that. The outputs were copied and now you're replacing that with something else. I don't really understand why. If you look down, you can see that we do even modify this `existing` thing, which would suggest that whatever you have on this line cannot be correct. But I'm not sure

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -401,16 +416,21 @@ public async Task<uint256> Rewind()
                 if (this.blockHash == this.innerBlockHash)
                     this.unspents.Clear();
 
-                if (this.unspents.Count != 0)
+                // Check if rewind data is available in local cache. If it is 
+                // we can rewind and there is no need to check underlying storage.
+                if (this.cachedRewindDataList.Any())
```
Aprogiena: Better to use Count on lists

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -420,6 +440,36 @@ public async Task<uint256> Rewind()
             }
         }
 
+        private void RestoreOutputs(RewindData rewindData)
```
Aprogiena: We need entry and exit logs for both of these new methods

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -420,6 +440,36 @@ public async Task<uint256> Rewind()
...
+            foreach (UnspentOutputs unspentToRestore in rewindData.OutputsToRestore)
+            {
+                this.logger.LogTrace("Outputs of transaction ID '{0}' will be restored.", unspentToRestore.TransactionId);
+                if (this.unspents.TryGetValue(unspentToRestore.TransactionId, out CacheItem existing))
+                {
+                    existing.UnspentOutputs = unspentToRestore;
+                }
+                else
+                {
+                    this.logger.LogTrace(
```
Aprogiena: Does not really deserve to be on three lines

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -420,6 +440,36 @@ public async Task<uint256> Rewind()
...
+                }
+            }
+        }
+
+        private void RemoveTransactions(RewindData rewindData)
+        {
+            foreach (uint256 transactionToRemove in rewindData.TransactionsToRemove)
+            {
+                this.logger.LogTrace("Attempt to remove transaction with ID '{0}'.", transactionToRemove);
+                if (!this.unspents.ContainsKey(transactionToRemove)) continue;
```
Aprogiena: Empty line after continue please

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CoinView.cs
```diff
@@ -34,7 +34,8 @@ public interface ICoinView
         /// and which were not.</param>
         /// <param name="oldBlockHash">Block hash of the current tip of the coinview.</param>
         /// <param name="nextBlockHash">Block hash of the tip of the coinview after the change is applied.</param>
-        Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash, uint256 nextBlockHash);
+        /// <param name="rewindDataList">List of rewind data items to be persisted. This should only be used when calling <see cref="DBreezeCoinView.SaveChangesAsync" /></param>
```
Aprogiena: Missing `.`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: So do I understand that correctly –  `existing` is the old version before the new block was applied and so we want to have the old version of outputs array in the rewind data?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -401,16 +416,21 @@ public async Task<uint256> Rewind()
                 if (this.blockHash == this.innerBlockHash)
```
Aprogiena: This whole segment I don't understand. Could you please explain the logic here? For example why this if stayed in the code and the other if was removed?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -401,16 +416,21 @@ public async Task<uint256> Rewind()
...
+                    RewindData lastRewindData = this.cachedRewindDataList.Last();
+
+                    this.RemoveTransactions(lastRewindData);
+                    this.RestoreOutputs(lastRewindData);
 
-                    this.logger.LogTrace("(-)[REWOUND_TO_INNER]:'{0}'", this.blockHash);
-                    return this.blockHash;
+                    this.cachedRewindDataList.Remove(lastRewindData);
+                    this.logger.LogTrace("(-)[REMOVED_FROM_BATCH]:'{0}'", lastRewindData.PreviousBlockHash);
+                    return lastRewindData.PreviousBlockHash;
```
Aprogiena: For example here, I don't understand, why don't we need to assign new value into this.blockHash

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -401,16 +416,21 @@ public async Task<uint256> Rewind()
...
-                    this.unspents.Clear();
-                    this.blockHash = this.innerBlockHash;
+                    RewindData lastRewindData = this.cachedRewindDataList.Last();
+
+                    this.RemoveTransactions(lastRewindData);
+                    this.RestoreOutputs(lastRewindData);
 
-                    this.logger.LogTrace("(-)[REWOUND_TO_INNER]:'{0}'", this.blockHash);
-                    return this.blockHash;
+                    this.cachedRewindDataList.Remove(lastRewindData);
```
Aprogiena: RemoveAt

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -420,6 +440,36 @@ public async Task<uint256> Rewind()
...
+                if (this.unspents.TryGetValue(unspentToRestore.TransactionId, out CacheItem existing))
+                {
+                    existing.UnspentOutputs = unspentToRestore;
+                }
+                else
+                {
+                    this.logger.LogTrace(
+                        "Outputs of transaction ID '{0}' not found in cache, inserting them.",
+                        unspentToRestore.TransactionId);
+                    existing = new CacheItem { UnspentOutputs = unspentToRestore };
```
Aprogiena: Not really existing, consider choosing better name and new variable

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: And also – can it even happen? Is it possible that the transaction that should be removed would not be found?

Aprogiena: In any case, the remove method will not throw if the element does not exist, so you don't need this line. The only question is whether we should check if remove returns false and if that is an error condition or not

Aprogiena:
@dangershony @noescape00 Could you guys please check the logic here, this needs as many eyes as possible

bokobza:
I'll have a look at this tomorrow.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: Yes, this is my understanding as well. We want to get original outputs before we mutate it.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: That shouldn't be here, good spotting. I forgot to set `blockHash`. Changed now

fenix2222:
Updated all as per feedback. Need more eyes on it.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: done

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: done

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -398,19 +416,24 @@ public async Task<uint256> Rewind()
...
 
-                    this.logger.LogTrace("(-)[REWOUND_TO_INNER]:'{0}'", this.blockHash);
+                    this.RemoveTransactions(lastRewindData);
+                    this.RestoreOutputs(lastRewindData);
+
+                    // Change current block hash to the one from the rewind data.
+                    this.blockHash = lastRewindData.PreviousBlockHash;
+
+                    this.cachedRewindDataList.RemoveAt(this.cachedRewindDataList.Count - 1);
+                    this.logger.LogTrace("(-)[REMOVED_FROM_BATCH]:'{0}'", lastRewindData.PreviousBlockHash);
```
Aprogiena: The argument here should match returned object, just cosmetics

noescape00: Unless there could be a race condition and it's possible that there is an another thread that might touch `this.blockHash`

bokobza:
great work guys!

PR #1950, opened 2018-08-18
[SC] Add limp receipt root to smart contract block headers.

codingupastorm:


codingupastorm:
All SC + Integration tests pass.

Issue #1934, opened 2018-08-16
[Activation] PowBlockDefinition - ensure we are calling validator correctly


noescape00:
There is a TODO in PowBlockDefinition.cs, fix it

herbepau:
Following a discussion with Ivan and Apro, it was determined that the method TestBlockValidity is not required and can be removed from the codebase.  This is because block validation is already happening during the mining phase in ConsensusManager.
Also, this is a sanity check and should never happen, and that not removing it from here would lead to performing partial validation twice.

bokobza:
fixed with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1987

PR #1951, opened 2018-08-18
[Activation] Fix ChainState's IsAtBestTip

fassadlr:
@Aprogiena @noescape00 Ive had to remove the exception in PowBlockDefinition to check that the tests pass. Can we create an issue so that we remember?


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1028,6 +1028,8 @@ private void AddOrReplacePeerTip(int networkPeerId, uint256 newTip)
 
             this.peerTipsByPeerId.Add(networkPeerId, newTip);
 
+            this.chainState.IsAtBestChainTip = this.IsAtBestChainTip();
```
Aprogiena: I think this does not belong here, this should be in consensus manager. It is missing there as we only have eight in a single place – when consensus tip changes and it is missing when the peer presents new header

Aprogiena:
I think this is good find, definitely we are missing something, but at the same time and quite convinced that we need to do more to fix a problem. I would like to fully understand the issue first.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: It's already in CM in ConnectBlockAsync after we call CHT.FullValidationSucceeded

src/Stratis.Bitcoin.Features.Miner/PowBlockDefinition.cs
```diff
@@ -77,7 +74,6 @@ public void TestBlockValidity()
         {
             this.logger.LogTrace("()");
 
-            throw new Exception("FIX ME");
```
noescape00: Why removing this? 
It was left here so we don't forget about it while fixing the tests

fassadlr: @noescape00 I removed this as it causes the mining integration tests to fail. Can we create an issue for this? Or does this link to #1934 ?

noescape00: Yes, issue #1934 is for that. 

>I removed this as it causes the mining integration tests to fail. 

Ok

noescape00:
@fassadlr I don't understand why we need to change this. 

IsAtBestChainTip is used only in BSQ at the moment in that check where we are flushing the batch.

Aprogiena:
@noescape00 So the idea is that before we considered to be at tip when the consensus was the last block anyone claimed, but imagine that the new block was mined on the network so for a fraction of second before we download and validate and connect that a block, we would consider ourselves is out of sync. But the purpose of this value on the chain state is not to be perfectly accurate. Rather than that we are just interested to know whether we should be in the mode of saving everything one by one. And there are various race condition that could make a lot of things harder to implement or harder to test, if the value reflects exactly the correct state at every given point of time. Instead we go now with this range of heights that are considered to be at tip regardless of occasional events such as new block being delivered or even a short reorg being presented.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1204,10 +1209,21 @@ private void CheckMaxReorgRuleViolated(ChainedHeader chainedHeader)
         }
 
         /// <inheritdoc />
-        public bool IsAtBestChainTip()
+        public bool IsConsensusConsideredToBeSynced()
         {
             this.logger.LogTrace("()");
 
+            // If the height of consensus is less than the tip margin, the node is considered to be synced.
```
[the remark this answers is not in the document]
fassadlr: Yes you are right, this should be the opposite. However this would mean that in all the integration we would have to mine at least 5 blocks before we persist? What is youe thoughts on this?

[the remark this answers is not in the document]
Aprogiena: How so? Let's say there are four blocks on the network and the node connects to it having zero blocks. We just need to make sure that the expression below – the normal one – compares correctly:

`our height + threshold > best = 0 + 5 > 4`

=> yes => save even the first block

so instead of minus, use plus below to avoid overflow and negative numbers and it should be fine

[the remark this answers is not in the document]
Aprogiena: So I think we can remove this part of the code and instead just have a good equation down below and we should be fine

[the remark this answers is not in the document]
fassadlr: Ok yeah, so we essentially invert the expression 👍 DONE.

PR #1958, opened 2018-08-20
[SC] Consensus components of receipt root implementation

codingupastorm:
Quite a few stubs in there, and a bit of repetition, but provides some kind of skeleton off which to build the rest of the receipts implementation. 

All SC and integration tests pass.


src/Stratis.SmartContracts.Core/BloomData.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public byte[] ToBytes()
+        {
+            return this.data;
+        }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as BloomData;
+            return other != null && Enumerable.SequenceEqual(this.data, other.data);
```
rowandh: Do we want to override == as well?

codingupastorm: Good call. Updated.

src/Stratis.SmartContracts.Core/BloomData.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public byte[] ToBytes()
+        {
+            return this.data;
+        }
+
+        public static bool operator ==(BloomData obj1, BloomData obj2)
+        {
+            if (object.ReferenceEquals(obj1, null))
```
rowandh: Nice

PR #1961, opened 2018-08-20
[SC] Bloom filter for receipts

codingupastorm:
Should go in after #1958


src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/SmartContractCoinviewRule.cs
```diff
@@ -58,6 +60,8 @@ public async override Task RunAsync(RuleContext context)
             this.ContractCoinviewRule.OriginalStateRoot.SyncToRoot(((SmartContractBlockHeader)context.ConsensusTip.Header).HashStateRoot.ToBytes());
             IContractStateRepository trackedState = this.ContractCoinviewRule.OriginalStateRoot.StartTracking();
 
+            this.receipts = new List<Receipt>();
```
rowandh: Not a fan of this being an instance field because this object is reused, but I think the alternative is hard/impossible to implement due to the way inheritance works on these objects?

PR #1966, opened 2018-08-20
Block to chainedheaderblock

### LINKED ISSUE #1813: Change signaling from Block to ChainedHeaderBlock
https://github.com/stratisproject/StratisBitcoinFullNode/issues/1768

Can be done at same time as #1814

majikandy:
fixes #1813 

A phenomenal amount of unit tests are breaking but cannot be related to this change so must be breaking in core-way-activation.

Obviously don't merge until that is resolved

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
@@ -1,6 +1,7 @@
 ﻿using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
+using Stratis.Bitcoin.Primitives;
 using Stratis.Bitcoin.Signals;
 
 namespace Stratis.Bitcoin.Features.MemoryPool
@@ -8,7 +9,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
     /// <summary>
     /// Mempool observer on disconnected block notifications.
     /// </summary>
-    public class BlocksDisconnectedSignaled : SignalObserver<Block>
+    public class BlocksDisconnectedSignaled : SignalObserver<ChainedHeaderBlock>
     {
         private readonly IMempoolValidator mempoolValidator;
         private readonly MempoolSchedulerLock mempoolLock;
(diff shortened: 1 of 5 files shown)


noescape00:
OnBlockConnected also should signal chained header block instead of Block

majikandy:
The issue only mentioned block disconnected, but happy to do block connected too if a priority?

noescape00:
>The issue only mentioned block disconnected

The issue doesn't specify it. There is basically no description except for the issue name and issue name suggests that block signaling in general should be changed to signal CHB, 

>but happy to do block connected too if a priority?

Yes, please do.

majikandy:
Ok this is now done.  I have run the unit tests and compared against  core-way-activation that the same number are still failing (that aren't relevant to this change).

src/Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
```diff
@@ -107,13 +108,18 @@ public void BenchStats()
...
-                    this.BenchStats();
-            }
+            if (this.LastSnapshotWasWithinLast5Seconds()) 
+                return;
+            
+            if (this.initialBlockDownloadState.IsInitialBlockDownload())
+                this.BenchStats();
+        }
+
+        private bool LastSnapshotWasWithinLast5Seconds()
```
noescape00: Please move this logic back to if because it's outside of the scope of the PR and ConsensusStats will be completely rewritten after activation is merged

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -22,11 +23,11 @@ public class BlocksDisconnectedSignaled : SignalObserver<Block>
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
         }
 
-        protected override void OnNextCore(Block block)
+        protected override void OnNextCore(ChainedHeaderBlock block)
         {
-            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.GetHash());
+            this.logger.LogTrace("({0}:'{1}')", nameof(block), block.Block.GetHash());
```
noescape00: ChainedHeaderBlock overrides .ToString()
log the whole ChainedHeaderBlock, no need for `block.Block.GetHash()`

noescape00:
Looks good to me. Just a few minor changes

src/Stratis.Bitcoin.Features.Consensus/ConsensusStats.cs
majikandy: In that case, there is no harm in it staying, but I see your point and will revert anyway.

majikandy:
Thanks @noescape00 These changes are now done.

PR #1976, opened 2018-08-21
[Activation] Fix mempool integration tests

fassadlr:
This PR fixes some issues in the mempool integration tests as well making the `ConnectBlocksResult` more intuitive. The main issue was that the 'IsNodeSynced` method was checking the wallet tip hash with nodes that wasn't started with a wallet.

The second was a null reference exception where the `Error` property on the `ConnectBlocksResult` was not being set properly.

image


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -33,20 +33,42 @@ public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reas
...
+            if (node1.FullNode.ChainBehaviorState.ConsensusTip.HashBlock != node2.FullNode.ChainBehaviorState.ConsensusTip.HashBlock)
+                return false;
+
+            if (node1.FullNode.GetBlockStoreTip().HashBlock != node2.FullNode.GetBlockStoreTip().HashBlock)
+                return false;
+
+            if (node1.FullNode.MempoolManager().InfoAll().Count != node2.FullNode.MempoolManager().InfoAll().Count)
+                return false;
+
+            if ((node1.FullNode.WalletManager().Wallets.Count != 0) && (node2.FullNode.WalletManager().Wallets.Count != 0))
```
bokobza: you can do `node1.FullNode.WalletManager().ContainsWallets`

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -33,20 +33,42 @@ public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reas
...
+
             return true;
         }
 
         public static bool IsNodeSynced(CoreNode node)
         {
-            if (node.FullNode.Chain.Tip.HashBlock != node.FullNode.ChainBehaviorState.ConsensusTip.HashBlock) return false;
-            if (node.FullNode.Chain.Tip.HashBlock != node.FullNode.GetBlockStoreTip().HashBlock) return false;
-            if (node.FullNode.Chain.Tip.HashBlock != node.FullNode.WalletManager().WalletTipHash) return false;
+            if (node.FullNode.Chain.Tip.HashBlock != node.FullNode.ChainBehaviorState.ConsensusTip.HashBlock)
```
bokobza: I think something was missed in the merge, as @ferdeen already fixed this in https://github.com/stratisproject/StratisBitcoinFullNode/commit/dd1056f6ea117c978106d3ec5a12875419e398fc

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -532,7 +532,7 @@ private async Task<ConnectBlocksResult> FullyValidateLockedAsync(ChainedHeader n
                 // In that case we try to load the blocks from store, if store is not present we disconnect all peers.
                 this.HandleMissingBlocksGap(currentTip);
 
-                var result = new ConnectBlocksResult(false);
+                var result = ConnectBlocksResult.Fail();
```
bokobza: Not sure this is actually clearer. It reads like you're making it fail, rather than expressing that "it's failed".

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
bokobza: It seems this was missed during the merge 😢

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
fassadlr: OK will do :)

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
fassadlr: All good, at least we found the same problem 👍

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Will rename to `Failed`

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -532,7 +532,7 @@ private async Task<ConnectBlocksResult> FullyValidateLockedAsync(ChainedHeader n
                 // In that case we try to load the blocks from store, if store is not present we disconnect all peers.
                 this.HandleMissingBlocksGap(currentTip);
 
-                var result = new ConnectBlocksResult(false);
+                var result = ConnectBlocksResult.Failed();
```
bokobza: I still find this construct weird.
what was the problem with the previous one?

fassadlr: @bokobza the previous constructor had one required parameter and 4 defaulted parameters which I think is quiet messy and not intuitive at all. It is so easy to NOT understand what constructor to use and this way it is more expressive as to what the intent is of the `ConnectBlocksResult`. It aids in helping the dev choose the right constructor via a static one and improves readability.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -184,31 +184,34 @@ public void TxMempoolBlockDoublespend()
                 }
 
                 // Test 1: block with both of those transactions should be rejected.
-                Block block = stratisNodeSync.GenerateBlockManually(spends);
+                var tipBeforeBlockCreation = stratisNodeSync.FullNode.Chain.Tip;
+                Assert.Throws<ConsensusException>(() => { Block block = stratisNodeSync.GenerateBlockManually(spends); });
                 TestHelper.WaitLoop(() => stratisNodeSync.FullNode.ConsensusManager().Tip.HashBlock == stratisNodeSync.FullNode.Chain.Tip.HashBlock);
```
bokobza: Why do we need to wait for syncing? Nothing has changed since the sync at line 164, no?

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -184,31 +184,34 @@ public void TxMempoolBlockDoublespend()
                 }
 
                 // Test 1: block with both of those transactions should be rejected.
-                Block block = stratisNodeSync.GenerateBlockManually(spends);
+                var tipBeforeBlockCreation = stratisNodeSync.FullNode.Chain.Tip;
+                Assert.Throws<ConsensusException>(() => { Block block = stratisNodeSync.GenerateBlockManually(spends); });
```
bokobza: Did this not throw before?

fassadlr: @bokobza yes prior to activation, we didn't let consensus exceptions bubble to PowMining etc. So now we have to catch it and ensure that the tip didnt advance cause an invalid block was pushed to consensus.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
fassadlr: @bokobza yeah those were there before? I can look at removing them. Let me check quick 👍

fassadlr: @bokobza I checked and I could remove quiet a few one them 👍

src/Stratis.Bitcoin/Consensus/ValidationResults/ConnectBlocksResult.cs
```diff
@@ -16,13 +16,40 @@ internal class ConnectBlocksResult : ValidationResult
...
+        {
+            var result = new ConnectBlocksResult
+            {
+                ConsensusTipChanged = consensusTipChanged,
+                Succeeded = false
+            };
+            return result;
+        }
+
+        public static ConnectBlocksResult FailedAndBanPeers(ConsensusError error, List<int> peersToBan, int banDurationSeconds)
```
bokobza: Should we not have bool `consensusTipChanged = true` here as well?

src/Stratis.Bitcoin/Consensus/ValidationResults/ConnectBlocksResult.cs
```diff
@@ -16,13 +16,40 @@ internal class ConnectBlocksResult : ValidationResult
 
         public ChainedHeader LastValidatedBlockHeader { get; set; }
 
-        public ConnectBlocksResult(bool succeeded, bool consensusTipChanged = true, List<int> peersToBan = null, string banReason = null, int banDurationSeconds = 0)
+        private ConnectBlocksResult() { }
+
+        public static ConnectBlocksResult Failed(bool consensusTipChanged = true)
```
bokobza: I think if we have `Success` then it should be `Failure` rather than `Failed` (my bad for suggesting it).

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
bokobza: I really don't think it's clear. You have method called Success(), a method called Failed() and a method called `FailedAndBanPeers()` where I'm not sure what it does. Does it ban peers?
Something that I think could be clearer is 

`public ConnectBlocksResult(bool success)` + a bunch of properties.

so you can do 
```
var result = new ConnectBlocksResult(false);
result.banDurationSeconds = 222;
result.BanReason = error.Message;
result.ConsensusTipChanged = false;
result.Error = error;
result.PeersToBan = peersToBan;
```

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -685,7 +685,7 @@ private async Task<ConnectBlocksResult> ReconnectOldChainAsync(ChainedHeader cur
 
             if (connectBlockResult.Succeeded)
             {
-                var result = new ConnectBlocksResult(false, false);
```
fassadlr: @Aprogiena @noescape00 can you guys comment here, I think the false here is a bug? I changed it true?

Aprogiena: Why do you think it is a bug? It looks correct as it was. Notice who consumes that result

fassadlr: This `if` is stating that the result succeeded. So why would we return a result with `Suceeded` being false?

Aprogiena: Because those are two different operations. The operation for which we want to return the result failed.

fassadlr: 👍

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
bokobza: great

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
bokobza: makes sense

Issue #1936, opened 2018-08-16
[Activation] fix failing tests


noescape00:
fix unit tests on activation branch. 
this issue can be assigned to multiple developers

noescape00:
@fassadlr is working on integration tests

noescape00:
Upd: currently only around 60 unit tests are failing. 
So we are almost there.

bokobza:
@fenix2222: PowMiningTest
@zeptin : PeerBanningTest
@bokobza: Stratis.Bitcoin.Features.Notifications.Tests ---- in PR
@majikandy : MempoolValidatorTest
@herbepau : BlockPullerTest
@thefazzer : CoinbaseHeightRuleTest
@ferdeen : TransactionLocktimeActivationRuleTest

![image 2](https://user-images.githubusercontent.com/1867877/44585693-147d5280-a7a5-11e8-8db2-1433a96f0fe4.png)

noescape00:
Will fix following ones: 

Stalling_DoesntAffectPeersThatFailedToDeliverNotImportantBlocksAsync
Stalling_PeerStallsButQualityScoreIsTheBestBecausePeerIsTheOnlyOneAsync
ProcessHeadersAsync_PeerThatSentInvalidHeaderIsBannedAsync

noescape00:
Fixed

PR #1979, opened 2018-08-21
[Activation] Fixed ChainRepositoryTest tests 👷‍♂️

bokobza:
Related to https://github.com/stratisproject/StratisBitcoinFullNode/issues/1936

This fixes a couple of tests.


bokobza:
@zeptin We should be able to save the height of the block to the DB without first loading the height of the block from the DB 😁

noescape00:
Yeah, fix makes sense. 
FinalizedBlockInfo should never be null. It's initial state is genesis, not null.

Issue #1837, opened 2018-08-03
[SC] Need to store receipts for all smart contract execution


codingupastorm:
More details in design rationale.

PR #1983, opened 2018-08-22
[SC] Saving logs inside contracts and reaching consensus with receipt root and logs bloom.

codingupastorm:


src/Stratis.SmartContracts.Executor.Reflection/ContractLogging/ContractLogHolder.cs
```diff
@@ -0,0 +1,66 @@
...
+        public ContractLogHolder(IContractPrimitiveSerializer serializer, Network network)
+        {
+            this.serializer = serializer;
+            this.network = network;
+            this.rawLogs = new List<RawLog>();
+        }
+
+        public void Log<T>(ISmartContractState smartContractState, T toLog)
+        {
+            this.rawLogs.Add(new RawLog(smartContractState.Message.ContractAddress.ToUint160(this.network), toLog));
```
rowandh: Because ContractLogHolder is local to the VM create/call methods, could we pass the address in the constructor and avoid the need to pass in the entire state? Will make testing slightly nicer as well.

rowandh:
Looks good, but I don't see any handling of the logs produced for internally generated transactions or internal contract creation. Am I missing something?

codingupastorm:
Closes #1837. Rest will be tracked via #2002 and #2003 and #2004

PR #1985, opened 2018-08-22
[SC] Implement fallback functions

### LINKED ISSUE #1831: [SC] Implement Fallback functions
If a smart contract is called without a method being specified, its fallback should be invoked.

The fallback should be declared via an attribute on a method. 

I believe we have more info available on our internal design rationale.

rowandh:
Closes #1831.

* Adds virtual void fallback function to Stratis.SmartContracts.SmartContract, which can be overridden by contracts to implement fallback processing. This is an API-breaking change.
* Adds logic for resolving and invoking the fallback method.


src/Stratis.SmartContracts.Executor.Reflection/Contract.cs
```diff
@@ -103,13 +130,34 @@ public IContractInvocationResult Invoke(string methodName, IReadOnlyList<object>
...
+        private IContractInvocationResult InvokeFallback()
+        {
+            // Handles the scenario where no fallback was defined, but it is attempted to be invoked anyway.
+            // This could occur if a method invocation is directly made to the fallback via a transaction.
+            if (this.Fallback == null)
+                return ContractInvocationResult.Failure(ContractInvocationErrorType.MethodDoesNotExist);
+
+            EnsureInitialized();
+
+            return this.InvokeInternal(this.Fallback, null);
```
codingupastorm: Could be worth looking into just doing instance.Fallback() if we stick with that approach.

src/Stratis.SmartContracts/SmartContract.cs
```diff
@@ -144,5 +144,13 @@ protected void Assert(bool condition, string message = "Assert failed.")
...
                 throw new SmartContractAssertException(message);
         }
+
+        /// <summary>
+        /// The fallback method. Override this method to define behaviour when the contract receives funds and the method name in the calling transaction equals <see cref="string.Empty"/>.
+        /// <para>
+        /// This occurs when a contract sends funds to another contract using <see cref="Transfer"/>.
+        /// </para>
+        /// </summary>
+        public virtual void Fallback() {}
```
codingupastorm: I like this. Requiring override will prevent misspellings.

src/Stratis.SmartContracts.Executor.Reflection/MethodCall.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <summary>
+        /// Returns true if the method call is a fallback method call.
+        /// </summary>
+        public bool IsFallbackCall
+        {
+            get
+            {
+                // The fallback method must always be the override with no parameters,
+                // so it's not enough just to check if the method name is correct.
+                return (this.Parameters == null || this.Parameters.Length == 0)
```
codingupastorm: I believe this means that a call to "" that has parameters is therefore not going to be considered a fallback call. 

I wonder what we decide here?

src/Stratis.SmartContracts.Executor.Reflection/MethodCall.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// Represents a call to a contract method.
+    /// </summary>
+    public class MethodCall
+    {
+        private readonly string methodName;
+
+        public const string ReceiveHandlerName = nameof(SmartContract.Receive);
+
+        public const string ExternalReceiveHandlerName = "";
```
codingupastorm: Could be good to have brief docs for these 2 for future devs ^

PR #1987, opened 2018-08-22
[Activation] PowBlockDefinition - ensure we are calling validator cor…

herbepau:
…rectly #1934.
@noescape00 @bokobza

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner.Tests/PowBlockAssemblerTest.cs
+++ b/src/Stratis.Bitcoin.Features.Miner.Tests/PowBlockAssemblerTest.cs
@@ -250,33 +250,6 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
         }
 
         [Fact]
-        public void TestBlockValidity_UsesRuleContextToValidateBlock()
-        {
-            var newOptions = new ConsensusOptions();
-
-            this.ExecuteWithConsensusOptions(newOptions, () =>
-            {
-                ConcurrentChain chain = GenerateChainWithHeight(5, this.testNet, new Key());
-                this.consensusManager.Setup(c => c.Tip).Returns(chain.GetBlock(5));
-
-                ValidationContext validationContext = null;
-                var powRuleContext = new PowRuleContext(new ValidationContext(), this.dateTimeProvider.Object.GetTimeOffset());
-                this.consensusRules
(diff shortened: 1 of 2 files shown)


noescape00:
@Aprogiena

src/Stratis.Bitcoin.Features.Miner.Tests/PowBlockAssemblerTest.cs
```diff
@@ -249,33 +249,6 @@ public void UpdateHeaders_UsingChainAndNetwork_PreparesBlockHeaders()
...
-                this.consensusManager.Setup(c => c.Tip).Returns(chain.GetBlock(5));
-
-                ValidationContext validationContext = null;
-                var powRuleContext = new PowRuleContext(new ValidationContext(), this.dateTimeProvider.Object.GetTimeOffset());
-                this.consensusRules
-                    .Setup(s => s.CreateRuleContext(It.IsAny<ValidationContext>())).Callback<ValidationContext>((r) => validationContext = r)
-                    .Returns(powRuleContext);
-
-                var powBlockAssembler = new PowTestBlockDefinition(this.consensusManager.Object, this.dateTimeProvider.Object, this.LoggerFactory.Object, this.txMempool.Object, new MempoolSchedulerLock(), this.minerSettings.Object, this.testNet, this.consensusRules.Object);
-                Block block = powBlockAssembler.TestBlockValidity();
```
bokobza: do we need to remove the whole test here or can we test the block validity using another way?

fassadlr: @bokobza I think because `TestBlockValidity` is not apart of block construction any longer these tests can be removed. There are various tests that call the `BlockMined` moethod on CM, that tests various scenarios where the blocks are invalid. @Aprogiena @noescape00 back me up here ? ;)

bokobza: ok cool, thanks @fassadlr.

Aprogiena:
Okay I think

src/Stratis.Bitcoin.Features.Miner.Tests/PowBlockAssemblerTest.cs
Aprogiena: Yes I think that is also true

PR #1995, opened 2018-08-22
Added timeoffset option to adjust node time

maciejzaleski:
The timeoffset option will allow to run a node with a different time to the current system time. This is required in order to set up end to end testing where nodes are required to run few seconds in the future (using positive value for the setting) or in the past (by using negative value for the setting).

As we are planning to setup test network on Docker we need to have ability to adjust time on each of the nodes. Running each node in a different Docker container is great in terms of isolating the FullNode process, assigning unique IP address "to the node", controlling resource utilisation, etc. Because all Docker containers are using the same linux kernel it is impossible to have two containers with a different system time.
I have experimented with faking Linux calls for the date time with https://github.com/wolfcw/libfaketime (doesn't work probably due to how dotNet is loading libraries) 
And replacing method implementation by changing method's pointers at the IL level https://www.codeproject.com/Articles/37549/CLR-Injection-Runtime-Method-Replacer (gave up after about half a day; I suspect that for this to work you need to clear the compiled versions of the CLR libraries)
There is also an option to run each node on a different machine but this is too resource consuming and would limit how many nodes could run per machine.

The alternative to the above approach is to have the ability to adjust time within the node. This is one of the 2 PR to achieve this.

**PR 1992: Replaced all references to DateTime.Now and DateTime.UtcNow with DateTimeProvider**
I have replaced all references to the DateTime properties Now, UtcNow and Today (last one not used anywhere) with calls to DateTimeProvider.GetUtcNow(). As a result the current time could be changed. The limitation is that any external classes (e.g. NLog) which still use DateTime.Now property will return system time. This however should not be a problem as we are only changing the time for all activities related to the blockchain.

**PR 1995: Added timeoffset option to adjust node time**
This is to add ability to configure time offset from the command line and config file. With the above PR 1992 we have ensured that all calls to get current time are made through DateTimeProvider class so that the time offset have to be set in only one place.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/DateTimeProviderTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/DateTimeProviderTest.cs
@@ -10,9 +10,11 @@ namespace Stratis.Bitcoin.Tests.Utilities
         [Fact]
         public void GetUtcNowReturnsCurrentUtcDateTime()
         {
+            int timeOffset = -96;
+            DateTimeProvider.Default.SetSystemTimeOffset(timeOffset);
             DateTime result = DateTimeProvider.Default.GetUtcNow();
 
-            Assert.Equal(DateTime.UtcNow.ToString("yyyyMMddhhmmss"), result.ToString("yyyyMMddhhmmss"));
+            Assert.Equal(DateTime.UtcNow.AddSeconds(timeOffset).ToString("yyyyMMddhhmmss"), result.ToString("yyyyMMddhhmmss"));
         }
 
         [Fact]
--- a/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
+++ b/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
@@ -76,6 +76,9 @@ namespace Stratis.Bitcoin.Configuration
         /// <summary>Minimum relay transaction fee for network.</summary>
         public FeeRate MinRelayTxFeeRate { get; private set; }
 
+        /// <summary>Node's time offset in relation to system time expresed as either positive or negative number of seconds.</summary>
+        public int TimeOffset { get; set; }
+
         /// <summary>
         /// Initializes a new instance of the object.
         /// </summary>
@@ -266,7 +269,12 @@ namespace Stratis.Bitcoin.Configuration
             this.MinTxFeeRate = new FeeRate(config.GetOrDefault("mintxfee", this.Network.MinTxFee, this.Logger));
             this.FallbackTxFeeRate = new FeeRate(config.GetOrDefault("fallbackfee", this.Network.FallbackFee, this.Logger));
             this.MinRelayTxFeeRate = new FeeRate(config.GetOrDefault("minrelaytxfee", this.Network.MinRelayTxFee, this.Logger));
+
+            // Node time adjustment
+            this.TimeOffset = this.ConfigReader.GetOrDefault<int>("timeoffset", 0, this.Logger);
+            DateTimeProvider.Default.SetSystemTimeOffset(TimeOffset);
         }
+
 
         /// <summary>
         /// Creates default data directories respecting different operating system specifics.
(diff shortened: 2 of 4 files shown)


Aprogiena:
can't you change the system time of that system where this runs (+ possibly disable timesync feature)?

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -266,8 +269,13 @@ private void LoadConfiguration()
             this.MinTxFeeRate = new FeeRate(config.GetOrDefault("mintxfee", this.Network.MinTxFee, this.Logger));
             this.FallbackTxFeeRate = new FeeRate(config.GetOrDefault("fallbackfee", this.Network.FallbackFee, this.Logger));
             this.MinRelayTxFeeRate = new FeeRate(config.GetOrDefault("minrelaytxfee", this.Network.MinRelayTxFee, this.Logger));
+
+            // Node time adjustment
+            this.TimeOffset = this.ConfigReader.GetOrDefault<int>("timeoffset", 0, this.Logger);
```
bokobza: Can you also please adapt the `PrintHelp` and `BuildDefaultConfigurationFile`?
Pleas specify that it is a testing-only launch argument.

src/Stratis.Bitcoin/Utilities/DateTimeProvider.cs
```diff
@@ -76,7 +84,7 @@ public virtual long GetTime()
         /// <inheritdoc />
         public virtual DateTime GetUtcNow()
         {
-            return DateTime.UtcNow;
+            return DateTime.UtcNow.AddSeconds(this.systemTimeOffset);
```
bokobza: You should probably add this offset to the other methods (at least until your other PR is merged).

Aprogiena: how about GetTimeOffset?

Aprogiena:
Another question – why do we want to modify the system time and not just the adjusted time?

maciejzaleski:
The call to GetAdjustedTime has only 9 references and will probably change time only for some aspects of the blockchain interaction. I want to be able to control the time for as much as I possibly can.
I wouldn't think that every call to GetUtcNow could be substituted with GetAdjustedTime.

src/Stratis.Bitcoin/Utilities/DateTimeProvider.cs
maciejzaleski: Do you mean that the systemTimeOffset should be returned by GetTimeOffset method?

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
bokobza: I think it's better to have that with some explicit warning on what it's for.

src/Stratis.Bitcoin.Tests/Utilities/DateTimeProviderTest.cs
```diff
@@ -10,9 +10,11 @@ public class DateTimeProviderTest
         [Fact]
         public void GetUtcNowReturnsCurrentUtcDateTime()
         {
+            int timeOffset = -96;
```
bokobza: Can you please create a new test in addition to this one and leave this one as is?

maciejzaleski:
**Example**
https://github.com/stratisproject/StratisBitcoinFullNode/blob/a73249d8a285f6957257c794b4c5a29573c7f86a/src/Stratis.Bitcoin/Base/InvalidBlockHashStore.cs#L93

My analysis is based on the calls to the DateTime.UtcNow call but this is probably not sufficient. We could be assigning current time to a class Field/Property and using it later in the invocation call. 

Thinking about now it is a different functionality.
**AdjustedTimeOffset** - represents node time which has been adjusted with cooperation with network peers using `TimeSyncBehaviorState`
**SystemTimeOffset** - it is a current system time with an offset supplied as a node configuration option

If we want to be able to test node time synchronisation we need to start two nodes each with a different SystemTimeOffset and prove that the AdjustedTimeOffset has been updated to get both nodes to agree on current network time.

Aprogiena:
I do understand there is code that is not using adjusted time. However, I'm still unconvinced that we have a use case which would rely on that. Maybe I just cannot remember such a use case and that obviously is one, but I just cannot see it now.

In any case, if we want this, both methods should be adjusted, not just the one

@noescape00 do you think we need this or not?

Aprogiena:
Talk to Maciej and we agreed to close this one and shrink 1992

bokobza:
Closing.

PR #1996, opened 2018-08-22
vs xunit runner needs explicit ref for jetbrains rider

majikandy:
Added explicit package ref to xunit.runner.visualstudio because jetbrains rider ide doesn't detect tests even though this package is inherited from the test.common project.

This solves the issue and allows all tests to be detected by the IDE now.


bokobza:
I'll check it.

bokobza:
Seems to be ok
image

bokobza:
Waiting for the build to merge.

PR #1998, opened 2018-08-22
[Activation] Fix a few API integration tests

fassadlr:
Consensus Manager wasn't set in the ConsensusController.


src/Stratis.Bitcoin/Controllers/FeatureController.cs
```diff
@@ -32,15 +33,15 @@ public abstract class FeatureController : Controller
 
         protected Connection.IConnectionManager ConnectionManager { get; set; }
```
majikandy: seems the Connection prefix could be removed here too ?

Issue #1938, opened 2018-08-16
[Activation] Do BlockIntegrity check outside of lock


noescape00:
integrity check is performed in CHT. Do it in CM outside of the cht's lock

bokobza:
Closed with https://github.com/stratisproject/StratisBitcoinFullNode/pull/1999

PR #1999, opened 2018-08-22
[Activation] Do BlockIntegrity check outside of lock

zeptin:
For #1938

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -95,7 +95,7 @@ namespace Stratis.Bitcoin.Consensus
         List<int> ConsensusTipChanged(ChainedHeader newConsensusTip);
 
         /// <summary>
-        /// Finds the header and verifies block integrity.
+        /// Finds the header for a given block.
         /// </summary>
         /// <param name="block">The block.</param>
         /// <returns>Chained header for a given block.</returns>
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -95,13 +95,6 @@ namespace Stratis.Bitcoin.Consensus
         List<int> ConsensusTipChanged(ChainedHeader newConsensusTip);
 
         /// <summary>
-        /// Finds the header for a given block.
-        /// </summary>
-        /// <param name="blockHash">The hash of the block.</param>
-        /// <returns>Chained header for a given block.</returns>
-        ChainedHeader FindHeader(uint256 blockHash);
-
-        /// <summary>
         /// Handles situation when the blocks data is downloaded for a given chained header.
         /// </summary>
         /// <param name="chainedHeader">Chained header that represents <paramref name="block"/>.</param>
@@ -638,21 +631,6 @@ namespace Stratis.Bitcoin.Consensus
         }
 
         /// <inheritdoc />
-        public ChainedHeader FindHeader(uint256 blockHash)
-        {
-            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);
-
-            if (!this.chainedHeadersByHash.TryGetValue(blockHash, out ChainedHeader chainedHeader))
-            {
-                this.logger.LogTrace("(-)[HEADER_NOT_FOUND]");
-                return null;
-            }
-
-            this.logger.LogTrace("(-):'{0}'", chainedHeader);
-            return chainedHeader;
-        }
-
-        /// <inheritdoc />
         public bool BlockDataDownloaded(ChainedHeader chainedHeader, Block block)
         {
             this.logger.LogTrace("({0}:'{1}')", nameof(chainedHeader), chainedHeader);
@@ -1019,7 +997,7 @@ namespace Stratis.Bitcoin.Consensus
 
             this.CreateNewHeaders(new List<BlockHeader>() { block.Header });
 
-            ChainedHeader chainedHeader = this.FindHeader(block.GetHash());
+            ChainedHeader chainedHeader = this.GetChainedHeader(block.GetHash());
             this.BlockDataDownloaded(chainedHeader, block);
 
             this.logger.LogTrace("(-):'{0}'", chainedHeader);
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -100,8 +100,7 @@ namespace Stratis.Bitcoin.Consensus
         /// <param name="block">The block.</param>
         /// <returns>Chained header for a given block.</returns>
         /// <exception cref="BlockDownloadedForMissingChainedHeaderException">Thrown when block data is presented for a chained block that doesn't exist.</exception>
-        /// <exception cref="IntegrityValidationFailedException">Thrown in case integrity validation failed.</exception>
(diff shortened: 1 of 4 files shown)


zeptin:
I am getting `ProcessHeadersAsync_PeerThatSentInvalidHeaderIsBannedAsync` failing, but that seems to be failing even with the latest activation commits, so I don't think it is related to the changes in this PR.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -956,27 +958,33 @@ private void BlockDownloaded(uint256 blockHash, Block block)
...
                 }
             }
 
+            if (block != null)
+            {
+                ValidationContext result = this.integrityValidator.VerifyBlockIntegrity(chainedHeader, block);
+
+                if (result.Error != null)
+                {
+                    this.peerBanning.BanAndDisconnectPeer(result.Peer, result.BanDurationSeconds,
```
bokobza: I don't understand where the peer is set in the ValidationContext object.
Not in the tests, but in the normal code.

noescape00: Good find.
The peer is not set anywhere. 

@dangershony

noescape00: However it's not an issue of this PR, it's more about consensus rules engine so let's review this PR assuming it's there.

noescape00: Created an issue for that https://github.com/stratisproject/StratisBitcoinFullNode/issues/2008

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -95,13 +95,12 @@ public interface IChainedHeaderTree
...
         /// <summary>
-        /// Finds the header and verifies block integrity.
+        /// Finds the header for a given block.
         /// </summary>
         /// <param name="block">The block.</param>
         /// <returns>Chained header for a given block.</returns>
         /// <exception cref="BlockDownloadedForMissingChainedHeaderException">Thrown when block data is presented for a chained block that doesn't exist.</exception>
-        /// <exception cref="IntegrityValidationFailedException">Thrown in case integrity validation failed.</exception>
-        ChainedHeader FindHeaderAndVerifyBlockIntegrity(Block block);
+        ChainedHeader FindHeader(Block block);
```
noescape00: Maybe change from block to hash? 
CHT needs only hash and it looks like all the callers already have hash in the variable somewhere

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -656,14 +652,6 @@ public ChainedHeader FindHeaderAndVerifyBlockIntegrity(Block block)
                 throw new BlockDownloadedForMissingChainedHeaderException();
```
noescape00: Probably should change it to return null instead of throwing exception and allow caller handle it

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -956,27 +958,33 @@ private void BlockDownloaded(uint256 blockHash, Block block)
...
-                    }
                 }
                 else
                 {
                     this.logger.LogDebug("Block '{0}' failed to be delivered.", blockHash);
                 }
             }
 
+            if (block != null)
+            {
```
noescape00: this block of code should be right after finding the header

Aprogiena:
Want to see this one as well.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
zeptin: That wasn't really in the original scope, but I will do it if you want me to.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
zeptin: You said outside the lock? Or were you referring to a different lock?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: Yes, please. 

I know that's extra but that would make perfect sense.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Correct, outside of the peer lock. 

I mean when we can have header null there (return null instead of throwing) and return in that case like we do now we won't need to check if the block is null again. By that time we know that block and header are not null . 

So just remove that if basically so we have integrity outside of the lock and right after we find header and handle if it's null

zeptin: I'm not entirely sure I follow - if `block` is null then the `else` branch above gets executed, but the method does not return. Therefore coming out of the lock `block` could either be `null` or not `null`, and we need to check it again.

noescape00: Ah, yes you are right, we do need to check, sorry. 

Forgot about cases where we want to process block being null (when puller failed to deliver)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -643,25 +638,14 @@ private void DisconnectChainHeader(ChainedHeader header)
...
+        public ChainedHeader FindHeader(uint256 blockHash)
         {
-            uint256 blockHash = block.GetHash();
-            this.logger.LogTrace("({0}:'{1}')", nameof(block), blockHash);
-
-            ChainedHeader chainedHeader;
+            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);
 
-            if (!this.chainedHeadersByHash.TryGetValue(blockHash, out chainedHeader))
+            if (!this.chainedHeadersByHash.TryGetValue(blockHash, out ChainedHeader chainedHeader))
```
bokobza: why is this not throwing an exception anymore?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -643,25 +638,14 @@ private void DisconnectChainHeader(ChainedHeader header)
         }
 
         /// <inheritdoc />
-        public ChainedHeader FindHeaderAndVerifyBlockIntegrity(Block block)
+        public ChainedHeader FindHeader(uint256 blockHash)
```
Aprogiena: How is this different from GetChainedHeader?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
noescape00: >why is this not throwing an exception anymore?

Before it made sense to throw because we would also validate the header. 

Now the method just looks up the header so I thought it's a good idea to return null because we have similar behavior in other methods in the codebase that do the lookups and have same naming pattern

bokobza:
@zeptin please the conflict and maybe we can merge it.
@Aprogiena @noescape00 are you ok with it.

noescape00:
conflicts

Aprogiena:
Still conflict

Aprogiena:
Looks good to me, it seems like there is a bug in the logic of method block downloaded, but that is out of scope of this PR, so I think this PR can go in

PR #2001, opened 2018-08-23
[SC] Unit test for contract upgradeability

### LINKED ISSUE #1982: [SC] Contract upgrading unit test
We need to implement a test for what happens when `Stratis.SmartContracts` NuGet/API is upgraded.

Criteria:
* Compile a contract against an earlier version of `Stratis.SmartContracts`
* Attempt to load the contract into an isolated ALC with a newer version of `Stratis.SmartContracts`

rowandh:
Closes #1982.

Tests the scenario where a contract compiled against `Stratis.SmartContracts.SmartContract` v1.0.0-TEST is loaded into an ALC with `Stratis.SmartContracts.SmartContract` v2.0.0-TEST. v2.0.0-TEST contains an additional method, `TestMethod`.


src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            // Load the contract compiled against v1.0.0-TEST Stratis.SmartContracts
+            var ms = new MemoryStream(version1CompiledContract);          
+            var assembly = alc.LoadFromStream(ms);
+            ms.Dispose();
+
+            var type = assembly.ExportedTypes.First(t => t.Name == "TestContract");
+
+            var method = type.BaseType.GetMethod("TestMethod");
+            
+            // If this condition is true, we have a V1 contract referencing a V2 assembly
```
majikandy: if null you mean ?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Emit;
+using Xunit;
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Tests
+{
+    public class ContractUpgradeabilityTests
+    {
+        public class TestAlc : AssemblyLoadContext
```
majikandy: what is ALC ?  should this be long form ?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            protected override Assembly Load(AssemblyName assemblyName)
+            {
+                return null;
+            }
+        }
+
+        [Fact]
+        public void Load_Contract_Compiled_Against_V1_With_V2_SmartContracts()
```
majikandy: extend test name here to explain what happens in this scenario,  eg  it fails, it works fine ?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+                EmitResult emitResult = compilation.Emit(dllStream);
+                
+                Assert.True(emitResult.Success);
+
+                version1CompiledContract = dllStream.ToArray();
+            }
+
+            var alc = new TestAlc();
+
+            // Load the v2.0.0-TEST Stratis.SmartContracts assembly into the ALC
```
majikandy: the line of code expresses the same as the comment?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            var ms = new MemoryStream(version1CompiledContract);          
+            var assembly = alc.LoadFromStream(ms);
+            ms.Dispose();
+
+            var type = assembly.ExportedTypes.First(t => t.Name == "TestContract");
+
+            var method = type.BaseType.GetMethod("TestMethod");
+            
+            // If this condition is true, we have a V1 contract referencing a V2 assembly
+            Assert.NotNull(method);
```
majikandy: does this method allow a custom string for the failure?   if so  you don't need the comment above...  if it doesn't you could use  Fluent Assertions   .Should().NotBeNull("TestMethod is only available in version 2")  or whatever message is appropriate

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            var alc = new TestAlc();
+
+            // Load the v2.0.0-TEST Stratis.SmartContracts assembly into the ALC
+            var v2Assembly = alc.LoadFromAssemblyPath(version2DllPath);
+
+            Assert.Equal(Version.Parse("2.0.0.0"), v2Assembly.GetName().Version);
+
+            // Load the contract compiled against v1.0.0-TEST Stratis.SmartContracts
+            var ms = new MemoryStream(version1CompiledContract);
```
majikandy: ms -> memoryStream or  version1compliledContractMemoryStream

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+                
+                Assert.True(emitResult.Success);
+
+                version1CompiledContract = dllStream.ToArray();
+            }
+
+            var alc = new TestAlc();
+
+            // Load the v2.0.0-TEST Stratis.SmartContracts assembly into the ALC
+            var v2Assembly = alc.LoadFromAssemblyPath(version2DllPath);
```
majikandy: version1   used elsewhere so either this to version2 or version1 -> v1

src/Stratis.Bitcoin.Features.SmartContracts.Tests/ContractUpgradeabilityTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            var type = assembly.ExportedTypes.First(t => t.Name == "TestContract");
+
+            var method = type.BaseType.GetMethod("TestMethod");
+            
+            // If this condition is true, we have a V1 contract referencing a V2 assembly
+            Assert.NotNull(method);
+        }
+    }
+}
```
majikandy: do you need the other scenarios when the versions match or are they elsewhere? also what about a version2 with version1 or is that not a valid scenario?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/Stratis.Bitcoin.Features.SmartContracts.Tests.csproj
```diff
@@ -118,9 +118,19 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Update="Packages\1.0.0-TEST\Stratis.SmartContracts.dll">
```
majikandy: So these dlls will be checked in, is this the only way this can be done ?

rowandh: They are part of the test and contain different versions of Stratis.SmartContracts. I guess we could somehow dynamically compile both of them.

rowandh: There is also a dependency on a specific version of System.Runtime, which is why that's included. It's hard/impossible to know if the local machine has that specific version.

majikandy: ok

codingupastorm:
@rowandh did you try invoking the method?

PR #2005, opened 2018-08-23
A couple of typo/sentencing corrections

majikandy:
As reading the documentation to try to understand it, noticed a couple of sentence improvements

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -20,11 +20,11 @@ namespace Stratis.Bitcoin.Consensus
     /// This component is an extension of <see cref="ConsensusManager"/> and is strongly linked to its functionality, it should never be called outside of CM.
     /// <para>
     /// View of the chains that are presented by connected peers might be incomplete because we always
-    /// receive only chunk of headers claimed by the peer in one message.
+    /// receive only a chunk of headers claimed by the peer in each message.
     /// </para>
     /// <para>
-    /// It is a role of the consensus manager to decide which of the presented chains is going to be treated as our best chain.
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -20,11 +20,11 @@ namespace Stratis.Bitcoin.Consensus
     /// This component is an extension of <see cref="ConsensusManager"/> and is strongly linked to its functionality, it should never be called outside of CM.
     /// <para>
     /// View of the chains that are presented by connected peers might be incomplete because we always
-    /// receive only chunk of headers claimed by the peer in one message.
+    /// receive only a chunk of headers claimed by the peer in each message.
     /// </para>
     /// <para>
-    /// It is a role of the consensus manager to decide which of the presented chains is going to be treated as our best chain.
-    /// <see cref="ChainedHeaderTree"/> only advices which chains it might be interesting to download.
+    /// It is a role of the <see cref="ConsensusManager"/> to decide which of the presented chains is going to be treated as our best chain.
```
bokobza: It is the* role

bokobza: of maybe not actually

bokobza: If it's one of its roles, then "One of the roles of the <see cref="ConsensusManager"/> is to decide" would be clearer.

majikandy: Haha, I only changed the words consensus manager to the cref but I think that it has the same meaning without saying “one of...” just because it says it is the role of, doesn’t mean that is its only role.


A wider point on this documentation is that the interface is documented as if it were the concrete class and the implementation of the concrete class very explicitly mentioned on the interface.

Issue #1820, opened 2018-08-02
local js attack on API


noescape00:

monsieurleberre:
this should be resolved when https://github.com/stratisproject/StratisBitcoinFullNode/issues/1263 is resolved

monsieurleberre:
more details from a discussion with @noescape00 
```
Basically the idea is that you go to a website and there is js code that calls API methods. 
It doesn't get the response because of cross domain origin calls policy, but the method is called anyway.

Fix would be doing smth that's prevents such calls to go through. It can be done using some token for example

ferdeen:
[Enabling CORS](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.1)

![image](https://user-images.githubusercontent.com/16000683/44086762-7d8141d0-9fb5-11e8-9e6f-856f0c20148c.png)

monsieurleberre:
Thanks for the input guys! I will have a look now, this is not my forte but I am very interested to learn :)

Aprogiena:
the scenario is that you visit a malicious website which contains a JavaScript, which will connect to the API of your wallet and execute the command

Aprogiena:
If I remember correctly, the browser will allow the JavaScript to invoke a call to arbitrary URL, but it will prevent the JavaScript to read the response from the target server. Therefore, if we make sure that all GET methods are strictly used for querying only (which means that you are unable to change any state by calling them), which I believe we already do have it like that, it should just probably be reviewed once again, then the only methods that can be problematic should be POST methods. However, if I remember correctly again, the browser will not send a POST request unless it is allowed by the policy or it has empty body. Therefore, if we change all the POST requests that can be currently called without any body, to have a body with some value that we check in the controller, for example `a=1`, then it should be impossible for the JavaScript to invoke such a request and therefore would fix the issue.

Aprogiena:
Alternative solution would be to do CSRF tokens as mentioned above.

Aprogiena:
But please verify what I wrote, I could have been mistaken

monsieurleberre:
That seems to indicate that if we are not creating sessions or other mechanisms by which we keep track of the client's state, then CSRF is not needed :)
https://security.stackexchange.com/questions/166724/should-i-use-csrf-protection-on-rest-api-endpoints

monsieurleberre:
I have had a look through our controllers and have put some questions I will come back to.
I leave them here for the moment.

[...]

PR #2013, opened 2018-08-23
Api js attack

monsieurleberre:
There are still some "Get" requests that seem to be doing a lot of work and for which we might want to do some changes (maybe GetHistory in the wallet for instance)

Closes https://github.com/stratisproject/StratisBitcoinFullNode/issues/1820


src/Stratis.Bitcoin/Controllers/NodeController.cs
```diff
@@ -358,8 +358,10 @@ public async Task<IActionResult> GetTxOutAsync([FromQuery] string trxid, uint vo
         [HttpPost]
         [Route("shutdown")]
         [Route("stop")]
-        public IActionResult Shutdown()
+        public async Task<IActionResult> ShutdownAsync([FromBody] int delayInSeconds = 0)
         {
+            await Task.Delay(TimeSpan.FromSeconds(delayInSeconds)).ConfigureAwait(false);
```
Aprogiena: Why would you change the method to be async just to insert a delay?

Aprogiena: Or is it somehow required by `[FromBody]`?

monsieurleberre: I did that because I was looking for a parameter that would not be completely useless.
I could have added a boolean "areYouSure" but it is stupid. @bokobza  and I agreed this was adding a little value, as well as making the method safe

Aprogiena: you could always go with Thread.Sleep though Without changing the method to async. That is what I'm talking about here. I will need to check what you are talking about hopefully later today

src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
```diff
@@ -91,7 +91,7 @@ private RPCResponse SendRPCRequest(RPCRequest request)
         /// </summary>
         /// <returns>A JSON result that varies depending on the RPC method.</returns>
         [Route("callbyname")]
-        [HttpGet]
+        [HttpPost]
```
bokobza: in this POST, the body is still empty.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/Controllers/WatchOnlyWalletController.cs
```diff
@@ -34,7 +34,7 @@ public WatchOnlyWalletController(IWatchOnlyWalletManager watchOnlyWalletManager)
         /// <param name="address">The base58 address to add to the watch list.</param>
         [Route("watch")]
         [HttpPost]
-        public IActionResult Watch([FromQuery]string address)
+        public IActionResult Watch([FromBody]string address)
```
bokobza: This needs to be checked with the wallet (Tumblebit in particular) to make sure it doesn't break anything. If it does, then please create an issue reminding them to update the code at the same time they get the latest FN version.

src/Stratis.Bitcoin/Controllers/NodeController.cs
bokobza: If the user wants to delay the shutdown for 30 min, Thread.Sleep might be a problem.

src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
monsieurleberre: oh putain, merci!

src/Stratis.Bitcoin.Features.WatchOnlyWallet/Controllers/WatchOnlyWalletController.cs
monsieurleberre: Cool I'll do that

Aprogiena:
So we have checked whether this fixes the problem and it looks like in a way it does, but it is relying on some kind of specific behaviour of the kestrel or the behaviour of frombody attribute or maybe both. So the JavaScript is still able to send the empty post request, but because of the from body attribute, the HTTP engine requires content type to be specified and this is not possible to do with empty post request. So we got back error 415 which means wrong media type. So we are currently unaware of how to go past this limitation, but it feels quite hacky.

The introduction of delay in seconds into those methods is also quite ugly, I was initially thinking about something like `corsProtection=1` and the code on the method but actually check that it is set to 1 with no default value.

In any case, given that we don't know how to bypass it currently, and willing to accept the solution, but each using of the extra delay in seconds must be properly commented that its main purpose is to prevent this attack.

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
```diff
@@ -126,8 +127,9 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
         /// <returns>An <see cref="OkResult"/> object that produces a status code 200 HTTP response.</returns>
         [Route("stopstaking")]
         [HttpPost]
-        public IActionResult StopStaking()
+        public async Task<IActionResult> StopStakingAsync([FromBody] int delayInSeconds = 0)
         {
+            await Task.Delay(TimeSpan.FromSeconds(delayInSeconds)).ConfigureAwait(false);
```
Aprogiena: So please insert comment here or on the parameter

src/Stratis.Bitcoin/Controllers/NodeController.cs
Aprogiena: So please insert comment here or on the parameter

Aprogiena: You don't have cancellation token so you will have problem even with task delay, if you want to remove that problem, you need to have cancellation token node lifetime

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
```diff
@@ -123,25 +123,31 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
         /// <summary>
         /// Stop staking.
         /// </summary>
+        /// <param name="delayInSeconds">Delay, in seconds, after which the staking will stop.</param>
         /// <returns>An <see cref="OkResult"/> object that produces a status code 200 HTTP response.</returns>
         [Route("stopstaking")]
         [HttpPost]
-        public IActionResult StopStaking()
+        public IActionResult StopStakingAsync([FromBody] int delayInSeconds = 0)
```
ferdeen: Hey, how come this method has an `async` suffix ? I can't see any 'ayncy' stuff going on in the method.

src/Stratis.Bitcoin.Tests/Controllers/NodeControllerTest.cs
```diff
@@ -61,9 +61,9 @@ public NodeControllerTest()
         }
 
         [Fact]
-        public void Stop_WithFullNode_DisposesFullNodeAsync()
+        public async Task Stop_WithFullNode_DisposesFullNodeAsync()
         {
-            IActionResult result = this.controller.Shutdown();
+            IActionResult result = await this.controller.ShutdownAsync(0);
```
ferdeen: In `NodeController` the method is still called `Shutdown()`.  

When I loaded this PR locally I get "NodeController does not contain a definition for nodeShutdownAsync...".

src/Stratis.Bitcoin.Features.WatchOnlyWallet/Controllers/WatchOnlyWalletController.cs
ferdeen: Jeremy if you didn't remember that then this could potentially have caused a problem in Tumblebit.

Thinking out load - is there any mileage is adding a remark at controller class level if the controller is used in other Stratis related Projects ? This would help if devs make a change that has a dependency.

For Breeze-Tumblebit in particular this problem would go away if and when it starts referencing the FullNode nuget packages.

src/Stratis.Bitcoin.Tests/Controllers/NodeControllerTest.cs
ferdeen: Also does `ShutdownAsync()` need `.ConfigureAwait(false)` in tests ?

src/Stratis.Bitcoin/Controllers/NodeController.cs
```diff
@@ -354,15 +354,17 @@ public async Task<IActionResult> GetTxOutAsync([FromQuery] string trxid, uint vo
...
-        public IActionResult Shutdown()
+        public IActionResult Shutdown([FromBody] int delayInSeconds = 0)
         {
-            // Start the node shutdown process.
-            this.fullNode?.Dispose();
-
+            var timer = new System.Timers.Timer(delayInSeconds * 1000);
+            // Start the node shutdown process on timer elapsed.
+            timer.Elapsed += (a, s) => this.fullNode?.Dispose();
+            timer.Start();
```
ferdeen: Nice.  Not now, but maybe useful to implement the delay in the `Dispose()` method directly and pass in the `delayInSeconds` (default 0 seconds).

src/Stratis.Bitcoin.IntegrationTests/API/AllControllersTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        private static Func<MethodInfo, bool> IsHttpPostMethod()
+        {
+            return m => m.CustomAttributes.Any(a => a.AttributeType == typeof(HttpPostAttribute));
+        }
+
+        private static Func<ParameterInfo, bool> HasAnyFromBodyAttribute()
+        {
+            return p => p.CustomAttributes.Any(a => a.AttributeType == typeof(FromBodyAttribute));
+        }
+    }
```
ferdeen: Nice!

bokobza:
I'll take over this one, @monsieurleberre being on hols.

src/Stratis.Bitcoin.Features.WatchOnlyWallet/Controllers/WatchOnlyWalletController.cs
monsieurleberre: If anything, I'd add a quick test with a name like "TestControllerIsDoingWhatTumblebitExpects" and make sure it breaks when a breaking change is made. Then the person changing it will ask why this test is there and hopefully get some help.
But if tumblebit is using a nuget package, then before upgrading to a new package it should be tested and that issue should come out.
I think creating dependency to downstream projects through comments is not the way forward. It can get messy if you start using the same project in many others (which is what we should hope for), and then have to update your comments in top level project everytime a downstream projects evolve...

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
monsieurleberre: Correct thanks, I changed it but not the name.

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
```diff
@@ -124,25 +124,31 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
         /// <summary>
         /// Stop staking.
         /// </summary>
+        /// <param name="delayInSeconds">Delay, in seconds, after which the staking will stop.</param>
```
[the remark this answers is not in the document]
monsieurleberre: OK I can add a word about that

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
```diff
@@ -124,25 +124,31 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
...
-            {
-                if (!this.fullNode.Network.Consensus.IsProofOfStake)
+            if (!this.fullNode.Network.Consensus.IsProofOfStake)
                     return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", "Method not available for Proof of Stake");
 
-                this.fullNode.NodeFeature<MiningFeature>(true).StopStaking();
-                return this.Ok();
-            }
-            catch (Exception e)
+            var timer = new System.Timers.Timer(delayInSeconds * 1000);
```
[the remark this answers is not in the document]
monsieurleberre: I'll look into it, thanks

src/Stratis.Bitcoin.Tests/Controllers/NodeControllerTest.cs
monsieurleberre: Apologies, I might have screwed it up. Let me check.

src/Stratis.Bitcoin/Controllers/NodeController.cs
monsieurleberre: I am not sure that I like the idea of a "complex" Dispose() method like that. But we can put some thoughts into it.

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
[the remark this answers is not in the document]
Aprogiena: I actually think this idea with timer is not a good one, introducing the cancellation token from the node lifetime be much easier solution

src/Stratis.Bitcoin.Tests/Controllers/NodeControllerTest.cs
```diff
@@ -61,11 +63,16 @@ public NodeControllerTest()
...
+        public void Stop_WithFullNode_DisposesFullNode()
         {
+            var isDisposed = false;
+            this.fullNode.Setup(f => f.Dispose()).Callback(() => isDisposed = true);
+
             IActionResult result = this.controller.Shutdown(0);
 
-            this.fullNode.Verify(f => f.Dispose());
+            result.Should().BeOfType<OkResult>();
+            Thread.Sleep(100);
```
monsieurleberre: Ideally I would have brought the waitloop from integration tests or raised a public event to test, but I thought this would be deemed out of scope. Hopefully this is not too bad.

monsieurleberre:
I fixed a few mistakes now it looks better
image

monsieurleberre:
Yes agreed it is a bit more complicated than expected, not sure this is worth the headache. What race condition are you talking about though ? Can you detail ? @bokobza what is your opinion on this ?

Aprogiena:
Just imagine what happens if you receive two API calls to stop staking for example at the same time. You will possibly dispose the timer twice and then you will create two timers and one of them will never be disposed and the other could have two events on it.

Aprogiena:
Also from the design point of view, this is just a controller. It should not implement any logic. It should just identify the underlying component and call it. If you look into the staking component you can find the actual logic to prevent such a race condition there. We don't want to duplicate such a mechanism here.

bokobza:
The last thing I want to do is to introduce extra avenues for bugs and extra testing.
The idea of having timers has legs but I'm afraid this "little extra feature" will come back at us.
@monsieurleberre what do you suggest to make it as simple as possible?

monsieurleberre:
well I think my first ugly idea was good enough, just replace that delay with a bool like "areYouSure=true" or equivalent, and we are safe... Same as @Aprogiena 's "corsProtection=1" really...

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
```diff
@@ -123,10 +123,14 @@ public IActionResult StartStaking([FromBody]StartStakingRequest request)
...
         /// </summary>
+        /// <param name="corsProtection">This body parameter is here to prevent a CORS call from triggering method execution.</param>
+        /// <remarks>
+        /// <seealso cref="https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests"/>
+        /// </remarks>
         /// <returns>An <see cref="OkResult"/> object that produces a status code 200 HTTP response.</returns>
         [Route("stopstaking")]
         [HttpPost]
-        public IActionResult StopStaking()
+        public IActionResult StopStaking([FromBody] bool corsProtection = true)
```
bokobza: What will happen if we don't pass a body here?

src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
```diff
@@ -91,27 +92,28 @@ private RPCResponse SendRPCRequest(RPCRequest request)
...
         /// <returns>A JSON result that varies depending on the RPC method.</returns>
         [Route("callbyname")]
-        [HttpGet]
-        public IActionResult CallByName([FromQuery]string methodName)
+        [HttpPost]
+        public IActionResult CallByName([FromBody]JObject body)
         {
             try
             {
+                var ignoreCase = StringComparison.InvariantCultureIgnoreCase;
```
bokobza: Do we need to have this as a variable?

monsieurleberre: just found it more readable, but I am not too fussed about it honestly.

monsieurleberre:
image

src/Stratis.Bitcoin.Features.Miner/Controllers/StakingController.cs
monsieurleberre: it's OK, I've tested with this ajax (both with and without content) and the code is not called
```
        $('#ajax_post').click(function () {
            $.ajax({
                type: "POST",
                crossDomain: true,
                headers: {},
                content: {},
                //content: { "corsProtection" : "true" },
                dataType: "json",
                url: "http://localhost:38221/api/Staking/stopstaking",
                success: function (data) {
                    alert(data);
                },
                error: function (error) {
                    alert(error);
                }
            });
        });
```

src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
```diff
@@ -91,27 +92,28 @@ private RPCResponse SendRPCRequest(RPCRequest request)
...
-                var param = new object[paramInfo.Count];
-                for (int i = 0; i < paramInfo.Count; i++)
-                {
-                    ControllerParameterDescriptor pInfo = paramInfo[i];
-                    KeyValuePair<string, StringValues> stringValues = this.Request.Query.FirstOrDefault(p => p.Key.ToLower() == pInfo.Name.ToLower());
-                    param[i] = (stringValues.Key == null) ? pInfo.ParameterInfo.HasDefaultValue ? pInfo.ParameterInfo.DefaultValue.ToString() : null : stringValues.Value[0];
-                }
+                List<ControllerParameterDescriptor> paramInfos = actionDescriptor.Parameters.OfType<ControllerParameterDescriptor>().ToList();
+
+                object[] paramsAsObjects = paramInfos.Select(paramInfo =>
```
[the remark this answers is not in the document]
monsieurleberre: fair enough, however I think the previous version is even worse (maybe it is personal)
do you have a better one ? I can make an explicit method of the logic in select... But I know you dislike mini methods too...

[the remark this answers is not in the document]
Aprogiena: Just don't use linq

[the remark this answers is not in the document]
bokobza: Just leave it as it was.

src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
```diff
@@ -90,27 +90,40 @@ private RPCResponse SendRPCRequest(RPCRequest request)
...
                 
                 // Prepare the named parameters that were passed via the query string in the order that they are expected by SendCommand.
                 List<ControllerParameterDescriptor> paramInfos = actionDescriptor.Parameters.OfType<ControllerParameterDescriptor>().ToList();
 
-                object[] paramsAsObjects = paramInfos.Select(paramInfo => 
-                    body.TryGetValue(paramInfo.Name, ignoreCase, out JToken jValue) && !jValue.HasValues
-                        ? jValue.ToString()
-                        : paramInfo.ParameterInfo.DefaultValue?.ToString())
-                    .Cast<object>().ToArray();
+                var paramsAsObjects = new object[paramInfos.Count];
```
monsieurleberre: Bad taste creates many more millionaires than good taste.
*Charles Bukowski* 😂

bokobza: we're on then!

Issue #2020, opened 2018-08-24
[Activation] Fix BlockNotification class and BlockNotificationTest


bokobza:
This was broken during the Activation work.
It needs to be fixed.

noescape00:
issue resolved

bokobza:
The light wallet is currently broken because these classes have been commented out during the activation work.
This issue can be resolved when either these classes are adapted to the new way of doing things or the light node will take the shape of a pruned node instead.

maciejzaleski:
Migrated to: https://dev.azure.com/Stratisplatformuk/StratisBitcoinFullNode/_workitems/edit/3055

PR #2021, opened 2018-08-24
[Activation] Fixed Notifications tests

bokobza:
image

Linked to https://github.com/stratisproject/StratisBitcoinFullNode/issues/2020

### DIFF
--- a/src/Stratis.Bitcoin.Features.Notifications.Tests/BlockNotificationTest.cs
+++ b/src/Stratis.Bitcoin.Features.Notifications.Tests/BlockNotificationTest.cs
@@ -17,7 +17,7 @@ namespace Stratis.Bitcoin.Features.Notifications.Tests
 {
     public class BlockNotificationTest : LogsTestBase
     {
-        private const string SkipReason = "Revisit these tests when BlockNotification is fixed.";
+        private const string RevisitWhenBlockNotificationFixed = "Revisit these tests when BlockNotification is fixed.";
 
         private readonly NodeLifetime lifetime;
         private readonly Mock<ISignals> signals;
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Notifications.Tests/BlockNotificationTest.cs
```diff
@@ -17,6 +17,8 @@ namespace Stratis.Bitcoin.Features.Notifications.Tests
 {
     public class BlockNotificationTest : LogsTestBase
     {
+        private const string SkipReason = "Revisit these tests when BlockNotification is fixed.";
```
majikandy: Create an issue for this so that it doesn't get lost when everyone is happy happy that we have a green build

bokobza: I created one earlier but didn't link it, thanks for spotting this!

src/Stratis.Bitcoin.Features.Notifications.Tests/BlockNotificationTest.cs
```diff
@@ -34,7 +36,7 @@ public BlockNotificationTest()
         /// Tests that <see cref="BlockNotification.Notify(System.Threading.CancellationToken)"/> exits due
         /// to <see cref="BlockNotification.StartHash"/> being null and no blocks were signaled.
         /// </summary>
-        [Fact]
+        [Fact(Skip = SkipReason)]
```
majikandy: wouldn't it be better if the text right here rather than a const ?   It is the same text so a const is valid, but since you would want to see the reason, it makes sense have it right next to it here.

Compromise would be to name the const   RevisitWhenBlockNotificationFixed

bokobza:
@majikandy how's this?

Issue #2014, opened 2018-08-23
Mark block invalid if partial of full validation fails


noescape00:
After FV or PV fail we should mark blocks as invalid AND when we receive a header we check against invalid blocks list and if presented header is there- ban the peer.

noescape00:
fixed

PR #2022, opened 2018-08-24
[Activation] Mark block invalid and don't create chained header for block that was already marked invalid

noescape00:
Removed `IsMarkedInvalid` and `MarkBlockInvalid` from `ChainState` because now we can access `IInvalidBlockStore` from consensus validation code. 
Instead started using there an instance of `IInvalidBlockStore` directly

Moved `ConsensusErrors` to Stratis.Bitcoin.Consensus and added there one more error for cases when hash was already banned

In `CHT.CreateAndValidateNewChainedHeader` check hash against banned hashes before we create chained header. If it's there- throw.

In `ConsensusRuleEngine` add hash of invalid block to banned hashes

Fix: #2014

### DIFF
--- a/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
+++ b/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
@@ -45,7 +45,7 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
             this.loggerFactory.AddConsoleWithFilters();
 
             this.CallbacksCalled = new Dictionary<uint256, Block>();
-            this.ChainState = new ChainState(new InvalidBlockHashStore(new DateTimeProvider())) {ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
+            this.ChainState = new ChainState(){ ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
 
             this.Puller = new ExtendedBlockPuller(this.ChainState, new NodeSettings(new StratisMain()), new DateTimeProvider(), this.loggerFactory);
         }
--- a/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
+++ b/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
@@ -45,7 +45,7 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
             this.loggerFactory.AddConsoleWithFilters();
 
             this.CallbacksCalled = new Dictionary<uint256, Block>();
-            this.ChainState = new ChainState(){ ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
+            this.ChainState = new ChainState() {ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
 
             this.Puller = new ExtendedBlockPuller(this.ChainState, new NodeSettings(new StratisMain()), new DateTimeProvider(), this.loggerFactory);
         }
--- a/src/Stratis.Bitcoin/Consensus/ConsensusErrors.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusErrors.cs
@@ -1,5 +1,4 @@
-﻿
-namespace Stratis.Bitcoin.Consensus
+﻿namespace Stratis.Bitcoin.Consensus
 {
     /// <summary>
     /// A class that holds consensus errors.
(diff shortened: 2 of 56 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -220,6 +226,13 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
             catch (ConsensusErrorException ex)
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
+
+                uint256 hashToBan = ruleContext.ValidationContext.ChainedHeaderToValidate.HashBlock;
+
+                if (ruleContext.ValidationContext.RejectUntil != null)
```
noescape00: Here should be 2nd rule that doesn't lead to permanent block hash banning. Who remembers what is it?

noescape00: Fixed. It was mutated merkle tree exception

src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
```diff
@@ -45,7 +45,7 @@ public BlockPullerTestsHelper()
             this.loggerFactory.AddConsoleWithFilters();
 
             this.CallbacksCalled = new Dictionary<uint256, Block>();
-            this.ChainState = new ChainState(new InvalidBlockHashStore(new DateTimeProvider())) {ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
+            this.ChainState = new ChainState(){ ConsensusTip = ChainedHeadersHelper.CreateGenesisChainedHeader()};
```
Aprogiena: Inconsistent spacing

src/Stratis.Bitcoin/Consensus/ConsensusErrors.cs
```diff
@@ -1,6 +1,5 @@
-﻿using Stratis.Bitcoin.Consensus;
-
-namespace Stratis.Bitcoin.Features.Consensus
+﻿
```
Aprogiena: Empty line

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -220,6 +226,16 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
             catch (ConsensusErrorException ex)
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
+
+                if (ruleContext.ValidationContext.Error != ConsensusErrors.BadTransactionDuplicate)
```
Aprogiena: I'm not sure if this is the best Place to have this codeIt feels very hacky here

src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerTestsHelper.cs
noescape00: fixed

src/Stratis.Bitcoin/Consensus/ConsensusErrors.cs
noescape00: fixed

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
noescape00: Feel free to suggest another place, I don't mind. 

I thought about following places as well: 
in block validator- that will require copy-pasting this code in 3 places or moving to a separated class
in a new rule- that will require making public property on consensus rule engine to expose invalid block hashes store so the rule can access it- felt like an ugly hack to me.

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -226,17 +232,26 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
...
+        private void HandleConsensusError(ValidationContext validationContext)
+        {
+            this.logger.LogTrace("()");
 
-                    if (ruleContext.ValidationContext.RejectUntil != null)
-                        this.invalidBlockHashStore.MarkInvalid(hashToBan, ruleContext.ValidationContext.RejectUntil);
-                    else
-                        this.invalidBlockHashStore.MarkInvalid(hashToBan);
-                }
+            if (validationContext.Error == ConsensusErrors.BadWitnessNonceSize)
```
Aprogiena: Probably a comment would be nice saying that this is handled in consensus manager and why the block should not be banned

noescape00: fixed

Issue #2015, opened 2018-08-23
Cache blocks that are being disconnected and use cache for reconnection if needed


noescape00:
Here is a scenario: 

We are at height 1000a and we see chain of height 1005b with a forkpoint at 700a.
 
We download the chain and we start a reorg- first rewind till 700a and then connect one by one new blocks. When we connect we signal which leads to pushing 701b to BSQ. That leads to reorg in BSQ. 
Now we find out that 704b is invalid and we rewind back to 700a to reconnect old chain. 

However now we can't reconnect the chain because blocks from 700a are not in the block store and not in CHT (CHT should have 900a-1000a and 701b-1005b). 

Therefore when we start disconnecting blocks we should save those to a list and use in case we need reconnect the chain. 
When we exit reorg method we should clean the list since now we don't need those blocks.

PR #2025, opened 2018-08-24
[Activation] Cache disconnected blocks

noescape00:
Fix: #2015

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -517,7 +517,7 @@ namespace Stratis.Bitcoin.Consensus
 
             if (!isExtension)
             {
-                // Save blocks that will be disconnected in case we will need to reconnect them (this might happen connection of a new chain fails).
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -513,8 +513,23 @@ private async Task<ConnectBlocksResult> FullyValidateLockedAsync(ChainedHeader n
             // If the new block is not on the current chain as our current consensus tip then rewind consensus tip to the common fork.
             bool isExtension = fork == oldTip;
 
+            List<ChainedHeaderBlock> disconnectedBlocks = null;
+
             if (!isExtension)
             {
+                // Save blocks that will be disconnected in case we will need to reconnect them (this might happen if connection of a new chain fails).
```
Aprogiena: This whole new code can go into separated method

Aprogiena: Maybe even better, you could add a new argument to RewindToForkPointAsync, and if that argument would be true, that method would return the list of disconnected blocks because it essentially has those blocks because it signals them. Or maybe it could return the list every time because it is not that common that we don't need the result

Issue #2027, opened 2018-08-24
Unimplemented Mempool validator tests


majikandy:
Removed the following tests from the Mempool validator test suite as they are "fake green" tests and give an illusion that things are covered that aren't

Saved here so that they can be brought in as and when (and if) they are implemented.

```
        [Fact]
        public void CheckFinalTransaction_WithStandardLockTimeAndValidTxTime_ReturnsTrue()
        {
            // TODO: Implement test
        }

        [Fact]
        public void CheckFinalTransaction_WithNoLockTimeAndValidTxTime_ReturnsTrue()
        {
            // TODO: Implement test
        }

        [Fact]
        public void CheckFinalTransaction_WithStandardLockTimeAndExpiredTxTime_Fails()
        {
            // TODO: Implement test
        }

        [Fact]
        public void CheckFinalTransaction_WithNoLockTimeLockTimeAndExpiredTxTime_Fails()
        {
            // TODO: Implement test
        }

        [Fact]
        public void CheckSequenceLocks_WithExistingLockPointAndValidChain_ReturnsTrue()
        {
            // TODO: Test case- Check two cases, chain with/without previous block
            // No Previous - time of lock == 0
            // Previous - time of lock < tip chain median time
        }

        [Fact]
        public void CheckSequenceLocks_WithExistingLockPointAndChainWithPrevAndExpiredBlockTime_Fails()
        {
            // TODO: Test case - The lock point MinTime exceeds chain previous block median time
        }

        [Fact]
        public void CheckSequenceLocks_WithExistingLockPointAndChainWihtNoPrevAndExpiredBlockTime_Fails()
        {
            // TODO: Test case - the lock point MinTime exceeds 0
        }

        [Fact]
        public void CheckSequenceLocks_WithExistingLockPointAndBadHeight_Fails()
        {
            // TODO: Test case - lock point height exceeds chain height
        }

        [Fact]
        public void GetTransactionWeight_WitnessTx_ReturnsWeight()
        {
            // TODO: Test getting tx weight on transaction with witness
        }

        [Fact]
        public void GetTransactionWeight_StandardTx_ReturnsWeight()
        {
            // TODO: Test getting tx weight on transaction without witness
        }

        [Fact]
        public void CalculateModifiedSize_CalcWeightWithTxIns_ReturnsSize()
        {
            // TODO: Test GetTransactionWeight - InputSizes = CalculateModifiedSize
            // Test weight calculation by input nTxSize = 0
[...]

PR #2028, opened 2018-08-24
[Activation] Removed mempool validator tests that aren't implemented

majikandy:
Removed 45 unimplemented tests.

Added them to Issue #2027 so they can be implemented when it is a priority.

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
@@ -19,6 +19,82 @@ namespace Stratis.Bitcoin.Features.MemoryPool.Tests
         {
         }
 
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckFinalTransaction_WithStandardLockTimeAndValidTxTime_ReturnsTrue()
+        {
+            // TODO: Implement test
+        }
+
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckFinalTransaction_WithNoLockTimeAndValidTxTime_ReturnsTrue()
+        {
+            // TODO: Implement test
+        }
+
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckFinalTransaction_WithStandardLockTimeAndExpiredTxTime_Fails()
+        {
+            // TODO: Implement test
+        }
+
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckFinalTransaction_WithNoLockTimeLockTimeAndExpiredTxTime_Fails()
+        {
+            // TODO: Implement test
+        }
+
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckSequenceLocks_WithExistingLockPointAndValidChain_ReturnsTrue()
+        {
+            // TODO: Test case- Check two cases, chain with/without previous block
+            // No Previous - time of lock == 0
+            // Previous - time of lock < tip chain median time
+        }
+
+        [Fact(Skip = "Not implemented yet.")]
+        public void CheckSequenceLocks_WithExistingLockPointAndChainWithPrevAndExpiredBlockTime_Fails()
+        {
+            // TODO: Test case - The lock point MinTime exceeds chain previous block median time
+        }
+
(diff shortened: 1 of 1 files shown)


bokobza:
Instead of removing them, can you please add Skip="some reason" or maybe actually implement them?

majikandy:
It is better to remove them

majikandy:
>Imho they should be implemented instead of being removed.

Absolutely agree - which is why an issue is created, so that can be prioritised into appropriate sprint

noescape00:
>Absolutely agree - which is why an issue is created, so that can be prioritised into appropriate sprint

So let's have the issue but close the PR. 
I don't see a reason why they should be removed from the codebase and stored in the issue

majikandy:
YAGNI is the main reason for removing. Worth noting that it is a copy-paste away from going back into the codebase when the time comes to implement them.   It is good that the thought has gone in to these scenarios so we don't want to lose that.

bokobza:
When you skip them, they will appear as skipped in your test window so we'll always be reminded that they need to be done.
Not keen for removing them at all though.

bokobza:
By all means, please create the issue.

majikandy:
>When you skip them, they will appear as skipped in your test window so we'll always be reminded that they need to be done.

No you won't , you'll get used to ignoring them

majikandy:
Issue #2027  already created and contains the exact stubs to paste back in when the time comes.

bokobza:
and how is removing them going to help?

majikandy:
makes you think less when you don't see anything there in the file - so you don't have to interrogate why skipped

majikandy:
and YAGNI - You ain't gonna need it - we might never ever need to implement these tests, so they might sit there skipped forever causing team thinking time over and over and over again.

majikandy:
Note though, I'm totally up for them being implemented, but is it a priority right now over getting back to a working codebase/test suite ?

bokobza:
Cool, maybe we can rediscuss.
Can you please close this PR? YAGNI or not, I don't want to delete them just to put them back next week.

majikandy:
I totally disagree though,  compromise of skip?

majikandy:
final point... Why so scared of deleting them? we have source control

bokobza:
yeah skip is good.
Try it, you'll see it pops off the screen. is nice

majikandy:
I know it does, but drawing attention to them is what make developers spend too much time on them

majikandy:
Skip reason="Not implemented yet" ?

bokobza:
Sounds good

majikandy:
done

src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
```diff
@@ -19,76 +19,76 @@ public MempoolValidatorTest() : base(KnownNetworks.RegTest)
...
-        [Fact]
+        [Fact(Skip = "Not implemented yet.")]
         public void CalculateModifiedSize_CalcWeightWithTxIns_ReturnsSize()
         {
             // TODO: Test GetTransactionWeight - InputSizes = CalculateModifiedSize
             // Test weight calculation by input nTxSize = 0
         }
 
-        [Fact]
+        [Fact(Skip = "Not implemented yet.")]
```
majikandy: oops

majikandy: Is this test below valid??? has a TODO in it.....

src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
```diff
@@ -19,76 +19,76 @@ public MempoolValidatorTest() : base(KnownNetworks.RegTest)
...
         public void CalculateModifiedSize_PreCalcWeightWithTxIns_ReturnsSize()
         {
             // TODO: Test GetTransactionWeight - InputSizes = CalculateModifiedSize
```
majikandy: This test ignored too as it has this todo in it - so unknown if this is a valid test or not

bokobza: Fair enough

PR #2036, opened 2018-08-26
Make fee estimate conditional on whether blocks are full

### LINKED ISSUE #1283: Min fee estimation for a wallet is too high
Wallet min fee estimation is ~60 times bigger than actual. 
With empty blocks the min fee to get into the block is: `tx serialized size / 100000000` (`PowBlockAssembler.cs:499`)

So for a tx that is 229 bytes fee is 0.00000229 and the wallet estimation algho says its 0.00013800

mikedennis:
Fixes #1283 

Will look at number of transaction in last 3 blocks to determine whether to use minRelayTxFee (0.0001) or fallbackTxFee *0.0006) rate for fee estimation on full node wallet.

Have tested with FullNodeUI wallet on testnet, with empty blocks.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Features.BlockStore;
 using Stratis.Bitcoin.Features.Wallet.Interfaces;
 
 namespace Stratis.Bitcoin.Features.Wallet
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
```diff
@@ -33,17 +36,35 @@ public class WalletFeePolicy : IWalletFeePolicy
...
+
         /// <summary>
         /// Constructs a wallet fee policy.
         /// </summary>
         /// <param name="nodeSettings">Settings for the the node.</param>
-        public WalletFeePolicy(NodeSettings nodeSettings)
+        /// <param name="chain">Header chain.</param>
+        /// <param name="blockRepository">Repository containing blocks.</param>
+        /// <param name="loggerFactory">Factory for creating instance logger.</param>
+        public WalletFeePolicy( NodeSettings nodeSettings,
```
Aprogiena: Spacing and indentation for the following lines

bokobza:
I'll review this today.

src/Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
```diff
@@ -33,17 +36,35 @@ public class WalletFeePolicy : IWalletFeePolicy
         /// </summary>
         private readonly FeeRate minRelayTxFee;
 
+        /// <summary>Header chain.</summary>
+        private readonly ConcurrentChain chainState;
+
+        /// <summary>Repository containing blocks.</summary>
+        private readonly IBlockRepository blockRepository;
```
dangershony: Dont use `IBlockRepository` use IBlockStore instead.
Also I think wallet does not have a dependency on IBlockStore being present so you need to allow it to be null and I guess use `minRelayTxFee` if store is null.

src/Stratis.Bitcoin.Features.Wallet/WalletFeePolicy.cs
```diff
@@ -33,17 +36,35 @@ public class WalletFeePolicy : IWalletFeePolicy
         /// </summary>
         private readonly FeeRate minRelayTxFee;
 
+        /// <summary>Header chain.</summary>
+        private readonly ConcurrentChain chainState;
```
dangershony: You should probably use ChainState not CuncurrentChain

bokobza:
I can't find any trace of 60K satoshis per Kb in StratisX.
I'll be happy with the fallback fee to be the same as the min relay trx fee, as suggested by Mike here: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1283#issuecomment-416066744

mikedennis:
As per Jeremy's comment, these are our current defaults for tx fees:
Bitcoin
```
            this.MinTxFee = 1000;
            this.FallbackFee = 20000;
            this.MinRelayTxFee = 1000;
```

Stratis
```
            this.MinTxFee = 10000;
            this.FallbackFee = 60000;
            this.MinRelayTxFee = 10000;
```

Looks like StratisX uses a harded coded fee of 10000.

dangershony:
I have no idea where the `  this.FallbackFee = 60000;` came from, perhaps a mistake I made long time ago.
So lets default to StratisX and replace the 60k to the 10k of stratisX

mikedennis:
Ok sounds good. I will make a new PR and then close this one out as it's a one liner now 😜

mikedennis:
Replaced by PR https://github.com/stratisproject/StratisBitcoinFullNode/pull/2049

PR #2038, opened 2018-08-27
[SC] Remove unused fields on TransactionContext

rowandh:
Removes some unused fields on TransactionContext. Changes the Type of the contract transaction data to be bytes rather than having to pull them from the Script.

### DIFF
--- a/src/Stratis.SmartContracts.Core/SmartContractTransactionContext.cs
+++ b/src/Stratis.SmartContracts.Core/SmartContractTransactionContext.cs
@@ -19,18 +19,6 @@ namespace Stratis.SmartContracts.Core
         private readonly Money mempoolFee;
 
         /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


src/Stratis.SmartContracts.Core/SmartContractTransactionContext.cs
```diff
@@ -56,13 +43,11 @@ public uint Nvout
         }
 
         /// <inheritdoc />
-        public IEnumerable<byte> ContractData
+        public byte[] Data
```
fassadlr: Isn't `Data` like a reserved keyword or something? 😋 I think we can come with a better name? I personally thought ContractData wasnt too bad?

codingupastorm: Maybe SerializedData ?

PR #2039, opened 2018-08-27
[SC] Byte serialization

### LINKED ISSUE #1714: Within smart contracts we need to be able to convert all allowed types to byte arrays and back.
We don't want to open direct access to System.Text or BitConverter, however we still need to convert between bytes and the below types. Useful for hashing etc.

Bool should go back and forwards with byte.

The others should be translatable to bytes:
* int
* uint 
* long
* ulong
* string (Encoding.UTF8, and to/from hex)
* Address

We should have a NuGet package purely for this implementation, likely something like Stratis.SmartContracts.ByteHelper.

codingupastorm:
Closes #1714.

### DIFF
--- a/src/Stratis.SmartContracts/ISerializer.cs
+++ b/src/Stratis.SmartContracts/ISerializer.cs
@@ -50,22 +50,22 @@
         Address ToAddress(byte[] val);
 
         /// <summary>
-        /// Serializes bytes into an integer.
+        /// Serializes first 4 bytes of a byte array into an integer.
         /// </summary>
         int ToInt32(byte[] val);
 
         /// <summary>
-        /// Serializes bytes into an unsigned integer.
+        /// Serializes first 4 bytes of a byte array into an unsigned integer.
         /// </summary>
         uint ToUInt32(byte[] val);
 
         /// <summary>
-        /// Serializes bytes into a long. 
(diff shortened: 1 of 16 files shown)


src/Stratis.SmartContracts/ISerializer.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        /// <summary>
+        /// Serializes 20-bytes into an address.
+        /// </summary>
+        Address ToAddress(byte[] val);
+
+        /// <summary>
+        /// Serializes bytes into an integer.
+        /// </summary>
+        int ToInt32(byte[] val);
```
rowandh: Documentation - val must be 4 bytes long.

src/Stratis.SmartContracts/ISerializer.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Serializes bytes into an unsigned integer.
+        /// </summary>
+        uint ToUInt32(byte[] val);
+
+        /// <summary>
+        /// Serializes bytes into a long. 
+        /// </summary>
+        long ToInt64(byte[] val);
+
```
rowandh: Documentation - val must be 8 bytes long.

Issue #2030, opened 2018-08-24
When partial validation fails with BadWitnessNonceSize require peer to provide same block but with witness data


noescape00:
Basically implement what we used to have in consensus loop but do it in consensus manager. 

ConsensusLoop's code: 

```
if (validationContext.Error == ConsensusErrors.BadWitnessNonceSize)
{
    this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
    this.connectionManager.AddDiscoveredNodesRequirement(NetworkPeerServices.NODE_WITNESS);
    this.Puller.RequestOptions(TransactionOptions.Witness);
    return;
}
```

noescape00:
@bokobza can you please add this issue to the sprint?

bokobza:
Closed with https://github.com/stratisproject/StratisBitcoinFullNode/pull/2043

PR #2043, opened 2018-08-28
[Activation] Witness requirments

noescape00:
Fix: #2030

### DIFF
--- a/src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
+++ b/src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
@@ -37,6 +37,15 @@ namespace Stratis.Bitcoin.BlockPulling
     {
         void Initialize(BlockPuller.OnBlockDownloadedCallback callback);
 
+        /// <summary>
+        /// Adds <see cref="NetworkPeerServices.NODE_WITNESS"/> to list of services that are required from all peer nodes.
+        /// </summary>
+        /// <remarks>
+        /// In case some of the nodes that we are already requesting block from don't support new
+        /// requirements those nodes will be released from their assignments.
+        /// </remarks>
+        void RequestWitnessOptions();
+
         /// <summary>Gets the average size of a block based on sizes of blocks that were previously downloaded.</summary>
         double GetAverageBlockSizeBytes();
 
@@ -240,6 +249,35 @@ namespace Stratis.Bitcoin.BlockPulling
 
             this.assignerLoop = this.AssignerLoopAsync();
             this.stallingLoop = this.StallingLoopAsync();
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public void RequestWitnessOptions()
+        {
+            this.logger.LogTrace("()");
+
+            this.networkPeerRequirement.RequiredServices |= NetworkPeerServices.NODE_WITNESS;
+
+            var peerIdsToRemove = new List<int>();
+
+            lock (this.peerLock)
+            {
+                foreach (KeyValuePair<int, IBlockPullerBehavior> peerIdToBehavior in this.pullerBehaviorsByPeerId)
+                {
+                    INetworkPeer peer = peerIdToBehavior.Value.AttachedPeer;
+
+                    if ((peer == null) || !this.networkPeerRequirement.Check(peer.PeerVersion))
+                    {
+                        this.logger.LogDebug("Peer Id {0} does not meet requirements.", peerIdToBehavior.Key);
+                        peerIdsToRemove.Add(peerIdToBehavior.Key);
+                    }
+                }
+            }
+
+            foreach (int peerId in peerIdsToRemove)
+                this.PeerDisconnected(peerId);
 
             this.logger.LogTrace("(-)");
         }
--- a/src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
+++ b/src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
@@ -38,10 +38,10 @@ namespace Stratis.Bitcoin.BlockPulling
         void Initialize(BlockPuller.OnBlockDownloadedCallback callback);
 
         /// <summary>
-        /// Adds <see cref="NetworkPeerServices.NODE_WITNESS"/> to list of services that are required from all peer nodes.
+        /// Adds <see cref="NetworkPeerServices.NODE_WITNESS"/> to list of services that are required from all peers.
         /// </summary>
         /// <remarks>
-        /// In case some of the nodes that we are already requesting block from don't support new
+        /// In case some of the peers that we are already requesting block from don't support new
         /// requirements those nodes will be released from their assignments.
         /// </remarks>
         void RequestWitnessOptions();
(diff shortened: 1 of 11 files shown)


src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -244,6 +253,35 @@ public void Initialize(OnBlockDownloadedCallback callback)
             this.logger.LogTrace("(-)");
         }
 
+        /// <inheritdoc />
+        public void RequestWitnessOptions()
```
noescape00: here we might want to add a param for NetworkPeerServices to be able to request different options than `NODE_WITNESS` but I'm not sure if we really need that so this method is for witness only right now.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -37,6 +37,15 @@ public interface IBlockPuller : IDisposable
...
         void Initialize(BlockPuller.OnBlockDownloadedCallback callback);
 
+        /// <summary>
+        /// Adds <see cref="NetworkPeerServices.NODE_WITNESS"/> to list of services that are required from all peers.
+        /// </summary>
+        /// <remarks>
+        /// In case some of the peers that we are already requesting block from don't support new
+        /// requirements those nodes will be released from their assignments.
+        /// </remarks>
+        void RequestWitnessOptions();
```
dangershony: This is problematic we are making witness part of the puller.
I think we can solve it differently bu passing the Network Enum here..

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
dangershony: Ah yes I agree

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -99,7 +103,8 @@ public class ConsensusManager : IConsensusManager
             ConcurrentChain chain,
             IBlockPuller blockPuller,
             IBlockStore blockStore,
-            IInvalidBlockHashStore invalidHashesStore)
+            IInvalidBlockHashStore invalidHashesStore,
```
dangershony: What is `IInvalidBlockHashStore` hmm need to check it out.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -355,14 +361,13 @@ private async Task OnPartialValidationCompletedCallbackAsync(ValidationContext v
...
-                //    this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
-                //    this.connectionManager.AddDiscoveredNodesRequirement(NetworkPeerServices.NODE_WITNESS);
-                //    this.Puller.RequestOptions(TransactionOptions.Witness);
-                //    return;
-                //}
+                if (validationContext.Error == ConsensusErrors.BadWitnessNonceSize)
+                {
+                    this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
+                    this.connectionManager.AddDiscoveredNodesRequirement(NetworkPeerServices.NODE_WITNESS);
+                    this.blockPuller.RequestWitnessOptions();
```
dangershony: I suggest we do the same as with the connection manager and pass the network service enum as a method parameter, and this can be set in the Validation context this will make the low level oblivious to any witness changes.
Then just before we throw `ConsensusErrors.BadWitnessNonceSize` you can set the enum.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
noescape00: It used to be part of the old puller before. 

Basically we need a way to filter out peers and pick only those that can provide us with the blocks we want.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: It's a list of banned block's hashes. 
We used to use it in consensusLoop and then forgot to add to rules engine.
I've wired it up a few days ago.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -355,14 +361,13 @@ private async Task OnPartialValidationCompletedCallbackAsync(ValidationContext v
...
 
-                // TODO ACTIVATION implement following (used to be in consensus loop)
-                //if (validationContext.Error == ConsensusErrors.BadWitnessNonceSize)
-                //{
-                //    this.logger.LogInformation("You probably need witness information, activating witness requirement for peers.");
-                //    this.connectionManager.AddDiscoveredNodesRequirement(NetworkPeerServices.NODE_WITNESS);
-                //    this.Puller.RequestOptions(TransactionOptions.Witness);
-                //    return;
-                //}
+                if (validationContext.Error == ConsensusErrors.BadWitnessNonceSize)
```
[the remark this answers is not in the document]
noescape00: ConsensusErrors are moved to Base

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Ok. Thought that would be needed. Will fix.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
noescape00: fixed

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: fixed

dangershony: This is still only half the fix.
`NetworkPeerServices` can be set as a nullable type on context

Then we do 
```
if(validationContext.RequiredNetworkService != null)
   bla..
```

noescape00:
@dangershony Apro suggested that we don't touch the puller so puller always works with same set of nodes we are connected to. 
That will allow reducing this pr to 4 lines but it will remove some flexibility (which we don't need currently but we might need in the future). 
I don't mind going with that. 

What do you think?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: I'm not sure if we want to pass services using context. 
Data in the context should be input data for the rules. So here in order to allow a rule to provide output data we will change the context for every single rule, when instead we can handle particular error like we are doing now.

dangershony: Well what I want to try to avoid is put witness related in CM so if we keep in ValidationContext the such a property (similar to how we keep the node banning flag) should be ok no?

noescape00:
>Also you should out to do to your list to add new test for consensus manager for the new functionality

done

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Fixed: https://github.com/stratisproject/StratisBitcoinFullNode/pull/2043/commits/5d6499da16f0e4e00e53d92df71e52a033b9c5bd

PR #2050, opened 2018-08-28
Fixed  Stratis.Bitcoin.Features.Consensus.Tests.Rules.CommonRules

bokobza:


src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/RulesRegistrationOrderTest.cs
```diff
@@ -54,9 +57,10 @@ public void StratisMainPowConsensusRulesRegistrationTest()
         [Fact]
         public void StratisMainPosConsensusRulesRegistrationTest()
         {
-            new FullNodeBuilderConsensusExtension.PosConsensusRulesRegistration().RegisterRules(this.network.Consensus);
+            Network network = new StratisTest();
```
Aprogiena: This is weird change. The test name specifically includes Stratis main and you're now using Stratis test.?

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/RulesRegistrationOrderTest.cs
```diff
@@ -1,19 +1,22 @@
...
 namespace Stratis.Bitcoin.Features.Consensus.Tests.Rules.CommonRules
 {
-    public class ConsensusRulesRegistrationTest : TestConsensusRulesUnitTestBase
+    public class ConsensusRulesRegistrationTest
     {
         [Fact]
         public void StratisMainPowConsensusRulesRegistrationTest()
         {
-            new FullNodeBuilderConsensusExtension.PowConsensusRulesRegistration().RegisterRules(this.network.Consensus);
+            Network network = new BitcoinTest();
```
Aprogiena: Same here – the network does not match the test name

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/CommonRules/RulesRegistrationOrderTest.cs
bokobza: ah yes true, didn't see that.
There is no difference between the two so I'll remove Main from the test names.

bokobza:
@zeptin at the moment, the rules are the same. If we introduce new rules specific to a particular network, we will definitely add a test for it.

PR #2051, opened 2018-08-28
Listmethods return value

zeptin:
The `listmethods` API call returns an ugly stacktrace if the `-server` config option was not supplied. This fixes it so it merely returns an empty list instead.


src/Stratis.Bitcoin.Features.RPC/Controllers/RPCController.cs
```diff
@@ -138,7 +144,15 @@ public IActionResult ListMethods()
...
             {
                 var listMethods = new List<Models.RpcCommandModel>();
-                foreach (ControllerActionDescriptor descriptor in this.GetActionDescriptors().Values.Where(desc => desc.ActionName == desc.ActionName.ToLower()))
+                Dictionary<string, ControllerActionDescriptor> actionDescriptors = this.GetActionDescriptors();
+
+                // -server config option was not supplied, so the RPCHost is not set up.
+                if (actionDescriptors == null)
+                    return this.Json(listMethods);
+
+                IEnumerable<ControllerActionDescriptor> descriptors = actionDescriptors.Values.Where(desc => desc.ActionName == desc.ActionName.ToLower());
```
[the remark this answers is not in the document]
zeptin: The RPC methods are all in lowercase, I presume this is why this was done in the first place. I will look into cleaning it up per your suggestions.

bokobza:
@zeptin Good catch.
A better solution would be to add
```
if (!this.rpcSettings.Server)
{
    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", "Method not allowed when RPC is disabled.");
}
```
at the beginning of both methods.

src/Stratis.Bitcoin.Features.RPC.Tests/Controller/RPCControllerTest.cs
```diff
@@ -151,6 +152,9 @@ public void CallByName_CorrectQueryString_SendsRpcRequest_ReturnsResult()
                 });
             }
             this.descriptors.Add(descriptor);
+            var values = new Dictionary<string, StringValues>();
```
bokobza: Is this needed?

zeptin: This must have shown up while fixing the conflict, not exactly sure where it came from. Removed.

Issue #2008, opened 2018-08-23
Peer endpoint is never set in the validation context


noescape00:
Peer endpoint is never set in the validation context of the consensus rules engine but it's used for banning in case validation failed. 

We need to set it for every validation in case header/block was provided by a peer and not mined

noescape00:
@dangershony would you like to fix it or I can do it?

bokobza:
Closed with https://github.com/stratisproject/StratisBitcoinFullNode/pull/2052

PR #2052, opened 2018-08-28
Peer endpoint in validation context

zeptin:
Relates to the following issues:

#2008
#1936 (specifically the PeerBanningTests)

The PeerBanningTests are not yet completely working. This is because, for the most part, they are using the `BlockMinedAsync` method in tests relating to peer banning. This is nonsensical, as there is no banning performed within `BlockMinedAsync` by design (as the local node is the source of the block and obviously cannot ban itself). It is further presumed that the local node will not knowingly deliver invalid blocks to peers, so the tests as currently envisioned have to be rethought. Possibly `CM.BlockDownloaded` could be made public and used instead of `BlockMinedAsync`, but this introduces the difficulty of having to obtain the peer ID within the tests.

At least one test also has contradictory naming and asserts, i.e. `NodeIsSynced_PeerSendsABadBlockAndPeerDisconnectedAndAddressIsNull_ThePeerGetsBanned_Async` has `Assert.False(context.PeerBanning.IsBanned(peerEndPoint));`, where you would expect the Assert to be true based on the name of the test alone.

This PR is therefore primarily to get a preliminary review of the changes to the validation context etc., the final fixes to the PeerBanningTests will follow pending additional discussion.


src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -99,7 +99,8 @@ public class BlockPuller : IBlockPuller
 
         /// <param name="blockHash">Hash of the delivered block.</param>
         /// <param name="block">The block.</param>
-        public delegate void OnBlockDownloadedCallback(uint256 blockHash, Block block);
+        /// <param name="peerId">The ID of the peer delivering the block.</param>
```
noescape00: The block was already delivered.
Maybe `Id of a peer that delivered the block`

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -738,7 +741,7 @@ private async Task<ConnectBlocksResult> ConnectBlockAsync(ChainedHeaderBlock blo
             }
 
             // Call the validation engine.
-            ValidationContext validationContext = await this.consensusRules.FullValidationAsync(blockToConnect).ConfigureAwait(false);
+            ValidationContext validationContext = await this.fullValidator.ValidateAsync(blockToConnect.ChainedHeader, blockToConnect.Block).ConfigureAwait(false);
```
noescape00: Please change interface of the validator to accept `ChainedHeaderBlock`

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -915,9 +918,9 @@ private void DownloadBlocks(ChainedHeader[] chainedHeaders, OnBlockDownloadedCal
             this.logger.LogTrace("(-)");
         }
 
-        private void BlockDownloaded(uint256 blockHash, Block block)
+        private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
         {
-            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);
+            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash); // TODO: Log peerID?
```
noescape00: Yes, please log peer Id

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -959,7 +962,11 @@ private void BlockDownloaded(uint256 blockHash, Block block)
 
                 if (result.Error != null)
                 {
-                    this.peerBanning.BanAndDisconnectPeer(result.Peer, result.BanDurationSeconds, $"Integrity validation failed: {result.Error.Message}");
+                    // It is possible for multiple peers to be claiming a particular ChainedHeader. We only ban the
```
noescape00:
```
When integrity validation fails we want to ban only particular peer that provided invalid block. Integrity validation failing for this block doesn't automatically make other blocks with the same hash invalid, therefore banning other peers that claim to be on a chain that contains a block with the same hash is not required. 
```

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
```diff
@@ -13,6 +14,12 @@ namespace Stratis.Bitcoin.Consensus.Validators
     /// <param name="validationContext">Result of the validation including information about banning if necessary.</param>
     public delegate Task OnPartialValidationCompletedAsyncCallback(ValidationContext validationContext);
 
+    /// <summary>
+    /// A callback that is invoked when <see cref="IFullValidator.StartFullValidation"/> completes validation of a block.
+    /// </summary>
+    /// <param name="validationContext">Result of the validation including information about banning if necessary.</param>
+    public delegate Task OnFullValidationCompletedAsyncCallback(ValidationContext validationContext);
```
noescape00: Not needed. FV is always done synchronously unlike PV.

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
```diff
@@ -48,6 +55,31 @@ public interface IPartialValidator : IDisposable
...
+        /// <summary>
+        /// Schedules a block for background full validation.
+        /// <para>
+        /// Full validation may involve changes to the underlying store like rewinding or updating the database.
+        /// </para>
+        /// </summary>
+        /// <param name="header">The chained header that is going to be validated.</param>
+        /// <param name="block">The block that is going to be validated.</param>
+        /// <param name="onFullValidationCompletedAsyncCallback">A callback that is called when validation is complete.</param>
+        void StartFullValidation(ChainedHeader header, Block block, OnFullValidationCompletedAsyncCallback onFullValidationCompletedAsyncCallback);
```
noescape00: Not needed. FV is always done synchronously unlike PV.

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
```diff
@@ -198,4 +230,92 @@ public override string ToString()
...
+                ChainedHeader = header,
+                Block = block,
+                FullValidationCompletedAsyncCallback = onFullValidationCompletedAsyncCallback
+            });
+
+            this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public async Task<ValidationContext> ValidateAsync(ChainedHeader header, Block block)
```
noescape00: nothing in this class is needed except for this method

noescape00:
As discussed on slack: in case of PV or FV failing we need to ban the peers that claim bogus block. 
We are currently collecting the peers that claimed FV failed block but we don't ban them. 
This logic is missing from this PR.

zeptin:
@noescape00 I also thought it was missing, but realised it was not and reverted adding banning logic for FV. It is quite a circuitous call path in `ConsensusManager`, but it does happen:

`BlockDownloaded` gets called when a block arrives
It in turn calls the `OnBlockDownloadedCallback` i.e. `ProcessDownloadedBlock`
PV is then performed.
Callback `OnPartialValidationCompletedCallbackAsync` called.
I will ignore the case where PV fails, as it is relatively clear it results in banning.
`OnPartialValidationSucceededAsync` then checks if FV is needed.
`FullyValidateLockedAsync` gets called if so, and returns a `ConnectBlocksResult`.
The `ConnectBlocksResult` contains a `PeersToBan` property, which results in banning if non-empty.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
zeptin: Do you mean the signature for `this.consensusRules.FullValidationAsync` (which has a different signature to the PV version for some reason)? All the validators in BlockValidator.cs currently have a `(header, block)` signature, unless you want that to be changed unilaterally.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
zeptin: Done

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
zeptin: Done

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
zeptin: Done

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Yeah, change the signature. 
It looks like internally validator constructs ChainedHeaderBlock again so we might easily save allocation here by avoiding creating another ChainedHeaderBlock there.

noescape00:
So when we are doing the connection the trace is the following: 

`OnPartialValidationSucceededAsync` => `FullyValidateLockedAsync` (call on line 415) => `ConnectChainAsync` (call on line 546) => `ConnectBlockAsync` (call on line 628) => `fullValidator.ValidateAsync` (call on line 744).

Let's say we failed FV, that's what will happen: 

```
ValidationContext validationContext = await this.fullValidator.ValidateAsync
```
will return error and we will get the peers to ban in CHT and return them as return variable of `ConnectBlockAsync`. 

On line 635 `ConnectChainAsync` will propagate this result to the caller. 

Now we are on line 546 and `connectBlockResult.Succeeded` is false. 

Since we failed we will try to reconnect the chain and call `ReconnectOldChainAsync` and return whatever this method returns. 
**Here is where we lost peer Ids to ban.** Those peers are still in this variable: `ConnectBlocksResult connectBlockResult` (line 546)

So now we get back to `OnPartialValidationSucceededAsync` line 415 and then we will ban an empty list of peers on line 437.


On another note: 

Since we call `CHT.PartialOrFullValidationFailed` in `ConnectBlockAsync` we don't need to call it again on BlockMined. (that's a bit extra to your PR but it looks like it's a mistake)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -915,9 +918,9 @@ private void DownloadBlocks(ChainedHeader[] chainedHeaders, OnBlockDownloadedCal
             this.logger.LogTrace("(-)");
         }
 
-        private void BlockDownloaded(uint256 blockHash, Block block)
+        private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
         {
-            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);
+            this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(blockHash), blockHash, nameof(peerId), peerId);
```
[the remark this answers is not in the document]
zeptin: ?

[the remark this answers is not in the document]
noescape00: Change line to 

```
this.logger.LogTrace("({0}:'{1}',{2}:{3})", nameof(blockHash), blockHash, nameof(peerId), peerId);
```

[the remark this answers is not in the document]
noescape00: '{3}' - for strings
{3} - for integers.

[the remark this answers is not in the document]
zeptin: Ah yes, fixed

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -572,6 +571,10 @@ private async Task<ConnectBlocksResult> FullyValidateLockedAsync(ChainedHeader n
 
             ConnectBlocksResult reconnectionResult = await this.ReconnectOldChainAsync(fork, blocksToReconnect).ConfigureAwait(false);
 
+            // Include peers that needed to be banned as a result of a failure to connect blocks.
+            // Otherwise they get lost as we are returning a different ConnnectBlocksResult.
+            reconnectionResult.PeersToBan.AddRange(connectBlockResult.PeersToBan);
```
noescape00:
```
reconnectionResult.PeersToBan = connectBlockResult.PeersToBan;
```

no need to add range- we are not supposed to ban anyone during reconnection

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
Aprogiena: Okay, if it saves allocation then we should do it

noescape00:
Looks good yo me. 

@Aprogiena passing to you

Aprogiena:
@noescape00 I think it is good, check out the latest commit and we can go. But please run the tests also before you merge

noescape00:
Ah, yeah, almost forgot: in BlockMiner we are still calling 
```
this.chainedHeaderTree.PartialOrFullValidationFailed(chainedHeader);
```
It was already called by that time. @zeptin can you please remove that?

Aprogiena:
@noescape00 https://github.com/stratisproject/StratisBitcoinFullNode/pull/2052/files#diff-a846346684dd7e0733412fa86f448f63L259

noescape00:
Some tests won't build.

zeptin:
Were they passing previously?

noescape00:
Not all of them. We currently have around 25 tests failing on activation branch. But they were always building. Current branch fails to build.

noescape00:
Conflicts

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: Why not instead pass `ChainedHeaderBlock` everywhere?

fassadlr:
@zeptin leading on from Dan's comment, can you please also revert the Assert.Throws in the peer banning tests?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
Aprogiena: I wouldn't mind either, but I have unsupported feeling that it would lead to more allocations

noescape00:
@zeptin discard my comments about tests not building. 
Looks like it's not your problem, activation branch failed to build too. 
Fixed already so please merge latest stuff in.

fassadlr:
@zeptin thanks, I'm happy with the changes now 👍

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: Ok lets keep it like this for now its fine

PR #2060, opened 2018-08-29
[Activation] coinview fix

noescape00:


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -371,8 +371,14 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
                         // We'll need to restore the original outputs, so we clone it
                         // and save it in rewind data.
                         UnspentOutputs clone = unspent.Clone();
-                        clone.Outputs = cacheItem.UnspentOutputs?.Outputs.ToArray() ?? Array.Empty<TxOut>();
-                        rewindData.OutputsToRestore.Add(clone);
+
+                        if (cacheItem.UnspentOutputs != null)
```
fassadlr: @noescape00 we can now combine this `if` with the one below as both of them does:

`if (cacheItem.UnspentOutputs != null)`

Aprogiena: We could, but I would just try not to touch things that shouldn't be touched

dangershony: Hang on I have some more changes coming.

fassadlr: @Aprogiena I really don't like this way of coding where we aren't allowed to make a simple improvement like combining an `if` statement. It is a small enough change that improves the code and if we are in this area we should fix it. I mean you and I went over it for 2 hours this morning, surely it won't be an issue reviewing it 😉. We are adding sub-optimal code to the codebase and upping the amount of tech-debt we have.

Its either fix it now, or after this PR has been committed (and the tests are shown to be working) I will submit another one to clean that up.

Also, same for me, I'm just busy fixing the `CanRewind` test, so wait before merging 👍

fassadlr:
@Aprogiena I will add now with my commit 👍

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: No it is not a small change, not in this consensus critical code, we already messed up. I totally dislike all those PRs that instead of just fixing the issue also include 10 times more code just refactor something. You can submit PR later, just don't include it into this one. This one is focused on fixing the problems.

PR #2062, opened 2018-08-29
Fix rewind tests

dangershony:


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -42,8 +42,11 @@ public static bool AreNodesSynced(CoreNode node1, CoreNode node2)
             if (node1.FullNode.GetBlockStoreTip().HashBlock != node2.FullNode.GetBlockStoreTip().HashBlock)
                 return false;
 
-            if (node1.FullNode.MempoolManager().InfoAll().Count != node2.FullNode.MempoolManager().InfoAll().Count)
-                return false;
+            if(!ignoreMempool)
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -49,6 +49,12 @@ private async Task AddBackToMempoolAsync(Block block)
 
                 foreach (Transaction transaction in block.Transactions)
                 {
+                    if(transaction.IsCoinBase)
```
Aprogiena: Spacing

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -49,6 +49,12 @@ private async Task AddBackToMempoolAsync(Block block)
...
+                        continue;
+
+                    if(transaction.IsCoinStake)
```
Aprogiena: Spacing

Aprogiena: You don't want to have this in general case, consider using isprotocoltransaction equivalent

src/Stratis.Bitcoin.Features.MemoryPool/BlocksDisconnectedSignaled.cs
```diff
@@ -49,6 +49,9 @@ private async Task AddBackToMempoolAsync(Block block)
 
                 foreach (Transaction transaction in block.Transactions)
                 {
+                    if(transaction.IsProtocolTransaction())
```
Aprogiena: Spacing

src/NBitcoin/BlockStake.cs
```diff
@@ -173,6 +173,11 @@ public PosTransaction(byte[] bytes) : this()
         {
             this.FromBytes(bytes);
         }
+
+        public override bool IsProtocolTransaction()
+        {
+            return this.IsCoinStake || this.IsCoinBase;
```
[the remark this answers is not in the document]
dangershony: I did consider that but its more readable as it is now

>implies we know exactly what is there  
But we do coinbase is integral part of POS trx.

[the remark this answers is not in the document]
noescape00: It's not integral part imho. 
Only reason why we have it are the leftovers from btc. 

If we change node to be PoS only (no PoW and mixed stage) or improve it in a way that after last PoW block coinbase is not created we won't need to have coinbase tx at all. Currently it's empty already and just occupies first place in the block.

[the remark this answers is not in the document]
dangershony: I agree actually on that comment, my meaning is that at this point it is part of the protocol.
I dont mind changing it later on your free time now I will commit.

PR #2064, opened 2018-08-30
Initial block weight depends on Network.

codingupastorm:
Setting the initial block weight to 4000 in the BlockDefinition works for the Bitcoin network, but it does not make sense for any networks without weighting (i.e. WitnessScaleFactor=1) such as Stratis. 

The result without this change would just be that blocks mined via the C# node have a maximum of BlockSize - 3KB (4000 - 1000).

This change also makes it much easier for us to write certain test cases in SC :)

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
+++ b/src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
@@ -170,7 +170,7 @@ namespace Stratis.Bitcoin.Features.Miner
             this.BlockSize = 1000;
             this.BlockTemplate = new BlockTemplate(this.Network);
             this.BlockTx = 0;
-            this.BlockWeight = 4000;
+            this.BlockWeight = 1000 * this.Network.Consensus.Options.WitnessScaleFactor;
             this.BlockSigOpsCost = 400;
             this.fees = 0;
             this.inBlock = new TxMempool.SetEntries();


noescape00:
I'm on it. Will review tmr.

codingupastorm:
image

All integration tests passing except for `A_cut_off_miner...`

codingupastorm:
Actually, why do these values start at 1000 / 4000 ? 

A standard empty block is somewhere around 250 bytes, so why start at 1000? Is it just to give room to move?

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
```diff
@@ -170,7 +170,7 @@ private void Configure()
             this.BlockSize = 1000;
             this.BlockTemplate = new BlockTemplate(this.Network);
             this.BlockTx = 0;
-            this.BlockWeight = 4000;
+            this.BlockWeight = 1000 * this.Network.Consensus.Options.WitnessScaleFactor;
```
dangershony: This is not a consensus rule so, I guess this is ok (for now)

dangershony:
@codingupastorm this is a network specific argument however its not a consensus rules so perhaps this is something that can be set on the miner setting?

Do we consider putting this on the Network itself (not on any of the consensus classes)?.

Feels like we need to have a generally Default setting of the `Settings` file that is created per network.

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
noescape00: The only bad consequence that could be here is that we will have less space in a block for txes than we should or we fit too much and the block is not valid. 

As far as I understand initializing BlockWeight with this value is needed just to reserve some room for protocol txes. As long as protocol txes are lighter than 1000 bytes we are cool.

codingupastorm: Yep @noescape00 nailed it!

Aprogiena:
Makes sense

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
dangershony: Ah some witness information is kept in coinbase thats why the increase to 4k?

Aprogiena: Actually this should not be increased at all. If we are using block weight, it should always be 4000 here regardless of the network

Aprogiena: I just checked that we are using these in quite messy way. We are using block weight with scaling factor of one which does not make sense

dangershony: But I think the SC guys have very big transactions so the weight limits the number of trx they can push in one block.

dangershony: Perhaps it will be more correct to rename `WitnessScaleFactor`  to `BlockScaleFactor`?

dangershony:
>Why? It's block template constructor specific.

But its also network specific.

Issue #585, opened 2017-10-13
NStratis - extend the method FindAncestorOrSelf


dangershony:
This method will go all the way to genesis and might be useless in some cases when (when a reorg is checked and checkpoints are in place, or reorg is not allowed beyond a certain size)  

As part of this task we should bring the checkpoints in to the Network class.  
The method can take an additional overload bottom height `FindAncestorOrSelf(hash, bottomHeight)` and stop at that height.

Once its done this can be applied to the WalletSyncManager (and maybe also ChainHeaderBehaviour)

Aprogiena:
I've created #659 to aim at the core problem in this, which is inefficient implementation of basic chain functions. Fixing that would make this a non-issue (speaking solely about extension of FAOS).

dangershony:
#659

dangershony:
I think this can be closed right?

Aprogiena:
@mikedennis ?

mikedennis:
Well FindAncestorOrSelf is now fast (uses skiplist) but will still go to genesis if the header isn't found. So not sure if we still want to add an bottomHeight parameter or not. How would we know what to set it to?

Aprogiena:
the specific use case is for stratis network that has a maximal reorg length defined
therefore in some cases, it does not make sense to try to look any further than that
the perf gain from that, however, would be negligible in any foreseeable future
OTOH, it is very trivial to add
so :-|

noescape00:
what is the status of this issue? Wasn't it done on master already?

zeptin:
Just to put a clarifying note here for future reference - there are two versions of the function, and the skip list version is not the one being referred to in this issue.

It may be worth seeing if the few places the non-skiplist version is used can be changed to the other version, with the corresponding performance benefit if the height parameter is not being used to limit the search.

noescape00:
Fixed with activation merge.

noescape00:
@dangershony I think it can be closed

PR #2068, opened 2018-08-30
NStratis - extend the method FindAncestorOrSelf #585

herbepau:
- Optional height param added to allow checking of block height in the search

### DIFF
--- a/src/NBitcoin/ChainedHeader.cs
+++ b/src/NBitcoin/ChainedHeader.cs
@@ -289,12 +289,16 @@ namespace NBitcoin
         /// Finds the ancestor of this entry in the chain that matches the block hash given.
         /// </summary>
         /// <param name="blockHash">The block hash to search for.</param>
+        /// <param name="height">Optional height to search for.</param>
         /// <returns>The ancestor of this chain that matches the block hash.</returns>
-        public ChainedHeader FindAncestorOrSelf(uint256 blockHash)
+        public ChainedHeader FindAncestorOrSelf(uint256 blockHash, int? height = null)
         {
             ChainedHeader currentBlock = this;
-            while ((currentBlock != null) && (currentBlock.HashBlock != blockHash))
+            while (currentBlock != null)
             {
+                if (currentBlock.HashBlock == blockHash || (height != null && currentBlock.Height == height))
+                    break;
+
                 currentBlock = currentBlock.Previous;
             }
 


dangershony:
https://github.com/stratisproject/StratisBitcoinFullNode/issues/585

dangershony:
@majikandy indeed correct I was thinking that, what we need is to find the point in CM (or CHT) that finds a previous block and if max reorg is enabled pass that number.

@herbepau please add that code as well

src/NBitcoin/ChainedHeader.cs
```diff
@@ -286,15 +286,19 @@ public ChainedHeader FindAncestorOrSelf(ChainedHeader chainedHeader)
...
+        /// <param name="height">Optional height to search for.</param>
         /// <returns>The ancestor of this chain that matches the block hash.</returns>
-        public ChainedHeader FindAncestorOrSelf(uint256 blockHash)
+        public ChainedHeader FindAncestorOrSelf(uint256 blockHash, int height = -1)
         {
             ChainedHeader currentBlock = this;
-            while ((currentBlock != null) && (currentBlock.HashBlock != blockHash))
+            while (currentBlock != null)
             {
+                if (currentBlock.HashBlock == blockHash || currentBlock.Height == height)
```
Aprogiena: Brackets please

src/NBitcoin/ChainedHeader.cs
```diff
@@ -286,15 +286,19 @@ public ChainedHeader FindAncestorOrSelf(ChainedHeader chainedHeader)
...
 
         /// <summary>
-        /// Finds the ancestor of this entry in the chain that matches the block hash given.
+        /// Finds the ancestor of this entry in the chain that matches the block hash, or optionally, the block height given.
         /// </summary>
         /// <param name="blockHash">The block hash to search for.</param>
+        /// <param name="height">Optional height to search for.</param>
         /// <returns>The ancestor of this chain that matches the block hash.</returns>
-        public ChainedHeader FindAncestorOrSelf(uint256 blockHash)
+        public ChainedHeader FindAncestorOrSelf(uint256 blockHash, int height = -1)
```
Aprogiena: The default value of zero would be even better

dangershony: Then you have to do the extra check in the while loop, its fine like I recon.

Aprogiena: Why would you need any extra check?

dangershony: Because zero is also the genesis block.

Aprogiena: Yes, so why would you need any extra checks?

Aprogiena: Okay, it should return null, you are right. Returns comment should be updated to mention what should happen when the element is not found.

dangershony: Ah should we return null in case the max limit was reached?

Aprogiena: LOL, that is a correct question and answer is probably yes. We should totally change the comment to know what we are doing

src/NBitcoin/ChainedHeader.cs
```diff
@@ -286,15 +286,19 @@ public ChainedHeader FindAncestorOrSelf(ChainedHeader chainedHeader)
...
         {
             ChainedHeader currentBlock = this;
-            while ((currentBlock != null) && (currentBlock.HashBlock != blockHash))
+            while (currentBlock != null)
             {
+                if ((currentBlock.HashBlock == blockHash) || (currentBlock.Height == height))
```
noescape00: why not make it part of the while condition?

dangershony: Its easier to read like this

noescape00: but has more overhead. 

I guess if formatted correctly, having this in while statememnt will be readable

dangershony: I don't think there is more overhead.
I generally prefer readability because the CLI will anyway compile and optimize the code

zeptin:
I have attempted to make the logic match what we want from this, i.e. stop the search at a given height, not look for a header with a given height.

Andy's comment about actually using the functionality somewhere is valid, I will see if the two usages of this function in the code can benefit.

zeptin:
So the two usages are in `ConsensusManager.InitializeAsync` and `ChainHeadersBehavior.ProcessHeadersAsync`.

`InitializeAsync`: There is probably minimal improvement to be gained by introducing the parameter here, but I have done so. Is it possible for the node to start up in a state where it will need to rewind more than the max reorg to find a common header? If it is, then I will have to revert this and the new parameter is essentially useless in the codebase as it stands.

`ProcessHeadersAsync`: I don't see a sensible setting for the height parameter here, as peers may (and often will) request headers right back to genesis. Not added.

src/NBitcoin/ChainedHeader.cs
Aprogiena: Okay, so now, as we have the new implementation, let me again ask my original question – would it not be better if the default value was zero on the index?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -143,9 +143,12 @@ public async Task InitializeAsync(ChainedHeader chainTip)
 
             uint256 consensusTipHash = await this.consensusRules.GetBlockHashAsync().ConfigureAwait(false);
 
+            // If we are on a network with a maximum reorg setting, don't bother searching further back than the limit.
```
Aprogiena: I'm not sure this is correct. The initialisation should be allowed to go beyond finalised the block. Also note that all the code should work with finalised block instead of maximum reorganisation directly. But such limitation should not be here

dangershony: @zeptin I prefer if in this PR we only add the code and not activate it, less problems to worry about if something is broken so please revert change to CM

zeptin: Yes, I think that is better. I will do that shortly.

zeptin: Done

src/NBitcoin/ChainedHeader.cs
```diff
@@ -286,19 +286,24 @@ public ChainedHeader FindAncestorOrSelf(ChainedHeader chainedHeader)
...
-        /// Finds the ancestor of this entry in the chain that matches the block hash given.
+        /// Finds the ancestor of this entry in the chain that matches the block hash.
+        /// It will not search lower than the optional height parameter.
         /// </summary>
         /// <param name="blockHash">The block hash to search for.</param>
-        /// <returns>The ancestor of this chain that matches the block hash.</returns>
-        public ChainedHeader FindAncestorOrSelf(uint256 blockHash)
+        /// <param name="height">Optional height to restrict the search to.</param>
+        /// <returns>The ancestor of this chain that matches the block hash, or null if it was not found.</returns>
+        public ChainedHeader FindAncestorOrSelf(uint256 blockHash, int height = -1)
```
Aprogiena: So now as we have the new implementation, I raise my question again, would it not be better to have default value zero?

dangershony:
Ok I am happy with this change now, but this will go in after we merge activation to master

bokobza:
@dangershony is this ok to go in now?

src/NBitcoin/ChainedHeader.cs
```diff
@@ -286,19 +286,24 @@ public ChainedHeader FindAncestorOrSelf(ChainedHeader chainedHeader)
...
-        public ChainedHeader FindAncestorOrSelf(uint256 blockHash)
+        /// <param name="height">Optional height to restrict the search to.</param>
+        /// <returns>The ancestor of this chain that matches the block hash, or null if it was not found.</returns>
+        public ChainedHeader FindAncestorOrSelf(uint256 blockHash, int height = 0)
         {
             ChainedHeader currentBlock = this;
-            while ((currentBlock != null) && (currentBlock.HashBlock != blockHash))
+            while ((currentBlock != null) && (currentBlock.Height > height))
             {
+                if (currentBlock.HashBlock == blockHash)
```
dangershony: Supposedly you can save an additional `if` statement  if you replace the condition `currentBlock.HashBlock == blockHash` with `currentBlock.Height > height`

And in the if do this
 ```
if(currentBlock.Height > height)
{
    currentBlock = null;
    break;
}

then the return is just `currentBlock`

```

But I am happy with this as is as well.

PR #2069, opened 2018-08-30
Fix mempool tests

dangershony:


src/Stratis.Bitcoin.Features.Consensus/ConsensusQuery.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
```
fassadlr: Are all these `usings` used?

src/Stratis.Bitcoin.Features.Consensus/ConsensusQuery.cs
```diff
@@ -0,0 +1,49 @@
...
+        public async Task<UnspentOutputs> GetUnspentTransactionAsync(uint256 trxid)
+        {
+            CoinViews.FetchCoinsResponse response = null;
+
+            if (this.coinView != null)
+            {
+                response = await this.coinView.FetchCoinsAsync(new[] { trxid }).ConfigureAwait(false);
+            }
+
+            return response?.UnspentOutputs?.SingleOrDefault();
```
Aprogiena: The response is never null

Aprogiena: And the unspent outputs inside of the response is also never null

Aprogiena: And also, you don't need single or default, you can just take the first element

dangershony: yeah correct was a copy past from the version in master that was used before, I will fix.

src/Stratis.Bitcoin.Features.Consensus/ConsensusQuery.cs
dangershony: fixed

src/Stratis.Bitcoin.Features.MemoryPool/MempoolFeature.cs
```diff
@@ -171,9 +171,9 @@ public static IFullNodeBuilder UseMempool(this IFullNodeBuilder fullNodeBuilder)
                         services.AddSingleton<BlockPolicyEstimator>();
                         services.AddSingleton<IMempoolValidator, MempoolValidator>();
                         services.AddSingleton<MempoolOrphans>();
-                        services.AddSingleton<MempoolManager>();
-                        services.AddSingleton<IPooledTransaction, MempoolManager>();
-                        services.AddSingleton<IPooledGetUnspentTransaction, MempoolManager>();
+                        services.AddSingleton<MempoolManager>()
+                            .AddSingleton<IPooledTransaction, MempoolManager>(provider => provider.GetService<MempoolManager>())
+                            .AddSingleton<IPooledGetUnspentTransaction, MempoolManager>(provider => provider.GetService<MempoolManager>());
```
majikandy: Is it correct that these ones are chained on to the previous addsingleton and the others here are on their own line of code?

src/Stratis.Bitcoin.Features.Consensus/ConsensusQuery.cs
```diff
@@ -0,0 +1,40 @@
...
+        public ConsensusQuery(
+            ICoinView coinView,
+            ILoggerFactory loggerFactory)
+        {
+            this.coinView = coinView;
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+        }
+
+        /// <inheritdoc />
+        public async Task<UnspentOutputs> GetUnspentTransactionAsync(uint256 trxid)
```
majikandy: Parameter could be named trxId or transactionId

PR #2072, opened 2018-08-30
[Activation] introduced block mined bool

noescape00:
Apparently we never set `bool BlockMined` in the context. That was the reason why at least 10 tests were failing. 

Introduced it. 
However it's not propagated from high level to low level code so I don't really like it.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
+++ b/src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
@@ -19,8 +19,9 @@ namespace Stratis.Bitcoin.Consensus.Validators
         /// Validates a block header.
         /// </summary>
         /// <param name="chainedHeader">The chained header to be validated.</param>
+        /// <param name="blockMined">Whether the block was mined or obtained from a peer.</param>
         /// <returns>Context that contains validation result related information.</returns>
-        ValidationContext ValidateHeader(ChainedHeader chainedHeader);
+        ValidationContext ValidateHeader(ChainedHeader chainedHeader, bool blockMined);
     }
 
     public interface IPartialValidator : IDisposable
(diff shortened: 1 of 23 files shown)


src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
```diff
@@ -19,8 +19,9 @@ public interface IHeaderValidator
         /// Validates a block header.
         /// </summary>
         /// <param name="chainedHeader">The chained header to be validated.</param>
+        /// <param name="blockMined">Whether the block was mined or obtained from a peer.</param>
         /// <returns>Context that contains validation result related information.</returns>
-        ValidationContext ValidateHeader(ChainedHeader chainedHeader);
+        ValidationContext ValidateHeader(ChainedHeader chainedHeader, bool blockMined);
```
Aprogiena: Do we need this?

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
```diff
@@ -59,8 +62,9 @@ public interface IIntegrityValidator
         /// </remarks>
         /// <param name="header">The chained header that is going to be validated.</param>
         /// <param name="block">The block that is going to be validated.</param>
+        /// <param name="blockMined">Whether the block was mined or obtained from a peer.</param>
         /// <returns>Context that contains validation result related information.</returns>
-        ValidationContext VerifyBlockIntegrity(ChainedHeader header, Block block);
+        ValidationContext VerifyBlockIntegrity(ChainedHeader header, Block block, bool blockMined);
```
Aprogiena: Do we need this?

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
dangershony: Can we not solve this in a different way?
this is so ugly.

noescape00: I'm open for suggestions. 

Basically we need this thing to be populated for all types of validation so the only way I found without having a refactoring is to provide this bool in all validators. That leaded to propagation of this bool from high level code to the very bottom.  

Another solution would be to remove `BlockMined` from context and all the rules should treat the stuff being validated as something downloaded from the peer. I'd actually go with that but it's a bit big refactoring.

src/Stratis.Bitcoin/Consensus/Validators/BlockValidator.cs
noescape00: yeah. there is one integration rule that uses context.BlockMined

noescape00:
will submit a pr that removes BlockMined from context

PR #2075, opened 2018-08-31
Virtuals for testing

codingupastorm:
Hey guys, the retrieval of concrete classes in the BlockDefinition isn't really conducive to testing, but we can get around it if we mark the given methods virtual. 

Hopefully this is okay!


codingupastorm:
@majikandy this is to separate FN changes from SC, but alas I think I won't be needing this at the moment anyhow :(

PR #2076, opened 2018-08-31
[SC] Block buffer for generated transactions

codingupastorm:


src/Stratis.Bitcoin.Features.SmartContracts.Tests/BlockBufferGeneratorTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Tests
+{
+    public class BlockBufferGeneratorTests
+    {
+        private readonly BlockBufferGenerator bufferGen;
+
+        public BlockBufferGeneratorTests()
+        {
+            this.bufferGen = new BlockBufferGenerator();
```
majikandy: Use the full words.... bufferGenerator

src/Stratis.Bitcoin.Features.SmartContracts/BlockBufferGenerator.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Stratis.Bitcoin.Features.Miner;
+
+namespace Stratis.Bitcoin.Features.SmartContracts
+{
+    public class BlockBufferGenerator : IBlockBufferGenerator
+    {
+        public BlockDefinitionOptions GetOptionsWithBuffer(BlockDefinitionOptions options)
+        {
+            uint percentageBuffer = 5; // For 1MB blocks, 50 KB reserved for generated transactions / txouts
```
majikandy: Variable name could represent the comment?

src/Stratis.Bitcoin.Features.SmartContracts/SmartContractPosPowBlockDefinition.cs
```diff
@@ -67,6 +68,13 @@ public sealed class SmartContractPosPowBlockDefinition : BlockDefinition
             this.stakeChain = stakeChain;
             this.stakeValidator = stakeValidator;
             this.stateRoot = stateRoot;
+
+            // When building smart contract blocks, we will be generating and adding both transactions to the block and txouts to the coinbase. 
+            // At the moment, these generated objects aren't accounted for in the block size and weight accounting. 
+            // This means that if blocks started getting full, this miner could start generating blocks greater than the max consensus block size.
+            // To avoid this without significantly overhauling the BlockDefinition, for now we just lower the block size by a percentage buffer.
+            // If in the future blocks are being built over the size limit and you need an easy fix, just increase the size of this buffer.
```
majikandy: This comment is a repeat of the one in the other file. Is there any way to avoid the duplication ?

codingupastorm: Sorry Andy, but I think this is more a reflection of the issues with the modularity of the BlockDefinition. Pains me but not going to move it for now. We also don't really know which BlockDefinition we'll end up using.

src/Stratis.Bitcoin.Features.SmartContracts/SmartContractBlockDefinition.cs
```diff
@@ -52,6 +53,13 @@ public sealed class SmartContractBlockDefinition : BlockDefinition
             this.stateRoot = stateRoot;
             this.refundOutputs = new List<TxOut>();
             this.receipts = new List<Receipt>();
+
+            // When building smart contract blocks, we will be generating and adding both transactions to the block and txouts to the coinbase. 
+            // At the moment, these generated objects aren't accounted for in the block size and weight accounting. 
+            // This means that if blocks started getting full, this miner could start generating blocks greater than the max consensus block size.
+            // To avoid this without significantly overhauling the BlockDefinition, for now we just lower the block size by a percentage buffer.
+            // If in the future blocks are being built over the size limit and you need an easy fix, just increase the size of this buffer.
+            this.Options = blockBufferGenerator.GetOptionsWithBuffer(this.Options);
```
rowandh: I'm not sure of the implications of this, but in the `BlockDefinition` constructor it uses the options to set `NeedSizeAccounting`, which is based on the original value of `BlockMaxSize`.

codingupastorm: For weight-based block size consensus, you may want to track size as well as weight, if the user adds a setting via the command line to have a max block size mined by this miner. Hence `NeedSizeAccounting` is useful to check if a block has ever gone over a certain size in bytes when building it up in weight.

For size-based block size consensus, checking weight and checking size are actually the same calculation - whether `NeedSizeAccounting` is set or not, the result will be identical. We don't actually want to touch that setting in the code I've changed above.

This is because the node is trying to do 2 things at once (Account for weight and size) in the same code. Quite problematic but I will fix this in the FN at some point.

PR #2077, opened 2018-08-31
[SC] Introduce State object

rowandh:
Introduces a 'State' object, which represents the state of the world during an entire contract operation.

The initial state is set up in the executor after receipt of contract data from the outside. External messages invoke create/calls on a contract, which themselves generate internal messages. State handles each of these scenarios, invoking the VM, metering gas, and updating logs and balances as appropriate.


codingupastorm:
Going to take a while for me to understand, but so far:

- Why do we `startTracking()` in the internalexecutor only?

- Imagine we make a call from `SmartContractsController` (e.g. ContractBalance) while a block is going through the CoinViewRule. We could get results that shouldn't be there if we query at the wrong time, because we're using the top-level staterepositoryroot. That top-level root should remain untouched - maybe it needs its own interface to only query information, and get a `MutableRepository` like the ones we already have, which we can happily pass wherever we want.

- StateSnapshot, State, ContractState, ContractStateRoot, AccountState. We have to do something here, it's quite difficult to follow

- The vars in State.cs make it hard to follow - could you apply the 1-click VS fix? Thanks

codingupastorm:
Woops, it's ready to go in, just the ExecutorSpecification test needs fixing

PR #2079, opened 2018-08-31
[Activation] Fix TxMempoolBlockDoublespend

dangershony:


src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -386,10 +386,10 @@ private async Task<ConnectNewHeadersResult> PresentHeadersLockedAsync(List<Block
                 // Resync in case can't connect.
                 await this.ResyncAsync().ConfigureAwait(false);
             }
-            catch (ConsensusErrorException exception)
+            catch (HeaderInvalidException)
```
majikandy: I find the mismatch between ConsensusErrors.BannedHash.Throw and HeaderInvalidException awkward to follow. I assume the .Throw throws that exception but I would say that it isn’t obvious.

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -386,10 +386,10 @@ private async Task<ConnectNewHeadersResult> PresentHeadersLockedAsync(List<Block
...
             {
-                this.logger.LogDebug("Peer's header is invalid. Peer will be banned and disconnected. Error: {0}.", exception.ConsensusError);
-                this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, $"Peer presented invalid header, error: {exception.ConsensusError}.");
+                this.logger.LogDebug("Peer's header is invalid. Peer will be banned and disconnected.");
+                this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, $"Peer presented invalid header.");
```
majikandy: Magic strings, potentially should be constants

majikandy: And no need for the $ string interpolation

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
dangershony: They sort of have the same meaning so `HeaderInvalidException` is more general.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1097,7 +1097,7 @@ private ChainedHeader CreateAndValidateNewChainedHeader(BlockHeader currentBlock
             if (result.Error != null)
             {
                 this.logger.LogTrace("(-)[INVALID_HEADER]");
-                result.Error.Throw();
+                throw new ConsensusRuleException(result.Error);
```
fassadlr: Shouldn't this be `HeaderInvalidException` as we validated the header above?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1087,7 +1087,7 @@ private ChainedHeader CreateAndValidateNewChainedHeader(BlockHeader currentBlock
             if (this.invalidHashesStore.IsInvalid(newHeaderHash))
             {
                 this.logger.LogTrace("(-)[HEADER_HASH_MARKED_INVALID]");
-                ConsensusErrors.BannedHash.Throw();
+                throw new HeaderInvalidException();
```
fassadlr: Perhaps pass the "HEADER_HASH_MARKED_INVALID" message to the exception?

src/Stratis.Bitcoin/Consensus/ConsensusExceptions.cs
```diff
@@ -28,6 +28,23 @@ public ConnectHeaderException() : base()
         }
     }
 
+    public class HeaderInvalidException : ConsensusException
```
fassadlr: Document these new exceptions.

src/Stratis.Bitcoin/Consensus/ConsensusExceptions.cs
```diff
@@ -28,6 +28,23 @@ public ConnectHeaderException() : base()
...
+    {
+        public HeaderInvalidException() : base()
+        {
+        }
+    }
+
+    public class ConsensusRuleException : ConsensusException
```
fassadlr: Document these new exceptions.

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -386,11 +386,16 @@ private async Task<ConnectNewHeadersResult> PresentHeadersLockedAsync(List<Block
                 // Resync in case can't connect.
                 await this.ResyncAsync().ConfigureAwait(false);
             }
-            catch (ConsensusErrorException exception)
+            catch (ConsensusRuleException exception)
             {
                 this.logger.LogDebug("Peer's header is invalid. Peer will be banned and disconnected. Error: {0}.", exception.ConsensusError);
                 this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, $"Peer presented invalid header, error: {exception.ConsensusError}.");
             }
+            catch (HeaderInvalidException)
```
fassadlr: If we are going to pass the specific message to the exception we might want to write this to the log as well?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: No this is a specific exception.

noescape00:
looks good. 

however I'm against spawning more and more different types of consensus errors.

src/Stratis.Bitcoin/Consensus/ConsensusExceptions.cs
```diff
@@ -28,6 +28,30 @@ public ConnectHeaderException() : base()
...
+    {
+        public HeaderInvalidException() : base()
+        {
+        }
+    }
+
+    /// <summary>
+    /// An exception that is contains exception coming from the <see cref="IConsensusRuleEngine"/> execution.
+    /// </summary>
+    public class ConsensusRuleException : ConsensusException
```
noescape00: this is basically a copy-paste of ConsensusErrorException

dangershony: LOL yes it is even the name is wrong they should be swapped but didnt want to make such a big change yet.

PR #2096, opened 2018-09-02
Disconnect peers that send large locators

### LINKED ISSUE #2094: Introduce a maximum size for locators
https://github.com/bitcoin/bitcoin/pull/13907

zeptin:
Fixes #2094.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -41,6 +41,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         private ConsensusManagerBehavior consensusManagerBehavior;
 
+        private readonly IPeerBanning peerBanning;
+
         /// <summary>Instance logger.</summary>
         private readonly ILogger logger;
 
@@ -75,7 +77,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <inheritdoc cref="IChainState"/>
         private readonly IChainState chainState;
 
-        public BlockStoreBehavior(ConcurrentChain chain, IBlockStore blockStore, IChainState chainState, ILoggerFactory loggerFactory)
+        public BlockStoreBehavior(ConcurrentChain chain, IBlockStore blockStore, IChainState chainState, ILoggerFactory loggerFactory, IPeerBanning peerBanning)
         {
             Guard.NotNull(chain, nameof(chain));
             Guard.NotNull(blockStore, nameof(blockStore));
@@ -86,6 +88,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             this.chainState = chainState;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.loggerFactory = loggerFactory;
+            this.peerBanning = peerBanning;
 
             this.CanRespondToGetBlocksPayload = true;
             this.CanRespondToGetDataPayload = true;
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -186,6 +189,16 @@ private async Task ProcessGetBlocksAsync(INetworkPeer peer, GetBlocksPayload get
         {
             this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(peer), peer.RemoteSocketEndpoint, nameof(getBlocksPayload), getBlocksPayload);
 
+            if (getBlocksPayload.BlockLocators.Blocks.Count > ConsensusManagerBehavior.MaxLocatorSize)
+            {
+                this.logger.LogTrace("Peer '{0}' sent getblocks with oversized locator, disconnecting.", peer.RemoteSocketEndpoint);
+
+                this.peerBanning.BanAndDisconnectPeer(peer.RemoteSocketEndpoint, 0);
```
Aprogiena: Why zero?

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -64,6 +64,11 @@ public class ConsensusManagerBehavior : NetworkPeerBehavior
         /// <seealso cref="https://en.bitcoin.it/wiki/Protocol_documentation#getheaders"/>
         private const int MaxItemsPerHeadersMessage = 2000;
 
+        /// <summary>Maximum number of hashes that we will accept in a message from a peer.</summary>
+        /// <seealso cref="https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-August/016285.html"/>
+        /// <seealso cref="https://github.com/bitcoin/bitcoin/pull/13907"/>
+        public const int MaxLocatorSize = 101;
+
```
Aprogiena: I'm not sure if this is the right file to have this. I'd probably inside of the block locator class. There are other possibilities, but the specific behaviour does not look like a good place to have it

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -179,6 +184,16 @@ private async Task ProcessGetHeadersAsync(INetworkPeer peer, GetHeadersPayload g
                 return;
             }
 
+            if (getHeadersPayload.BlockLocator.Blocks.Count > MaxLocatorSize)
+            {
+                this.logger.LogTrace("Peer '{0}' sent getheader with oversized locator, disconnecting.", peer.RemoteSocketEndpoint);
+
+                this.peerBanning.BanAndDisconnectPeer(peer.RemoteSocketEndpoint, 0);
```
Aprogiena: Why zero?

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
zeptin: If it was acceptable for `MaxItemsPerHeadersMessage` to be in this file then it seems to me that, at least, we are no worse off including `MaxLocatorSize`.

Aprogiena: It is because the difference between those two is that one is private and the other one is used elsewhere.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
Aprogiena: If you only want to disconnect, just disconnect the peer, don't introduce peer banning dependency

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
dangershony: Block locator class seems more correct indeed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -186,6 +186,16 @@ private async Task ProcessGetBlocksAsync(INetworkPeer peer, GetBlocksPayload get
         {
             this.logger.LogTrace("({0}:'{1}',{2}:'{3}')", nameof(peer), peer.RemoteSocketEndpoint, nameof(getBlocksPayload), getBlocksPayload);
 
+            if (getBlocksPayload.BlockLocators.Blocks.Count > BlockLocator.MaxLocatorSize)
+            {
+                this.logger.LogTrace("Peer '{0}' sent getblocks with oversized locator, disconnecting.", peer.RemoteSocketEndpoint);
+
+                peer.Disconnect("Peer sent getblocks with oversized locator");
```
dangershony: Ah hang on should we not ban the node?

zeptin: I think core only disconnect from the node:

https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L2013

dangershony: Ah I think its so they dont ban old nodes that may provide locator information differently (BitcoinJ I think) so perhaps we don't care about such cases.
But maybe its so trivial so its not worth adding an entry in the db.

dangershony: Ah I see @Aprogiena commented on this not adding the banning dependency yeah ok lets push this as is (same as core)

PR #2097, opened 2018-09-02
Peer banning overload

### LINKED ISSUE #2088: Peer banning overload that doesn't have banning time parameter
Peer banning is used 95% of the time with default banning time from connection manager. That introduces dependency on connection manager in every component that uses banning. 
We can simplify design of a lot of components a lot if we add an overload to peer banning which won't accept banning time and use default banning time instead.

zeptin:
Fixes  #2088.

Attempts to decrease unnecessary usage of `ConnectionManager` by introducing an overload for the `BanAndDisconnectPeer` method. This overload bans for the default interval set in `ConnectionManagerSettings.BanTimeSeconds` instead of requiring the caller to have `ConnectionManagerSettings` available to retrieve this value.

Also removes some unneeded fields from `ValidationContext` and `ConnectBlocksResult`, as banning should probably be determined outside the consensus rules. These values also did not appear to be getting set to values outside the default.

### DIFF
--- a/src/Stratis.Bitcoin/Connection/PeerBanning.cs
+++ b/src/Stratis.Bitcoin/Connection/PeerBanning.cs
@@ -24,6 +24,14 @@ namespace Stratis.Bitcoin.Connection
         void BanAndDisconnectPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = null);
 
         /// <summary>
+        /// Overload for the BanAndDisconnectPeer method that uses the connection manager's default ban interval.
+        /// This allows features to depend solely on the peer banning interface and not the connection manager directly.
+        /// </summary>
+        /// <param name="endpoint">The endpoint to set that it was banned.</param>
+        /// <param name="reason">An optional reason for the ban, the 'reason' is only use for tracing.</param>
+        void BanAndDisconnectPeerForDefaultInterval(IPEndPoint endpoint, string reason = null);
+
+        /// <summary>
         /// Check if a peer is banned.
         /// </summary>
         /// <param name="endpoint">The endpoint to check if it was banned.</param>
@@ -96,6 +104,12 @@ namespace Stratis.Bitcoin.Connection
             this.logger.LogDebug("Peer '{0}' banned for reason '{1}', until {2}.", endpoint, reason, peerAddress.BanUntil.ToString());
 
             this.logger.LogTrace("(-)");
+        }
+
+        /// <inheritdoc />
+        public void BanAndDisconnectPeerForDefaultInterval(IPEndPoint endpoint, string reason = null)
+        {
+            this.BanAndDisconnectPeer(endpoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, reason);
         }
 
         /// <inheritdoc />
--- a/src/Stratis.Bitcoin/Connection/PeerBanning.cs
+++ b/src/Stratis.Bitcoin/Connection/PeerBanning.cs
@@ -24,12 +24,12 @@ namespace Stratis.Bitcoin.Connection
         void BanAndDisconnectPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = null);
 
         /// <summary>
-        /// Overload for the BanAndDisconnectPeer method that uses the connection manager's default ban interval.
+        /// Bans and disconnects the peer using the connection manager's default ban interval.
         /// This allows features to depend solely on the peer banning interface and not the connection manager directly.
         /// </summary>
         /// <param name="endpoint">The endpoint to set that it was banned.</param>
         /// <param name="reason">An optional reason for the ban, the 'reason' is only use for tracing.</param>
-        void BanAndDisconnectPeerForDefaultInterval(IPEndPoint endpoint, string reason = null);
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin/Connection/PeerBanning.cs
```diff
@@ -23,6 +23,14 @@ public interface IPeerBanning
         /// <param name="reason">An optional reason for the ban, the 'reason' is only use for tracing.</param>
         void BanAndDisconnectPeer(IPEndPoint endpoint, int banTimeSeconds, string reason = null);
 
+        /// <summary>
+        /// Overload for the BanAndDisconnectPeer method that uses the connection manager's default ban interval.
```
Aprogiena: The method has different name, so it doesn't look like overload. And when it is overload, you don't need to have it in a comment,

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -383,7 +383,7 @@ private async Task OnPartialValidationCompletedCallbackAsync(ValidationContext v
                 }
 
                 foreach (INetworkPeer peer in peersToBan)
-                    this.peerBanning.BanAndDisconnectPeer(peer.RemoteSocketEndpoint, validationContext.BanDurationSeconds, $"Invalid block received: {validationContext.Error.Message}");
+                    this.peerBanning.BanAndDisconnectPeerForDefaultInterval(peer.RemoteSocketEndpoint, $"Invalid block received: {validationContext.Error.Message}");
```
Aprogiena: This is wrong, leave the original as we need to be able to set duration on per rule basis

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -436,7 +436,7 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
                     this.logger.LogTrace("{0} peers will be banned.", peersToBan.Count);
 
                     foreach (INetworkPeer peer in peersToBan)
-                        this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, connectBlocksResult.BanDurationSeconds, connectBlocksResult.BanReason);
+                        this.peerBanning.BanAndDisconnectPeerForDefaultInterval(peer.PeerEndPoint, connectBlocksResult.BanReason);
```
Aprogiena: Same here

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Do we actually need banning time on per rule basis? 
The only way we ban is when the rule throws (so the context's error is set), and the rule throws when the block is invalid. It seems reasonable to me that no matter why the block is invalid we would like to ban the peer for the same time. 

But in case we don't- why rules should decide for how long the ban should happen? I think higher level logic should be responsible for that decision, not the rule itself.

Aprogiena: I think this discussion is outside of the scope of this PR. Bitcoin core does that, so possibly it is not a terrible idea. It would allow us to have better control about how strongly we ban peers for which problems.

noescape00: It's not a bad idea, I agree. 
But why rules get to decide the duration? It sounds like a high level logic to me comparing to just checking some block validity aspect. 

Also we might be validating our own blocks in which case ban duration is irrelevant (even though those are sanity checks).

Aprogiena: It might be for a discussion but it does not belong here. This issue and this PR is supposed to address something and is not supposed to do this. So if we want this, let's open an issue for discussion.

noescape00: Ok, makes sense. 
@zeptin can you please apply fixes?

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -389,17 +385,17 @@ private async Task<ConnectNewHeadersResult> PresentHeadersLockedAsync(List<Block
             catch (ConsensusErrorException exception)
             {
                 this.logger.LogDebug("Peer's header is invalid. Peer will be banned and disconnected. Error: {0}.", exception.ConsensusError);
-                this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, this.connectionManager.ConnectionSettings.BanTimeSeconds, $"Peer presented invalid header, error: {exception.ConsensusError}.");
```
dangershony: If the only reason we  needed Connection Manager is for the settings then we could just inject the settings.

src/Stratis.Bitcoin/Connection/PeerBanning.cs
dangershony: Agree I would just keep the name as is.

dangershony:
@zeptin please add a correct overload (not a new name) and revert CM changes, this is not even needed for activation so was not needed to fix right now, keep the changes of the behaviour and lets push it.

PR #2102, opened 2018-09-03
[Activation] Smart contract integration tests PASS

fassadlr:


src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/GenerateCoinsFastDateTimeProvider.cs
```diff
@@ -13,33 +12,36 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
     /// </summary>
     public sealed class GenerateCoinsFastDateTimeProvider : SignalObserver<ChainedHeaderBlock>, IDateTimeProvider
     {
-        private TimeSpan adjustedTimeOffset;
-        private DateTime startFrom;
+        private static TimeSpan adjustedTimeOffset;
```
dangershony: hmm why this needs to be static?

fassadlr: If we have two nodes we need to share the same instance of the `GenerateCoinsFastDateTimeProvider` so that both nodes' date and time is in sync. Otherwise each node gets a new instance of `GenerateCoinsFastDateTimeProvider` which means they track their own date and time.

For example, if we DONT do this and Node B syncs from Node A, its block and transaction's time stamps will start again from 2018-01-01 00:00:00 and not from where Node A's last time stamp left off.

dangershony: I have statics but this time you are right (unless we introduce a more complicated fix)

Issue #2089, opened 2018-08-31
Block puller- improve peer speed estimation


noescape00:
currently block puller peer's estimation is a bit hacky and it should be analyzed and improved. 

one of the approaches: 

```
Block Puller - check out alternative approach to peer speed estimation. 
Measure the delay based on time from the last block delivery and not
from the time that particular block was requested. If that approach is better- use it. fix speed estimation

add +10% margin (ask for 10% more)
```

noescape00:
fixed

PR #2104, opened 2018-09-03
[Activation] block puller peers speed estimation

noescape00:
Fix: #2089

### DIFF
--- a/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
+++ b/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
@@ -22,7 +22,7 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
             var loggerFactory = new ExtendedLoggerFactory();
             loggerFactory.AddConsoleWithFilters();
 
-            this.behavior = new BlockPullerBehavior(puller.Object, ibdState.Object, new DateTimeProvider(), loggerFactory);
+            this.behavior = new BlockPullerBehavior(puller.Object, ibdState.Object, DateTimeProvider.Default, loggerFactory);
         }
 
         [Fact]
--- a/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
+++ b/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
@@ -36,9 +36,9 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
         {
             // Add a lot of bad samples to push quality score down. After that peer will have only bad samples.
             for (int i = 0; i < 10; i++)
-                this.behavior.AddSample(100, 10);
+                this.behavior.AddSample(1, 10);
 
-            this.behavior.RecalculateQualityScore(1000);
+            this.behavior.RecalculateQualityScore(100000);
 
             Assert.Equal(BlockPullerBehavior.MinQualityScore, this.behavior.QualityScore);
         }
--- a/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
+++ b/src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
@@ -36,7 +36,7 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
         {
             // Add a lot of bad samples to push quality score down. After that peer will have only bad samples.
             for (int i = 0; i < 10; i++)
-                this.behavior.AddSample(10);
+                this.behavior.AddSample(100, 10);
 
             this.behavior.RecalculateQualityScore(1000);
 
@@ -46,8 +46,8 @@ namespace Stratis.Bitcoin.Tests.BlockPulling
         [Fact]
         public void QualityScoreCanGoToMaximum()
         {
-            this.behavior.AddSample(1);
-            this.behavior.RecalculateQualityScore(1);
+            this.behavior.AddSample(100, 1);
+            this.behavior.RecalculateQualityScore(100);
 
             Assert.Equal(BlockPullerBehavior.MaxQualityScore, this.behavior.QualityScore);
         }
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -100,31 +121,71 @@ public BlockPullerBehavior(IBlockPuller blockPuller, IInitialBlockDownloadState
...
         /// <inheritdoc/>
-        public void AddSample(double delaySeconds)
+        public void AddSample(long blockSizeBytes, double delaySinceRequestedSeconds)
         {
-            this.logger.LogTrace("({0}:{1})", nameof(delaySeconds), delaySeconds);
+            this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(blockSizeBytes), blockSizeBytes, nameof(delaySinceRequestedSeconds), delaySinceRequestedSeconds);
+
+            double adjustedDelay;
+
+            if (this.lastDeliveryTime == null)
```
Aprogiena: Brackets around please

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -100,31 +121,71 @@ public BlockPullerBehavior(IBlockPuller blockPuller, IInitialBlockDownloadState
...
+                adjustedDelay = delaySinceRequestedSeconds;
+            else
+            {
+                double deliveryDiff = (DateTime.Now - this.lastDeliveryTime).Value.TotalSeconds;
+
+                adjustedDelay = Math.Min(delaySinceRequestedSeconds, deliveryDiff);
+            }
 
-            this.averageDelaySeconds.AddSample(delaySeconds);
+            this.lastDeliveryTime = DateTime.Now;
```
Aprogiena: Date time provider

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -100,31 +121,71 @@ public BlockPullerBehavior(IBlockPuller blockPuller, IInitialBlockDownloadState
...
-            this.logger.LogTrace("({0}:{1})", nameof(delaySeconds), delaySeconds);
+            this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(blockSizeBytes), blockSizeBytes, nameof(delaySinceRequestedSeconds), delaySinceRequestedSeconds);
+
+            double adjustedDelay;
+
+            if (this.lastDeliveryTime == null)
+                adjustedDelay = delaySinceRequestedSeconds;
+            else
+            {
+                double deliveryDiff = (DateTime.Now - this.lastDeliveryTime).Value.TotalSeconds;
```
Aprogiena: datetime provider

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: Even better, you can simplify by initialising the values on line 128 with this and then apply adjustment in case last delivery time is available

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: And probably move down below to be executed as late as possible

Aprogiena: Actually, maybe it would be better if this is set at the end of OnMessageReceivedAsync

Aprogiena: Otherwise penalise would also update it, do we want that? And also to avoid time for processing in the callback

noescape00: Don't wanna put it to OnMessageReceivedAsync because it will update in case we received a block we didn't ask for. 

Unless we change BP.PushBlock to return bool. 

Also overhead is not that bad when we penalize the peer. 

Let me know what you think

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
noescape00: fixed

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
noescape00: fixed

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
Aprogiena: Doesn't matter which block it is, we need to count it

src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
```diff
@@ -21,7 +22,7 @@ public BlockPullerBehaviorTests()
             var loggerFactory = new ExtendedLoggerFactory();
             loggerFactory.AddConsoleWithFilters();
 
-            this.behavior = new BlockPullerBehavior(puller.Object, ibdState.Object, loggerFactory);
+            this.behavior = new BlockPullerBehavior(puller.Object, ibdState.Object, new DateTimeProvider(), loggerFactory);
```
Aprogiena: We usually use the default instance

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
noescape00: fixed

src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
noescape00: fixed

PR #2106, opened 2018-09-03
[Activation] Fix test sending_from_fifty_addresses_to_one_address

zeptin:
The problem appears to have been that no block was getting mined between `node1_sends_funds_to_node2_TO_fifty_addresses` and `node2_receives_the_funds`, so the wait loop in the latter would generally time out without the transaction appearing in the wallet.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
@@ -93,17 +93,18 @@ namespace Stratis.Bitcoin.IntegrationTests.Wallet
 
         private void node2_receives_the_funds()
         {
+            // Need to mine a block prior to checking for spendable transactions in node2's wallet.
+            this.sharedSteps.MineBlocks(1, this.nodes[NodeOne], WalletAccountName, WalletName, WalletPassword);
+
             TestHelper.WaitLoop(() => this.nodes[NodeTwo].FullNode.WalletManager().GetSpendableTransactionsInWallet(WalletName).Any());
 
             this.nodeTwoBalance = this.nodes[NodeTwo].WalletBalance(WalletName);
 
             this.nodeTwoBalance.Should().Be(Money.Coins(50));
 
-            this.nodes[NodeTwo].WalletHeight(WalletName).Should().BeNull();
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -93,18 +93,17 @@ private void Mine100Coins(CoreNode node)
 
         private void node2_receives_the_funds()
         {
+            // Need to mine a block prior to checking for spendable transactions in node2's wallet.
+            this.sharedSteps.MineBlocks(1, this.nodes[NodeOne], WalletAccountName, WalletName, WalletPassword);
```
bokobza: Why does the block need to be mined?
The transaction will appear in node2 and be spendable even with 0 confirmations.

src/Stratis.Bitcoin.IntegrationTests/Wallet/SendingToAndFromManyAddressesSpecification_Steps.cs
```diff
@@ -93,18 +93,17 @@ private void Mine100Coins(CoreNode node)
...
+
             TestHelper.WaitLoop(() => this.nodes[NodeTwo].FullNode.WalletManager().GetSpendableTransactionsInWallet(WalletName).Any());
 
             this.nodeTwoBalance = this.nodes[NodeTwo].WalletBalance(WalletName);
 
             this.nodeTwoBalance.Should().Be(Money.Coins(50));
 
-            this.nodes[NodeTwo].WalletHeight(WalletName).Should().BeNull();
```
bokobza: This test is correct, because if the block hasn't been mined, the block height will be null.

zeptin: This is the wallet height, not the block height.

bokobza: the method is badly named but it checks the block height of the latest transaction.

zeptin:
This needs further investigation, the underlying problem seems to be that node 2 occasionally does not sync from node 1, so the unconfirmed transaction never appears in the wallet. Mining another block tends to kickstart the syncing process and causes the test to pass, but we should probably fix the sync problem instead.

PR #2109, opened 2018-09-04
[SC] No P2PKH to contracts

### LINKED ISSUE #2007: [SC] P2PKH should not be allowed to be sent to smart contracts.
Would be a new consensus rule / mempool rule.

We don't want to see people losing money to this, and the alternative, where we execute p2pkh calls, is not good either. 

If the performance impact is severe we can always remove this later without significant impact to the network.

codingupastorm:
Closes #2007

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
@@ -20,23 +20,31 @@ namespace Stratis.Bitcoin.Features.SmartContracts
         /// <summary>
         /// These rules can be checked instantly. They don't rely on other parts of the context to be loaded.
         /// </summary>
-        private static readonly List<ISmartContractMempoolRule> preTxRules = new List<ISmartContractMempoolRule>
-        {
-            new MempoolOpSpendRule(),
-            new TxOutSmartContractExecRule()
-        };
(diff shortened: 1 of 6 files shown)


src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
```diff
@@ -20,23 +20,31 @@ public class SmartContractMempoolValidator : MempoolValidator
...
-        private static readonly List<ISmartContractMempoolRule> feeTxRules = new List<ISmartContractMempoolRule>
-        {
-            new SmartContractFormatRule()
-        };
+        private readonly List<ISmartContractMempoolRule> feeTxRules;
 
         public SmartContractMempoolValidator(ITxMempool memPool, MempoolSchedulerLock mempoolLock, IDateTimeProvider dateTimeProvider, MempoolSettings mempoolSettings, ConcurrentChain chain, ICoinView coinView, ILoggerFactory loggerFactory, NodeSettings nodeSettings, IConsensusRules consensusRules)
             : base(memPool, mempoolLock, dateTimeProvider, mempoolSettings, chain, coinView, loggerFactory, nodeSettings, consensusRules)
         {
+            var p2pkhRule = new P2PKHNotContractRule();
```
rowandh: Can we pass parent in to the constructor instead of using `Initialize`?

PR #2110, opened 2018-09-04
[Activation] Fix Given_NodesAreSynced_When_ABigReorgHappens_Then_TheReorgIsIgnored

fassadlr:


src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -189,9 +187,9 @@ public void Given_NodesAreSynced_When_ABigReorgHappens_Then_TheReorgIsIgnored()
                 stratisSyncer.CreateRPCClient().RemoveNode(stratisReorg.Endpoint);
                 TestHelper.WaitLoop(() => !TestHelper.IsNodeConnected(stratisReorg));
 
-                Task<List<uint256>> t1 = Task.Run(() => stratisMiner.GenerateStratisWithMiner(11));
-                Task<List<uint256>> t2 = Task.Delay(1000).ContinueWith(t => stratisReorg.GenerateStratisWithMiner(12));
-                Task.WaitAll(t1, t2);
+                stratisMiner.GenerateStratisWithMiner(11);
```
dangershony: How does this fix the test?
IT was done like this for a reason because we will get collisions (hmm unless we use the overlaods)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -650,9 +648,9 @@ private async Task<ConnectBlocksResult> ConnectChainAsync(ChainedHeader newTip,
 
                 await this.ResyncPeersAsync(peersToResync).ConfigureAwait(false);
 
-                if (this.network.Consensus.MaxReorgLength != 0)
+                if (this.chainState.MaxReorgLength != 0)
```
dangershony: Whats the difference?

fassadlr: The main issue was that we set the MaxReorgLength on ChainState in the test but we check the `MaxReorgLength`  on network when we want to save it, so `FinalizedBlock` doesn't get saved.

We CAN however still check network's `MaxReorgLength` in CM but then we need to set network's MaxReorgLength in the test (this might be an issue as network's MaxReorgLength  currently has a protected set).

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
fassadlr: If I run the tests in a task like that it looks like GenerateCoinsFastProvider gets out of sync. I don't understand why we need to run it like this?

dangershony: It was a hack to avoid block collusion,  (I think...)

dangershony: Actually there will be no collusion because it will be done sequentially so I don't  remember why we did it like this maybe @Aprogiena remembers he helped review this.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: In this case try to use reflection?

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
```diff
@@ -23,10 +27,50 @@ public class NodeSyncTests
...
+                subsidyHalvingInterval: 210000,
+                majorityEnforceBlockUpgrade: 750,
+                majorityRejectBlockOutdated: 950,
+                majorityWindow: 1000,
+                buriedDeployments: base.Consensus.BuriedDeployments,
+                bip9Deployments: base.Consensus.BIP9Deployments,
+                bip34Hash: new uint256("0x000000000000024b89b42a942fe0d9fea3bb44ab7bd1b19115dd6a759c0808b8"),
+                ruleChangeActivationThreshold: 1916, // 95% of 2016
+                minerConfirmationWindow: 2016, // nPowTargetTimespan / nPowTargetSpacing
+                maxReorgLength: 10,
```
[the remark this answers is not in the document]
fassadlr: I can do another PR where we clone the consensus object? Then we only need to set one property ;)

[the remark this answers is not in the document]
dangershony: It will be useful no? for future changes as well

src/Stratis.Bitcoin.IntegrationTests/NodeSyncTests.cs
Aprogiena: It was certainly needed before just to generate different timestamp. We wanted to different chains to be created and without this we only created one chain by two different miners

Aprogiena: If you remove the task, you must assert that the two chains are different exactly at the first block that is created here, otherwise you could have false pass

fassadlr: @Aprogiena I believe we do check for this a few lines down:

```
 // make sure the nodes are actually on different chains.
Assert.NotEqual(stratisMiner.FullNode.Chain.GetBlock(2).HashBlock, stratisReorg.FullNode.Chain.GetBlock(2).HashBlock);
```

dangershony: But this check means the test will fail.
But I think there is no problem because we are doing it sequential not parallel, it was done in parallel because at the time we mined many more blocks then it was reduced by overriding the maxreorg property I remember now.

Aprogiena: The test should pass and it needs to pass with those two chains being different

PR #2113, opened 2018-09-04
[Activation] Periodic log improvements

noescape00:
Refactoring and improvements that achieve: 

1) Ability to inject INodeStats to the component that produces the stats instead of how we are doing it now- feature implements stats interface and then goes one or 2 layers inside to grab the logs from specific component
2) Ability to set the priority of logs (before we had priority based on in which order features are registered)

Ps
`NodeStats.cs` is probably the only thing to review here.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
@@ -18,7 +18,7 @@ using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.BlockStore
 {
-    public class BlockStoreFeature : FullNodeFeature, INodeStats, IFeatureStats
+    public class BlockStoreFeature : FullNodeFeature
     {
         private readonly ConcurrentChain chain;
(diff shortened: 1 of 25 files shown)


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -59,26 +60,21 @@ public class BlockStoreFeature : FullNodeFeature, INodeStats, IFeatureStats
...
-                                     highestBlock.HashBlock);
-            }
-        }
+                string log = $"BlockStore.Height: ".PadRight(LoggingConfiguration.ColumnLength + 1) + highestBlock.Height.ToString().PadRight(8) +
+                             $" BlockStore.Hash: ".PadRight(LoggingConfiguration.ColumnLength - 1) + highestBlock.HashBlock;
 
-        /// <inheritdoc />
-        public void AddFeatureStats(StringBuilder benchLog)
-        {
-            this.blockStoreQueue.ShowStats(benchLog);
```
dangershony: Was these not showing anything?

PR #2114, opened 2018-09-04
Make Consensus/IConsensus an object that can be merged

fassadlr:
Should be pretty self-explanatory but if not, this PR makes the Consensus object on Network merg-able so that a developer can merge an instance of Consensus to the current one. In other words provide the ability to makes changes to single properties, for example MaxReorgLength, to the current Consensus instance without affecting the immutability of it or having to construct a full new instance.

### DIFF
--- a/src/NBitcoin/Consensus.cs
+++ b/src/NBitcoin/Consensus.cs
@@ -11,76 +11,76 @@ namespace NBitcoin
         public long CoinbaseMaturity { get; set; }
 
         /// <inheritdoc />
-        public Money PremineReward { get; }
+        public Money PremineReward { get; private set; }
 
         /// <inheritdoc />
-        public long PremineHeight { get; }
+        public long PremineHeight { get; private set; }
 
         /// <inheritdoc />
-        public Money ProofOfWorkReward { get; }
+        public Money ProofOfWorkReward { get; private set; }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


rowandh:
Sorry, but this change doesn't make much sense to me. 
>provide the ability to makes changes to single properties ... without affecting the immutability

Allowing changes to be made to the object means it _is_ no longer immutable.

It seems like we're only doing this to enable a test to be written, in which case the correct approach would be as you say - create a new consensus instance with the desired properties, and set it on the network.

fassadlr:
@rowandh I suppose this is what you were trying to achieve with MutableConsensus?

I dont want to create a whole Consensus object and set all the properties just to change one item on it?

bokobza:
This is to be used for tests only, right?

rowandh:
@fassadlr right, that's what I was trying to achieve. Creating a new Consensus object just to change one item for a test seems for more preferable over adding a confusing new "merge" function to Consensus' API.

bokobza:
what about just overriding the property using reflection, just for the tests?

Like this:
```
Type consensusType = typeof(Consensus);
consensusType.GetProperty("CoinbaseMaturity").SetValue(network.Consensus, 1);
```

fassadlr:
Closing this for now, until we come up with a better solution.

PR #2120, opened 2018-09-05
Fix MemoryPoolTests integration tests

fenix2222:


bokobza:
Nice, so what does it looks like in the end?

fenix2222:
image

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -156,10 +153,12 @@ public CoreNode CreateCustomNode(bool start, Action<IFullNodeBuilder> callback,
 
         private string GetNextDataFolderName(string folderName = null)
         {
-            var numberedFolderName = string.Join(".",
-                new[] { this.lastDataFolderIndex.ToString(), folderName }.Where(s => s != null));
```
monsieurleberre: really OK for me, I might written the {foldername, hash} instead of {hash, foldername} because now it makes more sense to have folder sorted by intended foldername than by a piece of GUID, but this is not a blocker at all :) 👍

fenix2222: > Did we consider using Interlocked.Increment(lastDataFolderIndex)

Yeah, that's actually the first thing I tried. Also tried locking it. I think the problem is that tests are trying to reuse the same indexes with different runs.

Issue #2092, opened 2018-08-31
speedup uint256.GetHashCode


noescape00:

zeptin:
Did you have in mind some kind of caching here to prevent recomputing it? I doubt we can outdo the compiler in optimising the actual implementation of `GetHashCode`.

noescape00:
Yeah. I'd have a bool `hashCodeRecalculationRequired` which will be set to true each time we modify uint256. On gethashcode I'd check against the bool and if it's true- calculate it and save to a variable. if it's false- return hashed value. 

Btw @Aprogiena suggested to create this issue so I'd also check with him.

noescape00:
Or because when we compare uint256 instances we are not using hashcodes we can also generate a random number in constructor and it will be our permanent hash code

noescape00:
Had a chat with Apro and the follwoing solution should be implemented: 

`this.pn7` should be returned as a hashcode instead of advanced hashcode calculation because: 

1) btc core does that
2) we use uint256.GetHashCode only for balancing buckets in the dictionary and not for comparing 2 instances of uint256

noescape00:
@Aprogiena pn7 doesn't work for btc because first few bits are 0- because of PoW
I think it will be reasonable to use pn1

PR #2122, opened 2018-09-05
speedup uint256.GetHashCode

noescape00:
Fix: #2092

### DIFF
--- a/src/NBitcoin/UInt256.cs
+++ b/src/NBitcoin/UInt256.cs
@@ -431,21 +431,10 @@ namespace NBitcoin
 
         public override int GetHashCode()
         {
-            int hash = 17;
-            unchecked
-            {
-                hash = hash * 31 + (int) this.pn0;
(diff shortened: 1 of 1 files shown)


noescape00:
>Is this assuming that the uint256 has been randomly generated ?

Yeah, it's under assumption that uint256 is a sha256 hash 

>Also wondering if you had the opportunity to measure the perf improvement ?

Nope, but I'd guess it's 5-10 times better

noescape00:
>Ok I need to have a bit more understanding of this comment

Sure. Call me or talk to Apro, he knows way more about this than me.

dangershony:
Ok looks good but @bokobza suggested we do this on master directly

noescape00:
Good suggestion. Did it on activation because it will be merged to master this week anyway.

But if you insist I can close this one and open another PR for master.

bokobza:
Please do @noescape00, because we're gonna keep a version of master somewhere so if this change is good for both, let's put it in master.

PR #2128, opened 2018-09-06
[Activation] Fix MempoolReceiveFromManyNodes

fassadlr:
image


fassadlr:
@dangershony yes it was but not on my PC, its definitely not right to create hundreds of logger and connection manager instances like this ;)

PR #2134, opened 2018-09-06
[Activation] feature executor fix

noescape00:
1) we didn't log aggregated exception properly (no stack trace which is 95% of what you need if you see an error)
2) before we would continue initializing the node after some feature failed. No need to do that.

### DIFF
--- a/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
+++ b/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
@@ -66,7 +66,6 @@ namespace Stratis.Bitcoin.Builder
         /// </summary>
         /// <param name="callback">Delegate to run start or stop method of the feature.</param>
         /// <param name="reverseOrder">Reverse the order of which the features are executed.</param>
-        /// <remarks>This method catches exception of start/stop methods and then, after all start/stop methods were called for all features.</remarks>
         private void Execute(Action<IFullNodeFeature> callback, bool reverseOrder = false)
         {
             this.logger.LogTrace("({0}:{1})", nameof(reverseOrder), reverseOrder);
--- a/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
+++ b/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
@@ -51,30 +51,14 @@ namespace Stratis.Bitcoin.Builder
         /// <inheritdoc />
         public void Initialize()
         {
-            try
-            {
-                this.Execute(service => service.ValidateDependencies(this.node.Services));
-                this.Execute(service => service.Initialize());
-            }
-            catch (Exception ex)
-            {
-                this.logger.LogError("An error occurred starting the application: {0}", ex);
-                throw;
-            }
+            this.Execute(service => service.ValidateDependencies(this.node.Services));
+            this.Execute(service => service.Initialize());
         }
 
         /// <inheritdoc />
         public void Dispose()
         {
-            try
-            {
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
```diff
@@ -51,42 +51,25 @@ public FullNodeFeatureExecutor(IFullNode fullNode, ILoggerFactory loggerFactory)
...
         }
 
         /// <summary>
         /// Executes start or stop method of all the features registered with the associated full node.
         /// </summary>
         /// <param name="callback">Delegate to run start or stop method of the feature.</param>
         /// <param name="reverseOrder">Reverse the order of which the features are executed.</param>
-        /// <remarks>This method catches exception of start/stop methods and then, after all start/stop methods were called
-        /// for all features, it throws AggregateException if there were any exceptions.</remarks>
+        /// <remarks>This method catches exception of start/stop methods and then, after all start/stop methods were called for all features.</remarks>
```
Aprogiena: Incomplete comment

bokobza: In fact, you can remove this comment now.

src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
```diff
@@ -103,21 +86,13 @@ private void Execute(Action<IFullNodeFeature> callback, bool reverseOrder = fals
                     }
                     catch (Exception ex)
                     {
-                        if (exceptions == null)
-                        {
-                            exceptions = new List<Exception>();
-                        }
-
-                        exceptions.Add(ex);
+                        this.logger.LogError("Exception occurred: {0}", ex.ToString());
```
bokobza: I agree that for intialization, we don't want to continue.
For disposing of resources, however, I'd be quite keen for the node to dispose of as many resources as possible, rather than bomb out on the first problem and then not finishing the others gracefully.

src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
noescape00: fixed

src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
noescape00: We are not supposed to throw from initialize or dispose methods. 
I think this is something that should be for the component to decide. For example if we expect some kind of error in initialization or disposal and it's not critical (we will be able to run node properly after initialize failed or run it next time when dispose fails)- then the components itself should handle this exception and swallow it with some kind of error message to the console. 

Before we used to catch all exceptions which didn't seem correct. To me this was more like a sanity check- that kind of stuff should never happen.

noescape00:
>In fact, you can remove this comment now.

fixed

src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
bokobza: > We are not supposed to throw from initialize or dispose methods.
but if we do throw in Dispose, and we do, you'll exit at the first problem and won't try to dispose the other features.

>  the components itself should handle this exception and swallow it with some kind of error message to the console.
but we don't have that yet.

dangershony:
I disgaree with this change, each feature is independent and failure of one should not stop the dispose of another (initialise is perhaps a different issue) we should stop the node on the first feature that fails.

noescape00:
ok, makes sense.

PR #2135, opened 2018-09-07
[SC] State transition error improvements

### LINKED ISSUE #2121: [SC] Move thrown state transition exceptions to ITE
AC:
* Return all expected errors as part of the StateTransitionResult
* In the ITE, observe the error and throw the appropriate exception

@codingupastorm

rowandh:
Closes #2121.
* Adds success and error result types which are returned on `StateTransitionResult`.
* Returns an error type for all expected state transition errors.
* Returns any exceptions generated within the VM execution.
* Pushes the handling of internally generated transaction errors back to the ITE.

For review only. Do not merge until we establish what's going on with core-way-activation.

### DIFF
--- a/src/Stratis.SmartContracts.Executor.Reflection/StateTransitionResult.cs
+++ b/src/Stratis.SmartContracts.Executor.Reflection/StateTransitionResult.cs
@@ -86,7 +86,7 @@ namespace Stratis.SmartContracts.Executor.Reflection
 
         /// <summary>
         /// An exception thrown by the VM. This value is null unless <see cref="Kind"/>
-        /// equals <see cref="StateTransitionErrorKind.VmError"/>.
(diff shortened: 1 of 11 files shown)


src/Stratis.SmartContracts.Executor.Reflection/StateTransitionResult.cs
```diff
@@ -34,9 +70,104 @@ public class StateTransitionResult
...
+
+        public bool IsSuccess { get; }
+
+        public bool IsFailure => !this.IsSuccess;
+
+        /// <summary>
+        /// The successful result of the state transition. This value is null unless
+        /// <see cref="IsSuccess"/> equals true.
+        /// </summary>
+        public StateTransitionSuccess Success { get; }
```
codingupastorm: This is interesting. Not opposed to it, but I think we should rename this then, if we have `IsSuccess` as well.

PR #2138, opened 2018-09-07
Should stake store flush be inside the lock?

dangershony:


Aprogiena:
Quite nontrivial change in terms of possible impact, yet no rationale provided

dangershony:
Will it not prevent flushing out of sync from coinview?

Aprogiena:
No, if there is power failure, locking does not help

If you have different scenario in mind, you need to describe it

dangershony:
ok this is solved and was never a problem

PR #2150, opened 2018-09-08
[SC] Exception in stored receipt

### LINKED ISSUE #2058: [SC] Add Exception to the stored receipts
As per title. 

Only stored on the node, not part of consensus.

codingupastorm:
Think we actually just want to pass a string around - not an Exception. Hence this.

Closes #2058


src/Stratis.SmartContracts.Executor.Reflection/ContractInvocationResult.cs
```diff
@@ -15,32 +16,46 @@ private ContractInvocationResult(object result)
...
         }
 
-        public static ContractInvocationResult Failure(ContractInvocationErrorType errorType, Exception exception)
+        /// <summary>
+        /// Return invocation failure for cases related to execution inside contract code.
+        /// </summary>
+        public static ContractInvocationResult ExecutionFailure(ContractInvocationErrorType errorType, Exception exception)
         {
-            return new ContractInvocationResult(errorType);
+            return new ContractInvocationResult(errorType, new ContractErrorMessage(exception.ToString()));
```
rowandh: Should we be using Exception.Message rather than Exception.ToString? ToString will be very verbose, including a stack trace.

src/Stratis.SmartContracts.Executor.Reflection/ReflectionVirtualMachine.cs
```diff
@@ -56,7 +56,8 @@ public VmExecutionResult Create(IContractState repository, ISmartContractState c
                 if (!validation.IsValid)
                 {
                     this.logger.LogTrace("(-)[CONTRACT_VALIDATION_FAILED]");
-                    return VmExecutionResult.Error(new SmartContractValidationException(validation.Errors));
+                    // TODO: List errors by string.
+                    return VmExecutionResult.Error(new ContractErrorMessage(new SmartContractValidationException(validation.Errors).ToString()));
```
rowandh: .ToString or .Message?

src/Stratis.SmartContracts.Executor.Reflection/ContractInvocationResult.cs
codingupastorm: Yeah, as this is only for the non-consensus version of the receipts, for convenience we will store the whole thing.

PR #2154, opened 2018-09-10
[Activation] bugfixes for coinview

noescape00:


src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -222,7 +225,7 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
                 {
                     transaction.ValuesLazyLoadingIsOn = false;
                     transaction.SynchronizeTables("BlockHash", "Coins", "Rewind");
-                    transaction.Technical_SetTable_OverwriteIsNotAllowed("Coins");
+                    //transaction.Technical_SetTable_OverwriteIsNotAllowed("Coins"); // Why it was there and what is it for? No one knows.
```
fassadlr: Isn't this exactly to prevent the "coins" table from overwrites? I.e. overwriting existing data in the table? I would think that is an absolute must to include.

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -207,10 +207,10 @@ public override void Initialize()
 
             this.consensusRules.Initialize().GetAwaiter().GetResult();
 
-            this.consensusManager.InitializeAsync(this.chain.Tip).GetAwaiter().GetResult();
-
             this.consensusRules.Register();
 
+            this.consensusManager.InitializeAsync(this.chain.Tip).GetAwaiter().GetResult();
```
fassadlr: So where are we doing this here now?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +367,47 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
+                        cacheItem.OriginalOutputs = unspentOutput?.Outputs.ToArray();
+
+                        this.unspents.TryAdd(unspent.TransactionId, cacheItem);
+                    }
+                    else
                         this.logger.LogTrace("Outputs of transaction ID '{0}' are in cache already, updating them.", unspent.TransactionId);
-                        if (cacheItem.UnspentOutputs != null) cacheItem.UnspentOutputs.Spend(unspent);
-                        else cacheItem.UnspentOutputs = unspent;
+
+
```
fassadlr: Extra line.

src/Stratis.Bitcoin/Base/BaseFeature.cs
noescape00: yes

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
noescape00: but we want to override them (for example when some outputs were spend and some were not) and we did that before

src/Stratis.Bitcoin/Base/BaseFeature.cs
noescape00: yes

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
noescape00: fixed

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
Aprogiena: Why would we need it? I can rather imagine situations when this is the problem, which would be when you have transaction with multiple outputs which is in the database already and you just want to update one of the output. But maybe such an update is never done? But do we know that for sure? From the code it is not obvious for me why such an update could not happen.

src/Stratis.Bitcoin/Base/BaseFeature.cs
fassadlr: So I meant "why"?

noescape00:
unit tests passing
coinview and node sync integration tests are passing

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -237,7 +238,7 @@ public async Task<FetchCoinsResponse> FetchCoinsAsync(uint256[] txIds, Cancellat
                     var cache = new CacheItem();
                     cache.ExistInInner = unspent != null;
                     cache.IsDirty = false;
-                    cache.UnspentOutputs = unspent;
+                    cache.UnspentOutputs = unspent?.Clone();
```
dangershony: Ah ok I don't think this is the correct fix, it requires fetching the coins later from disk as @fassadlr mentions.
Cloning at the UTXO set level is more correct in my opinion.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +367,47 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
-                        if (cacheItem.UnspentOutputs != null)
-                        {
-                            clone.Outputs = cacheItem.UnspentOutputs.Outputs.ToArray();
-                            rewindData.OutputsToRestore.Add(clone);
-                        }
-                        else
-                        {
-                            rewindData.TransactionsToRemove.Add(unspent.TransactionId);
-                        }
+                        FetchCoinsResponse result = await this.inner.FetchCoinsAsync(new[] {unspent.TransactionId}).ConfigureAwait(false);
```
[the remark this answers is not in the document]
dangershony: I agree this is not needed, you do this because of incorrect cloning earlier.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
dangershony: Ah ok this does not require fetching again however not sure it addresses the issue of the cache item directly being manipulated when we modify the coinview set in the rules engine.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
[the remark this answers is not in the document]
dangershony: Ok yeah this is definitely needed because of the evict method that is called in the fetch method.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
Aprogiena: It does, because the result of the call which goes to the coin view rule set is now different instance. So you have one thing going to the cache and another one being returned from the call which is used by the rule

Aprogiena: Currently the only modification of the cache item should happen in the save change async, so there should not be anything else touching the cache item

src/NBitcoin/Transaction.cs
```diff
@@ -1654,6 +1654,11 @@ public string ToHex()
             return Encoders.Hex.EncodeData(this.ToBytes());
         }
 
+        public override string ToString()
+        {
+            return $"TrxHash={this.GetHash()}";
```
Aprogiena: just `return this.GetHash().ToString()`;

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -368,7 +362,7 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
                 foreach (UnspentOutputs unspent in unspentOutputs)
                 {
                     if (!this.unspents.TryGetValue(unspent.TransactionId, out CacheItem cacheItem))
-                    {
+                    {
```
Aprogiena: ee

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -379,34 +373,35 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
 
                     // We'll need to restore the original outputs, so we clone it
                     // and save it in rewind data.
                     UnspentOutputs clone = unspent.Clone();
 
-                    if (cacheItem.OriginalOutputs != null)
+                    if (cacheItem.UnspentOutputs != null)
                     {
-                        clone.Outputs = cacheItem.OriginalOutputs.ToArray();
+                        // We take the original items that are in cache and put them in rewind data
```
Aprogiena: `.`

src/Stratis.Bitcoin.Features.Consensus/UnspentOutputSet.cs
```diff
@@ -60,7 +60,10 @@ public void SetCoins(UnspentOutputs[] coins)
             foreach (UnspentOutputs coin in coins)
             {
                 if (coin != null)
-                    this.unspents.Add(coin.TransactionId, coin);
+                {
+                    // We clone every UnspentOutputs as those will get mutated.
+                    this.unspents.Add(coin.TransactionId, coin.Clone());
+                }
```
Aprogiena: this is worse than our solution because now you don't have TrySetCoins covered below

Aprogiena: basically we were protecting every fetch, you are only protecting coin view rule. this means that someone working with fetched coin (e.g. miner) could have their coin modified silently without noticing

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -379,34 +373,35 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
                         cacheItem.ExistInInner = unspentOutput != null;
                         cacheItem.IsDirty = false;
 
-                        cacheItem.UnspentOutputs = unspentOutput?.Clone();
-                        cacheItem.OriginalOutputs = unspentOutput?.Outputs.ToArray();
+                        cacheItem.UnspentOutputs = unspentOutput;
```
Aprogiena: Here I believe we are making the same again because we have having the same reference as the actual coin in the set that is used by coin view Rule

src/Stratis.Bitcoin.Features.Consensus/UnspentOutputSet.cs
dangershony: yes correct you convinced me I will change back

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +361,54 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
-                        }
+                        this.logger.LogTrace("Outputs of transaction ID '{0}' are not found in cache, creating them.", unspent.TransactionId);
 
+                        FetchCoinsResponse result = await this.inner.FetchCoinsAsync(new[] {unspent.TransactionId}).ConfigureAwait(false);
+
+                        UnspentOutputs unspentOutput = result.UnspentOutputs[0];
+
+                        cacheItem = new CacheItem();
+                        cacheItem.ExistInInner = unspentOutput != null;
+                        cacheItem.IsDirty = false;
```
fassadlr: Isnt this false by default?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +361,54 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
+
+                        cacheItem.UnspentOutputs = unspentOutput?.Clone();
+
+                        this.unspents.TryAdd(unspent.TransactionId, cacheItem);
```
fassadlr: Can we perhaps rename `unspents` to be more specific? I.e. I'm not always sure which unspent set we are talking about. Perhaps `cachedUTXOset? or `cachedUnspentItems`?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +361,54 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
+                    }
+                    else
+                    {
                         this.logger.LogTrace("Outputs of transaction ID '{0}' are in cache already, updating them.", unspent.TransactionId);
-                        if (cacheItem.UnspentOutputs != null) cacheItem.UnspentOutputs.Spend(unspent);
-                        else cacheItem.UnspentOutputs = unspent;
+                    }
+
+                    // If cacheItem.UnspentOutputs is null this means the trx was not stored in the disk,
```
fassadlr:
```
// If cacheItem.UnspentOutputs is null, the trx was not stored on disk and as such
// the trx (and UTXO) is new. All the UTXOs need to be stored in cache
// otherwise we persist to cache only the UTXOs that have been spent.
```

ferdeen: It says if null, then the statement on line 391 checks for not null - is this correct ?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -366,36 +361,54 @@ public async Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, I
...
+                    {
+                        // To handle rewind we'll need to restore the original outputs,
+                        // so we clone it and save it in rewind data.
+                        UnspentOutputs clone = unspent.Clone();
+
+                        // We take the original items that are in cache and put them in rewind data.
+                        clone.Outputs = cacheItem.UnspentOutputs.Outputs.ToArray();
+                        rewindData.OutputsToRestore.Add(clone);
+
+                        // Now modify the cached items with the mutated data.
```
ferdeen: What do you mean by mutated data ? Is the the code below setting the `unspent` item as **spent** ?

dangershony: Yes correct its marking as spent the UTXO that have been marked as spent after passing in the rules engine.

PR #2158, opened 2018-09-11
Include extra LogInformation around application shutdown.

ferdeen:
I've included `LogInformation` when objects are being disposed at shutdown.  I looked at making this trace level logging, but it was difficult to see which objects were shutdown, with other trace level logging going on.

Also noticed there were times where Consensus stats were shown during shutdown, which looked messy.  I've stopped this happening.

### DIFF
--- a/src/Stratis.Bitcoin/Base/BaseFeature.cs
+++ b/src/Stratis.Bitcoin/Base/BaseFeature.cs
@@ -278,25 +278,37 @@ namespace Stratis.Bitcoin.Base
             this.logger.LogInformation("Flushing peers...");
             this.flushAddressManagerLoop.Dispose();
 
+            this.logger.LogInformation("peerAddressManager...");
             this.peerAddressManager.Dispose();
 
+            this.logger.LogInformation("partialValidator...");
             this.partialValidator.Dispose();
 
-            this.logger.LogInformation("Flushing headers chain...");
-            this.flushChainLoop?.Dispose();
+            if (this.flushChainLoop != null)
+            {
+                this.logger.LogInformation("Flushing headers chain...");
+                this.flushChainLoop.Dispose();
+            }
 
+            this.logger.LogInformation("Saving chainRepository...");
             this.chainRepository.SaveAsync(this.chain).GetAwaiter().GetResult();
 
             foreach (IDisposable disposable in this.disposableResources)
             {
+                this.logger.LogInformation($"{disposable.GetType().Name}...");
                 disposable.Dispose();
             }
 
+            this.logger.LogInformation("blockPuller...");
             this.blockPuller.Dispose();
 
+            this.logger.LogInformation("consensusManager...");
             this.consensusManager.Dispose();
+
+            this.logger.LogInformation("consensusRules...");
             this.consensusRules.Dispose();
 
+            this.logger.LogInformation("blockStore...");
             this.blockStore.Dispose();
         }
     }
--- a/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
+++ b/src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
@@ -68,7 +68,11 @@ namespace Stratis.Bitcoin.Builder
         {
             try
             {
-                this.Execute(feature => feature.Dispose(), true);
+                this.Execute(feature =>
+                {
+                    this.logger.LogInformation($"{feature.GetType().Name}...");
+                    feature.Dispose();
+                }, true);
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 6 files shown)


src/Stratis.StratisD/Properties/launchSettings.json
```diff
@@ -5,7 +5,7 @@
     },
     "Stratis.StratisD Test": {
       "commandName": "Project",
-      "commandLineArgs": "-testnet"
+      "commandLineArgs": "-testnet -datadir=D:\\\\Stratis\\\\ShutDownCore -agentprefix=ferdeen"
```
dangershony: Delete this lol

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -278,25 +278,37 @@ public override void Dispose()
             this.logger.LogInformation("Flushing peers...");
             this.flushAddressManagerLoop.Dispose();
 
+            this.logger.LogInformation("peerAddressManager...");
```
dangershony: You probably want to add `Disposing bla...`

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -278,25 +278,37 @@ public override void Dispose()
...
             this.peerAddressManager.Dispose();
 
+            this.logger.LogInformation("partialValidator...");
```
dangershony: Partial validator should probably get disposed inside CM come to think about it

src/Stratis.Bitcoin/Base/BaseFeature.cs
ferdeen: I did, as it's in the Dispose() method this is shown in the output.  It would say "... Dispose() Disposing peerAddressManager..."

Aprogiena: What is the user experience here seeing `peerAddressManager...`? How about writing English sentence? Also don't expose names of classes or variables or properties, write normal English sentence `Disposing partial block validator...`

src/Stratis.Bitcoin/Builder/FullNodeFeatureExecutor.cs
```diff
@@ -68,7 +68,11 @@ public void Dispose()
         {
             try
             {
-                this.Execute(feature => feature.Dispose(), true);
+                this.Execute(feature =>
+                {
+                    this.logger.LogInformation($"{feature.GetType().Name}...");
```
dangershony: This I think we don't need as we have logging inside each feature already

src/Stratis.Bitcoin/Base/BaseFeature.cs
ferdeen: Ok makes sense.

src/Stratis.Bitcoin/Base/BaseFeature.cs
ferdeen: I'll check this out.

src/Stratis.Bitcoin/Base/BaseFeature.cs
ferdeen: I'll add Disposing back in and we can see what it looks like.

monsieurleberre: I think it is OK to put the name of the class, or something that doesn't come up in the Oxford dictionary, we are really logging info about implementation details, so it is not shocking to me

Aprogiena: Then it doesn't belong on an information level

src/Stratis.Bitcoin/FullNode.cs
```diff
@@ -264,17 +264,24 @@ public void Dispose()
             // Fire INodeLifetime.Stopping.
             this.nodeLifetime.StopApplication();
 
+            this.logger.LogInformation("Disposing connection manager...");
             this.ConnectionManager.Dispose();
 
             foreach (IDisposable disposable in this.Resources)
+            {
+                this.logger.LogInformation($"{disposable.GetType().Name}...");
```
Aprogiena: i'd remove this one

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -278,25 +276,34 @@ public override void Dispose()
             this.logger.LogInformation("Flushing peers...");
             this.flushAddressManagerLoop.Dispose();
 
+            this.logger.LogInformation("Disposing peer address manager...");
             this.peerAddressManager.Dispose();
 
-            this.partialValidator.Dispose();
```
Aprogiena: what is the rationale here?

ferdeen: @dangershony wanted this object to be disposed of in ConsensusManager, not base feature.

Also we noticed this object wasn't used in base feature.

PR #2159, opened 2018-09-11
[Activation] block puller behavior stack overflow bugfix

noescape00:


src/Stratis.Bitcoin.Tests/BlockPulling/BlockPullerBehaviorTests.cs
```diff
@@ -62,6 +64,32 @@ public void QualityScoreIsRelativeToBestSpeed()
             Assert.True(this.DoubleEqual(0.5, this.behavior.QualityScore));
         }
 
+        [Fact]
+        public void SpeedCalculatedCorrectlyWhenSeveralBehaviorsStall()
+        {
+            var behaviors = new List<BlockPullerBehavior>();
+
+            for (int i = 0; i < 125; i++)
```
ferdeen: Why 125, can't you just do a couple of samples ? Line 86 is always setting `blockSizeBytes` to zero, so the total some will always be zero, regardless of how many `behaviors` your iterate over.

dangershony: Looks ok like this

noescape00: >Why 125

That's max amount of connections we can have. Therefore max amount of behaviors we can have

PR #2163, opened 2018-09-11
Core way of syncing

dangershony:
This PR represents several months in the making.

We now use the same methods as bitcoin core to synchronise blocks from peers.
Notable components introduced in this PR:
- **ConsensusManager:**  A single point where decisions are made on how to achieve consensus.
- **ChainedHeaderTree:** A component that will store potential branches of interesting chains we may want to download blocks from.
- **BlockPuller:** An enhanced and improved puller to fetch blocks from peers.
- **ConsensusRuleEngine:** Responsible for executing consensus rules (consensus rules are ultimately going to be attached per network).

The consensus code is now part of the core framework. 
In order to synchronise with nodes that do not require full validation, a node can just subscribe to the rules it requires (light wallets for example).

Current Status:
A few known bugs that sometimes crash the node when syncing the Stratis network.

All tests pass.
image


dangershony:
Please note after the latest merge I am not running tests again and syncing the networks.

src/Stratis.Bitcoin.Features.Consensus.Tests/PeerBanningTest.cs
```diff
@@ -39,14 +39,14 @@ private async Task<(TestChainContext context, IPEndPoint peerEndPoint)> Initiali
             return (context, peerEndPoint);
         }
 
-        [Fact]
+        [Fact(Skip = "Revisit with ConsensusManager tests")]
```
codingupastorm: Hey are these tests happening for ConsensusManager? Can delete from here now?

dangershony: Well we still did not write the unit tests for CM.

Issue #2153, opened 2018-09-10
[Activation] Node does not advance after failed Coinstake signature verification


fassadlr:
My node has stopped syncing after block 448135.

After a bit of debugging I found out that block 448136 (the next block) contains an UTXO which which fails signature verification here:

```
private bool VerifySignature(UnspentOutputs coin, Transaction txTo, int txToInN, ScriptVerify flagScriptVerify)
{
    this.logger.LogTrace("({0}:'{1}/{2}',{3}:{4},{5}:{6})", nameof(coin), coin.TransactionId, coin.Height, nameof(txToInN), txToInN, nameof(flagScriptVerify), flagScriptVerify);

    TxIn input = txTo.Inputs[txToInN];

    if (input.PrevOut.N >= coin.Outputs.Length)
        return false;
```

This then causes the peers that contains the blocks, to be banned and as such my node does not advance (rightly so).

I suspect that this issue relates to the incorrect shutdown problem and/or the stake store load error?

For whatever reason my node cannot process my block data at height 448136.

@dangershony @noescape00 @Aprogiena

fassadlr:
Could this just be because my data is corrupt due to previously mentioned bugs?

zeptin:
@fassadlr Can you add the specific txid failing verification to this thread? And what are the values that the trace log is logging on entry to the method?

fassadlr:
Update:

From my logs it appears that when I connected to IP 13.64.76.48, which is an older Stratis X node (2.0.0.3), block validation failed and the peer got banned. However, even though I was connected to other peers with higher versions and/or block heights it failed to advance.

I'm investigating why.

PR #2171, opened 2018-09-12
Fix unhandled exceptions in ConsensusRuleEngine

fassadlr:
Currently if we encounter an unhandled exception in the `ConsensusRuleEngine`, it is not handled and as such ConsensusManager does not know what to do with the peer it received the invalid block from.

An example is being connected to a Stratis-X node that was upgraded and after the upgrade activation point we receive a block which is now not valid anymore and we cant load it from the stake store, an exception is thrown which is not handled. This causes the peer to still be connected and as such the node fails to advance.

I am not sure whether we should do this as now ALL unhandled exceptions in the `ConsensusRuleEngine` will now cause the peer it got the block from, to be banned and disconnected?

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
@@ -8,7 +8,6 @@ using Stratis.Bitcoin.Base;
 using Stratis.Bitcoin.Base.Deployments;
 using Stratis.Bitcoin.Configuration.Settings;
 using Stratis.Bitcoin.Consensus.Rules;
-using Stratis.Bitcoin.Primitives;
 using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Consensus
@@ -233,6 +232,10 @@ namespace Stratis.Bitcoin.Consensus
             catch (ConsensusErrorException ex)
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
+            }
+            catch (Exception ex)
+            {
+                ruleContext.ValidationContext.Error = new ConsensusError("unhandled-exception", ex.Message);
             }
         }
 
--- a/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
@@ -235,7 +235,7 @@ namespace Stratis.Bitcoin.Consensus
             }
             catch (Exception ex)
             {
-                ruleContext.ValidationContext.Error = new ConsensusError("unhandled-exception", ex.Message);
+                ruleContext.ValidationContext.Error = new ConsensusError("unhandled-exception", ex.Message.ToString());
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -234,6 +233,10 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
             }
+            catch (Exception ex)
+            {
+                ruleContext.ValidationContext.Error = new ConsensusError("unhandled-exception", ex.Message);
```
ferdeen: Could you do `ex.ToString()` here ? As there's an override in `ConsensusError` which includes the error code, as well as the message.

fassadlr: Done :)

fassadlr:
Fixes #2153

src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -234,6 +233,10 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
             }
+            catch (Exception ex)
```
dangershony: Probably not the correct solution, but also I don't think this is the problem to the bug you got.
I would say we need to catch such exception at the validators and handles there

dangershony:
>An example is being connected to a Stratis-X node that was upgraded and after the upgrade activation point we receive a block which is now not valid anymore and we cant load it from the stake store  

Is it confirmed this is the problem?
The description does not make sense so much if we dont have the stake info of the current block only of the previous one, which is what is failing to load.

dangershony:
I think the problem most likely is that we don't flush stake chain and coinview in the same time.

fassadlr:
Guys for me the issue is here in `OnPartialValidationSucceededAsync`. If  `connectBlocksResult` result contains an error why do we still continue after the `if` and further onto `StartPartialValidation`?

```
           if (connectBlocksResult != null)
            {
                if (connectBlocksResult.PeersToBan != null)
                {
                    var peersToBan = new List<INetworkPeer>();

                    lock (this.peerLock)
                    {
                        foreach (int peerId in connectBlocksResult.PeersToBan)
                        {
                            if (this.peersByPeerId.TryGetValue(peerId, out INetworkPeer peer))
                                peersToBan.Add(peer);
                        }
                    }

                    this.logger.LogTrace("{0} peers will be banned.", peersToBan.Count);

                    foreach (INetworkPeer peer in peersToBan)
                        this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, connectBlocksResult.BanDurationSeconds, connectBlocksResult.BanReason);
                }

                if (connectBlocksResult.ConsensusTipChanged)
                    await this.NotifyBehaviorsOnConsensusTipChangedAsync().ConfigureAwait(false);

                lock (this.peerLock)
                {
                    this.ProcessDownloadQueueLocked();
                }
            }

            if (chainedHeaderBlocksToValidate != null)
            {
                this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);

                // Start validating all next blocks that come after the current block,
                // all headers in this list have the blocks present in the header.
                foreach (ChainedHeaderBlock toValidate in chainedHeaderBlocksToValidate)
                    this.partialValidator.StartPartialValidation(toValidate.ChainedHeader, toValidate.Block, this.OnPartialValidationCompletedCallbackAsync);
            }
```

Aprogiena:
Yes, this actually looks like the real problem. So please discard the changes in this PR and create a new one that will fix that cause

noescape00:
@fassadlr would you like to fix it or should I?

fassadlr:
@noescape00 I will do the fix , thanks 👍

PR #2186, opened 2018-09-13
Updated verbose Nlog config

noescape00:
Update NLog config to log new components and namespaces that were introduced during the activation. 

Removed autocopy of the empty NLog config file to the build directory since it would usually override customly added config file which is a pain during debugging with logs (actually it was deleted in some other PR but now also deleting the config file that was in the daemon project)

Removed copypaste of the verbose logging config.


Documentation/NLog.config
```diff
@@ -1,138 +1,132 @@
...
-    
+
     <logger name="Stratis.Bitcoin.Features.Wallet.*" minlevel="Trace" writeTo="debugWalletFile" />
-    
+
     <logger name="Stratis.Bitcoin.Features.WatchOnlyWallet.*" minlevel="Trace" writeTo="debugWatchOnlyWalletFile" />
-    
-    <logger name="Stratis.Bitcoin.Base.*" minlevel="Trace" writeTo="debugBaseFile" />    
-    <logger name="Stratis.Bitcoin.Base.ChainHeadersBehavior" minlevel="Trace" writeTo="debugChainHeadersBehaviorFile" />
-    <logger name="Stratis.Bitcoin.Base.TimeSyncBehaviorState" minlevel="Trace" writeTo="debugTimeSyncFile" />
```
Aprogiena: Why to remove these two?

noescape00: ChainHeadersBehavior - now we don't have such class
TimeSyncBehaviorState- already covered in base. In the past 6 month we never needed logs explicitly for TimeSyncBehavior and in general it's already stable, no need to have separated target for it, it's ok for it to be in base log

bokobza:
No more log file in StratisD? 😢 
Fair enough though.

bokobza:
If a new developer wants to get a NLog.config file, where do they need to go?
To the documentation?

noescape00:
>No more log file in StratisD? 😢

Yeah. When we need to enable verbose logging we will copy log config from \StratisRepo\Documentation\NLog.config to putput dir

>If a new developer wants to get a NLog.config file, where do they need to go?
To the documentation?

Correct

>It'd be cool to have the file that was in src/Stratis.StratisD/NLog.config as an example somewhere.

We have it as an example in `using-logging.md`

Documentation/NLog.config
Aprogiena: How do you know we never needed logs for time sync? You're making claim that you cannot backup. I would suggest to revert it back

Aprogiena:
@noescape00 You should also make changes to using-logging.md

Documentation/NLog.config
noescape00: Fixed

noescape00:
>@noescape00 You should also make changes to using-logging.md

Fixed

PR #2196, opened 2018-09-13
handle unexpected error in rules engine

dangershony:
This will at least log out unexpected errors, but the greater question is what to do when that happens, we could for example inject the lifetime token and just shutdown the node for such errors.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
@@ -236,7 +236,7 @@ namespace Stratis.Bitcoin.Consensus
             }
             catch (Exception exception)
             {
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusRuleEngine.cs
```diff
@@ -234,6 +234,11 @@ private async Task ExecuteRulesAsync(IEnumerable<AsyncConsensusRule> asyncRules,
             {
                 ruleContext.ValidationContext.Error = ex.ConsensusError;
             }
+            catch (Exception exception)
+            {
+                this.logger.LogCritical("Not handled error in the rules engine, exception: {0}.", exception.ToString());
```
ferdeen: "Unhandled exception in Consensus rules engine : {0}", maybe  ?

dangershony: yes better

PR #2197, opened 2018-09-14
[SC] Memory tracking via RuntimeObserver

### LINKED ISSUE #1839: [SC] Track gas for memory - AKA ensure OutOfMemoryExceptions are virtually impossible.
The gas tracking code currently definitely counts instructions, and persistent storage, however we're unsure how well it currently accounts for memory. It definitely has to expend some gas to save things to memory (instruction costs) so we need to try storing large amounts of data in memory and ensure that a significant cost is incurred for doing so.

codingupastorm:
Closes #1839.

2 tests I still need to make pass - for string.join and string.concat.

Also will refactor quite a bit but thought it would be good to get any initial feedback while it's available!


src/Stratis.SmartContracts.Executor.Reflection/Contract.cs
```diff
@@ -190,7 +193,12 @@ private IContractInvocationResult InvokeInternal(MethodBase method, object[] par
             {
                 // Method threw an exception that was an OutOfGasException.
                 return ContractInvocationResult.ExecutionFailure(ContractInvocationErrorType.OutOfGas, targetException.InnerException);
-            }           
+            }
+            catch (TargetInvocationException targetException) when (targetException.InnerException is MemoryConsumptionException)
+            {
+                // Method threw an exception that was an OutOfGasException.
```
rowandh: Comment

rowandh:
Looks good! I notice that `MemoryLimitRewriter` has not actually been wired up outside of the tests. Was this intentional?

You mention future refactoring so a couple of things stood out to me which you may have already considered:
- `MemoryLimitRewriter` must be applied to a module after the ObserverRewriter. If there's an ordering dependency like this we should code the ObserverRewriter so that it's impossible to write it up incorrectly. It feels like there's two sub-components to the ObserverRewriter - gas measurement and memory limit measurement. Perhaps these two things become injected components of the ObserverRewriter, which then decides on ordering.
- Above will also help with `/// TODO: Pass around the Observer instance for performance rather than retrieving a second time.`

rowandh:
Test_MemoryLimit_BigStringConcat_Fails and Test_MemoryLimit_BigStringJoin_Fails are failing for me. In `Contract.cs` the exception being caught is a `System.OutOfMemoryException` and not a `MemoryConsumptionException`.

PR #2198, opened 2018-09-14
[SC] State transition improvements

rowandh:
* Add state snapshotting. State is 'snapshotted' and the snapshot is modified. If the modifications are successful, the current state can be updated to the snapshot.
* Add state processor, which extracts the logic for modifying state
* Add SmartContractStateFactory, which creates ISmartContractState objects
* Add unit tests
* Begin state transition specifications, which define what should happen when the state is being updated


rowandh:
SendAndReceiveSmartContractTransactionsUsingController is failing (failing in master too).

PR #2200, opened 2018-09-14
Improved shutdown performance

noescape00:
Also fixed shutdown sequence in the base feature

### DIFF
--- a/src/Stratis.Bitcoin/Base/BaseFeature.cs
+++ b/src/Stratis.Bitcoin/Base/BaseFeature.cs
@@ -282,6 +282,7 @@ namespace Stratis.Bitcoin.Base
                 this.flushChainLoop.Dispose();
             }
 
+            this.logger.LogInformation("Disposing time sync behavior.");
             this.timeSyncBehaviorState.Dispose();
 
             this.logger.LogInformation("Disposing block puller.");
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -301,7 +300,7 @@ public async Task<ChainedHeader> BlockMinedAsync(Block block)
 
         /// <summary>
         /// Called after a peer was disconnected.
-        /// Informs underlying components about the even.
+        /// Informs underlying components about the event but only if node is not being shut down at the moment.
```
ferdeen: ..event, but only if **the** node...

noescape00: fixed

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -285,24 +282,27 @@ public override void Dispose()
...
+            this.partialValidator.Dispose();
+
             this.logger.LogInformation("Disposing consensus manager.");
             this.consensusManager.Dispose();
 
             this.logger.LogInformation("Disposing consensus rules.");
             this.consensusRules.Dispose();
 
+            this.logger.LogInformation("Saving chain repository.");
+            this.chainRepository.SaveAsync(this.chain).GetAwaiter().GetResult();
```
dangershony: Ahh so this needs to be disposed after the rule ok.

PR #2201, opened 2018-09-14
Make CheckDifficultyHybridRule a full validation rule

dangershony:


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosTimeMaskRule.cs
```diff
@@ -39,6 +39,7 @@ public override Task RunAsync(RuleContext context)
             this.Logger.LogTrace("Height of block is {0}, block timestamp is {1}, previous block timestamp is {2}, block version is 0x{3:x}.", chainedHeader.Height, chainedHeader.Header.Time, chainedHeader.Previous?.Header.Time, chainedHeader.Header.Version);
 
             var posRuleContext = context as PosRuleContext;
+            posRuleContext.BlockStake = BlockStake.Load(context.ValidationContext.BlockToValidate);
```
ferdeen: Could this ever be null ?

dangershony: No because his is a partial validation rule so a block will exist

dangershony:
Some integration tests are failing, but not related to this PR.
I will check it out anyway

dangershony:
Ok this is ready to commit, two tests are failing but this is likely because of the BSQ race condition

PR #2203, opened 2018-09-14
CM bugfix

noescape00:
Bug found by @dangershony 

Bug setup: we start the node at height 100 and advance 1-2 blocks to height 102. After that we find a better chain with fork at 80 and length of 120. We rewind but we don't have blocks that are being rewinded in memory and therefore `disconnectedBlocks` list will contain nulls for blocks that are missing. 
Then we find out that that better chain is invalid and we try to reconnect old blocks. But we can't- we have nulls in the list. 

Fixed by loading from DB in case it's not found in cache.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -623,7 +623,7 @@ namespace Stratis.Bitcoin.Consensus
 
                     if (block == null)
                     {
-                        // Sanity check.
+                        // Sanity check. Block being disconnected should always be in the block store before we rewind.
                         this.logger.LogTrace("(-)[BLOCK_NOT_FOUND]");
                         throw new Exception("Block that is about to be rewinded wasn't found in cache or database!");
                     }
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -590,7 +590,7 @@ namespace Stratis.Bitcoin.Consensus
         /// <returns>List of blocks that were disconnected.</returns>
         private async Task<List<ChainedHeaderBlock>> RewindToForkPointAsync(ChainedHeader fork, ChainedHeader oldTip)
         {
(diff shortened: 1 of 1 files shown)


Aprogiena:
looks ok

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -614,7 +614,23 @@ private async Task<List<ChainedHeaderBlock>> RewindToForkPointAsync(ChainedHeade
...
 
-                var disconnectedBlock = new ChainedHeaderBlock(current.Block, current);
+                Block block = current.Block;
+
+                if (block == null)
+                {
+                    this.logger.LogTrace("Block '{0}' wasn't cached. Loading it from the database.", current.HashBlock);
+                    block = await this.blockStore.GetBlockAsync(current.HashBlock).ConfigureAwait(false);
+
+                    if (block == null)
```
dangershony: What I am not sure about here is so we really need to throw in case the block is not found.
I know we said rewind will include the block, but it does not have to and it simplifies the code greatly.
Currently the only component that depends on a block being there is mempool rewind.

noescape00: We are throwing here just because it's a sanity check and should never happen. 

However I kinda see your point so moved this sanity check right before we need those blocks so we can see a clear error in case it ever happens.

noescape00: >Currently the only component that depends on a block being there is mempool rewind.

We potentially will need blocks being disconnected for other components in the future so it's worth it to have this logic.

dangershony: No the fix you had was good, my point was about requiring the block to be present in the signaler.

noescape00:
Reverted last change. 
@Aprogiena tx

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: >my point was about requiring the block to be present in the signaler.

Hm, ok. 
I think having the block always set is a good guarantee that other components need. Otherwise this signaler is not reliable and can't be used for different things.

PR #2204, opened 2018-09-15
Register Network if passed in to NodeSettings

codingupastorm:
If a network is passed into the `NodeSettings` to be used by the node, it also needs to be registered in our list of networks so that we can create a wallet, as the `Network.Parse` method relies on `GetCandidates` using registered networks. Insanely gross, we should fix that eventually, but for now at least we can patch it here.

It's otherwise misleading for anyone else consuming our API trying to run a new network.

### DIFF
--- a/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
+++ b/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
@@ -157,6 +157,10 @@ namespace Stratis.Bitcoin.Configuration
 
                 this.Logger.LogDebug("Network set to '{0}'.", this.Network.Name);
             }
+            else
+            {
+                NetworkRegistration.Register(network);
+            }
 
             // Set the full data directory path.
             if (this.DataDir == null)
--- a/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
+++ b/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
@@ -158,8 +158,8 @@ namespace Stratis.Bitcoin.Configuration
                 this.Logger.LogDebug("Network set to '{0}'.", this.Network.Name);
             }
 
-            // Ensure the network being used is registered.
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -157,6 +157,10 @@ public class NodeSettings : IDisposable
 
                 this.Logger.LogDebug("Network set to '{0}'.", this.Network.Name);
             }
+            else
```
fassadlr: Good point @codingupastorm regarindg smart contract node startup but whilst we are here lets just do this better.

Lets go `if this.Network == null` then only create the network (i.e. we dont create and register at the same time) and after we always call `NetworkRegistration.Register(network);`

Does this make sense?

dangershony: I don't even like this ugly if condition, maybe network should just be mandatory in the constructor?

codingupastorm: Sounds good guys I'll do it as @fassadlr suggested

codingupastorm:
Ready for review again guys.

Also want to reiterate that the real issue is that the node *should not* depend on the network being set in a static class and required in things like `Network.Parse`. If we were going to fix this - that would be how.

fassadlr:
@codingupastorm can you check as there as some test failing that possibly relates to you change?

image

dangershony:
Yeah looks good now, just check the test comment.

codingupastorm:
Okay fixed - there is an issue here where `new StratisTest() == new StratisTest()` returns false. This is slightly nonsensical - if we are truly going to use inheritance for networks then we need to either enforce a singleton or override `ToEquals`.

Ideally lets not define networks by inheritance though 😕

PR #2207, opened 2018-09-16
Fix BlockStoreQueue race condition

### LINKED ISSUE #2162: Fix race condition between block announcement & block saving to disk
In `BlockStoreSignaled.OnNextCore`:

```
            this.blockStoreQueue.AddToPending(blockPair);

            if (this.initialBlockDownloadState.IsInitialBlockDownload())
            {
                this.logger.LogTrace("(-)[IBD]");
                return;
            }

            // Add to cache if not in IBD.
            this.blockStoreCache.AddToCache(block);

            this.logger.LogTrace("Block header '{0}' added to the announce queue.", chainedHeader);
            this.blocksToAnnounce.Enqueue(chainedHeader);
```

It is possible for the `blocksToAnnounce` queue to get processed quickly enough that a peer node requests the announced block while the block is still in the `blockStoreQueue`. In the background `BlockStoreQueue.DequeueBlocksContinuouslyAsync` is grabbing blocks off the aforesaid queue and putting them into the batch to be saved.

However, the GetBlockAsync method expects the block to either be in the batch or in the repository, nowhere else:

```
        public async Task<Block> GetBlockAsync(uint256 blockHash)
        {
            this.logger.LogTrace("({0}:'{1}')", nameof(blockHash), blockHash);

            Block block = null;

            lock (this.getBlockLock)
            {
                block = this.batch.FirstOrDefault(x => x.ChainedHeader.HashBlock == blockHash)?.Block;
            }

            if (block == null)
                block = await this.blockRepository.GetBlockAsync(blockHash).ConfigureAwait(false);
            else
                this.logger.LogTrace("Block was found in the batch.");

            this.logger.LogTrace("(-)");
            return block;
        }
```

So the block doesn't get found and the peer waits its timeout interval to request the block again. This is bad if our node mined the block that is getting announced, as it introduces substantial block propagation latency.

zeptin:
Fixes #2162.

Removing the queue that was the source of the race condition seemed like the most straightforward approach.

This impacted the loop that was previously responsible for both dequeuing blocks & checking whether the 37 second timer elapsed. It now only needs to check the timer in addition to the (existing) batch size & consensus tip checks.

This change necessitated altering how the batch gets cleared. Previously blocks would be in the batch, a save condition would be triggered, and the chain without reorged blocks would be determined & written to the repository. The batch would then be cleared.

Since the batch now gets directly added to, the saved elements have to be removed from the batch under a lock. The save method therefore returns a list of blocks that got saved to facilitate this. Note that we do not perform the actual save under a lock, as this would stop consensus while it waits to acquire the lock in the `AddToPending` method.

The removal of reorged blocks also needs to be done more carefully - previously the code relied on the batch being cleared to ensure that the reorged blocks were removed from the batch. We now do this more explicitly by replacing the batch with the cleaned version.

Some minor changes to the block store tests were necessary, as they were originally written with a queue in mind.


noescape00:
will review tmr

src/Stratis.Bitcoin/Interfaces/IblockStoreQueue.cs
```diff
@@ -4,8 +4,8 @@ namespace Stratis.Bitcoin.Interfaces
 {
     public interface IBlockStoreQueue : IBlockStore
     {
```
Aprogiena: Please rename the file to have B

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -271,35 +267,40 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
...
-
-                // Wait for one of the tasks: dequeue or timer (if available) to finish.
-                Task task = (timerTask == null) ? dequeueTask : await Task.WhenAny(dequeueTask, timerTask).ConfigureAwait(false);
-
-                bool saveBatch = false;
-
                 try
                 {
-                    await task.ConfigureAwait(false);
+                    await Task.Delay(1000);
```
Aprogiena: Remove this active waiting

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -271,35 +267,40 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
...
+
+                    cumulativeTicks++;
+
+                    if (cumulativeTicks >= BatchMaxSaveIntervalSeconds)
+                    {
+                        saveBatch = true;
+                        cumulativeTicks = 0;
```
noescape00: reset this when the batch was saved. If you do it here cumulativeTicks won't be reset in case we saved the batch because it was too big and not because of the timer. That will lead to additional save when we don't need it.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -310,50 +311,27 @@ private async Task DequeueBlocksContinuouslyAsync()
...
                     {
-                        await this.SaveBatchAsync().ConfigureAwait(false);
+                        List<ChainedHeaderBlock> saved = await this.SaveBatchAsync().ConfigureAwait(false);
 
                         lock (this.getBlockLock)
                         {
-                            this.batch.Clear();
-                        }
+                            // Only remove items from the batch that have explicitly been saved.
+                            foreach (ChainedHeaderBlock savedHeaderBlock in saved)
```
noescape00: this will lead to race condition when we would remove stuff from a batch that we need. 
in case we are connecting a chain and then disconnecting it because it's invalid and reconnecting old one it's possible that we will have 2 instances of same block in the batch but it will be removed here. 
That will lead to the absence in the block which is a part of the best chain

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -367,11 +345,25 @@ private async Task DequeueBlocksContinuouslyAsync()
...
-        private async Task SaveBatchAsync()
+        /// <returns>List of saved blocks.</returns>
+        private async Task<List<ChainedHeaderBlock>> SaveBatchAsync()
         {
             this.logger.LogTrace("()");
 
-            List<ChainedHeaderBlock> clearedBatch = this.GetBatchWithoutReorgedBlocks();
+            List<ChainedHeaderBlock> clearedBatch;
+
+            lock (this.getBlockLock)
```
noescape00: it's not a get block lock anymore. please rename

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
zeptin: If so, it seems to me it was a pre-existing race condition. The previous implementation would clear the entire batch (which at that point would contain reorged blocks) after saving.

I can try rectify your concern as part of this PR if you want, or I can do it as part of implementing additional integration tests. This seems like a good test candidate.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -285,32 +290,39 @@ private async Task CheckBatchContinuouslyAsync()
...
-                    await Task.Delay(1000);
-
-                    cumulativeTicks++;
-
-                    if (cumulativeTicks >= BatchMaxSaveIntervalSeconds)
+                    using (CancellationTokenSource timeoutTokenSource = CancellationTokenSource.CreateLinkedTokenSource(this.nodeLifetime.ApplicationStopping))
                     {
-                        saveBatch = true;
-                        cumulativeTicks = 0;
+                        timeoutTokenSource.CancelAfter(BatchMaxSaveIntervalSeconds * 1000);
```
Aprogiena: Why use cancellation token source instead of timer? Either use timer or a separated task which completes after certain time

zeptin: The previous implementation used `Task.Delay` as a timer anyway, which you wanted changed when I used it in the new implementation. I don't see using cancellation as a timer like this as being particularly problematic, we do the same thing in the P2P code, for example.

Aprogiena: that feels very heavy. Task.Delay is quite slim compared to that. in P2P there is a onetime thing, this is repeated timer, so it feels wrong

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -271,89 +275,77 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
...
-                // or the max batch size is reached or the node is shutting down.
-                if (dequeueTask.Status == TaskStatus.RanToCompletion)
-                {
-                    ChainedHeaderBlock item = dequeueTask.Result;
-
-                    // Set the dequeue task to null so it can be assigned on the next iteration.
-                    dequeueTask = null;
-
-                    lock (this.getBlockLock)
+                    using (CancellationTokenSource timeoutTokenSource = CancellationTokenSource.CreateLinkedTokenSource(this.nodeLifetime.ApplicationStopping))
```
dangershony: Wont this be much simpler by just using an the AsyncLoop class and iterate every second on the batch condition?
You guys are complicating things unnecessarily.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -271,89 +275,77 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
...
+                // Apart from node shutdown, we also trigger a save if the batch is sufficiently full,
+                // or we are at the consensus tip.
+                saveBatch = saveBatch || (this.currentBatchSizeBytes >= BatchThresholdSizeBytes) || this.chainState.IsAtBestChainTip;
+
                 if (saveBatch)
                 {
                     if (this.batch.Count != 0)
                     {
-                        await this.SaveBatchAsync().ConfigureAwait(false);
+                        List<ChainedHeaderBlock> saved = await this.SaveBatchAsync().ConfigureAwait(false);
```
dangershony: `this.batch.Count` should probably be inside the lock no?

zeptin: Perhaps. But all we really care about is that the count is nonzero. Since this is the only thread that actually removes items from the batch, it should be safe.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -367,11 +359,25 @@ private async Task DequeueBlocksContinuouslyAsync()
...
+            List<ChainedHeaderBlock> clearedBatch;
+
+            lock (this.batchLock)
+            {
+                clearedBatch = this.GetBatchWithoutReorgedBlocks();
+
+                // We do not want reorged blocks to be left inside the batch once saving is complete.
+                // Therefore replace the batch with the cleaned version.
+                this.batch.Clear();
+                this.batch.AddRange(clearedBatch);
```
dangershony: Why push blocks back to the batch?

dangershony:
I suggest this flow, we only need to hold the lock inside the mthod `bla`


So you can do this
```
   private List<ChainedHeaderBlock> GetBatchWithoutReorgedBlocks()
        {
            this.logger.LogTrace("()");

 lock() - just lock here
 {
            // Initialize current with highest block from the batch.
            ChainedHeaderBlock current = this.batch.Last();

            // List of consecutive blocks. It's a cleaned out version of batch that doesn't have blocks that were reorged.
            var batchCleared = new List<ChainedHeaderBlock>(this.batch.Count) { current };

            // Select only those blocks that were not reorged away.
            for (int i = this.batch.Count - 2; i >= 0; i--)
            {
                if (this.batch[i].ChainedHeader.HashBlock != current.ChainedHeader.Previous.HashBlock)
                {
                    this.logger.LogDebug("Block '{0}' removed from the batch because it was reorged.", this.batch[i].ChainedHeader);
                    continue;
                }

                batchCleared.Add(this.batch[i]);
                current = this.batch[i];
            }
this.batch.Clear()
}
            batchCleared.Reverse();

            this.logger.LogTrace("(-):*.{0}={1}", nameof(batchCleared.Count), batchCleared.Count);
            return batchCleared;
        }
```

Then you can remove the lock from the outside method

dangershony:
>pls no

Why not? what's the reason behind not using a loop?

Aprogiena:
active waiting is not a way to go, we try to remove it everywhere and now you suggest to introduce it again? this is supposed to be event driven - someone adds to a queue, something happens, timer expires, something happens, there is no loop

dangershony:
If you have a timer then you have active waiting either way. 

>we try to remove it everywhere and now you suggest to introduce it again?
I was not aware if that 🤔 

If we want this logic to be evenet driven then we need to bring back the AsyncQueue but we will still need a timer regardless (so then you end up with some weird hack of waiting on a task completion to achieve a time etc...)

If we dont have an AsyncTimerQueue (wich we plan to implemet) then use an AsyncQueue and an AsyncLoop together and not do weired hacks.

What if we create an AsymcQueue that can be triggered with two types of messages (a block or a timer ping) then you still get a no-lock approach and the code is way simpler.

The result of the AsyncQueue will always push to the batch.

We still need a lock though arount he batch because of peers asking for pending blocks.

src/Stratis.Bitcoin/Consensus/IConsensusManager.cs
```diff
@@ -59,6 +59,10 @@ public interface IConsensusManager : IDisposable
         /// <param name="onBlockDownloadedCallback">The callback that will be called for each downloaded block.</param>
         Task GetOrDownloadBlocksAsync(List<uint256> blockHashes, OnBlockDownloadedCallback onBlockDownloadedCallback);
 
+        /// <summary>Loads the block data from <see cref="chainedHeaderTree"/> or block store if it's enabled.</summary>
+        /// <param name="blockHash">The block hash.</param>
+        Task<ChainedHeaderBlock> LoadBlockDataAsync(uint256 blockHash);
```
dangershony: Did we not say we will rename this to GetBlockData?

dangershony:
This looks much better, all integration tests passing?

zeptin:
There are a handful failing, but they do not seem to be related to this PR.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -75,17 +78,19 @@ public class BlockStoreBehavior : NetworkPeerBehavior, IBlockStoreBehavior
         /// <inheritdoc cref="IChainState"/>
         private readonly IChainState chainState;
 
-        public BlockStoreBehavior(ConcurrentChain chain, IBlockStore blockStore, IChainState chainState, ILoggerFactory loggerFactory)
+        public BlockStoreBehavior(ConcurrentChain chain, IBlockStore blockStore, IChainState chainState, ILoggerFactory loggerFactory, IConsensusManager consensusManager)
         {
             Guard.NotNull(chain, nameof(chain));
             Guard.NotNull(blockStore, nameof(blockStore));
```
noescape00: BlockStore now not required for the BlockStoreBehavior, please remove it.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -317,14 +322,14 @@ private async Task ProcessGetDataAsync(INetworkPeer peer, GetDataPayload getData
             foreach (InventoryVector item in getDataPayload.Inventory.Where(inv => inv.Type.HasFlag(InventoryType.MSG_BLOCK)))
             {
                 // TODO: check if we need to add support for "not found"
-                Block block = await this.blockStore.GetBlockAsync(item.Hash).ConfigureAwait(false);
+                ChainedHeaderBlock chainedHeaderBlock = await this.consensusManager.GetBlockDataAsync(item.Hash).ConfigureAwait(false);
 
-                if (block != null)
+                if (chainedHeaderBlock.Block != null)
```
noescape00: chainedHeaderBlock can be null here in case peer asked for a bogus block. That will crash the node. Please add support for null as it says in the todo above

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -74,6 +74,9 @@ public class BlockStoreQueue : IBlockStoreQueue
         /// <summary>Protects the batch from being modifying while <see cref="GetBlockAsync"/> method is using the batch.</summary>
         private readonly object getBlockLock;
 
+        /// <summary>Represents all blocks currently in the queue & pending batch, so that <see cref="GetBlockAsync"/> is able to return a value directly after enqueuing.</summary>
+        private Dictionary<uint256, ChainedHeaderBlock> blocksBeingPersisted;
```
noescape00: rename to blocksBeingProcessed. Because not all the blocks in the batch will be persisted. Sometimes we can have a reorg inside a batch so some of the blocks will be removed without even touching the disk

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -181,17 +184,21 @@ public async Task<Block> GetBlockAsync(uint256 blockHash)
...
+                    this.logger.LogTrace("(-)[FOUND_IN_DICTIONARY]");
+                    return chainedHeaderBlock.Block;
+                }
             }
 
+            Block block = await this.blockRepository.GetBlockAsync(blockHash).ConfigureAwait(false);
+
             if (block == null)
-                block = await this.blockRepository.GetBlockAsync(blockHash).ConfigureAwait(false);
+                this.logger.LogTrace("Block was not found in the repository.");
```
noescape00: those 4 lines can be replaced with 
`this.logger.LogTrace("Block was{0} found in the repository.", (block == null)? " not" : "");`

noescape00:
@zeptin after you apply changes please confirm that unit and integration tests are passing.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
dangershony: Ahh good catch

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
dangershony: Perhaps in that case just call it `pendingBlocks`

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
dangershony: @zeptin please also fix this comment

dangershony: The todo above does not apply to missing blocks but rather to investigating if a response to the peer is needed.
So just check that also  `chainedHeaderBlock ` is not null probably just do
```
if (chainedHeaderBlock?.Block != null)

```

noescape00:
@zeptin checked again. Looks really good, just fix this: 

```
// TODO: check if we need to add support for "not found"
ChainedHeaderBlock chainedHeaderBlock = await this.consensusManager.GetBlockDataAsync(item.Hash).ConfigureAwait(false);

if (chainedHeaderBlock.Block != null)
```

dangershony:
I think there is a problem with the logic here please do not merge yet

dangershony:
ok the problem was a known bug in master this is good to go in

Issue #2087, opened 2018-08-31
Feature initialization- change to InitializeAsync


noescape00:
Currently we don't support async on features initialization. All the initialization code is basically `GetAwaiter.GetResult()`. Initialize code should support async.

PR #2215, opened 2018-09-17
Feature.InitializeAsync & logging & PoSMinting don't use getawaiter.getResult

noescape00:
Fix: #2087

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using NBitcoin;
 using NBitcoin.Rules;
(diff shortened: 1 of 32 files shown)


src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -58,13 +59,15 @@ public class ConsensusFeature : FullNodeFeature
         }
 
         /// <inheritdoc />
-        public override void Initialize()
+        public override Task InitializeAsync()
```
Aprogiena: Why no logs here?

src/Stratis.Bitcoin.Features.LightWallet/LightWalletFeature.cs
```diff
@@ -123,7 +123,7 @@ public static void PrintHelp(Network network)
         }
 
         /// <inheritdoc />
-        public override void Initialize()
+        public override Task InitializeAsync()
```
Aprogiena: Why no logs here?

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
noescape00: no logger. 
decided to avoid touching too much- added logging only where the logger was already there

src/Stratis.Bitcoin.Features.LightWallet/LightWalletFeature.cs
noescape00: Same reason as above.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
noescape00: This PR already touches 32 files so probably better will be to introduce logging to classes that lack it in another PRs

fassadlr:
@noescape00 I'm approving but if these tests are related to your work then fix them otherwise merge when ready.

image

noescape00:
Those are just generally unstable. 
I'll rerun the tests on my machine just in case and then merge.

PR #2216, opened 2018-09-17
Fix partial validation sequence when a fork occurs

### LINKED ISSUE #2153: [Activation] Node does not advance after failed Coinstake signature verification

fassadlr:
**Review but DONT MERGE until BSQ issue with persisting has been resolved.**

Fixes #2153 

When a fork occurs we don't want to continue full validation of each block on the forked chain until the proof of work is higher of a block on the fork. 

This will stop until the fork is 500 blocks from the fork point and the MaxReorgViolation is thrown.

There are some integration test changes in this PR which was solely put there to help write the integration tests.


fassadlr:
PS this PR will only apply once the BSQ issue has been resolved that @zeptin is working on.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -464,7 +463,7 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
                 }
             }
 
-            if (chainedHeaderBlocksToValidate != null)
+            if (fullValidationRequired && chainedHeaderBlocksToValidate != null)
```
Aprogiena: This condition should have been extended with check that the full validation succeeded. So, if there was no full validation, partial validation should continue. And if there was full validation, partial validation should continue only if full validation succeeded.

fassadlr: Ok thankals @aprogiena. That makes it more clear. I think im going to get FullyValidateAsync to return a result class for better feedback. Let me check what I can do.

Aprogiena: all you need is 
```
bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
bool fullValidationFailed = (connectBlocksResult != null) && connectBlocksResult.Succeeded;
if (nextValidationNeeded && !fullValidationFailed)
```
or something like that

src/NBitcoin/Consensus.cs
```diff
@@ -23,7 +23,7 @@ public class Consensus : IConsensus
         public Money ProofOfStakeReward { get; }
 
         /// <inheritdoc />
-        public uint MaxReorgLength { get; }
+        public uint MaxReorgLength { get; private set; }
```
fenix2222: Isn't `private set;` redundant in latest c#?

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPosRunner.cs
```diff
@@ -14,15 +14,23 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.Runners
...
+            this.agentPrefix = agentPrefix;
             this.Network = network;
         }
 
         public override void BuildNode()
         {
-            var settings = new NodeSettings(this.Network, ProtocolVersion.ALT_PROTOCOL_VERSION, args: new string[] { "-conf=stratis.conf", "-datadir=" + this.DataFolder });
+            NodeSettings settings;
+
+            if (this.agentPrefix == string.Empty)
```
fenix2222: Should this be `.IsNullOrEmpty()` or `.IsNullOrWhitespace()`?

src/NBitcoin/Consensus.cs
dangershony: Why do you need this change at all? (are you reflection changing MaxReorgLength ?)

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        private const string walletName = "mywallet";
+        private const string walletPassword = "123456";
+
+        public ConsensusManagerTests()
+        {
+            this.network = new StratisRegTest();
+            this.network.Consensus.Options = new ConsensusOptionsTest();
+
+            Type consensusType = typeof(NBitcoin.Consensus);
+            consensusType.GetProperty("MaxReorgLength").SetValue(this.network.Consensus, (uint)20);
```
dangershony: I you set this here I see ok

src/NBitcoin/Consensus.cs
Aprogiena: even if it is redundant, it is better to see it explicitly

fassadlr:
OK @Aprogiena and @dangershony please review again.

I had to refactor `OnPartialValidationSucceededAsync` so that we work with a `PartialValidationSucceededResult` to determine what should happen next. Also, the logic is now a lot tighter so that it reads and flows a lot better (which plays a big role in why these bugs keep on creeping in). Once you have approved I will fix the unit tests (which will be broker at the moment).

I synced the node past block 488095 whilst connected to the un-upgraded stratis-x core and still successfully threw the `MaxReorgViolationExpception` which caused the Stratis-X node to be disconnected correctly. The node then proceeded to sync further.

There is however an issue in CHT with reconnecting the old chain after the fork got disconnected where block 448096 does not yet have the block loaded (which I think is correct). It was throwing a null reference when it tries to calculate the `UnconsumedBlocks`.

Aprogiena:
pls only address 1 issue at the time, i can't see why not just do those 3 lines of code mentioned previously

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -446,8 +448,12 @@ public List<int> ConsensusTipChanged(ChainedHeader newConsensusTip)
             // Consider blocks that became a part of our best chain as consumed.
             while (currentHeader != fork)
             {
-                this.UnconsumedBlocksDataBytes -= currentHeader.Block.BlockSize.Value;
-                this.logger.LogTrace("Size of unconsumed block data is decreased by {0}, new value is {1}.", currentHeader.Block.BlockSize.Value, this.UnconsumedBlocksDataBytes);
+                if (currentHeader.BlockDataAvailability == BlockDataAvailabilityState.BlockAvailable)
```
fassadlr: @noescape00 @Aprogiena @dangershony Can you please comment on this? I ran into a null reference for `currentHeader.Block`  for the block just after the fork point. It was full validated but the block was null and `BlockDataAvailabilityState` was `HeaderOnly`.

Aprogiena: pls create issue for that

fassadlr: This fixes the issue? Or do you just want to have it for our records?

Aprogiena:
Could you please convince me that the fix I suggest of changing just three lines of code would not work?

fassadlr:
> Could you please convince me that the fix I suggest of changing just three lines of code would not work?

Can you convince me that the logic in that method now is not better?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -414,67 +414,89 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
 
-                    foreach (INetworkPeer peer in peersToBan)
-                        this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, connectBlocksResult.BanDurationSeconds, connectBlocksResult.BanReason);
-                }
+        /// <summary>
+        /// Fully validates the chained header and should it fail, ban peers that has this header.
+        /// </summary>
+        /// <param name="chainedHeader">The chained header to validate.</param>
+        /// <returns>Validation related information.</returns>
+        private async Task<ConnectBlocksResult> FullyValidateAsync(ChainedHeader chainedHeader)
```
fassadlr: @Aprogiena this is just a copy paste of the peer banning code in the previous version of the code. No logic changes.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: Do you have a way to reproduce this constantly?
You are fixing more then one issue in one PR this makes it harder to be sure a fix is correct, or if a fix needs to be rolled back etc...

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -325,25 +325,25 @@ public void FullValidationSucceeded(ChainedHeader chainedHeader)
         }
 
         /// <inheritdoc />
-        public List<ChainedHeaderBlock> PartialValidationSucceeded(ChainedHeader chainedHeader, out bool fullValidationRequired)
+        public PartialValidationSucceededResult PartialValidationSucceeded(ChainedHeader chainedHeader)
```
dangershony: Perhaps this is a good change but it could be done in a pure env and not as part of an important fix

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
fassadlr: @dangershony I could not advance after this without this fix, this is why I asked your thoughts on this as I am unsure why `block` would be null here.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
fassadlr: I don't understand why, if improvements are introduced which is only going to help us debug the problem faster, they are frowned upon and asked not to be included in a PR. To create a new PR and go through the whole build and testing process will just waste a lot of time.

I really don't agree with this way of writing code.

If you can honestly tell me that the new code is worse then sure, no problem but just dismissing code because its slightly more difficult to review is not good enough reason.

noescape00:
Won't just
`if ((chainedHeaderBlocksToValidate != null) && ((connectBlocksResult == null) || connectBlocksResult.Succeeded))`

instead of 

`if (chainedHeaderBlocksToValidate != null)` 

on line 467 fix the bug?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
dangershony: I am not sure myself I will need to reproduce somehow, I got the same bug but after a restart I could continue to sync

Aprogiena: This is much more complicated, because this should never happen so that is why we believe it only happens because of the first bug. If that is the case, there is nothing to fix here. If that is not the case, it needs much deeper look, because this does not fix it. But we first need to fix the first bug and then to see if there is another one.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: You need to add `Locked` on this method as its under a lock.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -414,67 +414,89 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
 
-                this.logger.LogTrace("Full validation is{0} required.", fullValidationRequired ? "" : " NOT");
+                this.logger.LogTrace("Full validation is{0} required.", partialResult.FullValidationRequired ? string.Empty : " NOT");
 
-                if (fullValidationRequired)
+                if (partialResult.FullValidationRequired)
                 {
-                    connectBlocksResult = await this.FullyValidateLockedAsync(chainedHeader).ConfigureAwait(false);
+                    fullResult = await this.FullyValidateAsync(chainedHeader).ConfigureAwait(false);
+                    shouldPartialValidationContinue = fullResult.Succeeded;
```
dangershony: Do you need this boolean? `shouldPartialValidationContinue ` the result is anyway available in `fullResult.Succeeded`

Aprogiena:
> Can you convince me that the logic in that method now is not better?

That is a subjective metric.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        private class ConsensusOptionsTest : PosConsensusOptions
+        {
+            public override int GetStakeMinConfirmations(int height, Network network)
+            {
+                return height < 10 ? 5 : 10;
+            }
+        }
+
+        [Fact]
+        public void ConsensusManager_Fork_Scenario_1()
```
dangershony: ConsensusManager_Fork_Scenario_1 is referring to the test scenario?
Better is to describe the test in the method name and in the comment write the scenario number (as we did with CHT)

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -296,7 +296,7 @@ private long GetTotalSpeedOfAllPeersBytesPerSec()
         {
             lock (this.peerLock)
             {
-                return this.pullerBehaviorsByPeerId.Sum(x => x.Value.SpeedBytesPerSecond);
+                return this.pullerBehaviorsByPeerId.Where(p => p.Value.SpeedBytesPerSecond > 0).Sum(x => x.Value.SpeedBytesPerSecond);
```
dangershony: Is this really faster?

Aprogiena: more like ... is this really related to this PR?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: i can't find anything relevant to it, can you link or copy paste please?

Aprogiena: or do you mean https://github.com/stratisproject/StratisBitcoinFullNode/pull/2216#issuecomment-422406237 ?

Aprogiena: if yes, that should not happen and if it does we should investigate separately, that is what i'm explaining above

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: We only check the full validation result if it is required. Please see the comment a little bit further down which explains the logic from @aprogiena.

"So, if there was no full validation, partial validation should continue. And if there was full validation, partial validation should continue only if full validation succeeded."

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
fassadlr: This fix I can revert.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
fassadlr: Will update :)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -414,67 +414,89 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
 
             using (await this.reorgLock.LockAsync().ConfigureAwait(false))
             {
-                bool fullValidationRequired;
-
                 lock (this.peerLock)
                 {
-                    chainedHeaderBlocksToValidate = this.chainedHeaderTree.PartialValidationSucceeded(chainedHeader, out fullValidationRequired);
+                    partialResult = this.chainedHeaderTree.PartialValidationSucceeded(chainedHeader);
+                    if (partialResult.ChainedHeaderBlocksToValidate == null)
```
dangershony: if `partialResult.ChainedHeaderBlocksToValidate == null` we may still need to fully validate so returning here seems wrong

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: i disagree with your disagreement. this is critical fix, it  should be as small as possible, not touching ANYTHING else, no agent prefix, no block puller, no CHT, NOTHING. this only makes it harder for everyone to review this and i have asked multiple times not to do this exact thing to include unrelated code in PR, especially if it is bug fixing PR. yet you ignored it again :( that is why i feel i need to block it until everything unrelated is out of it

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
Aprogiena: it seems wrong (though i believe it is not, just seemingly), that's why i think the new code is actually worse. if you spend 5 minutes on it (which you should not), you will see that it is probably correct as this has a special meaning. It's comment says 

```
/// <summary>List of chained header blocks with block data that should be partially validated next. Or <c>null</c> if none should be validated.</summary>
``` 

but it is hardly obvious from that that not even the current block should not be validated (that it implies that full validation is not needed.

dangershony: But this is a classic case of the tip, a new block arrived it needs to be fully validated but no more blocks after that require partial validation (and this is similar to the way it worked before)

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
fassadlr: @Aprogiena I disagree, any changes realting to the bug and slight improvements should be regarded and reviewed on merit. 

Granted, the other fixes to BlockPuller shouldn't have been in here as well as the "supposed" fix for the `UnconsumedBytes` null reference, although if I was the reviewer here I wouldn't have had an issue with this. A simple message to me asking why the change would've explained it.

Changes to the tests were there for me to better understand which nodes are syncing and when, hence the agent prefix improvement. Again, a simple "why was this changed" would have sufficed.

Apart from the BlockPuller change, ALL of these relates to this ONE issue @Aprogiena and as such they are good and relevant.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: OK will fix now 👍

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
Aprogiena: i know, but the new logic is that if CHBTV is null it means no FV, but if it is not null, just empty list, then you may do FV. you don't need to convince me that it smells though :)

fassadlr: @Aprogiena ah ok now it makes sense. Let me fix this.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
Aprogiena: I did ask you that question and you replied and you said

```
I had to refactor OnPartialValidationSucceededAsync so that we work with a PartialValidationSucceededResult to determine what should happen next. 
```

which is wrong because that was not necessary (as my suggestion of fix proves). Therefore your approach does not stand. Also you make arbitrary new allocation which is not needed and then you have a hidden functionality based on internal property being null or just empty, which is quite smelly approach. Then you violate logging policies, but that's just tiny little thing at the end. 

Changes in tests that are not required for the new test you implemented don't belong here as well. 

So the whole change is based on wrong reasoning based on you not understanding the code.  That is not a reason that has merit and does not justify the change.

So I repeat, this is important bugfix, it should be minimal, implemented with at most 3 line change + comment + test. Nothing else.

You can't argue that any change that is good belongs everywhere. That's total nonsense, because you would suddenly combine everything everywhere. This is such a waste of time and I am asking you again - please don't do it.

fassadlr: - which is wrong because that was not necessary (as my suggestion of fix proves).

Why is returning a result class wrong? Just because you say so? Working with result class is more intuitive than calling a Validate method and returning a set of headers to validate. In no world is this something a new dev would understand as normally calling a validate method should return a validation result (and not some `out` parameter). We are working with C# here you know, we are allowed to make cahnges to make our life easier.

- hidden functionality based on internal property being null or just empty, which is quite smelly approach

What hidden functionality? You say this is a waste of time but again (and just like the last time you just dismissed my PR) you are just commenting without any constructive criticism or actually telling me what is wrong with the code. This is very unprofessional and not helping the cause. 

- Then you violate logging policies, but that's just tiny little thing at the end.

Again, why cant you just tell me what was being violated?

- So the whole change is based on wrong reasoning based on you not understanding the code. That is not a reason that has merit and does not justify the change.

This is **exactly** a reason to change the code, I am an experience developer and have been here just as long as you and if even I cant understand the code then surely a new dev won't so why not improve on it a bit here and now?

- You can't argue that any change that is good belongs everywhere. That's total nonsense, because you would suddenly combine everything everywhere. 

I didn't say that, the changes I made was for the betterment of the code and for me (and future devs) to understand what was done better.

It pains me that we are in a place where code cant be reviewed on merit. If you actually spent 5-10minutes reviewing the code (3 hours ago) and making constructive comments on them we would have had this PR merged by now and everyone would have been able to move along.

You can block this, but you are now holding up everyone with this.

Aprogiena: 1) three people are telling you that simple fix is better
2) your code is confusing (https://github.com/stratisproject/StratisBitcoinFullNode/pull/2216#discussion_r218489274)
3) there exists trivial solution that fixes the problem

just this should be enough, if it is not, then i can't do anything, we have simply too different approaches and are not compatible

PR #2223, opened 2018-09-19
Fix partial validation sequence when a fork occurs

fassadlr:
Slimmed down version #2216.

@noescape00 after this change I get this null reference in the block puller when calculating speed? Without that where clause one of the peers has speed of -173382348 something like that. Maybe you can shed some light on this?

### DIFF
--- a/src/NBitcoin/Consensus.cs
+++ b/src/NBitcoin/Consensus.cs
@@ -23,7 +23,7 @@ namespace NBitcoin
         public Money ProofOfStakeReward { get; }
 
         /// <inheritdoc />
-        public uint MaxReorgLength { get; }
+        public uint MaxReorgLength { get; private set; }
 
         /// <inheritdoc />
         public long MaxMoney { get; }
--- a/src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using NBitcoin;
+using Stratis.Bitcoin.IntegrationTests.Common;
+using Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers;
+using Stratis.Bitcoin.Networks;
+using Xunit;
+
+namespace Stratis.Bitcoin.IntegrationTests
+{
+    public class ConsensusManagerTests
+    {
+        private readonly Network network;
+
+        private const string walletAccount = "account 0";
+        private const string walletName = "mywallet";
+        private const string walletPassword = "123456";
+
+        public ConsensusManagerTests()
+        {
+            this.network = new StratisRegTest();
+            this.network.Consensus.Options = new ConsensusOptionsTest();
+
+            Type consensusType = typeof(NBitcoin.Consensus);
+            consensusType.GetProperty("MaxReorgLength").SetValue(this.network.Consensus, (uint)20);
+        }
+
+        private class ConsensusOptionsTest : PosConsensusOptions
+        {
+            public override int GetStakeMinConfirmations(int height, Network network)
+            {
+                return height < 10 ? 5 : 10;
+            }
+        }
+
+        [Fact]
+        public void ConsensusManager_Fork_Occurs_Node_Gets_Disconnected_Due_To_MaxReorgViolation()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create(this))
+            {
+                var minerA = builder.CreateStratisPosNode(this.network);
+                var minerB = builder.CreateStratisPosNode(this.network);
+                var syncer = builder.CreateStratisPosNode(this.network);
+
+                builder.StartAll();
+
+                minerA.NotInIBD().WithWallet();
+                minerB.NotInIBD().WithWallet();
+                syncer.NotInIBD();
+
+                // MinerA mines to height 10.
+                TestHelper.MineBlocks(minerA, walletName, walletPassword, walletAccount, 10);
+
+                // Sync the network to height 10.
+                TestHelper.Connect(syncer, minerA);
+                TestHelper.Connect(syncer, minerB);
+                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(syncer, minerA));
+                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(syncer, minerB));
+
+                // Disconnect Miner A and B.
+                TestHelper.Disconnect(syncer, minerA);
+                TestHelper.Disconnect(syncer, minerB);
+
+                // Ensure syncer does not have any connections.
+                TestHelper.WaitLoop(() => !TestHelper.IsNodeConnected(syncer));
+
+                // Miner A continues to mine to height 20 whilst disconnected.
+                TestHelper.MineBlocks(minerA, walletName, walletPassword, walletAccount, 10);
+
+                // Miner B continues to mine to height 14 whilst disconnected.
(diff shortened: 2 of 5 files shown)


fassadlr:
I have retested this scenario with the un-upgraded Stratis-X node and the node still advances.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
```diff
@@ -296,7 +296,7 @@ private long GetTotalSpeedOfAllPeersBytesPerSec()
         {
             lock (this.peerLock)
             {
-                return this.pullerBehaviorsByPeerId.Sum(x => x.Value.SpeedBytesPerSecond);
+                return this.pullerBehaviorsByPeerId.Where(p => p.Value.SpeedBytesPerSecond > 0).Sum(x => x.Value.SpeedBytesPerSecond);
```
dangershony: @noescape00 can you comment on this?

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            Type consensusType = typeof(NBitcoin.Consensus);
+            consensusType.GetProperty("MaxReorgLength").SetValue(this.network.Consensus, (uint)20);
+        }
+
+        private class ConsensusOptionsTest : PosConsensusOptions
+        {
+            public override int GetStakeMinConfirmations(int height, Network network)
+            {
+                return height < 10 ? 5 : 10;
```
dangershony: Why this condition?
Under 10 min confirmations is only 5, can you explain more?

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                // Miner B continues to mine to height 14 whilst disconnected.
+                TestHelper.MineBlocks(minerB, walletName, walletPassword, walletAccount, 4);
+
+                // Syncer now connects to both miners.
+                TestHelper.Connect(syncer, minerA);
+                TestHelper.Connect(syncer, minerB);
+
+                // Ensure that Syncer has synced with Miner A which is the node with the best tip.
+                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(syncer, minerA));
```
dangershony: I think this is wrong, Max reorg is 20 but it seems to me the reorg size is 4 so both nodes should move to the new chain if you wait enough time.

Whats odd here is why miner B wont reorg away.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -464,7 +464,9 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
                 }
             }
 
-            if (chainedHeaderBlocksToValidate != null)
+            bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
+            bool fullValidationFailed = (connectBlocksResult != null) && !connectBlocksResult.Succeeded;
+            if (nextValidationNeeded && !fullValidationFailed)
```
dangershony: Here I think its more readable but no big deal.
`   if (fullValidationSuccess && nextValidationNeeded)`

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
fassadlr: @dangershony I have checked this again and the re-orged does in fact happen BUT this is only with BlockStoreQueue constantly pushing to disk. So we will need to wait for @noescape00 to make those changes before this test really becomes relevant.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
fassadlr: This was to simulate a block validation failure so that we can eventually disconnect it due to a `MaxReorgViolation`.  I'm investigating further.

fassadlr: OK so this doesnt actually get triggered in this integration test as Im not actually working with stake blocks.

I'm going to create a second test to do that.

src/NBitcoin/Consensus.cs
```diff
@@ -23,7 +23,7 @@ public class Consensus : IConsensus
         public Money ProofOfStakeReward { get; }
 
         /// <inheritdoc />
-        public uint MaxReorgLength { get; }
+        public uint MaxReorgLength { get; private set; }
```
noescape00: `private set;` is redundant and inconsistent with other fields in the class.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
noescape00: this change is redundant. 

SpeedBytesPerSecond can't and shouldn't be negative, only >= 0. 

So there is no difference between summing collection of: {0, 5, 8, 0, 0, 3} and collection of {5, 8, 3} .
Please revert the change.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -464,7 +464,9 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
                 }
             }
 
-            if (chainedHeaderBlocksToValidate != null)
+            bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
+            bool fullValidationFailed = (connectBlocksResult != null) && !connectBlocksResult.Succeeded;
```
noescape00: I'd add a comment saying that in case `connectBlocksResult` is null it means that full validation didn't happen

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
fassadlr: @noescape00 please read the PR description on this.

src/NBitcoin/Consensus.cs
fassadlr: We need this to set values on Consensus via reflection.

src/Stratis.Bitcoin/BlockPulling/BlockPuller.cs
noescape00: I've read. Looks like it's a bug. 
It should be fixed by eliminating the source of the problem, not the consequence.
Ignoring the negative values is not the fix, it's a workaround. 

I've created an issue for it: https://github.com/stratisproject/StratisBitcoinFullNode/issues/2224

src/NBitcoin/Consensus.cs
noescape00: for some test? why value of 500 isn't good?

fassadlr: Mining 500 blocks in integration tests takes too long duded ; I dont see an issue with a private setter. We need in anycase look into being able to set consensus settings in test after the network has been created.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Will add a comment 👍

fassadlr:
I have 2 more tests to add to ConsensusManager after this PR gets merge:

image

dangershony:
Ok synced networks for a while looks good

PR #2230, opened 2018-09-20
[SC] Renaming

codingupastorm:
No actual code or functionality changes here, this just changes 'SmartContract' to 'Contract' in all classes in .Core and .Reflection.Executor, as well as a couple of other minor changes.


src/Stratis.SmartContracts.Executor.Reflection/ExternalExecutor.cs
```diff
@@ -12,23 +12,23 @@ namespace Stratis.SmartContracts.Executor.Reflection
     /// <summary>
     /// Deserializes raw contract transaction data, creates an external create/call message, and applies the message to the state.
     /// </summary>
-    public class Executor : ISmartContractExecutor
+    public class ExternalExecutor : IContractExecutor
```
fassadlr: I'm wondering whether we shouldn't call this `ContractExecutor` perhaps?

codingupastorm: I like this suggestion, will update

PR #2231, opened 2018-09-20
[Fix] BlockPullerBehavior SpeedBytesPerSecond can be negative sometimes

### LINKED ISSUE #2224: BlockPullerBehavior SpeedBytesPerSecond can be negative sometimes

fassadlr:
Fix #2224


noescape00:
I wonder if the bug is connected to `this.averageDelaySeconds.Average` being so small so it goes negative because of small calculation precision of `double`.
Otherwise I can't see why would it go negative.

PR #2235, opened 2018-09-20
FinalizedBlockRepository speedup

noescape00:
1) Moved finalized block info logic from ChainRepo to a separated component
2) Speeded up saving process so when we save it we save in memory and it goes to DB in the background

The reason to do that PR is FV speed. Every block we validate we update finalized block info and saving this to DB under the reorg lock is too costly. With this PR we increase FV speed by 130 ms per block (for checkpointed blocks total FV  time is around 160 sec)

ps
This PR requires deletion of the stratis folder and syncing from scratch.
This is due to separation of chainRepo and finalizedBlockHash in the database schema. 

Fix: #2236


src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+        private readonly AsyncManualResetEvent queueUpdatedEvent;
+
+        public FinalizedBlockInfoRepository(string folder, ILoggerFactory loggerFactory)
+        {
+            Guard.NotEmpty(folder, nameof(folder));
+            Guard.NotNull(loggerFactory, nameof(loggerFactory));
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            Directory.CreateDirectory(folder);
```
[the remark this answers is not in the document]
noescape00: `Directory.CreateDirectory` creates a directory only if it doesn't exist.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+            this.dbreeze = new DBreezeEngine(folder);
+
+            this.finalizedBlockInfosToSave = new Queue<HashHeightPair>();
+            this.queueLock = new object();
+
+            this.queueUpdatedEvent = new AsyncManualResetEvent(false);
+            this.cancellation = new CancellationTokenSource();
```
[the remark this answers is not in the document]
noescape00: No, it's better to have it's own source that is cancelled on disposal. Otherwise we might have problems with testing (would be kinda weird if the component hangs on disposal because inner loop is waiting for nodeLifetime's token)

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+            this.logger.LogTrace("(-):true");
+            return true;
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            this.logger.LogTrace("()");
+
+            this.cancellation.Cancel();
```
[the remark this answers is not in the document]
noescape00: Answered in the discussion above.

src/Stratis.Bitcoin.Tests/FinalizedBlockInfoRepositoryTest.cs
```diff
@@ -0,0 +1,57 @@
...
+using Stratis.Bitcoin.Networks;
+using Stratis.Bitcoin.Tests.Common;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests
+{
+    public class FinalizedBlockInfoRepositoryTest : TestBase
+    {
+        public FinalizedBlockInfoRepositoryTest() : base(KnownNetworks.StratisRegTest)
+        {
```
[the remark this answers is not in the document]
noescape00: Good point. Fixed

noescape00:
@fassadlr fixes applied.

src/Stratis.Bitcoin.Tests/FinalizedBlockInfoRepositoryTest.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public class FinalizedBlockInfoRepositoryTest : TestBase
+    {
+        private readonly ILoggerFactory loggerFactory;
+        private readonly Network network;
+
+        public FinalizedBlockInfoRepositoryTest() : base(KnownNetworks.StratisRegTest)
+        {
+            this.loggerFactory = new LoggerFactory();
+            this.network = KnownNetworks.StratisRegTest;
```
fassadlr: Ah actually you don't need to do this as it is set in the base, so you can just use `this.Network` everywhere ;)

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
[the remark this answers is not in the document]
fassadlr: Ah yeah makes sense! 👍

src/Stratis.Bitcoin.Tests/FinalizedBlockInfoRepositoryTest.cs
noescape00: fixed

noescape00:
@fassadlr fixes applied

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            this.logger.LogTrace("({0}:{1})", nameof(height), height);
+
+            if (this.finalizedBlockInfo != null && height <= this.finalizedBlockInfo.Height)
+            {
+                this.logger.LogTrace("(-)[CANT_GO_BACK]:false");
+                return false;
+            }
+
+            var finalizedInfo = new HashHeightPair(hash, height);
```
fassadlr: Do we need to create a variable here?

Can't we just use do `this.finalizedBlockInfo = new HashHeightPair(hash, height)` below?

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+                {
+                }
+
+                this.queueUpdatedEvent.Reset();
+
+                HashHeightPair lastFinalizedBlock = null;
+
+                lock (this.queueLock)
+                {
+                    while (this.finalizedBlockInfosToSave.Count != 0)
```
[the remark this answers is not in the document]
noescape00: in the queue we will have: 

456, 457, 458, 459.

Only  459 should be saved. 

So the logic is correct and intended. It's there in case we are pushing new finalized block infos faster than we can persist them on the harddrive. When this happens we ONLY need the most recent one.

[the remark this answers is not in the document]
dangershony: What if the node crashes? how do we make sure finalized is in sync with consensus?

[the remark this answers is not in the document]
noescape00: If the node crashes we will have finalized height that was saved 100ms before the crush. 
And we will update it with the first block being connected so it's all good. 
Plus on startup we recalculate the actual finalized block hash (because we rewind consensus in case of crush and finalized block info with it)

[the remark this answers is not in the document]
fassadlr: "So the logic is correct and intended. It's there in case we are pushing new finalized block infos faster than we can persist them on the harddrive. When this happens we ONLY need the most recent one."

OK 💯 makes perfect sense. But perhaps add a comment on top to explain the logic? Unless I'm just being stupid here ;)

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
fassadlr: Again, this makes complete sense bu maybe just document that?

noescape00: Sure, will do

noescape00: Kk, sure

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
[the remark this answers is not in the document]
noescape00: Fixed

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
noescape00: Fixed

noescape00:
@fassadlr fixes applied

noescape00:
@fassadlr thanks mate 🙂

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        public FinalizedBlockInfoRepository(string folder, ILoggerFactory loggerFactory)
+        {
+            Guard.NotEmpty(folder, nameof(folder));
+            Guard.NotNull(loggerFactory, nameof(loggerFactory));
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            Directory.CreateDirectory(folder);
+            this.dbreeze = new DBreezeEngine(folder);
```
dangershony: One think I don't understand is why create a new store we could like before just reuse chain store (this will create a new folder and the whole dbreeze structure just to store a hash and a height.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        private async Task PersistFinalizedBlockInfoContinuouslyAsync()
+        {
+            this.logger.LogTrace("()");
+
+            while (!this.cancellation.IsCancellationRequested)
+            {
+                try
+                {
+                    await this.queueUpdatedEvent.WaitAsync(this.cancellation.Token).ConfigureAwait(false);
```
dangershony: Why not AsyncQueue? exactly the same result.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// <remarks>
+    /// Finalized block height is the height of the last block that can't be reorged.
+    /// Blocks with height greater than finalized height can be reorged.
+    /// <para>Finalized block height value is always <c>0</c> for blockchains without max reorg property.</para>
+    /// </remarks>
+    public interface IFinalizedBlockInfoRepository : IDisposable
+    {
+        /// <summary>Gets the finalized block hash and height.</summary>
+        /// <returns>Hash and height of a block that can't be reorged away from.</returns>
+        HashHeightPair GetFinalizedBlockInfo();
```
dangershony: This could return null wich might crash the caller.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -710,7 +710,7 @@ private async Task<ConnectBlocksResult> ConnectChainAsync(ChainedHeader newTip,
                         {
                             uint256 newFinalizedHash = blockToConnect.ChainedHeader.GetAncestor(newFinalizedHeight).HashBlock;
 
-                            await this.finalizedBlockInfo.SaveFinalizedBlockHashAndHeightAsync(newFinalizedHash, newFinalizedHeight).ConfigureAwait(false);
+                            this.finalizedBlockInfo.SaveFinalizedBlockHashAndHeight(newFinalizedHash, newFinalizedHeight);
```
dangershony: Could this piece of code go in to a rule?

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
noescape00: Yeah, that's for logical separation. 
We can have chain store on any chain, but finalized block is only stratis thing.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
noescape00: We don't know how many blocks are in the queue. It doesn't implement .Count and therefore we can't decide if we got the last item or there are more. 
And we need to save only last item.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
noescape00: No.
Only if we didn't call `LoadFinalizedBlockInfoAsync` which is a part of initialization.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Yeah, sounds like a good idea. Will add to my todo list

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
dangershony: yes but in that case it will just be empty no big deal, better then creating a whole new store just for a hash.

src/Stratis.Bitcoin/FinalizedBlockInfoRepository.cs
dangershony: Ah yes that is a limitation.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: In fact this should be part of the CoinView right now, becuse of flushing.
There is a race condition if node will crush and coin view did not persist.

noescape00: Good point

PR #2243, opened 2018-09-20
improved the CanFindFork test

MithrilMan:
commented the tested scenarios and added a missing one (scenario 4)

previously the test was simplier, were comparing 2 small chains (around 5 blocks long) and missed a test case.

I changed this in order to have a longer chain to test and commented the 5 tested scenarios (4th is new)

```
//test scenario 1:
//chain2 is empty, so the fork point is supposed to be the Genesis
//chain1) G a b c d e
//chain2) G

//test scenario 2:
//chain2 is a chain on another network (main vs test), null expected
//chain1) G a b c d e
//chain2) Gtest

//test scenario 3:
//chain2 is a subset of chain2 and stops at the fork point "c", fork point expected
//chain1) G a b c d e
//chain2) G a b c

//test scenario 4:
//chain2 is a forked chain (at the fork point "c") that has other blocks on top of it, fork point expected
//chain1) G a b c d e
//chain2) G a b c d2 e2 f2


//test scenario 5:
//chain2 is at the same tip of chain1, no fork happened, tip point expected
//chain1) G a b c d e
//chain2) G a b c d e
```

### DIFF
--- a/src/NBitcoin.Tests/ChainTests.cs
+++ b/src/NBitcoin.Tests/ChainTests.cs
@@ -172,74 +172,64 @@ namespace NBitcoin.Tests
         {
             var chain1 = new ConcurrentChain(this.network);
 
-				//create the main chain with a (commonChainSize) blocks before the fork
-				int commonChainSize = 150000;
-				for (int i = 0; i < commonChainSize; i++) {
-					this.AppendBlock(chain1);
-				}
+            // Create the main chain with a commonChainSize blocks before the fork.
+            int commonChainSize = 150000;
+            for (int i = 0; i < commonChainSize; i++) {
+	            this.AppendBlock(chain1);
+            }
 
-				ChainedHeader fork = this.AppendBlock(chain1);
-				//add some blocks from the fork point to the tip
-				int chain1ForkDepth = 10;
-				for (int i = 0; i < chain1ForkDepth; i++) {
-					this.AppendBlock(chain1);
-				}
-				//tip of the main chain
-				ChainedHeader tip = chain1.Tip;// this.AppendBlock(chain);
+            ChainedHeader fork = this.AppendBlock(chain1);
 
+            // Add some blocks from the fork point to the tip.
+            int chain1AppendixSize = 100;
+            for (int i = 0; i < chain1AppendixSize; i++)
+	            this.AppendBlock(chain1);
 
-				{ 
-					//test scenario 1:
-					//chain2 is empty, so the fork point is supposed to be the Genesis
-					//chain1) G a b c d e
-					//chain2) G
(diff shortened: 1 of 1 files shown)


src/NBitcoin.Tests/ChainTests.cs
```diff
@@ -170,21 +170,76 @@ public void CanBuildChain()
...
+				}
+
+				ChainedHeader fork = this.AppendBlock(chain1);
+				//add some blocks from the fork point to the tip
+				int chain1ForkDepth = 10;
+				for (int i = 0; i < chain1ForkDepth; i++) {
+					this.AppendBlock(chain1);
+				}
+				//tip of the main chain
+				ChainedHeader tip = chain1.Tip;// this.AppendBlock(chain);
```
zeptin: Generally we avoid introducing commented out code to the codebase

src/NBitcoin.Tests/ChainTests.cs
```diff
@@ -170,21 +170,76 @@ public void CanBuildChain()
...
-            this.AssertFork(chain, chain2, null);
-            chain2 = new ConcurrentChain(this.network);
-            chain2.SetTip(fork);
-            this.AssertFork(chain, chain2, fork);
-            chain2.SetTip(tip);
-            this.AssertFork(chain, chain2, tip);
+            var chain1 = new ConcurrentChain(this.network);
+
+				//create the main chain with a (commonChainSize) blocks before the fork
+				int commonChainSize = 150000;
```
zeptin: Is there benefit to adding this quantity of blocks? Would we get equivalent test coverage by decreasing it to something much lower?

MithrilMan: I let it as a parameter because i'm not sure, I added so much block because i thought there was a bug in the ChainHeader.FindFork and tried to see if it was happening with a longer chain.

With a very short chain, the skip list used by ConcurrentChain doesn't come into action at full, so this code was the reason i thought that with an higher chain, a bug could be exposed (but it didn't)
```
while (highChain.HashBlock != lowChain.HashBlock)
            {
                if (highChain.Skip != lowChain.Skip)
                {
                    highChain = highChain.Skip;
                    lowChain = lowChain.Skip;
                }
                else
                {
                    lowChain = lowChain.Previous;
                    highChain = highChain.Previous;
                }

                if ((lowChain == null) || (highChain == null))
                    return null;
            }
```

it takes 3s on my pc to test this method, anyway we can lower that parameter if it's a problem

src/NBitcoin.Tests/ChainTests.cs
MithrilMan: oversight

noescape00:
Waiting for @zeptin to dismiss his review.

PR #2244, opened 2018-09-21
[Integration Tests] Fix TestHelper's IsNodeSynced methods

fassadlr:
If the node are at genesis the nodes should be considered to be synced.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
@@ -37,6 +37,10 @@ namespace Stratis.Bitcoin.IntegrationTests.Common
 
         public static bool AreNodesSynced(CoreNode node1, CoreNode node2, bool ignoreMempool = false)
         {
+            // If the nodes are at genesis they are considered synced.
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -37,6 +37,10 @@ public static void WaitLoop(Func<bool> act, string failureReason = "Unknown Reas
 
         public static bool AreNodesSynced(CoreNode node1, CoreNode node2, bool ignoreMempool = false)
         {
+            // If the nodes are at genesis they are considered synced.
+            if (node1.FullNode.Chain.Tip.Height == 0 && node2.FullNode.Chain.Tip.Height == 0)
```
dangershony: What if node1=0 and node2=10

then the condition bellow will fail no?
```
 if (node2.FullNode.BlockStore().GetBlockAsync(node1.FullNode.ChainBehaviorState.ConsensusTip.HashBlock).Result == null)
       return true
```

fassadlr: Well yes then the nodes arent in sync? Isnt correct?

dangershony: ahhh yes ofcourse awesome

PR #2245, opened 2018-09-21
Disable logs for testing

dangershony:
Logs have been accumulated for each tests (as its a static collection so each node start will create logs)

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
@@ -8,6 +8,7 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using NBitcoin;
 using NBitcoin.Protocol;
+using NLog;
 using Stratis.Bitcoin.Builder;
 using Stratis.Bitcoin.Features.BlockStore;
 using Stratis.Bitcoin.Features.Consensus;
@@ -160,6 +161,10 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
             {
                 node.Start();
             }
+
+            // disabled logs for testing.
+            LogManager.Configuration.LoggingRules.Clear();
+            LogManager.ReconfigExistingLoggers();
         }
 
         public void Dispose()
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
@@ -155,22 +155,33 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers
             return dataFolderName;
         }
 
-        public void StartAll()
+        public void StartAll(bool enableLogs = false)
         {
             foreach (CoreNode node in this.Nodes.Where(n => n.State == CoreNodeState.Stopped))
             {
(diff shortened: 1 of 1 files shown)


dangershony:
Alternatively if we want logs in tests we can clear them on the dispose but I don't see why we need logs.

bokobza:
Can we look for a solution where we can turn the logs ON for a test if we want to?
IMO this is an easy but less than desirable solution.

zeptin:
Just to add my two cents to the discussion, when you have multiple nodes running in an integration test, their logs all get logged together in the last node's data folder. This makes it extremely difficult (impossible) to piece together a logical sequence of events for a given node. There is therefore little point having logs at all during most non-trivial tests, unless this can be resolved.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -160,6 +161,10 @@ public void StartAll()
             {
                 node.Start();
             }
+
+            // disabled logs for testing.
+            LogManager.Configuration.LoggingRules.Clear();
```
noescape00: maybe provide a bool in StartAll so it can be manually enabled for tests debuging?

noescape00:
Approved but would very much like to have an option to turn logs on instead of hardcoded logs disabling.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
MithrilMan: isn't DisableLogging enough?
https://nlog-project.org/documentation/v4.5.0/html/M_NLog_LogManager_DisableLogging.htm

so if a test need it, you can call EnableLogging again and DisableLogging at the end

noescape00: Agreed. Very good suggestion. @dangershony what do you think?

I'll block the PR so it's not merged just yet.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/NodeBuilder.cs
```diff
@@ -154,18 +155,33 @@ private string GetNextDataFolderName(string folderName = null)
...
+        public void StartAll(bool enableLogs = false)
         {
             foreach (CoreNode node in this.Nodes.Where(n => n.State == CoreNodeState.Stopped))
             {
                 node.Start();
             }
+
+            if (enableLogs)
+            {
+                while (!LogManager.IsLoggingEnabled())
```
noescape00: why do we need while loop? 
If you call it once it doesn't work?

MithrilMan: @noescape00 internally it uses counter, look at Slack conversation

dangershony: Its just in case, if you look in the code there is a logs counter and the method `DisableLogging()` decrease the counter by 1

noescape00: Ah, ok. Makes perfect sense

MithrilMan:
I've completed (with some changes) the change within this PR: 
https://github.com/stratisproject/StratisBitcoinFullNode/pull/2267

PR #2246, opened 2018-09-21
Trace logging for consensus rules

noescape00:


noescape00:
Found a better solution. See https://github.com/stratisproject/StratisBitcoinFullNode/issues/2249

PR #2248, opened 2018-09-22
[Consensus] Implement PosBlockContextRule

### LINKED ISSUE #2151: Implement PosBlockContextRule
Implement PosBlockContextRule and rename it to have a somewhat meaningful name
Currently we have TODO there

### LINKED ISSUE #2247: [Consensus] Possible signature validation bug
While working on #2151 I found the following apparent discrepancy between stratisX and SBFN:

stratisX:
```
bool IsDERSignature(const valtype &vchSig, bool haveHashType) {
    // See https://bitcointalk.org/index.php?topic=8392.msg127623#msg127623
    // A canonical signature exists of: <30> <total len> <02> <len R> <R> <02> <len S> <S> <hashtype>
    // Where R and S are not negative (their first byte has its highest bit not set), and not
    // excessively padded (do not start with a 0 byte, unless an otherwise negative number follows,
    // in which case a single 0 byte is necessary and even required).
    if (vchSig.size() < 9)
        return error("Non-canonical signature: too short");
    if (vchSig.size() > 73)
        return error("Non-canonical signature: too long");
    if (vchSig[0] != 0x30)
        return error("Non-canonical signature: wrong type");
    if (vchSig[1] != vchSig.size() - (haveHashType ? 3 : 2))
        return error("Non-canonical signature: wrong length marker");
```

SBFN:
```
        public static bool IsValidSignatureEncoding(byte[] sig)
        {
            // Format: 0x30 [total-length] 0x02 [R-length] [R] 0x02 [S-length] [S] [sighash]
            // * total-length: 1-byte length descriptor of everything that follows,
            //   excluding the sighash byte.
            // * R-length: 1-byte length descriptor of the R value that follows.
            // * R: arbitrary-length big-endian encoded R value. It must use the shortest
            //   possible encoding for a positive integers (which means no null bytes at
            //   the start, except a single one when the next byte has its highest bit set).
            // * S-length: 1-byte length descriptor of the S value that follows.
            // * S: arbitrary-length big-endian encoded S value. The same rules apply.
            // * sighash: 1-byte value indicating what data is hashed (not part of the DER
            //   signature)

            int signLen = sig.Length;

            // Minimum and maximum size constraints.
            if(signLen < 9 || signLen > 73)
                return false;

            // A signature is of type 0x30 (compound).
            if(sig[0] != 0x30)
                return false;

            // Make sure the length covers the entire signature.
            if(sig[1] != signLen - 3)
                return false;
```

In stratisX the `ProcessBlock` method ultimately calls `IsDERSignature` with the boolean set to `false`, implying that the `hashType` is missing from the signature & that the constant `2` should be used to subtract from the length. In SBFN we do not have the bool, but by using `3` we are assuming the `hashType` is in fact affixed to the signature.

This does not appear to always be the case, e.g. while processing block 35 on Stratis testnet the following block signature was extracted:

`BlockSignature` = `3045022100ee488514f8f8dbad8176b4a314d47e22b4be30a335ac4032225b4d85cb83985a02206bcfd07da2fd1622fda056af2881fdfa1a2e27ad9dbecd35ab73ea91a31e8bf9`

```
            // 71 bytes in total (>= 9 && <= 73) [signLen = 71]

            // 0x30 -> type is correct
            // 0x45 = total len = 69
            // 0x02
            // 0x21 = len R = 33
            // 00ee488514f8f8dbad8176b4a314d47e22b4be30a335ac4032225b4d85cb83985a = R
            // 0x02
            // 0x20 = len S = 32
            // 6bcfd07da2fd1622fda056af2881fdfa1a2e27ad9dbecd35ab73ea91a31e8bf9
            // -> no hashType
```

Therefore this block fails the tests in `IsValidSignatureEncoding`. These tests are not currently used in the codebase but will be introduced at some point via #2151.

zeptin:
Fixes #2151.
Fixes #2247.

The rule was renamed to `PosBlockSignatureRepresentationRule` as this more accurately describes what it is validating.

The `PosBlockValidator` functions were for the most part already implemented, but a slight difference with the stratisX equivalents was found.

Tests still need to be added to this PR, but the consensus code change (via the addition of the rule) can be reviewed in the interim.

Useful stratisX source file cross-references:

`IsDERSignature` - script.cpp
`IsLowDERSignature` - script.cpp
`IsCanonicalBlockSignature` - main.cpp


dangershony:
Ah excellent, ok I will need time to review that properly.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRepresentationRule.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public class PosBlockSignatureRepresentationRule : PartialValidationConsensusRule
+    {
+        /// <inheritdoc />
+        public override Task RunAsync(RuleContext context)
+        {
+            if (!PosBlockValidator.IsCanonicalBlockSignature((PosBlock)context.ValidationContext.BlockToValidate, false))
+            {
+                // In stratisX CANONICAL_BLOCK_SIG_VERSION (version.h) is defined as 70000, the same as ProtocolVersion.ALT_PROTOCOL_VERSION.
+                if (this.Parent.NodeSettings.ProtocolVersion >= ProtocolVersion.ALT_PROTOCOL_VERSION)
```
ferdeen: Can't you add CANONICAL_BLOCK_SIG_VERSION to ProtocolVersion, with a value 7000 then no need to add the comment and CANONICAL_BLOCK_SIG_VERSION could be used elsewhere, if need be ?

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRepresentationRule.cs
```diff
@@ -0,0 +1,37 @@
...
+                    ConsensusErrors.BadBlockSignature.Throw(); // bad block signature encoding
+            }
+
+            if (!PosBlockValidator.IsCanonicalBlockSignature((PosBlock)context.ValidationContext.BlockToValidate, true))
+            {
+                // In stratisX CANONICAL_BLOCK_SIG_LOW_S_VERSION (version.h) is defined as 70000, the same as ProtocolVersion.ALT_PROTOCOL_VERSION.
```
ferdeen: Same here.

src/Stratis.Bitcoin.Features.Consensus/ConsensusFeature.cs
```diff
@@ -203,7 +203,7 @@ public void RegisterRules(IConsensus consensus)
                     new WitnessCommitmentsRule(), // BIP141, BIP144
                     new BlockSizeRule(),
 
-                    new PosBlockContextRule(), // TODO: this rule needs to be implemented
+                    new PosBlockSignatureRepresentationRule(), // TODO: this rule needs to be implemented
```
ferdeen: Is the comment to the right of this still the case ?

zeptin: No, this should have been removed. I'll fix it.

zeptin:
@noescape00 And stratisX does not?

noescape00:
it looks like stratis x doesn't

zeptin:
`ProcessBlock` in `main.cpp`:

```
    if (!IsCanonicalBlockSignature(pblock, false)) {
        if (pfrom && pfrom->nVersion >= CANONICAL_BLOCK_SIG_VERSION) {
            pfrom->Misbehaving(100);
        }

        return error("ProcessBlock(): bad block signature encoding");
    }

    if (!IsCanonicalBlockSignature(pblock, true)) {
        if (pfrom && pfrom->nVersion >= CANONICAL_BLOCK_SIG_LOW_S_VERSION) {
            pfrom->Misbehaving(100);
            return error("ProcessBlock(): bad block signature encoding (low-s)");
        }

        if (!EnsureLowS(pblock->vchBlockSig))
            return error("ProcessBlock(): EnsureLowS failed");
    }
```

Aprogiena:
not only netsplit, but looks like node crash also possible due to wrong logic in the rule, we should probably get rid of the peer version checking completely

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRepresentationRule.cs
dangershony: you dont need settings we plan to put protocol version on the network class (most likely)

dangershony: you also have a bug here, in stratisx we throw regardless of protocol version (PV is only used to determine if we need to ban the peer)

dangershony: to avoid passing settings everywhere I suggest we even just use numbers to determine banning

dangershony: also you are checking the protocol version of our node not the peers

Aprogiena:
oh actually Dan is right, the logic is different from sX because sX uses peer's version while here we use our own version, which i missed. so what we stated is not true for current code, it would become true if we used peer's version as sX does.

Aprogiena:
moreover, the placement of the rule seems incorrect, because we need it to be executed before PosBlockSignatureRule is executed otherwise  PosBlockSignatureRule can crash if signature is in invalid format

Aprogiena:
to conclude 
- move the rule in front of  PosBlockSignatureRule  (this by itself makes sure block is not banned, just peer, which is to avoid netsplit)
- remove checking protocol version
- remove EnsureLowS

dangershony:
I'll review this tomorrow

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockContextRule.cs
```diff
@@ -1,45 +0,0 @@
...
-    /// Context checks on a POS block.
-    /// </summary>
-    public class PosBlockContextRule : PartialValidationConsensusRule
-    {
-        /// <inheritdoc />
-        public override Task RunAsync(RuleContext context)
-        {
-            // TODO: fix this validation code
-
-            //// check proof-of-stake
```
dangershony: What a bout this ?`Limited duplicity on stake`

dangershony: If we don't fix this as part of this PR we should at least create an issue to fix it.

src/Stratis.Bitcoin.Features.SmartContracts/SmartContractPosRuleRegistration.cs
```diff
@@ -23,6 +23,7 @@ public void RegisterRules(IConsensus consensus)
             consensus.IntegrityValidationRules = new List<IIntegrityValidationConsensusRule>()
             {
                 new BlockMerkleRootRule(),
+                new PosBlockSignatureRepresentationRule(),
```
dangershony: Yeah seems right, this is more of an integrity check.

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRepresentationRule.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Check that the block signature for a POS block is in the canonical format.
+    /// </summary>
+    public class PosBlockSignatureRepresentationRule : IntegrityValidationConsensusRule
+    {
+        /// <inheritdoc />
+        /// <exception cref="ConsensusErrors.BadBlockSignature">The block signature is not in the canonical format.</exception>
+        public override void Run(RuleContext context)
+        {
+            if (!PosBlockValidator.IsCanonicalBlockSignature((PosBlock)context.ValidationContext.BlockToValidate, true))
```
dangershony: We test only once with the sighash, in X there is also the test without the sighash

src/Stratis.Bitcoin.Features.SmartContracts/SmartContractPosRuleRegistration.cs
dangershony: >move the rule in front of PosBlockSignatureRule (this by itself makes sure block is not banned, just peer, which is to avoid netsplit)

Following this comment its very correct integrity failure will ban the node not the block

zeptin: Yes, agreed

src/Stratis.Bitcoin.Features.Consensus/PosBlockValidator.cs
```diff
@@ -12,7 +12,7 @@ public static bool IsCanonicalBlockSignature(PosBlock block, bool checkLowS)
 
             return checkLowS ?
```
dangershony: Seems this check is not similar to whats in X

https://github.com/stratisproject/stratisX/blob/master/src/main.cpp#L2161

dangershony: Also seems to me we now only call IsCanonicalBlockSignature once with boolan to check lowS

zeptin: As discussed, `IsDERSignature` is named `IsValidSignatureEncoding` in SBFN. Or was it a different part of the check you're referring to?

src/NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -1797,7 +1797,7 @@ private static int CompareBigEndian(byte[] c1, int ic1, int c1len, byte[] c2, in
         }
 
 
-        public static bool IsValidSignatureEncoding(byte[] sig)
+        public static bool IsValidSignatureEncoding(byte[] sig, bool haveSigHash)
```
dangershony: I think this needs to be by default true so `bool haveSigHash = true` and only from the method `IsCanonicalBlockSignature` we call it with false (thats when we validate POS blocks)

dangershony: please default this to true, BTC assumes always a sigh hash, we only remove the sighash if we are checking a POS signed block.

src/Stratis.Bitcoin.Features.Consensus/PosBlockValidator.cs
```diff
@@ -12,7 +12,7 @@ public static bool IsCanonicalBlockSignature(PosBlock block, bool checkLowS)
...
                 ScriptEvaluationContext.IsLowDerSignature(block.BlockSignature.Signature) :
-                ScriptEvaluationContext.IsValidSignatureEncoding(block.BlockSignature.Signature);
+                ScriptEvaluationContext.IsValidSignatureEncoding(block.BlockSignature.Signature, false);
```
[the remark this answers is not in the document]
dangershony: also please add a comment here that says "For POS blocks that have a signature we do not append a sighash type at the end of the signature."

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosBlockSignatureRepresentationRule.cs
dangershony: ok so as @Aprogiena suggested we only check lows canonical blocks

dangershony:
Adding so explanation to @Aprogiena suggestions

>only call IsCanonicalBlockSignature once with true - we dont need to distinguish between wrong format and low S

is because sig encoding is checked already in the method check lows so no need to do that twice.

>remove EnsureLowS  

we dont support lows anymore see this link (thats why we dont need the encoding check without the lows check) just reject peers that provide lows blocks.

>remove checking protocol version  

we dont support lows protocol nodes so no need to check that anymore

src/Stratis.Bitcoin.Features.Consensus/PosBlockValidator.cs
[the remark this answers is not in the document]
zeptin: Done

src/NBitcoin/Crypto/ECDSASignature.cs
```diff
@@ -88,6 +88,20 @@ public ECDSASignature MakeCanonical()
...
+        /// </summary>
+        /// <remarks>Not to be used under normal circumstances</remarks>
+        public ECDSASignature MakeNonCanonical()
+        {
+            if (!this.IsLowS)
+            {
+                return this;
+            }
+            else
+                return new ECDSASignature(this.R, ECKey.CURVE_ORDER.Subtract(this.S));
```
dangershony: Did you check that this is correct?

src/NBitcoin/ScriptEvaluationContext.cs
```diff
@@ -1632,7 +1632,7 @@ internal bool CheckSignatureEncoding(byte[] vchSig)
             {
                 return true;
             }
-            if((this.ScriptVerify & (ScriptVerify.DerSig | ScriptVerify.LowS | ScriptVerify.StrictEnc)) != 0 && !IsValidSignatureEncoding(vchSig))
+            if((this.ScriptVerify & (ScriptVerify.DerSig | ScriptVerify.LowS | ScriptVerify.StrictEnc)) != 0 && !IsValidSignatureEncoding(vchSig, true))
```
dangershony: remove the `true` here its now redundant

src/NBitcoin/Crypto/ECDSASignature.cs
zeptin: Yes. I found multiple sources that confirm that, for an ECDSA signature, the `R` and `S` values are both modulo `N` (aka the curve order). Further, a signature `(R, S)` is equivalent to `(R, -S mod N)`.

In other words there are always 2 valid `S` values, call them `S` and `S'`. Clearly `S + S' = N`, so there will always be one value greater than half the curve order (`N / 2`) and one less than this, unless both somehow happen to be `= N / 2`. The canonical signature is required to be the so-called 'low S' value, the one less than `N / 2`.

Therefore to get the other `S'` value (the complement) we calculate `S' = N - S`.

We can switch between the canonical and non-canonical form by calculating the complement of whichever representation we currently have in a signature.

Using `N + S` will give a valid signature too, but will not give the complement, as `(N + S) mod N = S`

zeptin: A small example of why `S + S' = N`:

`N = 7`
`S = 4`

`((N - S) % N)` = `3` <- `S'`, therefore `S` + `S'` = `7` = `N`
`((-S) % N)` = `3`

And `(N + S) mod N = S`:

`((N + S) % N)` = `4`

dangershony: Ok thanks very much for this explanation @zeptin

dangershony:
This is ready to be merged all tests passed (apart for those we know fail)

noescape00:
@zeptin please add comments on what is canonical signature and how does it affect sig size. 
Also add comment to the rule on why it should be integrity rule (info that we can't have it as PV or FV because then we will ban the block and QT won't do that so it may lead to chainsplit because C# and C++ nodes will handle non-canonical sigs differently)

noescape00:
PR looks good to me, but some comments are missing.

noescape00:
Guys, imagine what happens if someone takes our FN and starts shuffling the rules around- he will introduce a chainsplit bug lust like that. 

@dangershony that's my main argument why we should use old school validators over the rules- rules depend too much on their order and type. And if you mess with it you won't notice that you've introduced a problem but the problem might be critical.

dangershony:
@noescape00 I can argue the same about validator style.

>imagine what happens if someone takes our FN and starts shuffling the rules around  

Imagine someone takes the node and change the validator, if you change things without you know what you are doing things can break this is true for all software systems.

zeptin:
I don't really see the need to add detail about the size, it is a relatively minor thing and was already documented in `IsLowDerSignature` already, as far as I can see.

PR #2251, opened 2018-09-23
Simplify conditions when PV success

dangershony:
This is a proposed fix to simplify the condition when partial validation completes.

This is to be discussed on the PR and not merged yet.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -472,20 +472,19 @@ namespace Stratis.Bitcoin.Consensus
                     }
                 }
 
-                // Partial validation should continue if:
-                //  1: There are more blocks to validate.
-                //  2: A full validation did not fail (if full validation was needed).
-                bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
-                bool fullValidationFailed = (connectBlocksResult != null) && !connectBlocksResult.Succeeded;
-                if (nextValidationNeeded && !fullValidationFailed)
+                // If more blocks are available continue validation.
+                if (chainedHeaderBlocksToValidate != null)
                 {
-                    this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);
-
+                    // Validate the next blocks if validation was not needed, or if needed then it succeeded.
+                    if ((connectBlocksResult == null) || connectBlocksResult.Succeeded)
+                    {
+                        this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);
 
                         // Start validating all next blocks that come after the current block,
                         // all headers in this list have the blocks present in the header.
                         foreach (ChainedHeaderBlock toValidate in chainedHeaderBlocksToValidate)
                             this.partialValidator.StartPartialValidation(toValidate.ChainedHeader, toValidate.Block, this.OnPartialValidationCompletedCallbackAsync);
+                    }
                 }
             }
 
@@ -666,6 +665,8 @@ namespace Stratis.Bitcoin.Consensus
             }
 
             disconnectedBlocks.Reverse();
+
+            this.logger.LogInformation("Reorg from block '{0}' to '{1}'", oldTip, fork);
 
             this.logger.LogTrace("(-):*.{0}={1}", nameof(disconnectedBlocks.Count), disconnectedBlocks.Count);
             return disconnectedBlocks;


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -472,20 +472,19 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
-                bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
-                bool fullValidationFailed = (connectBlocksResult != null) && !connectBlocksResult.Succeeded;
-                if (nextValidationNeeded && !fullValidationFailed)
+                // If more blocks are available continue validation.
+                if (chainedHeaderBlocksToValidate != null)
                 {
-                    this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);
-
+                    // Validate the next blocks if validation was not needed, or if needed then it succeeded.
+                    if ((connectBlocksResult == null) || connectBlocksResult.Succeeded)
```
fassadlr: Can we maybe rename `connectBlocksResult` to `fullValidationResult`? Just so that logically it reads better ;)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -472,20 +472,19 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
                 }
 
-                // Partial validation should continue if:
-                //  1: There are more blocks to validate.
-                //  2: A full validation did not fail (if full validation was needed).
-                bool nextValidationNeeded = chainedHeaderBlocksToValidate != null;
-                bool fullValidationFailed = (connectBlocksResult != null) && !connectBlocksResult.Succeeded;
-                if (nextValidationNeeded && !fullValidationFailed)
+                // If more blocks are available continue validation.
+                if (chainedHeaderBlocksToValidate != null)
```
fassadlr: I always prefer to rather return here than having nested ifs. So how about:

```
if (chainedHeaderBlocksToValidate = null)
{    
    this.logger.LogTrace("(-)[NO_FURTHER_BLOCKS]");
    return;
}
```

dangershony: hmmm to me this is more readable because we act on the fact its not null, but I am not sure if this entire change is more readable then before, its a suggestion.

fassadlr: Its definitely an improvement @dangershony ;)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
MithrilMan: I agree on changing the name variable if the meaning is that a full validation result happened

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -667,6 +666,8 @@ private async Task<List<ChainedHeaderBlock>> RewindToForkPointAsync(ChainedHeade
 
             disconnectedBlocks.Reverse();
 
+            this.logger.LogInformation("Reorg from block '{0}' to '{1}'", oldTip, fork);
```
MithrilMan: what about using string interpolation when logging?
e.g.
 this.logger.LogInformation($"Reorg from block '{oldTip}' to '{fork}'");

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
MithrilMan: What about creating an extension for `ValidationResult` like `IsNullOrSucceeded`
this way can be used on any inherited validation results you have and can simplify the checks

e.g.
`if ((connectBlocksResult == null) || connectBlocksResult.Succeeded)`

becomes
`if(connectBlocksResult.IsNullOrSucceeded())`

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: We dont use string interpolation for logs because they have a perf impact, if logging is disabled the object will not  try to resolve the ToString()

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: I would not necessary change the param name because this is not only validation its also connecting the block on top of the previous block when we do full validation.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -472,20 +472,19 @@ private async Task OnPartialValidationSucceededAsync(ChainedHeader chainedHeader
...
-                if (nextValidationNeeded && !fullValidationFailed)
+                // If more blocks are available continue validation.
+                if (chainedHeaderBlocksToValidate != null)
                 {
-                    this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);
-
+                    // Validate the next blocks if validation was not needed, or if needed then it succeeded.
+                    if ((connectBlocksResult == null) || connectBlocksResult.Succeeded)
+                    {
+                        this.logger.LogTrace("Partial validation of {0} block will be started.", chainedHeaderBlocksToValidate.Count);
```
noescape00: blocks

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: dot missing

MithrilMan: oh yes i remember now that discussion and my suggestion about `ConditionalAttribute`

Issue #1761, opened 2018-07-28
Missing RPC methods


noescape00:
Currently we implement 16 RPC methods

BTC core supports 58: https://en.bitcoin.it/wiki/Original_Bitcoin_client/API_calls_list

I think we should implement the rest of them (except for those that deal with non-HD wallet or with features we don't have).

dangershony:
We should implement RPC methods as and when we need them else it might be useless work.
If we dont need them for our core code we can delegate to the foundation

noescape00:
>We should implement RPC methods as and when we need them else it might be useless work.

We don't really need them because we use API. 

However around 4 people contacted me in the last 2 months who were tying to integrate FN to their project and they had problems with missing RPC methods. 2 of them chose QT wallet over FN just because of that reason. 

So I'd say RPC methods are really needed in case your goal is to make it easy for non-stratis developers to use FN.

dangershony:
Yes so please create a list of the RPC methods they require and we can ask the foundation to create bounties.

noescape00:
Basically all the commands that are dealing with wallets- 
get total amount of money that was received by address
get current address money amount (with argument of how many confirmations required). 

Those are 2 that pop up in my head right away.

sondreb:
These 6 methods are used by at least one exchange, not sure what others are relying on. The two first, is supported today, the 4 others are not (at least not returned in the listmethods results:

- **validateaddress**
- **getblockcount**
- getnewaddress
- sendtoaddress
- getwalletinfo
- gettransaction

sondreb:
I see that getnewaddress is supposed to have been implemented, but not returned as part of the listmethods. So this is a separate issue really, that perhaps listmethods doesn't return everything.

sondreb:
It is either a bug, or intentional change. I think it is a bug, with with the latest modifications to the Stratis daemon, it uses the cold staking wallet, and not the normal wallet. When it does this, the RPC controller in the wallet is not loaded and initialized in the RPC host. So all the RPC methods in the wallet is missing, unless the daemon instead launches with regular wallet, and not staking wallet.

zeptin:
Ah, nicely spotted sondreb, that makes perfect sense. We will need to resolve that.

maciejzaleski:
This issue relates to #2758

dangershony:
Closing as duplicate

PR #2255, opened 2018-09-24
Add more RPC methods to support NBitcoin Altcoin

mikedennis:
Adds the following rpc methods:
* sendrawtransaction
* getnewaddress
* getblock

Also fixed up a issue with getblock not able to return genesis block.

Partially address issue #1356


src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
```diff
@@ -271,6 +270,33 @@ public ValidatedAddress ValidateAddress(string address)
             return res;
         }
 
+        /// <summary>
+        /// RPC method for returning a block.
+        /// Currently only supports raw (hex) format, Json format is not yet supported.
```
zeptin: I think that, where possible, the RPC call could just invoke the underlying API methods to avoid the possibility of code duplication. That would, probably, solve the JSON issue, as the `BlockStore` controller appears to already implement this logic with JSON support in the `block` endpoint.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -29,13 +41,51 @@ public uint256 SendToAddress(BitcoinAddress bitcoinAddress, Money amount)
...
         }
 
+        /// <summary>
+        /// Broadcasts a raw transaction from hex to local node and network.
+        /// </summary>
+        /// <param name="hex">Raw transaction in hex.</param>
+        /// <returns>The transaction hash.</returns>
+        [ActionName("sendrawtransaction")]
+        [ActionDescription("Submits raw transaction (serialized, hex-encoded) to local node and network.")]
+        public async Task<uint256> SendTransactionAsync(string hex)
```
zeptin: Again, maybe you could just call `send-transaction` here.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -29,13 +41,51 @@ public uint256 SendToAddress(BitcoinAddress bitcoinAddress, Money amount)
...
+        }
+             
+        /// <summary>
+        /// RPC method that gets a new address for receiving payments.
+        /// Uses the first wallet and account.
+        /// </summary>
+        /// <returns>The new address.</returns>
+        [ActionName("getnewaddress")]
+        [ActionDescription("Returns a new wallet address for receiving payments.")]
+        public NewAddressModel GetNewAddress()
```
zeptin: `unusedaddress`/`unusedaddresses`. Note that this probably means you can remove the `NewAddressModel` entirely.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -0,0 +1,34 @@
...
+        public void TestRpcGetBlockWithValidHashIsSuccessful()
+        {
+            using (NodeBuilder builder = NodeBuilder.Create(this))
+            {
+                var node = builder.CreateStratisPowNode(KnownNetworks.RegTest);
+                builder.StartAll();
+                RPCClient rpcClient = node.CreateRPCClient();
+                node.NotInIBD();
+                node.SetDummyMinerSecret(new BitcoinSecret(new Key(), node.FullNode.Network));
+                node.GenerateStratisWithMiner(2);
```
zeptin: I think we're changing to `TestHelper.MineBlocks` here. I believe there's also a `.WithWallet()` helper method on the node that does some of the boilerplate so you don't have to set the miner secret etc.

fassadlr: Hi @mikedennis yes, please wait for my PR to be merged and then you can use the methods as described by @zeptin here 👍

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -445,6 +445,12 @@ public Task<Block> GetBlockAsync(uint256 hash)
                         res = blockRow.Value;
                 }
 
+                // If searching for genesis block, return it.
+                if (res == null && hash == this.network.GenesisHash)
```
dangershony: Hmm my intention was to implement this in IBlockStore, I guess here is also fine.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -29,13 +41,51 @@ public uint256 SendToAddress(BitcoinAddress bitcoinAddress, Money amount)
...
+
+            this.logger.LogTrace("(-):'{0}'", base58Address);
+            return new NewAddressModel(base58Address);
+        }
+
         private WalletAccountReference GetAccount()
         {
             //TODO: Support multi wallet like core by mapping passed RPC credentials to a wallet/account
-            string w = this.WalletManager.GetWalletsNames().FirstOrDefault();
+            string w = this.walletManager.GetWalletsNames().FirstOrDefault();
```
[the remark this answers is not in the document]
zeptin: `account 0` is the default, but I think that long-standing TODO has a point - ultimately some form of credential needs to be made available to the RPC functionality so that the methods that sign transactions can be implemented. Not saying that has to be in this PR, but it should be borne in mind.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTests.cs
```diff
@@ -214,6 +214,27 @@ public void CanGenerateByStringArgs()
...
 
+        /// <summary>
+        /// Tests that RPC method 'SendRawTransaction' can be called with a new transaction.
+        /// </summary>
+        [Fact]
+        public void SendRawTransaction()
+        {
+            var tx = new Transaction();
+            tx.Outputs.Add(new TxOut(Money.Coins(1.0m), new Key()));
+            this.rpcTestFixture.RpcClient.SendRawTransaction(tx);
```
dangershony: What is this test checking? there are not asserts.

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
dangershony: Do we already do that? (call the api endpoints from an RPC endpoint)

mikedennis: No I don't think that is the current design. It is a good point of discussion though whether we should be.  Should RPC sit on top of API? I'm a bit concerned about all the coupling between RPC, API, and Features.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: Unfortunately not, RPC has some funky serialization logic so need a separate model in this case.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
mikedennis: Ok thanks. I will take a look!

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
mikedennis: I might be able to do that, I will take a quick look.

zeptin:
Also note related issue #1761. Referencing it here for the implementers of the remaining methods.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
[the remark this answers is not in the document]
mikedennis: We are going to have to have a bigger discussion on wallet RPC methods in general. I need to implement sendtoaddress which requires signing, but I'll create a separate issue for discussion on how best to do this.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTests.cs
mikedennis: That the RPC method executes without throwing an exception. It basically broadcasts the transaction to the mempool I think, so maybe I could some how check that it was received in mempool. Not sure exactly what else to do here. 

The transaction should get rejected cause it isn't signed

src/Stratis.Bitcoin.Features.RPC/Controllers/FullNodeController.cs
zeptin: I think my main problem with that is that we then need two sets of tests for everything. But the coupling is problematic, yes.

I suppose it is an unnecessary concern if the API methods themselves diverge significantly in functionality from the RPC versions in future.

dangershony: If we don't already do that then leave this as is, we will indeed have too many shared code between the two, if we can though we use a Service that both API and RPC can call for the lower level stuff (when too much logic is inside the controller)

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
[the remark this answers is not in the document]
mikedennis: I created a issue #2266 if you guys have any suggestions on sendtoaddress

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTests.cs
dangershony: ok yeah you could ask the trx form the node and get back null but it wont help much so I guess like this is ok.

Issue #2260, opened 2018-09-24
BSQ check blocks in batch when looking for transaction


noescape00:
BSQ check blocks in batch when looking for transaction- first check batch and then go to BlockRepository

PR #2262, opened 2018-09-24
BSQ check tx in batch before going to ChainRepository + Tests

noescape00:
Fix: #2260


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -168,25 +169,55 @@ public async Task InitializeAsync()
...
+
+            return this.blockRepository.GetTransactionByIdAsync(trxid);
         }
 
         /// <inheritdoc/>
-        public Task<uint256> GetTrxBlockIdAsync(uint256 trxid)
+        public Task<uint256> GetBlockIdByTransactionIdAsync(uint256 trxid)
         {
-            return this.blockRepository.GetTrxBlockIdAsync(trxid);
+            lock (this.blocksCacheLock)
```
dangershony: I suggest reverse the order,
First look in disk  (as before), then in memory, to avoid holding the lock unnecessary if not needed.

zeptin: I think this is logical, because the common case will be when we are out of IBD, when we typically serve blocks that are near the tip via the BlockStoreBehavior. These are cached in memory already, so there is no performance hit going to disk.

It will ostensibly be peers trying to sync from genesis that need old uncached blocks, and there would be no point checking the pending cache for these.

noescape00: Discussed on slack. 

@dangershony are you ok with RAM lookup first given the arguments provided on slack?

dangershony: Yes ok this are miner arguments, I also realise there may be a minor race condition if we look at the disk first (and get back null) and between the time we acquire the lock someone is pushing to disk we will have a race condition.

So this is fine

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -65,17 +65,18 @@ public class BlockStoreQueue : IBlockStoreQueue
...
         private Task dequeueLoopTask;
 
         /// <summary>Protects the batch from being modifying while <see cref="GetBlockAsync"/> method is using the batch.</summary>
-        private readonly object getBlockLock;
+        private readonly object blocksCacheLock;
 
         /// <summary>Represents all blocks currently in the queue & pending batch, so that <see cref="GetBlockAsync"/> is able to return a value directly after enqueuing.</summary>
-        private readonly Dictionary<uint256, ChainedHeaderBlock> pendingBlocks;
+        /// <remarks>Write access should be protected by <see cref="blocksCacheLock"/>.</remarks>
+        private readonly Dictionary<uint256, ChainedHeaderBlock> pendingBlocksCache;
```
dangershony: There was no need to change the names of `getBlockLock` and `pendingBlocks` it only bloats the PR (remember for next time).

PR #2264, opened 2018-09-24
[API] Add stopmining method to API + tests

fassadlr:
As part of the integration test work we need the ability to:

1> Start mining on a POS network via the API.
2> Stop mining via an API call.

Currently if the node is on a POS network, the `Generate` method throws an exception when called. This is not entirely correct as we are able to mine up until the node has reached the height of `Consensus.LastPowBlock`.

If the testers spin up nodes in a test VM, they should be able to start the node on a POS network, start mining via an API call and stop it after the node has reached `LastPowBLock`'s height.


src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -20,40 +20,17 @@ namespace Stratis.Bitcoin.Features.Miner.Controllers
...
-        /// <summary>Instance logger.</summary>
+        private const string exceptionOccurredMessage = "Exception occurred: {0}";
+        private readonly IFullNode fullNode;
         private readonly ILogger logger;
 
-        /// <summary>PoW miner.</summary>
-        private readonly IPowMining powMining;
-
-        /// <summary>Wallet manager.</summary>
-        private readonly IWalletManager walletManager;
```
dangershony: Wait why remove the dependencies?
This is the opposite of what we want using `this.fullNode.NodeService<IWalletManager>()` is bad practice.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -62,16 +39,17 @@ public MiningController(Network network, IPowMining powMining, ILoggerFactory lo
...
-        public IActionResult Generate([FromBody]MiningRequest request)
+        public IActionResult StartMining([FromBody]MiningRequest request)
         {
             Guard.NotNull(request, nameof(request));
 
             try
             {
-                if (this.network.Consensus.IsProofOfStake)
-                    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", "Method not available for Proof of Work");
+                if (this.fullNode.Network.Consensus.IsProofOfStake &&
```
dangershony: Interesting, but not required that much becuase such blocks will anyway be rejected.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: OK I have reverted this 👍

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: Yes but we WANT to be able to mine on a POS network (see PR description) before LastPowBLock height ;) So there this check is needed otherwise we just block mining if the network's IsProofOfStake setting is true. This is to help Lukasz's testing ;)

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
```diff
@@ -70,31 +85,100 @@ public void Generate_Blocks_When_Model_Is_Invalid_ReturnsBadRequest()
...
             HdAddress address = WalletTestsHelpers.CreateAddress();
-            this.walletManager.Setup(w => w.GetUnusedAddress(new WalletAccountReference(wallet, account))).Returns(address);
+            walletManager.Setup(f => f.GetUnusedAddress(new WalletAccountReference(wallet, account))).Returns(address);
 
-            this.powMining.Setup(p => p.GenerateBlocks(It.Is<ReserveScript>(r => r.ReserveFullNodeScript == address.Pubkey), 1, int.MaxValue))
-                .Returns(new List<uint256> { new uint256(1255632623) });
+            var powMining = new Mock<IPowMining>();
+            powMining.Setup(f => f.GenerateBlocks(It.Is<ReserveScript>(r => r.ReserveFullNodeScript == address.Pubkey), 1, int.MaxValue)).Returns(new List<uint256> { new uint256(1255632623) });
+            powNode.Setup(f => f.NodeFeature<MiningFeature>(false)).Returns(new MiningFeature(KnownNetworks.RegTest, new MinerSettings(Configuration.NodeSettings.Default()), Configuration.NodeSettings.Default(), this.loggerFactory, new Mock<ITimeSyncBehaviorState>().Object, powMining.Object, null));
```
zeptin: Further to my other comment, I am not sure why we bother having `this.network` defined if the individual tests in the class operate across different networks.

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
```diff
@@ -70,31 +85,100 @@ public void Generate_Blocks_When_Model_Is_Invalid_ReturnsBadRequest()
...
+            IActionResult result = controller.StartMining(new MiningRequest { BlockCount = 1 });
+
+            var errorResult = Assert.IsType<ErrorResult>(result);
+            var errorResponse = Assert.IsType<ErrorResponse>(errorResult.Value);
+            Assert.Single(errorResponse.Errors);
+
+            ErrorModel error = errorResponse.Errors[0];
+            Assert.Equal(405, error.Status);
+            Assert.Equal("Method not allowed", error.Message);
+            Assert.Equal(string.Format("This is a POS node and it's consensus tip is higher that the allowed last POW block height of {0}", this.network.Consensus.LastPOWBlock), error.Description);
```
zeptin: `its`, `than`

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -62,16 +51,17 @@ public MiningController(Network network, IPowMining powMining, ILoggerFactory lo
         /// <returns>List of block header hashes of newly mined blocks.</returns>
         /// <remarks>It is possible that less than the required number of blocks will be mined because the generating function only
         /// tries all possible header nonces values.</remarks>
-        [Route("generate")]
+        [Route("startmining")]
```
zeptin: Are changes required for the RPC version of this?

Note that the name change should not apply to the RPC version as there is an expectation for `generate` to exist.

I don't think we currently implement the separate `setgenerate` RPC call. My feeling is that naming this method `StartMining` is somewhat misleading, as it is intended for once-off block generation. The equivalent of `setgenerate`, which would more correctly be called StartMining, is done via a config key that triggers block generation in a loop in the mining feature.

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
```diff
@@ -1,44 +1,57 @@
...
+        private readonly Mock<IFullNode> fullNode;
+        private readonly ILoggerFactory loggerFactory;
         private readonly Network network;
 
         public MiningControllerTest()
         {
-            this.powMining = new Mock<IPowMining>();
-            this.walletManager = new Mock<IWalletManager>();
-            this.network = KnownNetworks.TestNet;
+            this.network = KnownNetworks.StratisRegTest;
```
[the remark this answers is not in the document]
fassadlr: @zeptin most of the tests test against Stratis, however I see your point. I will define the networks so that they are local to each test.

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
fassadlr: See previous comment ;)

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
fassadlr: Will fix 👍

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: Actually good point re generation, let me revert this. We should actually create separate start mining API call? We can have this as a another PR.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -70,8 +59,9 @@ public IActionResult Generate([FromBody]MiningRequest request)
 
             try
             {
-                if (this.network.Consensus.IsProofOfStake)
-                    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", "Method not available for Proof of Work");
+                if (this.network.Consensus.IsProofOfStake &&
```
noescape00: brackets missing

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -70,8 +59,9 @@ public IActionResult Generate([FromBody]MiningRequest request)
...
+                    this.consensusManager.Tip.Height > this.network.Consensus.LastPOWBlock)
+                    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", string.Format("This is a POS node and it's consensus tip is higher than the allowed LastPowBlock height of {0}", this.network.Consensus.LastPOWBlock));
```
noescape00: Instead of `string.Format("This is a POS node and it's consensus tip is higher than the allowed LastPowBlock height of {0}"`

we can use `$"This is a POS node and it's consensus tip is higher than the allowed LastPowBlock height of {this.network.Consensus.LastPOWBlock}"`

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: Ah yes 👍

src/Stratis.Bitcoin.Features.Miner.Tests/Controllers/MiningControllerTest.cs
```diff
@@ -70,31 +85,100 @@ public void Generate_Blocks_When_Model_Is_Invalid_ReturnsBadRequest()
...
 
             Assert.NotNull(result);
             var viewResult = Assert.IsType<JsonResult>(result);
             var resultValue = Assert.IsType<GenerateBlocksModel>(viewResult.Value);
             Assert.NotNull(resultValue);
-        }     
+        }
+
+        [Fact]
+        public void StartMining_PosNetwork_ConsensusTip_IsBeforeLastPowBlock_ReturnsSuccess()
```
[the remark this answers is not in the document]
fassadlr: I did rename Generate to StartMining... so maybe I should rename the tests back.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -20,40 +21,28 @@ namespace Stratis.Bitcoin.Features.Miner.Controllers
     [Route("api/[controller]")]
     public class MiningController : Controller
     {
-        /// <summary>Instance logger.</summary>
-        private readonly ILogger logger;
+        private const string exceptionOccurredMessage = "Exception occurred: {0}";
```
bokobza: `const`s are pascal case.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -70,8 +59,8 @@ public IActionResult Generate([FromBody]MiningRequest request)
 
             try
             {
-                if (this.network.Consensus.IsProofOfStake)
-                    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", "Method not available for Proof of Work");
+                if (this.network.Consensus.IsProofOfStake && (this.consensusManager.Tip.Height > this.network.Consensus.LastPOWBlock))
+                    return ErrorHelpers.BuildErrorResponse(HttpStatusCode.MethodNotAllowed, "Method not allowed", $"This is a POS node and it's consensus tip is higher than the allowed LastPowBlock height of {this.network.Consensus.LastPOWBlock}");
```
bokobza: it's => its
This message contains too many details IMO.
"Mining is not allowed past block {this.network.Consensus.LastPOWBlock}." is simpler.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
```diff
@@ -20,40 +21,28 @@ namespace Stratis.Bitcoin.Features.Miner.Controllers
...
-        /// <param name="powMining">PoW miner.</param>
-        /// <param name="fullNode">Full Node.</param>
-        /// <param name="loggerFactory">Factory to be used to create logger for the node.</param>
-        /// <param name="walletManager">The wallet manager.</param>
-        public MiningController(Network network, IPowMining powMining, ILoggerFactory loggerFactory, IWalletManager walletManager) 
+        public MiningController(IConsensusManager consensusManager, IFullNode fullNode, ILoggerFactory loggerFactory, Network network, IPowMining powMining, IWalletManager walletManager)
         {
-            Guard.NotNull(network, nameof(network));
-            Guard.NotNull(powMining, nameof(powMining));
+            Guard.NotNull(consensusManager, nameof(consensusManager));
```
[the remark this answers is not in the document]
fassadlr: Have added all of them back.

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.Miner/Controllers/MiningController.cs
fassadlr: I want to include the fact this will only happen if the node is POS, so have amended to:
`"This is a POS node and mining is not allowed past block {0}"`

PR #2270, opened 2018-09-25
[SC] Integration test when contract doesn't exist.

codingupastorm:
- Added error message to receipt when contract fails due to StateTransitionError.
- Added integration test for calling a contract that doesn't exist.
- TestHelper.MineBlocks -> MockChainNode.MineBlocks to make SC tests great again.


src/Stratis.SmartContracts.Executor.Reflection/StateTransitionErrors.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Stratis.SmartContracts.Executor.Reflection
+{
+    public class StateTransitionErrors
```
rowandh: Make class static

PR #2275, opened 2018-09-25
[Magic Network] Give the ability to create custom networks

fassadlr:
To test on the Magic network we need to be able to run Stratis on custom networks so that Lukasz can test various scenarios with scripts (like reorgs) without having to wait for certain block heights etc.

This way we can create numerous test networks and included them in Stratis.StratisD.Test without it interfering with the main node.

### DIFF
--- a/src/Stratis.StratisD.Test/Program.cs
+++ b/src/Stratis.StratisD.Test/Program.cs
@@ -14,6 +14,7 @@ using Stratis.Bitcoin.Features.Miner;
 using Stratis.Bitcoin.Features.RPC;
 using Stratis.Bitcoin.Features.Wallet;
 using Stratis.Bitcoin.Utilities;
+using Stratis.StratisD.Test.Networks;
 
 namespace Stratis.StratisD.Test
 {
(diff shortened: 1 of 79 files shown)


bokobza:
I feel strongly against creating a network that bears literally no resemblance to the networks it's supposed to test. Strong Nack from me.

noescape00:
Or we can allow specifying network file which will be json serialized network class. And in case new testnet is required without recompilation a new file will be provided in the command line: `network=myCustomNetwork.json`

fassadlr:
@bokobza I don't see the issue with this? Remember these networks sit outside of the main Stratis projects and as @noescape00 said, every time we run the node, it gets built anyway. I really think allowing this is a good way to test various consensus scenarios without modifying the existing networks. We want to test consensus logic  here, not network configuration.

That said, a json serialized version of network is also a really good suggestion but it will force us to make network serializable or we have to create INetwork interface so that we can implement a serializable version of Network

src/Stratis.StratisD.Test/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    public class Program
+    {
+        public static async Task Main(string[] args)
+        {
+            try
+            {
+                Network network = null;
+                if (args.Contains("-network=StratisPosTestNetwork"))
+                    network = new StratisPosTestNetwork();
```
dangershony: probably its not correct to call it Stratis maybe just `PosTestNetwork` and `PowTestNetwork`?

dangershony: Or maybe even add the word Generated `GeneratedPowTestNetwork`

dangershony:
Generally I am fine with this because its all in test projects (and this should help the testers to easy bootstrap a networks)

Alternatively we can create a list of knowen test networks in advance for the test effort.

bokobza:
How can you test the stratis network if the network you use for those tests is different from the stratis network?

If anything, why don't you inherit from Stratis Main and change what you need to change?

bokobza:
The way it is now, it's impossible to see, at a glance, what the difference is with the Stratis/Bitcoin networks.

fassadlr:
@bokobza I just want to re-iterate a few points again.

1. These networks are here to test consensus logic and how the node behaves with various consensus configurations. They are specifically for the Magic network and not to be used by integration tests. Therefore it doesn't matter that they aren't inheriting from Stratis. 
2. I don't want to be forced to include Stratis.Bitcoin.Networks here as it will kind of defeat the purpose. I want this to be completely isolated from the current solution.
3. Lukasz needs to be able to test the above scenarios and for obvious reasons cant connect to the test net to do this, as for example, on Stratis test, the node is already past the mine-able height.

noescape00:
@fassadlr @bokobza @dangershony what is the status of this PR? Do we want this to go in? 

I'd personally go with a different approach and make networks more modular (separate PoS stuff from PoW) and come up with some kind of network modules where you can specify what do you need while building the network. Only base things like base58 prefixes, ports and so on will go to base class, rest will be implemented as modules.

Issue #1791, opened 2018-08-01
Create proven headers rules and add to POS Network class + Test


noescape00:
Also consider issue #979.

dangershony:
For this task we have to be careful not to mix PH in to the low level code (for example PH are not needed for Bitcoin)  This means PH should be part of rules or an override of one of the validators.

noescape00:
See for additional info: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1792

noescape00:
Depends on: https://github.com/stratisproject/StratisBitcoinFullNode/issues/1790

PR #2279, opened 2018-09-25
[PH] Proven header validation rules - Part 1 

fenix2222:
This is a first PR for a series of PRs for proven header rules as per issue #1791 

Includes only two rules size and version rules.

Registration of rules will happen in one of the last PRs.

There are also few TODOs in this PR to be actioned in later PRs:
Set activation height (atm it is a fake one)
Set max proven header deserialisation size (atm it is = MaxBlockSerializationSize)


src/NBitcoin/ProvenBlockHeader.cs
```diff
@@ -31,6 +31,9 @@ namespace NBitcoin
     /// </remarks>
     public class ProvenBlockHeader : PosBlockHeader
     {
+        /// <inheritdoc />
+        public override int CurrentVersion => 8;
```
Aprogiena: why this?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Consensus;
+using Stratis.Bitcoin.Consensus.Rules;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
+{
+    /// <summary>
+    /// Rule to check if the serialized size of the proven header is less than 1 000 512 bytes.
```
Aprogiena: thinking about this little more, it might be a better idea to check size of each component
- coinstake size is at max 1000000
- block signature i cant remember, but something like 80 ?
- merkle proof no more than 512

reason being to prevent accepting merkle proof with size 1mb itself

src/NBitcoin/ProvenBlockHeader.cs
noescape00: That's my idea. 
So other stratis-only components can check if the header is PH without typecast.

noescape00: @fenix2222 had a call with Apro. I't a bad idea indeed because version goes to block's hash and we can't easily change it. Sorry, my bad. 
Could you please remove it?

fenix2222: Sure, I can remove it, but what are we checking with version rule then? Should we just use `StratisHeaderVersionRule`?

src/NBitcoin/ProvenBlockHeader.cs
```diff
@@ -63,6 +61,15 @@ public class ProvenBlockHeader : PosBlockHeader
         /// </summary>
         public BlockSignature Signature => this.signature;
 
+        /// <summary>Gets the size of the merkle proof in bytes, the header must be serialized for this property to be set.</summary>
```
noescape00: `must be serialized or deserialized`

same for other 2

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRule.cs
```diff
@@ -0,0 +1,58 @@
...
+using Stratis.Bitcoin.Networks;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
+{
+    /// <summary>
+    /// Base rule to be used by all proven header validation rules. 
+    /// </summary>
+    /// <seealso cref="Stratis.Bitcoin.Consensus.Rules.HeaderValidationConsensusRule" />
+    public abstract class ProvenHeaderValidationConsensusRule : HeaderValidationConsensusRule
```
noescape00: ProvenHeaderValidationConsensusRule => ProvenHeaderValidationConsensusRuleBase

src/NBitcoin/ProvenBlockHeader.cs
noescape00: >Sure, I can remove it, but what are we checking with version rule then? Should we just use StratisHeaderVersionRule?

Yeah

src/NBitcoin/ProvenBlockHeader.cs
```diff
@@ -80,19 +87,34 @@ public ProvenBlockHeader(PosBlock block)
...
+            this.MerkleProofSize = GetProcessedBytes(stream) - baseByteSize;
+
             stream.ReadWrite(ref this.signature);
+            this.SignatureSize = GetProcessedBytes(stream) - baseByteSize - this.MerkleProofSize;
+
             stream.ReadWrite(ref this.coinstake);
+            this.CoinstakeSize = GetProcessedBytes(stream) - baseByteSize - this.MerkleProofSize - this.SignatureSize;
+        }
+
+        private static long GetProcessedBytes(BitcoinStream stream)
```
Aprogiena: why static?

Aprogiena: also why not to the stream class itself?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -0,0 +1,44 @@
...
+        public override void Run(RuleContext context)
+        {
+            Guard.NotNull(context.ValidationContext.ChainedHeaderToValidate, nameof(context.ValidationContext.ChainedHeaderToValidate));
+
+            if (context.SkipValidation || !this.IsProvenHeaderActivated(context))
+                return;
+
+            var header = (ProvenBlockHeader)context.ValidationContext.ChainedHeaderToValidate.Header;
+
+            if (header.MerkleProofSize == null || header.MerkleProofSize > PosConsensusOptions.MaxMerkleProofSerializedSize)
```
Aprogiena: brackets everywhere

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRuleBase.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Determines whether proven headers are activated based on the proven header activation height and applicable network.
+        /// </summary>
+        /// <param name="context">The rule context.</param>
+        /// <returns>
+        /// <c>true</c> if proven header height is past the activation height for the corresponding network;
+        /// otherwise, <c>false</c>.
+        /// </returns>
+        public bool IsProvenHeaderActivated(RuleContext context)
```
Aprogiena: ok, but have you considered having just height as parameter? to me it would be cleaner logic

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRuleBase.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            return false;
+        }
+
+        /// <summary>
+        /// Determines whether header is a proven header.
+        /// </summary>
+        /// <param name="context">The rule context.</param>
+        /// <returns><c>true</c> if header is a <see cref="ProvenBlockHeader"/>.</returns>
+        public bool IsProvenHeader(RuleContext context)
```
Aprogiena: similarly here i'd expect header to be parameter, any particular reason ?

src/NBitcoin/ProvenBlockHeader.cs
fenix2222: Static because it is a helper function and it is not dealing with the instance itself. I don't really mind either way, it just seems logical for functions of this type

src/NBitcoin/ConsensusOptions.cs
```diff
@@ -112,6 +112,33 @@ public class PosConsensusOptions : ConsensusOptions
         /// <summary>Coinstake minimal confirmations softfork activation height for the testnet.</summary>
         public const int CoinstakeMinConfirmationActivationHeightTestnet = 436000;
 
+        /// <summary>
```
ferdeen: Can remove the word 'the'.

ferdeen: Can remove the word 'the'.

ferdeen: Can remove the word 'the'.

ferdeen: Can remove the word 'the'.

src/NBitcoin/ConsensusOptions.cs
```diff
@@ -112,6 +112,33 @@ public class PosConsensusOptions : ConsensusOptions
...
+        /// Proven headers activation height for the mainnet.
```
ferdeen: Can remove the word 'the'.

src/NBitcoin/ConsensusOptions.cs
```diff
@@ -112,6 +112,33 @@ public class PosConsensusOptions : ConsensusOptions
...
+        /// TODO: Set it to the real value once it is known.
+        /// </summary>
+        public const int ProvenHeadersActivationHeightMainnet = 20_000_000;
+
+        /// <summary>
+        /// Proven headers activation height for the testnet.
```
ferdeen: Same as above.

noescape00:
@Aprogiena @ferdeen I'm about to merge it. Any objections?

src/NBitcoin/ConsensusOptions.cs
```diff
@@ -106,12 +106,39 @@ public ConsensusOptions()
...
 
-        /// <summary>Coinstake minimal confirmations softfork activation height for the testnet.</summary>
+        /// <summary>Coinstake minimal confirmations softfork activation height for testnet.</summary>
         public const int CoinstakeMinConfirmationActivationHeightTestnet = 436000;
 
+        /// <summary>
+        /// Proven headers activation height for mainnet.
+        /// TODO: Set it to the real value once it is known.
+        /// </summary>
+        public const int ProvenHeadersActivationHeightMainnet = 20_000_000;
```
dangershony: Lets go over again why we need activation?

We should just know that a peer has or has not PH and if it has we ask PH from it otherwise we don't.
Eventually when no more X nodes we will only use PH.

I guess the issue is that then an attacker can just not send PH and fake blocks so at some point we can push in the code a condition that we need to have at least 50% outbound as PH nodes

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -0,0 +1,46 @@
...
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
+{
+    /// <summary>
+    /// Rule to check if the serialized sizes of the proven header components, such as merkle proof (max 512 bytes),
+    /// signature (max 80 bytes) and coinstake (max 1,000,000 bytes), do not exceed maximum possible size allocation.
+    /// </summary>
+    /// <seealso cref="ProvenHeaderValidationConsensusRuleBase" />
+    public class ProvenHeaderSizeRule : ProvenHeaderValidationConsensusRuleBase
```
dangershony: `ProvenHeaderValidationConsensusRuleBase` wait what?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -0,0 +1,46 @@
...
+            Guard.NotNull(context.ValidationContext.ChainedHeaderToValidate, nameof(context.ValidationContext.ChainedHeaderToValidate));
+
+            int height = context.ValidationContext.ChainedHeaderToValidate.Height;
+
+            if (context.SkipValidation || !this.IsProvenHeaderActivated(height))
+                return;
+
+            var header = (ProvenBlockHeader)context.ValidationContext.ChainedHeaderToValidate.Header;
+
+            if ((header.MerkleProofSize == null) || (header.MerkleProofSize > PosConsensusOptions.MaxMerkleProofSerializedSize))
```
dangershony: Do we use this any other place then in this class?
`PosConsensusOptions.MaxMerkleProofSerializedSize`

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRuleBase.cs
```diff
@@ -0,0 +1,58 @@
...
+using Stratis.Bitcoin.Networks;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
+{
+    /// <summary>
+    /// Base rule to be used by all proven header validation rules. 
+    /// </summary>
+    /// <seealso cref="Stratis.Bitcoin.Consensus.Rules.HeaderValidationConsensusRule" />
+    public abstract class ProvenHeaderValidationConsensusRuleBase : HeaderValidationConsensusRule
```
dangershony: Ah ok fine `ProvenHeaderValidationConsensusRuleBase ` is part of `HeaderValidationConsensusRule`

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRuleBase.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Determines whether proven headers are activated based on the proven header activation height and applicable network.
+        /// </summary>
+        /// <param name="height">The height of the header.</param>
+        /// <returns>
+        /// <c>true</c> if proven header height is past the activation height for the corresponding network;
+        /// otherwise, <c>false</c>.
+        /// </returns>
+        public bool IsProvenHeaderActivated(int height)
+        {
+            if (this.PosParent.Network is StratisMain)
```
dangershony: Right... this becomes a StratisNetwork rule only, in this case we need to change the rule to be `StratisProvenHeaderValidationConsensusRuleBase ` and only inject it for stratis network.

This is probably wrong.
Any other network then stratis PH will be activated by default.

So the solution is to use overrides, but this is an abstract class so the logic `IsProvenHeaderActivated` should not be here at all but in some rule that is CheckProvenHeaderActivation (or something).

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <seealso cref="ProvenHeaderValidationConsensusRuleBase" />
+    public class ProvenHeaderSizeRule : ProvenHeaderValidationConsensusRuleBase
+    {
+        public override void Run(RuleContext context)
+        {
+            Guard.NotNull(context.ValidationContext.ChainedHeaderToValidate, nameof(context.ValidationContext.ChainedHeaderToValidate));
+
+            int height = context.ValidationContext.ChainedHeaderToValidate.Height;
+
+            if (context.SkipValidation || !this.IsProvenHeaderActivated(height))
```
dangershony: I suggest you split this in two classes `ProvenHeaderSizeRule` and `StratisProvenHeaderSizeRule`.
The later will have the check `IsProvenHeaderActivated` and then call the base if yes.

src/NBitcoin/ConsensusOptions.cs
Aprogiena: after activation heighr node will require proven headers

noescape00: >Eventually when no more X nodes we will only use PH.

Node will use normal headers for checkpointed blocks since we don't need to validate those headers.

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
noescape00: It will be used in other PH rules in @fenix2222 's follow up prs

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
noescape00: We probably can make it shorter: `ProvenHeaderRuleBase`. Sounds good?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderValidationConsensusRuleBase.cs
fenix2222: I am confused with this. What exactly should we do?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
fenix2222: Problem is that you can't have double inheritance. i.e. if `StratisProvenHeaderSizeRule` inherits from `ProvenHeaderSizeRule` it cannot then also inherit a `StratisProvenHeaderRuleBase` (the one you want to keep separate with a `IsProvenHeaderActivated` method).

fenix2222: I can just rename 
`ProvenHeaderRuleBase` => `StratisProvenHeaderRuleBase`
and
`ProvenHeaderSizeRule` => `StratisProvenHeaderSizeRule`

If we ever have other POS implementations, we can action it then

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderSizeRule.cs
```diff
@@ -10,8 +10,8 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
     /// Rule to check if the serialized sizes of the proven header components, such as merkle proof (max 512 bytes),
     /// signature (max 80 bytes) and coinstake (max 1,000,000 bytes), do not exceed maximum possible size allocation.
     /// </summary>
-    /// <seealso cref="ProvenHeaderValidationConsensusRuleBase" />
-    public class ProvenHeaderSizeRule : ProvenHeaderValidationConsensusRuleBase
+    /// <seealso cref="ProvenHeaderRuleBase" />
+    public class ProvenHeaderSizeRule : ProvenHeaderRuleBase
```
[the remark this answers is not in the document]
fenix2222: @dangershony But the whole purpose of having that base rule was to contain that `IsProvenHeaderActivated` helper property so that all rules related to PH can use it, otherwise I have to duplicate this code into every rule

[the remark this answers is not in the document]
fenix2222: @dangershony Updated as per our discussion

src/Stratis.Bitcoin.Features.SmartContracts/Networks/SmartContractsRegTest.cs
```diff
@@ -39,7 +39,8 @@ public SmartContractsRegTest()
                 maxBlockBaseSize: 1_000_000,
                 maxStandardVersion: 2,
                 maxStandardTxWeight: 100_000,
-                maxBlockSigopsCost: 20_000
+                maxBlockSigopsCost: 20_000,
+                provenHeadersActivationHeight: 10_000_000 // TODO: Set it to the real value once it is known.
```
dangershony: SC wont have proven headers so this could just be zero but no big deal this will be fixed later.

PR #2284, opened 2018-09-26
[SC] Blacklist new object creation

### LINKED ISSUE #1921: [SC] Contracts shouldn't be able to be instantiated in code via their constructors.
Contracts being instantiated in code via their constructors is nonsensical - contracts should only be created via the `.Create<>` method on the `SmartContract` class, which will create them in persistent storage permanently. 

We should likely prevent this in the Format Validator.

rowandh:
Closes #1921. Prevents all usages of `newobj` except when it is being used to create value types.


rowandh:
All SC unit/integration tests passing except for some unrelated specs.

PR #2288, opened 2018-09-26
[SC] Remove TransferFundsToContract

### LINKED ISSUE #2283: [SC] Remove TransferFundsToContract
No longer used and can be removed from Stratis.SmartContracts.

Will require a version increase as it removes functionality.

rowandh:
Closes #2283.


codingupastorm:
Are we going to increase the version every time we update the functionality then? Could be a lot between releases

PR #2289, opened 2018-09-26
[SC] Fix mempool op spend rule and stop using ToOps

### LINKED ISSUE #2281: [SC] Stop using ToOps() from NBitcoin for SC serialization and opcode checking
When given contract bytecode inside script, ToOps() and other related functionality behaves erratically, as Script and related classes were never expecting to interact with smart contract code.

All of the smart contract checks should instead check the value of the first byte only, which tells the node how to handle this transaction.

rowandh:
Closes #2281.

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts.Tests/SmartContractScriptTests.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts.Tests/SmartContractScriptTests.cs
@@ -8,39 +8,71 @@ namespace Stratis.Bitcoin.Features.SmartContracts.Tests
         [Fact]
         public void IsSmartContractExec_CallContract_Success()
         {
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.SmartContracts.Tests/SmartContractScriptTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            var script = new NBitcoin.Script(new byte[] { (byte)ScOpcodeType.OP_SPEND });
+
+            Assert.True(script.IsSmartContractSpend());
+        }
+
+        [Fact]
+        public void IsSmartContractSpend_Failure()
+        {
+            var script = new NBitcoin.Script(new byte[] { 0xFF, (byte)ScOpcodeType.OP_SPEND, 0xFF });
```
codingupastorm: 💯

PR #2291, opened 2018-09-26
CoinView: moved evict under lock

noescape00:
Minor perf optimization. 

Instead of exiting async lock and then entering it inside EvictAsync I've moved eviction under the lock and changed the method to be EvictLocked. 

Also fixed TODO

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
@@ -110,6 +110,8 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 
         private CachePerformanceSnapshot latestPerformanceSnapShot;
 
+        private readonly Random random;
+
         /// <summary>
         /// Initializes instance of the object based on DBreeze based coinview.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -316,23 +319,19 @@ public async Task FlushAsync(bool force = true)
...
-            using (await this.lockobj.LockAsync().ConfigureAwait(false))
+            foreach (KeyValuePair<uint256, CacheItem> entry in this.cachedUtxoItems.ToList())
             {
-                // TODO: Do not create new random source every time.
-                var rand = new Random();
-                foreach (KeyValuePair<uint256, CacheItem> entry in this.cachedUtxoItems.ToList())
+                if (!entry.Value.IsDirty && entry.Value.ExistInInner)
                 {
-                    if (!entry.Value.IsDirty && entry.Value.ExistInInner)
+                    if (this.random.Next() % 3 == 0)
```
Aprogiena: brackets

noescape00: fixed

dangershony:
Looks good but I didn't run tests.

PR #2292, opened 2018-09-26
Fix reorg bug and some CM tests

dangershony:


dangershony:
Note I still need to run all integration tests.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -196,5 +246,198 @@ public void ConsensusManager_Fork_Occurs_Node_Gets_Disconnected_Due_To_MaxReorgV
...
+                TestHelper.Disconnect(syncer, minerB);
+
+                // Miner A and syncer continues to mine to height 20.
+                TestHelper.MineBlocks(minerA, 10);
+                TestHelper.WaitLoop(() => TestHelper.AreNodesSynced(syncer, minerA));
+
+                // Miner B continues to mine to height 30 on a new and longer chain whilst disconnected.
+                TestHelper.MineBlocks(minerB, 20);
+
+                // Inject a rule that will fail at block 15 of the new chain
```
Aprogiena: 11?

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -196,5 +246,198 @@ public void ConsensusManager_Fork_Occurs_Node_Gets_Disconnected_Due_To_MaxReorgV
...
+
+                var minerA = builder.CreateStratisPosNode(this.posNetwork);
+                var syncer = builder.CreateStratisPosNode(syncerNetwork);
+
+                builder.StartAll();
+
+                minerA.NotInIBD().WithWallet();
+                syncer.NotInIBD();
+
+                // MinerA mines to height 10.
```
Aprogiena: 11?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -685,6 +685,14 @@ private async Task<ConnectBlocksResult> ConnectChainAsync(ChainedHeader newTip,
             {
                 using (this.performanceCounter.MeasureBlockConnectionFV())
                 {
+                    if (blockToConnect.ChainedHeader.Block == null)
+                    {
+                        // In some rare cases block can be null.
+                        // This can happen after a reorg failed and
+                        // we are trying to reconnect the old chain
+                        blockToConnect.ChainedHeader.Block = blockToConnect.Block;
```
Aprogiena: i am now unconvinced this is good, can we make call tomorrow with ivan?

dangershony: Ok but is it critical or perhaps we push and make the fix in another PR, because right now node is unstable and this fixes it.

Aprogiena: i understand, but i think we need to fix correctly

dangershony: Lets discuss tomorrow.

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -261,17 +262,21 @@ public TransactionNode(Transaction tx)
         public static void Connect(CoreNode from, CoreNode to)
         {
             from.CreateRPCClient().AddNode(to.Endpoint, true);
```
ferdeen: I know this code hasn't change as part of this work, but can the logic check if the nodes are already connected - or wouldn't that really matter ?

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -29,14 +36,57 @@ public override int GetStakeMinConfirmations(int height, Network network)
...
+            }
+        }
+
+        public class BitcoinOverrideRegTest : BitcoinRegTest
+        {
+            public BitcoinOverrideRegTest() : base()
+            {
+                this.Name = Guid.NewGuid().ToString();
+            }
+        }
```
ferdeen: Rather then create override type classes, can't the derived classes include an extra ctor that passes in/sets Name ?

Otherwise any future tests would need to follow this code pattern just to set the Name.

src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
dangershony: hmm well as this is a test we should know exactly who is connected (I could add a test ensure not connected perhaps?)

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
dangershony: we want to keep Network as immutable as possible ;)

Aprogiena:
briefly looks good, just make sure cht tests pass

dangershony:
@Aprogiena you seem to be correct but other things are tested in this flow so I am not sure.

src/Stratis.Bitcoin.Tests/Consensus/ChainedHeaderTreeTest.cs
```diff
@@ -2418,7 +2418,7 @@ public void PresentHeaders_BlocksDownloaded_UnconsumedBlocksDataBytes_Equals_Ser
         /// have no block pointers and block data availability == headers only.
         /// </summary>
         [Fact]
-        public void NodeHasReorg_ReorganisedHeaders_HaveNoBlockPointers_BlockDataAvailability_SetToHeadersOnly()
+        public void NodeHasReorg_ReorganisedHeaders_BlockDataAvailability()
```
Aprogiena: still the comment is wrong

dangershony: ok test is gone!

PR #2300, opened 2018-09-28
CM: changed constant

noescape00:
MaxBlocksToAskFromPuller changed from 30k to 10k


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -32,7 +32,7 @@ public class ConsensusManager : IConsensusManager
         private const long ConsumptionThresholdBytes = MaxUnconsumedBlocksDataBytes / 10;
 
         /// <summary>The maximum amount of blocks that can be assigned to <see cref="IBlockPuller"/> at the same time.</summary>
-        private const int MaxBlocksToAskFromPuller = 30000;
+        private const int MaxBlocksToAskFromPuller = 10000;
```
dangershony: yeah seems right we have way too many blocks in memory as the bottle neck is validation.

Issue #1830, opened 2018-08-03
[SC] Serialization of method parameters should be simpler.


codingupastorm:
When sending parameters in to a method inside of a transaction, we should just send them in string form, rather than the current form that prefixes the items with a number etc.

When working out which method to call, use the parameters on the class via reflection to work out which types to deserialize to.

If ambiguous, should just select the first one. This is a risk we will warn developers of at the compiler level.

rowandh:
Investigate use of `ContractPrimitiveSerializer` to perform param serialization as bytes rather than a string.

PR #2302, opened 2018-09-28
[SC] Improve method param serialization

rowandh:
Beginning of #1830.

* Separate string-based method param serialization out of ContractCarrier.
* Remove ContractCarrier. It has been replaced by ContractTxData.


src/Stratis.SmartContracts.Executor.Reflection/CallDataSerializer.cs
```diff
@@ -13,13 +14,13 @@ public class CallDataSerializer : ICallDataSerializer
         // TODO this is ugly but there is poor DI support for rules so we can't inject it yet
         public static ICallDataSerializer Default = new CallDataSerializer(new MethodParameterSerializer());
 
-        private readonly IMethodParameterSerializer methodParamSerializer;
+        public IMethodParameterSerializer MethodParamSerializer { get; }
```
codingupastorm: This feels a little lazy

rowandh: It is. It imagined it could be fixed/removed in part 2, when the param serialization is changed.

PR #2320, opened 2018-09-28
Increase flush interval

dangershony:


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -59,7 +59,7 @@ private class CacheItem
 
         /// <summary>Length of the coinview cache flushing interval in seconds.</summary>
         /// <seealso cref="lastCacheFlushTime"/>
-        public const int CacheFlushTimeIntervalSeconds = 60;
+        public const int CacheFlushTimeIntervalSeconds = 600;
```
ferdeen: If your increasing the flush time X 10 - maybe change the constant to CacheFlushTimeIntervalInMinutes =  10;

Unless it may come back down to seconds in the future.

dangershony:
>We have 100k cache items and we evict prunable onces   

We don't evict prunable we evict non dirty cache items 
`if(!entry.Value.IsDirty && entry.Value.ExistInInner)`


>If we flush every 10m we will have 99% of cache being non prunable   

I am not sure I get the argument, more memory will be occupied by items pending flush indeed however it wont make disable caching (that depends on weather a UTXO is in disk on in memory)

However I agree perhaps if increasing intervals we should also increase memory size.
This goes back to the argument of making coinview size in memory configurable .

dangershony:
>you also won't get any boost out of it

@Aprogiena I don't think there is a correlation between frequency of flush to length the flush will take,
But agree this change needs to show some stats before we can say its improving per.

I will try to provide some stats.

dangershony:
>it won't be a cache but a batch

Same thing in this context as every batched item is potentially a cached item.

Aprogiena:
> @Aprogiena I don't think there is a correlation between frequency of flush to length the flush will take,

i think there is correlation ...

noescape00:
>Same thing in this context as every batched item is potentially a cached item.

Yeah, but with this approach we will effectively keep only the last N items when for prefetch to work properly we need old items to be loaded a few seconds before they are needed and we don't wan't them to be deleted right after they were loaded.

bokobza:
There is no description to this PR. Can we have an explanation that states why this needs to be done? Otherwise how would you expect anyone to judge whether it's a good call or not?

noescape00:
@dangershony are we ok to close this PR or you are convinced that it will be to benefit and will get metrics?

bokobza:
I'd appreciate a response to this: https://github.com/stratisproject/StratisBitcoinFullNode/pull/2320#issuecomment-426562204

dangershony:
>Can we have an explanation that states why this needs to be done  

Yeah the idea is to reduce the frequency of flushing to disk, every time we flush we slow down the node, the idea is that UTXOs that are currently in memory and are spent in a block before the flush will not need to go to disk (however @Aprogiena is right we will still flush the rewind data) so we will just have bigger flush time when we do go to disk, the question is will it be proportionally bigger to the amount of time we waited.

Additionally it may means we have more items in memory that are pending flush (and so this cannot be deleted when we purge memory, so this change should probably come with an increase of cached items)

I agree such a change should come with some metrics and trying out several configurations and measuring perf.

bokobza:
Thanks for the explanation. I'm not convinced that this should be merged without proper testing.

dangershony:
Ill close this now its just useless noise

PR #2322, opened 2018-09-29
changes to ConsensusManagerPerformance Counter/Snapshot

MithrilMan:
simplified the ConsensusManagerPerformance Counter/Snapshot
fixed a bug in Tick to MS conversion

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
+++ b/src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
@@ -1,18 +1,27 @@
 ﻿using System;
 using System.Text;
+using System.Threading;
 
 namespace Stratis.Bitcoin.Consensus.PerformanceCounters.ConsensusManager
 {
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
```diff
@@ -34,18 +43,24 @@ public override string ToString()
...
-            this.TotalExecutionsCount = 0;
-            this.TotalDelayTicks = 0;
+            this.totalExecutionsCount = 0;
+            this.totalDelayTicks = 0;
         }
 
         public double GetAvgExecutionTimeMs()
         {
-            return Math.Round((this.TotalDelayTicks / (double)this.TotalExecutionsCount) / 1000.0, 4);
+            return Math.Round((this.totalDelayTicks / (double)this.totalExecutionsCount) / 10000.0, 4);
```
zeptin: This is a nice find, I see that `StopwatchDisposable` utilises `TimeSpan.Ticks` on disposal, so 10 000 would indeed appear to be the correct conversion factor to give milliseconds.

PR #2323, opened 2018-09-29
Minor test fix

zeptin:
1. Some new RPCs have been added, which broke a relatively pointless test.

2. Tests using `bitcoind` on Windows are highly sensitive to path length. While this cannot be solved from within the tests, it can at least be made to fail rapidly, rather than wasting 5 minutes of execution time per test in some cases.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Stratis.Bitcoin.Tests.Common;
 
@@ -52,6 +53,15 @@ namespace Stratis.Bitcoin.IntegrationTests.Common.Runners
         {
             string logMode = Debugger.IsAttached ? "-debug=net" : string.Empty;
             TimeSpan duration = TimeSpan.FromSeconds(15);
+
+            // The complete path bitcoind uses to locate (e.g.) the block files consists of the source code build folder,
+            // the relative path within the test case folders, and the bitcoind network-specific path to its block database
+            // This adds roughly 37 characters onto the full data folder path: \regtest\blocks\index/MANIFEST-000001
+
+            // By throwing here we avoid a pointless 5-minute wait for bitcoind to start up (it will 'start' and then soon
+            // crash, which results in the getblockhash RPC call timing out later on in the startup sequence).
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && (new FileInfo(this.DataFolder).FullName.Length > 222))
+                throw new Exception("Path is too long for bitcoind to function.");
 
             TestHelper.WaitLoop(() =>
             {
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
```diff
@@ -53,6 +54,15 @@ public override void Start()
             string logMode = Debugger.IsAttached ? "-debug=net" : string.Empty;
             TimeSpan duration = TimeSpan.FromSeconds(15);
 
+            // The complete path bitcoind uses to locate (e.g.) the block files consists of the source code build folder,
```
ferdeen: e.g in brackets ? Not too sure what that means.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
```diff
@@ -53,6 +54,15 @@ public override void Start()
...
+            // the relative path within the test case folders, and the bitcoind network-specific path to its block database
+            // This adds roughly 37 characters onto the full data folder path: \regtest\blocks\index/MANIFEST-000001
+
+            // By throwing here we avoid a pointless 5-minute wait for bitcoind to start up (it will 'start' and then soon
+            // crash, which results in the getblockhash RPC call timing out later on in the startup sequence).
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && (new FileInfo(this.DataFolder).FullName.Length > 222))
```
ferdeen: Magic number - 222. Probably better as a constant.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSteps.cs
```diff
@@ -416,7 +416,7 @@ private void a_full_list_of_available_commands_is_returned()
         {
             var commands = JsonDataSerializer.Instance.Deserialize<List<RpcCommandModel>>(this.responseText);
 
-            commands.Count.Should().Be(16);
+            commands.Count.Should().Be(20);
```
ferdeen: Not related to this change, but this test could use reflection to figure out the rpc commands.  In the same manner as this : 

https://github.com/stratisproject/StratisBitcoinFullNode/blob/68eb2c25932e2095fcae5788c63740d1f008b2ef/src/Stratis.Bitcoin.IntegrationTests/API/AllControllersTests.cs#L17

zeptin: I guess it depends on whether we want to explicitly list all the known RPCs to test against.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
zeptin: You can read it as "The complete path bitcoind uses to locate, for example, the block files,".

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/BitcoinCoreRunner.cs
zeptin: I'm not sure it's even worth it, this check is only added to 'fail fast' in case the developer isn't aware that having paths that are too long causes bitcoind to not start up properly.

PR #2325, opened 2018-10-01
Add missing nuget package reference to Tracer.4NLog.Fody

rowandh:
Master won't build locally without this.


noescape00:
Tracer.4NLog.Fody is nother adapter. It shouldn't be required because we already have a custom adapter. 

But let me recheck and build master

noescape00:
However on a server it won't build. Hm. Let me investigate this.

PR #2327, opened 2018-10-01
[SC] Use shared nonce state during tx execution

### LINKED ISSUE #2259: [SC] Specify 'nonce' behaviour when contract execution fails.
When an internal create fails, should the nonce still increment? 

(I think yes)

Define expected behaviour and test.

rowandh:
Closes #2259.

Uses a shared `NonceGenerator` during the lifetime of `State`, which corresponds to the lifetime of a transaction execution. If `State` is snapshotted, the snapshot will use the original `NonceGenerator` ensuring that even if execution fails, nonces will still increment.


rowandh:
All SC tests + SC integration tests pass.

PR #2328, opened 2018-10-01
[SC] Serialize defaults to byte0

codingupastorm:


codingupastorm:
Working on as part of bigger issues

Issue #2307, opened 2018-09-28
Stratis.PoAChainD feature startup


noescape00:
On startup following checks should be done:

1. fedpubkeys are configured in the network. There should be at least 1 key.
2. If there is a keypair file in the folder we need to check if pubkey from it is included in fedpubkeys in network.
3. Check if private key in keypair corresponds to public key (in case keypair file exist).
4. Show in log pub key loaded for node if there is one

In case any of those checks fail we should spawn an error and stop the node.

PR #2335, opened 2018-10-01
PoA feature startup checks

noescape00:
Fix: #2307


src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
```diff
@@ -0,0 +1,60 @@
...
+            this.network = network as PoANetwork;
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+        }
+
+        public void Initialize()
+        {
+            this.LoadKey();
+
+            if (this.FederationMemberKey != null)
```
dangershony: If you are a federation member how can this be null?

src/Stratis.Bitcoin.Features.PoA/KeyTool.cs
```diff
@@ -29,6 +37,9 @@ public void SavePrivateKey(Key privKey, string path)
 
         public Key LoadPrivateKey(string path)
         {
+            if (!File.Exists(path))
```
dangershony: Please check how we load files in the wallet.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
```diff
@@ -14,6 +14,13 @@ public class PoANetwork : Network
         /// <summary> The default name used for the Stratis configuration file. </summary>
         private const string NetworkDefaultConfigFilename = "poa.conf";
 
+        /// <summary>Public keys of all federation members.</summary>
+        /// <remarks>
+        /// Blocks that are not signed with private keys that correspond
+        /// to public keys from this list are considered to be invalid.
+        /// </remarks>
+        public List<PubKey> FederationPublicKeys { get; }
```
dangershony: This should proabably go in PoaConsensusOptions no?

dangershony: Actually I am not so sure.

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Configuration;
+
+namespace Stratis.Bitcoin.Features.PoA
+{
+    public class FederationManager
```
dangershony: Why do you need a manager I wonder.

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
noescape00: When the node loads it should decide if we are federation member or not. It does that by loading a key. if the key is not there- we are normal node. 
So here we tried to load the key and if it's null we set FederationMemberKey to null and IsFederationMember to false. Otherwise we set 2 values. 

Later miner will check if we are fed member and start mining if we are. 

that way we don't need much changes in the code or user interaction to become a fed member.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
noescape00: No, that's not a part of consensus. 
We can have several PoA networks with same consensus rules but different federations. Sj it's a part of network settings, not consensus.

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
noescape00: To check if we are federation and provide this info to other components that may or may not work based on that information (or to initialize some components for being federation member and not initialize if normal node. miner class- for example)

Issue #2331, opened 2018-10-01
Make the dbreeze serialization instance based and plug decerializaiton components 


dangershony:
The `DBreezeSerializer` class could have injected a serialization provider, this can then have serialization types injected to that provider and how they get handled.

This will make it easier to add new serialization types at the higher level.

dangershony:
However this still does not solve serialization between different networks as the serializers are static and we have no way to know which dbreeze engine instance is requesting deserialization.

dangershony:
This was fixed by @rowandh 
https://github.com/stratisproject/StratisBitcoinFullNode/commit/a3654a694339cb0dff4eb845bff56247a6ff32cf

PR #2343, opened 2018-10-02
deserialise any IBitcoinSerializable

monsieurleberre:
This is to allow us deserialise any types that implements IBitcoinSerialisable
This is a smaller change than the one suggested in #2331 for now.
It should help in https://github.com/stratisproject/FederatedSidechains/pull/118

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Utilities/DBreezeTest.cs
+++ b/src/Stratis.Bitcoin.Tests/Utilities/DBreezeTest.cs
@@ -3,9 +3,9 @@ using System.Linq;
 using System.Text;
 using DBreeze;
 using DBreeze.DataTypes;
+using FluentAssertions;
 using NBitcoin;
 using NBitcoin.BitcoinCore;
-using Stratis.Bitcoin.Networks;
 using Stratis.Bitcoin.Tests.Common;
 using Stratis.Bitcoin.Utilities;
 using Xunit;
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Tests/Utilities/DBreezeTest.cs
```diff
@@ -137,6 +137,28 @@ public void DeserializerWithNotSupportedClassThrowsException()
             });
         }
 
+        private class UnknownBitcoinSerialisable : IBitcoinSerializable
```
zeptin: It looks better if we consistently keep the American spelling of serializable I think, as that has less potential for confusion.

monsieurleberre: I can't decide if I should please you or the code policy enforcement administration (from codereview-policy.md, l.43):
* **Typos and grammar**: make sure the code reads smoothly and is in British English.

dangershony:
Looks good but we need to sync networks as this touches low level parts of the code

monsieurleberre:
OK let me try

src/Stratis.Bitcoin.Tests/Utilities/DBreezeTest.cs
zeptin: That is already not 100% achievable unless we substantially modify NBitcoin. But as this is just a test case there is no need to change it.

bokobza: UK spelling would make more sense for us, since we're a UK company ;-)

PR #2347, opened 2018-10-02
improved the logging of starting tasks

MithrilMan:
Previous log had the problem of printing that the task was Starting even if delayStart was set

As a result of this, when you launched the node you were reading `Periodic peer flush starting.` and it seemed that it took so long to execute, when actually it had a delayStart of 5 minutes.

with current changes, in case of delay you see something like
`Periodic peer flush starting in 300 seconds.`

and when it start for real, you'll see the previous `Periodic peer flush starting.`

### DIFF
--- a/src/Stratis.Bitcoin/Utilities/AsyncLoop.cs
+++ b/src/Stratis.Bitcoin/Utilities/AsyncLoop.cs
@@ -119,11 +119,11 @@ namespace Stratis.Bitcoin.Utilities
                 {
                     if (delayStart != null)
                     {
-                        this.logger.LogInformation($"{this.Name} starting in {delayStart.Value.TotalSeconds} seconds.");
+                        this.logger.LogInformation("{0} starting in {1} seconds.", this.Name, delayStart.Value.TotalSeconds);
                         await Task.Delay(delayStart.Value, cancellation).ConfigureAwait(false);
                     }
 
-                    this.logger.LogInformation($"{this.Name} starting.");
+                    this.logger.LogInformation("{0} starting.", this.Name);
 
                     if (this.RepeatEvery == TimeSpans.RunOnce)
                     {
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Utilities/AsyncLoop.cs
```diff
@@ -115,11 +115,15 @@ private Task StartAsync(CancellationToken cancellation, TimeSpan? delayStart = n
             return Task.Run(async () =>
             {
                 Exception uncaughtException = null;
-                this.logger.LogInformation(this.Name + " starting.");
                 try
                 {
                     if (delayStart != null)
+                    {
+                        this.logger.LogInformation($"{this.Name} starting in {delayStart.Value.TotalSeconds} seconds.");
```
noescape00: don't use string interpolation in logs

MithrilMan: fixed.
btw there are lot of logs actually using string interpolation, like this
https://github.com/stratisproject/StratisBitcoinFullNode/blob/a0004c768207fcaf624bdc88ddf40ca3316000f2/src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs#L1027

that should be fixed too, what's the best thing to do, change it when you put hands on a file containing wrong logs or fix with a PR whenever spot some?

P.S.
for the same reason, using previous `this.Name + "some test" should be discouraged too

noescape00: >btw there are lot of logs actually using string interpolation, like this

Yeah, this is incorrect

>change it when you put hands on a file containing wrong logs or fix with a PR whenever spot some?

Fix it whenever you spot it around the code you are working on. Or just create a separated PR for this and nothing else.

PR #2354, opened 2018-10-03
[SC] Assorted integration tests

codingupastorm:
- Calling to private method
- Calling with incorrect parameters
- Creating with value transfers
- Creating multiple contracts and checking nonces


src/Stratis.SmartContracts.IntegrationTests/ContractInternalTransferTests.cs
```diff
@@ -248,15 +248,104 @@ public void InternalTransfer_BetweenContracts()
...
+            Assert.True(receipt.Success);
+            Assert.True(receipt.GasUsed > GasPriceList.BaseCost);
+            Assert.Null(receipt.NewContractAddress);
+            Assert.Equal(this.node1.MinerAddress.Address, receipt.From);
+            Assert.Null(receipt.Error);
+            Assert.Equal(preResponse.NewContractAddress, receipt.To);
+        }
+
+        [Fact]
+        public void InternalTransfer_CreateMultipleContracts_FromConstructor()
```
rowandh: Cosider renaming to reflect that the test is to see if nonces are incrementing

Issue #2355, opened 2018-10-03
[SC] Enable MultiSig Scripts through the AllowedScriptTypesRule.


codingupastorm:
The implementation of AllowedScriptTypesRule is complete, but we need to test for MultiSig addresses, and ensure that it is flexible enough to be used for the sidechains implementation.

codingupastorm:
Get some empirical idea about how much time will be consumed, what we need the block time to be, and thus where to go next

dangershony:
Basically there is a Multisig Standard Template, I believe this should be enough to support for federation Multisig (up to 15 signers)

zeptin:
@dangershony Note my comments previously on #2356 as we currently use P2SH for the federation cross chain transfers. My understanding of the PoA feature is that it is transfer-agnostic and leaves the transfers up to other code (in this case the federation gateway feature).

PR #2356, opened 2018-10-03
[SC] Restrict ScriptPubKeys on network

### LINKED ISSUE #1835: [SC] Restrict ScriptPubKey formats allowed on SC network.
At the moment the only allowed transaction types should be smart contracts and P2PKH.

Other transaction types will limit our capacity to reach the 'sender' address, and the functionality should be provided by SC anyhow.

codingupastorm:
Closes #1835.

After this we need to do #2355

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/AllowedScriptTypesRule.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/AllowedScriptTypesRule.cs
@@ -1,0 +1,85 @@
+﻿using System.Threading.Tasks;
+using NBitcoin;
+using Stratis.Bitcoin.Consensus;
+using Stratis.Bitcoin.Consensus.Rules;
+using Stratis.Bitcoin.Features.MemoryPool;
+using Stratis.SmartContracts;
+using Stratis.SmartContracts.Core;
+using Block = NBitcoin.Block;
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Consensus.Rules
+{
+    /// <summary>
+    /// Enforces that only certain script types are used on the network.
+    /// </summary>
+    public class AllowedScriptTypesRule : PartialValidationConsensusRule, ISmartContractMempoolRule
+    {
+        protected ISmartContractCoinviewRule ContractCoinviewRule { get; private set; }
+
+        /// <inheritdoc />
+        public override void Initialize()
+        {
+            base.Initialize();
+            this.ContractCoinviewRule = (ISmartContractCoinviewRule)this.Parent;
+        }
+
+        public override Task RunAsync(RuleContext context)
+        {
+            Block block = context.ValidationContext.BlockToValidate;
+
+            foreach (Transaction transaction in block.Transactions)
+            {
+                CheckTransaction(transaction);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public void CheckTransaction(MempoolValidationContext context)
+        {
+            CheckTransaction(context.Transaction);
+        }
+
+        private void CheckTransaction(Transaction transaction)
+        {
+            foreach (TxOut output in transaction.Outputs)
+            {
+                CheckOutput(output);
+            }
+
+            // Coinbase inputs are funny-lookin so don't validate them
+            if (!transaction.IsCoinBase)
+            {
+                foreach (TxIn input in transaction.Inputs)
+                {
+                    CheckInput(input);
+                }
+            }
+        }
+
+        private void CheckOutput(TxOut output)
+        {
+            if (output.ScriptPubKey.IsSmartContractExec())
+                return;
+
+            if (output.ScriptPubKey.IsSmartContractInternalCall())
+                return;
+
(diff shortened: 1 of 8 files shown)


src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/AllowedScriptTypesRule.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (output.ScriptPubKey.IsSmartContractExec())
+                return;
+
+            if (output.ScriptPubKey.IsSmartContractInternalCall())
+                return;
+
+            if (PayToPubkeyHashTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
+                return;
+
+            if (PayToScriptHashTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
```
zeptin: I am not sure this is what you meant, shouldn't it be `PayToMultiSigTemplate` for non-P2SH (i.e. raw) multisig?

If you do it this way then arbitrary P2SH scriptPubKeys will be allowed through.

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/AllowedScriptTypesRule.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        private void CheckInput(TxIn input)
+        {
+            if (input.ScriptSig.IsSmartContractSpend())
+                return;
+
+            if (PayToPubkeyHashTemplate.Instance.CheckScriptSig(this.ContractCoinviewRule.Network, input.ScriptSig))
+                return;
+
+            if (PayToScriptHashTemplate.Instance.CheckScriptSig(this.ContractCoinviewRule.Network, input.ScriptSig, null)) // TODO: This last parameter is a ScriptPubKey. Which ScriptPubKey??
```
zeptin: I would expect (without looking too deeply into the underlying code, admittedly) that this is the P2SH redeemscript's corresponding scriptPubKey. It is not really possible to validate a scriptSig without the corresponding scriptPubKey it was originally created against.

In the case of a P2SH transaction the redeemscript would appear in the scriptSig, and the validation would verify that what was provided hashes/signs correctly.

zeptin: And my comment for `CheckOutput` applies here too, obviously.

src/Stratis.SmartContracts.IntegrationTests/SmartContractMemoryPoolTests.cs
```diff
@@ -99,6 +99,13 @@ public void SmartContracts_AddToMempool_OnlyValid()
                 tx.Sign(stratisNodeSync.FullNode.Network, stratisNodeSync.MinerSecret, false);
                 stratisNodeSync.Broadcast(tx);
 
+                // OP_RETURN (Anything non-P2PKH
```
zeptin: I would say that we should probably test multiple different varieties instead of just the one, but as a first implementation I suppose it is ok.

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/AllowedScriptTypesRule.cs
zeptin: Basically they are two different script structures:

1. Native multisig where `OP_CHECKMULTISIG` appears in a (long) scriptPubKey containing the pubkeys of all the multisig participants. The spending scriptSig presents the signatures of `m` of the `n` participants. This is space-inefficient for receiving funds as every 'incoming' transaction has all the keys in it. Spending takes a moderate amount of space.

2. P2SH multisig, where the scriptPubKey only contains the hash of the redeem script. This is considerably more space-efficient for receiving funds, and facilitates arbitrary redeem scripts. Spending therefore takes a variable amount of space depending on the complexity of the redeem script.

Basically we can use either for sidechains, it's just that we are currently using P2SH as it is usually the 'preferred' methodology (as the `3xxxxxx` addresses that result are easy to work with from a user perspective). How would you determine the sender for either variant, in any case? P2SH in particular might make it impossible to do so in a meaningful way.

codingupastorm: Oooh you raise a really important point - if we do allow a non-P2PKH, then we need some rules about when they can be used because we need the 'sender'

I believe the most reasonable is: If calling a smart contract, the first input to the transaction MUST be P2PKH.

codingupastorm: Is the multisig / p2sh implementation likely to change @zeptin ?

If so, it may be worth delaying that part of this rule until it's finalised

zeptin: Well, we would only change it if you need it to be changed, I think.

I need to go over how you call smart contracts in your implementation. A user could try to get the federation to send funds to a smart contract destination address, which seems like it would cause problems.

PR #2358, opened 2018-10-03
[Integration Tests] Clean up NotInIBD

fassadlr:
Move this method to the builder so that it checks after the node has started whether or not to set it's IBD state.

After this I'll move the WithWallet() method which will clean up the code even more.


bokobza:
@fassadlr I'm on it

bokobza:
Do you think we can have the `Corenode`s always start not in IBD by default? At the moment it's a lot of repetition, and I see very few cases where we don't care about IBD.

Let's do withWallet first and then see how we can simplify this.

fassadlr:
@bokobza yeah I agree, maybe we should have it NotInIbd by default and then we can just set it to IBD when needed.

Issue #2308, opened 2018-09-28
Create PoABlockHeader


noescape00:
Create PoAHeader that will inherit BlockHeader and will have additional field- signature. That way we won't need proven headers to make sure headers are not bogus.

Signature- it's a sig(blockHash).

Signature doesn't participate in block's hash calculation.

PR #2359, opened 2018-10-03
PoA BlockHeader

noescape00:
Fix: #2308


src/NBitcoin/BlockHeader.cs
```diff
@@ -56,7 +56,6 @@ public DateTimeOffset BlockTime
             }
         }
 
-        [Obsolete("Please use the Load method outside of consensus.")]
```
noescape00: Removed on purpose because it's not obsolete if we need to use it.

dangershony: hmmm.... it is, we should only use it from the ConsensusFactory, whis should actually be internal (but we need NBitcoin to be in Stratis.Bitcoin for that)

noescape00: Yes. Instead of being obsolete it should have a remark that it should be only used from consensus factory

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
```diff
@@ -0,0 +1,52 @@
+﻿using NBitcoin;
+using NBitcoin.Crypto;
+
+namespace Stratis.Bitcoin.Features.PoA
+{
+    public class PoABlockHeader : BlockHeader
+    {
+        private BlockSignature federationSignature;
```
dangershony: hmm I don't like the name why not just call it `blockSignature`

src/Stratis.Bitcoin.Tests/Utilities/AsyncLoopTest.cs
```diff
@@ -28,8 +28,6 @@ public async Task RunOperationCanceledExceptionThrownBeforeCancellationTokenIsCa
 
             await asyncLoop.Run(new CancellationTokenSource(800).Token, TimeSpan.FromMilliseconds(300)).RunningTask;
 
-            this.AssertLog(this.FullNodeLogger, LogLevel.Information, "TestLoop starting.");
```
dangershony: Why delete those?

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
bokobza: why not just have `public BlockSignature BlockSignature { get; set; }`?

src/Stratis.Bitcoin.Tests/Utilities/AsyncLoopTest.cs
noescape00: Those are failing on master. 
And it's second time we have tests failing because of string assertions. 
Imho it's a bad idea to verify that a particular log was added

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
noescape00: fixed

src/NBitcoin/BlockHeader.cs
dangershony: I prefer you put it back so to avoid mistakes by other devs

noescape00: Makes sense. Resoted

noescape00:
@dangershony fixes applied

src/Stratis.Bitcoin.Tests/Utilities/AsyncLoopTest.cs
bokobza: I guess it's failing because @MithrilMan moved the `... starting` logs to be logged when the node is actually starting, which is good.
Agree that we shouldn't test logs.

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public PoAHeadersPayload(params PoABlockHeader[] headers)
+        {
+            this.Headers.AddRange(headers);
+        }
+
+        [NoTrace]
+        public override void ReadWriteCore(BitcoinStream stream)
+        {
+            if (stream.Serializing)
```
bokobza: why do we need to do all this?
The readWrite method does it all by itself I think.
https://github.com/stratisproject/StratisBitcoinFullNode/blob/d3a1952a85a47d3a041c137f970162dd5a3e5da9/src/NBitcoin/BitcoinStream.cs#L215-L248

Looks like we can just have:
```
[NoTrace]
public override void ReadWriteCore(BitcoinStream stream)
{
    stream.ReadWrite(ref this.headers);
}
```

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Stratis.Bitcoin.Features.PoA
+{
+    public class PoABlockHeader : BlockHeader
+    {
+        private BlockSignature blockSignature;
+
+        public BlockSignature BlockSignature
+        {
+            get => this.blockSignature;
```
bokobza: Can we use `public BlockSignature BlockSignature { get; set; }`?

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
```diff
@@ -0,0 +1,52 @@
...
+            set => this.blockSignature = value;
+        }
+
+        public override void ReadWrite(BitcoinStream stream)
+        {
+            base.ReadWrite(stream);
+            stream.ReadWrite(ref this.blockSignature);
```
bokobza: I guess we could have a comment at the top of this class to explain why we add the block signature here and why we exclude it from the hash calculation.

src/Stratis.Bitcoin.Features.PoA/PoABlockHeaderValidator.cs
```diff
@@ -0,0 +1,50 @@
...
+        public void Sign(Key key, PoABlockHeader header)
+        {
+            ECDSASignature signature = key.Sign(header.GetHash());
+
+            header.BlockSignature = new BlockSignature { Signature = signature.ToDER() };
+        }
+
+        public bool VerifySignature(PubKey pubKey, PoABlockHeader header)
+        {
+            if (header.BlockSignature.IsEmpty())
```
bokobza: maybe check for null as well ` if (header.BlockSignature == null || header.BlockSignature.IsEmpty())`

src/Stratis.Bitcoin.Features.PoA/PoABlockHeaderValidator.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public void Sign(Key key, PoABlockHeader header)
+        {
+            ECDSASignature signature = key.Sign(header.GetHash());
+
+            header.BlockSignature = new BlockSignature { Signature = signature.ToDER() };
+        }
+
+        public bool VerifySignature(PubKey pubKey, PoABlockHeader header)
```
bokobza: Nice implementation.

src/Stratis.Bitcoin.Features.PoA/PoABlockHeaderValidator.cs
```diff
@@ -0,0 +1,50 @@
...
+        private readonly ILogger logger;
+
+        public PoABlockHeaderValidator(ILoggerFactory factory)
+        {
+            this.logger = factory.CreateLogger(this.GetType().FullName);
+        }
+
+        public void Sign(Key key, PoABlockHeader header)
+        {
+            ECDSASignature signature = key.Sign(header.GetHash());
```
bokobza: Do we need to make sure the signature is canonical before creating the BlockSignature?
```
ECDSASignature signature = key.Sign(header.GetHash());
signature = signatuee.MakeCanonical();
```

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public PoAHeadersPayload()
+        {
+            this.headers = new List<PoABlockHeader>();
+        }
+
+        public PoAHeadersPayload(params PoABlockHeader[] headers)
+        {
+            this.Headers.AddRange(headers);
```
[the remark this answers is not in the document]
noescape00: It can't be because we initialize it in constructor

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+using Stratis.Bitcoin.P2P.Protocol.Payloads;
+using TracerAttributes;
+
+namespace Stratis.Bitcoin.Features.PoA.Payloads
+{
+    /// <summary>
+    /// Block headers received after a getheaders messages.
```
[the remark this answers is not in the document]
noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    [Payload("poahdr")]
+    public class PoAHeadersPayload : Payload
+    {
+        private List<PoABlockHeader> headers;
```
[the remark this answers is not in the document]
noescape00: we can't use ref with fields. And we need ref for stream.ReadWrite

src/Stratis.Bitcoin.Tests/Utilities/AsyncLoopTest.cs
MithrilMan: Ops my bad I didn't ran tests when I changed the log output, I didn't expected it was tested

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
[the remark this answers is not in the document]
bokobza: but this is also a constructor

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
noescape00: you are right. fixed

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
noescape00: no, for the same reason as above- need this `ref` for btc stream

src/Stratis.Bitcoin.Features.PoA/PoABlockHeader.cs
noescape00: fixed. 

didn't add info why we exclude it from hash calculation- well, because there is no reason to do so.

src/Stratis.Bitcoin.Features.PoA/PoABlockHeaderValidator.cs
noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/PoABlockHeaderValidator.cs
noescape00: It's always canonical be default.

noescape00:
@bokobza fixes applied, please check

src/Stratis.Bitcoin.Tests/Utilities/AsyncLoopTest.cs
noescape00: No problem mate. I also merged it too quickly, my bad.

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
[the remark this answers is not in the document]
noescape00: yes, fixed 🙂

src/Stratis.Bitcoin.Features.PoA/Payloads/PoAHeadersPayload.cs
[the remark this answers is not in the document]
bokobza: with *properties.  ah yeah, true, completely forgot about that.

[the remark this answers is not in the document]
noescape00: ah, yeah. with properties 🙂

PR #2361, opened 2018-10-03
Fixed issue #2294.

### LINKED ISSUE #2294: FullNode BTC network memory leak [bug] [sp]
@dangershony reported an issue with Bitcoin network synching using fullnode that smells of memory leak

here the screenshot

image

Dan guess it's a background task not disposing, I'll investigate the issue reporting here the findings

MithrilMan:
Fixes #2294 

Prevent enqueue too many BlockDownload that may cause the process to Fill up all the available computer RAM.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -1100,7 +1100,15 @@ namespace Stratis.Bitcoin.Consensus
                     return;
                 }
 
+                // To fix issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/2294#issue-364513736
+                // if there are no samples, assume the worst scenario (you are going to donwload full blocks).
                 long avgSize = (long)this.blockPuller.GetAverageBlockSizeBytes();
+                if (avgSize == 0)
+                {
+                    // TODO: move MaxBlockSize on a common place.
+                    avgSize = Block.MaxBlockSize;
+                }
+
                 int maxBlocksToAsk = avgSize != 0 ? (int)(freeBytes / avgSize) : DefaultNumberOfBlocksToAsk;
 
                 if (maxBlocksToAsk > freeSlots)
--- a/src/Stratis.Bitcoin/Utilities/AverageCalculator.cs
+++ b/src/Stratis.Bitcoin/Utilities/AverageCalculator.cs
@@ -17,14 +17,6 @@ namespace Stratis.Bitcoin.Utilities
         /// <summary>Samples used in calculation of the average value.</summary>
         private CircularArray<double> samples;
 
-        /// <summary>
-        /// Gets a value indicating whether this instance is empty.
-        /// </summary>
-        /// <value>
-        ///   <c>true</c> if this instance is empty; otherwise, <c>false</c>.
-        /// </value>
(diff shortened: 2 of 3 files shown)


src/Stratis.Bitcoin/Utilities/AverageCalculator.cs
```diff
@@ -17,6 +17,14 @@ public class AverageCalculator
         /// <summary>Samples used in calculation of the average value.</summary>
         private CircularArray<double> samples;
 
+        /// <summary>
+        /// Gets a value indicating whether this instance is empty.
+        /// </summary>
+        /// <value>
+        ///   <c>true</c> if this instance is empty; otherwise, <c>false</c>.
+        /// </value>
+        public bool IsEmpty { get { return this.samples.Count == 0; } }
```
noescape00: No need for that- blocks can't be smaller than header size so in GetAverageBlockSizeBytes you can check average to be less or equal to 80. 
Also this change breaks 2 BP tests

MithrilMan: moving the bugfix logic into ProcessDownloadQueueLocked method

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1100,7 +1100,15 @@ private void ProcessDownloadQueueLocked()
                     return;
                 }
 
+                // To fix issue https://github.com/stratisproject/StratisBitcoinFullNode/issues/2294#issue-364513736
+                // if there are no samples, assume the worst scenario (you are going to donwload full blocks).
                 long avgSize = (long)this.blockPuller.GetAverageBlockSizeBytes();
+                if (avgSize == 0)
+                {
+                    // TODO: move MaxBlockSize on a common place.
+                    avgSize = Block.MaxBlockSize;
```
dangershony: No this is wrong it should come from `Netowrk.Consensus`

noescape00: Yes. We have this value copypasted all over. Should just refactor the code to make sure it's in one place. 

Probably next thing @MithrilMan can do is to find all block size values and making sure we use only one from Network.Consensus

noescape00: That's why there is a TODO. 
I think @MithrilMan already discussed that in fn channel

MithrilMan: yes, sparse delcarations all around
I can try to replace all Block.MaxBlockSize to `network.Consensus.Options.MaxBlockBaseSize` if that's the current property (what does `Base` refers to?)

dangershony: yes @MithrilMan please replace to `network.Consensus.Options.MaxBlockBaseSize` and get rid of all other places with a block size.

MithrilMan: actually there are 16 references, would be quick to just update last one and then looking to change them all (not just Block. but even other const around, like Transaction.MAX_BLOCK_SIZE)

MithrilMan: #2375

PR #2362, opened 2018-10-04
[SC] Fix #1698 GasInjector and MemoryLimit rewriters create invalid programs

### LINKED ISSUE #1698: SmartContractGasInjector creates an invalid program in methods with nested loops.
When there are nested loops of non-constant iterations, the GasInjector injects the SpendGas code in such a way that the resulting CIL is invalid. When we try to execute the method via reflection we get the following exception: `System.InvalidProgramException: Common Language Runtime detected an invalid program.`

Playing with it myself, I was able to find the following code that will produce this bug:
```
using Stratis.SmartContracts;

public sealed class ThrowExceptionContract : SmartContract
{
    public ThrowExceptionContract(ISmartContractState state)
        : base(state)
    {
    }

    string SampleString { get; set;}

    public void ThrowException()
    {
        SampleString = "test";
        for(int i = 0; i < SampleString.Length; i++)
        {
            SampleString += "1";
            for(int j=0; j < SampleString.Length; j++)
            {
                SampleString = "example";
                SampleString += "2";
            }
        }
    }
}
```

Yes, the code is nonsensical  but it should still execute (possibly just until it runs out of gas).

We should have a test for code similar to the above (ideally a more realistic use case) in SmartContractExecutorTests, and fix the issue.

Current Bounty: 50 STRAT.

nathanielcook:
As mentioned [here](https://stackoverflow.com/a/7267875/89590), Cecil updates branch locations on our behalf, but we need some extra commands to convert "br.s" op codes to "br"; otherwise the IL command can point to an invalid location.


rowandh:
Thanks @nathanielcook! Will take a look soon.

codingupastorm:
Confirmed both new tests failed before the change and pass after.

codingupastorm:

src/Stratis.SmartContracts.Executor.Reflection/ILRewrite/GasInjectorRewriter.cs
```diff
@@ -133,14 +134,6 @@ public void Rewrite(MethodDefinition methodDefinition, ILProcessor il, ObserverR
 
                 AddSpendGasMethodBeforeInstruction(methodDefinition, context.Observer, context.ObserverVariable, injectAfterInstruction, gasToSpendForSegment[instruction]);
             }
-
-            foreach (Instruction instruction in branches)
```
codingupastorm: Yessss very glad to get rid of this :)

nathanielcook:
Hi Jordan. Yes, almost. The offset is signed (as in, it can be negative), which limits it to 127. By the way, since it is an offset 127 means the target instruction can be no more than 127 bytes from the beginning of the instruction following the current instruction. (That might have been what you meant when you said location, but just to clarify.)

nathanielcook:
By the way, remember I don't have merge permissions, so feel free to merge whenever you guys see fit. 🙂

PR #2363, opened 2018-10-04
[SC] Deploy Attribute validator

### LINKED ISSUE #1752: Smart contract compilation should fail for assemblies without only 1 contract or without a Deploy attribute.

codingupastorm:
Closes #1752


codingupastorm:
All integration tests passing.

PR #2375, opened 2018-10-04
Changed MaxBlockSize property

MithrilMan:
using this.network.Consensus.Options.MaxBlockBaseSize instead of Block.MaxBlockSize


noescape00:
@MithrilMan please fix other places where we have max block size constant copypasted. Everywhere use `network.Consensus.Options.MaxBlockBaseSize;`

noescape00:
Okay, change other places in next PR 🙂

Issue #2310, opened 2018-09-28
Make all header-related components work with PoABlockHeader


noescape00:
That mostly touches CMB and ChainRepository.

Make sure we are saving those headers to the database and relay them.

PR #2377, opened 2018-10-04
Dummy miner that will be needed to check syncing and payloads

noescape00:
This PR is needed for testing https://github.com/stratisproject/StratisBitcoinFullNode/issues/2310

Later we will fix the miner to produce blocks according to poa consensus rules.

untitled

### DIFF
--- a/src/Stratis.Bitcoin.Features.PoA/ConsensusRules/PoACoinviewRule.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/ConsensusRules/PoACoinviewRule.cs
@@ -1,19 +1,48 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using System.Threading.Tasks;
+﻿using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Consensus;
 using Stratis.Bitcoin.Consensus.Rules;
+using Stratis.Bitcoin.Features.Consensus.Rules.CommonRules;
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.PoA.ConsensusRules
 {
-    public class PoACoinviewRule : FullValidationConsensusRule
+    public class PoACoinviewRule : CoinViewRule
     {
-        /// <inheritdoc />
-        public override Task RunAsync(RuleContext context)
-        {
-            // TODO POA implement rule
+        // TODO POA check this rule properly later, it's a dummy implementation needed for miner
 
-            return Task.CompletedTask;
+        /// <inheritdoc/>
+        protected override bool IsProtocolTransaction(Transaction transaction)
+        {
+            return transaction.IsCoinBase;
+        }
+
+        /// <inheritdoc/>
+        public override void CheckBlockReward(RuleContext context, Money fees, int height, Block block)
+        {
+            if (block.Transactions[0].TotalOut > fees)
+            {
+                this.Logger.LogTrace("(-)[BAD_COINBASE_AMOUNT]");
+                ConsensusErrors.BadCoinbaseAmount.Throw();
+            }
+        }
+
+        /// <inheritdoc/>
+        public override Money GetProofOfWorkReward(int height)
+        {
+            return 0;
+        }
+
+        /// <inheritdoc/>
+        public override void CheckMaturity(UnspentOutputs coins, int spendHeight)
+        {
+            base.CheckCoinbaseMaturity(coins, spendHeight);
+        }
+
+        /// <inheritdoc/>
+        public override void UpdateCoinView(RuleContext context, Transaction transaction)
+        {
+            base.UpdateUTXOSet(context, transaction);
         }
     }
 }
--- a/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
@@ -20,11 +20,11 @@ namespace Stratis.Bitcoin.Features.PoA
 
         private readonly PoANetwork network;
 
-        private readonly PoAMiner miner;
+        private readonly IPoAMiner miner;
 
         private readonly ILogger logger;
 
-        public FederationManager(NodeSettings nodeSettings, Network network, PoAMiner miner, ILoggerFactory loggerFactory)
+        public FederationManager(NodeSettings nodeSettings, Network network, IPoAMiner miner, ILoggerFactory loggerFactory)
         {
             this.settings = Guard.NotNull(nodeSettings, nameof(nodeSettings));
             this.network = Guard.NotNull(network as PoANetwork, nameof(network));
--- a/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
@@ -51,7 +51,7 @@ namespace Stratis.Bitcoin.Features.PoA
                 this.logger.LogInformation("Federation key pair was successfully loaded. Your public key is: {0}.", this.FederationMemberKey.PubKey);
 
                 // Enable mining because we are a federation member.
-                this.miner.StartMining();
+                this.miner.InitializeMining();
             }
         }
 
--- a/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
@@ -4,6 +4,7 @@ using System.Text;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.PoA
 {
@@ -25,9 +26,9 @@ namespace Stratis.Bitcoin.Features.PoA
 
         public FederationManager(NodeSettings nodeSettings, Network network, PoAMiner miner, ILoggerFactory loggerFactory)
         {
-            this.settings = nodeSettings;
-            this.network = network as PoANetwork;
-            this.miner = miner;
+            this.settings = Guard.NotNull(nodeSettings, nameof(nodeSettings));
+            this.network = Guard.NotNull(network as PoANetwork, nameof(network));
+            this.miner = Guard.NotNull(miner, nameof(miner));
 
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
         }
--- a/src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
@@ -1,0 +1,56 @@
+﻿using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Consensus;
+using Stratis.Bitcoin.Features.MemoryPool;
+using Stratis.Bitcoin.Features.MemoryPool.Interfaces;
+using Stratis.Bitcoin.Features.Miner;
+using Stratis.Bitcoin.Mining;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.PoA
+{
+    public class PoABlockDefinition : BlockDefinition
+    {
+        /// <summary>Instance logger.</summary>
+        private readonly ILogger logger;
+
+        public PoABlockDefinition(
+            IConsensusManager consensusManager,
+            IDateTimeProvider dateTimeProvider,
+            ILoggerFactory loggerFactory,
+            ITxMempool mempool,
+            MempoolSchedulerLock mempoolLock,
+            Network network)
+            : base(consensusManager, dateTimeProvider, loggerFactory, mempool, mempoolLock, new MinerSettings(), network)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+        }
+
+        /// <inheritdoc/>
+        public override void AddToBlock(TxMempoolEntry mempoolEntry)
+        {
+            this.AddTransactionToBlock(mempoolEntry.Transaction);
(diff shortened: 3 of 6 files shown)


src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            return this.BlockTemplate;
+        }
+
+        /// <inheritdoc/>
+        public override void UpdateHeaders()
+        {
+            base.UpdateBaseHeaders();
+
+            this.block.Header.Bits = Target.Difficulty1; // TODO POA use some constant difficulty all the time so chainwork rises
```
bokobza: Did you mean "Should use some constant difficulty so the chainwork **doesn't** rise"?
Good idea on the rule.

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
```diff
@@ -62,6 +63,7 @@ public static IFullNodeBuilder UsePoAConsensus(this IFullNodeBuilder fullNodeBui
                         services.AddSingleton<FederationManager>();
                         services.AddSingleton<PoABlockHeaderValidator>();
                         services.AddSingleton<PoAMiner>();
+                        services.AddSingleton<PoABlockDefinition, PoABlockDefinition>();
```
bokobza: Should be `services.AddSingleton<BlockDefinition, PoABlockDefinition>();`

src/Stratis.Bitcoin.Features.PoA/ConsensusRules/PoACoinviewRule.cs
```diff
@@ -1,19 +1,48 @@
...
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.PoA.ConsensusRules
 {
-    public class PoACoinviewRule : FullValidationConsensusRule
+    public class PoACoinviewRule : CoinViewRule
     {
-        /// <inheritdoc />
-        public override Task RunAsync(RuleContext context)
+        // TODO POA check this rule properly later, it's a dummy implementation needed for miner
```
bokobza: Ok so I won't review it.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,121 @@
...
+        private readonly ILogger logger;
+
+        private readonly Network network;
+
+        /// <summary>
+        /// A cancellation token source that can cancel the mining processes and is linked to the <see cref="INodeLifetime.ApplicationStopping"/>.
+        /// </summary>
+        private CancellationTokenSource cancellation;
+
+        private readonly IInitialBlockDownloadState indState;
```
bokobza: ibdState*

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,121 @@
...
+using NBitcoin;
+using Stratis.Bitcoin.Consensus;
+using Stratis.Bitcoin.Interfaces;
+using Stratis.Bitcoin.Mining;
+using Stratis.Bitcoin.Utilities;
 
 namespace Stratis.Bitcoin.Features.PoA
 {
-    public class PoAMiner
+    public class PoAMiner : IDisposable
```
bokobza: Can we have an interface so that we have `PoaMining : IPoaMining, Idisposable` (like in IPosMinting and IPowMining)?

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,121 @@
...
+            }
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
         {
+            this.cancellation.Cancel();
+            this.miningTask?.GetAwaiter().GetResult();
 
+            this.cancellation.Dispose();
```
bokobza: since we're doing a new Dispose method, can we maybe do dispose thew proper way?
Look here: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose
and for an example, look how it's done in the DNS project.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,121 @@
...
 namespace Stratis.Bitcoin.Features.PoA
 {
-    public class PoAMiner
+    public class PoAMiner : IDisposable
     {
-        // TODO POA. First add creation of block template and later apply signatures and other stuff. We need bare minimum to test syncing and payloads
+        // TODO POA. Implement miner properly later. Right now we need bare minimum to test syncing and payloads.
 
-        public void StartMining()
+        private const bool MineDuringIBD = true; // TODO POA just for tests
```
bokobza: this is bound to be very confusing. Maybe we should never mine during IBD and for the test, we make sure we're not in IBD.

src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
noescape00: Chainwork is a cumulative of difficulties of all blocks. So the chainwork HAVE to rise with every block. And it should rise by equal amount. 

For example we might have shorter chains on stratis that are better in terms of chainwork. For PoA logest chain will always mean the best chain.

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
noescape00: BlockDefinition has an interface I don't want to use really. It makes me provide a script. But script is not needed without a subsidy for a block.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Yeah. Later when we work on production version of miner we shouldn't forget about adding the interface and commenting it properly.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: It's for testing (but not in unit tests- testing during the development). Without it every time I want to run the miner I'll need to change network so genesis block have timestamp that is close to my machine's timestamp, otherwise it wont mine because we will be in ibd.

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
```diff
@@ -25,9 +26,9 @@ public class FederationManager
 
         public FederationManager(NodeSettings nodeSettings, Network network, PoAMiner miner, ILoggerFactory loggerFactory)
         {
-            this.settings = nodeSettings;
-            this.network = network as PoANetwork;
-            this.miner = miner;
+            this.settings = Guard.NotNull(nodeSettings, nameof(nodeSettings));
```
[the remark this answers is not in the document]
noescape00: We used this overload in FN.cs or basefeature.cs. And for some reason everywhere except for those two we decided to just use more lines than we need.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: will fix

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: But it's not override, would would we need overidable dispose(bool)?

src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
bokobza: makes sense :-)

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
bokobza: but you're inheriting from BlockDefinition already.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: I really think adding the interface now is a good move. When you want to test etc, you have to have an interface. If it changes, than it's very quick and easy to change.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: ok

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
dangershony: I don't like the name `FederationManager` I dont think we have similar equivalents with POW/POS (like a PowManager)
So you could consider just calling this PoaMiner or something similar (future PR is ok).

src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
```diff
@@ -0,0 +1,58 @@
...
+            // TODO maybe make it a consensus rule to ensure that bits are always constant (!!!!!!)
+            // If we do that we can reuse selecting chain using best chainwork
```
dangershony: Maybe put the constant difficulty in the Consensus Options (yeah to make a rule sounds definitely correct.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
dangershony: The fix is probably not to create such a flag `MineDuringIBD` but rather in the calculation of difficulty you check the testnet flag on the network class and allow blocks when in testnet.

Also we solved the time issue in POS by injecting a fake datetime provider that will simulate the date for what the tests needs to really no need for this flag.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,122 @@
...
+            PoABlockDefinition blockDefinition)
+        {
+            this.consensusManager = consensusManager;
+            this.dateTimeProvider = dateTimeProvider;
+            this.network = network;
+            this.ibdState = ibdState;
+            this.blockDefinition = blockDefinition;
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.cancellation = CancellationTokenSource.CreateLinkedTokenSource(new[] { nodeLifetime.ApplicationStopping });
```
dangershony: Why create a linked cancellation, just wondering.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -1,16 +1,122 @@
...
+        }
+
+        /// <summary>Starts mining loop.</summary>
+        public void InitializeMining()
+        {
+            if (this.miningTask == null)
```
dangershony: If this is called from apis you will have race conditions here, what is the plan on mining is it enabled by default for a block producer? (in that case this is ok)
Also consider using asyncloop (see POS)

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: It's the pattern that's like that. this is so that you can call Dispose many times and the freeing of resources is properly managed + it doesn't throw exceptions. 
Look at this bit: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#dispose-and-disposeboolean

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
noescape00: Yeah, but when I inherit I set default value to the script to be null. And in `BlockDefinition` there is no default value so caller will have to guess if putting null there is ok, when in fact that's the only thing that caller should do

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: ok, fixed

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Currently dispose method already supports calling it multiple times without a problem. 
What is the benefit of having 2 dispose (one with bool and one without) in this case?

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: it's not for unit tests, it's for running the node, so injection is just a more difficult approach.
this bool will be removed as soon as the network can run stable.

src/Stratis.Bitcoin.Features.PoA/PoABlockDefinition.cs
noescape00: In the next PR adding it to rule

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: copied from PoW miner. It's for having an ability to stop miner manually or via lifetime

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: It's not called from API or RPC and it won't in the future.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
dangershony: OK but integration tests should be created with injection of override tool and not this flag (if the flag is intended to be deleted that fine with me)

PR #2380, opened 2018-10-05
[SC] Make TransferInfo immutable

rowandh:


src/Stratis.SmartContracts.Core/State/AccountAbstractionLayer/TransferInfo.cs
```diff
@@ -9,6 +9,13 @@ namespace Stratis.SmartContracts.Core.State.AccountAbstractionLayer
...
     public class TransferInfo
     {
+        public TransferInfo(uint160 from, uint160 to, ulong value)
+        {
+            this.From = from;
+            this.To = to;
+            this.Value = value;
+        }
+
         public uint160 From { get; set; }
```
codingupastorm: Hmm, immutable? Should the set; be removed?

rowandh: :joy: probably, that was the whole point of this PR

PR #2382, opened 2018-10-05
Add sendtoaddress, walletpassphrase, walletlock RPC methods

### LINKED ISSUE #2266: Implement sendtoaddress RPC method
Bitcoin Core Documentation:
https://bitcoin.org/en/developer-reference#sendtoaddress

mikedennis:
Fixes #2266 

Implements wallet locking/unlocking by using the key cache already present in `WalletTransactionHandler`. Also implemented **sendtoaddress** rpc method which was just stubbed out in the current code base.

Also added some integration tests for this functionality:

`TestRpcSendToAddressCantSpendWhenLocked()` does take a while to run as it tests the expiry of the unlocking which is only checked once per minute. 

@fassadlr Also removed an existing **sendtoaddress** wallet test that my new test `TestRpcSendToAddressIsSuccessful()` replaces. The original test did not properly handle coinbase requirement before spending so I just deleted it as the new one handles that case and also checks a couple of additional things.


src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -25,20 +29,81 @@ public class WalletRPCController : FeatureController
...
+        [ActionName("sendtoaddress")]
+        [ActionDescription("Sends money to a bitcoin address. Requires wallet to be unlocked using walletpassphrase.")]
+        public async Task<uint256> SendToAddressAsync(BitcoinAddress address, decimal amount, string commentTx, string commentDest)
+        {
+            WalletAccountReference account = this.GetAccount(); 
+            TransactionBuildContext context = new TransactionBuildContext(this.fullNode.Network)
+            {
+                AccountReference = this.GetAccount(),
+                Recipients = new [] {new Recipient { Amount = Money.Coins(amount), ScriptPubKey = address.ScriptPubKey } }.ToList(),
+                WalletPassword = "_" // Want private key to pull from cache, so pass in non empty string so tx will be signed.
```
zeptin: This feels a bit ugly

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -29,5 +32,67 @@ public void TestRpcGetBlockWithValidHashIsSuccessful()
...
+                node.NotInIBD().WithWallet();
+                int blocksToMine = (int)network.Consensus.CoinbaseMaturity + 1;
+                TestHelper.MineBlocks(node, blocksToMine);
+                TestHelper.WaitLoop(() => node.FullNode.GetBlockStoreTip().Height == blocksToMine);
+
+                var alice = new Key().GetBitcoinSecret(network);
+                var aliceAddress = alice.GetAddress();
+                rpcClient.WalletPassphrase("password", 60);
+                var txid = rpcClient.SendToAddress(aliceAddress, Money.Coins(1.0m));
+                rpcClient.SendCommand(RPCOperations.walletlock); // TODO: create a method for walletlock.
```
zeptin: This TODO can be removed I think.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -29,5 +32,67 @@ public void TestRpcGetBlockWithValidHashIsSuccessful()
...
+
+                var alice = new Key().GetBitcoinSecret(network);
+                var aliceAddress = alice.GetAddress();
+
+                // Not unlockedcase.                
+                Assert.Throws<RPCException>(() => rpcClient.SendToAddress(aliceAddress, Money.Coins(1.0m)));
+
+                // Unlock and lock case.
+                rpcClient.WalletPassphrase("password", 60);
+                rpcClient.SendCommand(RPCOperations.walletlock); // TODO: create a method for walletlock.
```
zeptin: Remove TODO.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -29,5 +32,67 @@ public void TestRpcGetBlockWithValidHashIsSuccessful()
...
+                Assert.Throws<RPCException>(() => rpcClient.SendToAddress(aliceAddress, Money.Coins(1.0m)));
+
+                // Unlock timesout case.
+                rpcClient.WalletPassphrase("password", 5);
+                Thread.Sleep(120 * 1000); // 2 minutes.
```
zeptin: Maybe use a much shorter duration than this? It will just slow the tests down, we expect the wallet to be locked again within 5 seconds, yes?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -25,20 +29,81 @@ public class WalletRPCController : FeatureController
...
-        public uint256 SendToAddress(BitcoinAddress bitcoinAddress, Money amount)
+        [ActionName("walletpassphrase")]
+        [ActionDescription("Stores the wallet decryption key in memory for the indicated number of seconds. Issuing the walletpassphrase command while the wallet is already unlocked will set a new unlock time that overrides the old one.")]
+        [NoTrace]
+        public bool UnlockWallet(string passphrase, int timeout)
+        {
+            Guard.NotEmpty(passphrase, nameof(passphrase));
+
+            // As per RPC method definition this should be the max allowable expiry duration.
+            const int maxDurationInSeconds = 1073741824;
```
dangershony: const at top of class

dangershony: max duration should probably go in the trx handler not in the controller

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
dangershony: yeah it is ugly just use an empty string

dangershony:
Good job, looks good to me, please address  some of @zeptin concerns and lets also let @bokobza review this.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
zeptin: Ah, I see you mention this in the PR description.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
zeptin: Well, the empty-ness is used inside `BuildTransaction` to check whether signing should occur. But there should be a less hacky solution than this. Maybe a future PR could require an explicit bool in `TransactionBuildContext` for signing.

mikedennis: Yes zeptin is correct, I would have to either add a parameter to the context, add a different method or add a parameter to the method. Not sure which is best so will wait to see what Jeremy thinks about this code.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: This max duration comes from bitcoin core. I'm not sure how this magic number was derived. Do we even want to restrict this?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
dangershony: Ok I see the argument, the correct approach is a flag that says to look in cache or use password.
But for now I guess this is ok.

dangershony: Perhaps this code should go in the else if key is not in cache.
```
            if (string.IsNullOrEmpty(context.WalletPassword))
                return;
```

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -27,20 +33,78 @@ public class WalletRPCController : FeatureController
...
+        public bool LockWallet()
         {
             WalletAccountReference account = this.GetAccount();
-            return uint256.Zero;
+            this.walletTransactionHandler.ClearCachedSecret(account);
+            return true; // NOTE: Have to return a value or else RPC middleware doesn't serialize properly.
+        }
+
+        [ActionName("sendtoaddress")]
+        [ActionDescription("Sends money to a bitcoin address. Requires wallet to be unlocked using walletpassphrase.")]
```
bokobza: Can we just say "Sends money to an address."?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -27,20 +33,78 @@ public class WalletRPCController : FeatureController
...
+
+            try
+            {
+                Transaction transaction = this.walletTransactionHandler.BuildTransaction(context);
+                await this.broadcasterManager.BroadcastTransactionAsync(transaction);
+
+                uint256 hash = transaction.GetHash();
+                return hash;
+            }
+            catch (SecurityException exception)
```
bokobza: Did you plan to have separate RPC server exception here?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
bokobza: This is a hard one. I'm happy for this to go in like this, but @mikedennis can you please follow with a PR that adds a `bool Sign` to the TransactionBuildContext? True by default.
If sign = true and there is no password then we rely on the cache, if sign is false and there is password we ignore it etc.

mikedennis: Yes for sure. Sounds like a good solution.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: Good idea. It's updated now with a better error code.

PR #2384, opened 2018-10-05
[SC] Change method param serialization scheme for transactions

rowandh:
Use an RLP-encoded list of byte arrays as our method parameter transaction serialization wire format.

Part of #1830.


src/Stratis.SmartContracts.Executor.Reflection/Serialization/ContractPrimitiveSerializer.cs
```diff
@@ -59,6 +59,9 @@ public byte[] Serialize(object o)
 
             if (o is string s)
                 return Serialize(s);
+
+            if (o is uint160 u160)
```
codingupastorm: Why is this type included?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/MethodParameterByteSerializerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        [Fact]
+        public void Roundtrip_Serialize_Multiple_Params()
+        {
+            object[] methodParameters = GetData(0).SelectMany(o => o).ToArray();
+
+            var serialized = this.Serializer.Serialize(methodParameters);
+
+            var deserialized = this.Serializer.Deserialize(serialized);
```
codingupastorm: Assert length of deserialized == methodParameters or similar here, just in case it returns an empty array, test would still pass

src/Stratis.Bitcoin.Features.SmartContracts.Tests/MethodParameterByteSerializerTests.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (paramType.IsArray && paramType.GetElementType() == typeof(byte))
+            {
+                Assert.True(((byte[])value).SequenceEqual((byte[])methodParamObjects[0]));
+            }
+        }
+
+        [Fact]
+        public void Serialized_Method_Params_Are_Smaller_Than_Strings()
+        {
+            // Single comparative case for a sample byte vs. string encoded method params array
```
codingupastorm: 💯

src/Stratis.SmartContracts.Executor.Reflection/Serialization/MethodParameterByteSerializer.cs
```diff
@@ -0,0 +1,87 @@
...
+using Nethereum.RLP;
+
+namespace Stratis.SmartContracts.Executor.Reflection.Serialization
+{
+    /// <summary>
+    /// Serializes method parameters using RLP-encoded byte arrays.
+    /// </summary>
+    public class MethodParameterByteSerializer : IMethodParameterSerializer
+    {
+        private readonly IContractPrimitiveSerializer primitiveSerializer;
```
codingupastorm: Because this uses the primitive serializer, the value will be determined by our serializer behaviour, which I believe will be returning the default value in case of null. Could be misleading? Lets discuss.

src/Stratis.SmartContracts.Executor.Reflection/Serialization/Prefix.cs
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        public static Prefix ForObject(object o)
+        {
+            var type = (byte) GetPrimitiveType(o);
+            return new Prefix(type);
+        }
+
+        private static MethodParameterDataType GetPrimitiveType(object o)
```
codingupastorm: If the primitives are the same among a few serializers and we end up with many methods like this, I think there is a better approach here. A static readonly list of the primitive types somewhere with references to the required method paramters and serialization etc? Would help us in future with ensuring each type is covered for each serializer rather than copying these long if returns

Not a blocker, just thoughts!

rowandh:
Closing until we resolve some other serialization decisions.

Issue #2391, opened 2018-10-05
Care about Stopwatch ElapsedTicks property


MithrilMan:
Worning on an Issue about DNS feature, I spotted a point where Stopwatch was used old fashioned way to measure dns performance instead using our StopwatchDisposable class, so I refactored the code.
In the process of refactoring, I saw that we were getting the elapsed Ticks in two different ways:
- using stopwatch.ElapsedTicks property (a long value=
- using stopwatch.Elapsed property (a Timespan value) accessing then it's Ticks property

While at first I thought it was the same, I googled to confirm that.

What I discovered was this: https://stackoverflow.com/a/1438984/1716620

now looking at dotnet source, one isn't just a wrapper of another

https://github.com/dotnet/corefx/blob/f267aed8b4a8f8314b2f6773cd1d825a82b0075f/src/System.Runtime.Extensions/src/System/Diagnostics/Stopwatch.cs#L115-L128

that lead to this:
https://github.com/dotnet/corefx/blob/f267aed8b4a8f8314b2f6773cd1d825a82b0075f/src/System.Runtime.Extensions/src/System/Diagnostics/Stopwatch.cs#L144-L175
```
// Get the elapsed ticks.        
private long GetRawElapsedTicks()
{
    long timeElapsed = _elapsed;


    if (_isRunning)
    {
        // If the Stopwatch is running, add elapsed time since
        // the Stopwatch is started last time. 
        long currentTimeStamp = GetTimestamp();
        long elapsedUntilNow = currentTimeStamp - _startTimeStamp;
        timeElapsed += elapsedUntilNow;
    }
    return timeElapsed;
}


// Get the elapsed ticks.        
private long GetElapsedDateTimeTicks()
{
    long rawTicks = GetRawElapsedTicks();
    if (IsHighResolution)
    {
        // convert high resolution perf counter to DateTime ticks
        double dticks = rawTicks;
        dticks *= s_tickFrequency;
        return unchecked((long)dticks);
    }
    else
    {
        return rawTicks;
    }
}
```

IsHighResolution doc: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.ishighresolution?redirectedfrom=MSDN&view=netframework-4.7.2

so the conlusion of that stackoverlow post seems legit

> I just found out that ElapsedTicks in the StopWatch class doesn't mean real "ticks" if StopWatch.isHighResolution is True
> 
[...]

PR #2396, opened 2018-10-05
refactory of Stopwatch usage in DnsSeedServer

MithrilMan:
refactory of Stopwatch usage in DnsSeedServer class and documented Stopwatch implication of ElapsedTicks usage as per issue #2391

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
+++ b/src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
@@ -54,7 +54,7 @@ namespace Stratis.Bitcoin.Consensus.PerformanceCounters.ConsensusManager
 
         public double GetAvgExecutionTimeMs()
         {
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin/Consensus/PerformanceCounters/ConsensusManager/ConsensusManagerPerformanceSnapshot.cs
```diff
@@ -54,7 +54,7 @@ public ExecutionsCountAndDelay()
 
         public double GetAvgExecutionTimeMs()
         {
-            return Math.Round((this.totalDelayTicks / (double)this.totalExecutionsCount) / 10000.0, 4);
+            return Math.Round(TimeSpan.FromTicks(this.totalDelayTicks).TotalMilliseconds / this.totalExecutionsCount, 4);
```
dangershony: milliseconds is the same as dividing ticks by 1k?

MithrilMan: Not 1k but 10k, it was 1k until few days ago when I found it was a bug and fixed with 10k but I think getting rid of numbers and using timespan is cleaner (and I saw we did already in other places)

Issue #2371, opened 2018-10-04
MiningAndPropagatingPOS [Mining and Staking tests impl]


thefazzer:
**MiningAndPropagatingPOS [[Mining and Staking tests impl]**
- MineBlockCheckPeerHasNewBlock | Mine a block and propagate to peers (success)
- MineBlockStakeAtInsufficientHeightError | Mine a block insufficient height to stake (error)

PR #2397, opened 2018-10-05
MiningAndPropagatingPOS [Mining and Staking tests impl] (#2371).

thefazzer:
**MiningAndPropagatingPOS [Mining and Staking tests impl]**
- MineBlockCheckPeerHasNewBlock | Mine a block and propagate to peers (success)
- MineBlockStakeAtInsufficientHeightError | Mine a block insufficient height to stake (error)

Preferred: @fassadlr


fassadlr:
@thefazzer I hate to be a pain here but can we write these tests in the older:

`using(NodeBuilder nodeBuilder = nodeBuilder.Create())`

style? I don't think we are going to continue with the BDD style of tests.

Just to get some clarity here, these tests specifically test the scenarios where the node mines on a POS network? If this is the case then `MineBlockCheckPeerHasNewBlock` only has to mine one block on a POS test network (StratisRegTest) and then connect to another peer and ensure that the peer synced. Job done!

Re `MineBlockCheckPeerHasNewBlock` is this test supposed to test the fact that we can't mine after the network's LastPowBlock` height? If that's the case you only need to setup a test where you mine one block, set the `LastPowBlock`, value to 2 and then try and mine another block.

Any questions let me know! :)

thefazzer:
@fassadlr 
1) Separate test file for non-BDD spec mining & staking POS
2) MineBlockCheckPeerHasNewBlock & MineBlockCheckPeerHasNewBlock both follow the suggestions in review comments above

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSpecification.cs
```diff
@@ -15,5 +15,26 @@ public void Staking_wallet_can_mint_new_coins()
             When(pos_node_starts_staking);
             Then(pos_node_wallet_has_earned_coins_through_staking);
         }
+
+        [Fact]
+        public void MiningAndPropagatingPOS_MineBlockCheckPeerHasNewBlock()
```
fassadlr: @thefazzer you can now delete these files/tests.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeMintCoinsSteps.cs
```diff
@@ -45,5 +45,20 @@ private void pos_node_wallet_has_earned_coins_through_staking()
         {
             this.proofOfStakeSteps.PremineNodeWalletHasEarnedCoinsThroughStaking();
         }
+
+        private void pos_node_adds_peernode_and_propagate_blocks()
```
fassadlr: @thefazzer you can now delete these files/tests.

src/Stratis.Bitcoin.IntegrationTests/Miners/ProofOfStakeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Stratis.Bitcoin.IntegrationTests.Miners
+{
+    public class ProofOfStakeTests
+    {
+        [Fact]
+        public void MiningAndPropagatingPOS_MineBlockCheckPeerHasNewBlock()
+        {
+            using (NodeBuilder nodeBuilder = NodeBuilder.Create(this))
+            {
+                CoreNode node = nodeBuilder.CreateStratisPosNode(KnownNetworks.StratisRegTest).NotInIBD().WithWallet();
```
fassadlr: @thefazzer here you need to pass in a new instance of StratisRegTest otherwise, when you update the `LastPowBlock` it effects all other instances, which will affect tests that use that instance.

So pass in `new StratisRegTest` here and everywhere else in these tests.

src/Stratis.Bitcoin.IntegrationTests/ProofOfStakeSteps.cs
```diff
@@ -1,14 +1,17 @@
-﻿using System.Collections.Generic;
+﻿using System;
```
fassadlr: Revert all changes in this file.

Issue #2395, opened 2018-10-05
[Integration Tests] Implement coverage for consensus


fassadlr:

fassadlr:
![cm-cht](https://user-images.githubusercontent.com/4281271/48259635-7ea2ac80-e410-11e8-9c85-18b993cc7ea0.png)

fassadlr:
All of these scenarios has been implemented by myself.

PR #2398, opened 2018-10-05
[Consensus] Fix issue where miner reorgs + Tests

fassadlr:
This is to fix an issue where a miner reorgs but the reorg chain fails.

This also adds tests for for #2395 .


fassadlr:
I confirm that all the current re-org integration tests still pass 👍

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -393,6 +400,25 @@ private void ClaimPeerTip(int networkPeerId, uint256 tipHash)
             peersClaimingThisHeader.Add(networkPeerId);
         }
 
+        /// <inheritdoc />
+        public void ClaimPeerTipForMiner(uint256 tipHash)
+        {
+            HashSet<int> peersClaimingThisHeader;
```
dangershony: I suggest just call the existing method `ClaimPeerTip` then at the bottom of this method find the last MinerPeerId on the same branch and remove it.

Though I am not sure if w should remove it at all, we could always keep mined branches of blocks.

@noescape00 please review this fix as well (and maybe also @Aprogiena if available )

The reason we need to mark a mined block with its own special peer is that it does not get reorged away in case of a longer chain is presented and the node jumps to that chain.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -393,6 +400,25 @@ private void ClaimPeerTip(int networkPeerId, uint256 tipHash)
...
+            if (!this.peerIdsByTipHash.TryGetValue(tipHash, out peersClaimingThisHeader))
+            {
+                this.peerIdsByTipHash.Add(tipHash, new HashSet<int>(new[] { MinerPeerId }));
+            }
+            else
+            {
+                // Update the last miner tip hash with the new tip
+                var lastMinerTip = this.peerIdsByTipHash.FirstOrDefault(p => p.Value.Contains(MinerPeerId));
```
dangershony: This is probably wrong we need to walk back the current branch otherwise we risk removing mined blocks from parallel branches

fassadlr: Mining will always happen at the tip so how can parallel branches with the miner peer id happen? So every time we mine a block we are at tip so we need to update the last mined tip to the new one?

If my understanding is correct then "branches" refers to tips by peer id, so I'm not how a miner in this case can claim more than one tip? Seeing as its always at tip.

Just thinking out loud here trying to understand this better :)

fassadlr: Maybe give me a scenario of where this will be possible?

dangershony: yeah basically we can have more then one branch, but the miner will always extend the longest validated branch, however after a reorg we do not want to delete blocks that we mined on a shorter branch.
In the case bellow if we mined 6A, and then a reorg happened and we moved to chain B and then we mined block 8B we should have 6A marked as -2 and 8B, so you dont want to delete -2 from 6A
```
1A-2A-3A-4A-5A-6A
                -4B-5B-6B-7B-8B-9B
```

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -258,7 +258,9 @@ public async Task<ChainedHeader> BlockMinedAsync(Block block)
                     if (fullValidationRequired)
                     {
                         ConnectBlocksResult fullValidationResult = await this.FullyValidateLockedAsync(validationContext.ChainedHeaderToValidate).ConfigureAwait(false);
-                        if (!fullValidationResult.Succeeded)
+                        if (fullValidationResult.Succeeded)
+                            this.chainedHeaderTree.ClaimPeerTipForMiner(validationContext.ChainedHeaderToValidate.HashBlock);
```
dangershony: If we have an else with braces then I prefer them also on the if if you dont mind.

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -214,11 +230,39 @@ private void StartStratisRunner()
...
                 ((InitialBlockDownloadStateMock)this.FullNode.NodeService<IInitialBlockDownloadState>()).SetIsInitialBlockDownload(false, DateTime.UtcNow.AddMinutes(5));
 
             if (this.builderWithWallet)
                 this.Mnemonic = this.FullNode.WalletManager().CreateWallet(this.builderWalletPassword, this.builderWalletName, this.builderWalletPassphrase);
+
+            if (this.builderNoValidation)
+                DisableValidation();
+        }
+
+        public void DisableValidation()
```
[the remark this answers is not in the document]
fassadlr: OK yeah I have removed this switch 👍 and will just pass in a network that has no validation.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Done 👍

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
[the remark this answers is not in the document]
fassadlr: Actually, this isn't going to work as we always construct the rule set when we start up the node and then set the rules on Consensus after the fact. So in the tests we first need to start up the node and THEN clear the rules.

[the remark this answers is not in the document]
fassadlr: We can however pass in another network instance to this node that requires no validation, so that we don't accidentally affect the node that DO require validation ;)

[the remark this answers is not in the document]
dangershony: yes that's what I mean either pass an instance or a generic instance and create the network and remove the rules.

PR #2399, opened 2018-10-05
try xUnit retry package

monsieurleberre:
Please don't shout, I am just doing this PR as a test to see if the CI gives better results

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
@@ -194,7 +194,7 @@ namespace Stratis.Bitcoin.IntegrationTests.API
             Then(a_method_not_allowed_error_is_returned);
         }
 
-        [Fact]
+        [Retry]
         public void Proof_of_work_node_calls_generate_and_receives_error()
         {
             Given(a_proof_of_stake_node_with_api_enabled);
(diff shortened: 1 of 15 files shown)


src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
```diff
@@ -194,7 +194,7 @@ public void Proof_of_work_node_calls_getstakinginfo_and_receives_error()
             Then(a_method_not_allowed_error_is_returned);
         }
 
-        [Fact]
+        [Retry]
```
dangershony: What is the retry default attempts?

noescape00:
> I am just doing this PR as a test to see if the CI gives better results

But the CI doesn't work with integration tests because they are ignored

bokobza:
I think it's ok to have a retry for now and remove it later. This will help us during the build to see the failures for what they really are.

src/Stratis.Bitcoin.IntegrationTests/API/ApiSpecification.cs
bokobza: it's 3 by default.

noescape00:
Unit tests don't have severe issues that require retrying and integration tests are not ran by CI. 
I don't think we need retrying.

bokobza:
There is no retry on unit tests and we want to run the integration tests from the CI.
Also, even locally, I think we'd benefit from having some tests retry by themselves.

bokobza:
@dangershony also OK with this.

noescape00:
>There is no retry on unit tests and we want to run the integration tests from the CI.
>Also, even locally, I think we'd benefit from having some tests retry by themselves.

Ok. Then it's ok as a temporary thing till we fix tests to be stable but in a long run I'd prefer not having this.

PR #2404, opened 2018-10-07
[SC] Rule to check that we can get sender

codingupastorm:
Some grossness around the intialisation of concrete classes in the constructor, not sure if there's much we can do about it atm


src/Stratis.SmartContracts.Core/Util/ISenderRetriever.cs
```diff
@@ -15,5 +17,10 @@ public interface ISenderRetriever
         /// Get the 'sender' of a transaction.
         /// </summary>
         GetSenderResult GetSender(Transaction tx, ICoinView coinView, IList<Transaction> blockTxs);
+
+        /// <summary>
+        /// Get the 'sender' of a transaction in the mempool. Necessary because the MempoolCoinView has a very peculiar API.
+        /// </summary>
+        Task<GetSenderResult> GetSenderAsync(Transaction tx, MempoolCoinView coinView);
```
rowandh: Any reason this is Async/returns a Task?

src/Stratis.Bitcoin.Features.SmartContracts/Consensus/Rules/CanGetSenderRule.cs
```diff
@@ -0,0 +1,66 @@
...
+                processedTxs.Add(transaction);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private void CheckTransactionInsideBlock(Transaction transaction, ICoinView coinView, IList<Transaction> blockTxs)
+        {
+            // If wanting to execute a contract, we must be able to get the sender.
+            if (transaction.Outputs.Any(x => x.ScriptPubKey.IsSmartContractExec()))
```
rowandh: Could reuse this logic as it's the same as below.

codingupastorm: Thanks you are right but the GetSender call is different so it unfortunately isn't  a clean methodisation. Could  look at it in future though

PR #2407, opened 2018-10-07
Bug/2406 node shutdown

### LINKED ISSUE #2406: Process does not shutdown gracefully when shutdown API is used
There is a difference in the process of shutting down, depending if the daemon is shut down using CTRL-C (Windows) and stopped using the /api/Node/shutdown.

When using the API, which is what wallet UIs does, the main process will terminate before Dispose on the FullNode completes. It will end at random places in the Dispose, depending on how slow or fast the computer is.

This can cause various different errors, since it will end shutdown prematurely and persistent storage, etc. might have state not fully correct.

The main reason it does this, is that the Shutdown method on the API controller calls the Dispose directly:

```cs
// Start the node shutdown process.
this.fullNode?.Dispose();
```

Within Dispose, this is called:

```cs
// Fire INodeLifetime.Stopping.
this.nodeLifetime.StopApplication();
```

This will notify the main RunAsync method, which will start processing shutdown as well, and call Dispose, at which time the guard clause will make the code skip and continue:

```cs
if (this.State == FullNodeState.Disposing || this.State == FullNodeState.Disposed)
    return;
```

After a bit more processing, the main process exists, while there is another thread running the shutdown in Dispose, which never completes after the Exit.

The fix is a simple one line, it might have some consequences I'm not aware of, I have done some testing locally. PR coming very soon!

sondreb:
Closes #2406

- The daemon/node does not gracefully shut down when calling the shutdown API.
- Instead of calling Dispose directly in the Shutdown method, the StopApplication should be called, which will signal to the RunAsync wait to continue processing, which in turn calls node.Dispose.

### DIFF
--- a/src/Stratis.Bitcoin/Controllers/NodeController.cs
+++ b/src/Stratis.Bitcoin/Controllers/NodeController.cs
@@ -391,8 +391,9 @@ namespace Stratis.Bitcoin.Controllers
         [Route("stop")]
         public IActionResult Shutdown([FromBody] bool corsProtection = true)
         {
-            // Start the node shutdown process.
-            this.fullNode?.Dispose();
+            // Start the node shutdown process, by calling StopApplication, which will signal to
+            // the full node RunAsync to continuee processing, which calls Dispose on the node.
+            this.fullNode?.NodeLifetime.StopApplication();
 
             return this.Ok();
         }
--- a/src/Stratis.Bitcoin/Controllers/NodeController.cs
+++ b/src/Stratis.Bitcoin/Controllers/NodeController.cs
@@ -392,7 +392,7 @@ namespace Stratis.Bitcoin.Controllers
         public IActionResult Shutdown([FromBody] bool corsProtection = true)
         {
             // Start the node shutdown process, by calling StopApplication, which will signal to
-            // the full node RunAsync to continuee processing, which calls Dispose on the node.
+            // the full node RunAsync to continue processing, which calls Dispose on the node.
             this.fullNode?.NodeLifetime.StopApplication();
 
             return this.Ok();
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin/Controllers/NodeController.cs
```diff
@@ -391,8 +391,9 @@ public async Task<IActionResult> GetTxOutAsync([FromQuery] string trxid, uint vo
         [Route("stop")]
         public IActionResult Shutdown([FromBody] bool corsProtection = true)
         {
-            // Start the node shutdown process.
-            this.fullNode?.Dispose();
+            // Start the node shutdown process, by calling StopApplication, which will signal to
```
dangershony: Ah brilliant ofcourse this makes sense good spot @sondreb

dangershony:
I will test this tomorrow then commit.

src/Stratis.Bitcoin/Controllers/NodeController.cs
sondreb: Thanks Dan! One of those issues that takes some time to understand, but ends up being the simplest changes in code ;-)

TjadenFroyda: Rpc method uses dispose as well. May need similar update.

dangershony:
@sondreb please also fix RPC

src/Stratis.Bitcoin/Controllers/NodeController.cs
sondreb: Fixed now in the RPC stop method, changes to same as API stop/shutdown.

sondreb:
Sorry for not including the unit test fix, the builds are not successful, but I don't think the tests are related to my modifications? More work needed, or is it OK now?

PR #2413, opened 2018-10-08
Wallet manual tests fixes

bokobza:
**Should be reviewed commit by commit.**

These fixes bugs found during manually testing the wallet functionality.


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -408,6 +408,9 @@ public HdAccount GetUnusedAccount(Wallet wallet, string password)
 
                 // No unused account was found, create a new one.
                 account = wallet.AddNewAccount(password, this.coinType, this.dateTimeProvider.GetTimeOffset());
+                IEnumerable<HdAddress> newReceivingAddresses = account.CreateAddresses(this.network, this.walletSettings.UnusedAddressesBuffer);
```
fassadlr: Won't it be better to move this "creation" logic into `AddNewAccount`?

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -591,9 +594,11 @@ public AddressBalance GetAddressBalance(string address)
 
             lock (this.lockObject)
             {
+                HdAddress hdAddress = null;
+
                 foreach (Wallet wallet in this.Wallets)
                 {
-                    HdAddress hdAddress = wallet.GetAllAddressesByCoinType(this.coinType).FirstOrDefault(a => a.Address == address);
+                    hdAddress = wallet.GetAllAddressesByCoinType(this.coinType).FirstOrDefault(a => a.Address == address);
```
fassadlr: Maybe pass address into `GetAllAddressesByCoinType`, thereby creating an overload for the method?

bokobza:
Added commit that fixes https://github.com/stratisproject/StratisBitcoinFullNode/issues/2417

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
zeptin: Not entirely sure that would make sense, given the name of the method and its presumed purpose.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
```diff
@@ -328,7 +328,7 @@ public HdAccount GetFirstUnusedAccount()
             if (this.Accounts == null)
                 return null;
 
-            List<HdAccount> unusedAccounts = this.Accounts.Where(acc => !acc.ExternalAddresses.Any() && !acc.InternalAddresses.Any()).ToList();
+            List<HdAccount> unusedAccounts = this.Accounts.Where(acc => !acc.ExternalAddresses.SelectMany(add => add.Transactions).Any() && !acc.InternalAddresses.SelectMany(add => add.Transactions).Any()).ToList();
```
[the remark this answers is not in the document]
bokobza: yes, an unused account is defined as an account that has no transaction against any of its addresses. Look here: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery

We used to just look at the number of addresses, which was wrong.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: In theory yes. However, sometimes (like in cold staking), you want to just create an account and be in control of what you generate.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: I think this overload would be confusing given the name of the function.

src/Stratis.Bitcoin.Features.Wallet/Wallet.cs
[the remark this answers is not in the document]
fassadlr: 👍

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
fassadlr: OK makes sense, I'm just worried that if someone calls `AddNewAccount` from anywhere and expect addresses to be in there, there won't be?

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
fassadlr: `GetAllAddressesByCoinTypeAndAddress` ? LOL maybe that is too much. Anyway, all good then 👍

PR #2420, opened 2018-10-08
[PoA] Add rule for header bits

noescape00:


src/Stratis.Bitcoin.Features.PoA.Tests/PoAHeaderDifficultyRuleTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        public void ChainworkIsIncreasedBySameConstantValue()
+        {
+            BigInteger pow256 = BigInteger.ValueOf(2).Pow(256);
+            Target defaultTarget = PoAHeaderDifficultyRule.PoABlockDifficulty;
+            int headersCount = 200;
+
+            List<ChainedHeader> headers = ChainedHeadersHelper.CreateConsecutiveHeaders(headersCount, null, false, defaultTarget);
+
+            foreach (ChainedHeader chainedHeader in headers)
+                chainedHeader.InvokeMethod("CalculateChainWork");
```
bokobza: Not to do with your change, but I think this method `CalculateChainWork` should be made public and should be returning a `BigInteger` instead of assigning the `this.chainWork` field (so you 'd have `this.chainwork = CalculateChainWork()`).
Effectively, this method lies about what it's doing. It says it's calculating, but there is an undocumented side-effect.

bokobza:
Can you just look at my comment above please?

bokobza:
ah wait I didn't see the new changes.

src/Stratis.Bitcoin.Features.PoA.Tests/PoAHeaderDifficultyRuleTests.cs
noescape00: Agreed with your point. 
That call was when I was investigating why the test doesn't work. Found out it wasn't an issue but forgot to remove this call

noescape00:
@bokobza yes, removed those 2 lines per your comment, they were not needed.

Issue #2316, opened 2018-09-28
Slots manager - determine who mines when


noescape00:
We should know at which timestamp which pubkey should be used. 

Implement method:
```
    GetPubKeyForTimestamp(unixTimestamp)
```
It will throw if timestamp is invalid.

PR #2427, opened 2018-10-08
[PoA] Slots manager

noescape00:
Fix: #2316


src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
```diff
@@ -70,12 +74,19 @@ public override Task InitializeAsync()
 
             this.federationManager.Initialize();
 
+            if (this.federationManager.IsFederationMember)
```
bokobza: Concerning the keys:
At the moment a federation member will have to have a file in a certain format, in the right place, and the node will start to mine by itself because discovery of the file means they're a miner. It feels counter-intuitive to me and isn't in keeping with how other miners do it.

A better approach is to 
- have the keys in the config file
- ask the miners to start the node with -mine.

It's simpler, clearer for the miners and easier to test.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
...
+                    }
 
-                    // TODO sign the block with our signature
+                    // TODO POA sign the block with our signature
 
                     // Update merkle root.
                     blockTemplate.Block.UpdateMerkleRoot();
 
-                    ChainedHeader chainedHeader = await this.consensusManager.BlockMinedAsync(blockTemplate.Block).ConfigureAwait(false); // TODO POA That should also do interg vaidation
+                    // TODO POA That should also do intergity vaidation
```
bokobza: `integrity validation` but that's not that important right now.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
...
                     ChainedHeader tip = this.consensusManager.Tip;
 
                     BlockTemplate blockTemplate = this.blockDefinition.Build(tip);
 
+                    blockTemplate.Block.Header.Time = myTimestamp;
+
                     // Timestamp should only greater than prev one.
                     if (blockTemplate.Block.Header.Time <= tip.Header.Time)
-                        continue; // TODO POA Log
+                    {
```
bokobza: should always* be greater

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
             {
                 try
                 {
-                    // Don't mine in IBD.
-                    if (this.ibdState.IsInitialBlockDownload() && !MineDuringIBD)
+                    // Don't mine in IBD in case we are connected to any node.
+                    if (this.ibdState.IsInitialBlockDownload() && this.connectionManager.ConnectedPeers.Any())
```
bokobza: If we're in IBD and connected to no one, it will start mining. Is that the intention?

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
...
                         int attemptDelayMs = 20_000;
                         await Task.Delay(attemptDelayMs, this.cancellation.Token).ConfigureAwait(false);
 
                         continue;
                     }
 
+                    uint timeNow = (uint) this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
+                    uint myTimestamp = this.slotsManager.GetMiningTimestamp(timeNow);
+
+                    if (myTimestamp < timeNow)
```
bokobza: Can we do this in `GetMiningTimestamp`? I'd rather make sure we don't have a wacky method in there. :-)

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
...
+                    {
+                        // Sanity check.
+                        throw new Exception("Invalid mining timestamp generated.");
+                    }
 
-                    // TODO POA check if our timestamp and if not wait for our timestamp (wait with cancellation token)
+                    uint waitingTime = myTimestamp - timeNow - 1;
```
bokobza: why -1?

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +92,68 @@ private async Task CreateBlocksAsync()
...
 
-                    // TODO will set nonce and target to const. Check todo in PoABlockDefenition
+                    if (waitingTime > 0)
+                    {
+                        // Wait till we can mine.
+                        this.logger.LogInformation("Waiting {0} seconds before block can be mined.", waitingTime);
```
bokobza: We can even have this outside the if. "Waiting 0 seconds before .. " is also a useful message.
Also, before block -> until a block.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
```diff
@@ -19,10 +19,14 @@ public class PoANetwork : Network
...
         /// to public keys from this list are considered to be invalid.
         /// </remarks>
-        public List<PubKey> FederationPublicKeys { get; }
+        public List<PubKey> FederationPublicKeys { get; protected set; }
+
+        public uint TargetSpacingSeconds { get; protected set; }
 
         public PoANetwork()
         {
+            this.TargetSpacingSeconds = 16;
```
bokobza: Can PoANetwork be inheriting from Network? 
TargetSpacingSeconds could be part of `PoaConsensusOptions` and FederationPublicKeys don't have to be in Network, they can be in the .config file and that makes them easier to be modified by the user (no need to rebuild).
Also PoA should be Poa everywhere (small 'a').

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
```diff
@@ -49,10 +46,15 @@ public void Initialize()
                 }
 
                 this.logger.LogInformation("Federation key pair was successfully loaded. Your public key is: {0}.", this.FederationMemberKey.PubKey);
-
-                // Enable mining because we are a federation member.
-                this.miner.InitializeMining();
             }
+
+            var builder = new StringBuilder();
```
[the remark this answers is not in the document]
noescape00: That will produce: 

```
[2018-10-09 10:55:04.4463 3] INFO: Stratis.Bitcoin.FullNode.Method
Federation contains...
[2018-10-09 10:55:04.4463 3] INFO: Stratis.Bitcoin.FullNode.Method
"Key1"
"Key2"
```

I want to avoid having that `[2018-10-09 10:55:04.4463 3] INFO: ` suffix on each line

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Yes

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: To compensate time we spend mining the block.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: It's already done there. Here I have a 2nd layer of protection. That's why the comment says it's a sanity check- that should never happen.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
noescape00: >Also PoA should be Poa everywhere (small 'a').

Why? 

>Can PoANetwork be inheriting from Network?

It is inherited from network

>TargetSpacingSeconds could be part of PoaConsensusOptions and FederationPublicKeys don't have to be in Network, they can be in the .config file and that makes them easier to be modified by the user (no need to rebuild).

I think this is network specific. In config we never have values that can change consensus. 
Also not sure if having to rebuild it is a bad thing- creator of federation will have his own custom network and I expect him to override not only target spacing and fed keys but also other network values like base58 prefixes, premine amount, default ports and so on.

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
noescape00: The reason behind having this as a separated file is mostly the ability to back it up easily. Also separated file feels more important than a config value. I want ppl to treat it like a wallet file 🙂

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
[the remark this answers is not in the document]
bokobza: So:
```
this.logger.LogInformation($"Federation contains {this.network.FederationPublicKeys.Count} members. Their public keys are: {string.Join(Environment.NewLine, this.network.FederationPublicKeys)}");
```

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
bokobza: Still, it's not a wallet file and these should be backed up somewhere else. A couple of values are easier to backup in a password manager than a file.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: don't we want to be out of IBD first?

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: I don't see the point of adding it there if I'm perfectly honest.
It's not only that it should never happen, it's already checked. That should be enough.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: How do you evaluate this to be 1?

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
bokobza: > Why?

Poa like Pow or Pos or Dns or Api or Mvc :-)

> It is inherited from network

ah yes lol sorry.

> I think this is network specific. 

TargetSpacingSeconds is going to be checked in consensus rules so I see no reason why it can't be a consensus option.

For the keys, I don't think they're network specific. Network values should never change (apart from checkpoints and a few others). The pubkeys of the participants should be configurable. We already know that these are likely to be changed from time to time (members added or removed, members losing their keys or they get compromised).

> Also not sure if having to rebuild it is a bad thing

For user experience it's better if a member says to the other "my key has been compromised, here is the new one" and they can update their config file, rather than having to build, test and redistribute an executable.

> to override not only target spacing

Yeah he can do that in the PoaNetwork file, but the target spacing property is declared as part of PoaconsensusOptions.

> also other network values like base58 prefixes, premine amount, default ports and so on.

Of course

src/Stratis.Bitcoin.Features.PoA/FederationManager.cs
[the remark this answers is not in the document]
noescape00: Good point. Fixed

src/Stratis.Bitcoin.Features.PoA/PoAFeature.cs
noescape00: Ok, reasonable. 
I've added those changes to my TODO list. Will do them when other tasks are completed.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: We can't if we are not connected to anyone. 

So let's say we ran the node (we are the only miner on network), then we shutted it down for 2 days and then started up. Now we will never mine because we are in IBD. 

Also it's ok to mine even if we are on a disconnected part of the network- we will reorg to better chain anyway always (there is no MaxReoirg property so it doesn't matter for how long we are disconnected)

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Ok, removed

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Just a random value. 
Better to mine slightly earlier than always be behind

noescape00: Also it compensates precision problems since unix timestamp is in seconds.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
noescape00: >TargetSpacingSeconds is going to be checked in consensus rules so I see no reason why it can't be a consensus option.

Because network creator should be able to change this value. It's a network property, not consensus algho property.

>For user experience it's better if a member says to the other "my key has been compromised, here is the new one" and they can update their config file, rather than having to build, test and redistribute an executable.

Currently we don't support editing fed members except for via hardfork which also requires recompilation.

noescape00: However ok, not a bad idea on the second thought. 
I'll move those values to consensus options

noescape00: In consensus options we usually have variables that we can override or leave as it is. FedKeys are the values that we want to force people to set.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: Good point.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
bokobza: ok fair enough.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -75,42 +100,65 @@ private async Task CreateBlocksAsync()
...
+
+                    uint waitingTime = myTimestamp - timeNow - 1;
 
-                    // TODO POA check if our timestamp and if not wait for our timestamp (wait with cancellation token)
+                    this.logger.LogInformation("Waiting {0} seconds until block can be mined.", waitingTime);
 
-                    // TODO will set nonce and target to const. Check todo in PoABlockDefenition
+                    if (waitingTime > 0)
+                    {
+                        // Wait till we can mine.
```
bokobza: until*

noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/SlotsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+            uint roundTime = this.GetRoundLengthSeconds();
+
+            // Index of a slot that current node can take in each round.
+            uint slotIndex = (uint)this.network.FederationPublicKeys.IndexOf(this.federationManager.FederationMemberKey.PubKey);
+
+            // Time when current round started.
+            uint roundStartTimestamp = (currentTime / roundTime) * roundTime;
+            uint timestamp = roundStartTimestamp + slotIndex * this.network.TargetSpacingSeconds;
+
+            if (timestamp < currentTime)
```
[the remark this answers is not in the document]
noescape00: Fixed

src/Stratis.Bitcoin.Features.PoA/SlotsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            // Round length in seconds.
+            uint roundTime = this.GetRoundLengthSeconds();
+
+            // Index of a slot that current node can take in each round.
+            uint slotIndex = (uint)this.network.FederationPublicKeys.IndexOf(this.federationManager.FederationMemberKey.PubKey);
+
+            // Time when current round started.
+            uint roundStartTimestamp = (currentTime / roundTime) * roundTime;
+            uint timestamp = roundStartTimestamp + slotIndex * this.network.TargetSpacingSeconds;
```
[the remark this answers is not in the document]
noescape00: renamed to nextTimestampForMining since it's not timestamp in current round. might be in next round in case in this round our timestamp already passed

src/Stratis.Bitcoin.Features.PoA/SlotsManager.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            if (timestamp < currentTime)
+            {
+                // Get timestamp from next round.
```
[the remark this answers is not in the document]
noescape00: fixed

src/Stratis.Bitcoin.Features.PoA/SlotsManager.cs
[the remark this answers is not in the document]
bokobza: excellent

PR #2433, opened 2018-10-09
[SC] Storage improvements

### LINKED ISSUE #2379: [SC] Unused StateRepository in SmartContractCoinViewRule
Online 52 `IStateRepository trackedState = this.ContractCoinviewRule.OriginalStateRoot.StartTracking();`

This is inited but never actually used. I believe it should be used. That it's not suggests that the next-level-down may be pushing to db. Worth investigating.

codingupastorm:
- CoinviewRule processes blocks without putting the global state service in an "in-between" state
- Replaced overkill MultiCache classes with a repository of StorageCaches
- Removed ReadCache classes
- Appear to have fixed a bug with contract state not being persisted correctly

Closes #2379 

I'm very confident we can reduce the state down even more now that a bunch of the readcache stuff is gone, I think we can get rid of the whole 'Chained' idea. Will look into it in spare time.


src/Stratis.Bitcoin.Features.SmartContracts.Tests/ExecutorSpecification.cs
```diff
@@ -12,6 +12,7 @@ public class ExecutorSpecification
         [Fact]
         public void Create_Contract_Success()
         {
+            // TODO: Fix
```
codingupastorm: Just check what's happening in the ContractExecutor is aligned with your thoughts and I'll update these

PR #2451, opened 2018-10-11
[SC] Byte array param serialization

### LINKED ISSUE #2450: [SC] Improve Serialization of byte arrays in strings
We need to support string-based serialization of byte arrays. Currently they are being interpreted as strings and then back to byte arrays, which is unnecessary and susceptible to issues with encoding.

Recommend either:
* Encoding byte arrays as hex strings, eg. "0xAABBCC"
or
* Using the BitConverter.ToString format "AA-BB-CC

@codingupastorm

codingupastorm:
Closes #2450 

All SC tests and integration tests passing.


src/Stratis.SmartContracts.Core/Extensions.cs
```diff
@@ -14,10 +15,15 @@ public static string ToHexString(this byte[] ba)
...
         public static byte[] HexToByteArray(this string hex)
         {
-            int NumberChars = hex.Length;
+            string toHex = hex;
+
+            if (hex.StartsWith("0x"))
+                toHex = hex.Substring(2);
+
+            int NumberChars = toHex.Length;
             byte[] bytes = new byte[NumberChars / 2];
```
rowandh: This will not behave as expected if NumberChars is not even. We should consider zero-padding the hex if that's the case.

rowandh: Ah, I see there's a unit test for that already. I think zero-padding is probably preferable over outright failure.

rowandh: Fair point.

Issue #2146, opened 2018-09-07
Disconnect peers when node can't advance


dangershony:
Create an additional behaviour to support this functionality.
Possibly put this in `PosConsensusManagerBehaviour`

If a node is connected to the max outbounds limit, and all the nodes are not PH enabled or whitelisted we need to disconnect some nodes to try to make space for PH enabled nodes or our, otherwise we will not be able to advance our our consensus height.

dangershony:
A suggestion is that 50% of slots for outbound are saved for PH nodes.
We can also boost the number of maximum outbound nodes.

MithrilMan:
so do you want a settings like "phConnectionReserved" and disconnect any nodes that doesn't support PH, trying to occupy that slots?

MithrilMan:
I'll create a new behavior because I see PosConsensusManagerBehaviour is not (yet?) available

noescape00:
ProvenHeadersReservedSlotsBehavior (PHRSB)



Behavior should be subscribed to version message.

When new peers connects and starts handshaking (sends us version payload) we estimate if peer supports PH or not based on the version provided. 

If peer supports PH we check how many outbound connections we have. If we have all outbound slots taken we will disconnect single non-PH peer. 

If peer doesn't support PH and all outbound slots are taken we will drop the connection with this peer. 



In both cases if after this peer connects to us we still have at least one outbound slot we don't do anything, without checking if peer is PH or not.



Basically the idea is to have single slot reserved at all times except for cases when we have all our outbound connections to PH nodes.

MithrilMan:
ok so no more slots at all, makes sense and doesn't involve many external components

dangershony:
I think we also need to be aware that we do this check only for outbound connections (I think there is a flag in the behaviour or the node to check that), so check that the incoming connection is an outbound we initiated and its the last node in N allowed outbounds

PR #2403, opened 2018-10-06
[PH] DNS Seed nodes to prefer using PH enabled nodes

MithrilMan:
Implementation of #2147 

nodes that connects to our DNS node when PH gets activated, will be disconnected if they can't serve PH.

Extended the IFullNodeServiceProvider/FullNodeServiceProvider and used a better way to instance and register PeerBehavior like per issue #2402

Fixed tests

PR #2400, opened 2018-10-05
[PH] Extending ConsensusManagerBehaviour #2145

kogot:
this is for extend ConsensusManagerBehaviour to handle ProvenHeaders payload

PR #2469, opened 2018-10-11
[PH] Disconnect peers when node can't advance

MithrilMan:
implementation of #2146 

Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.

During IBD, all nodes are allowed.


src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -55,9 +66,57 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
+        /// <summary>
+        /// Processes "version" message received from the peer.
+        /// Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="version">Payload of "version" message to process.</param>
+        protected Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
+        {
+            // We enforce having free slots for PH enabled nodes only when we are not in IBD.
+            if (!this.initialBlockDownloadState.IsInitialBlockDownload())
```
dangershony: I wonder if we should not reserve place for PH nodes even when in IBD @noescape00 any idea?

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -55,9 +66,57 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
                 case GetProvenHeadersPayload getHeaders:
                     await this.ProcessGetHeadersAsync(peer, getHeaders).ConfigureAwait(false);
                     break;
+                case VersionPayload version:
+                    await this.ProcessVersionAsync(peer, version).ConfigureAwait(false);
```
dangershony: Do we need to check if PH is activated?

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -55,9 +66,57 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
+                        .Count();
+
+                    bool hasToReserveSlots = legacyPeersConnectedCount >= maxLegacyPeersAllowed;
+                    if (hasToReserveSlots)
+                    {
+                        // If we were previously in IBD state, we should remove legacy peers to reserve slots for PH enabled peers.
+                        int legacyPeersToDisconnect = legacyPeersConnectedCount - maxLegacyPeersAllowed;
+                        if (legacyPeersToDisconnect > 0)
+                        {
+                            var peersToDisconnect = connector.ConnectorPeers.OrderBy(p => p.PeerVersion.StartHeight).Take(legacyPeersToDisconnect).ToList();
```
dangershony: `p.PeerVersion.StartHeight` is not a good indicator, you need to find the correct `ProvenHeadersConsensusManagerBehavior` for that peer and look at its tip also you are not reading the correct linq query you need to filter out `connector.ConnectorPeers` bellow `PROVEN_HEADER_VERSION` also here right? if not you might disconnect PH enabled nodes.

But I think ther eis also another way to check if a peer is a PH peer

noescape00:
This logic shouldn't be in ProvenHeadersConsensusManagerBehavior, it should be in a separated behavior

ProvenHeadersConsensusManagerBehavior has only one responsibility- working with headers.

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
noescape00: We should. 
We shouldn't only before the last checkpoint.

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -22,11 +23,19 @@ public class ProvenHeadersConsensusManagerBehavior : ConsensusManagerBehavior
...
         private readonly ILoggerFactory loggerFactory;
         private readonly PosConsensusFactory consensusFactory;
+        private readonly IConnectionManager connectionManager;
+
+        /// <summary>
+        /// The proven header peers reserved slots threshold (%).
+        /// Rapresents the percentage of maximum connectable peers that we reserve for peers that are able to serve proven headers.
+        /// </summary>
+        private const decimal ProvenHeaderPeersReservedSlotsThreshold = 0.4M;
+
```
noescape00: redundant line

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -22,11 +23,19 @@ public class ProvenHeadersConsensusManagerBehavior : ConsensusManagerBehavior
         private readonly IPeerBanning peerBanning;
         private readonly ILoggerFactory loggerFactory;
         private readonly PosConsensusFactory consensusFactory;
+        private readonly IConnectionManager connectionManager;
+
+        /// <summary>
+        /// The proven header peers reserved slots threshold (%).
+        /// Rapresents the percentage of maximum connectable peers that we reserve for peers that are able to serve proven headers.
```
noescape00: typo

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -55,9 +66,57 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
             }
         }
 
+        /// <summary>
+        /// Processes "version" message received from the peer.
+        /// Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="version">Payload of "version" message to process.</param>
+        protected Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
```
noescape00: private

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -55,9 +66,57 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
+                    if (hasToReserveSlots)
+                    {
+                        // If we were previously in IBD state, we should remove legacy peers to reserve slots for PH enabled peers.
+                        int legacyPeersToDisconnect = legacyPeersConnectedCount - maxLegacyPeersAllowed;
+                        if (legacyPeersToDisconnect > 0)
+                        {
+                            var peersToDisconnect = connector.ConnectorPeers.OrderBy(p => p.PeerVersion.StartHeight).Take(legacyPeersToDisconnect).ToList();
+                            foreach (var peerToDisconnect in peersToDisconnect)
+                            {
+                                peer.Disconnect("Reserving connection slot for a Proven Header enabled peer.");
```
noescape00: you are disconnecting wrong peer 🙂

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
noescape00: Agreed. We should enable this policy only after the last checkpoint && PH are activated

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
noescape00: Also we shouldn't care about start height as long as policy is activated after we only can sync with PH and those nodes can't do it no matter the height

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: oooops

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: Did I have to change any logic here then?

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: While I agree because we call a method to process this payload, isn't better to don't have logic in this switches?
Even thinking about having a better publisher/subscriber implementation where a behavior can register at specific payloads and not every payload

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: I ended up doing this helper method 
```
private IEnumerable<INetworkPeer> GetConnectedLegacyPeersSortedByTip(NetworkPeerCollection connectedPeers)
{
    return from peer in connectedPeers.ToList() // not sure if connectedPeers can change, so i use ToList to get a snapshot
            let isLegacy = peer.PeerVersion.Version < NBitcoin.Protocol.ProtocolVersion.PROVEN_HEADER_VERSION
            let tip = peer.Behavior<ProvenHeadersConsensusManagerBehavior>()?.ExpectedPeerTip?.Height ?? 0
            where isLegacy
            orderby tip
            select peer;
}
```

MithrilMan:
> This logic shouldn't be in ProvenHeadersConsensusManagerBehavior, it should be in a separated behavior
> 
> ProvenHeadersConsensusManagerBehavior has only one responsibility- working with headers.

@dangershony you asked to include it in `ProvenHeadersConsensusManagerBehavior`, what should i do?

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
dangershony: Yes you need to check if the current header is bellow checkpoint/assumed-valid, if it is then we don't need to reserve slots, if its past both of them we need to reserve slots.

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
dangershony: yeah logic should go inside the method agreed

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
dangershony: yeah `connectedPeers` can suddenly change so good you took a snapshot.

Ok this linq query is nice but how easy will it be to change it in the future?
When we refactor `Protocol.ProtocolVersion` this will have to change a bit.

But this looks correct

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
dangershony: Well spotted @noescape00 🌝

dangershony:
>ProvenHeadersConsensusManagerBehavior has only one responsibility- working with headers.  

hmmm well I think as they are very much bound together (we will never separate the two logics) then its ok they are in the same behaviour, but there is not that much overhead with adding a new behaviour so I leave this decision to you @MithrilMan

MithrilMan:
What could be the name of an eventuale behaviour of this kind, something like ProvenHeadersConnectionManagerBehavior if it has to deal with connections?

src/Stratis.Bitcoin/Consensus/ProvenHeadersConsensusManagerBehavior.cs
dangershony: so basically what I think you need to do here is 
- First check if PH is activated
- Check if checkpoints or assumed valid (CP/AV)
- If not CP/AV enable slots reservation
- if yes CP/AV do nothing

Same logic of reservation needs to be applied when we pass the CP/AV but how do we trigger that.
Perhaps we actually need a timer (or background async loop) in this behaviour or a parent class that is shared with all this kind of behaviour types that will monitor consensus tip (CT.Tip) and trigger activation of slot reservation?

MithrilMan:
ok splitted the behavior in two and moved to consensus feature.

I'm aware that the check of the last checkpoint couldn't be the right one, please discuss about this.
`bool isBelowLastCheckpoint = this.consensusSettings.UseCheckpoints && (this.chainState.ConsensusTip.Height <= this.checkpoints.GetLastCheckpointHeight());`

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+
+      /// <summary>
+      /// Processes "version" message received from the peer.
+      /// Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.
+      /// </summary>
+      /// <param name="peer">Peer from which the message was received.</param>
+      /// <param name="version">Payload of "version" message to process.</param>
+      private Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
+      {
+         bool isBelowLastCheckpoint = this.consensusSettings.UseCheckpoints && (this.chainState.ConsensusTip.Height <= this.checkpoints.GetLastCheckpointHeight());
```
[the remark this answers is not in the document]
dangershony: You have an example in ChainedHeaderTree
```
            bool isAssumedValidEnabled = this.consensusSettings.BlockAssumedValid != null;
            bool isBelowLastCheckpoint = this.consensusSettings.UseCheckpoints && (earliestNewHeader.Height <= this.checkpoints.GetLastCheckpointHeight());
```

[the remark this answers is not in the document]
dangershony: @noescape00 can you please confirm this? the problem is we do not know we are on the assumed valid chain until we get the header that is associated with assumed valid, again if we find the header and we are on a different branch we will just disconnect the block.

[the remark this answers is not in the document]
dangershony: Ah I do remember that the ChainedHeader will have a BlockValidation flag that will be assumed valid for the entire branch of assumed valid

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+
+         // We enforce having free slots for PH enabled nodes only when we have passed last checkpoint.
+         if (!isBelowLastCheckpoint)
```
dangershony: I also noticed we don't check if PH is activated.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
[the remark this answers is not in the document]
dangershony: So this are the conditions to check.
- is PH active
- are we bellow AssumedValid (check the ChainedHeader attribute)
- are there less then treshold slots

noescape00:
>hmmm well I think as they are very much bound together (we will never separate the two logics) then its ok they are in the same behaviour, but there is not that much overhead with adding a new behaviour so I leave this decision to you @MithrilMan

After we switch to C# nodes and update min protocol version we won't need this logic anymore. 
So I'd have it as a separated behavior which will be deleted later on.

noescape00:
Conflicts

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,109 @@
...
+      private readonly IConsensusManager consensusManager;
+      private readonly IPeerBanning peerBanning;
+      private readonly ILoggerFactory loggerFactory;
+      private readonly PosConsensusFactory consensusFactory;
+
+      /// <summary>
+      /// The proven header peers reserved slots threshold (%).
+      /// Represents the percentage of maximum connectable peers that we reserve for peers that are able to serve proven headers.
+      /// </summary>
+      private const decimal ProvenHeaderPeersReservedSlotsThreshold = 0.4M;
```
noescape00: redundant

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,109 @@
...
+using Stratis.Bitcoin.P2P.Protocol;
+using Stratis.Bitcoin.P2P.Protocol.Payloads;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Behaviors
+{
+   /// <summary>Behavior that takes care of proven headers protocol. It also keeps the notion of peer's consensus tip.</summary>
+   public class ProvenHeadersConsensusManagerBehavior : ConsensusManagerBehavior
+   {
+      private readonly IInitialBlockDownloadState initialBlockDownloadState;
```
noescape00: most of those are needed for cloning and are already saved in the base class. Maybe it will be cleaner to make them protected on the base class so you don't need to create variables for them here

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,109 @@
...
+
+      /// <inheritdoc />
+      /// <summary>
+      /// Processes and incoming message from the peer.
+      /// </summary>
+      /// <param name="peer">Peer from which the message was received.</param>
+      /// <param name="message">Received message to process.</param>
+      protected override async Task OnMessageReceivedAsync(INetworkPeer peer, IncomingMessage message)
+      {
+         await base.OnMessageReceivedAsync(peer, message).ConfigureAwait(false);
```
noescape00: add empty line after that one

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -0,0 +1,109 @@
...
+         {
+            this.logger.LogTrace("(-)[INVALID_LOCATOR]:null");
+            return null;
+         }
+
+         var headers = new ProvenHeadersPayload();
+         foreach (ChainedHeader header in this.chain.EnumerateToTip(fork).Skip(1))
+         {
+            var posBock = new PosBlock(header.Header);
+            ProvenBlockHeader provenBlockHeader = this.consensusFactory.CreateProvenBlockHeader(posBock);
```
noescape00: this is wrong. we should get PH from PH store, not create it

noescape00: also we should get them in 1 batch from PH store

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
```
noescape00: lines 4,7,12 are redundant

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+using Stratis.Bitcoin.P2P;
+using Stratis.Bitcoin.P2P.Peer;
+using Stratis.Bitcoin.P2P.Protocol;
+using Stratis.Bitcoin.P2P.Protocol.Behaviors;
+using Stratis.Bitcoin.P2P.Protocol.Payloads;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.Consensus.Behaviors
+{
+   public class ProvenHeadersReservedSlotsBehavior : NetworkPeerBehavior
```
noescape00: Add it to template behaviors for pos

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+      }
+
+      /// <summary>
+      /// Processes and incoming message from the peer.
+      /// </summary>
+      /// <param name="peer">Peer from which the message was received.</param>
+      /// <param name="message">Received message to process.</param>
+      protected async Task OnMessageReceivedAsync(INetworkPeer peer, IncomingMessage message)
+      {
+         switch (message.Message.Payload)
```
noescape00: no need to do switch case here, if will be sufficient

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
noescape00: no need to check UseCheckpoints because if they are disabled last checkpointed height will be 0

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+         bool isBelowLastCheckpoint = this.consensusSettings.UseCheckpoints && (this.chainState.ConsensusTip.Height <= this.checkpoints.GetLastCheckpointHeight());
+
+         // We enforce having free slots for PH enabled nodes only when we have passed last checkpoint.
+         if (!isBelowLastCheckpoint)
+         {
+            PeerConnectorDiscovery connector = this.connectionManager.PeerConnectors.OfType<PeerConnectorDiscovery>().FirstOrDefault();
+            // If PeerConnectorDiscovery is not found means we are using other ways to connect peers (like -connect) and thus we don't enforce the rule.
+            if (connector != null)
+            {
+               int slotsReservedForProvenHeaderEnabledPeers = (int)Math.Round(connector.MaxOutboundConnections * ProvenHeaderPeersReservedSlotsThreshold, MidpointRounding.ToEven);
```
noescape00: please use shorter name and add comment to provide info you currently have in variable name

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,150 @@
...
+                  {
+                     var peersToDisconnect = this.GetConnectedLegacyPeersSortedByTip(connector.ConnectorPeers).Take(legacyPeersToDisconnect);
+                     foreach (var peerToDisconnect in peersToDisconnect)
+                     {
+                        peerToDisconnect.Disconnect("Reserving connection slot for a Proven Header enabled peer.");
+                     }
+                  }
+
+                  // If current connected peer doesn't serve Proven Header, disconnect it.
+                  if (version.Version < NBitcoin.Protocol.ProtocolVersion.PROVEN_HEADER_VERSION)
```
noescape00: I think that will disconnect 1 more peer than we intend to

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
noescape00: 👍 this logic should be executed when we are ahead of last checkpoint and ahead of PH activation height

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: ProvenHeadersConsensusManagerBehavior.cs is not my code, it shows as file changed because i moved it into consensus feature. I've removed my changes from there and moved on its own behavior. This changes comes from #2400 that was approved this way, I wouldn't change in this PR

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
MithrilMan: In a previous behavior i used a straight if on a single condition but Dan asked me to use switch.
https://github.com/stratisproject/StratisBitcoinFullNode/pull/2403#discussion_r223388625

Please guys, agree on this and update the code style guidline if this is a subject of PR reviews.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
MithrilMan: funny enough i pointed this out in Slack #fullnode channel before using it
https://stratisgroupworkspace.slack.com/archives/C9TP7RP2R/p1539269982000100

>does `decimal slotsReservedForProvenHeaderEnabledPeers` sounds too verbose? :nerd_face:

and i just got 2 ACK about this.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: Don't touch this class if its wrong we will fix it in a new PR

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
dangershony: We use switch in all other places lets stay consistent

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
dangershony: >please use shorter name  
 
What why is this a problem?

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
dangershony: Thi sis how I see it, we can use BlockValidationState to know if a header is inside checkpoints/assumedvalid.

However this is  a payload of version so we cant know any info about the peers height yes, ah but maybe it does tno matter.

When a peer connects we neeed to find the sate of our node, if our node is in assumed valid we dont need to reserve place for PH nodes.

The problem as I see it is how do we reserve slots once we are out of assumed valid.

noescape00:
conflicts

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <summary>
+        /// Processes "version" message received from the peer.
+        /// Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="version">Payload of "version" message to process.</param>
+        private Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
+        {
+            PeerConnectorDiscovery connector = this.connectionManager.PeerConnectors.OfType<PeerConnectorDiscovery>().FirstOrDefault();
+            // If PeerConnectorDiscovery is not found means we are using other ways to connect peers (like -connect) and thus we don't enforce the rule.
```
noescape00: add space before that line

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    // There is at least one free slot, so we don't enforce this peer to be PH enabled.
+                    return Task.CompletedTask;
+                }
+
+                bool peerSupportsPH = peer.Version >= NBitcoin.Protocol.ProtocolVersion.PROVEN_HEADER_VERSION;
+                if (peerSupportsPH)
+                {
+                    INetworkPeer nodeToDisconnect = GetConnectedLegacyPeersSortedByTip(connector.ConnectorPeers).FirstOrDefault();
+                    if (nodeToDisconnect != null)
```
noescape00: add space before that line

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+                    peer.Disconnect("Reserving connection slot for a Proven Header enabled peer.");
+                }
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private IEnumerable<INetworkPeer> GetConnectedLegacyPeersSortedByTip(NetworkPeerCollection connectedPeers)
+        {
+            return from peer in connectedPeers.ToList() // not sure if connectedPeers can change, so i use ToList to get a snapshot
```
noescape00: yes, connected peers can change

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+                if (freeSlots >= 1)
+                {
+                    // There is at least one free slot, so we don't enforce this peer to be PH enabled.
+                    return Task.CompletedTask;
+                }
+
+                bool peerSupportsPH = peer.Version >= NBitcoin.Protocol.ProtocolVersion.PROVEN_HEADER_VERSION;
+                if (peerSupportsPH)
+                {
+                    INetworkPeer nodeToDisconnect = GetConnectedLegacyPeersSortedByTip(connector.ConnectorPeers).FirstOrDefault();
```
noescape00: this.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
noescape00: not other ways but just -connect. with -addNode we will still have discovery enabled

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (connector != null)
+            {
+                int freeSlots = connector.MaxOutboundConnections - connector.ConnectorPeers.Count;
+                if (freeSlots >= 1)
+                {
+                    // There is at least one free slot, so we don't enforce this peer to be PH enabled.
+                    return Task.CompletedTask;
+                }
+
+                bool peerSupportsPH = peer.Version >= NBitcoin.Protocol.ProtocolVersion.PROVEN_HEADER_VERSION;
```
noescape00: it's not set at this point most likely. use version you get from payload

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="version">Payload of "version" message to process.</param>
+        private Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
+        {
+            PeerConnectorDiscovery connector = this.connectionManager.PeerConnectors.OfType<PeerConnectorDiscovery>().FirstOrDefault();
+            // If PeerConnectorDiscovery is not found means we are using other ways to connect peers (like -connect) and thus we don't enforce the rule.
+            if (connector != null)
+            {
+                int freeSlots = connector.MaxOutboundConnections - connector.ConnectorPeers.Count;
```
noescape00: ConnectorPeers.Count will give you only handshaked peers. So most likely you need to subtract 1 on top of it.

noescape00:
and still conflicts

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,111 @@
...
+            this.loggerFactory = loggerFactory;
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName, $"[{this.GetHashCode():x}] ");
+        }
+
+        /// <summary>
+        /// Processes and incoming message from the peer.
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="message">Received message to process.</param>
+        protected async Task OnMessageReceivedAsync(INetworkPeer peer, IncomingMessage message)
```
dangershony: Ah one more thing please look at `BlockStoreBehavior` I think we need to put a try catch for `OperationCanceledException`, this might be thrown when the node just disconnects

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
dangershony: I think we don't really care about -addnode, it does not take from the max outbound slots.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
MithrilMan: if you declare a variable used in the if, you space it with a blank line?

MithrilMan: ok as per slack, it is fine to keep it without space since prev line is a declaration of the variable used in the if

dangershony: yes its fine, this is dev preference so no problem

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
MithrilMan: As discussed, should happen only if you send something to node, so not needed.

noescape00:
Looks good. 
Also if you have time please add logs to ProcessVersionAsync so when we disconnect someone we log it on debug level.

MithrilMan:
logs added

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersReservedSlotsBehavior.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Processes "version" message received from the peer.
+        /// Ensures we leave some available connection slots to nodes that are enabled to serve Proven Headers, if the number of minimum PH enabled nodes connected isn't met.
+        /// </summary>
+        /// <param name="peer">Peer from which the message was received.</param>
+        /// <param name="version">Payload of "version" message to process.</param>
+        private Task ProcessVersionAsync(INetworkPeer peer, VersionPayload version)
+        {
+            PeerConnectorDiscovery connector = this.connectionManager.PeerConnectors.OfType<PeerConnectorDiscovery>().FirstOrDefault();
+            // If PeerConnectorDiscovery is not found means we are using other ways to connect peers (like -connect) and thus we don't enforce the rule.
```
noescape00:
```suggestion
            // If PeerConnectorDiscovery is not found it means we are using -connect and thus we don't enforce the rule.
```

PR #2472, opened 2018-10-11
Rpc allow ip any

majikandy:
bitcoind and stratisd support -rpcallowip=*  so added this support to Full node so that it is an easier direct swap for other services.  Eg Block explorer like Nako @dangershony


bokobza:
I'll check this one as well.

bokobza:
So we actually don't want to support that as it represents a security risk.
Bitcoin actually banned it from version 10.
Check here:
https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.10.0.md#rpc-access-control-changes

image

majikandy:
Ah ok, good find!  Feel free to close

bokobza:
Actually @maciejzaleski said that it already works with * so he'll create a bug so that we can disallow it.

bokobza:
Closing.

PR #2478, opened 2018-10-12
[SC] OptimizationLevel Release for Compiler

codingupastorm:
- Compiler allows choosing which OptimisationLevel. Default is Release.
- Wherever necessary, made tests Debug (mainly to make determinism pass)
- Updated the GasInjectorRewriter to be more cleaner, injection is now correct. Could use more tests here though.


src/Stratis.SmartContracts.Executor.Reflection/ILRewrite/CodeSegment.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+namespace Stratis.SmartContracts.Executor.Reflection.ILRewrite
+{
+    /// <summary>
+    /// A segment of code inside a method.
+    /// </summary>
+    public class CodeSegment
```
rowandh: 🌠

Issue #2483, opened 2018-10-12
DBreezeCoinView performance improvement


MithrilMan:
Ok, let's start linking the DBreeze documentation: https://github.com/hhblaze/DBreeze/blob/master/Documentation/_DBreeze.Documentation.actual.pdf

and the DBreezeCoinView class
https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs

Subject of the issue is how SaveChangeAsync is implemented.
I was trying to sync bitcoin chain, I've a super connection so I can download blocks very fast and so download speed isn't an issue.
What i was noticing was that my Consensus.Height was progressing veeeery slow, 1 block every few minutes

Digging into it, I saw that the bottleneck was the SaveChangeAsync that was taking too much (note that I'm not using an SSD for this sync)

I started to tweak from this line
https://github.com/stratisproject/StratisBitcoinFullNode/blob/22abd08c3dfc624f0c60c06cba887da78b43ce05/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs#L203

where the comment reason was "nobody knows what this do".
Well, actually looking at the documentation you can guess something

> 20130529 - Speeding up batch modifications (updates, random inserts) with
> Technical_SetTable_OverwriteIsNotAllowed instruction

> We CAN SEE SPEED DEGRADE, when ​we update batch of values or data-blocks or if we
> insert a batch of keys in random order and, especially, if these keys have high entropy.
> For such cases we have integrated new methods for transactions and for nested tables:
> tran.Technical_SetTable_OverwriteIsNotAllowed(”t1”);
> or
> var tblABC = tran.InsertTable<byte[]>(“masterTable”, new byte[] { 1 }, 0);
> tblABC .Technical_SetTable_OverwriteIsNotAllowed();
> - This technique is interesting for the transactions with specific batch modifications,
> where speed really matters. Only developer can answer this question and find a
> balance.
> - This technique is not interesting for the memory based data stores.
> - These methods work only inside of one transaction and must be called for every
> table or nested table separately, before table modification command.
> - When new transaction starts, overwrite automatically will be allowed again for all
> tables and nested tables.
> - Overwriting concerns all: search trie nodes, values and data blocks.
> - Remember always to sort batch ascending by key, before insert - it will economize
> HDD space.
[...]

PR #2484, opened 2018-10-12
Performance enhancment on SaveChangesAsync

MithrilMan:
refers to issue #2483

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
@@ -188,19 +188,20 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
         }
 
         /// <inheritdoc />
-        public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash, uint256 nextBlockHash, List<RewindData> rewindDataList = null)
+        public Task SaveChangesAsync(IList<UnspentOutputs> unspentOutputs, IEnumerable<TxOut[]> originalOutputs, uint256 oldBlockHash, uint256 nextBlockHash, List<RewindData> rewindDataList = null)
         {
-            List<UnspentOutputs> all = unspentOutputs.ToList();
-
-            int insertedEntities = 0;
-
             Task task = Task.Run(() =>
             {
+                int insertedEntities = 0;
+
                 using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
                 {
                     transaction.ValuesLazyLoadingIsOn = false;
                     transaction.SynchronizeTables("BlockHash", "Coins", "Rewind");
 
+                    // Speed can degrade when keys are in random order and, especially, if these keys have high entropy.
+                    // This settings helps with speed, see DBreeze documentations about details.
+                    // We should double check if this settings help in our scenario, or sorting keys and operations is enough.
                     // Refers to issue #2483. https://github.com/stratisproject/StratisBitcoinFullNode/issues/2483
                     transaction.Technical_SetTable_OverwriteIsNotAllowed("Coins");
 
@@ -215,23 +216,28 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
 
                         this.SetBlockHash(transaction, nextBlockHash);
 
-                        var toRemove = all.Where(utxo => utxo.IsPrunable == true).ToList();
-                        toRemove.Sort(UnspentOutputsComparer.Instance);
-                        for (int i = 0; i < toRemove.Count; i++)
-                        {
-                            var coin = toRemove[i];
-                            this.logger.LogTrace("Outputs of transaction ID '{0}' are prunable and will be removed from the database.", coin.TransactionId);
+                        // Here we'll add items to be inserted in a second pass.
+                        List<UnspentOutputs> toInsert = new List<UnspentOutputs>();
 
-                            transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
+                        foreach (var coin in unspentOutputs.OrderBy(utxo => utxo.TransactionId, new UInt256Comparer()))
+                        {
+                            if (coin.IsPrunable)
+                            {
+                                this.logger.LogTrace("Outputs of transaction ID '{0}' are prunable and will be removed from the database. {1}/{2}.", coin.TransactionId);
+                                transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
+                            }
+                            else
+                            {
+                                // Add the item to another list that will be used in the second pass.
+                                // This is for performance reasons: DBreeze is optimized to run the same kind of operations, sorted.
+                                toInsert.Add(coin);
+                            }
                         }
 
-
-                        var toInsert = all.Where(utxo => utxo.IsPrunable == false).ToList();
-                        toInsert.Sort(UnspentOutputsComparer.Instance);
                         for (int i = 0; i < toInsert.Count; i++)
                         {
                             var coin = toInsert[i];
-                            this.logger.LogTrace("Outputs of transaction ID '{0}' are NOT PRUNABLE and will be inserted into the database.", coin.TransactionId);
+                            this.logger.LogTrace("Outputs of transaction ID '{0}' are NOT PRUNABLE and will be inserted into the database. {1}/{2}.", coin.TransactionId, i, toInsert.Count);
 
                             transaction.Insert("Coins", coin.TransactionId.ToBytes(false), coin.ToCoins());
                         }
@@ -248,7 +254,7 @@ namespace Stratis.Bitcoin.Features.Consensus.CoinViews
                             }
                         }
 
-                        insertedEntities += all.Count;
+                        insertedEntities += unspentOutputs.Count;
                         transaction.Commit();
                     }
                 }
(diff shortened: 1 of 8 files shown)


src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -213,15 +215,25 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
 
                         this.SetBlockHash(transaction, nextBlockHash);
 
-                        all.Sort(UnspentOutputsComparer.Instance);
-                        foreach (UnspentOutputs coin in all)
+                        var toRemove = all.Where(utxo => utxo.IsPrunable == true).ToList();
```
dangershony: I suggest this could even be made better if you iterate over `all` only once by creating two lists in advance.
Then sort them independently.

MithrilMan: Yes it's true but very negligible because those are few ms per SaveChangesAsyn code (it's not in a loop) and keep the code compact.

I'd be worried more about using `IEnumerable<UnspentOutputs>` as a signature for this method, because we are always passing lists and arrays so we should use `IReadOnlyCollection` here, I already tried changed it in another local branch and it requires only to change the signature on classes that implement this method, no need to change anything else and the pros is that now you can handle `unspentOutputs` as a List (with Count property too) and no need to call `all = unspentOutputs.ToList()`

I could then change the code to have only an iteration on it, splitting items on these 2 arrays

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -200,7 +200,9 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
                 {
                     transaction.ValuesLazyLoadingIsOn = false;
                     transaction.SynchronizeTables("BlockHash", "Coins", "Rewind");
-                    //transaction.Technical_SetTable_OverwriteIsNotAllowed("Coins"); // Why it was there and what is it for? No one knows.
+
+                    // Refers to issue #2483. https://github.com/stratisproject/StratisBitcoinFullNode/issues/2483
```
noescape00: probably better just mention what it does

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -213,15 +215,25 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
...
+                            transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
+                        }
+
+
+                        var toInsert = all.Where(utxo => utxo.IsPrunable == false).ToList();
+                        toInsert.Sort(UnspentOutputsComparer.Instance);
+                        for (int i = 0; i < toInsert.Count; i++)
                         {
-                            this.logger.LogTrace("Outputs of transaction ID '{0}' are {1} and will be {2} to the database.", coin.TransactionId, coin.IsPrunable ? "PRUNABLE" : "NOT PRUNABLE", coin.IsPrunable ? "removed" : "inserted");
+                            var coin = toInsert[i];
```
noescape00: unvar

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -213,15 +215,25 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
 
                         this.SetBlockHash(transaction, nextBlockHash);
 
-                        all.Sort(UnspentOutputsComparer.Instance);
-                        foreach (UnspentOutputs coin in all)
+                        var toRemove = all.Where(utxo => utxo.IsPrunable == true).ToList();
+                        toRemove.Sort(UnspentOutputsComparer.Instance);
```
noescape00: why not instead of creating a list enumerate all.where(...).OrderBy() ?
same for code below

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -213,15 +215,25 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
...
+                            var coin = toRemove[i];
+                            this.logger.LogTrace("Outputs of transaction ID '{0}' are prunable and will be removed from the database.", coin.TransactionId);
+
+                            transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
+                        }
+
+
+                        var toInsert = all.Where(utxo => utxo.IsPrunable == false).ToList();
+                        toInsert.Sort(UnspentOutputsComparer.Instance);
+                        for (int i = 0; i < toInsert.Count; i++)
```
noescape00: foreach?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -213,15 +215,25 @@ public Task SaveChangesAsync(IEnumerable<UnspentOutputs> unspentOutputs, IEnumer
...
+                        var toRemove = all.Where(utxo => utxo.IsPrunable == true).ToList();
+                        toRemove.Sort(UnspentOutputsComparer.Instance);
+                        for (int i = 0; i < toRemove.Count; i++)
+                        {
+                            var coin = toRemove[i];
+                            this.logger.LogTrace("Outputs of transaction ID '{0}' are prunable and will be removed from the database.", coin.TransactionId);
+
+                            transaction.RemoveKey("Coins", coin.TransactionId.ToBytes(false));
+                        }
+
```
noescape00: 1 new line instead of 2

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
dangershony: ok sure this sounds like a good idea, we anyway just use lists so if you want to add this change to this PR fine with me.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
dangershony: Issue is ok but also yeah feel free to just write here stuff even if its bulky

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
dangershony: For example RewindData is a list so the other components should be the same

MithrilMan: >I'd be worried more about using IEnumerable as a signature for this method
What i mean here is that when we passed IEnumerable you did a .ToList() that just copied the passed list (because it's already a list or an array) so we were doing something unuseful.

About the iteration, I could optimize them doing these steps:
- changing the signature to be `IReadOnlyCollection`
- removing `List<UnspentOutputs> all = unspentOutputs.ToList();` because I've already a list (or better a IReadOnlyCollection)
- sort all the items
- iterate through all items in the list, checking if `IsPrunable` is true: if true, i remove the item, if false, I add the item to a `toInsert` list

at this point, i can iterate the toInsert list, that will be a subset of the initial list, to insert what I have to.

Total iterations: 1 and a half (or better 1 full iteration + 1 partial iteration of only `toInsert` items :)

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
MithrilMan: look my comment above https://github.com/stratisproject/StratisBitcoinFullNode/pull/2484#discussion_r225306605

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
MithrilMan: about foreach i see 2 cons
1 - performance ( see https://www.dotnetperls.com/for-foreach )

> Summary. In micro-benchmarking, introducing extra local variables with foreach-loops can impact performance. However, if those local variables are reused several times in the loop body, they can lead to a performance improvement.
> Thus:
> The for-loop is faster than the foreach-loop if the array must only be accessed once per iteration.
> 

- impossibility to know at which iteration we are
so we can't know how many items we are at or how many items are left and we can't log that information if we need, or debug easily.
I consider foreach a syntactic sugar that help readability in some scenario, but I prefer for over foreach when the loop is more than just `foreach x do ActionX`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
MithrilMan: Having said this, I don't think it's important how we create these lists because the impact is few ms per method call, and this method is called once every 30 or 60 seconds I don't remember.
The game changer, performance wise, on this method was the mixing of removal and insertion of items in a single loop and the `Technical_SetTable_OverwriteIsNotAllowed` instruction.

MithrilMan:
ok improved optimization and changed unspent arguments signature, from IEnumerable to IList

PR #2487, opened 2018-10-13
RPC getbalance - only use mature coins

mikedennis:
Bitcoin core only shows balance of coins that are mature with getbalance. Do similar.


src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -21,11 +21,18 @@ public void TestRpcGetBalanceIsSuccessful()
                 builder.StartAll();
                 RPCClient rpcClient = node.CreateRPCClient();
 
-                TestHelper.MineBlocks(node, 2);
-                TestHelper.WaitLoop(() => node.FullNode.GetBlockStoreTip().Height == 2);
+                int maturity = (int)KnownNetworks.RegTest.Consensus.CoinbaseMaturity;
+
+                // This shouldn't be necessary but is required to make this test pass.
+                // - looks like a bug in the wallet?
+                maturity--;
```
mikedennis: @bokobza I think this is a off by one bug in the wallet? This tests on bitcoin core with the decrement so I think we have bug or maybe them 😜

zeptin: Hmm, you may be onto something there, although I need to investigate in further detail:

`if (coin.IsCoinBase() && nSpendHeight - coin.nHeight < COINBASE_MATURITY)`

One way to know for sure would be to connect an SBFN regtest to a bitcoind regtest and try to create such a transaction as early as possible on bitcoind to see if SBFN accepts it.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -79,16 +79,16 @@ public NewAddressModel GetNewAddress()
         /// RPC method that returns the spendable balance of all accounts.
         /// Uses the first wallet and account.
         /// </summary>
+        /// <param name="minConfirmations">Only include transactions confirmed at least this many times. (default=1)</param>
+        /// <param name="isWatchOnly">Also include balance in watch-only addresses. (not used yet)</param>
```
zeptin: This does ultimately introduce a cross-feature dependency, which we may not want. But a parameter is harmless enough for the time being.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
zeptin: Ok, so if we think about this:

1. Genesis has height = 0
2. Let's say coinbase maturity = 100, which is the normal value
3. Generate a number of blocks exactly equal to maturity
4. Run `getbalance` on Core regtest - gives a zero result (this is backed up by examples that always use `generate 101` to give spendable coins)
5. Generate one more block (i.e. the equivalent of `generate 101` overall)
6. `getbalance` now returns 50

mikedennis: If that's the case I think we are different than everyone else - This test in NBitcoin passes on bitcoin core and a bunch of other coins. 
https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin.Tests/AltcoinTests.cs#L156-L169

mikedennis: Gosh even more confusing the bitcoin core docs say minconf parameter should default to 1 but I just checked the source and it's defaulted to 0. That being said, there still is something else going on as even with changing the default to 0 I still have to change that >= to > to get the test to pass.

zeptin: Yes, I think you are right and that there is a mistake in that method. I am trying to discern whether our original rationale was valid for incrementing the `countFrom` variable by 1.

I double checked, and the actual definition of a confirmation includes the block a transaction is initially mined in. So my example above is not strictly correct. I do not know why core only gets 50 spendable coins at height 101 then, though.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -79,16 +79,16 @@ public NewAddressModel GetNewAddress()
...
         /// <returns>Total spendable balance of the wallet.</returns>
         [ActionName("getbalance")]
         [ActionDescription("Gets wallets spendable balance.")]
-        public decimal GetBalance()
+        public decimal GetBalance(int? minConfirmations, bool? isWatchOnly)
```
noescape00: isWatchOnly isn't used anywhere

noescape00: maybe just default minConfirmations to 1?

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
dangershony: >or maybe them

I like your optimism 😄 if this was the case then we helped bitcoin ;)

dangershony: I think its true we need to change to `confirmationCount > network.Consensus.CoinbaseMaturity` @bokobza will also check it out now.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
bokobza: Is watchOnly needed for compatibility?
I prefer the signature `public decimal GetBalance(int minConfirmations = 1, bool isWatchOnly = false)`

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -21,11 +21,18 @@ public void TestRpcGetBalanceIsSuccessful()
                 builder.StartAll();
                 RPCClient rpcClient = node.CreateRPCClient();
 
-                TestHelper.MineBlocks(node, 2);
-                TestHelper.WaitLoop(() => node.FullNode.GetBlockStoreTip().Height == 2);
+                int maturity = (int)KnownNetworks.RegTest.Consensus.CoinbaseMaturity;
```
bokobza: Can you please use `new BitcoinRegTest().Consensus.CoinbaseMaturity;`, we're trying to move away from these static networks.

```suggestion
                int maturity = (int)new BitcoinRegTest().Consensus.CoinbaseMaturity;
```

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -21,11 +21,18 @@ public void TestRpcGetBalanceIsSuccessful()
...
+                // - looks like a bug in the wallet?
+                maturity--; 
+                
+                TestHelper.MineBlocks(node, maturity);
+                TestHelper.WaitLoop(() => node.FullNode.GetBlockStoreTip().Height == maturity);                             
+                Assert.Equal(Money.Zero, rpcClient.GetBalance());
 
-                Money balance = rpcClient.GetBalance();
-                Assert.Equal(Money.Coins(100), balance);
+                rpcClient.Generate(1);
```
bokobza: Why not use MineBlocks?
```suggestion
                TestHelper.MineBlocks(node, 1);
```

bokobza: Maybe the error by 1 you see is because rpcClient.Generate(1); doesn't sync the block and the wallet on time?
MineBlocks generates a block and wait for everything to sync. Can we try it?

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
dangershony: I think its ok `KnownNetworks` are created specifically for tests (I dont think its static)

mikedennis:
Made some updates as per some discussion here. I still think there is a wallet issue as changes Jeremy suggested still required the maturity--  or modifying the > for confirmationCount as mentioned above in order to pass the maturity check. 

Also I've found lots of the documentation about this rpc method out there is out of date, it's best to refer to core source code directly if you want to compare implementations https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcwallet.cpp#L676

bokobza:
@mikedennis @zeptin @dangershony the >= should be changed to a > in the wallet.
The maturity is the number of blocks **between** the output being mined and being able to spend it. So if the confirmation is 100, and we mine at block 15, the coins will be spendable at block 15 + 100 + 1.

Look here https://bitcoin.stackexchange.com/questions/1991/what-is-the-block-maturation-time

mikedennis:
Sounds good @bokobza I've added that change to this PR.

PR #2491, opened 2018-10-15
[SC] Transaction data serialization

### LINKED ISSUE #1830: [SC] Serialization of method parameters should be simpler.

rowandh:
Changes the byte encoding format for smart contract transaction data.
* Use fixed offsets for unchanging fields (opcode, gas price, gas limit)
* Use RLP to encode variable-length fields

Closes #1830.

### DIFF
--- a/src/Stratis.SmartContracts.Executor.Reflection/CallDataSerializer.cs
+++ b/src/Stratis.SmartContracts.Executor.Reflection/CallDataSerializer.cs
@@ -161,11 +161,6 @@ namespace Stratis.SmartContracts.Executor.Reflection
             }
         }
 
-        private static bool IsCreateContract(byte type)
-        {
-            return type == (byte)ScOpcodeType.OP_CREATECONTRACT;
-        }
-
         private static bool IsCallContract(byte type)
         {
             return type == (byte)ScOpcodeType.OP_CALLCONTRACT;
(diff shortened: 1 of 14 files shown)


src/Stratis.SmartContracts.Executor.Reflection/CallDataSerializer.cs
```diff
@@ -1,56 +1,65 @@
...
+        private readonly IMethodParameterSerializer methodParamSerializer;
+        private readonly IContractPrimitiveSerializer primitiveSerializer;
 
-        public CallDataSerializer(IMethodParameterSerializer methodParameterSerializer)
-        {
-            this.methodParamSerializer = methodParameterSerializer;
+        public CallDataSerializer(IContractPrimitiveSerializer primitiveSerializer)
+        {            
+            this.primitiveSerializer = primitiveSerializer;
+            this.methodParamSerializer = new MethodParameterByteSerializer(primitiveSerializer);
```
rowandh: Normally I'd inject this, but I feel this is OK here because method param serialization could be a responsibility of `CallDataSerializer`. It's only advantageous to keep it separate for testing purposes.

codingupastorm: Makes sense, it was the responsibility of this prior

PR #2504, opened 2018-10-15
[Integration Tests] Consensus reorg tests part 2

fassadlr:


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -96,6 +96,22 @@ public static bool IsNodeSynced(CoreNode node)
             return true;
         }
 
+        /// <summary>
+        /// Ensures a node is internally synced and at a given height.
+        /// </summary>
+        /// <param name="node">This node.</param>
+        /// <param name="height">At which height should it be synced to.</param>
+        public static bool IsNodeSyncedAtHeight(CoreNode node, int height)
```
dangershony: hmm perhaps when we create tests that have internal loops that wait for something we should put that on the method name?
Maybe something like `WaitForNodeSyncedToHeight`?

dangershony: Not on this PR just something to consider.

Issue #1695, opened 2018-07-21
Implement API call to run smart contract methods locally in SmartContractsController


codingupastorm:
Developers may want to return the result of a particular method locally without running the code on-chain. The primary use case will be in the retrieval of some static information. E.g. maybe there is a `GetAverageHoldings` method in an ICO contract that returns `totalSupply / investors.Count`.

They should be able to do this via the API.

We need a public API call, built into SmartContractsController that takes in all the required parameters  (contract address, method name, method parameters), executes a method, and returns a result. Ideally it will also return the amount of gas used. 

This functionality already exists roughly in the Executor.cs file, so it should be a matter of using that and returning the result.

Current Bounty: 20 STRAT

mwherman2000:
A nice extension (no bounty from me) would be to enable a symmetric programming model where an SC could also be deployed off-chain to support the wrapping of an identical implementation of a set of business rules around off-chain data. The existing Visual Studio SSC template test harness provides most of the source code you need to implement this capability.

codingupastorm:
When resolving a method, if given no params this should check all properties first.

PR #2512, opened 2018-10-16
[SC] Add standard token contract

rowandh:
In the same solution for now. Can be broken out into its own repository once IStandardToken is in a NuGet package.


rowandh:
Passing validation in SCT once the reference to `IStandardContract` is removed.

rowandh:
Passing validation in SCT once the reference to `IStandardToken` is removed.

src/Stratis.SmartContracts.Token/StandardToken.cs
```diff
@@ -0,0 +1,123 @@
+﻿using Stratis.SmartContracts;
+using Stratis.SmartContracts.Token;
+
+public class StandardToken : SmartContract, IStandardToken
+{
+    public StandardToken(ISmartContractState smartContractState, uint totalSupply)
```
dangershony: Beautiful, this is music to my ears 😄

src/Stratis.SmartContracts.Token/StandardToken.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            return false;
+        }
+
+        uint toBalance = GetBalance(to);
+
+        SetBalance(Message.Sender, senderBalance - amount);
+
+        checked
+        {
```
codingupastorm: Checked is enabled by default in our compiler - I wonder what we do to establish the best practice here: Suggest only using our compiler?

src/Stratis.SmartContracts.Token/StandardToken.cs
codingupastorm: This is so cool I'm tearing up 😂

src/Stratis.SmartContracts.Token/StandardToken.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public uint Allowance(Address owner, Address spender)
+    {
+        return PersistentState.GetUInt32($"Allowance:{owner}:{spender}");
+    }
+
+    public struct TransferLog
+    {
+        public Address From;
```
codingupastorm: [Index] all of these fields

src/Stratis.SmartContracts.Token/StandardToken.cs
```diff
@@ -0,0 +1,123 @@
...
+        public Address To;
+        public uint Amount;
+    }
+
+    public struct ApprovalLog
+    {
+        public Address Owner;
```
codingupastorm: [Index] also

src/Stratis.SmartContracts.Token/IStandardToken.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Interface for a standard smart contract token
+    /// </summary>
+    public interface IStandardToken
+    {
+        /// <summary>
+        /// Gets the total supply of tokens.
+        /// </summary>
+        uint TotalSupply { get; }
```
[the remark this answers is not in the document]
codingupastorm: Requiring users to call a method `get_PropName` isn't that user friendly otherwise

src/Stratis.SmartContracts.Token/StandardToken.cs
codingupastorm: * Maybe we don't need to index the amount? OpenZep seems to not

src/Stratis.SmartContracts.Token/StandardToken.cs
rowandh: I considered this, but figured it would be better practice for us to always use `checked` rather than rely on any specific compiler flags being set.

src/Stratis.SmartContracts.Token/IStandardToken.cs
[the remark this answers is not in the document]
codingupastorm: Keep the property, we will allow easy querying of properties. Tracked by #1695

src/Stratis.SmartContracts.Token/StandardToken.cs
```diff
@@ -0,0 +1,123 @@
+﻿using Stratis.SmartContracts;
+using Stratis.SmartContracts.Token;
+
+public class StandardToken : SmartContract, IStandardToken
```
codingupastorm: I think this class should just have a brief summary stating how it works - primarily what TotalSupply means and who receives the tokens on deployment.

PR #2524, opened 2018-10-17
[SC] Smart contracts on PoA

codingupastorm:
All injection is complete, the nodes are initialized. 

Need to start porting the integration tests but not sure where to start as the block generation API is different - we will discuss.


src/Stratis.Bitcoin.Features.SmartContracts/Consensus/SmartContractPoABlockHeader.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using NBitcoin;
+using NBitcoin.Crypto;
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Consensus
+{
+    public class SmartContractPoABlockHeader : SmartContractBlockHeader
```
codingupastorm: Ideally this would inherit from PoABlockHeader and add the SC header fields in, but because it uses the base serialization to get the hash right now, it's much easier to have the PoA fields in the "most" derived class

src/Stratis.Bitcoin.Features.SmartContracts/SmartContractPoABlockDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+using Stratis.SmartContracts.Core;
+using Stratis.SmartContracts.Core.State;
+using Stratis.SmartContracts.Core.Util;
+
+namespace Stratis.Bitcoin.Features.SmartContracts
+{
+    /// <summary>
+    /// Pushes everything to the <see cref="SmartContractBlockDefinition"/>, just amends the block difficulty for PoA.
+    /// </summary>
+    public class SmartContractPoABlockDefinition : SmartContractBlockDefinition
```
codingupastorm: Because there's more functionality in the SCBlockDefinition, it makes more sense to derive from it and add the PoA code rather than the other way around.

Unfortunately this sucks because sometimes we derive from the PoA class and sometimes we derive from the SC class

rowandh:
`PoAMockChain_Node_Builds` is failing

PR #2531, opened 2018-10-17
[Integration Tests] Consensus reorg tests part 3

fassadlr:
With this test we introduce a block observer feature for integration tests which can catch when reorg happened and execute  a given func at a certain block height. Once the func has been executed it will be executed again. This can be changed at a later stage to make this more configurable etc.


src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    if (chainedHeaderBlock.ChainedHeader.Previous.Height == 5)
+                    {
+                        TestHelper.Disconnect(minerA, minerC);
+                        return true;
+                    }
+
+                    return false;
+                }
+                minerA.BlockDisconnectInterceptor(interceptor);
```
dangershony: Would be cleaner if you did 
```
 var minerA = builder.CreateStratisPosNode(new StratisRegTest()).NotInIBD().WithWallet().BlockDisconnectInterceptor(interceptor).Start();
```

Or even just call it `Intercept(`

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
```diff
@@ -0,0 +1,75 @@
...
+                TestHelper.BuildBlocks.OnNode(minerB).Amount(5).Invalid(8, (node, block) => BlockBuilder.InvalidCoinbaseReward(node, block)).BuildAsync();
+
+                // Reconnect MinerA to MinerB.
+                TestHelper.Connect(minerA, minerB);
+
+                // MinerC should be disconnected from MinerA.
+                TestHelper.WaitLoop(() => !TestHelper.IsNodeConnectedTo(minerA, minerC));
+
+                // This will cause the reorg chain to fail at block 8 and roll back any changes.
+                TestHelper.WaitLoop(() => TestHelper.IsNodeSyncedAtHeight(minerA, 9));
```
dangershony: What the point of this test? you dont check there was a reorg (using the interceptor)

fassadlr: Have discussed on Slack 👍

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
fassadlr: In this case I need to first create the node that I wish to disconnect from (minerA) so I first have to set up the nodes and then inject the interceptor function.

Also, I want to be specific on what the interceptor is intercepting, so thought it better to call `BlockDisconnectInterceptor`?

Thoughts?

dangershony: ah yeah its going to be tricky so like this is good

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
dangershony: Good job very creative mate

PR #2542, opened 2018-10-17
[Integration Tests] TS101_Connectivity_CallAddNode

ferdeen:
Related to #2390.

When connecting with addnode connect to peer and any peers in the address manager database.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/Connectivity/ConnectivityTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Connectivity/ConnectivityTests.cs
@@ -7,8 +7,8 @@ using NBitcoin;
 using Stratis.Bitcoin.Connection;
 using Stratis.Bitcoin.IntegrationTests.Common;
 using Stratis.Bitcoin.IntegrationTests.Common.EnvironmentMockUpHelpers;
+using Stratis.Bitcoin.Networks;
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.IntegrationTests/Connectivity/ConnectivityTests.cs
```diff
@@ -0,0 +1,66 @@
...
+                var peerAddNodeConnector = syncerConnectionMgr.PeerConnectors.
+                    Where(p => p.GetType() == typeof(PeerConnectorAddNode)).First() as PeerConnectorAddNode;
+
+                peerAddNodeConnector.Should().NotBeNull();
+                peerAddNodeConnector.Should().BeOfType<PeerConnectorAddNode>();
+
+                // Adding these endpoints will add the items to the internal address manager.
+                peerAddNodeConnector.ConnectionSettings.AddNode = new List<IPEndPoint>() { node2.Endpoint, node3.Endpoint };
+                peerAddNodeConnector.Initialize(syncerConnectionMgr);
+                await peerAddNodeConnector.OnConnectAsync();
```
fassadlr: I dont (think) that you need to call this as the PeerConnector base class runs an async loop that periodically calls each connector anyway. Maybe remove this and check that the test still pass?

PR #2545, opened 2018-10-18
Fixing the ReadWriteHash functionality

codingupastorm:
I merged #2538 but unfortunately it wasn't quite correct. 

This fixes the broken build.

We have separated the serialization of blocks from hash retrieval, as there are cases where we wish to add to the serialization but not to the hash (e.g. PH, PoA).

And base.ReadWrite wasn't going to last because we have to add to add extra fields to the hash in the case of smart contracts or other extensions

### DIFF
--- a/src/NBitcoin/BlockHeader.cs
+++ b/src/NBitcoin/BlockHeader.cs
@@ -98,7 +98,12 @@ namespace NBitcoin
         /// <summary>Populates stream with items that will be used during hash calculation.</summary>
         protected virtual void ReadWriteHashingStream(BitcoinStream stream)
         {
-            this.ReadWrite(stream);
+            stream.ReadWrite(ref this.version);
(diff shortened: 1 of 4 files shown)


dangershony:
These does not seem so correct to me, but I'll apply the fix in a separate PR

src/NBitcoin/BlockHeader.cs
```diff
@@ -98,7 +98,12 @@ public virtual void ReadWrite(BitcoinStream stream)
         /// <summary>Populates stream with items that will be used during hash calculation.</summary>
         protected virtual void ReadWriteHashingStream(BitcoinStream stream)
         {
-            this.ReadWrite(stream);
+            stream.ReadWrite(ref this.version);
```
noescape00: if this.readwrite doesn't work for your maybe we should move those lines to separated method and call this method from readwrite and readwriteHashingStream

PR #2517, opened 2018-10-16
[Integration Tests] Mempool additional tests 2

zeptin:
Several of these are placeholders until the relevant issues are fixed.

PR #2551, opened 2018-10-18
Mempool time bug

dangershony:
This is a mempool consensus change so should be reviewed carefully.

Required for https://github.com/stratisproject/StratisBitcoinFullNode/pull/2517


src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -621,6 +609,18 @@ private void CheckStandardTransaction(MempoolValidationContext context, bool wit
                 context.State.Fail(MempoolErrors.Version).Throw();
             }
 
+            if (this.network.Consensus.IsProofOfStake)
+            {
+                long adjustedTime = this.dateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
+                PosFutureDriftRule futureDriftRule = this.consensusRules.GetRule<PosFutureDriftRule>();
+
+                // nTime has different purpose from nLockTime but can be used in similar attacks
+                if (tx.Time > futureDriftRule.GetFutureDrift(adjustedTime))
```
noescape00: Do we have exactly same logic for C++ nodes? 
I mean exactly the same adjusted time calculation, future drift and this rule in place for mempool validation?

dangershony:
Referencing this from the other PR
```
         // Timestamp limit
            if (tx.nTime > GetAdjustedTime() || (fProofOfStake && tx.nTime > pblock->vtx[0].nTime))
                continue;
```
This PR should probably also implement this code, as this is an attack vector if not implemented together.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
zeptin: @dangershony you may have an error here - `GetFutureDrift()` only returns the constant `15`. It does not add the drift amount to the timestamp (the implementation is somewhat different to X, hence the discrepancy). So this would reject every transaction from the mempool I think, as every `.Time` would be `>15`.

dangershony:
I will write a test tomorrow for this PR

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
```diff
@@ -471,7 +471,7 @@ private void OnlyUnconfirmed(TxMempool.SetEntries testSet)
         /// <summary>
         /// Test if a new package would "fit" in the block.
         /// </summary>
-        private bool TestPackage(long packageSize, long packageSigOpsCost)
+        protected virtual bool TestPackage(TxMempoolEntry entry, long packageSize, long packageSigOpsCost)
```
zeptin: There are also classes inheriting `BlockDefinition` in the smart contracts feature. We will possibly want the future drift logic in the PoS version of them?

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
```diff
@@ -875,5 +885,43 @@ public void MiningAndPropagatingPOW_MineBlockCheckNodeConsensusTipIsCorrect()
                 Assert.Equal(5, miner.FullNode.ConsensusManager().Tip.Height);
             }
         }
+
+        [Fact]
+        public void Mniner_PosNetwork_CreatePowTransaction_AheadOfFutureDrift_ShouldNotBeIncludedInBlock()
```
zeptin: Typo, `Miner`

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -572,7 +560,7 @@ protected virtual void PreMempoolChecks(MempoolValidationContext context)
 
             // Coinstake is only valid in a block, not as a loose transaction
             // TODO: mempool needs to have seprate checks for POW/POS as part of the change to rules.
-            if (context.Transaction.IsCoinStake)
+            if (this.network.Consensus.IsProofOfStake && context.Transaction.IsCoinStake)
```
zeptin: This breaks one of the mempool unit tests but I'll fix the test in my open PR.

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
```diff
@@ -875,5 +885,43 @@ public void MiningAndPropagatingPOW_MineBlockCheckNodeConsensusTipIsCorrect()
...
+                // Send coins to the receiver
+                var context = WalletTests.CreateContext(network, new WalletAccountReference(WalletName, Account), Password, new Key().PubKey.GetAddress(network).ScriptPubKey, Money.COIN * 100, FeeType.Medium, 1);
+                
+                Transaction trx = stratisMiner.FullNode.WalletTransactionHandler().BuildTransaction(context);
+
+                // This should make the mempool reject a POS trx.
+                trx.Time = Utils.DateTimeToUnixTime(Utils.UnixTimeToDateTime(trx.Time).AddMinutes(5));
+
+                // Sign trx again after changing the time property.
+                trx = context.TransactionBuilder.SignTransaction(trx);
```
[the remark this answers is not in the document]
dangershony: yes... come to think about it I think its should be on by default no?

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
dangershony: Smart contracts are POA no future drift ASAIK right @noescape00 ?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
dangershony: ah really? this is correct though right?

src/Stratis.Bitcoin.Features.Miner/PosBlockDefinition.cs
```diff
@@ -64,5 +70,22 @@ public override void UpdateHeaders()
...
+        {
+            if (this.futureDriftRule == null)
+                this.futureDriftRule = this.ConsensusManager.ConsensusRules.GetRule<PosFutureDriftRule>();
+
+            long adjustedTime = this.DateTimeProvider.GetAdjustedTimeAsUnixTimestamp();
+
+            if (entry.Transaction.Time > adjustedTime + this.futureDriftRule.GetFutureDrift(adjustedTime))
+                return false;
+
+            if (entry.Transaction.Time > this.block.Transactions[0].Time)
```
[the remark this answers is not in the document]
dangershony: Well this is the logic in X, but I think coinbase and block time are the same in any case.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
zeptin: Yes; the test was taking a shortcut defining a coinstake on a non-PoS network.

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
zeptin: Note that the bugfix rule does appear in the smart contracts PoS rule registrations, but the PoA version might have a different intended design.

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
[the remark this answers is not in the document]
zeptin: It is off for testnet/regtest networks unless explicitly enabled like you did for the other test in the PR. But ok, if this was intentional here then that's fine.

src/Stratis.Bitcoin.Features.Miner/BlockDefinition.cs
codingupastorm: Our PoS implementation is currently broken in other areas. So no worries - it would be nice if you could chuck in our SmartContractPosPowBlockDefinition a `TODO: If reimplementing Smart contract PoS look at time drift` but otherwise we won't be using that for now so doesn't affect us.

PoA I'm not sure about - @noescape00 if you could let us know if PoA requires anything here I'm happy to update

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
```diff
@@ -36,6 +41,11 @@ namespace Stratis.Bitcoin.IntegrationTests
 {
     public class MinerTests
     {
+        private const string Password = "password";
+        private const string WalletName = "mywallet";
+        private const string Passphrase = "passphrase";
```
fassadlr: This isn't being used.

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -6,16 +6,25 @@
...
+using Stratis.Bitcoin.Tests.Common;
 using Xunit;
 
 namespace Stratis.Bitcoin.IntegrationTests.Mempool
 {
     public class MemoryPoolTests
     {
+        private const string Password = "password";
+        private const string WalletName = "mywallet";
+        private const string Passphrase = "passphrase";
```
fassadlr: This isn't being used.

fassadlr:
@dangershony there's few tests failing here which I think is definitely related to your changes:

[CI Build](https://dev.azure.com/StratisProject/StratisBitcoinFullNode/_build/results?buildId=1605&view=ms.vss-test-web.test-result-details)

src/Stratis.Bitcoin.Features.Miner/PosBlockDefinition.cs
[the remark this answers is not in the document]
dangershony: hmm I think the reason is block time is not set at this point yet, but trx time is set to current adjusted time (so basically its a way to check that the trx in mempool is not ahead of current time)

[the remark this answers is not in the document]
dangershony: Anyway later when we stake we adjust the block and trx time to the coinstake masking

src/Stratis.Bitcoin.IntegrationTests/MinerTests.cs
dangershony: Its used in the test bellow I believe (you don't see the reference?)

src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
dangershony: Same here used in this test `Mempool_SendPosTransaction_AheadOfFutureDrift_ShouldRejectByMempool`

dangershony:
Ok merged master so tests should now pass (unless some mempool tests will fail but @zeptin is aware of that)

PR #2555, opened 2018-10-18
Performance improvements

dangershony:


src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
```diff
@@ -197,14 +197,18 @@ private Task OnMessageReceivedAsync(INetworkPeer peer, IncomingMessage message)
         {
             if (message.Message.Payload is BlockPayload block)
             {
+                block.Obj.Header.PrecomputeHash(true, true);
```
zeptin: Just an observation - there is potential for a DoS vector here, as the entire block & all transactions are getting hashed before any rules get applied to it.

Perhaps the risk is minor though, as ostensibly we will only request blocks for which we already have a valid header.

@noescape00 what do you think?

dangershony: No the PrecomputeHash method has an option to tell it to lazy load, so only when the hash is calculated we put it in cache.

zeptin: Ah ok, I made an (incorrect) assumption about what `PrecomputeHash` does based on the name.

bokobza:
Did you sync the chains from the beginning?

dangershony:
>Did you sync the chains from the beginning?  

No, I synced STRAT main to 300k and BTC main to 70k.
I will let my nodes run over night with this change

src/Stratis.Bitcoin/BlockPulling/BlockPullerBehavior.cs
noescape00: >as the entire block & all transactions are getting hashed before any rules get applied to it.

only stuff from header gets hashed. Txes affect merkle tree root which is inside of a header.
so should be ok to do it.

dangershony:
I did consider that as well, hmm not a bad idea in the merkle root rule we anyway iterate trx so we could also cache the hash, however that means it wont be cached if called anyplace before hitting that rule, but I don't think this is the case, ok I will make the change.

dangershony:
@noescape00 FYI

PR #2557, opened 2018-10-19
[SC] Gas amendments and recursive call integration tests

codingupastorm:


codingupastorm:
All SC unit tests pass. Running SC integration tests

codingupastorm:
Note: Just need to update the wallet tests which I believe have incorrect gas limits coded in.

codingupastorm:
All SC and integration tests passing

PR #2559, opened 2018-10-19
[Integration Tests] Remove IsNodeConnectedTo in TestHelper disconnect

fassadlr:


src/Stratis.Bitcoin.IntegrationTests.Common/TestHelper.cs
```diff
@@ -269,9 +269,6 @@ public static List<Transaction> Reorder(List<Transaction> transactions)
         /// <param name="nodeToDisconnect">The node that will be disconnected.</param>
         public static void Disconnect(CoreNode thisNode, CoreNode nodeToDisconnect)
         {
-            if (!IsNodeConnectedTo(thisNode, nodeToDisconnect))
```
dangershony: This was wrong finally?

fassadlr: This is wrong, it is the test that is correct. Will fix.

Issue #2558, opened 2018-10-19
ConsensusManagerBehavior ProcessHeadersAsync signature


MithrilMan:
Currently ProcessHeadersAsync is invoked by ConsensusManagerBehavior and some specialized classes like PoAConsensusManagerBehavior` and `ProvenHeaderConsensusManagerBehavior`

e.g. in ConsensusManagerBehavior 
https://github.com/stratisproject/StratisBitcoinFullNode/blob/aeffe82db781650cd2f201718519c8454d6b5567/src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs#L139-L141

in PoAConsensusManagerBehavior (similar to ProvenHeaderConsensusManagerBehavior)
https://github.com/stratisproject/StratisBitcoinFullNode/blob/aeffe82db781650cd2f201718519c8454d6b5567/src/Stratis.Bitcoin.Features.PoA/PoAConsensusManagerBehavior.cs#L39-L41

now the signature of the method is this
https://github.com/stratisproject/StratisBitcoinFullNode/blob/aeffe82db781650cd2f201718519c8454d6b5567/src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs#L244

This way, it differs a lot from other kind of method that handle messages, becfause instead of having a Payload as an argument, it has a property of that payload and it sounds wrong to me.

In order to solve this, we can create an interface like `IHeaderPayload` where we ask for a method like
`List<BlockHeaders> GetHeaders()`
this way we can get use that method to  get Headers property rather than relying on current headers injection in the ProcessHeadersAsync  method.

Issue #2553, opened 2018-10-18
GetProvenHeadersPayload is not used, nodes ask always legacy GetHeadersPayload


MithrilMan:
Current implementation of PH CMB
I see it inherits from `ConsensusManagerBehavior` and inside that class, there are a lot of call to a `ResyncAsync` method that sends a `getheader` payload to the peer and I don't see any call to `new GetProvenHeadersPayload` that I was expecting to see if we want to ask proven headers.

We should have `ResyncAsync` as virtual so we can override it.

noescape00:
@dangershony we need to prioritize this one. 
Good catch @MithrilMan 👍

Issue #1793, opened 2018-08-01
Enable synching from whitelisted nodes


noescape00:
Enable synching using old headers only from whitelisted nodes (stratisX)

MithrilMan:
I can't do this because it depends on https://github.com/stratisproject/StratisBitcoinFullNode/issues/2145

PR #2560, opened 2018-10-19
whitelisted nodes can sync in a legacy way (getheaders)

MithrilMan:
address #1793 and #2553 

I had to refactor a bit ConsensusManagerBehavior but I pointed out an issue (that I didn't solved in this PR) about how we process headers, see #2558


src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +102,88 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        /// Processes the legacy GetHeaders message.
+        /// Only whitelisted legacy peers are allowed to handle this message.
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <param name="headers">The headers.</param>
+        /// <returns></returns>
+        protected Task ProcessLegacyHeadersAsync(INetworkPeer peer, List<BlockHeader> headers)
+        {
+            // Only whitelisted legacy peers are allowed to handle this message.
+            if (!IsPeerPhEnabled(peer) && IsPeerWhitelisted(peer))
```
dangershony: Very nice so basically, if we get a normal header (without the proofs its not fake) and PH is not activated or the peer is white listed we allow to processes it.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +102,88 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+                return new GetProvenHeadersPayload()
+                {
+                    BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                    HashStop = null
+                };
+            }
+            // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
+            else if (IsPeerWhitelisted(peer))
+            {
+                return new GetHeadersPayload()
```
dangershony: return `base.BuildGetHeadersPayload()` perhaps?

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -442,25 +442,39 @@ public async Task ResyncAsync()
 
             if ((peer != null) && (peer.State == NetworkPeerState.HandShaked))
             {
-                var headersPayload = new GetHeadersPayload()
+                var headersPayload = this.BuildGetHeadersPayload();
```
dangershony: I wonder if this is correct though, this is about sending headers, we should be ok to send normal headers to none PH nodes (legacy X)

dangershony: The logic is if the peer is PH we send PH headers is enabled else we send regular headers (regardless of peer being whitelisted) that means X can still sync form #

MithrilMan: But here you are requesting headers, not sending, am I wrong?

MithrilMan: So my logic was
If we are connected to a legacy peer, we don't have to request normal headers to that peer except if he's whitelisted.

On the other hand If a legacy peer ask us blocks, we can serve it and it should work because we rely on 
```
// Rely on base.OnMessageReceivedAsync only if the message hasn't be already processed.
await base.OnMessageReceivedAsync(peer, message).ConfigureAwait(false);
```

that actually on my implementation can just handle the case
```
case GetHeadersPayload getHeaders:
   await this.ProcessGetHeadersAsync(peer, getHeaders).ConfigureAwait(false);
   break;
```

this mean that if a peer asks us legacy headers sending us a `getheaders` we serve it in a legacy way

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +102,81 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+
+            if (IsPeerPhEnabled(peer))
+            {
+                return new GetProvenHeadersPayload()
+                {
+                    BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                    HashStop = null
+                };
+            }
+            // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
```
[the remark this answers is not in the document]
noescape00: we should send normal headers to nodes that don't support them and PH to nodes that do. Doesn't matter if they are whitelisted or not.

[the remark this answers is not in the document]
dangershony: hmmm this raises another question when a legacy sends us headers we need to handle this (send them back headers) this will allow X to sync headers and blocks from # nodes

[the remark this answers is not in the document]
MithrilMan: when someone sends us legacy headers, `ProcessLegacyHeadersAsync` is executed and I think I should change
`if (!IsPeerPhEnabled(peer) && IsPeerWhitelisted(peer))`
to
`if (!IsPeerPhEnabled(peer))`

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,107 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        /// Processes the legacy GetHeaders message.
+        /// Only whitelisted legacy peers are allowed to handle this message.
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <param name="headers">The headers.</param>
+        /// <returns></returns>
+        protected Task ProcessLegacyHeadersAsync(INetworkPeer peer, List<BlockHeader> headers)
+        {
+            // Only legacy peers are allowed to handle this message, or ph enabled peers before PH activation.
+            if (!IsPeerPhEnabled(peer) || !IsProvenHeaderActivated())
```
dangershony: this. `this.IsPeerPhEnabled()` for both methods

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,107 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+                        BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                        HashStop = null
+                    };
+                }
+                // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
+                else if (IsPeerWhitelisted(peer) || this.IsProvenHeaderActivated())
+                    return base.BuildGetHeadersPayload();
+                // If the peer doesn't support PH and isn't whitelisted, return null (stop synch attempt with legacy StratisX nodes).
+                else
+                    return null;
```
dangershony: I still think we should return a normal GetHeaders if PH is not enabled and peer is not upgraded.

So we need to return `GetProvenHeadersPayload` if PH enabled and protocol >= PH protocol otherwise normal headers.

dangershony: The whitelistedness is for processing the headers not for getheaders so the other method is correct (i.e processes legacy) but getheaders is no big deal.
We can have a vid chant perhaps to talk about this when you have time.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,107 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+                if (this.IsPeerPhEnabled(peer))
+                {
+                    return new GetProvenHeadersPayload()
+                    {
+                        BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                        HashStop = null
+                    };
+                }
+                // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
+                else if (IsPeerWhitelisted(peer) || this.IsProvenHeaderActivated())
```
bokobza: `this.IsProvenHeaderActivated()` will always be true here.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: ok but AFAIU BuildGetHeadersPayload is used to create the GetHeaders payload sent by our node to the other, when we want to sync to a peer, so we are requiring nodes in the legacy way and this is what we don't want to: we want to synch in a legacy way only to whitelisted nodes or before ph activation.

Actually we serve GetHeaders requests (so HeadersPayload) to any node, so any node can ask us headers in a legacy way and we serve them, but not the opposite (we don't want to request legacy node to peers we don't trust or that don't serve PH headers)

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,107 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        /// Processes the legacy GetHeaders message.
+        /// Only whitelisted legacy peers are allowed to handle this message.
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <param name="headers">The headers.</param>
+        /// <returns></returns>
+        protected Task ProcessLegacyHeadersAsync(INetworkPeer peer, List<BlockHeader> headers)
+        {
+            // Only legacy peers are allowed to handle this message, or ph enabled peers before PH activation.
+            if (!this.IsPeerPhEnabled(peer) || !this.IsProvenHeaderActivated())
```
dangershony: Ah ok here is where you need to check whitelisted `IsPeerWhitelisted()`

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: @noescape00 what's your thought about this

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: yep!
```suggestion
                else if (IsPeerWhitelisted(peer))
```

noescape00:
Following tests are failing: 

StratisConsensusRulesRegistrationTest
RunRule_ProvenHeadersActive_And_ValidProvenHeader_NoErrorsAreThrown

@MithrilMan please run all the tests when submitting a PR

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -56,6 +66,15 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
                 case GetProvenHeadersPayload getHeaders:
                     await this.ProcessGetHeadersAsync(peer, getHeaders).ConfigureAwait(false);
                     break;
+
+                case HeadersPayload headers:
+                    await this.ProcessLegacyHeadersAsync(peer, headers.Headers).ConfigureAwait(false);
```
noescape00: Wouldn't call them legacy since on BTW network they are not outdated.
I'd probably rename it to process headers and for PH rename to ProcessProvenHeaders

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        }
+
+        /// <summary>
+        /// Determines whether the specified peer supports Proven Headers and PH has been activated.
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <returns>
+        ///   <c>true</c> if is peer is PH enabled; otherwise, <c>false</c>.
+        /// </returns>
+        private bool IsPeerPhEnabled(INetworkPeer peer)
```
noescape00: PH

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <returns>
+        ///   <c>true</c> if the specified peer is Whitelisted; otherwise, <c>false</c>.
+        /// </returns>
+        private bool IsPeerWhitelisted(INetworkPeer peer)
+        {
+            return peer.Behavior<IConnectionManagerBehavior>()?.Whitelisted == true;
+        }
+
```
noescape00: redundant line

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+
+        private bool IsProvenHeaderActivated()
+        {
+            if (this.network.Consensus.Options is PosConsensusOptions options)
+            {
+                long currentHeight = this.chainState.ConsensusTip.Height;
+                return (options.ProvenHeadersActivationHeight > 0) && (currentHeight >= options.ProvenHeadersActivationHeight);
```
noescape00: no need to check against 0. PH are supposed to be enabled on all PoS networks

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+        /// <returns>
+        ///   <c>true</c> if the specified peer is Whitelisted; otherwise, <c>false</c>.
+        /// </returns>
+        private bool IsPeerWhitelisted(INetworkPeer peer)
+        {
+            return peer.Behavior<IConnectionManagerBehavior>()?.Whitelisted == true;
+        }
+
+
+        private bool IsProvenHeaderActivated()
```
noescape00: AreProvenHeadersActivated

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+            {
+                long currentHeight = this.chainState.ConsensusTip.Height;
+                return (options.ProvenHeadersActivationHeight > 0) && (currentHeight >= options.ProvenHeadersActivationHeight);
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Builds the GetHeadersPayload.
```
noescape00:
```
<see cref="GetHeadersPayload"/>
```

same for other class references

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+                if (this.IsPeerPhEnabled(peer))
+                {
+                    return new GetProvenHeadersPayload()
+                    {
+                        BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                        HashStop = null
+                    };
+                }
+                // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
+                else if (IsPeerWhitelisted(peer))
```
noescape00: this.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -92,16 +111,108 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+            }
+        }
+
+        /// <summary>
+        /// Processes the legacy GetHeaders message.
+        /// Only whitelisted legacy peers are allowed to handle this message.
+        /// </summary>
+        /// <param name="peer">The peer.</param>
+        /// <param name="headers">The headers.</param>
+        /// <returns></returns>
```
noescape00: empty returns

noescape00:
@MithrilMan I've checked master and they don't fail there. 
If they were failing and then fixed on master then please merged the latest master in

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: In pascal cases, acronyms usually are presented by only having the the first letter capitalized, even if there are special rules for acronyms of length 2 that can be both capitalized. Which standard did we have?
Anyway for code consistency I rename it to extended ProvenHeaders

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: @noescape00 this is the most important comment to address in this PR.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: No this is only about triggering the sync, the handling of headers is done somewhere else (bellow in the method `ProcessLegacyHeadersAsync`) there we should ignore regular headers if we are PH activated.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: @noescape00 the main problem with renaming this stuff is that `ProcessGetHeadersAsync` is a base class defined in ConsensusManager and renaming it to ProcessProvenHeaderGetHeadersAsync is wrong.
This is why I needed a way to refer to the old way to handle headers, and legacy seemed appropriate to me, but if you have other suggestions let me know.
Actually I can't refactor the way you want.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: `GetHeadersPayload` is handled only in ConsensusManagerBehavior.cs
https://github.com/stratisproject/StratisBitcoinFullNode/blob/e7fd207adf857a3f2ecc85261ba3ca3fe87c2846/src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs#L135

This mean that whenever we receive a `getheaders` message from a peer, we handle that message only in that method (ProvenHeadersConsensusManagerBehavior inherit from ConsensusManagerBehavior so it handle the same way) and what we actually do is serving the request, so we allow any peer to ask us legacy headers. Isn't this right?
What we prevent, is asking legacy headers from peers that aren't whitelisted (after ph is activated)

Basically you started saying:
>I still think we should return a normal GetHeaders if PH is not enabled and peer is not upgraded.

And we are already doing it, because when a peer ask us headers using `getheaders` message, we handle it normally as before, see
https://github.com/stratisproject/StratisBitcoinFullNode/blob/e7fd207adf857a3f2ecc85261ba3ca3fe87c2846/src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs#L135-L137

note that `ProcessGetHeadersAsync` method is called both when legacy `getheaders` or new  `getprovhdr` is received, what change is message we send to that peer as a reply to its request:
https://github.com/stratisproject/StratisBitcoinFullNode/blob/e7fd207adf857a3f2ecc85261ba3ca3fe87c2846/src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs#L180

Now, taking a closer look to `ConstructHeadersPayload` method now i see another problem: that method actually always return a proven header payload because when overridden in `ConstructHeadersPayload` it doesn't check if the request was a legacy one or a PH one.
So this is what should be fixed (this wasn't part of my changes, I'm just noticing this now).

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
MithrilMan: Fixed this problem.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: Fixed ConstructHeadersPayload problem.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -85,23 +102,113 @@ protected override Payload ConstructHeadersPayload(BlockLocator locator, uint256
...
+                    {
+                        BlockLocator = (this.ExpectedPeerTip ?? this.consensusManager.Tip).GetLocator(),
+                        HashStop = null
+                    };
+                }
+                // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
+                else if (this.IsPeerWhitelisted(peer))
+                    return base.BuildGetHeadersPayload();
+                // If the peer doesn't support PH and isn't whitelisted, return null (stop synch attempt with legacy StratisX nodes).
+                else
```
dangershony: Its a bit confusing about what this `if` this `else` belongs to so maybe add braces?

bokobza:
merge with master, let's build.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: oh if it was for me I'd add always braces even on single-line blocks!
Braces added, let's merge

PR #2563, opened 2018-10-19
[Mempool] CheckSigOps using incorrect constant (#2470)

### LINKED ISSUE #2470: [Mempool] CheckSigOps using incorrect constant
Used in https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs#L768 ->

`if (context.SigOpsCost > this.ConsensusOptions.MaxBlockSigopsCost)`

Used in core https://github.com/bitcoin/bitcoin/blob/master/src/validation.cpp#L724 -> 

`if (nSigOpsCost > MAX_STANDARD_TX_SIGOPS_COST)`

These are two distinct constants with their own values (the max for blocks is 80 000 and the max for a single transaction is 80 000 / 5 = 16 000).

MithrilMan:
Added maxStandardTxSigopsCost ConsensusOptions property to fix #2470

### DIFF
--- a/src/NBitcoin/ConsensusOptions.cs
+++ b/src/NBitcoin/ConsensusOptions.cs
@@ -7,7 +7,7 @@
     public class ConsensusOptions
     {
         /// <summary>
-        /// Flag used to detect SegWit transactions.  
+        /// Flag used to detect SegWit transactions.
         /// </summary>
         public const int SerializeTransactionNoWitness = 0x40000000;
 
@@ -39,6 +39,9 @@
         /// <summary>The maximum allowed number of signature check operations in a block (network rule).</summary>
         public int MaxBlockSigopsCost { get; set; }
 
+        /// <summary>The maximum number of sigops we're willing to relay/mine in a single tx.</summary>
+        public int MaxStandardTxSigopsCost { get; set; }
+
         /// <summary>
         /// Initializes the default values. Currently only used for initialising Bitcoin networks and testing.
         /// </summary>
@@ -52,6 +55,7 @@
             this.MaxStandardTxWeight = 400000;
             this.MaxBlockBaseSize = 1000000;
             this.MaxBlockSigopsCost = 80000;
+            this.MaxStandardTxSigopsCost = this.MaxBlockSigopsCost / 5;
         }
 
         /// <summary>
@@ -64,7 +68,8 @@
             int witnessScaleFactor,
             int maxStandardVersion,
             int maxStandardTxWeight,
-            int maxBlockSigopsCost)
+            int maxBlockSigopsCost,
+            int maxStandardTxSigopsCost)
         {
             this.MaxBlockBaseSize = maxBlockBaseSize;
             this.MaxBlockWeight = maxBlockWeight;
@@ -73,6 +78,7 @@
             this.MaxStandardVersion = maxStandardVersion;
             this.MaxStandardTxWeight = maxStandardTxWeight;
             this.MaxBlockSigopsCost = maxBlockSigopsCost;
+            this.MaxStandardTxSigopsCost = maxStandardTxSigopsCost;
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


bokobza:
In fact, hold up.
@MithrilMan Are you sure MaxBlockSigopsCost is not used instead of MaxStandardTxSigopsCost in other part of the code?

MithrilMan:
check on bitcoin code
https://github.com/bitcoin/bitcoin/search?utf8=%E2%9C%93&q=MAX_STANDARD_TX_SIGOPS_COST&type=

evidences that it's used only there. Not sure if this is a valid proof

bokobza:
Look at the stratisX code as well.

src/NBitcoin/ConsensusOptions.cs
```diff
@@ -52,6 +55,7 @@ public ConsensusOptions()
             this.MaxStandardTxWeight = 400000;
             this.MaxBlockBaseSize = 1000000;
             this.MaxBlockSigopsCost = 80000;
+            this.MaxStandardTxSigopsCost = this.MaxBlockSigopsCost / 5;
```
dangershony: This is sort of a default value?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
```diff
@@ -765,7 +765,7 @@ private void CheckSigOps(MempoolValidationContext context)
             // itself can contain sigops MAX_STANDARD_TX_SIGOPS is less than
             // MAX_BLOCK_SIGOPS; we still consider this an invalid rather than
             // merely non-standard transaction.
-            if (context.SigOpsCost > this.ConsensusOptions.MaxBlockSigopsCost)
+            if (context.SigOpsCost > this.ConsensusOptions.MaxStandardTxSigopsCost)
```
dangershony: @zeptin found this bug so lets let him have a look as well, I think also @majikandy had issues here.

dangershony:
>check on bitcoin code
Yes ot be sure lets check stratisX as well

src/NBitcoin/ConsensusOptions.cs
MithrilMan: on bitcoin yes, it is
https://github.com/bitcoin/bitcoin/blob/78dae8caccd82cfbfd76557f1fb7d7557c7b5edb/src/policy/policy.h#L30

MithrilMan: `static const unsigned int MAX_STANDARD_TX_SIGOPS_COST = MAX_BLOCK_SIGOPS_COST/5;`

bokobza: same on StratisX @dangershony

majikandy:
yeah so your /5 makes it similar to the bitcoin code but that won't solve the bug that @sondreb and I both experienced (separately).  Where after staking on a new network you have lots of inputs to a largish value transactions (eg staking for a 1000 blocks) and then sending 1mil coins,  it has loads of inputs that when you debug show as ancestors....

yet on that mempool check - the sigops is 2...   but at mining time, the sig ops is 86,000 ish, way over the 20,000 stratis limit and so can't be mined....

futhermore it screws up the wallet a bit as it remains as unconfirmed whilst stuck in the memool, and deleting the mempool and resyncing (via api) doesn't even solve that

MithrilMan:
I'd consider this separated issue, being this PoS and not PoW so bitcoin related
We could leave issue open to another PR

majikandy:
Yes I agree it is a different issue - but it is covered on the issue this is linked to so either create a new more specific one from that or keep that one open as covering both the bit this fixes and awaiting another PR

majikandy:
PoS might not be related.....  (or perhaps it is, good point)

src/Stratis.Bitcoin.Features.MemoryPool/MempoolValidator.cs
zeptin: I'm pretty sure this `if` is now correct, at least as far as stratisX is concerned. I agree that Andy's issue is probably not resolved by this PR and still needs to be looked at.

PR #2569, opened 2018-10-22
[PoA] Integration tests

noescape00:


noescape00:
do not review before github shows that the last commit is 1cc34f7
it doesn't display all the commits and the diff is wrong

codingupastorm:
Just looked at briefly - gives me something to work off, great. Thanks @noescape00

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoAMiner.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            this.FastMiningEnabled = false;
+            this.cancellationSource = new CancellationTokenSource();
+        }
+
+        protected override async Task WaitBeforeCanMineAsync(int delayMs, CancellationToken cancellation = default(CancellationToken))
+        {
+            if (this.FastMiningEnabled)
+            {
+                TimeSpan span = (TimeSpan)this.timeProvider.GetMemberValue("adjustedTimeOffset");
```
dangershony: This is really ugly, instead I would suggest to cast to a new DTP type with those methods publicly available.
but its test to no  big deal

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoANetwork.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Features.PoA.IntegrationTests.Tools
+{
+    public class TestPoANetwork : PoANetwork
```
dangershony: Come to think about it instead of a PoaNetwork you should have a PoaConsensusOptions that holds additional POA network stuff, the federated keys dont belong in network rather in consensus.
WDYT

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -203,6 +176,60 @@ private async Task CreateBlocksAsync()
...
+
+            ChainedHeader chainedHeader = await this.consensusManager.BlockMinedAsync(blockTemplate.Block).ConfigureAwait(false);
+
+            if (chainedHeader == null)
+            {
+                // Block wasn't accepted because we already connected block from the network.
+                return null;
+            }
+
+            ValidationContext result = this.integrityValidator.VerifyBlockIntegrity(chainedHeader, blockTemplate.Block);
```
dangershony: An integrity check here? is this correct?

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/FullNodePoATestBuilderExtension.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            fullNodeBuilder.ConfigureFeature(features =>
+            {
+                foreach (IFeatureRegistration feature in features.FeatureRegistrations)
+                {
+                    feature.FeatureServices(services =>
+                    {
+                        ServiceDescriptor defaultProivider = services.FirstOrDefault(x => x.ServiceType == typeof(PoAMiner));
+
+                        services.Remove(defaultProivider);
```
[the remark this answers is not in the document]
noescape00: Not sure if needed. It will require even more code and will be harder to read- https://stackoverflow.com/questions/43590769/replace-service-registration-in-asp-net-core-built-in-di-container

also we use remove and add in other places

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/PoANodeBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public CoreNode CreatePoANode(PoANetwork network)
+        {
+            return this.CreateNode(new PoANodeRunner(this.GetNextDataFolderName(), network, this.TimeProvider), "poa.conf");
+        }
+
+        public CoreNode CreatePoANode(PoANetwork network, Key key)
+        {
+            string dataFolder = this.GetNextDataFolderName();
+            CoreNode node = this.CreateNode(new PoANodeRunner(dataFolder, network, this.TimeProvider), "poa.conf");
```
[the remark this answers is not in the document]
noescape00: yeah. Inherited networks will override the name.

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoAMiner.cs
noescape00: I'm not sure if we should change production code in favor of tests. 
Also we've agreed long time ago to use reflection over modifying production code.

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoANetwork.cs
noescape00: Fed keys are part of the network because we can have 2 networks which have exactly same consensus logic but different parameters of those that are network specific. 
So imho it's network specific.

noescape00: But we can move it to consensus options if needed because consensus option are in the network anyway. So I don't really mind

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Yeah. Just to make sure we don't violate integrity. 
It's more like a sanity check that I've added after noticing that I'm not updating merkle tree root but all the validators pass

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/FullNodePoATestBuilderExtension.cs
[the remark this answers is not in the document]
dangershony: https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.Features.SmartContracts/SmartContractFeature.cs#L122

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
noescape00: Also if we remove it it won't harm. It's just and additional sanity check which we can afford because integrity rules are pretty cheap

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoAMiner.cs
dangershony: No the new DTP can be a test class that you inject instead of the normal provider

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoANetwork.cs
dangershony: yes agree its per network but the options are where you can use override for that purpose exactly (so you dont need to inherit network) please move to options if you dont mind. (you can do it in a separate PR)

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoAMiner.cs
noescape00: Ok, fixed. 
But I had to change adjustedTimeOffset to be protected, which is not good, but ok.

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/TestPoANetwork.cs
noescape00: Ok, sure, will do

src/Stratis.Bitcoin.Features.PoA.IntegrationTests/Tools/EditableTimeProvider.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Features.PoA.IntegrationTests.Tools
+{
+    public class EditableTimeProvider : DateTimeProvider
```
dangershony: Nice name

PR #2584, opened 2018-10-24
[SC] Add PersistentState Get and Set for raw byte arrays

rowandh:


src/Stratis.SmartContracts/IPersistentState.cs
```diff
@@ -12,6 +12,11 @@ public interface IPersistentState
         /// </summary>
         bool IsContract(Address address);
 
+        /// <summary>
```
codingupastorm: Don't we want APIs for bytes as keys for every primitive type?

Issue #2268, opened 2018-09-25
[SC] Finalise gas costs


codingupastorm:
A couple issues with gas costs currently:

- For the gas price that we're using in tests (1), the BaseCost is extremely low. 100x less than the fee just to be included in a block.
- There are validaton / contract not existing cases where no fee is paid (even though we generate a condensing transaction)

So we need to specify the costs of certain scenarios and then test for these. 

Also take into account if we go with PoA then the funds are going to the A. Fees can be really cheap, but they do still need to not enable spamming of the network.

codingupastorm:
- Test

codingupastorm:
Test

codingupastorm:
Also need to enforce the min gas limit at the mempool validator

PR #2586, opened 2018-10-24
[SC] Amended costs for Call, Create, Transfer

codingupastorm:
Second last part of #2268


rowandh:
In `InternalExecutor.Transfer` we also check that the supplied gas is above `BaseCost`. This is now 12x higher than the required gas if the transfer is P2PKH only.

PR #2594, opened 2018-10-24
PoA - added periodic log and moved fed keys and target spacing to PoAConsensusOptions

noescape00:
untitled

### DIFF
--- a/src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
@@ -52,16 +52,6 @@ namespace Stratis.Bitcoin.Features.PoA
 
             var consensusFactory = new PoAConsensusFactory();
 
-            // Configure federation public keys.
-            // Keep in mind that order in which keys are added to this list is important
-            // and should be the same for all nodes operating on this network.
-            var federationPublicKeys = new List<PubKey>()
-            {
-                new PubKey("03e6f19ea3dc6c145d98a0e0838af952755798e5bc3950bbca4f9485aa23873d7f"),
-                new PubKey("02ddebcf18207072bdd172a25f85f2ea12e2de1d9d794f136722634aad08400fcb"),
-                new PubKey("02067b38d777690aaaf23a5b371a819e6ddc6d2aae734b0199fe59df28dc056dd7")
-            };
-
             // Create the genesis block.
             this.GenesisTime = 1513622125;
             this.GenesisNonce = 1560058197;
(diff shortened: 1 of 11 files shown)


src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
```diff
@@ -64,7 +55,7 @@ public PoANetwork()
             // Configure federation public keys.
             // Keep in mind that order in which keys are added to this list is important
             // and should be the same for all nodes operating on this network.
-            this.FederationPublicKeys = new List<PubKey>()
+            var federationPublicKeys = new List<PubKey>()
```
bokobza: Can you bring this closer to the new PoAConsensusOptions()?
You can even bring it inside the instantiation and not use a variable.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
```diff
@@ -168,7 +161,7 @@ public PoANetwork()
             Assert(this.Consensus.HashGenesisBlock == uint256.Parse("0x0621b88fb7a99c985d695be42e606cb913259bace2babe92970547fa033e4076"));
             Assert(this.Genesis.Header.HashMerkleRoot == uint256.Parse("0x9928b372fd9e4cf62a31638607344c03c48731ba06d24576342db9c8591e1432"));
 
-            if ((this.FederationPublicKeys == null) || (this.FederationPublicKeys.Count == 0))
+            if ((this.ConsensusOptions.FederationPublicKeys == null) || (this.ConsensusOptions.FederationPublicKeys.Count == 0))
```
bokobza: You can bring this almost all the way to the top of the class. Maybe just after `var consensusOptions = new PoAConsensusOptions`

src/Stratis.SmartContracts.Networks/SmartContractsPoARegTest.cs
```diff
@@ -38,7 +38,9 @@ public SmartContractsPoARegTest()
                 maxStandardVersion: 2,
                 maxStandardTxWeight: 100_000,
                 maxBlockSigopsCost: 20_000,
-                maxStandardTxSigopsCost: 20_000 / 5
+                maxStandardTxSigopsCost: 20_000 / 5,
+                federationPublicKeys: this.ConsensusOptions.FederationPublicKeys,
```
bokobza: You'd expect the keys to be inside this class rather than inherited from the base network class.
Which begs the question: do we need an instantiable PoaNetwork class at all?
am I right in thinking that the only difference in the number of properties (not their values) is the consensusOptions?

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
noescape00: >Can you bring this closer to the new PoAConsensusOptions()?

done

>You can even bring it inside the instantiation and not use a variable.

static code analyzer use are using doesn't like it

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
noescape00: not sure. 
network initialize has parts- first is setup and second is sanity checks. 
I think it's a good idea to avoid mixing them

src/Stratis.SmartContracts.Networks/SmartContractsPoARegTest.cs
noescape00: >You'd expect the keys to be inside this class rather than inherited from the base network class.

it's a consequence of moving fed keys to network options. 
basically we need to copy all values from inherited network except for fed keys. 

it's the only way since those are protected- so we can only set whole new instance of consensus options. 

That's only of the reasons why I think it's better to have those in network itself.

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
bokobza: We don't really need the sanity checks, but it's a shame to go through the whole method only to fail because some things weren't set up properly.
Even the genesis checks are a bit useless in my opinion.

src/Stratis.SmartContracts.Networks/SmartContractsPoARegTest.cs
bokobza: In Network rather than in ConsensusOptions?

src/Stratis.Bitcoin.Features.PoA/PoANetwork.cs
bokobza: ok fine

src/Stratis.SmartContracts.Networks/SmartContractsPoARegTest.cs
bokobza: ugh, but ok

PR #2596, opened 2018-10-24
Disabled staking when in IBD and add interceptors for tests

### LINKED ISSUE #2581: Check that POS is not mining at IBD
Can PoS minting generate a block while in IBD?

we have this pro protect against this kind of things:
```// Prevent mining if not fully synced.
if (this.initialBlockDownloadState.IsInitialBlockDownload() ||
    this.consensusManager.Tip != this.chain.Tip)
{
    this.logger.LogTrace("Waiting for synchronization before mining can be started.");

    await Task.Delay(TimeSpan.FromMilliseconds(this.minerSleep), cancellationToken).ConfigureAwait(false);
    continue;
}
```

but now it's useless since this.consensusManager.Tip will always be the same as this.chain.Tip.

So if we are not in IBD and mined >500 blocks behind header's tip we would announce this block and violate max reorg rule for other nodes. And they will ban us.

dangershony:
Fix #2581

### DIFF
--- a/src/Stratis.Bitcoin.Features.Miner.Tests/PowMiningTest.cs
+++ b/src/Stratis.Bitcoin.Features.Miner.Tests/PowMiningTest.cs
@@ -9,6 +9,7 @@ using Stratis.Bitcoin.Configuration;
 using Stratis.Bitcoin.Consensus;
 using Stratis.Bitcoin.Features.MemoryPool;
 using Stratis.Bitcoin.Features.MemoryPool.Interfaces;
+using Stratis.Bitcoin.Interfaces;
 using Stratis.Bitcoin.Mining;
 using Stratis.Bitcoin.Tests.Common;
 using Stratis.Bitcoin.Tests.Common.Logging;
@@ -27,6 +28,7 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
         private ConcurrentChain chain;
         private readonly Mock<IConsensusManager> consensusManager;
         private readonly Mock<IConsensusRuleEngine> consensusRules;
+        private readonly Mock<IInitialBlockDownloadState> initialBlockDownloadState;
         private readonly ConsensusOptions initialNetworkOptions;
         private readonly PowMiningTestFixture fixture;
         private readonly Mock<ITxMempool> mempool;
@@ -59,6 +61,9 @@ namespace Stratis.Bitcoin.Features.Miner.Tests
 
             this.nodeLifetime = new Mock<INodeLifetime>();
             this.nodeLifetime.Setup(n => n.ApplicationStopping).Returns(new CancellationToken()).Verifiable();
+
+            this.initialBlockDownloadState = new Mock<IInitialBlockDownloadState>();
+            this.initialBlockDownloadState.Setup(s => s.IsInitialBlockDownload()).Returns(false);
 
             this.mempoolLock = new MempoolSchedulerLock();
         }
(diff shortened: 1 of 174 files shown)


src/Stratis.Bitcoin.IntegrationTests.Common/BlockObserverFeature.cs
```diff
@@ -69,4 +91,29 @@ protected override void OnNextCore(ChainedHeaderBlock chainedHeaderBlock)
                 this.interceptorExecuted = this.interceptor(chainedHeaderBlock);
         }
     }
+
+    /// <summary>
+    /// Executes an action when a block has been disconnected at a certain height.
```
fassadlr: ..has been connected...

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -121,7 +123,18 @@ public CoreNode NoValidation()
         /// <returns>This node.</returns>
         public CoreNode BlockDisconnectInterceptor(Func<ChainedHeaderBlock, bool> interceptor)
         {
-            this.builderInterceptor = interceptor;
+            this.builderDisconnectInterceptor = interceptor;
+            return this;
+        }
+
+        /// <summary>
+        /// Executes a function when a block has disconnected.
```
fassadlr: ...when a block has connected

src/Stratis.Bitcoin.IntegrationTests.Common/EnvironmentMockUpHelpers/CoreNode.cs
```diff
@@ -121,7 +123,18 @@ public CoreNode NoValidation()
...
+        /// </summary>
+        /// <param name="interceptor">A function that is called when a block disconnects, it will return true if it executed.</param>
```
fassadlr: ... is called when a block connects

dangershony:
>I dont see the new interceptor being used in any integration tests?  

Yeah tried to simulate the behaviour but its very hard, I needed the interceptor but ended up not using it (yet)

src/Stratis.Bitcoin.Features.Miner/PowMining.cs
```diff
@@ -175,7 +180,7 @@ public List<uint256> GenerateBlocks(ReserveScript reserveScript, ulong amountOfB
                     continue;
 
                 if (!this.ValidateAndConnectBlock(context))
-                    break;
+                    continue;
```
fassadlr: So just to confirm if anything goes wrong with block connection we just continue mining yes?

dangershony: If a consensus rule fails it will throw, this flag is only true when the pre hash is not as the presented one which is ok.We are outside of the lock when we create the next block so in the mean time a new block could take the lead before we take the lock and make that check.

src/Stratis.Bitcoin.Features.Miner.Tests/PowMiningTest.cs
```diff
@@ -205,23 +210,6 @@ public void GenerateBlocks_SingleBlock_ReturnsGeneratedBlock()
             Assert.Equal(callbackBlock.GetHash(), blockHashes[0]);
         }
 
-        [Fact]
-        public void GenerateBlocks_SingleBlock_InvalidPreviousTip_ContinuesExecution_ReturnsGeneratedBlock()
```
fassadlr: This test isnt necessary as it is covered by`GenerateBlocks_SingleBlock_MaxTriesReached_StopsGeneratingBlocks_ReturnsEmptyList`

Issue #2401, opened 2018-10-06
[SC] The smart contract consensus rules can be optimized.


codingupastorm:
There are checks about the format of smart contract transactions, whether the sender can be retrieved, whether P2PKHs are to contracts etc, all in different rules. 

This means that the Scripts are deserialized and checked several times. 

This could really all be in one rule, or at least broken down in a different manner so as to minimise performing the same tasks many times. 

Will take some investigation but much progress to be made here.

codingupastorm:
Includes inside the `SmartContractMempoolValidator`

PR #2599, opened 2018-10-25
[SC] Min gas price in mempool validator

### LINKED ISSUE #2268: [SC] Finalise gas costs

codingupastorm:
Closes #2268

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using CSharpFunctionalExtensions;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Configuration;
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin.Features.SmartContracts/SmartContractMempoolValidator.cs
```diff
@@ -75,7 +81,22 @@ public override void CheckFee(MempoolValidationContext context)
...
+        private void CheckMinGasLimit(MempoolValidationContext context)
+        {
+            Transaction transaction = context.Transaction;
+
+            if (!transaction.IsSmartContractExecTransaction())
+                return;
+
+            // We know it has passed SmartContractFormatRule so we can deserialize it easily.
+            TxOut scTxOut = transaction.TryGetSmartContractTxOut();
+            Result<ContractTxData> callDataDeserializationResult = this.callDataSerializer.Deserialize(scTxOut.ScriptPubKey.ToBytes());
```
rowandh: Is there a nice way we can avoid performing the deserialization twice?

codingupastorm: We actually serialize and deserialize way too many times, it would be nice if we could pass a deserialized objects around - currently tracking with #2401

Issue #2579, opened 2018-10-23
SBFN node is failing to keep up with StratisX


maciejzaleski:
**Setup**
StratisX and SBFN are occasionally able to sync to the same block height.
Most of the time SBFN will be about 10 blocks behind StratisX and when SBFN falls 10-20 blocks behind StratisX SBFN will synchronise to the latest block height.

**Steps to reproduce**
 - Start SBFN node which is connected only to StratisX node
 - Start StratisX node which is connected to the network
 - Observe slow block height update (SBFN will be usually 10-20 blocks behind StratisX)

maciejzaleski:
Logs from two nodes described in this issue:
[2570.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2505754/2570.zip)

ArcSin2000X:
I have now tested this scenario on mainnet and I can see SBFN node is constantly running 6 blocks behind StratisX

zeptin:
I think I've found a possible reason for this, just running some additional analysis to be sure. It appears X initially announces blocks via unsolicited `inv` payloads, which we do not handle anywhere in the behaviours.

X generates block (done here via RPC `generate`, not sure if that makes a difference to the announcement logic):

```
CheckWork() : new proof-of-work block found  
  proof hash: 35a63154068506d2b52c7d34f2c22c17b1bddf76257c6a099b8db28e5b5f3df8  
target: 7fffff0000000000000000000000000000000000000000000000000000000000
CBlock(hash=d97d56067438b18588d48e97d5cde6a2a11bc1e2bee37a28ff7cc4ab74c81c1d, ver=7, hashPrevBlock=56e3c1b83edd9828a066c4ce73d4b8235f3bc635202419eb234b74cec9fb9c99, hashMerkleRoot=10586fad010d3122ceb9fd23f04b6b223e308e780f919419c5f27d08b1341122, nTime=1540432088, nBits=207fffff, nNonce=0, vtx=1, vchBlockSig=)
  Coinbase(hash=10586fad010d3122ceb9fd23f04b6b223e308e780f919419c5f27d08b1341122, nTime=1540432088, ver=1, vin.size=1, vout.size=1, nLockTime=0)
    CTxIn(COutPoint(0000000000, 4294967295), coinbase 570101)
[...]

PR #2601, opened 2018-10-25
[Mempool] Don't send getdata unnecessarily in response to inv

zeptin:
Just a small fix, this is related to issue #2579 but does not fix it.

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -318,7 +318,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
                 send.Inventory.Add(new InventoryVector(peer.AddSupportedOptions(InventoryType.MSG_TX), inv.Hash));
             }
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -318,7 +318,7 @@ private async Task ProcessInvAsync(INetworkPeer peer, InvPayload invPayload)
                 send.Inventory.Add(new InventoryVector(peer.AddSupportedOptions(InventoryType.MSG_TX), inv.Hash));
             }
 
-            if (peer.IsConnected)
+            if (peer.IsConnected && (send.Inventory.Count > 0))
```
noescape00: peer.IsConnected - it should probably be is handshaked instead

noescape00: or nothing, actually. since it looks like we will always be connected when we execute this method. 
and in case we are in the middle of disconnection this won't save us from the race condition

Issue #2595, opened 2018-10-24
Node is failing to synchronise after a reorg


maciejzaleski:
**Setup**
 - Two nodes are staking NodeDA and NodeDB
 - NodeAD is failing to sync consensus and blockstore.
 - Wallet hash is set to 0000000000000000000000000000000000000000000000000000000000000000

**Full Logs**
https://stratisplatform-my.sharepoint.com/:u:/p/maciej_zaleski/EQzF5PI43ZZArNrpWZZ7_JwBKixJYkSnoEEUFrwvY_KtYg?e=CasADm

```[2018-10-24 10:40:15.4594 36] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__76_0 ======Node stats====== 10/24/2018 10:40:15
Headers.Height:      2075     Headers.Hash:     f613bb91a34527471c7f8a696083c6c8c29282efc8b9d5915d584cf71d880c38
Consensus.Height:    2075     Consensus.Hash:   f613bb91a34527471c7f8a696083c6c8c29282efc8b9d5915d584cf71d880c38
BlockStore.Height:   2075     BlockStore.Hash:  f613bb91a34527471c7f8a696083c6c8c29282efc8b9d5915d584cf71d880c38
Wallet.Height:       2075     Wallet.Hash:      f613bb91a34527471c7f8a696083c6c8c29282efc8b9d5915d584cf71d880c38

======Connection====== agent StratisBitcoin:1.2.2
Peer:[::ffff:192.168.98.121]:36249,      connected:inbound,         height:2075,          agent:StratisBitcoin:1.2.2
Peer:[::ffff:192.168.98.190]:45329,      connected:inbound,         height:2066,          agent:StratisBitcoin:1.2.2
Peer:[::ffff:192.168.98.150]:44557,      connected:inbound,         height:2075,          agent:StratisBitcoin:1.2.2
Peer:[::ffff:192.168.98.193]:16178,      connected:outbound,        height:2075,          agent:/Stratis:2.0.0.5/
Peer:[::ffff:192.168.98.102]:42029,      connected:inbound,         height:2070,          agent:StratisBitcoin:1.2.2

======Consensus Manager======
Unconsumed blocks: -178 -- (-114 386 / 209 715 200 bytes). Cache is filled by: -0.05%

======Block Puller======
Blocks being downloaded: 0
Queueued downloads: 0
Average block size: 0.7 KB
[...]

PR #2603, opened 2018-10-25
Added logs

noescape00:


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -289,7 +289,17 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
...
+                bool added = this.pendingBlocksCache.TryAdd(chainedHeaderBlock.ChainedHeader.HashBlock, chainedHeaderBlock);
+
+                if (added)
+                {
+                    this.logger.LogTrace("Block '{0}' was added to pending.", chainedHeaderBlock.ChainedHeader);
+                }
+                else
+                {
+                    this.logger.LogCritical("Unable to add block '{0}' to pending!", chainedHeaderBlock.ChainedHeader);
+                    throw new Exception("Can't add block to pending!");
```
bokobza: What will happen to the node if this fails?

noescape00: node stops advancing

dangershony: Why do you suspect this will even happen?

noescape00: Because logs @maciejzaleski provided indications that https://github.com/stratisproject/StratisBitcoinFullNode/issues/2595 is caused by inability to get a block from CHT cache and BSQ. 
So need to investigate missing blocks and make sure that they were saved in BSQ correctly and then not deleted.  

Basically 2595 is about node trying to do a big reorg (100+ blocks) which leads to inability to get blocks for connection. 
This might be CM, CHT or BSQ issue.

dangershony: ok yes I am very much in agreement, by the way why did we use the number 100? and is it configurable?

noescape00: it's not configurable. 
value of 100 is pretty much random. 

for PoS networks reasonable value is between 10 and MaxReorg

Issue #2500, opened 2018-10-15
[PH] store stake info in memory for forward looking headers


dangershony:
This line is problematic in PH rule validation 
https://github.com/stratisproject/StratisBitcoinFullNode/pull/2381/files#diff-c0bed2cb85d3ca47e2fb7a8c5c5e12e8R220

*Problem:*
We ask for stake info (to get the stake modifier) from StakeStore but that will only calculate stake info up to current CT.

*The solution:*
At that point of validation we have all required info to calculate stake modifier for current PH so we calculate it and store it in the PH header (it will not get serialized to disk, or should it? see bellow)
After verifuing stake modifier and kernal is correct we will calculate the stake modifier for current block using `stakeValidator.ComputeStakeModifierV2()`

dangershony:
*One more option:*

Storing stake modifier in PH header AND passing it in payloads and storing it to disk.
- Benefit: this will reduce the complexity of the code 
- Downside: Increase the amount of data we store to disk (but only slightly)

I suggest we consider putting stake modifier in PH as a serialised citizen.

noescape00:
In order to validate PH we need to validate coisntake kernel which is inside of PH. This validation requires information from the stake chain, more specifically - stake modifier. 
Usually we get this from StakeStore but stake store only has this info for blocks that are already part of consensus chain. Now we need this info after consensus tip to validate proven headers. 

Therefore we need a way to access this information. 
I think we can implement it by adding a field to PH (it will not be serialized and saved to disk\sent to network\ affect hash)- this field will contain stake modifier. 

When we need to validate PH N we will go to PH N-1 and get modifier from there. 
After that we will be able to calculate modifier for N and then we should save it in N's modifier field.

MithrilMan:
To compute stakeModifier from previous header, I need that the previous header exposes that information.
Are we sure that if we validate a header after last checkpoint, the previous header is a ProvenHeader ?
Since we allow whitelisted stratisX peers to return us legacy headers, we can't ensure this is true, so the `StackModifier` property should be added to `PosBlockHeader` and not to `ProvenBlockHeader` and that property should be computed for both normal PoS headers and PH headers. Is this right? How we approach this?
@noescape00 @dangershony

noescape00:
>Are we sure that if we validate a header after last checkpoint, the previous header is a ProvenHeader ?

no. we might as well be syncing from whitelisted qt node- in that case we don't have proven header there

[...]

PR #2608, opened 2018-10-25
[PH] store stake info in memory for forward looking headers (#2500)

MithrilMan:
Initial PR related to #2500 , use this as base for discussion, it's not ready to be merged yet I suppose.


src/NBitcoin/ProvenBlockHeader.cs
```diff
@@ -93,6 +95,9 @@ public ProvenBlockHeader(PosBlock block)
             this.signature = block.BlockSignature;
             this.coinstake = block.Transactions[1];
             this.merkleProof = new MerkleBlock(block, new[] { this.coinstake.GetHash() }).PartialMerkleTree;
+
+            BlockStake blockStake = BlockStake.Load(block);
+            this.StakeModifierV2 = blockStake.StakeModifierV2;
```
dangershony: I dont think you can just do that unfortunately stake modifier needs to be created based on the previous SMV2

MithrilMan: this is old code, don't review this, we need to talk (new code is uncommitted because requires design decision)

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -217,16 +217,29 @@ private void CheckStakeKernelHash(PosRuleContext context, UnspentOutputs staking
...
             {
-                this.Logger.LogTrace("(-)[BAD_PREV_STAKE]");
-                ConsensusErrors.PrevStakeNull.Throw();
+                stakeModifierV2 = prevBlockStake.StakeModifierV2;
+            }
+            else
+            {
+                if (prevChainedHeader.Header is ProvenBlockHeader phHeader)
+                {
+                    stakeModifierV2 = phHeader.StakeModifierV2;
```
dangershony: maybe rename this to `preStakeModifierV2`

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -217,16 +217,29 @@ private void CheckStakeKernelHash(PosRuleContext context, UnspentOutputs staking
...
+                    stakeModifierV2 = null;
+                    this.Logger.LogTrace("(-)[BAD_PREV_STAKE]");
+                    ConsensusErrors.PrevStakeNull.Throw();
+                }
             }
 
             uint headerBits = chainedHeader.Header.Bits.ToCompact();
 
-            this.stakeValidator.CheckStakeKernelHash(context, headerBits, prevBlockStake, stakingCoins, prevOut, transactionTime);
+            this.stakeValidator.CheckStakeKernelHash(context, headerBits, prevBlockStake.StakeModifierV2, stakingCoins, prevOut, transactionTime);
```
dangershony: here you should use the variable `stakeModifierV2 ` right?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
dangershony: What I am missing is where to we set stake modifier on the PH header

src/NBitcoin/ProvenBlockHeader.cs
dangershony: So is this still valid?

src/NBitcoin/ProvenBlockHeader.cs
```diff
@@ -107,6 +113,8 @@ public override void ReadWrite(BitcoinStream stream)
 
             stream.ReadWrite(ref this.coinstake);
             this.CoinstakeSize = stream.ProcessedBytes - Size - this.MerkleProofSize - this.SignatureSize;
+
+            stream.ReadWrite(ref this.stakeModifierV2);
```
dangershony: I think we don't need this finally, we will not serialize stake modifier just set it.

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -48,19 +47,9 @@ public override void Initialize()
...
-        public override void Run(RuleContext context)
+        /// <inheritdoc />
+        protected override void ProcessRule(PosRuleContext context, ChainedHeader chainedHeader, ProvenBlockHeader header)
         {
-            Guard.NotNull(context.ValidationContext.ChainedHeaderToValidate, nameof(context.ValidationContext.ChainedHeaderToValidate));
-
-            ChainedHeader chainedHeader = context.ValidationContext.ChainedHeaderToValidate;
-            int height = chainedHeader.Height;
-
-            if (context.SkipValidation || !this.IsProvenHeaderActivated(height))
```
dangershony: Why delete this code?

MithrilMan: I moved that code, that was common to both 2 PH rules (the other was size rule) and moved to the ProvenHeaderRuleBase.cs
If you look this method is no more the `Run` but another that gets called by ProvenHeaderRuleBase.Run method

src/NBitcoin/ProvenBlockHeader.cs
MithrilMan: you need it to validate incoming headers: how can you know that when you start a node otherwise?

dangershony: Yes but we will calculate the stake modifier when validating

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
```diff
@@ -162,7 +162,7 @@ private void CheckAndComputeStake(RuleContext context)
                 // Compute stake modifier.
                 ChainedHeader prevChainedHeader = chainedHeader.Previous;
                 BlockStake blockStakePrev = prevChainedHeader == null ? null : this.stakeChain.Get(prevChainedHeader.HashBlock);
-                blockStake.StakeModifierV2 = this.stakeValidator.ComputeStakeModifierV2(prevChainedHeader, blockStakePrev, blockStake.IsProofOfWork() ? chainedHeader.HashBlock : blockStake.PrevoutStake.Hash);
+                blockStake.StakeModifierV2 = this.stakeValidator.ComputeStakeModifierV2(prevChainedHeader, blockStakePrev?.StakeModifierV2, blockStake.IsProofOfWork() ? chainedHeader.HashBlock : blockStake.PrevoutStake.Hash);
```
dangershony: Its probably not right to do this `blockStakePrev?.StakeModifierV2` we will fail and not know why.
At this stage we expect to have stake modifier and if not we need to throw.

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -214,19 +203,28 @@ private void CheckStakeKernelHash(PosRuleContext context, UnspentOutputs staking
...
-            BlockStake prevBlockStake = this.PosParent.StakeChain.Get(prevChainedHeader.HashBlock);
-            if (prevBlockStake == null)
+            if (chainedHeader.Height == this.ProvenHeadersActivationHeight)
             {
-                this.Logger.LogTrace("(-)[BAD_PREV_STAKE]");
-                ConsensusErrors.PrevStakeNull.Throw();
+                // If we are validating the first Proven Header, we don't have a StakeModifierV2 available and I should compute it but I can't.
+                // Proposed solution is to store the StakeModifierV2 value of the header at the height of ProvenHeaderActivation in a consensus variable
+                // called ProvenHeaderActivationStakeModifier and so I can return it
+                previousStakeModifier = uint256.Zero; //TODO to be changed after ProvenHeaderActivationStakeModifier has been created.
```
dangershony: Here you need to get Stakemodifier from store (and when we use checkpoints it will be form the last checkpoint, so perhaps throw an exception for now.

MithrilMan: yes we have to remove ProvenHeadersActivationHeight too (this will impact almost all PH code)

src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/PosCoinviewRule.cs
MithrilMan: throw here or null-check into ComputeStakeModifierV2 to see if prevStakeModifier is null?
previous code where throwing into ComputeStakeModifierV2 on the line `serializer.ReadWrite(blockStakePrev.StakeModifierV2);` so maybe I should put a Guard notnull there to have the same behavior

dangershony: yes put a guard there

PR #2611, opened 2018-10-26
Block announcement from legacy nodes

zeptin:
This is not complete, I need feedback from @noescape00 and @dangershony on how this should coexist with proven headers & whitelisting.

This is related to #2579 and does seem to improve the syncing in the limited integration tests I have tried so far. I am investigating further however, as there seems to be a possibly related transaction relay issue where X is either not sending or SBFN is not receiving a relayed transaction.


src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -474,6 +478,52 @@ protected virtual GetHeadersPayload BuildGetHeadersPayload()
...
+                {
+                    blockHashAnnounced = true;
+                    break;
+                }
+            }
+
+            if (blockHashAnnounced)
+            {
+                this.logger.LogTrace("Block was announced, sending getheaders.");
+                await ResyncAsync().ConfigureAwait(false);
```
dangershony: Is this the same way X handles inv? I am not so sure.
But the logic as I understand it, we receive an inv, if it has a block type we  retrigger sync of headers, right?

zeptin: I am not sure we need to preserve identical functionality, as we will in any case need to disregard `MSG_BLOCK` `inv`s from non-whitelisted nodes (this is the part I have not implemented yet).

My rationale for doing it this way is that an X node will send us an `inv` to announce a block. We then need to either disregard this `inv` or obtain the new header somehow. To avoid code duplication I opted to call `ResyncAsync` which constructs a `getheaders` message, which should achieve what we want.

`ResyncAsync` is perhaps not perfectly named. In the case where we are fully synced with the rest of the network except the newly announced block, we should get back a response with just the one header in it I expect, as our locator will indicate to the peer that we are only missing the latest block.

dangershony: this will cause a lot of traffic for every inv, perhaps limit the amount of allowed inv triggered resync to 1 per min per node?

zeptin: I am not sure rate limiting is the way to do this.

I have enumerated the following `inv` scenarios:

**Solicited:**
* If we solicited the `inv` in some way then rate limiting is the responsibility of the initiating code. In any case this PR primarily deals with the unsolicited `inv`s typically sent by X nodes for block announcement

**Unsolicited:**
* May receive MSG_TX inventory vector
* May receive mixture of vectors
* May receive payload with few vectors
* May receive payload with many vectors
* May receive MSG_BLOCK inventory vector

Examining the unsolicited scenarios:

*MSG_TX vector*

Handled by mempool behaviour. Out of scope of this PR, as the mempool has its own inventory cache logic.

*May receive mixture of vectors*
*May receive payload with few vectors*
*May receive payload with many vectors*

These all result in the same thing from the perspective of this PR. If they contain at least one `MSG_BLOCK` then the logic is triggered, else the `inv` is ignored by CMB. The performance impact of large `inv`s is therefore negligible for our purposes.

*MSG_BLOCK inventory vector*

This is therefore the only case of interest, the payload had at least one of these vectors in it.

This has a few possibilities:
* A block is being announced that we have a header for
* A block is being announced that we do not have a header for (this includes malicious peers that falsely advertise nonexistent blocks)

As a response to the `MSG_BLOCK` we can do one of the following, I think:

* Ignore it unconditionally
* Ignore it conditional upon us having the header already (this would involve a lookup of some sort), else request the header from the peer
* Request the header from the peer unconditionally

We cannot ignore it unconditionally as we would be back to where we were before the PR, with SBFN syncing poorly from X nodes.

Filtering based on whether we have the header already places a (small) processing burden on us for every `inv`. It does not protect us from falsely advertised block hashes, so we don't gain much.

Requesting the header unconditionally is what the PR currently does. This only has the overhead of building an sending a `getheaders` payload, i.e. constructing a block locator.
This is expected to be minimal as we mostly get blocks announced when we are fully synced (and a node on the network has staked a block).

Implicitly there is also the processing overhead of the `headers` response returned. The two main possibilities:
* We are in fact near the overall tip. We expect to get a very small number of headers back, which we want anyway as they represent blocks we probably want to download
* We are in IBD and get a large `headers` response back (we could also be on a fork, but that complicates the analysis)

The large `headers` response is not too much of a concern, as the CHT maintains a dictionary of header hashes so potential duplicates will be quickly filtered out.

So, should we ignore unsolicited `inv`s during IBD? I think the rest of the approach is sound.

dangershony: Ah I see there is definitely place to rate limit, once we send a getheaders then we start the processes of syncing headers so no need to send a get headers again for every inv that has blocks.


Also its probably a good idea to check that the hashes in the inv that is sent are not currently in our chain no?
So if we find any block hash we don't know about then we trigger a sync of headers

zeptin: For the edification of future readers, core seems to handle it in the same way:

```
            if (inv.type == MSG_BLOCK) {
                UpdateBlockAvailability(pfrom->GetId(), inv.hash);
                if (!fAlreadyHave && !fImporting && !fReindex && !mapBlocksInFlight.count(inv.hash)) {
                    // We used to request the full block here, but since headers-announcements are now the
                    // primary method of announcement on the network, and since, in the case that a node
                    // fell back to inv we probably have a reorg which we should get the headers for first,
                    // we now only provide a getheaders response here. When we receive the headers, we will
                    // then ask for the blocks we need.
                    connman->PushMessage(pfrom, msgMaker.Make(NetMsgType::GETHEADERS, chainActive.GetLocator(pindexBestHeader), inv.hash));
                    LogPrint(BCLog::NET, "getheaders (%d) %s to peer=%d\n", pindexBestHeader->nHeight, inv.hash.ToString(), pfrom->GetId());
                }
            }
```

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -474,6 +478,58 @@ protected virtual GetHeadersPayload BuildGetHeadersPayload()
...
```
dangershony: `this.Resync`

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -474,6 +478,58 @@ protected virtual GetHeadersPayload BuildGetHeadersPayload()
...
+                this.peerBanning.BanAndDisconnectPeer(peer.PeerEndPoint, "Peer sent oversize inventory message.");
+                this.logger.LogTrace("(-)[OVERSIZE_INVENTORY]");
+                return;
+            }
+
+            // There is no point sending getheaders for every MSG_BLOCK announced. Just send one if needed.
+            bool blockHashAnnounced = false;
+            foreach (InventoryVector inventoryVector in inventory)
+            {
+                if ((inventoryVector.Type == InventoryType.MSG_BLOCK) && (this.chain.GetBlock(inventoryVector.Hash) != null))
```
bokobza: == null!

PR #2612, opened 2018-10-26
Node is failing to synchronise after a reorg fix

noescape00:
Related to #2595

I think the bug appears under following scenario:
We connect chain A because it's better.
We advance some blocks on A and remove block data from CHT that is > 100 blocks behind. Block availability for those is still BlockAvailable because it's in block store.

We rewind more than 100 blocks and and connect chain B. But we still think that blocks from chain A are available in memory or store.
Now A is better again.
We try to connect it but we don't find blocks because we have them marked as available but they are reorged away from BS.

ps
I'm not 100% if this is the right fix, we need to test this PR on internal network.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -543,6 +543,17 @@ namespace Stratis.Bitcoin.Consensus
 
             if (connectBlockResult.Succeeded)
             {
+                // Mark disconnected blocks as not available.
+                // This is needed in case of large reorgs when we disconnect more than CHT.KeepBlockDataForLastBlocks blocks.
+                // In case we disconnect more than that we would expect some of disconnected blocks to be in memory and some in store,
+                // but those that are in store are reorged.
+                // Removing block data will result in redownloading blocks next time node would like to reconnect this chain.
+                foreach (ChainedHeaderBlock disconnectedBlock in disconnectedBlocks)
+                {
+                    disconnectedBlock.ChainedHeader.Block = null;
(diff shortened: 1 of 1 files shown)


noescape00:
Please do not merge before it's confirmed that the bug is not reproducible anymore with this branch.

dangershony:
This obviously looks like the right fix but I think we can actually write a test to simulate this bug.

noescape00:
Yeah, looks like we can reproduce such a scenario in integration test setup.

dangershony:
WE did not check reorgs bigger then 100 (we do change the max reorg property but such tests wont help here) so this should be easily producible in one of @fassadlr tests by just extending the reorg size.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerTests.cs
```diff
@@ -428,5 +428,42 @@ public void ConsensusManager_Connect_New_Block_Failed()
...
+                TestHelper.ConnectAndSync(syncer, minerA, minerB);
+
+                TestHelper.DisableBlockPropagation(syncer, minerA);
+                TestHelper.DisableBlockPropagation(syncer, minerB);
+
+                // Miner A mines 105 blocks.
+                TestHelper.MineBlocks(minerA, 105);
+                TestHelper.WaitForNodeToSync(syncer, minerA);
+
+                //TestHelper.EnableBlockPropagation(minerA, minerB);
```
fassadlr: Remove commented code 👍

dangershony:
This fixes reorgs over 100 blocks problem, your problem is different we are still investigating

PR #2613, opened 2018-10-26
[Integration Tests] Added second interceptor test for node disconnects

fassadlr:
This covers the scenario where minerA/B/C are synced to height 5. Miner C mines on its to height 10. When miner's A an B want to reorg to miner C's chain and their chain data has rewound to block 5, we intercept it and disconnect miner C. Miner's A and B should still synced to block 10.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
@@ -85,7 +85,7 @@ namespace Stratis.Bitcoin.IntegrationTests
             {
                 var minerA = builder.CreateStratisPosNode(new StratisRegTest()).WithDummyWallet();
                 var minerB = builder.CreateStratisPosNode(new StratisRegTest()).WithDummyWallet();
-                var minerC = builder.CreateStratisPosNode(new StratisRegTest()).WithDummyWallet();
+                var syncer = builder.CreateStratisPosNode(new StratisRegTest()).WithDummyWallet();
 
                 // Configure the interceptor to disconnect a node after a certain block has been disconnected (rewound).
                 bool interceptor(ChainedHeaderBlock chainedHeaderBlock)
@@ -94,7 +94,7 @@ namespace Stratis.Bitcoin.IntegrationTests
                     {
                         // Ensure that minerA's tips has rewound to 5.
                         TestHelper.WaitLoop(() => TestHelper.IsNodeSyncedAtHeight(minerA, 5));
-                        TestHelper.Disconnect(minerA, minerC);
+                        TestHelper.Disconnect(minerA, minerB);
                         return true;
                     }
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
```diff
@@ -67,9 +70,67 @@ bool interceptor(ChainedHeaderBlock chainedHeaderBlock)
...
+                    }
+
+                    return false;
+                }
+
+                minerA.BlockDisconnectInterceptor(interceptor);
+
+                // Start the nodes.
+                minerA.Start();
+                minerB.Start();
```
dangershony: miner B never mines perhaps better to keep the terminology of MinerA, MinerB and Syncer

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
```diff
@@ -67,9 +70,67 @@ bool interceptor(ChainedHeaderBlock chainedHeaderBlock)
...
+                minerC.Start();
+
+                // MinerA mines 5 blocks.
+                TestHelper.MineBlocks(minerA, 5);
+
+                // MinerB/C syncs with MinerA.
+                TestHelper.ConnectAndSync(minerA, minerB);
```
dangershony: Use the overload ll nodes in one line

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
```diff
@@ -67,9 +70,67 @@ bool interceptor(ChainedHeaderBlock chainedHeaderBlock)
...
+                TestHelper.MineBlocks(minerA, 4);
+                TestHelper.WaitLoop(() => minerA.FullNode.ConsensusManager().Tip.Height == 9);
+                TestHelper.WaitLoop(() => minerB.FullNode.ConsensusManager().Tip.Height == 9);
+                TestHelper.WaitLoop(() => minerC.FullNode.ConsensusManager().Tip.Height == 5);
+
+                // MinerC mines 5 more blocks so that a reorg is triggered.
+                TestHelper.MineBlocks(minerC, 5);
+
+                // Miner A and B should have reorged to the longer chain.
+                TestHelper.WaitLoop(() => TestHelper.IsNodeSyncedAtHeight(minerA, 10));
```
dangershony: Here is better to check A,B are synced with C then against a specific height.

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
fassadlr: Ah yes, good spot. Will rename 👍

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
fassadlr: 👍

src/Stratis.Bitcoin.IntegrationTests/ConsensusManagerInterceptorTests.cs
fassadlr: Done!

Issue #2588, opened 2018-10-24
PHBSB - ProvenHeadersBlockStoreBehavior


noescape00:
We need to create a new behavior- PHBSB which will be inherited from BSB. 
It should be present only on PoS networks (instead of BSB). 

It should override announcing logic- to PH peers we should announce with proven headers payload, for non-PH nodes- announce like we do it now.

noescape00:
In BSB we have 
```
                case SendHeadersPayload sendHeadersPayload:
                    this.PreferHeaders = true;
                    break;
```

Same should be for SendProvenHeadersPayload except for we also will save height from where it's preferred over normal headers. 

When we announce blocks we first check their height and if proven headers are preferred starting from before that height then we announce proven headers. If preferred height is after height of headers we will announce with normal headers.

noescape00:
Fixed

PR #2615, opened 2018-10-26
PHBSB - ProvenHeadersBlockStoreBehavior (#2588)

MithrilMan:
Created ProvenHeadersBlockStoreBehavior as per #2588.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -37,7 +37,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
         private const int MaxBlocksToAnnounce = 8;
 
         private readonly ConcurrentChain chain;
-        
+
         private readonly IConsensusManager consensusManager;
 
         private ConsensusManagerBehavior consensusManagerBehavior;
@@ -125,7 +125,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
             }
         }
 
-        private async Task ProcessMessageAsync(INetworkPeer peer, IncomingMessage message)
+        protected virtual async Task ProcessMessageAsync(INetworkPeer peer, IncomingMessage message)
         {
             switch (message.Message.Payload)
             {
@@ -380,7 +380,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     // We expect peer to answer with getheaders message.
                     if (bestSentHeader == null)
                     {
-                        await peer.SendMessageAsync(new HeadersPayload(blocksToAnnounce.Last().Header)).ConfigureAwait(false);
+                        await peer.SendMessageAsync(BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, blocksToAnnounce.Last().Header)).ConfigureAwait(false);
 
                         this.logger.LogTrace("(-)[SENT_SINGLE_HEADER]");
                         return;
@@ -434,7 +434,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                         this.lastSentHeader = bestIndex;
                         this.consensusManagerBehavior.UpdateBestSentHeader(this.lastSentHeader);
 
-                        await peer.SendMessageAsync(new HeadersPayload(headers.ToArray())).ConfigureAwait(false);
+                        await peer.SendMessageAsync(BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, headers.ToArray())).ConfigureAwait(false);
                         this.logger.LogTrace("(-)[SEND_HEADERS_PAYLOAD]");
                         return;
                     }
@@ -479,6 +479,20 @@ namespace Stratis.Bitcoin.Features.BlockStore
             }
         }
 
+        /// <summary>
+        /// Builds the announced header payload.
+        /// This method can be overridden to return different type of HeadersPayload, e.g. <see cref="ProvenHeadersPayload" />
+        /// </summary>
+        /// <param name="blockstoreTipHeight">Height of the <see cref="ChainState.BlockStoreTip"></see> height.</param>
+        /// <param name="headers">The headers.</param>
+        /// <returns>
+        /// The <see cref="HeadersPayload" /> instance to announce to the peer.
+        /// </returns>
+        protected virtual Payload BuildAnnouncedHeaderPayload(int blockstoreTipHeight, params BlockHeader[] headers)
+        {
+            return new HeadersPayload(headers);
+        }
+
         public override object Clone()
         {
             var res = new BlockStoreBehavior(this.chain, this.chainState, this.loggerFactory, this.consensusManager)
@@ -486,7 +500,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 CanRespondToGetBlocksPayload = this.CanRespondToGetBlocksPayload,
                 CanRespondToGetDataPayload = this.CanRespondToGetDataPayload
             };
-            
+
             return res;
         }
     }
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
@@ -380,7 +380,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     // We expect peer to answer with getheaders message.
                     if (bestSentHeader == null)
                     {
-                        await peer.SendMessageAsync(BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, blocksToAnnounce.Last().Header)).ConfigureAwait(false);
+                        await peer.SendMessageAsync(this.BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, blocksToAnnounce.Last().Header)).ConfigureAwait(false);
 
                         this.logger.LogTrace("(-)[SENT_SINGLE_HEADER]");
                         return;
@@ -434,7 +434,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                         this.lastSentHeader = bestIndex;
                         this.consensusManagerBehavior.UpdateBestSentHeader(this.lastSentHeader);
 
-                        await peer.SendMessageAsync(BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, headers.ToArray())).ConfigureAwait(false);
+                        await peer.SendMessageAsync(this.BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, headers.ToArray())).ConfigureAwait(false);
                         this.logger.LogTrace("(-)[SEND_HEADERS_PAYLOAD]");
                         return;
                     }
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <inheritdoc />
+    public class ProvenHeadersBlockStoreBehavior : BlockStoreBehavior
+    {
+        private readonly Network network;
+
+        /// <summary>
+        /// Gets or sets the height from which start serving Proven Headers, if > 0.
+        /// </summary>
+        public int AnnounceProvenHeaderFromHeight { get; set; } = 0;
```
zeptin: Should be `AnnounceProvenHeadersFromHeight` I think.

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -0,0 +1,82 @@
...
+            return Task.CompletedTask;
+        }
+
+        /// <inheritdoc />
+        /// <returns>The <see cref="HeadersPayload"/> instance to announce to the peer, or <see cref="ProvenHeadersPayload"/> if the peers requires it.</returns>
+        protected override Payload BuildAnnouncedHeaderPayload(int blockstoreTipHeight, params BlockHeader[] headers)
+        {
+            if (this.AnnounceProvenHeaderFromHeight > 0 && blockstoreTipHeight >= this.AnnounceProvenHeaderFromHeight)
+            {
+                ProvenHeadersPayload provenHeadersPayload = new ProvenHeadersPayload();
```
zeptin: I think you've duplicated this portion of code in 2 different ways.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -85,13 +88,21 @@ private void AddInlineStats(StringBuilder log)
 
         public override Task InitializeAsync()
         {
-            this.connectionManager.Parameters.TemplateBehaviors.Add(new BlockStoreBehavior(this.chain, this.chainState, this.loggerFactory, this.consensusManager));
+            // Use ProvenHeadersBlockStoreBehavior for PoS Networks
+            if (this.network.Consensus.Options is PosConsensusOptions)
```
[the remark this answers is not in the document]
MithrilMan: ok about IsProofOfStake, changing to that, about configuration it's out of scope now, I'm not sure what's the best thing to do actually.

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
MithrilMan: lol yes mb, I wanted to use Linq  expression and forget to remove the other part :)

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    /// <inheritdoc />
+    public class ProvenHeadersBlockStoreBehavior : BlockStoreBehavior
+    {
+        private readonly Network network;
+
+        /// <summary>
+        /// Gets or sets the height from which start serving Proven Headers, if > 0.
+        /// </summary>
+        public int AnnounceProvenHeadersFromHeight { get; set; } = 0;
```
dangershony: default is zero right so no need to set it.

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        /// <inheritdoc />
+        /// <returns>The <see cref="HeadersPayload"/> instance to announce to the peer, or <see cref="ProvenHeadersPayload"/> if the peers requires it.</returns>
+        protected override Payload BuildAnnouncedHeaderPayload(int blockstoreTipHeight, params BlockHeader[] headers)
+        {
+            if (this.AnnounceProvenHeadersFromHeight > 0 && blockstoreTipHeight >= this.AnnounceProvenHeadersFromHeight)
+            {
+                return new ProvenHeadersPayload(
+                    from header in headers
```
dangershony: I prefer to not use linq here (link is ok but we avoid it in the core protocol code so lets keep it that way)

src/Stratis.Bitcoin/P2P/Protocol/Payloads/ProvenHeadersPayload.cs
```diff
@@ -29,6 +29,11 @@ public ProvenHeadersPayload(params ProvenBlockHeader[] headers)
             this.Headers.AddRange(headers);
         }
 
+        public ProvenHeadersPayload(IEnumerable<ProvenBlockHeader> headers)
```
dangershony: Why not just use the method above?

MithrilMan: to save an .array() on that linq code, do you want to remove it?

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -0,0 +1,74 @@
...
+                    break;
+                default:
+                    await base.ProcessMessageAsync(peer, message).ConfigureAwait(false); ;
+                    break;
+            }
+        }
+
+        private Task ProcessSendProvenHeadersPayload(SendProvenHeadersPayload sendProvenHeadersPayload)
+        {
+            this.PreferHeaders = true;
```
dangershony: I am not sure this flag is part of PH, its about a node saying it preferes to receive headers, we should create a new flag if anything PreferProvenHeaders maybe?
But we dont use that because we default to anyway to prefer headers so its actually obsoulete

MithrilMan: I didn't created PreferProvenHeaders because PreferHeaders was enough to communicate that you prefer headers over Inv.
If you think it's obsolete then we should remove it completely from the project?
Or just leave it now then think about what to do later

src/Stratis.Bitcoin/P2P/Protocol/Payloads/ProvenHeadersPayload.cs
dangershony: yeah I prefer not to use linq in this components

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
dangershony: ok yeah leave it now but you dont need to set it do you? (ah yeah maybe you need ok)

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
```diff
@@ -59,11 +59,14 @@ protected override Payload BuildAnnouncedHeaderPayload(int blockstoreTipHeight,
...
             {
-                return new ProvenHeadersPayload(
-                    from header in headers
-                    let posBlock = new PosBlock(header)
-                    select ((PosConsensusFactory)this.network.Consensus.ConsensusFactory).CreateProvenBlockHeader(posBlock)
-                    );
+                ProvenHeadersPayload provenHeadersPayload = new ProvenHeadersPayload();
+                foreach (var header in headers)
+                {
+                    var posBock = new PosBlock(header);
```
dangershony: This is not correct, you need to use `Network.ConsensusFctory.CreateBlock()`, fix it in the next PR

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -84,7 +87,15 @@ private void AddInlineStats(StringBuilder log)
 
         public override Task InitializeAsync()
         {
-            this.connectionManager.Parameters.TemplateBehaviors.Add(new BlockStoreBehavior(this.chain, this.chainState, this.loggerFactory, this.consensusManager));
+            // Use ProvenHeadersBlockStoreBehavior for PoS Networks
+            if (this.network.Consensus.IsProofOfStake)
```
dangershony: I don't like to use this flag but for now ok.

MithrilMan: https://github.com/stratisproject/StratisBitcoinFullNode/pull/2615#discussion_r228702266

ehm... conflicts? :D
what's the best way and why?

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
MithrilMan: CreateBlock doesn't have a header argument, should we add it?

dangershony: Ah hang on this is probably not correct, you don't just create the PH like this, the block you created is missing infomration, you need the block itself
but also potentially you will have the PH available in the PH store so I guess this is the next PR

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs
```diff
@@ -434,7 +434,7 @@ public async Task AnnounceBlocksAsync(List<ChainedHeader> blocksToAnnounce)
                         this.lastSentHeader = bestIndex;
                         this.consensusManagerBehavior.UpdateBestSentHeader(this.lastSentHeader);
 
-                        await peer.SendMessageAsync(new HeadersPayload(headers.ToArray())).ConfigureAwait(false);
+                        await peer.SendMessageAsync(BuildAnnouncedHeaderPayload(this.chainState.BlockStoreTip.Height, headers.ToArray())).ConfigureAwait(false);
```
dangershony: Here you probably want to pass the consensus tip not store tip (but I am not entirely sure)

dangershony: Thats because store tip may take up to 37 seconds to updated (thats the persist batch interval)

dangershony: Mental note we should probably set Chain.StoreTip to be the tip in memory not in disk @fassadlr what do you think?

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreBehavior.cs
dangershony: I think also there is a `Network.ConsensusFctory.CreateProvenHeader` method

PR #2617, opened 2018-10-27
[Integration Tests] Additional stratisX tests

zeptin:
What it says on the tin.


fassadlr:
@zeptin can you merge latest from master so that we can see how the build runs?

zeptin:
@fassadlr done. Note that this may require #2611 to be merged before the tests will pass, but let's see.

Issue #1937, opened 2018-08-16
[Activation] tests for consensus manager


noescape00:
See test cases here: https://github.com/stratisproject/Research/blob/master/CoreSyncingAndProvenHeaders/ConsensusManager.txt

dangershony:
==================================================TESTS==================================================

=====HeadersPresented
1) Setup blockPuller.AverageBlockSize = 200 bytes. Checkpoints are disabled. 10 new headers are presented. Now BP has a peer tip at the last header (BP.NewTipClaimed is called with the last header), BP is asked to download 10 blocks (BP.RequestNewData called). Make sure that expectedBlockDataBytes equals to 200*10. Make sure that hashes of those 10 headers are inside expectedBlockSized list. Make sure that peer that was supplied was added to peersByPeerId.
2) Setup blockPuller.AverageBlockSize = 200 bytes. Checkpoints are disabled, max MaxUnconsumedBlocksDataBytes = 1000. Avg block size is 500. 3 headers are presented, 2 are asked for download. 1 item is in the ToDownloadQueue. 
Make sure that expectedBlockDataBytes equals to 200*2. Make sure that hashes of 2 asked for download headers are inside expectedBlockSized list. Make sure that peer that was supplied was added to peersByPeerId.
3) 10 headers are presented, 10 are consumed by CHT (HeadersPresented returns the last header)
4) HeadersPresented called on 10 headers with triggerDownload == false. Make sure that BP wasn't asked for those blocks.
5) Setup blockPuller.AverageBlockSize = 200 bytes. MaxBlocksToAskFromPuller=100. Checkpoints are disabled. 200 headers are presented, 100 are asked for download. 5 was delivered. Nothing happens. 5 more deleivered- 10 asked from puller. 

====ProcessDownloadedBlock
1) Checkpoints are disabled. CT is at block 10. Block 11 is received. partialValidatior.StartValidation is called
2) Checkpoints are disabled. CT is at 9. Block data for 11 is received. partialValidatior.StartValidation is not called

====BlockDownloaded
1) Callback is registered with the hash of block X. Block data arrived for for an unknown header- nothing happens (callback not called). 
2) Callback is registered with the hash of block X. ExpectedBlockSIzes list initialized with block X. Block X data arrived with fake signature is presented- callback not called, peer is banned. Make sure that ExpectedBlockSizes is empty.
[...]

PR #2620, opened 2018-10-28
Tests for consensusmanager (#1937)

Neurosploit:
First set of tests for https://github.com/stratisproject/StratisBitcoinFullNode/issues/1937
These are the tests up to and including OnPartialValidationCompletedCallbackAsync

* refactored consensusmanager and consensustestcontext to what I needed for these tests. CreateConsensusManager was in the way quite a bit, I initially tried to rework the ConsensusManagerHelper with automock to easily setup the consensus manager and allow for some overrides but I needed too much of the consensustestcontext so merged that into the consensustestcontext.
* added a testconsensusmanager with some helper methods.
* added an inmemory test coin view to help with testability.

There are some differences from the initial task:

HeadersPresented
> Setup blockPuller.AverageBlockSize = 200 bytes. Checkpoints are disabled, max MaxUnconsumedBlocksDataBytes = 1000. Avg block size is 500. 3 headers are presented, 2 are asked for download. 1 item is in the ToDownloadQueue.
> Make sure that expectedBlockDataBytes equals to 200*2. Make sure that hashes of 2 asked for download headers are inside expectedBlockSized list. Make sure that peer that was supplied was added to peersByPeerId.

Original test does not pass but added with skip so you guys can have a look at it. As far as I can tell that is not going to be possible without downloading the block which is a requirement not to do that early:
- HeadersPresented_NewHeaders_BlockSizeTotalHigherThanMaxUnconsumedBlocksDataBytes_UnexpectedlyBiggerBlocksThanAverage_LimitsDownloadedBlocks

Instead added another test of what I think is right:
- HeadersPresented_NewHeaders_BlockSizeTotalHigherThanMaxUnconsumedBlocksDataBytes_LimitsDownloadedBlocks

BlockDownloaded
- Merged most peer ban tests together and just went with block integrity invalidated as from the code logic they were testing the same thing.

GetOrDownloadBlocks/BlockMined
- This was unit testable so went with that instead.

### DIFF
--- a/src/Stratis.Bitcoin.Tests/Consensus/ConsensusManagerTests.cs
+++ b/src/Stratis.Bitcoin.Tests/Consensus/ConsensusManagerTests.cs
@@ -606,6 +606,7 @@ namespace Stratis.Bitcoin.Tests.Consensus
                 });
 
             Assert.Throws<Bitcoin.Consensus.ConsensusException>(() => builder.ConsensusManager.BlockMinedAsync(additionalHeaders.Block).GetAwaiter().GetResult());
+            Assert.Equal(builder.InitialChainTip, builder.ConsensusManager.Tip);
         }
 
         [Fact]
@@ -627,6 +628,7 @@ namespace Stratis.Bitcoin.Tests.Consensus
                 });
 
             Assert.Throws<Bitcoin.Consensus.ConsensusException>(() => builder.ConsensusManager.BlockMinedAsync(additionalHeaders.Block).GetAwaiter().GetResult());
+            Assert.Equal(builder.InitialChainTip, builder.ConsensusManager.Tip);
         }
 
         [Fact]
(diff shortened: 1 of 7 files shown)


dangershony:
Boom... he did it again!!!

fassadlr:
Amazing job on these @Neurosploit!

Seeing as we are about expose some fields of CM I want to propose that we implement a class that holds CM related information, like PeersById, Tip etc. Something to the effect of a`ConsensusManagerState` class. That way we call `consensusMangager.CurrentState` which returns this class and get loads of useful of information on CM and CHT at any time.

This way we dont change the visibility of CM fields and we can easily add to the class if we need to data.

@dangershony what do you think?

Neurosploit:
I think that is a good idea @fassadlr. One other thing I would like to see as a future improvement is pulling out the callbacks into their own classes to they can be tested seperately without having to include into the tests this already massive class.

dangershony:
Both good idea @fassadlr @Neurosploit 
Please merge conflicts

fassadlr:
@dangershony I can perhaps help @Neurosploit on this one? So that we can look at creating a Snapshot / State obejct that can be returned from CM at any given time.

I will keep in mind the locks as discussed.

Neurosploit:
@fassadlr much appreciated.

Pulling out the callbacks for both blockdownloaded and partialvalidationsucceeded in a nice way isn't as easy as I thought as there is code that is shared with other methods on the consensusmanager.

PR #2623, opened 2018-10-29
[Mempool] AttachedPeer can be null when relaying.

fassadlr:
This is to see if this fixes the timeout in integration tests as this test fails to complete:

`MempoolReceiveFromManyNodes`

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
@@ -314,7 +314,7 @@ namespace Stratis.Bitcoin.Features.MemoryPool
                     this.logger.LogInformation("Transaction ID '{0}' inventory sent in violation of protocol peer '{1}'.", inv.Hash, peer.RemoteSocketEndpoint);
                     continue;
                 }
-   
+
                 send.Inventory.Add(new InventoryVector(peer.AddSupportedOptions(InventoryType.MSG_TX), inv.Hash));
             }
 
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -487,6 +487,12 @@ public void RelayTransaction(uint256 hash)
             IEnumerable<MempoolBehavior> behaviours = peers.Select(s => s.Behavior<MempoolBehavior>());
             foreach (MempoolBehavior mempoolBehavior in behaviours)
             {
+                if (mempoolBehavior?.AttachedPeer == null)
```
zeptin: Do we have anything that cleans out old behaviours when they are no longer attached to anything?

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
```diff
@@ -487,6 +487,12 @@ public void RelayTransaction(uint256 hash)
...
+                {
+                    this.logger.LogTrace("Peer is null, skipped.");
+                    continue;
```
dangershony: I don't think `continue;` is correct if we have no attached peer no point to continue.

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
dangershony: ah sorry this is per behaviour so ok its correct

src/Stratis.Bitcoin.Features.MemoryPool/MempoolBehavior.cs
dangershony: you can still have race conditions later down the line,
I suggest taking a local var of the peer

```
var peer = mempoolBehavior?.AttachedPeer

if(peer == null)
  continue
```
form here us `peer`

fassadlr: Fixed 👍

PR #2624, opened 2018-10-29
Added an address book

bokobza:
Added an address book to be used by the wallet.
Users can then save the addresses they use the most: addresses of friends, but also addresses of sidechains federations and addresses of smart contracts they  use often.

Here is the api:

AddressBook:

**1. to add a new entry:**
POST `/api/AddressBook/address`
Example request:
```
{
  "label": "add7",
  "address": "TQNyrEPc4qHxWN96dBAjncBeB2ghJPqYVu"
}
```
Returns:
```
{
  "address": "TQNyrEPc4qHxWN96dBAjncBeB2ghJPqYVu",
  "label": "add7"
}
```

**2. to remove an entry:**
DELETE `/api/AddressBook/address`
Example request: http://localhost:38221/api/AddressBook/address?label=add7
Returns:
```
{
  "address": "TQNyrEPc4qHxWN96dBAjncBeB2ghJPqYVu",
  "label": "add7"
}
```

**3. to get the address book:**
GET `/api/AddressBook`
Optional query params: int skip, int take.
Example request:
http://localhost:38221/api/AddressBook
http://localhost:38221/api/AddressBook?skip=10&take=5
Returns:
```
{
  "addresses": [
    {
      "address": "TFE7R2FSAgAeJxt1fgW2YVCh9Zc448f3ms",
      "label": "add1"
    },
    {
      "address": "TDfosbE6ChGKdH9cVpfgbKzbvFJGLs1zgq",
      "label": "add2"
    },
    {
      "address": "TJcyatE4XYTkJ5DYUmoHdG9Eij7W3vjGZr",
      "label": "add3"
    },
    {
      "address": "TC52WGLwE1KE1bXvD6f4MC7i5QtxNUGiUb",
      "label": "add4"
    },
    {
      "address": "TM9i96uQDFDancRp5bUR5ea16CMWLkCYhK",
      "label": "add5"
    },
    {
      "address": "TWMxjBk5bVdv8dhDJ645Z5RoxfrbRUJewa",
      "label": "add6"
    }
  ]
}
```


src/Stratis.Bitcoin.Features.Wallet/AddressBookManager.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.fileStorage = new FileStorage<AddressBook>(dataFolder.RootPath);
+
+            if (this.fileStorage.Exists(AddressBookFileName))
+            {
+                this.addressBook = this.fileStorage.LoadByFileName(AddressBookFileName);
+            }
+            else
+            {
+                this.addressBook = new AddressBook();
+                this.fileStorage.SaveToFile(this.addressBook, AddressBookFileName);
```
dangershony: We should probably not save to disk on constructors, this is called form the DI and we have no control over when constructors are called and they should also be as light as possible.

bokobza: This will be done only the first time this code is run though.
Maybe in a start method?

PR #2627, opened 2018-10-29
Fix AddToPending bug and integration tests

fassadlr:


fassadlr:
Fix #2630

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -289,16 +294,19 @@ public void AddToPending(ChainedHeaderBlock chainedHeaderBlock)
         {
             lock (this.blocksCacheLock)
             {
-                bool added = this.pendingBlocksCache.TryAdd(chainedHeaderBlock.ChainedHeader.HashBlock, chainedHeaderBlock);
-
-                if (added)
+                ChainedHeaderBlock existing = null;
+                this.pendingBlocksCache.TryGetValue(chainedHeaderBlock.ChainedHeader.HashBlock, out existing);
```
dangershony: Why replace one operation with two?

Most of the time `TryAdd` will work so I would suggest just leave it like that.

src/Stratis.Bitcoin.IntegrationTests.Common/Runners/StratisBitcoinPowRunner.cs
```diff
@@ -23,17 +23,21 @@ public override void BuildNode()
...
+                        .UseBlockStore()
+                        .UsePowConsensus()
+                        .UseMempool()
+                        .AddMining()
+                        .UseWallet()
+                        .AddRPC()
+                        .UseApi()
+                        .MockIBD();
+
+            if (this.ServiceToOverride != null)
```
dangershony: Don't use block store use base feature instead else its confusing, it does nto matter so much because the services DI is shared with all features

PR #2629, opened 2018-10-29
Remove exception from the normal minting flow

### LINKED ISSUE #2473: [Mining] stakeValidator.CheckKernel should use return codes, not exceptions (stops API from responding)
The current implementation of CheckKernel relies on exceptions for normal processing, and not return codes. It is throwing a huge amount of exceptions when processing. It does this so much, that the API ends up becoming unresponsive for the wallet. The API starts throwing errors short time after enabling staking.

Changing the implementation of CheckKernel to return status codes (int), fixes the issues. I have tested for a few minutes, and the API responds with no issues in less than 100 ms for most requests.

With exceptions being thrown in this method, then requests starts taking 3+ seconds and fails.

Here is network log, showing what happens during the kernel execution:

image

And sometimes it is total stop:

image

Individual loop for each of the 30 threads that is spawned for POS "mining", is reduced from 6692346-185683984 ticks to only 20-5000 ticks. Huge performance increase by not relying on exceptions for normal processing (it checks for a specific exception and then continues to process the loop).

My suggestion is to do something similar to:

var result = this.stakeValidator.CheckKernel(contextInformation, chainTip, block.Header.Bits, txTime, prevoutStake);

And have CheckKernel return status code (integer) for the various steps in the validation.

Discovered this when I was attempting to debug why staking (not on Stratis network) wouldn't work properly, and I kept getting errors from the API whenever staking was enabled.

The CheckKernel is processed to the very end, this is where it "fails":

```
// Now check if proof-of-stake hash meets target protocol.
var hashProofOfStakeTarget = new BigInteger(1, context.HashProofOfStake.ToBytes(false));
if (hashProofOfStakeTarget.CompareTo(weightedTarget) > 0)
{
    this.logger.LogTrace("(-)[TARGET_MISSED]");
    ConsensusErrors.StakeHashInvalidTarget.Throw();
```
The wallet has more than 700 transactions which is part of the staking.

sondreb:
- The minting process spawns many threads, all attempting to find the correct hash. If the resulting hash does not qualify with correct target, it will return a boolean result.
- CPU usage spikes on minting, sometimes resulting in thousands of thrown exceptions. Since it is expeccted behavior to not find hash with correct target, this should not be an exceptional case.
- Closes #2473


fassadlr:
@sondreb this is a great change! Can you please latest from master that we can see that the build passes?

src/Stratis.Bitcoin.Features.BlockStore.Tests/BlockStoreTests.cs
```diff
@@ -86,7 +86,9 @@ public BlockStoreTests()
 
             this.chainState = new ChainState();
 
-            this.blockStoreQueue = new BlockStoreQueue(this.chain, this.chainState, new StoreSettings(NodeSettings.Default(this.network)),
+            var blockStoreFlushCondition = new BlockStoreQueueFlushCondition(this.chainState);
```
dangershony: This class was introduced by @fassadlr a few commits ago, @sondreb I think you did not do the merge correctly

sondreb: Have done a new merge and fixes the conflicts, now the PR only contains the correct changes in 5 files.

sondreb:
Merge and resolved conflicts, so only correct changes should be part of PR now. Thanks! :-)

fassadlr:
@sondreb I'm going to run these changes on my VM and confirm that staking still works as expected.

fassadlr:
@dangershony I can confirm that staking still works as usual. I am happy with this change if you are?

src/Stratis.Bitcoin.Features.Consensus.Tests/Rules/ProvenHeaderRules/ProvenBlockHeaderCoinstakeRuleTest.cs
```diff
@@ -308,11 +308,11 @@ public void RunRule_ProvenHeadersActive_And_NullPreviousStake_PrevStakeErrorIsTh
                 .Setup(m => m.VerifySignature(It.IsAny<UnspentOutputs>(), It.IsAny<Transaction>(), It.IsAny<int>(), It.IsAny<ScriptVerify>()))
                 .Returns(true);
 
-            // When we run the validation rule, we should hit previous stake null error.
+            // When we run the validation rule, we should hit bad merkle root.
```
fassadlr: @sondreb can you revert this? @MithrilMan is fixing this in another PR.

fassadlr:
Actually, @sondreb (and @dangershony ) I just realized that you will need to update `ProvenHeaderCoinStakeRule.CheckStakeKernelHash' to throw the exception which you removed (`StakeHashInvalidTarget `) as the rule depends on an exception being thrown so that it signals a failure.

So in `ProvenHeaderCoinStakeRule.CheckStakeKernelHash' check the kernel hash and if it doesnt succeed, throw StakeHashInvalidTarget. See how the other methods do it in ProvenHeaderCoinStakeRule.

Then you can update the unit test.

fassadlr:
@sondreb I've made the changes as I suggested and am busy staking to confirm that all still works as expected.

fassadlr:
I can confirm that Staking works as per normal:

[My staking address on TestNet](https://chainz.cryptoid.info/strat-test/address.dws?TZC65AhriPcVH7SK7wNhquHnU2tYryRmp2.htm)

sondreb:
Merged with latest changes upstream, have updated tests and made it all green. I have not tested the latest changes at runtime.

fassadlr:
@dangershony I managed to stake some coins with the new code 👍 

https://chainz.cryptoid.info/strat-test/address.dws?TCBTJQep8C1cNzWw9Rst4B8Tcn1pLAEXq7.htm

PR #2636, opened 2018-10-30
Add optional callback for message sending

zeptin:
I use this as a building block for recording the exact traffic exchanged between peers, for protocol diagnostics & test-writing. Incoming traffic can be intercepted with no node changes via a dummy behaviour, but outgoing traffic requires some sort of hook into the network code, hence this PR.


bokobza:
That's cool. How would you use it?

bokobza:
utACK

@dangershony

dangershony:
If tests pass this looks ok to me @zeptin

PR #2645, opened 2018-10-31
[Consensus] Fix ConnectBlockResult null reference

fassadlr:


src/Stratis.Bitcoin/Consensus/ValidationResults/ConnectBlocksResult.cs
```diff
@@ -26,6 +26,9 @@ public override string ToString()
             if (this.Succeeded)
                 return $"{nameof(this.Succeeded)}={this.Succeeded}";
 
+            if (this.PeersToBan == null)
```
bokobza: can't you remove the 
```
if (this.Succeeded)
   return $"{nameof(this.Succeeded)}={this.Succeeded}";
```

fassadlr: I suppose if Succeeed is true we dont want log all the other properties?

Issue #2589, opened 2018-10-24
PH BlockStoreSignaled


noescape00:
Create PH version of block store signaled class. 
Difference in logic comparing to base class: 
when block is signaled we should check if it's at height after PH activation ad if it is- check if PH is available in PH store. It it's not we should construct PH and put it in store. This logic should be executed before we announce those headers to our peers. 

PH BlockStoreSignaled should be enabled only for PoS networks. 

ps
this logic is needed for cases when we synced blocks using headers acquired from whitelisted QT nodes- for those we don't have proven headers and need to construct them to allow C# nodes sync from us.

MithrilMan:
I'm doing this as per specifics, but what happens if a reorg happens?

MithrilMan:
needed re-design, addressed in https://github.com/stratisproject/StratisBitcoinFullNode/issues/2662.

PR #2660, opened 2018-11-02
PH BlockStoreSignaled (#2589)

MithrilMan:
Implemented as per #2589 requirements

Actually it's not yet complete because it isn't clear what to do once a ProvenHeader is created because the arguments passed, are then used to announce the blocks and so if a header is a legacy header this wouldn't work.
Moreover the eventual assignment
//blockPair.Block.Header = blockPair.ChainedHeader.Header = createdProvenHeader;
can't actually be done because both those property setters aren't accessible and I'm not sure is a good idea to set them public or have a way to change it once a block has been instantiated

as a final note it seems we want to get rid of whitelisted legacy nodes so the creation shouldn't be needed (all nodes should talk using proven header)

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
@@ -1,8 +1,11 @@
-﻿using Microsoft.Extensions.Logging;
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Logging;
 using NBitcoin;
 using Stratis.Bitcoin.Base;
 using Stratis.Bitcoin.Connection;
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
```diff
@@ -0,0 +1,109 @@
...
+                return blockHeight >= options.ProvenHeadersActivationHeight;
+            }
+
+            return false;
+        }
+
+        /// <inheritdoc />
+        protected override void AddBlockToQueue(ChainedHeaderBlock blockPair)
+        {
+            base.AddBlockToQueue(blockPair);
```
dangershony: I think we may want to first create the PH then call `AddBlockToQueue`

Issue #1790, opened 2018-08-01
Rule to verify UTXO spent on another chain + Tests.


noescape00:
forkPoints data structure is a key-value storage where key is a TxId + N (N is an index of output in a transaction) and value is a rewind data index. This data structure will always contain as many entries as there are rewind data instances in the database (currently we do not delete old rewind data that is no longer needed but after the issue #5 is fixed we should also make sure that old fork point data is deleted as well).

When we need to fetch a UTXO first we check if it was spent in a rewind data with index that is associated with a block created after the fork point. If the coin was spent before the fork point then the coinstake where this coin is being used as an input is invalid. Otherwise we use rewind data id to fetch the rewind data instance and get a UTXO from it.

forkPoints data structure should be updated every time new rewind data instance is created or deleted. Since this data structure will be very lightweighted it should be kept in the RAM and persisted to the disk every time it is updated (this includes altering the fork point data starting from the fork point in case a reorg happens).

See: https://github.com/stratisproject/Research/blob/master/CoreSyncingAndProvenHeaders/ProvenHeaders.md

PR #2669, opened 2018-11-05
[PH] WIP - Rule to verify UTXO spent on another chain

fenix2222:
First PR for Rule to verify UTXO spent on another chain as per https://github.com/stratisproject/StratisBitcoinFullNode/issues/1790.

Preferred reviewers @dangershony or @noescape00


src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -150,17 +158,19 @@ private class CacheItem
...
             this.dateTimeProvider = dateTimeProvider;
             this.stakeChainStore = stakeChainStore;
+            this.rewindDataIndexStore = rewindDataIndexStore;
             this.MaxItems = CacheMaxItemsDefault;
             this.lockobj = new AsyncLock();
             this.cachedUtxoItems = new Dictionary<uint256, CacheItem>();
             this.performanceCounter = new CachePerformanceCounter(this.dateTimeProvider);
             this.lastCacheFlushTime = this.dateTimeProvider.GetUtcNow();
-            this.cachedRewindDataList = new List<RewindData>();
+            this.cachedRewindDataList = new List<(int height, RewindData rewindData)>();
```
dangershony: Why not a dictionary then? we will always have only one height ever stored.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -299,7 +313,7 @@ public async Task FlushAsync(bool force = true)
                     u.Value.ExistInInner = true;
                 }
 
-                await this.Inner.SaveChangesAsync(unspent.Select(u => u.Value.UnspentOutputs).ToArray(), null, this.innerBlockHash, this.blockHash, this.cachedRewindDataList).ConfigureAwait(false);
+                await this.Inner.SaveChangesAsync(unspent.Select(u => u.Value.UnspentOutputs).ToArray(), null, this.innerBlockHash, this.blockHash, 0, this.cachedRewindDataList.Select(c => c.rewindData).ToList()).ConfigureAwait(false);
```
dangershony: Still not sure how the height will be found?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -405,6 +420,15 @@ public async Task SaveChangesAsync(IList<UnspentOutputs> unspentOutputs, IEnumer
 
                     cacheItem.IsDirty = true;
 
+                    if (this.rewindDataIndexStore != null)
+                    {
+                        for (int i = 0; i < unspent.Outputs.Length; i++)
+                        {
+                            string key = $"{unspent.TransactionId}-{i}";
+                            indexItems[key] = checked((int)unspent.Height);
```
dangershony: just use `uint`

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <param name="loggerFactory">Factory to create a logger for this type.</param>
+        public RewindDataIndexRepository(string folder, ILoggerFactory loggerFactory)
+        {
+            Guard.NotNull(folder, nameof(folder));
+
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            Directory.CreateDirectory(folder);
+
+            this.dbreeze = new DBreezeEngine(folder);
```
dangershony: Should we not reuse the same transaction from DbreezeCoinview?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// and value is a rewind data index
+        /// </summary>
+        private readonly ConcurrentDictionary<string, int> items;
+
+        /// <summary>Time of the last cache flush.</summary>
+        private DateTime lastCacheFlushTime;
+
+        /// <summary>Length of the coinview cache flushing interval in seconds.</summary>
+        /// <seealso cref="lastCacheFlushTime"/>
+        public const int CacheFlushTimeIntervalSeconds = 60;
```
dangershony: Do we nee da flush interval? dont we just flush when coinview flushes?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (!force && ((now - this.lastCacheFlushTime).TotalSeconds < CacheFlushTimeIntervalSeconds))
+            {
+                this.logger.LogTrace("(-)[NOT_NOW]");
+                return;
+            }
+
+            IDictionary<string, int> itemsToSave = this.items.ToDictionary(i => i.Key, i => i.Value);
+            await this.rewindDataIndexRepository.PutAsync(itemsToSave).ConfigureAwait(false);
+
+            this.items.Clear();
```
dangershony: We probably want to keep in cache at least 500 items this is likely to be called most.

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -270,6 +275,31 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+        /// </summary>
+        /// <param name="header">The proven block header.</param>
+        /// <exception cref="ApplicationException">TODO: throw specific exception to be caught later</exception>
+        private void CheckIfCoinstakeIsSpentOnAnotherChain(ProvenBlockHeader header)
+        {
+            Transaction coinstake = header.Coinstake;
+            TxIn intput = coinstake.Inputs[0];
+
+            int? rewindDataIndex = this.PosParent.RewindDataIndexStore.GetAsync(intput.PrevOut.Hash, (int)intput.PrevOut.N).GetAwaiter().GetResult();
+            if (!rewindDataIndex.HasValue)
```
dangershony: just `return` is incorrect right?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderForkPointRule.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
+{
+    /// <summary>
+    /// Rule to check if the serialized sizes of the proven header components, such as merkle proof (max 512 bytes),
+    /// signature (max 80 bytes) and coinstake (max 1,000,000 bytes), do not exceed maximum possible size allocation.
+    /// </summary>
+    /// <seealso cref="ProvenHeaderRuleBase" />
+    public class ProvenHeaderForkPointRule : ProvenHeaderRuleBase
+    {
+        public override void Run(RuleContext context)
```
dangershony: These seems very similar to `ProvenHeaderSizeRule`

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: I wanted to keep changes in CachedCoinView to a minimum as changing it to dictionary changes the logic in CoinView (searching, getting last item, etc.). I have changed it to a SortedDictionary for now to keep it as close to the original logic as possible.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
fenix2222: `Inner` coin view is a dbreeze storage. All rewind data is stored in dbreeze with key=height. So for dbreeze we don't need to pass height in, I only need it in cached coin view. But because they both implement `ICoinView` I have to pass some value in.

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
fenix2222: Yeah, we probably don't need it if it is controlled by coin view. I will remove it for now.

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
fenix2222: How do we do this? All the disk access is hidden away in DbreezeCoinview and we don't have access to it.

fenix2222: Wouldn't we then tightly couple rewind data index store to a specific implementation of coin view - DBreezeCoinView. Maybe we should consider moving RewindDataIndexStore as a dependency to `DBreezeCoinView` so that DBreezeCoinView uses `RewindDataIndexStore` to create and manage rewind data and its index?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
fenix2222: So instead of clearing all items we always want to keep last 500 items?

MithrilMan:
Just a note about DBreeze (I could open a generic Issue) I don't think it's right to pass/access dbreeze instance, we should instead have a IDBreezeTransactionFactory as a service, so we use that to create transactions when we need it

e.g. in RewindDataIndexRepository.cs
`using (DBreeze.Transactions.Transaction transaction = this.DBreeze.GetTransaction())`

could be
`using (DBreeze.Transactions.Transaction transaction = IDbreezeTransactionFactory.Create())`

passing DBreeze allow to mess up with DBreeze settings, folder, etc... and we don't want it to be exposed I'd say.

fenix2222:
@MithrilMan I totally agree about factory pattern. I am just not sure if it is in scope for this PR and coinview will be refactored so we want to keep as less changes as possible. We discussed this at the standup and I already raised a concern about the current implementation being dirty.

MithrilMan:
ok fine, I had problem with speakers when it started :)

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/IRewindDataIndexRepository.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    public interface IRewindDataIndexRepository : IDisposable
+    {
+        /// <summary>
+        /// Persists key value pair for the rewind data index item.
+        /// </summary>
+        /// <param name="key">Transaction id + N (N is an index of output in a transaction).</param>
+        /// <param name="rewindDataIndex">Index of the rewind data index.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        Task PutAsync(string key, int rewindDataIndex, CancellationToken cancellationToken = default(CancellationToken));
```
MithrilMan: as pointed out in the implementation, this seems not used

src/Stratis.Bitcoin.Features.Consensus/CoinViews/InMemoryCoinView.cs
```diff
@@ -98,5 +98,10 @@ public Task<uint256> RewindAsync()
         {
             throw new NotImplementedException();
         }
+
+        public Task<RewindData> GetRewindData(int height)
+        {
+            throw new NotImplementedException();
```
MithrilMan: Is this expected?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
```diff
@@ -40,6 +37,9 @@ public class DBreezeCoinView : ICoinView, IDisposable
 
         private BackendPerformanceSnapshot latestPerformanceSnapShot;
 
+        /// <summary>Access to DBreeze database.</summary>
+        public readonly DBreezeEngine DBreeze;
```
MithrilMan: Instead of exposing this as a public property, I'd create in this class a CreateTransaction method that return a DBreeze Transaction to be used outside. In the future we'd want to use a ITransactionFactory service to inject into constructor of classes that needs a DBreeze Transaction. (copy&pasted observation done where we uses this exposed member)

MithrilMan: (RewindDataIndexRepository.cs)

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Initializes a new instance of the object.
+        /// </summary>
+        /// <param name="dBreezeCoinView"><see cref="DBreezeCoinView"/>.</param>
+        /// <param name="loggerFactory">Factory to create a logger for this type.</param>
+        public RewindDataIndexRepository(DBreezeCoinView dBreezeCoinView, ILoggerFactory loggerFactory)
+        {
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+
+            this.dbreeze = dBreezeCoinView.DBreeze;
```
MithrilMan: Instead of exposing this as a public property, I'd create in this class a CreateTransaction method that return a DBreeze Transaction to be used outside. In the future we'd want to use a ITransactionFactory service to inject into constructor of classes that needs a DBreeze Transaction.

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        /// <inheritdoc />
+        public Task<int?> GetAsync(string key, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Task<int?> task = Task.Run(() =>
+            {
+                using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
```
MithrilMan: as said above, probably would be better to use something like `dBreezeCoinView.CreateDBreezeTransaction()` so we don't expose DBreeze engine (we should anyway move to a IDBreezeTransactionFactory service)

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -268,6 +274,33 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+        /// Checks if coinstake is spent on another chain.
+        /// </summary>
+        /// <param name="header">The proven block header.</param>
+        /// <exception cref="UtxoNotFoundInRewindDataException">
+        /// Throws specific exception so that it can be handled later.
+        /// </exception>
+        private void CheckIfCoinstakeIsSpentOnAnotherChain(ProvenBlockHeader header)
+        {
+            Transaction coinstake = header.Coinstake;
+            TxIn intput = coinstake.Inputs[0];
```
MithrilMan: typo (intput => input)

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
MithrilMan: yes declare 
`var indexItems = new Dictionary<string, int>();`
as
`var indexItems = new Dictionary<string, uint>();`
and you can just add unspent.Height without casting it

MithrilMan: of course this affect all the other methods that are expecting a Dictionary<string, int> because now would be Dictionary<string, uint>
Nonetheless I'd use the same type for represents the same things along our codebase and actually we have a mix of int/uint, @dangershony what should we use ?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -268,6 +274,33 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+
+            int? rewindDataIndex = this.PosParent.RewindDataIndexStore.GetAsync(intput.PrevOut.Hash, (int)intput.PrevOut.N).GetAwaiter().GetResult();
```
MithrilMan: I'm not sure because I've to re-read the requirement, but it seems to me that we have to check in all RewindData we stored between new chain tip and fork point, so we have to fetch from RewindDataIndexStore an index to the RewindData next of the fork point  between the new chain and the current chain
Using the coinstake transaction of fork point as TX+N we should get the index of the RewindData and we need to check that from that RewindData to the last(?) our utxo isn't spent.

I may be wrong tho, can anyone confirm?

src/Stratis.Bitcoin.Features.Consensus/CoinViews/InMemoryCoinView.cs
fenix2222: Yes. This is not relevant for InMemoryCoinView. The whole coin view concept needs to be redesigned in the future.

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
```diff
@@ -0,0 +1,125 @@
...
+
+                    return (int?)null;
+                }
+            }, cancellationToken);
+
+            return task;
+        }
+
+        /// <inheritdoc />
+        public Task PutAsync(string key, int rewindDataIndex, CancellationToken cancellationToken = default(CancellationToken))
```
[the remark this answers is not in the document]
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            Task task = Task.Run(() =>
+            {
+                this.logger.LogTrace("Storing rewind data index ({items}.Count():{count})", nameof(items), items.Count);
+
+                using (DBreeze.Transactions.Transaction transaction = this.dbreeze.GetTransaction())
+                {
+                    transaction.SynchronizeTables(RewindDataIndexTable);
+
+                    foreach (KeyValuePair<string, int> item in items)
```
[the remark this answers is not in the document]
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/IRewindDataIndexRepository.cs
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Consensus/CoinViews/DBreezeCoinView.cs
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexRepository.cs
fenix2222: Fixed

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
fenix2222: @dangershony Can you have a look, please. This is important.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -427,17 +459,17 @@ public async Task<uint256> RewindAsync()
             {
                 // Check if rewind data is available in local cache. If it is
                 // we can rewind and there is no need to check underlying storage.
-                if (this.cachedRewindDataList.Count > 0)
+                if (this.cachedRewindDataIndex.Count > 0)
                 {
-                    RewindData lastRewindData = this.cachedRewindDataList.Last();
+                    RewindData lastRewindData = this.cachedRewindDataIndex.Last().Value;
```
dangershony: improve this by taking `var item = cachedRewindDataIndex.Last()` then `item.Value` and `item.Key` small improvment.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -299,14 +316,14 @@ public async Task FlushAsync(bool force = true)
                     u.Value.ExistInInner = true;
                 }
 
-                await this.Inner.SaveChangesAsync(unspent.Select(u => u.Value.UnspentOutputs).ToArray(), null, this.innerBlockHash, this.blockHash, this.cachedRewindDataList).ConfigureAwait(false);
+                await this.Inner.SaveChangesAsync(unspent.Select(u => u.Value.UnspentOutputs).ToArray(), null, this.innerBlockHash, this.blockHash, 0, this.cachedRewindDataIndex.Select(c => c.Value).ToList()).ConfigureAwait(false);
```
dangershony: Why not use the height in cachedRewindDataIndex?

MithrilMan:
We'll need to double check the logics, but we can fix them later if needed, for now it's fine to merge this

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            using (new StopwatchDisposable(o => this.performanceCounter.AddInsertTime(o)))
+            {
+                foreach (KeyValuePair<string, int> indexRecord in indexData)
+                {
+                    this.items[indexRecord.Key] = indexRecord.Value;
+                }
+
+                // Save the items to disk.
+                await this.rewindDataIndexRepository.PutAsync(indexData).ConfigureAwait(false);
```
dangershony: FYI I don think we need to push to disk on save only on flush, @MithrilMan @fenix2222  please fix this in the other PR

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -268,6 +272,34 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+        /// <exception cref="ConsensusException">Throws utxo not found error.</exception>
+        private void CheckIfCoinstakeIsSpentOnAnotherChain(ProvenBlockHeader header, PosRuleContext context)
+        {
+            Transaction coinstake = header.Coinstake;
+            TxIn input = coinstake.Inputs[0];
+
+            int? rewindDataIndex = this.PosParent.RewindDataIndexStore.GetAsync(input.PrevOut.Hash, (int)input.PrevOut.N).GetAwaiter().GetResult();
+            if (!rewindDataIndex.HasValue) {
+                // Not sure if we should throw this exception or another.
+                ConsensusErrors.ReadTxPrevFailed.Throw();
```
dangershony: here you also need to `SetFlagAndThrow` if we cant find a rewind maybe the utxo is not in the chain yet (if we are very much ahead of consensus

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -268,6 +272,34 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+            }
+
+            RewindData rewindData = this.PosParent.UtxoSet.GetRewindData(rewindDataIndex.Value).GetAwaiter().GetResult();
+            UnspentOutputs matchingUnspentUtxo =
+                rewindData.OutputsToRestore.FirstOrDefault(unspent => unspent.TransactionId == input.PrevOut.Hash);
+            if (matchingUnspentUtxo == null)
```
dangershony: this is actually problematic if we find the rewind data then we must also find the UTXO this is a bug if `matchingUnspentUtxo == null`

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -972,6 +972,8 @@ private List<ChainedHeader> CreateNewHeaders(List<BlockHeader> headers)
             var newChainedHeaders = new List<ChainedHeader>();
 
             ChainedHeader newChainedHeader = this.CreateAndValidateNewChainedHeader(headers[newHeaderIndex], previousChainedHeader);
+            if (newChainedHeader == null) return newChainedHeaders;
```
dangershony: Trace logs are needed here no?

Also better just use the bottom of the method as exit point so put an if here I would suggest

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -1023,6 +1027,11 @@ private ChainedHeader CreateAndValidateNewChainedHeader(BlockHeader currentBlock
 
             if (result.Error != null)
             {
+                if (result.MissingUtxoInformation)
+                {
+                    return null;
```
dangershony: Trace

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -268,6 +272,34 @@ private void CheckHeaderSignatureWithCoinstakeKernel(ProvenBlockHeader header, U
...
+
+            int? rewindDataIndex = this.PosParent.RewindDataIndexStore.Get(input.PrevOut.Hash, (int)input.PrevOut.N);
+            if (!rewindDataIndex.HasValue)
+            {
+                context.ValidationContext.SetFlagAndThrow(ConsensusErrors.ReadTxPrevFailed, c => c.InsufficientHeaderInformation = true);
+            }
+
+            RewindData rewindData = this.PosParent.UtxoSet.GetRewindData(rewindDataIndex.Value).GetAwaiter().GetResult();
+            UnspentOutputs matchingUnspentUtxo =
+                rewindData.OutputsToRestore.FirstOrDefault(unspent => unspent.TransactionId == input.PrevOut.Hash);
```
zeptin: Just wondering - it is possible that there is more than one UTXO from the same transaction in `OutputsToRestore`. We should match on the index as well.

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -32,57 +38,80 @@ public class RewindDataIndexStore : IRewindDataIndexStore
...
+            this.items = new ConcurrentDictionary<string, int>();
+            this.coinView = coinView;
+            this.chainedHeaderTree = chainedHeaderTree;
 
             this.performanceCounter = new BackendPerformanceCounter(dateTimeProvider);
         }
 
+        public async Task InitializeAsync()
+        {
+            ChainedHeader tip = this.chainedHeaderTree.GetBestPeerTip();
```
MithrilMan: we don't have peers connected when you initialize, did we?
shouldn't we get blockstore tip as reference?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -32,57 +38,80 @@ public class RewindDataIndexStore : IRewindDataIndexStore
...
+            int heightToSyncTo = tip.Height > NumberOfItemsToKeep ? tip.Height - NumberOfItemsToKeep : 0;
```
MithrilMan: NumberOfItemsToKeep is a const set to 500, we should use MaxReorg consensus property

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
MithrilMan: we don't have peers connected when you initialize, did we?
shouldn't we get blockstore tip as reference?

zeptin: This might have been meant as `GetConsensusTip`. But I think the block store tip is preferable, as on startup consensus will be unwound to the block store height anyway if it consensus was higher.

MithrilMan: yes actually I see GetConsensusTip is private while GetBestPeerTip look every peer except for current peer (our fullnode)

Anyway I saw that when we call InitializeAsync we have already the chain tip, so we can just pass that

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
MithrilMan: agree, RewindDataIndex is the utxo we have in RewindData, so this would mean we have a store not synched (or maybe race conditions, can be possible?)
What to do then? we'd need to invalidate the cache? or leave the header stuck? or throw and stop(not ideal because it's not a peer fault but our data not synched)

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
MithrilMan: agre

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -12,53 +12,43 @@ namespace Stratis.Bitcoin.Features.Consensus.ProvenBlockHeaders
...
         }
 
-        public async Task InitializeAsync()
+        /// <summary>
+        /// <inheritdoc />
+        public async Task InitializeAsync(IConsensus consensusParameters, ChainedHeader tip, ICoinView coinView)
         {
-            ChainedHeader tip = this.chainedHeaderTree.GetBestPeerTip();
-            int heightToSyncTo = tip.Height > NumberOfItemsToKeep ? tip.Height - NumberOfItemsToKeep : 0;
+            this.numberOfBlocksToKeep = (int)consensusParameters.MaxReorgLength;
```
zeptin: I am not entirely convinced we want to couple these together. If the rewind data size for the network starts getting very large (i.e. blocks are getting very full) you may not want to keep the entire cache in memory any longer, given that reorgs up to the maximum possible length are hopefully rare.

But perhaps there are good reasons to do it this way too.

dangershony: Well no, its essential to keep the number of max reorg because we may get branches that spend a reorged item at position CT- 499 and that may ultimately become the longest chain so we need to validate it as well.

Its rare yes but still possible

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
dangershony: correct

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
dangershony: yeah this should be `GetConsensusTip`

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    this.items[indexRecord.Key] = indexRecord.Value;
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public void Flush()
+        {
+            int maxStoredHeight = this.items.Select(i => i.Value).DefaultIfEmpty(0).Max();
```
dangershony: too wasteful lets keep an integer representing the height of the last pushed items.
Also do we need to consider what we do in case of a regorg?

fenix2222: Maybe we should instead pass current tip into the flush and remove all cache items where height is either greater than the tip or less then tip - 500?

fenix2222: Updated logic, this should cater for reorg

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
fenix2222: So what are we doing with this one?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
dangershony: yeah good idea about passing the height

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -73,14 +73,11 @@ public void Save(Dictionary<string, int> indexData)
...
+        public void Flush(int currentHeight)
         {
-            int maxStoredHeight = this.items.Select(i => i.Value).DefaultIfEmpty(0).Max();
-            int heightToKeepItemsTo = maxStoredHeight - this.numberOfBlocksToKeep;
+            int heightToKeepItemsTo = currentHeight - this.numberOfBlocksToKeep;
 
-            if (heightToKeepItemsTo <= 0) return;
-
-            List<KeyValuePair<string, int>> itemsToRemove = this.items.Where(i => i.Value < heightToKeepItemsTo).ToList();
+            List<KeyValuePair<string, int>> itemsToRemove = this.items.Where(i => i.Value < heightToKeepItemsTo || i.Value > currentHeight).ToList();
```
dangershony: You are iterating over the collection twice no need for that.

src/Stratis.Bitcoin.Features.Consensus/CoinViews/CachedCoinView.cs
```diff
@@ -427,17 +466,17 @@ public async Task<uint256> RewindAsync()
...
+                    RewindData lastRewindData = this.cachedRewindDataIndex.Last().Value;
 
                     this.RemoveTransactions(lastRewindData);
                     this.RestoreOutputs(lastRewindData);
 
                     // Change current block hash to the one from the rewind data.
                     this.blockHash = lastRewindData.PreviousBlockHash;
 
-                    this.cachedRewindDataList.RemoveAt(this.cachedRewindDataList.Count - 1);
+                    this.cachedRewindDataIndex.Remove(this.cachedRewindDataIndex.Keys.Last());
```
dangershony: We probably nee dot add `this.blockHeight--`

PR #2670, opened 2018-11-05
[SC] Method param string deserialization should expect base58 encoded Address

### LINKED ISSUE #2659: [SC] Support Base58 Addresses in Wallet method params
Convert them to hex before sending across the wire.

rowandh:
Closes #2659.

### DIFF
--- a/src/Stratis.SmartContracts.Tools.Sct/ValidatorService.cs
+++ b/src/Stratis.SmartContracts.Tools.Sct/ValidatorService.cs
@@ -55,7 +55,8 @@ namespace Stratis.SmartContracts.Tools.Sct
 
             Assembly smartContract = Assembly.Load(compilation);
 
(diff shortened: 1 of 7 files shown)


src/Stratis.SmartContracts.Tools.Sct/ValidatorService.cs
```diff
@@ -55,7 +55,8 @@ private static void ValidateContract(string fileName, IConsole console, string[]
 
             Assembly smartContract = Assembly.Load(compilation);
 
-            var serializer = new MethodParameterStringSerializer();
+            // Network does not matter here as we are only checking the deserialized Types of the params.
+            var serializer = new MethodParameterStringSerializer(new SmartContractsRegTest());
```
codingupastorm: Could use null then?

PR #2681, opened 2018-11-05
Consensus manager reorg bug

dangershony:
When we handle ChanedHeaders we should assume that at any given moment a peer may get disconnected and the chain will be removed form the tree.

**Unhandled yet**
The problematic scenario is when we are in the middle of a big reorg and the only peer that claims this branch gets disconnected.
For example a reorg of 10 blocks and at block 5 the only peer claiming this branch is disconnected we will remove the entire branch of the reorg from the tree.

Two ways to fix it
1. We observe the branch was disconnected and try to reconnect
2. We place a temporary claim on a branch that is about to be connected, then this branch will not get removed on disconnect.

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
+++ b/src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
@@ -322,7 +322,7 @@ namespace Stratis.Bitcoin.Consensus
         {
             fullValidationRequired = false;
 
-            if (!chainedHeader.Previous.Next.Contains(chainedHeader))
+            if (!chainedHeader.IsConnected)
             {
                 this.logger.LogTrace("(-)[HEADER_DISCONNECTED]:false");
                 return null;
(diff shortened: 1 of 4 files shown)


dangershony:
Ok after further reviewing the code I think this fix is enough.

Observing the method `PartialValidationSucceeded` if the header has more work then our current tip we already claim that tip by calling ` this.ClaimPeerTip(LocalPeerId, chainedHeader.HashBlock);` 
This means even if the peer disconencts the ChainedHeader will remain in the tree, if validation failes we just remove that subtree and the claimed tip along with it.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -322,6 +322,12 @@ public List<ChainedHeaderBlock> PartialValidationSucceeded(ChainedHeader chained
         {
             fullValidationRequired = false;
 
+            if (!chainedHeader.Previous.Next.Contains(chainedHeader))
+            {
+                this.logger.LogTrace("(-)[HEADER_DISCONNECTED]:false");
```
dangershony:
```suggestion
                this.logger.LogTrace("(-)[HEADER_DISCONNECTED]:null");
```

PR #2683, opened 2018-11-06
[SC] SmartContractWallet History

codingupastorm:


src/Stratis.StratisSmartContractsD/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "Stratis.StratisSmartContractsD": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "commandLineArgs": "-connect=52.189.213.175"
```
rowandh: Remove

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/ContractAccountHistoryModel.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
```
rowandh: Looks this this file and some others are duplicated between the two wallet projects. Was this intentional?

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/ContractWalletHistoryModel.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public class ContractWalletHistoryModel
+    {
+        public ContractWalletHistoryModel()
+        {
+            this.AccountsHistoryModel = new List<ContractAccountHistoryModel>();
+        }
+
+        [JsonProperty(PropertyName = "history")]
+        public ICollection<ContractAccountHistoryModel> AccountsHistoryModel { get; set; }
```
rowandh: Each wallet history request specifies the account name so I assume this should only ever have one item? Feels like unnecessary nesting if so.

src/Stratis.Bitcoin.Features.SmartContracts.Wallet/ContractAccountHistoryModel.cs
codingupastorm: They're nearly identical but with a different enum I believe is the only change. This little change forced all the concrete classes to change

PR #2688, opened 2018-11-06
Ensure we generate always a PH for signaled blocks (#2662 part 3)

MithrilMan:
Not completed yet.

Things to discuss are: 

- when we create a PH from a legacy Header, what's the best way to propagate it back to the CHT?

see
```
// setters aren't accessible, not sure setting them to public is a nice idea.
//chainedHeaderBlock.Block.Header = chainedHeaderBlock.ChainedHeader.Header = createdProvenHeader;
```
Also I'm not sure the value is retained and used even if those properties were available

- Is it safe to assume that PH at the same Hash are overwritten? And what happen if we switch chain and we need to invalidate some height? Did we leave them there waiting to be overwritten over time or should we remove them?


src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
```diff
@@ -33,77 +36,68 @@ public class ProvenHeadersBlockStoreSignaled : BlockStoreSignaled
...
+                    if (provenHeaderHash == signaledHeaderHash)
                     {
-                        logger.LogTrace("Proven Header {0} found.", blockHash);
-                        return;
+                        logger.LogTrace("Proven Header {0} found.", signaledHeaderHash);
                     }
                     else
                     {
-                        throw new BlockStoreException("Found a proven header with a different hash.");
+                        logger.LogTrace("Found a proven header with a different hash, recreating PH. Expected Hash: {0}, found Hash: {1}.", signaledHeaderHash, provenHeaderHash);
```
dangershony: This means a reorg happened.

src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
```diff
@@ -33,77 +36,68 @@ public class ProvenHeadersBlockStoreSignaled : BlockStoreSignaled
...
 
             uint256 provenHeaderHash = newProvenHeader.GetHash();
             this.provenBlockHeaderStore.AddToPendingBatch(newProvenHeader, new HashHeightPair(provenHeaderHash, blockHeight));
 
-            logger.LogTrace("Created Proven Header at height {0} with hash {1} and adding to the store (pending).", blockHeight, provenHeaderHash);
+            logger.LogTrace("Created Proven Header at height {0} with hash {1} and adding to the pending batch to be stored.", blockHeight, provenHeaderHash);
 
-            return newProvenHeader;
+            // setters aren't accessible, not sure setting them to public is a nice idea.
+            //chainedHeaderBlock.Block.Header = chainedHeaderBlock.ChainedHeader.Header = createdProvenHeader;
```
dangershony: I suggest `chainedHeaderBlock.ChainedHeader.SetHeader(newheader)`

MithrilMan:
Let reference the issue: #2662

src/Stratis.Bitcoin/Primitives/ChainedHeaderBlock.cs
```diff
@@ -32,5 +33,17 @@ public override string ToString()
...
+        /// <summary>
+        /// Update the <see cref="ChainedHeader" /> (if not null) with a new provided header.
+        /// </summary>
+        /// <param name="newHeader">The new header to set.</param>
+        /// <remarks>Use this method very carefully because it could cause race conditions if used at the wrong moment.</remarks>
+        public void SetHeader(BlockHeader newHeader)
+        {
+            Guard.NotNull(newHeader, nameof(newHeader));
+
+            this.ChainedHeader?.SetHeader(newHeader);
```
dangershony: This is a problematic method because we dont know for sure if header was set, we are better of doing `chainedHeaderBlock.ChainedHeader.SetBlock()`

dangershony: Ah actually ChainedHeader can never be null

MithrilMan: that method is inside ChainedHeaderBlock and just ensure that ChainedHeader is set, with `?.` preventing to invoke SetBlock if ChainedHeader is null (even if this should never happen, right?)

maybe I don't understand, why do you want to call `SetBlock` if we are setting an header, typo?

Issue #2676, opened 2018-11-05
Wallet is not reorging correctly


ArcSin2000X:
**Description:** 
When running internal testnet we have NodeFD not being able to advance Consensus, Blockstore and Wallet to the tip of the Header, node will never go above height 1845.
Node wallet hash becomes `0000000000000000000000000000000000000000000000000000000000000000`

`[2018-11-03 17:01:28.3594 21] INFO: Stratis.Bitcoin.Consensus.ConsensusManager+<RewindToForkPointAsync>d__50.MoveNext Reorg from block '1845-29265ba908294d9b6f88857147e910fe159db44fffd16dfd3dbfe5bb4967f7f2-FullyValidated' to '1839-0100b52070436df710ce133a86f73c6fdba67c955fd7fdbe73506220f0a9360c-FullyValidated'
[2018-11-03 17:01:28.3891 22] INFO: Stratis.Bitcoin.Features.Miner.Staking.PosMinting+<CreateCoinstakeAsync>d__43.MoveNext Node staking with 3293794.15820313 (120.27 % of the network weight 2738675.44907665), est. time to find new block is 00:00:53.
[2018-11-03 17:01:28.9692 27] INFO: Stratis.Bitcoin.FullNode.<StartPeriodicLog>b__76_0 ======Node stats====== 11/03/2018 17:01:28
Headers.Height:      1846     Headers.Hash:     1fa1c6e450d648461e2682b3318c7d9513f5003c25e8dd6ae1c96f6c3eacc912
Consensus.Height:    1839     Consensus.Hash:   0100b52070436df710ce133a86f73c6fdba67c955fd7fdbe73506220f0a9360c
BlockStore.Height:   1845     BlockStore.Hash:  29265ba908294d9b6f88857147e910fe159db44fffd16dfd3dbfe5bb4967f7f2
Wallet.Height:       1845     Wallet.Hash:      0000000000000000000000000000000000000000000000000000000000000000`

if I try and find all where search string == `Wallet.Height:` i get
`Wallet.Height:       1841     Wallet.Hash:      f320de68748d6fad97fda602f889780366620c639454df79ad3635149f16f483
Wallet.Height:       1841     Wallet.Hash:      f320de68748d6fad97fda602f889780366620c639454df79ad3635149f16f483
Wallet.Height:       1841     Wallet.Hash:      f320de68748d6fad97fda602f889780366620c639454df79ad3635149f16f483
[...]

PR #2700, opened 2018-11-07
Add Fody weaving to feature projects

### LINKED ISSUE #2697: [Logging] Fody does not weave the features
During testing it has become apparent that the feature projects (e.g. mempool, wallet) do not have entry/exit trace logs.

`FodyWeavers.xml` seems to imply that they should:

`<TraceOn namespace="Stratis.Bitcoin+*" class="private" method="private" />`

Upon further experimentation, the `Tracer.Fody` nuget package needs to be added to every csproj that we want to be weaved. I also added a reference to the `FodyNlogAdapter` project to the feature project I was testing, but I am not 100% sure if both are required.

Therefore, someone needs to go through all the feature projects and add these references so that we get adequate trace logging for them.

zeptin:
Fixes #2697 

Fody all the things


zeptin:
@maciejzaleski @ArcSin2000X Note that this PR may substantially increase the size of the trace logs getting generated.

dangershony:
Do we need to add logs on all these projects?
@bokobza @fassadlr 

Perhaps yes, we do.

zeptin:
At the very least we need wallet logs to determine whether the wallet reorg bug is related to the node reorg bug or not.

My feeling is that we removed entry/exit logs everywhere under the expectation that Fody would provide them going forwards, so we should add it to all the features. If there are particularly noisy method calls (and there will be, almost certainly) we can `[NoTrace]` them.

dangershony:
ok perhaps its true, FYI what wallet reorg bug?

zeptin:
@dangershony #2676

bokobza:
From here, it might be preferable if the SC team did it themselves: https://github.com/stratisproject/StratisBitcoinFullNode/pull/2250#discussion_r220107454

So maybe remove Fody from the SC feature.
Also, what about entry/exit logs? Are they all gone from the features? Also, did you add [NoTrace] to the methods that are called too often?

zeptin:
Many of the consensus rules should probably also be squelched but this PR is getting large enough as it is.

bokobza:
I'm happy with the PR so far, but I'm just afraid the quantity of logs is going to be really huge if we don't really do all the methods that are called often.
Maybe just concentrate on the feature you need to see logs for first?

Issue #2474, opened 2018-10-12
[Mempool] Transaction time checks


zeptin:
The mempool validator does not appear to currently validate the `.Time` property of submitted transactions.

This is a check done by stratisX on mempool addition:

```
inline bool IsDriftReduced(int64_t nTime) { return TestNet() || nTime > 1479513600; } // Drifting Bug Fix, hardfork on Sat, 19 Nov 2016 00:00:00 GMT

inline int64_t TestingDrift(int64_t nTime) { return nTime + 128 * 60 * 60; }
inline int64_t MainNetDrift(int64_t nTime) { return nTime + 15; }

inline int64_t FutureDriftV1(int64_t nTime) { return nTime + 10 * 60; }
inline int64_t FutureDriftV2(int64_t nTime) {
    return IsDriftReduced(nTime) ? MainNetDrift(nTime) : TestingDrift(nTime);
}
inline int64_t FutureDrift(int64_t nTime, int nHeight) { return IsProtocolV2(nHeight) ? FutureDriftV2(nTime) : FutureDriftV1(nTime); }

...

    if (tx.nTime > FutureDrift(GetAdjustedTime(), nBestHeight + 1)) {
        reason = "time-too-new";
        return false;
    }
```

It would appear that, to get the same behaviour from SBFN, transactions that have a time greater than the most recent block's time + 15 seconds should be rejected from the mempool.

The following code exists in the rule `PosCoinstakeRule`:

```
            foreach (Transaction transaction in block.Transactions)
            {
                // Check transaction timestamp.
                if (block.Header.Time < transaction.Time)
                {
                    this.Logger.LogTrace("Block contains transaction with timestamp {0}, which is greater than block's timestamp {1}.", transaction.Time, block.Header.Time);
                    this.Logger.LogTrace("(-)[TX_TIME_MISMATCH]");
                    ConsensusErrors.BlockTimeBeforeTrx.Throw();
                }
            }
```

However, `PosCoinstakeRule` does not run for mempool transactions (since coinbase/coinstake transactions are immediately rejected from the mempool anyway).

The PoS-specific rule that gets run in the mempool validator is `CheckPosTransactionRule`. However, this rule does not perform the time check.

It may therefore make sense to move the time check logic from one rule to the other, as it does not really apply to coinstake transactions (it is unconditionally applied to all of them, which means the name of the rule is at best misleading).

There is also a TODO in `MempoolValidator` about potentially having mempool-specific rules.

[...]

PR #2701, opened 2018-11-07
Correct tx logic assigning RequireStandard mempool setting (#2474).

thefazzer:
Preferred: @bokobza


bokobza:
Let's wait for the build.

PR #2702, opened 2018-11-07
Added pagination and search to History 

bokobza:
* added pagination using optional `skip` and `take` parameters
  Example query: 
```
http://localhost:38221/api/Wallet/history?WalletName=wallet1&AccountName=account%200&Skip=10&Take=20
```

* added search using `q` parameter, that can search by transaction id, receiving address or send address
  Example query: 
```
http://localhost:38221/api/Wallet/history?WalletName=wallet1&AccountName=account%200&SearchQuery=3cf9c9783ecd4db4900f1f1241b8fe8075729abcb1bfb32c5300810c7adfa756
```


bokobza:
Linked to https://github.com/stratisproject/StratisCore/issues/100 and https://github.com/stratisproject/StratisCore/issues/96

bokobza:
not sure if @codingupastorm and @rowandh can benefit from this?

codingupastorm:
Thanks, not immediately, we had to make some changes to our history. Useful skeleton for future though.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -375,7 +375,8 @@ public IActionResult GetHistory([FromQuery] WalletHistoryRequest request)
                 {
                     var transactionItems = new List<TransactionItemModel>();
 
-                    List<FlatHistory> items = accountHistory.History.OrderByDescending(o => o.Transaction.CreationTime).Take(200).ToList();
+                    List<FlatHistory> items = accountHistory.History.OrderByDescending(o => o.Transaction.CreationTime).ToList();
```
fassadlr: @bokobza a bit more readable:

```
IEnumerable<FlatHistory> items = accountHistory.History.OrderByDescending(o => o.Transaction.CreationTime);
if (string.IsNullOrEmpty(request.SearchQuery))
    items = items.Take(200);
```

Also, no reason to `ToList()` unless the intention is to take a copy.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -107,6 +107,13 @@ public class WalletHistoryRequest : RequestModel
         public string WalletName { get; set; }
 
         public string AccountName { get; set; }
+
+        public int? Skip { get; set; }
```
fassadlr: Cant you just default this to 0 then you don't have to do a coalesce?

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -107,6 +107,13 @@ public class WalletHistoryRequest : RequestModel
...
+
+        public int? Take { get; set; }
+
+        [JsonProperty(PropertyName = "q")]
```
fassadlr: I would call this `query` rather?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -497,9 +498,16 @@ public IActionResult GetHistory([FromQuery] WalletHistoryRequest request)
                         }
                     }
 
+                    // Sort and filter the history items.
+                    List<TransactionItemModel> itemsToInclude = transactionItems.OrderByDescending(t => t.Timestamp)
+                        .Where(x => string.IsNullOrEmpty(request.SearchQuery) || (x.Id.ToString() == request.SearchQuery || x.ToAddress == request.SearchQuery || x.Payments.Any(p => p.DestinationAddress == request.SearchQuery)))
```
fassadlr: What if SearchQuery is null or empty?

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza:
```
items = string.IsNullOrEmpty(request.SearchQuery) ? items.Take(200).ToList() : items;`
```

 is as readable as 

```
if (string.IsNullOrEmpty(request.SearchQuery))
    items = items.Take(200);
```

I need ToList, the type I get from `OrderByDescending` is `IOrderedEnumerable<FlatHistory>`.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
bokobza: For Skip that wouldn't be a problem, but then how do I interpret `Take = 0`?
If Take is 0 I should normally return no items, or write some code that says that if it's 0 then it's means it's off.
To avoid all confusion, a nullable is the easiest way.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
bokobza: q is kind of an API convention for search queries. I think it's ok.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
bokobza: Ah, that's the beauty of this statement!
The goal is to avoid evaluating `x.Id.ToString() == request.SearchQuery` (for example) if SearchQuery is empty.
So:
if SearchQuery is null or empty then the evaluation stops as the statement returns `true`.
if SearchQuery is not null or empty, then the evaluation moves on to the rest of the statement.

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
dangershony: This is the value included in the URL?

PR #2705, opened 2018-11-08
Fetch stake utxo from context

dangershony:


src/Stratis.Bitcoin.Features.Consensus/StakeValidator.cs
```diff
@@ -184,12 +184,7 @@ public void CheckProofOfStake(PosRuleContext context, ChainedHeader prevChainedH
 
             TxIn txIn = transaction.Inputs[0];
 
-            // First try finding the previous transaction in database.
-            FetchCoinsResponse coins = this.coinView.FetchCoinsAsync(new[] { txIn.PrevOut.Hash }).GetAwaiter().GetResult();
-            if ((coins == null) || (coins.UnspentOutputs.Length != 1))
```
zeptin: Are you sure the null check can be removed though?

zeptin: Disregard this

bokobza:
Any background for this? @dangershony

dangershony:
Yeah it is not intended to fix the stake signature bug, I just stumbled upon this when looking at that bug.

Basically we are just acquiring locks not necessarily (and potentially in extreme cases even calling the disk when not needed)

zeptin:
Ah ok, that makes sense.

PR #2714, opened 2018-11-08
Fix wallet 000 hash

bokobza:
Fixes https://github.com/stratisproject/StratisBitcoinFullNode/issues/2676

While in reorg and printing the wallet height and hash to the console, the height of the wallet is taken from the wallet but the hash is taken from the chain.
Since we're in a reorg and the block at height X doesn't exist, the hash shows uint256.Zero.

This is fixed to take the hash from the wallet so that the height and the hash are always in sync, and correct values are shown when the reorg is done and the wallet has been updated.


bokobza:
Only commit 91ea33c is relevant.

PR #2715, opened 2018-11-08
Corrective fixes to Whitebind functionality

thefazzer:
@maciejzaleski Please review the following for functional correctness and style

Have interpreted the following where the first endpoint is a network I/F the node is listening on and the second arises from a user specified whitebind 
- 0.0.0.0 != 127.0.0.1
- 127.0.0.1:16178 == 0.0.0.0:16178
- 127.0.0.1:16178 != 0.0.0.0:44556
- 0.0.0.0 == 0.0.0.0


src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -229,18 +229,17 @@ private bool AllowClientConnection(TcpClient tcpClient)
                 return true;
             }
 
-            var clientRemoteEndPoint = tcpClient.Client.RemoteEndPoint as IPEndPoint;
+            var clientLocalEndPoint = tcpClient.Client.LocalEndPoint as IPEndPoint;
```
dangershony: This is a big change it seems, what the logic behind using `LocalEndPoint`?

src/Stratis.Bitcoin/Stratis.Bitcoin.csproj
```diff
@@ -52,6 +52,12 @@
     <ProjectReference Include="..\NBitcoin\NBitcoin.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="xunit.assert">
```
dangershony: we dont need testfreamwork dlls in core

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
maciejzaleski: As per the bitcoin specification on which I believe we are basing our `-whitebind` flag:
```Bind to given address and whitelist peers connecting to it. Use [host]:port notation for IPv6```

Since the option defines interface onto which the node is binding I had to correct the logic in the `AllowClientConnection` to assert if the connection should be "whitelisted" based on the local interface (we do not whitelist remote interfaces as this wasn't the intention of the flag).

Comment (dangershony) [reply src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs]: Can you point to the code in core where this happens?

File: src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs [hunk 3]
```diff
@@ -158,6 +172,59 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine($"#iprangefiltering=<0 or 1>");
         }
 
+        public bool EndpointIsRoutableToAnotherEndpoint(string whiteBindEndpointString, out NodeServerEndpoint localEndpoint)
```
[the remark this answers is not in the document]
maciejzaleski: This should be an extension method: `public static bool CanBeMappedTo(this NodeEndpoint base, out NodeEndpoint target)`

src/Stratis.Bitcoin.Tests/Connection/ConnectionManagerTest.cs
```diff
@@ -0,0 +1,82 @@
...
+using Stratis.Bitcoin.Configuration;
+using Stratis.Bitcoin.Configuration.Settings;
+using Stratis.Bitcoin.Connection;
+using Stratis.Bitcoin.Interfaces;
+using Stratis.Bitcoin.Tests.Common;
+using Stratis.Bitcoin.Tests.Common.Logging;
+using Stratis.Bitcoin.Utilities.JsonErrors;
+using Xunit;
+
+namespace Stratis.Bitcoin.Tests.Connection
```
maciejzaleski: Add tests for:
- IPv6 localhost and network address
  **Example:** `::1`, `fe80::d111:a4c4:ce4:2bc7%21`

- Any whitebind IP can be mapped to any localhost IPV4/IPV6 address (e.g. `127.0.0.0/8` and `::1/128`)
   `0.0.0.0:16178` == `127.0.0.1:16178`
   `0.0.0.0:16178` == `127.0.0.2:16178`
   `::/128` == `::1`

- If the address could be matched but the port is not matching the method should return false
   `0.0.0.0:16178` != `127.0.0.1:999`
   `::/128:16178` == `::1:16178`

- If we have bound to all interfaces but whitebind only whitelists one interface
   `127.0.0.1:16178` != `0.0.0.0:16178`
   `::1:16178` !=  `::/128:16178`

- Specify only IP address and use the default port

maciejzaleski: Tests are not covering the code from `ConnectionManagerSettings.cs`

maciejzaleski: Personally I would prefer a single test with a list of addresses to compare and expected results.

src/Stratis.Bitcoin/Configuration/Settings/ConnectionManagerSettings.cs
```diff
@@ -158,6 +172,59 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
...
+        {
+            // refactor this with lib code.
+            string[] ipAndPortSplit = whiteBindEndpointString.Split(new char[]{':'});
+            var whiteBindEndpoint = new IPEndPoint(IPAddress.Parse(ipAndPortSplit[0]), int.Parse(ipAndPortSplit[1]));
+            var nodeServerWhiteBindEndpoint = new NodeServerEndpoint(whiteBindEndpoint, true);
+            NodeServerEndpoint networkInterfaceEndpoint = null;
+
+            // 127.0.0.1:16178 == 0.0.0.0:16178
+            // 127.0.0.1:16178 != 0.0.0.0:44556
+            var localhostIp = IPAddress.Parse("127.0.0.1").MapToIPv6Ex();
```
maciejzaleski: There is no reason to use loopback address in here

src/Stratis.Bitcoin/P2P/Peer/NetworkPeerServer.cs
```diff
@@ -240,7 +240,7 @@ private bool AllowClientConnection(TcpClient tcpClient)
             }
 
             this.logger.LogTrace("Node '{0}' is not white listed during initial block download.", clientRemoteEndPoint);
-
+
```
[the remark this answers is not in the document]
maciejzaleski: The line
`NodeServerEndpoint endpoint = this.connectionManagerSettings.Listen.FirstOrDefault(e => e.Endpoint.Match(clientLocalEndPoint));`

Should use CanBeMappedTo method to compare addresses.

thefazzer:
**Superceded by:** https://github.com/stratisproject/StratisBitcoinFullNode/pull/2752
**Reason:** PR clean commit

PR #2718, opened 2018-11-09
[SC] Allow structs through private method param validator

### LINKED ISSUE #2711: [SC] I can't use struct params in my smart contract

codingupastorm:
Closes #2711

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
@@ -720,10 +720,18 @@ namespace Stratis.Bitcoin.Features.SmartContracts.Tests
                                                 public Test(ISmartContractState state)
                                                     : base(state) {}         
 
-                                                private void F(TestStruct test)
+                                                private void PrivateStruct(TestStruct test)
                                                 {
                                                 }
-                                                
+
+                                                public void PublicStruct(TestStruct test)
+                                                {
+                                                }
+
+                                                internal void InternalStruct(TestStruct test)
+                                                {
+                                                }
+
                                                 public struct TestStruct
                                                 {
                                                     public int TestInt;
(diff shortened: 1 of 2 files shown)


codingupastorm:

src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
```diff
@@ -709,6 +709,36 @@ public void F(float param)
...
+        {
+            string adjustedSource = @"using System;
+                                            using Stratis.SmartContracts;
+
+                                            public class Test : SmartContract
+                                            {
+                                                public Test(ISmartContractState state)
+                                                    : base(state) {}         
+
+                                                private void F(TestStruct test)
```
rowandh: Can we add public, private, and internal methods here too? And check that the error returned contains the public methodname.

src/Stratis.SmartContracts.Core.Validation/Validators/Method/MethodParamValidator.cs
```diff
@@ -66,7 +66,13 @@ public static (bool, string) IsValidParam(MethodDefinition methodDefinition, Par
                 return (true, null);
             }
 
-            var allowedType = AllowedTypes.Contains(param.ParameterType.FullName);
+            // In private methods we allow structs
+            if (!methodDefinition.IsPublic && param.ParameterType.IsValueType)
```
rowandh: We can allow all param types here if the method is not public.

codingupastorm: Decided against that - weird to have a private method that allows DateTime but is hence unusable?

src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
codingupastorm: 👍

src/Stratis.SmartContracts.Core.Validation/Validators/Method/MethodParamValidator.cs
rowandh: It will still fail validation if a method param is a disallowed Type, so that should be picked up before deployment.

Otherwise it would be possible to write a method that uses a Type which is not an allowed public method param Type, but _is_ allowed internally (eg. int[]). One consequence of this is that methods could then not be further broken down into smaller pieces.

PR #2721, opened 2018-11-09
[SC] Private method structs + arrays

codingupastorm:


src/Stratis.SmartContracts.Core.Validation/Validators/Method/MethodParamValidator.cs
```diff
@@ -27,8 +27,17 @@ public class MethodParamValidator : IMethodDefinitionValidator
...
 
+        public static readonly IEnumerable<string> PrivateAllowedTypes = PublicAllowedTypes.ToList().Concat(new string[]
+        {
+            typeof(Array).FullName
+        });
+
+
         public IEnumerable<ValidationResult> Validate(MethodDefinition methodDef)
         {
+            if (!methodDef.IsPublic)
```
rowandh: Forgot to remove this?

PR #2723, opened 2018-11-09
[PeerAddressManager] Add inbound connections to the address book

fassadlr:


src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
```diff
@@ -132,6 +132,12 @@ private Task OnStateChangedAsync(INetworkPeer peer, NetworkPeerState previousSta
                     this.peerAddressManager.PeerHandshaked(peer.PeerEndPoint, this.dateTimeProvider.GetUtcNow());
             }
 
+            if ((peer.Inbound) && (peer.State == NetworkPeerState.HandShaked) &&
+                (this.Mode == PeerAddressManagerBehaviourMode.Advertise || this.Mode == PeerAddressManagerBehaviourMode.AdvertiseDiscover))
+            {
+                this.peerAddressManager.AddPeer(peer.PeerEndPoint, IPAddress.Loopback);
```
fassadlr: Not sure about the Loopback @dangershony ?

dangershony: The problem here is with `peer.PeerEndPoint` this might not be correct this is the endpoint the socket is connected on so the port can be incorrect.

fassadlr: OK so PeerEndPoint here is set to:

```
client.Client.RemoteEndPoint
```

in

`public INetworkPeer CreateNetworkPeer(TcpClient client, NetworkPeerConnectionParameters parameters = null, NetworkPeerDisposer networkPeerDisposer = null)`

This is not correct?

fassadlr: @dangershony Im going to add some extra logging in key places so that we can see what the respective values are. Will post my findings here.

dangershony:
Yeah but we cannot assume what the peers listening port is, notice it connects inbound on port 1749 
Perhaps the port on the AddressFrom is the listening port?
In anycase we need store the PeerEndpoint but not sure with what listening port

dangershony:
This now looks more correct to me.

zeptin:
This looks good in theory, but note that there are situations where it will not work.

E.g. if we are behind a router that NATs incoming connections to its own IP address. This will result in the inbounds all having PeerEndPoint IP address `192.168.x.y`, say, with varying ports. The AddressFrom could be anything. These variations will then pollute the peer database and offer little utility to the node.

A somewhat contrived/uncommon scenario, but something to be aware of.

fassadlr:
@zeptin good point! Will the NAT not always assign 192.168.x.x ? I.e. in a range? We could check AddressFrom for intranet addresses like these and range filter them?

In other words if AddressFrom is a Loopback or falls in a 192.186 range then we use PeerEndpoint?

It might not be exact but at least we are filtering more unwanted addresses?

noescape00:
Not really related to this PR but PeerAddressManagerBehaviourMode enum is marked as [Flags], but the values for that enum are 0,1,2,3 and therefore it can't be used as flags. Also enum contains AdvertiseDiscover which is supposed to be combination for Advertise and Discover, so there is no need for that enum to be marked [Flags]- it already has combined states defined there without flags.

src/Stratis.Bitcoin/P2P/PeerAddressManagerBehaviour.cs
```diff
@@ -132,6 +133,30 @@ private Task OnStateChangedAsync(INetworkPeer peer, NetworkPeerState previousSta
...
+                IPEndPoint inboundPeerEndPoint = null;
+
+                // Use AddressFrom if it is not a Loopback address as this means the inbound node was configured with a different external endpoint.
+                if (!peer.PeerVersion.AddressFrom.Match(new IPEndPoint(IPAddress.Loopback, this.AttachedPeer.Network.DefaultPort)))
+                {
+                    inboundPeerEndPoint = peer.PeerVersion.AddressFrom;
+                }
+                else
+                {
+                    // If it is a Loopback address use PeerEndpoint but combine it with the AdressFrom's port as that is the
```
noescape00: In case port forwarding is enabled it's possible that internal and external ports are different so it won't work for such scenarios.

dangershony:
@noescape00 yes good spot, @fassadlr will remove the flags from that enum in another PR.
We also need to check that we ban based on IP only and not port and IP

PR #2724, opened 2018-11-09
[Consensus] Add state objects to CM and CHT for inspection

fassadlr:
Acquiring and/or inspecting the state of various ConsensusManager and ChainedHeaderTree properties/fields is extremely limited and makes unit tests basically impossible unless reflection is used,

This PR introduces a `ConsenusManagerState` and `ChainedHeaderTreeState` class that can easily be added to, without changing the access modifiers of internal or private fields of CM or CHT.

I don't see the point of using locks around these as they will mostly be used in tests for now. Should we want to use CHT properties outside of CHT we can look at how to lock them. @dangershony thoughts on this?


fassadlr:
@Neurosploit this can then be used in your PR #2620

Neurosploit:
Great I'll try to merge when this is in. Is that ok?

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
```diff
@@ -227,6 +232,19 @@ public sealed class ChainedHeaderTree : IChainedHeaderTree
             this.UnconsumedBlocksCount = 0;
         }
 
+        /// <inheritdoc />
+        public ChainedHeaderTreeState State()
+        {
+            var state = new ChainedHeaderTreeState
+            {
+                ChainedHeadersByHash = this.chainedHeadersByHash,
```
dangershony: We probably actually want to create copies of each collection no? (or the intention is to get the changes as they are reflected on the tree)

Also these methods should be internal so they will not get abused.

dangershony:
@Neurosploit if you agree with my comments you can modify this PR directly.

Neurosploit:
Thanks @dangershony @fassadlr  I'll try it out over the weekend, first i'll make it internal and then create copies, trying out a test merge with my branch on the way to see if I run into anything.

src/Stratis.Bitcoin/Consensus/ChainedHeaderTree.cs
fassadlr: @dangershony these methods can't be internal as they sit on the IConsensusManager interface ;) We need public access to them.

I will make copies of the collections though 👍

fassadlr:
@Neurosploit I've made the changes as suggested by @dangershony  👍

fassadlr:
@Neurosploit We will greatly benefit from having this on the interface as there are looooads of tests that can be amended to rather use this instead of getting values via reflection.

Neurosploit:
Ok, I've already finished my rework in a testbranch so once this is merged i'll merge my branch with master and reapply it to my PR.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -146,6 +146,22 @@ public class ConsensusManager : IConsensusManager
             nodeStats.RegisterStats(this.AddBenchStats, StatsType.Benchmark, 1000);
         }
 
+        /// <inheritdoc />
+        public ConsensusManagerState State()
+        {
```
dangershony: This needs to have the correct locks I would imagine, otherwise tests might throw.

noescape00:
For unit tests or integration tests you can create TestCHT which will have those structures exposed- in that case you won't even need reflection.

fassadlr:
@noescape00 please note that we are not exposing internal properties of CHT or CM. We are merely creating copies of them.

noescape00:
>@noescape00 please note that we are not exposing internal properties of CHT or CM. We are merely creating copies of them.

That's also exposure, just read-only one (except for chained headers- they are references, not copies). 
And very expensive one too if it's ever used on production code.

fassadlr:
@Neurosploit we have decided not to go this route. We don't want to put the `State` on the interface as it might be misued. @noescape00 will create a PR where we will be using an overridden ChainedHeaderTree class that exposes the private properties of CHT.

Neurosploit:
Alright just reference it to my issue and I will have a look again. @fassadlr @noescape00.

Issue #2518, opened 2018-10-16
Wallet doesn't ignore empty UTXOs


noescape00:
From wallet history: 

```
"type": "received",
         "toAddress": "PXh9UeETiVKMPzcCUeGZ6SCYLU6CEKS75V",
         "id": "acca764bda2e3b607b72d85ed7e0279891e59b7652a5b8b13b2d1f3db2a56f47",
         "amount": 0,
         "payments": [],
         "confirmedInBlock": 54,
         "timestamp": "1539688048"
```

Basically if 0 amount was sent to some addr the wallet will track this utxo.

PR #2726, opened 2018-11-09
Wallet: skip processing of transactions with 0 in output

bokobza:
Closes https://github.com/stratisproject/StratisBitcoinFullNode/issues/2518

This will leave out from processing in the wallet:
- poa mining transactions
- coinbase transactions in pos
- first coinstake output

where we know we're not gonna save anything to the wallet.
This will also improve performance a bit.


codingupastorm:
Could we add some tests for this case?

bokobza:
not..yet.
I did sync stratis testnet and my wallets and everything is behaving as expected though.
So that counts for something amirite?

PR #2728, opened 2018-11-09
PH backward compatibility on handshake

### LINKED ISSUE #2727: [P2P] sendprovhdr gets sent unconditionally during handshake
The default `ConnectionManagerBehavior` gets overridden with `ProvenHeadersConnectionManagerBehavior` when PoS is being used. This behaviour overrides the default functionality of the method `OnHandshakedAsync`, which would normally instruct the peer to preferentially send headers. The PH version instead tells the peer to send proven headers, which is not valid as the functionality is not active on the network yet.

MithrilMan:
Fixes #2727

### DIFF
--- a/src/Stratis.Bitcoin/Connection/ProvenHeadersConnectionManagerBehavior.cs
+++ b/src/Stratis.Bitcoin/Connection/ProvenHeadersConnectionManagerBehavior.cs
@@ -31,16 +31,24 @@ namespace Stratis.Bitcoin.Connection
         /// <inheritdoc />
         protected override async Task OnHandshakedAsync(INetworkPeer peer)
         {
-            // Require from height is the highest between activation height and last checkpoint height.
-            int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
-            int activationHeight = (this.network.Consensus.Options as PosConsensusOptions).ProvenHeadersActivationHeight;
+            if (this.CanPeerProcessProvenHeaders(peer))
+            {
+                // Require from height is the highest between activation height and last checkpoint height.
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin/Connection/ProvenHeadersConnectionManagerBehavior.cs
```diff
@@ -31,16 +31,24 @@ public ProvenHeadersConnectionManagerBehavior(IConnectionManager connectionManag
...
         protected override async Task OnHandshakedAsync(INetworkPeer peer)
         {
-            // Require from height is the highest between activation height and last checkpoint height.
-            int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
-            int activationHeight = (this.network.Consensus.Options as PosConsensusOptions).ProvenHeadersActivationHeight;
+            if (this.CanPeerProcessProvenHeaders(peer))
+            {
+                // Require from height is the highest between activation height and last checkpoint height.
+                int lastCheckpointHeight = this.checkpoints.GetLastCheckpointHeight();
+                int activationHeight = (this.network.Consensus.Options as PosConsensusOptions).ProvenHeadersActivationHeight;
```
fenix2222: I don't think `this.network.Consensus.Options` can ever be `null`. Use direct cast, not `as`, i.e. `(PosConsensusOptions )this.network.Consensus.Options`

dangershony: This will indeed go away soon but for now its correct yes, we should not send PH to nodes under that protocol version

PR #2735, opened 2018-11-11
[SC Integration Tests] [Dont Merge] Remove ClassFixtures

fassadlr:
Hey @codingupastorm @rowandh check out this PR.

I have removed the IClassFixtures from the tests that are constantly failing and implemented a disposable when each test exits so that the node builder and its nodes get shut down properly.

I noticed that when the class fixtures are used, we only start the nodes once and re-use them throughout ALL the tests. This to me can be problematic as in an ideal scenario we want the tests to run contained, i.e. start nodes, run test, close nodes, start test, run test, close nodes etc.

However, when I did this loads of tests started failing due to the wallet not having spendable balances, asserts not being right etc.

This now leads me to believe that perhaps the tests were asserting values etc with that in mind? I..e possibly picking up values from other tests?

Have a look for yourself and see what ya think :)

### DIFF
--- a/src/Stratis.SmartContracts.IntegrationTests/PoA/ContractExecutionFailureTests.cs
+++ b/src/Stratis.SmartContracts.IntegrationTests/PoA/ContractExecutionFailureTests.cs
@@ -18,18 +18,20 @@ using Xunit;
 
 namespace Stratis.SmartContracts.IntegrationTests.PoA
 {
-    public class ContractExecutionFailureTests : IClassFixture<PoAMockChainFixture>
+    public class ContractExecutionFailureTests : IDisposable
     {
         private readonly PoAMockChain mockChain;
+        private readonly PoAMockChainFixture mockChainFixture;
         private readonly MockChainNode node1;
         private readonly MockChainNode node2;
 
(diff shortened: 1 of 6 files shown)


codingupastorm:
I think the fixtures are actually more useful for simulating a real network - our isolated environment is:
- Here's a network with some history that doesn't matter
- This node sends a contract tx
- Let's see the result

In the real world transactions aren't always going to happen on block 11 right after the premine (which is what would be happening if we remove the fixtures).

On top of that it also makes the tests a lot faster.

I do want our tests to work in the CI but I'm not sure this is the fix...

fassadlr:
@codingupastorm OK I'm closing this so that we can investigate why the tests are failing.

rowandh:
@fassadlr - we've figured out the PoA tests are failing due to a race condition waiting for the tx to hit the mempool. @codingupastorm has created a PR which changes the miner so it behaves as expected. Thanks for your help with this :+1:

fassadlr:
@rowandh ah ok great! Which is PR is this? Or was it merged already. Sorry Ive been away this week 👍

rowandh:
Hasn't been merged yet, it's #2760.

PR #2737, opened 2018-11-11
[RPC] Properly handle optional parameters for getnewaddress

mikedennis:
There are some optional parameters we were ignoring in previous implementation. Throw exceptions when they are not well formed.

Bitcoin core docs https://bitcoincore.org/en/doc/0.16.0/rpc/wallet/getnewaddress/

Related to https://github.com/MetacoSA/NBitcoin/pull/413

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
@@ -127,11 +127,22 @@ namespace Stratis.Bitcoin.Features.Wallet
         /// RPC method that gets a new address for receiving payments.
         /// Uses the first wallet and account.
         /// </summary>
+        /// <param name="account">Parameter is deprecated.</param>
+        /// <param name="addressType">Address type, currently only 'legacy' is supported.</param>
         /// <returns>The new address.</returns>
         [ActionName("getnewaddress")]
         [ActionDescription("Returns a new wallet address for receiving payments.")]
-        public NewAddressModel GetNewAddress()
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -127,11 +127,22 @@ public async Task<uint256> SendTransactionAsync(string hex)
         /// RPC method that gets a new address for receiving payments.
         /// Uses the first wallet and account.
         /// </summary>
+        /// <param name="account">Parameter is deprecated.</param>
+        /// <param name="addressType">Address type, currently only 'legacy' is supported.</param>
         /// <returns>The new address.</returns>
         [ActionName("getnewaddress")]
         [ActionDescription("Returns a new wallet address for receiving payments.")]
-        public NewAddressModel GetNewAddress()
+        public NewAddressModel GetNewAddress(string account, string addressType)
```
dangershony: I wonder if we should we create an overload for this method?

bokobza:
The test Listmethods_via_rpc_returns_non_empty_list is failing.
Can you please have a look @mikedennis ?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: If it's called via RPC no need as the parameters are optional so don't need to be included. If we want to call directly, then maybe. The wrapper in RPCClient actually doesn't pass these parameters at all.

mikedennis:
@bokobza fixed - that one always gets me!

Issue #2583, opened 2018-10-24
[SC] Put in place a block gas limit


codingupastorm:
There's currently no limit on how much gas can be executed as part of a block.

Because of this, someone could submit many many transactions to a PoA node and it would be building that block for a long-time, potentially filling it all the way up to 1MB and taking many seconds.

To counter this, we should add in a limit on the total gas allowed into a block being built. 

This would happen in `SmartContractBlockDefinition`. After each execution we increment a counter for this block's gas counter and if it reaches `10 * SmartContractFormatRule.GasLimitMaximum` we will stop including transactions in the block.

codingupastorm:
Closed by #2739

PR #2739, opened 2018-11-12
[SC] Gas limit per block

kogot:
#2583 
Added GasLimit per block for Smart contracts. Once we've riched the limit no more transactions will be added to block.


codingupastorm:
Looks great so far @kogot, I messaged you about testing

src/Stratis.Bitcoin.Features.SmartContracts/PoW/SmartContractBlockDefinition.cs
```diff
@@ -83,7 +86,9 @@ public override void AddToBlock(TxMempoolEntry mempoolEntry)
             {
                 this.logger.LogTrace("Transaction contains smart contract information.");
 
-                // We HAVE to first execute the smart contract contained in the transaction
+                if (this.blockGasConsumed >= GasPerBlockLimit) return;
```
codingupastorm: This is the 1 part I'm cautious of, I think this is right, but just want to make sure in tests that it doesn't think the tx was included in one block or the other when it wasn't

kogot: @codingupastorm updated PR. Added integration test.

src/Stratis.SmartContracts.IntegrationTests/PoW/ContractExecutionFailureTests.cs
```diff
@@ -744,5 +746,40 @@ public void ContractTransaction_Call_Method_Consumes_All_Gas_Throws_OutOfGasExce
...
+            for (int i = 0; i < txCount; i++)
+            {
+                preResponseList.Add(this.node1.SendCreateContractTransaction(compilationResult.Compilation, 0, gasPrice:10000));
+            }
+            this.node1.WaitMempoolCount(txCount);
+            this.node1.MineBlocks(1);
+
+            NBitcoin.Block lastBlock = this.node1.GetLastBlock();
+
+            int expectedTxQty = Convert.ToInt32(txGasPerBlockLimit / txGasPrice) + 2;
```
codingupastorm: Where does this + 2 come from?

I think a more obvious way to test this would be to deploy the InfiniteLoop contract in one block - then for the repeated transaction do a `SendCallContractTransaction` that triggers the `Loop` method on the contract - this would consume all of the gas limit per tx always, which is probably more reliable than hardcoding the 12009, which may change if we update our gas costs. Does this make sense @kogot ?

kogot: @codingupastorm updated. please check.

PR #2741, opened 2018-11-12
Updated FN documentation

Neurosploit:
Went through the documentation and corrected some typos and made some sentences easier to read.

### DIFF
--- a/Documentation/coldstaking.md
+++ b/Documentation/coldstaking.md
@@ -3,31 +3,31 @@
 
 ## Motivation
 
-In order to produce blocks on Stratis network, a miner has to be online with running node and have its wallet open. This is necessary because at each time slot, the minor is supposed to check whether 1 of its UTXOs is eligible to be used as so-called coinstake kernel input and if so, it needs to use the private key associated with this UTXO in order to produce the coinstake transaction.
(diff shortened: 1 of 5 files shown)


Documentation/coldstaking.md
```diff
@@ -3,31 +3,31 @@
...
 
 
 ## User interface flow
 
-We want to implement cold staking in a way that allows even inexperienced users to participate in production of new blocks. Since all the logic is implemented on the node side and the wallet is only using interfaces provided by the node, it is sufficient to describe the flow from inexperienced users point of view. Advanced users can then easily use the underlying functionality of the node.
+We want to implement cold staking in a way that allows even inexperienced users to participate in production of new blocks. Since all the logic is implemented on the node side and the wallet is only using interfaces provided by the node, it is sufficient to describe the flow from an inexperienced users point of view. Advanced users can then easily use the underlying functionality of the node.
```
bokobza:
```suggestion
We want to implement cold staking in a way that allows even inexperienced users to participate in production of new blocks. Since all the logic is implemented on the node side and the wallet is only using interfaces provided by the node, it is sufficient to describe the flow from inexperienced users point of view. Advanced users can then easily use the underlying functionality of the node.
```

PR #2745, opened 2018-11-13
[PH] Activation of Proven Header (as a Feature)

MithrilMan:
I've redesigned Proven Headers as a feature, so it will be much easier to create a daemon to use PH and we still have a chance to use not use the PH at all on other networks.
Basically decoupling PH from PoS network (but PH still need PoS network to work)

This way we can even merge without breaking previous integration tests because actually they don't use the PH feature.

I've created a Stratis.StratisProvenHeadersD project that uses PH and we can use that to test PH
Note that the difference between StratisD is just a 
`.UseProvenHeaders(allowLegacyHeadersForWhitelistedPeers: true)`
after `.UsePosConsensus()`


dangershony:
this approach is no used for now

PR #2756, opened 2018-11-14
Oversize tx rejected by mempool integration test

codingupastorm:
Ay, was working on some related things and punched this out - figure any test coverage is useful so feel free to accept or decline.

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
@@ -447,5 +447,41 @@ namespace Stratis.Bitcoin.IntegrationTests.Mempool
                 Assert.Equal("time-too-new", entry.ErrorMessage);
             }
         }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.IntegrationTests/Mempool/MemoryPoolTests.cs
```diff
@@ -447,5 +447,41 @@ public void Mempool_SendPosTransaction_AheadOfFutureDrift_ShouldRejectByMempool(
                 Assert.Equal("time-too-new", entry.ErrorMessage);
             }
         }
+
+        [Fact]
+        public void Mempool_SendOversizeTransaction_ShouldRejectByMempool()
```
zeptin: Note that we do have a test of the mempool validator already that tests for this condition. Is it necessary to introduce an integration test for this?

See: `AcceptToMemoryPool_TxIsNonStandardTransactionSizeInvalid_ReturnsFalseAsync`

bokobza: I'll accept it anyway.

PR #2757, opened 2018-11-14
[SC] Lottery MockChain integration test, uses keccak

### LINKED ISSUE #2516: [SC] Integration test for some Keccak use cases

codingupastorm:
Closes #2516 for the meanwhile


src/Stratis.SmartContracts.IntegrationTests/PoW/SmartContractWalletTests.cs
```diff
@@ -437,6 +437,63 @@ public void MockChain_AuctionTest()
             }
         }
 
+        [Fact]
+        public void MockChain_Lottery()
```
rowandh: Do we need to add this to PoA tests too?

codingupastorm: Good point, I only did it here because it's easier - I'll add it to this PR after hopefully the PoA change is merged in

PR #2760, opened 2018-11-15
[SC] PoA Spacing DateTimeProvider

codingupastorm:
Allows the SC integration tests to run "synchronously" on PoA.

Should fix the CI also.

### DIFF
--- a/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
@@ -156,9 +156,6 @@ namespace Stratis.Bitcoin.Features.PoA
                     builder.AppendLine($"Block was mined {chainedHeader}.");
                     builder.AppendLine("<<==============================================================>>");
                     this.logger.LogInformation(builder.ToString());
-
-                    int halfTargetSpacingMs = (int)this.network.ConsensusOptions.TargetSpacingSeconds * 1000 / 2;
-                    await this.TaskDelayAsync(halfTargetSpacingMs, this.cancellation.Token).ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
                 {
--- a/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
+++ b/src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
@@ -176,12 +176,12 @@ namespace Stratis.Bitcoin.Features.PoA
         {
             int waitingTime = this.GetWaitingTimeInSeconds(myTimestamp);
 
+            this.logger.LogInformation("Waiting {0} seconds until block can be mined.", waitingTime);
+
             // Accounts for different IDateTimeProvider implementations. The standard implementation would progress
             // after one iteration, but for other cases where we are emulating time, this may loop more than once.
             while (waitingTime > 0)
             {
-                this.logger.LogInformation("Waiting {0} seconds until block can be mined.", waitingTime);
-
                 await this.TaskDelayAsync(waitingTime * 1000, this.cancellation.Token).ConfigureAwait(false);
 
                 waitingTime = this.GetWaitingTimeInSeconds(myTimestamp);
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -146,10 +146,16 @@ private async Task CreateBlocksAsync()
 
                     this.logger.LogInformation("Waiting {0} seconds until block can be mined.", waitingTimeInSeconds );
 
-                    if (waitingTimeInSeconds > 0)
+                    // This is a loop to account for different IDateTimeProvider implementations.
```
codingupastorm: @bokobza @noescape00 if you guys could give an opinion on this part that would be great.

Could be a bit cleaner but this is the smallest change that works.

noescape00: This change introduces copy-pasted logic. 
If such logic is required for integration tests maybe it's better to create miner for integration tests that would behave differently in a way that tests require?

codingupastorm: It does introduce copy-pasted logic, to support different DateTimeProviders. It would be possible to `do while` it or similar - I can remove the copy paste but this is just simpler.

 Otherwise why are we injecting an `IDateTimeProvider` and then making assumptions about its implementation??

>maybe it's better to create miner for integration tests that would behave differently in a way that tests require?

I'm not on board with this, there are too many places where we do this using inheritance for tests and it's inflexible, we end up testing different implementations of core classes, and it's hard to maintain.

codingupastorm: To avoid "copy-pasted logic" I adjusted the PoAMiner

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -164,9 +156,6 @@ private async Task CreateBlocksAsync()
                     builder.AppendLine($"Block was mined {chainedHeader}.");
                     builder.AppendLine("<<==============================================================>>");
                     this.logger.LogInformation(builder.ToString());
-
-                    int halfTargetSpacingMs = (int)this.network.ConsensusOptions.TargetSpacingSeconds * 1000 / 2;
-                    await this.WaitBeforeCanMineAsync(halfTargetSpacingMs, this.cancellation.Token).ConfigureAwait(false);
```
codingupastorm: Also removed this, superfluous, don't need it when it would always loop over and enter the wait elsewhere anyway

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -179,7 +168,41 @@ private async Task CreateBlocksAsync()
...
+        /// <param name="myTimestamp">The next time slot this node should mine at.</param>
+        protected virtual async Task WaitUntilWeCanMineAsync(uint myTimestamp)
+        {
+            int waitingTime = this.GetWaitingTimeInSeconds(myTimestamp);
+
+            // Accounts for different IDateTimeProvider implementations. The standard implementation would progress
+            // after one iteration, but for other cases where we are emulating time, this may loop more than once.
+            while (waitingTime > 0)
+            {
+                this.logger.LogInformation("Waiting {0} seconds until block can be mined.", waitingTime);
```
bokobza: I think this should be outside of the while so that if we're ready to mine now, we have a log that says so anyway.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
```diff
@@ -179,7 +168,41 @@ private async Task CreateBlocksAsync()
...
+
+                await this.TaskDelayAsync(waitingTime * 1000, this.cancellation.Token).ConfigureAwait(false);
+
+                waitingTime = this.GetWaitingTimeInSeconds(myTimestamp);
```
bokobza: why do we need to recalculate this here?

codingupastorm: `GetWaitingTimeInSeconds` is what hits the `IDateTimeProvider`, so we get the new time and either continue out if it's 0, or keep waiting until the `IDateTimeProvider` returns a different result.

src/Stratis.Bitcoin.Features.PoA/PoAMiner.cs
codingupastorm: Makes sense, will do.

PR #2763, opened 2018-11-15
prevent to stuck peers forcing resync

MithrilMan:


src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -284,9 +294,15 @@ protected async Task ProcessHeadersAsync(INetworkPeer peer, List<BlockHeader> he
 
                 if (result == null)
                 {
-                    this.logger.LogTrace("Processing of {0} headers failed.", headers.Count);
-                    this.logger.LogTrace("(-)[PROCESSING_FAILED]");
-                    return;
+                    // If a peer has a new block and our current peer is not exactly at the tip,
+                    // we try to trigger a resync, this will bring us to the tip of the peer.
+                    if (headers.Count == 1 && (DateTime.UtcNow - this.lastForcedResync) >= this.forceResyncThreshold)
```
zeptin: Shouldn't this be using the date time provider?

dangershony: Well not a big deal we just use this to measure the last time this happened, but DTP may make it easier to change this for tests.

dangershony: I think we should put the two trace lines after the `if` becuse these are exist logs

MithrilMan: moved

Issue #2758, opened 2018-11-14
Missing implementation for RPC calls


maciejzaleski:
Following is a list of required RPC which are required in SBFN

Results for node at address: 192.168.98.200:16174
 - [Optional ] GetInfo: Pass
 - GetBlockCount: Pass
 - GetTransaction:  !!!!!! Fail (The remote server returned an error: (404) Not Found.) !!!!!
 - GetRawTransaction: Pass
 - DecodeRawTransaction: Pass
 - ValidateAddress: Pass
 - GetBlockHash: Pass
 - GetBlock: Pass
 - GetBalance:  !!!!!! Fail (The remote server returned an error: (404) Not Found.) !!!!!
 - ListUnspent:  !!!!!! Fail (The remote server returned an error: (404) Not Found.) !!!!!
 - ListAddressGroupings:  !!!!!! Fail (The remote server returned an error: (404) Not Found.) !!!!!
 - SendMany:  !!!!!! Fail (The remote server returned an error: (404) Not Found.) !!!!!

**Missing RPC calls details**
```gettransaction <txid>```
> Get detailed information about \<txid\>

```getbalance [account] [minconf=1]```
> If [account] is not specified, returns the server's total available balance.
> If [account] is specified, returns the balance in the account.

```listunspent [minconf=1] [maxconf=9999999] ["address",...]```
> Returns array of unspent transaction outputs with between minconf and maxconf (inclusive) confirmations.
> Optionally filtered to only include txouts paid to specified addresses.
> Results are an array of Objects, each of which has:
> {txid, vout, scriptPubKey, amount, confirmations}

```listaddressgroupings```
> Lists groups of addresses which have had their common ownership made public by common use as inputs or as the resulting change in past transactions

```sendmany <fromaccount> {address:amount,...} [minconf=1] [comment]```
> amounts are double-precision floating point numbers

mikedennis:
getbalance should be there. Here is a test https://github.com/stratisproject/StratisBitcoinFullNode/blob/master/src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs#L18-#L35

This RPC is in the wallet feature, is that included on the node?

mikedennis:
Ditto for gettransaction
It's also already in the wallet feature and requires a wallet.

sondreb:
I just want to mention, that #1761 has some comments regarding "getwalletinfo", which is not mentioned in this issue.

PR #2768, opened 2018-11-18
[RPC] Add listunspent

mikedennis:
Part of #2758

See https://bitcoin.org/en/developer-reference#listunspent for relevant documentation.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
@@ -667,7 +667,7 @@ namespace Stratis.Bitcoin.Features.Wallet.Controllers
                         Index = st.Transaction.Index,
                         IsChange = st.Address.IsChangeAddress(),
                         CreationTime = st.Transaction.CreationTime,
-                        Confirmations = st.Transaction.BlockHeight == null ? 0 : this.chain.Tip.Height - st.Transaction.BlockHeight.Value
+                        Confirmations = st.Confirmations
                     }).ToList()
                 });
             }
--- a/src/Stratis.Bitcoin.Features.Wallet/Models/UnspentCoinModel.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/Models/UnspentCoinModel.cs
@@ -1,0 +1,75 @@
+﻿using NBitcoin;
+using Newtonsoft.Json;
+using Stratis.Bitcoin.Utilities.JsonConverters;
+
+namespace Stratis.Bitcoin.Features.Wallet.Models
+{
+    /// <summary>
+    /// Model for Json response for listunspent RPC call.
+    /// </summary>
+    public class UnspentCoinModel
+    {
+        /// <summary>
+        /// Transaction id.
+        /// </summary>
+        [JsonProperty(PropertyName = "txid")]
+        [JsonConverter(typeof(UInt256JsonConverter))]
+        public uint256 Id { get; set; }
+
+        /// <summary>
+        /// The index of the output in the transaction.
+        /// </summary>
+        [JsonProperty(PropertyName = "vout")]
+        public int Index { get; set; }
+
+        /// <summary>
+        /// The Base58 representation of this address.
+        /// </summary>
+        [JsonProperty(PropertyName = "address")]
+        public string Address { get; set; }
+
+        /// <summary>
+        /// Account name.
+        /// </summary>
+        [JsonProperty(PropertyName = "account")]
+        public string Account { get; set; }
+
+        /// <summary>
+        /// The output script paid, encoded as hex.
+        /// </summary>
+        [JsonProperty(PropertyName = "scriptPubKey")]
+        public string ScriptPubKeyHex { get; set; }
+
+        /// <summary>
+        /// If the output is a P2SH whose script belongs to this wallet, this is the redeem script.
+        /// </summary>
+        [JsonProperty(PropertyName = "redeemScript")]
+        public string RedeemScriptHex { get; set; }
+
+        /// <summary>
+        /// The transaction amount.
+        /// Serialized in coins (BTC).
+        /// </summary>
+        [JsonProperty(PropertyName = "amount")]
+        [JsonConverter(typeof(MoneyInCoinsJsonConverter))]
+        public Money Amount { get; set; }
+
+        /// <summary>
+        /// The number of confirmations.
+        /// </summary>
+        [JsonProperty(PropertyName = "confirmations")]
+        public int Confirmations { get; set; }
+
+        /// <summary>
+        /// Whether the private key or keys needed to spend this output are part of the wallet.
+        /// </summary>
+        [JsonProperty(PropertyName = "spendable")]
+        public bool IsSpendable { get; set; }
+
+        /// <summary>
+        /// Whether the wallet knows how to spend this output.
+        /// </summary>
+        [JsonProperty(PropertyName = "solvable")]
+        public bool IsSolvable { get; set; }
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -67,6 +68,62 @@ public void TestRpcGetBlockWithValidHashIsSuccessful()
...
+                RPCClient rpcClient = node.CreateRPCClient();
+                const int numCoins = 5;
+                int blocksToMine = (int)network.Consensus.CoinbaseMaturity + numCoins;
+                TestHelper.MineBlocks(node, blocksToMine);
+                TestHelper.WaitLoop(() => node.FullNode.GetBlockStoreTip().Height == blocksToMine);
+
+                var minerAddress = BitcoinAddress.Create(node.MinerHDAddress.Address, network);
+
+                // validate existing address and minconf
+                var coins = rpcClient.ListUnspent((int)network.Consensus.CoinbaseMaturity+2, 99999, minerAddress);
```
zeptin: Doesn't this method take an array of `BitcoinAddress`?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -221,6 +225,46 @@ public GetTransactionModel GetTransaction(string txid)
...
+                    {
+                        unspentCoins.Add(new UnspentCoinModel()
+                        {
+                            Account = accountReference.AccountName,
+                            Address = spendableTx.Address.Address,
+                            Id = spendableTx.Transaction.Id,
+                            Index = spendableTx.Transaction.Index,
+                            Amount = spendableTx.Transaction.Amount,
+                            ScriptPubKeyHex = spendableTx.Transaction.ScriptPubKey.ToHex(),
+                            RedeemScriptHex = spendableTx.Transaction.ScriptPubKey.PaymentScript?.ToHex(), // Is this right?
```
zeptin: I don't think it is possible to trivially assign this like you are doing here (apart from the PaymentScript not being the same thing as the RedeemScript). It would require special handling within the wallet to support arbitrary P2SH payments, like we have partially done with the multisig wallet in the sidechain project. The crux of it is, the redeem script is not stored in the transaction data anywhere and cannot be derived from it. It has to be maintained somewhere else at the time of transaction creation, or communicated to the party wanting to spend the transaction out-of-band.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -221,6 +225,46 @@ public GetTransactionModel GetTransaction(string txid)
...
+                            Confirmations = confirmations,
+                            IsSpendable = spendableTx.Transaction.IsSpendable(),
+                            IsSolvable = spendableTx.Transaction.IsSpendable() // Is it's spendable we assume it's solvable.
```
zeptin: Seems like a reasonable assumption - if the transaction was sent to an address we own we at least nominally know (knew) how to construct the redeem scripts etc., presuming wallet support for such transactions.

There are probably edge cases, but those can be addressed in future.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -221,6 +225,46 @@ public GetTransactionModel GetTransaction(string txid)
             return model;
         }
 
+        [ActionName("listunspent")]
+        [ActionDescription("Returns an array of unspent transaction outputs belonging to this wallet.")]
+        public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmations = 9999999, string addressesJson = null)
```
bokobza: Is this method not the same as what we have in `WalletManager.GetSpendableTransactionsInAccount`?
Look also at `Walletcontroller.GetSpendableTransactions`.
I think the work might be a bit duplicated here and we could use the aforementioned for our purpose.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: Yes this IsSolvable is a fairly new feature to bitcoin core so not sure exactly the purpose behind it.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: Yes the logic is almost identical to the API spendable-transactions but formatting specific to the RPC specification.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
bokobza: Kevin is right here.
Our wallet only generates P2PKH addresses and not P2SH ones. As such, we don't store or even have redeem scripts in our outputs. You can give this a null and maybe add a TOD to say that this needs to be reviewed when we support P2SH addresses, whether it's multisig or arbitrary ones.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
bokobza: Ah yes indeed.
In this case, can we add the number of confirmations to `UnspentOutputReference` and update `walletManager.GetSpendableTransactionsInAccount` all the way down to `HdAccount.GetSpendableTransactions` ?
Then, if we're missing something in the `UnspentOutputReference` we can add to it.

src/Stratis.Bitcoin.Features.Wallet/Models/UnspentCoinModel.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "redeemScript")]
+        public string RedeemScriptHex { get; set; }
+
+        /// <summary>
+        /// The transaction amount.
+        /// Serialized in coins (BTC).
+        /// </summary>
+        [JsonProperty(PropertyName = "amount")]
+        [JsonConverter(typeof(MoneyInCoinsJsonConverter))]
```
bokobza: Doesn't `MoneyJsonConverter` work here?

mikedennis: MoneyJsonConverter serializes in Satoshis - according to RPC spec I need to serialize in BTC. That's the only difference.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
mikedennis:
```
        public UnspentCoin[] ListUnspent(int minconf, int maxconf, params BitcoinAddress[] addresses)
        {
            IEnumerable<string> addr = from a in addresses select a.ToString();
            RPCResponse response = SendCommand(RPCOperations.listunspent, minconf, maxconf, addr.ToArray());
            return response.Result.Select(i => new UnspentCoin((JObject)i, this.Network)).ToArray();
        }
```

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -667,7 +667,7 @@ public IActionResult GetSpendableTransactions([FromQuery] SpendableTransactionsR
                         Index = st.Transaction.Index,
                         IsChange = st.Address.IsChangeAddress(),
                         CreationTime = st.Transaction.CreationTime,
-                        Confirmations = st.Transaction.BlockHeight == null ? 0 : this.chain.Tip.Height - st.Transaction.BlockHeight.Value
+                        Confirmations = st.Confirmations
```
bokobza: excellent

PR #2769, opened 2018-11-18
[RPC] Use correct data for getpeerinfo response

### LINKED ISSUE #2732: [RPC] getpeerinfo shows incorrect user agents
The data coming back seems to be coming from the local node, not the actual peer:

```
                    if (peer.MyVersion != null)
                    {
                        peerNode.LocalAddress = peer.MyVersion.AddressReceiver?.ToString();
                        peerNode.Services = ((ulong)peer.MyVersion.Services).ToString("X");
                        peerNode.Version = (uint)peer.MyVersion.Version;
                        peerNode.SubVersion = peer.MyVersion.UserAgent;
                        peerNode.StartingHeight = peer.MyVersion.StartHeight;
                    }
```

It looks like these should be `PeerVersion.*` instead.

zeptin:
Fixes #2732


bokobza:
Good spot mate.

Issue #2731, opened 2018-11-10
call getrawtransaction RPC with null reply


xiewenlong1202:
As I know, Stratis is a C# implementation of Bitcoind, I trying to use the same RPC call of BTC to act with Stratis. But after try it, I got some problem with it.
1. I try to call the getrawtransaction RPC method and it return with 204 status code and nothing about the transaction
after that I try to use the Swagger API /api/BlockStore/block to get transaction details, also set ShowTransactionDetails to be true
this time I got transaction information but, the field is different from  a BTC transaction
In the vouts, I Can't find the address field

I want to know, how to get transactions infos in a block in Stratis blockchain.

zeptin:
@xiewenlong1202 Are you running the daemon with `-txindex=1`?

xiewenlong1202:
thanks for reply, I did not run the daemon with -txindex=1 argument
the command I used to start Stratis Node is as follows:
dotnet run -testnet -server=1 -rpcuser=oneroot -rpcpassword=oneroot123

the request to Swagger UI is as follows:
curl -X GET "http://localhost:38221/api/BlockStore/block?Hash=0d4971ab7f86b507e7a296c5e19c3d65e3c98864a8cb9f9be54129463432a95e&ShowTransactionDetails=true&OutputJson=true" -H "accept: application/json"

the reply I got is as follow:
{
  "transactions": [
    {
      "hex": "01000000506ee65b010000000000000000000000000000000000000000000000000000000000000000ffffffff050340b50900ffffffff0100000000000000000000000000",
      "txid": "4f5b305dbe09d755b7efa167613b1f52ddf87a2e62d73f74bd94a722ba956f22",
      "size": 69,
      "version": 1,
      "locktime": 0,
      "vin": [
        {
          "coinbase": "0340b50900",
          "sequence": 4294967295
        }
      ],
      "vout": [
        {
          "value": 0,
          "n": 0,
          "scriptPubKey": {
            "asm": "",
            "hex": "",
            "type": "nonstandard"
          }
        }
      ]
    },
    {
[...]

PR #2770, opened 2018-11-18
[RPC] Handle paytopubkey tx type

### LINKED ISSUE #2748: No "address" field in the reply of decoderawtransaction RPC
Hello there
recently I tried the decoderawtransaction RPC call of the StratisFullNode and in the Vout's "scriptPubKey" field, there is only there parameter value: asm, hex, type
the same call in swagger ui also return similar result
I can't find the "address" filed in the response. But, in the reply of Bitcoind for the same rpc call, the Vout indeed have the "address".

can anybody tell me how to get this field? with additional arguments or calculate it from the returned values?

zeptin:
Fixes #2748 (hopefully)

Also see #2731

The root issue is that NBitcoin did not seem to have a `TxDestination` for P2PK transactions. While these are not often used any more, the coinstake transactions in each block tend to be of this type. So when trying to decode them no address information was getting returned.

This is quite a low level change and I am not 100% sure it is correct. It does however decode the coinstake destination into the address I'd expect, per the block explorer.

Example Stratis transaction hex:

```01000000506ee65b0135c59cb8bf34d98590efe61fc94290c07a8b048b893199b71f7ee7b9aa6422200100000049483045022100b28977b2e530fd2466b6cc4b35276ae158a1e9975c862c8a6e9af9b2db292b0902201b5ddcf93944d23a45ec50ec4410d228ef331d00a4c3602bf0e95f49279c383a01ffffffff020000000000000000001c339aeb70000000232102a961c743372b3b79e4c6a1c2917fecbcf351158e4b01fde729bb70f308cd34f4ac00000000```

-> resulting address for Vout 1:

```"TPcVgKfg2wHgABG3U5Rim7hKfk8mQPDWAH"```

Compared with block explorer:

https://chainz.cryptoid.info/strat-test/tx.dws?10ebe9c14f0a2bacfd361642597b6ab359db7224c61ecb61e239539a583725ab.htm

### DIFF
--- a/src/NBitcoin/Script.cs
+++ b/src/NBitcoin/Script.cs
@@ -935,6 +935,9 @@ namespace NBitcoin
         /// <returns></returns>
         public TxDestination GetDestination(Network network)
         {
+            PubKey pubKeyParams = PayToPubkeyTemplate.Instance.ExtractScriptPubKeyParameters(this);
+            if (pubKeyParams != null)
+                return pubKeyParams.Hash;
             KeyId pubKeyHashParams = PayToPubkeyHashTemplate.Instance.ExtractScriptPubKeyParameters(this);
             if(pubKeyHashParams != null)
                 return pubKeyHashParams;

### CODE REVIEW

**Comment by zeptin** [comment]
Upon further research, I think the block explorer is actually presenting a false reality. P2PK transactions have no address; we could possibly opt to present the P2PKH address instead for user-friendliness.

**Comment by zeptin** [comment]
For reference, here is how X distinguishes between destinations. It is basically doing what I am doing in this PR:

```
bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
{
    vector<valtype> vSolutions;
    txnouttype whichType;
    if (!Solver(scriptPubKey, whichType, vSolutions))
        return false;

    if (whichType == TX_PUBKEY)
    {
        addressRet = CPubKey(vSolutions[0]).GetID();
        return true;
    }
    else if (whichType == TX_PUBKEYHASH)
    {
        addressRet = CKeyID(uint160(vSolutions[0]));
        return true;
    }
    else if (whichType == TX_SCRIPTHASH)
    {
        addressRet = CScriptID(uint160(vSolutions[0]));
        return true;
    }
    // Multisig txns have more than one address...
    return false;
}
```

PR #2793, opened 2018-11-21
[Coinview] Deduplicate coinview cache fetches for staking

### LINKED ISSUE #2761: Efficiency of coinview caching
Currently if the cached coinview receives an array of multiple copies of the same txid it appears it will do a lookup for all of them against the cache. It also appears that the DBreeze coinview will perform one lookup for each occurrence, rather than one lookup per unique txid.

This could either be handled in all the callers of `FetchCoinAsync`, or the txid array could be collapsed within `FetchCoinsAsync` prior to fetching.

zeptin:
Fixes #2761

I examined other calls to `CachedCoinview.FetchCoinsAsync` and they all either use the CoinviewHelper (which internally utilises a HashSet) or only fetch one transaction at a time. This was the only apparent place where de-duplication needed to be done.


bokobza:
Yeah looks ok to me.

PR #2797, opened 2018-11-22
[SC] Add gas refund label to wallet history

### LINKED ISSUE #2773: [SC] Add gas refund label to wallet
Add a label to `ContractTransactionItemType` that identifies when a tx is a gas refund.

rowandh:
Closes #2773.


src/Stratis.Bitcoin.Features.SmartContracts/Wallet/SmartContractWalletController.cs
```diff
@@ -135,13 +135,15 @@ public IActionResult GetHistory(string walletName, string address)
...
+                    bool isCoinBase = transaction.IsCoinBase.HasValue && transaction.IsCoinBase.Value;
+
                     // Record a receive transaction
                     transactionItems.Add(new ContractTransactionItem
                     {
                         Amount = transaction.Amount.ToUnit(MoneyUnit.Satoshi),
                         BlockHeight = transaction.BlockHeight,
                         Hash = transaction.Id,
-                        Type = ContractTransactionItemType.Received,
+                        Type = isCoinBase ? ContractTransactionItemType.GasRefund : ContractTransactionItemType.Received,
```
codingupastorm: For miner wallets this will say GasRefund when it's the MinerReward. Just checked the Payment and TransactionData classes, don't know that there's anything we can do about it - to change it we would need to TxOut index

codingupastorm: We can use `Index` on the `TransactionData` object to get whether this is a miner reward (index == 0)

PR #2799, opened 2018-11-22
[SC] Add method string parameter exception handling

### LINKED ISSUE #2772: [SC] Handle method param deserialization errors in wallet API
Deserialization can throw an exception but it remains unhandled in `SmartContractTransactionService`.

rowandh:
Closes #2772.

### DIFF
--- a/src/Stratis.SmartContracts.Executor.Reflection/Serialization/MethodParameterStringSerializer.cs
+++ b/src/Stratis.SmartContracts.Executor.Reflection/Serialization/MethodParameterStringSerializer.cs
@@ -124,9 +124,9 @@ namespace Stratis.SmartContracts.Executor.Reflection.Serialization
 
             var processedParameters = new List<object>();
 
-            try
+            foreach (var parameter in split)
             {
-                foreach (var parameter in split)
+                try
                 {
                     string[] parameterSignature =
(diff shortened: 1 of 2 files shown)


src/Stratis.SmartContracts.Executor.Reflection/Serialization/MethodParameterStringSerializer.cs
```diff
@@ -118,49 +118,58 @@ public object[] Deserialize(string parameters)
...
-                else
-                    throw new Exception(string.Format("{0} is not supported.", parameterSignature[0]));
+                    else
+                        throw new MethodParameterStringSerializerException(string.Format("{0} is not supported.",
+                            parameterSignature[0]));
+                }
+            }
+            catch (Exception e) when (e is FormatException || e is OverflowException || e is ArgumentException || e is ArgumentNullException)
+            {
+                throw new MethodParameterStringSerializerException(e.Message);
```
codingupastorm: How about we use the InnerException functionality and pass the exception into the constructor as well, to maintain any extra data on the exception object

rowandh: Good idea, will add.

PR #2802, opened 2018-11-22
[BlockStore / Consensus] Add check for node disposing

fassadlr:


thefazzer:
@fassadlr Have checked this twice in a new testnet environment. 
 In both cases the PR appears to fix the bug and we understand why the change in logic has fixed it.  Approved.

PR #2804, opened 2018-11-22
Tx finality checks

noescape00:
Fix: #2791 

Logic for finality checks was taken from QT. 
See `main.cpp:2070` and `main.cpp:371`

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
@@ -109,7 +109,10 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.CommonRules
         }
 
         /// <summary>Checks if transaction if final.</summary>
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/Rules/CommonRules/CoinviewRule.cs
```diff
@@ -120,6 +108,12 @@ public override async Task RunAsync(RuleContext context)
             else this.Logger.LogTrace("BIP68, SigOp cost, and block reward validation skipped for block at height {0}.", index.Height);
         }
 
+        /// <summary>Checks if transaction if final.</summary>
+        protected virtual bool IsTxFinal(Transaction transaction, RuleContext context)
+        {
+            return transaction.IsFinal(context.ValidationContext.ChainedHeaderToValidate);
```
dangershony: What's currently in `transaction.IsFinal` the same as X?

noescape00: Yeah

PR #2807, opened 2018-11-22
PH store to use sorted dictionary

dangershony:


src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
```diff
@@ -269,6 +272,12 @@ public async Task SaveAsync()
                 this.pendingTipHashHeight = null;
             }
 
+            if (pendingBatch.Count == 0)
```
bokobza: Is there any chance that after `this.PendingBatch.Clear();`, `pendingBatch.Count != 0`?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
```diff
@@ -44,10 +45,19 @@ public async Task InitializeAsync(IConsensus consensusParameters, ChainedHeader
 
             int heightToSyncTo = tip.Height > this.numberOfBlocksToKeep ? tip.Height - this.numberOfBlocksToKeep : 0;
 
-            for (int i = tip.Height; i >= heightToSyncTo; i--)
+            for (int i = tip.Height - 1; i >= heightToSyncTo; i--)
```
bokobza: Why the -1?

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
bokobza: ah wait it's not the same object. Well that's a bit confusing lol.

dangershony: yeah it should be renamed I agree

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/RewindDataIndexStore.cs
dangershony: We should not have rewind data of current tip, that was the reason there was a `continue;` there before, the continue is wrong because there should always be rewind data for past blocks, so better is to start form the right place and throw if not found.

PR #2813, opened 2018-11-23
[SC] Parametrize integration tests

### LINKED ISSUE #2778: [SC] Parametrize PoW and PoA integration tests

codingupastorm:
Closes #2778


src/Stratis.SmartContracts.Tests.Common/MockChain/PoWMockChain.cs
```diff
@@ -54,6 +57,26 @@ public void WaitForAllNodesToSync()
             }
         }
 
+        public void MineBlocks(int num)
+        {
+            for (int i = 0; i < num; i++)
+            {
+                int nodeToMineWith = this.height % this.Nodes.Count;
+
+                this.Nodes[nodeToMineWith].MineBlocks(num);
```
rowandh: I think `num` here should be replaced with `i`, or we should increment the height by `num` and remove the loop. Depends on whether the goal is to spread out the mining of `num` blocks across all miners, or have one miner mine `num` blocks.

Issue #2503, opened 2018-10-15
Component that will allows keeping all important tips in sync on startup. 


noescape00:
### Problem

When node shuts down incorrectly (exception\reboot or any other problem that led to it) we need to recover the node next time we launch it. In order to do that we need to initialize tips of all important components at the same height. Components that have to be initialized on the same height at startup are: StakeChain, ChainRepository, CoinView, BlockStore, ProvenheadersRepository (after PH activation height and last checkpoint).

### Solution

We need to introduce a new component that keeps track of last hash and height that was persisted by all components that need to be in sync on startup.

Let's call this new component **TipsManager** (please feel free to suggest a better name for it).

Here is a suggested interface for it: 

```
public interface ITipsManager : IDisposable
{
    // Type is the type of component that requires syncing the tips.
    void RegisterTipProvider(Type type);

    void CommitTipPersited(Type type, ChainedHeader tip);

    // Returns highest tip commited by all components.
    HashHeightPair GetLastCommonTip();
}
```

Every feature will be able to register components on tipsManager during feature initialization.

Every feature that owns a component that requires tips syncing (StakeChain, ChainRepository, CoinView, BlockStore, ProvenheadersRepository) will override initialize method and call `ITipsManager.RegisterTipProvider` with an argument of component that requires tips syncing.

When any component is being initialized it should initialize at height equal to `ITipsManager.GetLastCommonTip().Height` and do a sanity check that hash of the tip is correct.

When any component advances and persists data to the database during node syncing it should call `ITipsManager.CommitTipPersited` AFTER data is persisted. We require every component to guarantee that once tip was commited the node can't loose data of that tip. ITipsManger should also support cases when components commit tip that is below last commited tip. 

`ITipsManager` keeps track of all tips commited and it has it's own tip which is always equal to the highest common commited tip. For example let's say we have 3 components that are registered: component A,B and C.

```
A commits height 5
B commits height 5
C commits height 3
last common tip is 3
C commits height 5
last common tip is 5
C commits height 4
last common tip is 4
C commits height 8
B commits height 9
A commits height 7
last common tip is 7
```

When last common tip changes `LastCommonTip` should be persisted in the background to the database by `ITipsManager`.

### Benefits

Components that require syncing don't need to be initialized in particular order, they don't need to be in the same namespace or have references to each other, they don't need to be aware of other components and know anything about how many there are.

MithrilMan:
[...]

PR #2814, opened 2018-11-23
TipsManager - component to sync tips of all components that can have a tip

noescape00:
This PR creates `TipsManager` but doesn't activate it. 
Activation will be done in the follow-up PR. 

See: https://github.com/stratisproject/StratisBitcoinFullNode/issues/2503


noescape00:
I think that will only add complications. 

First: having GetTip on the interface won't help- we need last persisted tip. Making every component keep additional value is more complicated comparing to calling a method after component saves it's tip.

Here I don't see a need for a provider to implement any interface- we need nothing from the provider except for it's last persisted tip which provider should manually push.

src/Stratis.Bitcoin/Base/TipsManager.cs
```diff
@@ -0,0 +1,205 @@
...
+
+                var hashHeight = new HashHeightPair(tipToSave);
+                this.keyValueRepo.SaveValue(commonTipKey, hashHeight);
+
+                this.logger.LogDebug("Saved common tip '{0}'.", tipToSave);
+            }
+        }
+
+        /// <inheritdoc />
+        public void RegisterTipProvider(object provider)
```
dangershony: Feels a bit wrong to pass an object here.

dangershony: Why even register any component? registration could happen when the component calls set tip.

noescape00: >Feels a bit wrong to pass an object here.

All we need here is ability to compare references so we don't need anything more than just an `object`

>Why even register any component? registration could happen when the component calls set tip.

When we start the node and first component sets the tip we will think that it's the common tip, when other components might have different tips. And we will save wrong value to db. If node shuts down after that we won't be able to recover next time we start the node.

noescape00: String probably won't work since we might accidentally have collisions. 
Type will work as long as we know for sure we won't have 2 instances of the same type having their own tips. 

But still just comparing the reference using object provides the most flexible solution.

MithrilMan: I don't get why not using an interface.
The interface has to be tought as a contract, and if we have an interface called, let say, `IManagedTip`, you know that any component that uses that interface has the tip managed by the TipManager

This way we can get rid of the `RegisterTipProvider(object)` and just have RegisterTipProvider(IManagedTip).
This doesn't mean the solution is more complex, on the contrary, and you clearly declare your intention on components, because the interface IManagedTip declare the component has a tip, and it's managed.

With this change, you can then change your dictionary to be of type `Dictionary<IManagedTip, ChainedHeader>`.
Much more clean without having an object and the only need is to add the interface to components using the tipmanager, but you already need a change to that classes anyway to register anyway.

As a side node, consider that having an interface could even allow us to cycle all registered service that implement this interface and autoregister them

ah, of course even CommitTipPersisted should reference the interface rather then the object, but those are very trivial changes

noescape00: Such an interface will be empty, won't it? 

It feels weird to me to have an empty interface there just to avoid using bare object.
But I can do it. 

@dangershony what do you think?

MithrilMan: I'm not sure it will stay empty, e.g. having a GetCurrentTip could be useful.
I'm not sure about CommitTipPersisted neither because it causes lot of locks during sync, not sure we want it,, but I'll look at it later

MithrilMan:
Another thing I'd change is CommitTipPersisted.
Everytime a component commit a tip it gets a lock and inside the lock there are several operation in place, it can be optimized.

one thing to try is
- lock
- get a snapshot of current chainedheaders
- release the lock
- do all the checks you need to do (lock free)
- if you have to commit something
  - lock again 
  - check if the min common tip is still the same you want to commit
  - commit or ignore (if not on the same tip)

not sure about the logic (didn't thought about it deeply), anyway all these operations under locks, in a constant tip changing environment, could have performance impact

src/Stratis.Bitcoin/Base/TipsManager.cs
noescape00: >e.g. having a GetCurrentTip could be useful.

Some components have HashHeight pair as their tip representation, some have ChainedHeader. 

>I'm not sure about CommitTipPersisted neither because it causes lot of locks during sync

Not really, in IBD we always save stuff in batches. Plus the code inside a lock is pretty lightweighted

noescape00:
I don't think it will cause a perf impact because: 

1) logic inside the lock is pretty simple and will always be executed on a tiny dictionary 
2) tip is supposed to be commited only when component persists data to hard drive, which happens in batches in IBD and once per target spacing when at the tip of the chain. 

So I don't think it will cause noticeable perf problems.

src/Stratis.Bitcoin/Base/TipsManager.cs
fassadlr: I'm definitely in favor of the empty interface and not passing `object`. Its like you are asking for trouble by doing that ;)

noescape00: Ok, fixed 🙂

MithrilMan:
Questions based reading the issue related to this PR:
- I saw you liked the name `CommonTipManager`, any reason you fall back on TipsManager? (I've no preference on this)

- The goal of this class is storing the common tip and handling the startup of components based on the tip.
Why then not having an async task that every minute (or 30sec, or whatever) ask to every registered component its tip ITipProvider.GetTip() and to the logic only once every task run?
What's the cons, apart not being realtime-updated with component tips?

noescape00:
>I saw you liked the name CommonTipManager, any reason you fall back on TipsManager? (I've no preference on this)

It's just shorter. Also don't really have a preference on which name is the best. 

>What's the cons, apart not being realtime-updated with component tips?

In that case we might rewind 30 sec worth of blocks on startup.

MithrilMan:
> > What's the cons, apart not being realtime-updated with component tips?
> 
> In that case we might rewind 30 sec worth of blocks on startup.

That's the worst scenario, is it a big deal? the task could even be changed dynamically depending if we are in IBD or not

> Also asking the components (changing it from push to pull mechanic) makes all other components more complicated- now they will have to cache last persisted tip in some variable.

do you have a list of components/service that will use the TipManager?
anyway you can use anyway the mechanism to communicate the component Tip, but schedule the persistence of the common tip height in a separate task

src/Stratis.Bitcoin/Base/TipsManager.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /// <summary>Continuously persists <see cref="lastCommonTip"/> to hard drive.</summary>
+        private async Task PersistCommonTipContinuouslyAsync()
+        {
+            while (!this.cancellation.IsCancellationRequested)
+            {
+                try
+                {
+                    await this.newCommonTipSetEvent.WaitAsync(this.cancellation.Token).ConfigureAwait(false);
```
dangershony: feels very much like an async queue would do the job rather then an event sync

noescape00: Here we don't need a queue of items. 
So in case save takes 1 seconds and in 1 seconds we change common tip to 5a,6a,7a then we don't need to save them all one by one, only the most recent one.
That's why queue wasn't used but a reset event.

noescape00:
>That's the worst scenario, is it a big deal? the task could even be changed dynamically depending if we are in IBD or not

Not a big deal but it's an argument why current approach is better. 
If alt approach would give us more benefits I'd switch to it but so far I don't see its being better. 

>do you have a list of components/service that will use the TipManager?

block store, chain repo, coinview, ph store, stake chain. maybe some more that I've missed in that list.

>anyway you can use anyway the mechanism to communicate the component Tip, but schedule the persistence of the common tip height in a separate task

It's saved in the background already. We are not blocking caller by saving it to db.

noescape00:
Don't merge. This PR shouldn't go in before the release.

src/Stratis.Bitcoin/Base/TipsManager.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using NBitcoin;
+using Stratis.Bitcoin.Utilities;
+
+namespace Stratis.Bitcoin.Base
+{
+    /// <summary>Interface that every tip provider that uses <see cref="ITipsManager"/> should implement.</summary>
+    public interface ITipProvider
```
dangershony: I suggest we rename this to TipTracker (the component that implements this is not a tip provider its rather has a need to track its current tip) WDYT

noescape00:
>@noescape00 it seems this will not effect any existing code yet (apart for the FinalizedBlockInfoRepository but this will just create a new data folder so no harm here either) is that correct?

Yes, correct

noescape00:
Fixed conflicts. 
Please don't merge, I'll need to run tests and sync

PR #2816, opened 2018-11-23
Ph/allow sync from whitelisted peers only

MithrilMan:
Actually it prevents from to sync with non whitelisted peer but doesn't force whitelisted to send legacy header, so with the commits as it is, you can't whitelist a C# node

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
@@ -200,10 +200,11 @@ namespace Stratis.Bitcoin.Features.Consensus.Behaviors
         {
             INetworkPeer peer = this.AttachedPeer;
 
-            if (this.isGateway && !peer.IsWhitelisted())
+            if (this.isGateway && peer.IsWhitelisted())
             {
-                this.logger.LogTrace("Node is a gateway, cannot sync from non whitelisted peer.");
-                return null;
+                // A gateway node can only sync using regular headers and from whitelisted peers
+                this.logger.LogTrace("Node is a gateway, sync regular headers from whitelisted peer.");
+                return base.BuildGetHeadersPayload(); ;
             }
 
             bool aboveLastCheckpoint = this.GetCurrentHeight() >= this.lastCheckpointHeight;
@@ -217,14 +218,10 @@ namespace Stratis.Bitcoin.Features.Consensus.Behaviors
                         HashStop = null
                     };
                 }
-                else if (peer.IsWhitelisted())
-                {
-                    // If the peer doesn't supports PH but it's whitelisted, issue a standard GetHeadersPayload
-                    return base.BuildGetHeadersPayload();
-                }
                 else
                 {
-                    // If the peer doesn't support PH and isn't whitelisted, return null (stop synch attempt with legacy StratisX nodes).
+                    // If the peer doesn't support PH and we are above last checkpoint
+                    // return null (stop synch).
                     return null;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -35,6 +36,14 @@ public class ProvenHeadersConsensusManagerBehavior : ConsensusManagerBehavior
         private readonly IProvenBlockHeaderStore provenBlockHeaderStore;
         private readonly int lastCheckpointHeight;
         private readonly CheckpointInfo lastCheckpointInfo;
+        private readonly ConnectionManagerSettings connectionManagerSettings;
+
+        /// <summary>
+        /// Specify if the node is a gateway or not.
+        /// Gateway are used internally by Stratis to prevent network split during the transaction from Headers to ProvenHeaders protocol.
```
noescape00: transition

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -184,6 +200,12 @@ protected override GetHeadersPayload BuildGetHeadersPayload()
         {
             INetworkPeer peer = this.AttachedPeer;
 
+            if (this.isGateway && !peer.IsWhitelisted())
```
noescape00: Not sure if it's needed here. I mean the whole added block of code.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -222,6 +244,12 @@ protected override GetHeadersPayload BuildGetHeadersPayload()
         /// <param name="headers">The headers.</param>
         protected async Task ProcessLegacyHeadersAsync(INetworkPeer peer, List<BlockHeader> headers)
         {
+            if (this.isGateway && !peer.IsWhitelisted())
```
noescape00: Same should be for proven headers payload.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -184,6 +200,12 @@ protected override GetHeadersPayload BuildGetHeadersPayload()
         {
             INetworkPeer peer = this.AttachedPeer;
 
+            if (this.isGateway && !peer.IsWhitelisted())
+            {
+                this.logger.LogTrace("Node is a gateway, cannot sync from non whitelisted peer.");
+                return null;
+            }
+
```
noescape00: later in the code if we are the gateway and peer is whitelisted and it is a C# node we will ask for PH. 
We should ask for normal headers.

MithrilMan: yes, actually we don't support C# whitelisted node as gateway. If we need to, we can ask for legacy peer but we should implement even a way to communicate using legacy node when the non-gateway c# node has a block to announce to us

dangershony: why not just do the inverse?

```
if (this.isGateway && peer.IsWhitelisted())
  return base.BuildGetHeadersPayload();
```
And bellow remove any whitelisting reference

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -214,6 +236,18 @@ protected override GetHeadersPayload BuildGetHeadersPayload()
             }
         }
 
+        /// <inheritdoc />
+        protected override Task ProcessHeadersAsync(INetworkPeer peer, List<BlockHeader> headers)
+        {
+            if (this.isGateway)
```
dangershony: If we are a gateway we just don't processes PH

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: I suggest here the same handle gateway and whitelist and remove whitelist form bellow

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
MithrilMan: yes this change was to honor that

PR #2820, opened 2018-11-26
Fix stratisX tests

zeptin:


fassadlr:
Hi @zeptin,thanks for looking into this. However I see that the X tests are still failing? Dan merged something last night that relates to how we sync from X nodes so maybe this should be merged to your branch?

fassadlr:
A couple are still failing the CI but they all are passing locally for me.

Issue #1845, opened 2018-08-03
Move block store to Stratis.Bitcoin


noescape00:
Move block store to Stratis.Bitcoin because it's going to become a mandatory component and we would just implement pruning instead of not using a feature

dangershony:
I am not sure this will be done so please talk to me before acting on this.

noescape00:
Sure.

dangershony:
This might not be needed for prod release but we need to have an additional discussions about this task

noescape00:
Yes sure. 

My points are: 
1) We will be able to remove BSQ.Tip from CHainState and after that ChainState can be completely removed from the codebase (ChainState is a hacky workaround and should be removed at some point)
2) It makes sense to have BS in base because it's mandatory feature

dangershony:
Also consider this PR when implementing this code
https://github.com/stratisproject/StratisBitcoinFullNode/issues/2628

PR #2821, opened 2018-11-26
Moved blockstore to base feature

noescape00:
Moved block store to Stratis.Bitcoin

Fix: #1845


noescape00:
Closing the PR because of the feature freeze.
It's faster to recreate it than merging

PR #2828, opened 2018-11-26
Concurrent chain wrong tip usage fix

noescape00:
Minor fix of a bug that basically was disabling batching


src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -140,7 +136,6 @@ private async Task DequeueContinuouslyAsync()
                         // Set the dequeue task to null so it can be assigned on the next iteration.
                         dequeueTask = null;
                         batch.Add(item);
-                        sendBatch = item == this.chain.Tip;
```
dangershony: We don't need this?

noescape00: Before we used to trigger batch when we were synced to the tip. Now it's always true because consensus tip is equal to chain tip.

PR #2837, opened 2018-11-27
[BlockStore] Dequeue all items from the async queue on dispose

fassadlr:


src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
```diff
@@ -47,6 +47,18 @@ private class AsyncContext
         /// <remarks>All access to this object has to be protected by <see cref="lockObject"/>.</remarks>
         private readonly Queue<T> items;
 
+        /// <summary>The amount of items in the queue.</summary>
+        public int ItemCount
```
noescape00: Why not just `Count`?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -394,6 +394,20 @@ private async Task DequeueBlocksContinuouslyAsync()
...
+
+        /// <summary>
+        /// Ensures that any blocks queued in <see cref="blocksQueue"/> gets added to <see cref="batch"/>
+        /// so that it can be persisted on dispose.
+        /// </summary>
+        private async Task SaveBlocksInBatchAndAsyncQueueAsync()
+        {
+            while (this.blocksQueue.ItemCount > 0)
+            {
+                this.batch.Add(await this.blocksQueue.DequeueAsync());
```
noescape00: configureawait(false)

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -394,6 +394,20 @@ private async Task DequeueBlocksContinuouslyAsync()
                 }
             }
 
+            await SaveBlocksInBatchAndAsyncQueueAsync();
```
noescape00: `this.`

noescape00:
I think it will also be a good idea to add a check to `AddToPending` that cancellation token is not cancelled yet to avoid items consumption during disposure. Throw if we are attempting to add a new item after BSQ's token is cancelled.

noescape00:
I'm not sure if count should be added to the async queue. I remember we discussed that with @Aprogiena but I don't remember his point on why it was a bad idea.

src/Stratis.Bitcoin/Utilities/AsyncQueue.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
fassadlr: Fixed.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
fassadlr: Fixed

noescape00:
>@noescape00 will it make more sense to make TryDequeue a public method?

That can work as well as long as it's thread safe, yeah.

dangershony:
Yeah it is thread safe internally already

PR #2839, opened 2018-11-27
[Configuration] A gateway node can't mine or stake

fassadlr:


fassadlr:
I tested and it works as expected 👍

PR #2840, opened 2018-11-27
Change the agent to be like Stratis:1.2.5 (70012)

bokobza:
Used to be `StratisBitcoin:1.2.5`.
`StratisBitcoin` has changed to `Stratis` and we now include the protocol version.

### DIFF
--- a/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
+++ b/src/Stratis.Bitcoin/Configuration/NodeSettings.cs
@@ -96,7 +96,7 @@ namespace Stratis.Bitcoin.Configuration
         ///   name would be determined. In this case we first need to determine the network.
         /// </remarks>
         public NodeSettings(Network network = null, ProtocolVersion protocolVersion = SupportedProtocolVersion,
-            string agent = "StratisBitcoin", string[] args = null, NetworksSelector networksSelector = null)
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin/Configuration/NodeSettings.cs
```diff
@@ -96,7 +96,7 @@ public class NodeSettings : IDisposable
         ///   name would be determined. In this case we first need to determine the network.
         /// </remarks>
         public NodeSettings(Network network = null, ProtocolVersion protocolVersion = SupportedProtocolVersion,
-            string agent = "StratisBitcoin", string[] args = null, NetworksSelector networksSelector = null)
+            string agent = "Stratis", string[] args = null, NetworksSelector networksSelector = null)
```
noescape00: Maybe StratisFN will be better to avoid mixing stratis full node platform with stratis coin.

fassadlr:
NICE @bokobza ! Looks so much better 👍

src/Stratis.Bitcoin/Configuration/NodeSettings.cs
bokobza: We thought about StratisNode yesterday but in the end we thought Stratis was shorter (it is) and cuter.The coin is called strat, or $STRAT on twitter lol.

PR #2849, opened 2018-11-28
Error in PosConsensusFeature.InitializeAsync #2846

MithrilMan:
about #2846 
this fixes the exception message itself (triggers the guard) but the reason why this exception was rised ins't clear yet.
This can be merged as it is, the problem probably should be fixed on the sidechain repo

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/PosConsensusFeature.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/PosConsensusFeature.cs
@@ -70,7 +70,7 @@ namespace Stratis.Bitcoin.Features.Consensus
             NetworkPeerConnectionParameters connectionParameters = this.connectionManager.Parameters;
 
             // Replace CMB.
-            bool oldCMBRemoved = connectionParameters.TemplateBehaviors.Remove(connectionParameters.TemplateBehaviors.Single(x => x is ConsensusManagerBehavior));
+            bool oldCMBRemoved = connectionParameters.TemplateBehaviors.Remove(connectionParameters.TemplateBehaviors.FirstOrDefault(x => x is ConsensusManagerBehavior));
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.Consensus/PosConsensusFeature.cs
```diff
@@ -70,7 +70,7 @@ public override Task InitializeAsync()
             NetworkPeerConnectionParameters connectionParameters = this.connectionManager.Parameters;
 
             // Replace CMB.
-            bool oldCMBRemoved = connectionParameters.TemplateBehaviors.Remove(connectionParameters.TemplateBehaviors.Single(x => x is ConsensusManagerBehavior));
+            bool oldCMBRemoved = connectionParameters.TemplateBehaviors.Remove(connectionParameters.TemplateBehaviors.FirstOrDefault(x => x is ConsensusManagerBehavior));
```
bokobza: I think we should maybe stop the node if we don't have a `ConsensusManagerBehavior` as we should always have one. In the BaseFeature only, 5 of then are initialized and you can't run a node without the base feature.

So maybe check oldCMBRemoved is null and then throw an exception.

MithrilMan: actually it does the same, if null it will throw on the guard

fassadlr:
@MithrilMan I actually saw this error yesterday as well. But couldnt reproduce it. I t might be worth me spending a couple of hours to see if I can get the error again. I have no idea why `connectionParameters` will not have all the behaviors by that time.

MithrilMan:
@fassadlr that's weird, as long as we initialize features in the same order it should work.
Not sure if this can be an async issue

PR #2852, opened 2018-11-28
Fix issue with exceptions on startup

fassadlr:


dangershony:
Fixes this https://github.com/stratisproject/StratisBitcoinFullNode/issues/2846

@MithrilMan

bokobza:
So this is needed because the exception thrown in #2849 isn't stopping the node.

MithrilMan:
ACK

PR #2856, opened 2018-11-28
(R/C) tip height, refactoring, IBD status

thefazzer:
Preferred: @bokobza @dangershony

### DIFF
--- a/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
+++ b/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
@@ -230,13 +230,15 @@ namespace Stratis.Bitcoin.Connection
             {
                 var chainHeadersBehavior = peer.Behavior<ConsensusManagerBehavior>();
 
+                string peerHeights = $"(r/s):{(chainHeadersBehavior.BestReceivedTip != null ? chainHeadersBehavior.BestReceivedTip.Height.ToString() : peer.PeerVersion?.StartHeight.ToString() ?? "unknown")}";
+                peerHeights += $"/{(chainHeadersBehavior.BestSentHeader != null ? chainHeadersBehavior.BestSentHeader.Height.ToString() : peer.PeerVersion?.StartHeight.ToString() ?? "unknown")}";
+
                 string agent = peer.PeerVersion != null ? peer.PeerVersion.UserAgent : "[Unknown]";
                 builder.AppendLine(
                     "Peer:" + (peer.RemoteSocketEndpoint + ", ").PadRight(LoggingConfiguration.ColumnLength + 15) +
-                    (" connected:" + (peer.Inbound ? "inbound" : "outbound") + ",").PadRight(LoggingConfiguration.ColumnLength + 7) +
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin/Utilities/NodeStats.cs
```diff
@@ -62,7 +66,7 @@ public string GetStats()
             lock (this.locker)
             {
                 string date = this.dateTimeProvider.GetUtcNow().ToString(CultureInfo.InvariantCulture);
-                statsBuilder.AppendLine($"======Node stats====== {date}");
+                statsBuilder.AppendLine($"======Node stats====== {date} ====== IBD:({this.NodeIsInIbd})");
```
noescape00: IBD status probably should be on a separated line after the `====Node stats` header

src/Stratis.Bitcoin/Utilities/NodeStats.cs
```diff
@@ -19,6 +19,8 @@ public interface INodeStats
 
         /// <summary>Collects benchmark stats.</summary>
         string GetBenchmark();
+
+        bool NodeIsInIbd { get; set; }
```
noescape00: This shouldn't be on the interface. 

Stats are always for display so just having this info returned by `string GetStats()` is enough

src/Stratis.Bitcoin/Utilities/NodeStats.cs
```diff
@@ -38,6 +40,8 @@ public NodeStats(IDateTimeProvider dateTimeProvider)
             this.stats = new List<StatsItem>();
         }
 
+        public bool NodeIsInIbd { get; set; }
```
noescape00: Instead get IBD provider from DI and use it in GetStats method

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -230,13 +230,15 @@ private void AddComponentStats(StringBuilder builder)
             {
                 var chainHeadersBehavior = peer.Behavior<ConsensusManagerBehavior>();
 
+                string peerHeights = $"(r/s):{(chainHeadersBehavior.BestReceivedTip != null ? chainHeadersBehavior.BestReceivedTip.Height.ToString() : peer.PeerVersion?.StartHeight.ToString() ?? "unknown")}";
```
dangershony: instead of unknown I suggest just put a `-`

src/Stratis.Bitcoin/Utilities/NodeStats.cs
dangershony: Just inject `InitialBlockDownloadState` in to stats and get that value when log is called

dangershony:
@thefazzer please provide a screen shot of how it looks as well

bokobza:
sorry @noescape00 I didn't write my comments here yesterday.
In the end we decided to go with this:

```
======Consensus Manager======
IBD stage
Unconsumed blocks: 0 -- (0 / 209,715,200 bytes). Cache is filled by: 0%

```

noescape00:
@bokobza np

>In the end we decided to go with this:

I like it

PR #2857, opened 2018-11-28
Ability to access private CHT members in tests

noescape00:
Usage:

```
var testCHT = stratisNode1.FullNode.NodeService<TestChainedHeaderTree>();

Dictionary<uint256, ChainedHeader> chainedHeadersByHash = testCHT.ChainedHeadersByHash;
Dictionary<uint256, HashSet<int>> peerIdsByTipHash = testCHT.PeerIdsByTipHash;
Dictionary<int, uint256> peerTipsByPeerId = testCHT.PeerTipsByPeerId;
```

### DIFF
--- a/src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/FullNodeTestBuilderExtension.cs
+++ b/src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/FullNodeTestBuilderExtension.cs
@@ -2,6 +2,7 @@
 using Microsoft.Extensions.DependencyInjection;
 using Stratis.Bitcoin.Builder;
 using Stratis.Bitcoin.Builder.Feature;
+using Stratis.Bitcoin.Consensus;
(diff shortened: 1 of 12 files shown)


src/Stratis.Bitcoin.IntegrationTests.Common/Extensions/FullNodeTestBuilderExtension.cs
```diff
@@ -58,6 +59,27 @@ public static IFullNodeBuilder MockIBD(this IFullNodeBuilder fullNodeBuilder)
             return fullNodeBuilder;
         }
 
+        public static IFullNodeBuilder UseTestChainedHeaderTree(this IFullNodeBuilder fullNodeBuilder)
```
fassadlr: There is an extension called ReplaceService

fassadlr: Actually no, we will need to extend ReplaceService to take in a Type as the the replacement.

Issue #2862, opened 2018-11-29
[Consensus] Peer presented nonconsecutive headers


dangershony:
```
[2018-11-28 20:33:39.9782 5] INFO: Stratis.Bitcoin.Connection.ConnectionManagerBehavior+<OnStateChangedAsync>d__15.MoveNext Peer '[::ffff:51.140.231.125]:16178' offline, reason: 'The peer was banned, reason: Peer presented nonconsecutive headers.'.
```

noescape00:
Related to https://github.com/stratisproject/StratisBitcoinFullNode/pull/2887

noescape00:
Fixed

PR #2865, opened 2018-11-29
Fix consecutive header bug

fassadlr:


fassadlr:
Fix #2862

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -140,16 +141,26 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
                         this.logger.LogTrace("(-)[NO_PH_AVAILABLE]");
                         break;
                     }
+
+                    // We need to check that the headers we are returning are consecutive.
+                    // If a specific header is found not to follow on from the previous one,
+                    // we just return what we have that are consecutive.
+                    var previous = lastHeader ?? fork;
```
noescape00: unvar

src/Stratis.Bitcoin.Features.PoA/PoAConsensusManagerBehavior.cs
```diff
@@ -44,7 +44,7 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
 
         /// <inheritdoc />
         /// <remarks>Creates <see cref="PoAHeadersPayload"/> instead of <see cref="HeadersPayload"/> like base implementation does.</remarks>
-        protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersPayload, out ChainedHeader lastHeader)
+        public override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersPayload, out ChainedHeader lastHeader)
```
noescape00: This shouldn't be public. We don't want any external code to call it.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -140,16 +141,26 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
...
+                    if (provenBlockHeader.HashPrevBlock != previous.HashBlock)
```
noescape00:
```
if (provenBlockHeader.HashPrevBlock != header.Previous.HashBlock)
```

instead of this and prev line would probably be better

dangershony:
Proven header store looks at height not prev hash, so there can be a race condition where consensus is reorging and signaling new blocks, those will get in to the PH store but while a reorg is taking place this situation can still happen (if a `getprovenheader` is send in the middle of a reorg PH store will have none consecutive headers).

There is nothing to fix really this is the deign of PH store.

src/Stratis.Bitcoin.Features.PoA/PoAConsensusManagerBehavior.cs
dangershony: Its made public so the test can call in to the code, I think its not a big deal if it is public.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: We should compare against the value that was returned form the PH store right?

src/Stratis.Bitcoin.Features.PoA/PoAConsensusManagerBehavior.cs
noescape00: Imho it's exposes what shouldn't be exposed on public interface.
I think in general it's a bad practice to do changes that decrease quality of production code in favor of tests unless it's impossible to test something without doing that.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
noescape00: That line and what is currently in the code does it, yes

noescape00:
>Proven header store looks at height not prev hash, so there can be a race condition where consensus is reorging and signaling new blocks, those will get in to the PH store but while a reorg is taking place this situation can still happen (if a getprovenheader is send in the middle of a reorg PH store will have none consecutive headers).

If this is the case then the fix is correct. 

So we are looking at the scenario when chain reorged already and PH store not yet.

Is it possible to go into that if on ln 129? 
```
if (!(header.Header is ProvenBlockHeader provenBlockHeader))
```

All headers we received for the chain we've reorged to are PH.

Unless I'm missing smth it shouldn't be possible that we don't have PH in memory for chain we reorged to.

noescape00:
Actually, it looks like we can't go into that `if` but we still can get not consecutive header. 

ConcurrentChain's `EnumerateToTip` is just getting block's by height from fork's height to tip's height. 

So if we've reorged and set chain's tip to another header from a different chain while we are inside of `foreach (ChainedHeader header in this.chain.EnumerateToTip(fork).Skip(1))` we will get a bad header. 

In that case this check makes sense, but only if it's after the `if`, not inside of it.

fassadlr:
> In that case this check makes sense, but only if it's after the `if`, not inside of it.

@noescape00 to what `if` are you referring to exactly?

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.PoA/PoAConsensusManagerBehavior.cs
fassadlr: Again, then the design of these classes aren't built in a very TDD way as it makes it VERY difficult to unit test critical logic in the code.

We either expose these methods or we create a `ConstructHeadersPayload` class that we can inject with `chain` and `provenBlockHeaderStore` so that we can easily new this up in a unit test and write test scenarios around it.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
fassadlr: Does something need to change here then? (Confused.com)

dangershony:
I actually think the problem is indeed also in ConcurrentChain and we should avoid using it (yeah it will get deleted but we can remove it from here already now)

But also PH store can still be in an inconsistent state.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -123,12 +122,25 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
                 return null;
             }
 
-            var headers = new ProvenHeadersPayload();
-            foreach (ChainedHeader header in this.chain.EnumerateToTip(fork).Skip(1))
+            var provenHeadersPayload = new ProvenHeadersPayload();
+
+            // Do not return more than 2000 headers from the fork point.
+            ChainedHeader chainedHeaderToStartFrom = this.consensusManager.Tip;
+            if ((this.consensusManager.Tip.Height - fork.Height) > MaxItemsPerHeadersMessage)
```
dangershony: Why take an instance of chainedHeaderToStartFrom and not use it here in the `if`?

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -123,12 +122,25 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
...
+            {
+                // e.g. If fork = 3000 and tip is 6000 we need to start from block 5000.
+                var startFromHeight = (fork.Height + 1) + MaxItemsPerHeadersMessage;
+                chainedHeaderToStartFrom = this.consensusManager.Tip.GetAncestor(startFromHeight);
```
dangershony: Same here use `chainedHeaderToStartFrom` Tip might change as this is  not inside the lock of CT.

dangershony:
Looks good, reminder to make the same change in `ConesnsusManagerBehaviour`

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
noescape00: I think there is a -1 problem. 

If CM.Tip is 5k, fork is 3k we will have

`if (5000 - 3000 > 2000)`

so the if won't be executed. 
Then we will collect 1999 headers instead of 2000

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
fassadlr: Ah ok yes, fixed!

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
fassadlr: Oh yes  of course,  fixed 👍

fassadlr:
> Looks good, reminder to make the same change in `ConesnsusManagerBehaviour`

I have also made the changes in CMB 👍

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -123,12 +122,18 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
                 return null;
             }
 
-            var headers = new ProvenHeadersPayload();
-            foreach (ChainedHeader header in this.chain.EnumerateToTip(fork).Skip(1))
+            var provenHeadersPayload = new ProvenHeadersPayload();
+
+            var chainedHeaderToStartFrom = DetermineFirstHeaderToConstructPayloadFrom(fork);
```
dangershony: I suggest, for readability, this approach.
```
ChainedHeader header = this.FindHeaderToStartFrom(frok)
``` 
And delete `chainedHeaderToStartFrom`

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -217,17 +230,26 @@ protected virtual Payload ConstructHeadersPayload(GetHeadersPayload getHeadersPa
...
+            var chainedHeaderToStartFrom = this.DetermineFirstHeaderToConstructPayloadFrom(fork);
+
+            ChainedHeader header = chainedHeaderToStartFrom;
+
+            for (int heightIndex = chainedHeaderToStartFrom.Height; heightIndex > fork.Height; heightIndex--)
             {
                 lastHeader = header;
                 headersPayload.Headers.Add(header.Header);
 
                 if ((header.HashBlock == getHeadersPayload.HashStop) || (headersPayload.Headers.Count == MaxItemsPerHeadersMessage))
```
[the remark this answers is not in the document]
fassadlr: Fixed

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -124,7 +124,11 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
 
             var provenHeadersPayload = new ProvenHeadersPayload();
 
-            ChainedHeader header = this.FindHeaderToStartFrom(fork);
+            ChainedHeader header = null;
+            if (getHeadersPayload.HashStop != null && getHeadersPayload.HashStop != uint256.Zero)
+                header = this.FindHeaderToStartFromByHashStop(getHeadersPayload.HashStop);
```
dangershony: The content of `FindHeaderToStartFromByHashStop` and `FindHeaderToStartFrom` are practically identical you could easily just call `FindAncestorOrSelf` at the caller if hashstop is not null and just have one method

dangershony: This is also not so correct because if hashstop is has a value the fork needs to change to the value in hashstop

dangershony: Ah sorry no this is correct however we need to make sure hashstop is between fork and starting header

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -205,7 +205,24 @@ protected ChainedHeader FindHeaderToStartFrom(ChainedHeader fork)
...
+        }
+
+        protected ChainedHeader FindHeaderToStartFromByHashStop(uint256 hashStop)
+        {
+            // Find the header
+            var hashStopHeader = this.consensusManager.Tip.FindAncestorOrSelf(hashStop);
+
+            // Do not return more than 2000 headers from the fork point.
+            ChainedHeader chainedHeaderToStartFrom = hashStopHeader;
+            if ((chainedHeaderToStartFrom.Height - hashStopHeader.Height) > MaxItemsPerHeadersMessage)
```
dangershony: This is odd are you not deducting the same value form itself?

Above you have `chainedHeaderToStartFrom = hashStopHeader`
Then under its `chainedHeaderToStartFrom.Height - hashStopHeader.Height`

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -123,12 +122,16 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
                 return null;
             }
 
-            var headers = new ProvenHeadersPayload();
-            foreach (ChainedHeader header in this.chain.EnumerateToTip(fork).Skip(1))
+            var provenHeadersPayload = new ProvenHeadersPayload();
+
+            ChainedHeader header = this.FindStartingHeader(fork, getHeadersPayload.HashStop);
```
noescape00: Maybe rename to FindLastHeader ? 
Because it's not starting header, its the last header to include in the response payload.

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -199,6 +198,38 @@ protected async Task ProcessGetHeadersAsync(INetworkPeer peer, GetHeadersPayload
             }
         }
 
+        protected ChainedHeader FindStartingHeader(ChainedHeader fork, uint256 hashstop)
+        {
+            ChainedHeader tip = this.consensusManager.Tip;
+
+            if (hashstop != null && hashstop != uint256.Zero)
```
noescape00: brackets

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -199,6 +198,38 @@ protected async Task ProcessGetHeadersAsync(INetworkPeer peer, GetHeadersPayload
             }
         }
 
+        protected ChainedHeader FindStartingHeader(ChainedHeader fork, uint256 hashstop)
+        {
+            ChainedHeader tip = this.consensusManager.Tip;
```
noescape00: Not the tip but the highest\last header to include. Sometimes it will be tip, sometimes hashstop or other header in between.

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -123,12 +122,16 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
...
+
+            ChainedHeader header = this.FindLastHeaderForPayload(fork, getHeadersPayload.HashStop);
+
+            for (int heightIndex = header.Height; heightIndex > fork.Height; heightIndex--)
             {
                 if (!(header.Header is ProvenBlockHeader provenBlockHeader))
                 {
                     this.logger.LogTrace("Invalid proven header, try loading it from the store.");
+
                     provenBlockHeader = this.provenBlockHeaderStore.GetAsync(header.Height).GetAwaiter().GetResult();
```
noescape00: ConstructHeadersPayload is called from async methods. 
I think maybe changing it to async and awaiting here instead of `.GetAwaiter().GetResult();` will be a good idea.

fassadlr: Hmm we can but the base method in CMB doesn't do anything async?

PR #2868, opened 2018-11-29
nameof variables fix

thefazzer:
Bugfix introduced by refactoring.  Resolved and nameof introduced to prevent recurrence.


thefazzer:
Preferred: @fassadlr @dangershony

Issue #2864, opened 2018-11-29
Race condition that allows multiple connections to peers from the same subnet


noescape00:
How to reproduce: 
create 2 nodes, A and B. 
A.peers.json should contain B, B.peers.json should contain A. 
Run them at the same time from the same machine or same subnet (first 2 digits of IP address should be the same). 
Nodes will establish bidirectional connection. 

We have code to prevent connections to the nodes from the same subnet (`PeerConnector` checks `PeerIsPartOfExistingGroup`) but in case both connections are established within a certain timeframe this check fails to prevent such connections. 

So let's say we are node A and connecting to node B. We select IP and it passes the `PeerIsPartOfExistingGroup` check, we establish a connection and start handshake. When handshake is finished peer is added to connectionmanager.connectedPeers and after that `PeerIsPartOfExistingGroup` will taker this peer in count. 

If some other node from the same subnet connects to us after we selected the IP for connection but before it was added to peers collection we will end up with connections to peers from the same subnet. 

ps
I believe this is the reason why bidirectional connections are happening on the internal testnet. See https://github.com/stratisproject/StratisBitcoinFullNode/issues/2842

MithrilMan:
can't we just drop the last that handshake?

dangershony:
Perhaps we need to do this check also in the behaviour then?

maciejzaleski:
This has been proved to work as expected now

sondreb:
This means that there needs to be X amount of public nodes available for every person that is behind a NAT with the same IP address, right?

Have a scenario where I might have hundreds of users running on the same public Wi-Fi, sharing the same external IP or range. Turning off IP Range filtering would be one option to get around that.

PR #2869, opened 2018-11-29
Disconnect after handshake if already have a peer on the same group

noescape00:
Quick and dirty fix for https://github.com/stratisproject/StratisBitcoinFullNode/issues/2864

### DIFF
--- a/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
+++ b/src/Stratis.Bitcoin/Connection/ConnectionManager.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Sockets;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using NBitcoin;
@@ -82,6 +83,8 @@ namespace Stratis.Bitcoin.Connection
 
         private IConsensusManager consensusManager;
 
+        private AsyncQueue<INetworkPeer> connectedPeersQueue;
+
         public ConnectionManager(IDateTimeProvider dateTimeProvider,
             ILoggerFactory loggerFactory,
             Network network,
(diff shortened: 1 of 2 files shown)


noescape00:
@maciejzaleski I've tried to reproduce the problem that you've mentioned but had no luck. 
I'm using following setup: 

2 nodes, both fully synced, ran without any command line arguments, each has a peers file with another node's IP. 

When I start them both I see a single connection established between them that is displayed on both consoles at all times. 

Could you please provide info on how to reproduce the problem or logs?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -264,13 +268,69 @@ public void Dispose()
...
+            this.connectedPeersQueue.Dispose();
+
             this.networkPeerDisposer.Dispose();
         }
 
         /// <inheritdoc />
         public void AddConnectedPeer(INetworkPeer peer)
         {
             this.connectedPeers.Add(peer);
+            this.connectedPeersQueue.Enqueue(peer);
```
dangershony: Why asyncqueue this operations should be quite fast right?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -264,13 +268,69 @@ public void Dispose()
...
+                }
+            }
+
+            if (isAddNodeOrConnect)
+            {
+                this.logger.LogTrace("(-)[ADD_NODE_OR_CONNECT]:false");
+                return false;
+            }
+
+            byte[] peerGroup = peer.PeerEndPoint.MapToIpv6().Address.GetGroup();
```
dangershony: Dont we have a method to check same group already (the same method we check on connect)?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: To avoid race conditions. I don't want to use locks here but also don't want peers to be added to the same collection right after I've checked against it or in the middle of it.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: No. There is a method in peer connector - `PeerIsPartOfExistingGroup` but it's private and should be moved somewhere else and it's interface should change in order to make it good to be used for other components

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
dangershony: ah right yeah and asyncqueue is single threaded...  elegant solution actually.

noescape00:
> do we want to test this branch on its own in simulation network?

In case we do that we won't be able to establish connections on the test network according to the topology. 
First it will be required to change IPs on the test network so no 2 nodes are on the same group.

PR #2871, opened 2018-11-30
[SC] TokenStandard adjustments

rowandh:
* Replace uint with ulong
* Rename Transfer method
* Check existing approval amount when granting new approval


src/Stratis.SmartContracts.Standards/IStandardToken.cs
```diff
@@ -8,22 +8,22 @@ public interface IStandardToken
...
+        ulong GetBalance(Address address);
 
         /// <summary>
         /// Transfers tokens owned by the sender to the destination address.
         /// </summary>
         /// <param name="to">The address to transfer the tokens to.</param>
         /// <param name="amount">The amount of tokens to transfer.</param>
         /// <returns>A bool representing whether the transfer was successful.</returns>
-        bool Transfer(Address to, uint amount);
+        bool TransferTo(Address to, ulong amount);
```
codingupastorm: What's the idea behind this?

rowandh: Prevent confusion with the `Transfer` method that exists on all contracts.

PR #2875, opened 2018-11-30
Correctly store and load rewind data index, 

dangershony:
And correctly look up for  utxo spent on another chain.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
@@ -371,10 +371,6 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
                 }
             }
 
-            //UnspentOutputs matchingUnspentUtxo = rewindData.OutputsToRestore
-            //    .Where((unspent, i) => (unspent.TransactionId == input.PrevOut.Hash) && (i == input.PrevOut.N))
-            //    .FirstOrDefault();
-
             if (matchingUnspentUtxo == null)
             {
                 this.Logger.LogTrace("(-)[UTXO_NOT_FOUND_IN_REWIND_DATA]");
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
@@ -133,14 +133,16 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
             {
                 // The trx was found now check if the UTXO is spent.
                 prevUtxo = coins.UnspentOutputs[0];
-                if (txIn.PrevOut.N >= prevUtxo.Outputs.Length)
+
+                TxOut utxo = null;
+                if (txIn.PrevOut.N < prevUtxo.Outputs.Length)
                 {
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -355,9 +357,19 @@ private UnspentOutputs CheckIfCoinstakeIsSpentOnAnotherChain(ProvenBlockHeader h
             }
 
             RewindData rewindData = this.PosParent.UtxoSet.GetRewindData(rewindDataIndex.Value).GetAwaiter().GetResult();
-            UnspentOutputs matchingUnspentUtxo = rewindData.OutputsToRestore
-                .Where((unspent, i) => (unspent.TransactionId == input.PrevOut.Hash) && (i == input.PrevOut.N))
-                .FirstOrDefault();
+
+            UnspentOutputs matchingUnspentUtxo = null;
+            foreach (UnspentOutputs unspent in rewindData.OutputsToRestore)
```
dangershony: I wonder if here we should also add one more check 
```
            UnspentOutputs matchingUnspentUtxo = null;
            foreach (UnspentOutputs unspent in rewindData.OutputsToRestore)
            {
                if (unspent.TransactionId == input.PrevOut.Hash)
                {
                    if (input.PrevOut.N < unspent.Outputs.Length)
                    {
                        if (unspent.Outputs[input.PrevOut.N] != null)
                        {
                            matchingUnspentUtxo = unspent;
                            break;
                        }
                    }
                }
            }
```

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -355,9 +357,19 @@ private UnspentOutputs CheckIfCoinstakeIsSpentOnAnotherChain(ProvenBlockHeader h
             }
 
             RewindData rewindData = this.PosParent.UtxoSet.GetRewindData(rewindDataIndex.Value).GetAwaiter().GetResult();
-            UnspentOutputs matchingUnspentUtxo = rewindData.OutputsToRestore
-                .Where((unspent, i) => (unspent.TransactionId == input.PrevOut.Hash) && (i == input.PrevOut.N))
-                .FirstOrDefault();
+
```
dangershony: I wonder if here we should also add one more check 
```
            UnspentOutputs matchingUnspentUtxo = null;
            foreach (UnspentOutputs unspent in rewindData.OutputsToRestore)
            {
                if (unspent.TransactionId == input.PrevOut.Hash)
                {
                    if (input.PrevOut.N < unspent.Outputs.Length)
                    {
                        if (unspent.Outputs[input.PrevOut.N] != null)
                        {
                            matchingUnspentUtxo = unspent;
                            break;
                        }
                    }
                }
            }
```

dangershony: I am not sure because later we will check the UTXO is not null when we validate the signature, this is a permanent problem so if we do this check here we will just continue to try this header because later we mark this as `InsufficientHeaderInformation` so I suggest either to throw another error code or to not add this check at all.

MithrilMan: if we have a rewindIndex for that output shouldn't we have a rewindData too?
if yes, than yes we need to check nullity and throw because it's spent

if we don't throw, then matchingUnspentUtxo could mean we have insufficient informations

noescape00:
Looks correct to me.

Issue #2873, opened 2018-11-30
PHBS can serve reorged headers


noescape00:
`ProvenBlockHeaderStore.GetAsync` returns items from cache or repository. 

In case there was a reorg and pending batch wasn't saved yet PHBS will return reorged proven headers. 
Batch saving interval is 60 seconds which makes this problem show itself on almost every reorg, 

PHBS should first look into pending batch on GetAsync.

PR #2876, opened 2018-11-30
Fix for "PHBS can serve reorged headers"

noescape00:
GetAsync now looks into pending batch first. 

Also removed GetAsync overload from the codebase because it wasn't implemented properly and would ask repo for headers one by one. Also given that we don't use this method and don't plan to (mostly because we will keep PH always in memory like ChainedHeaders) it wasn't reasonable to fix it over removing it. 

Fix: #2873

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
@@ -160,8 +160,20 @@ namespace Stratis.Bitcoin.Features.Consensus.ProvenBlockHeaders
         {
             using (new StopwatchDisposable(o => this.performanceCounter.AddQueryTime(o)))
             {
+                lock (this.lockObject)
+                {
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
```diff
@@ -174,8 +160,20 @@ public async Task<ProvenBlockHeader> GetAsync(int blockHeight)
         {
             using (new StopwatchDisposable(o => this.performanceCounter.AddQueryTime(o)))
             {
+                lock (this.lockObject)
+                {
+                    if (this.PendingBatch.TryGetValue(blockHeight, out ProvenBlockHeader headerFromBatch))
```
dangershony: Ok yes this fixes the other issue I reported that if cache is depleted the header might be gone.
Unfortunately we need to lock on every call but its cheap as its a dictionary read.

noescape00: Could you please close that other issue if it's fixed? 
Wasn't able to find it

PR #2878, opened 2018-12-02
Mempool default values improved

codingupastorm:
Previous values of 101 kb are kind of nonsensical when maximum tx size is 100kb, at least this gives us some room to move. 

I'm wondering where these values came from though - Bitcoin Core and other clients don't enforce a limit, I think we should follow suit provided there are no known attack vectors

See https://github.com/bitcoin/bitcoin/blob/ed12fd83ca7999a896350197533de5e9202bc2fe/src/txmempool.cpp#L656


fassadlr:
@dangershony @noescape00 I don't see an issue with this seeing as Core doesnt' impose a limit? Do we check the size somewhere else ?

dangershony:
Too risky to look at this now, can it be that BTC have no limit on ancestors?

codingupastorm:
True @zeptin , I'll do what you suggested - tracked by #3019

PR #2882, opened 2018-12-02
Test for rewind data index

MithrilMan:
this test aim to check the RewindDataIndex and RewindData implementation


MithrilMan:
Test logic has to be adjusted because actually when the proven header is validated, it finds a RewindDataIndex and RewindData with nothing to restore.
Probably minerA and minerB aren't enough and I need a 3rd node

PR #2886, opened 2018-12-03
dns seeders to only connect to C# nodes

dangershony:


bokobza:
I don't really like this idea. Why not return all the nodes it finds?

PR #2889, opened 2018-12-03
Feature/default wallet

### LINKED ISSUE #2736: [Wallet] Create configuration to create a default wallet on startup
Bitcoin core by default creates a default unencrypted wallet on startup. To maintain compatibility with NBitcoin tests we would require a couple of configuration parameters to be passed into the node:
1. To create a default wallet on startup
2. To unlock the default wallet on startup

Needed to address https://github.com/MetacoSA/NBitcoin/pull/413

sondreb:
Add support for a default wallet, that is created when -defaultwallet flag is set as either an argument parameter, or configuration option.

Is there any value in calling the LoadWallet, after creation? At least it will verify that it was written to disk correctly, and read from disk correctly, but other than that?

I considered having password as an possible parameter, but I did not want to store the password in clear-text anywhere, which would be needed if I was suppose to auto-unlock the wallet, which was in the "requirement" for the task, but unlocking is not really unlocking, is it? It could of course still be implemented with a password parameter that user must supply on every launch, let me know if you want that option to be added.

Closes #2736


mikedennis:
Hi, yes I need autounlock for the nbitcoin use case. And perhaps a parameter for unlock time. If combined with -defaultwallet  you don't necessarily need to specify the password as a parameter as you know the password already. Might want to have a parameter for unlock duration.

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
```diff
@@ -66,6 +73,8 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
             builder.AppendLine("####Wallet Settings####");
             builder.AppendLine("#Save the hex of transactions in the wallet file. Default: 0.");
             builder.AppendLine("#savetrxhex=0");
+            builder.AppendLine("#Creates a default wallet and unlocks the wallet on startup when set to 1. Default: 0.");
```
mikedennis: I dont think this actually unlocks the wallet? LoadWallet I think just confirms password but doesnt unlock it if I understand correctly.

bokobza:
Nice, thank you @sondreb. I'll review this soon.

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
zeptin: I think Mike is right @sondreb, `LoadWallet` only loads JSON from disk. You would need to call `CacheSecret` on `WalletTransactionHandler` to unlock the default wallet for further RPC access.

sondreb: I have a solution that unlocks the default wallet, the issue though is that it is not very pretty as it needs some state shared between WalletManager and WalletTransactionHandler. When the PR #3017 from @mikedennis is completed, I will refresh this PR and see how I can make it better.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -267,6 +279,18 @@ public Mnemonic CreateWallet(string password, string name, string passphrase, Mn
...
+        private void CreateDefaultWallet(bool exists)
+        {
+            if (!exists)
+            {
+                var mnemonic = new Mnemonic(Wordlist.English, WordCount.Twelve);
+                this.CreateWallet(DefaultWalletPassword, DefaultWalletName, string.Empty, mnemonic);
+            }
+
+            // Load the wallet from disk after creation, to verify it and also verify that we can decrypt it.
+            this.LoadWallet(DefaultWalletPassword, DefaultWalletName);
```
bokobza: When the node starts, all the wallets (files ending in `*.wallet.json`) contained in the data folder are loaded. 
The LoadWallet method only loads wallets that weren't loaded when the node started, for example if you have a running node and you want to load a wallet you brought from another computer.
So when the node starts, the `default.wallet.json` file will be loaded automatically.
I think you should throw a `WalletException` inside the `if` block starting on line 202 if a default wallet already exists.

sondreb: I do think at least some exchanges rely on default wallet feature, which is normally not encrypted, so auto-unlock not needed for those, but since Stratis does not support unencrypted, it must be unlocked at startup to give same features.

I don't think the mnemonic is necessary to return or give to user.

sondreb: Does the stratisX wallet have a default wallet? I know at least one exchange that uses default wallet, if there are others I presume that exchanges must either modify their integrations to support Stratis in the future, or continue to use stratisX?

bokobza: All fair points, thanks for the clarification.
@sondreb can you please address my comments here https://github.com/stratisproject/StratisBitcoinFullNode/pull/2889#discussion_r244854871 about the Load methos?

bokobza: > Does the stratisX wallet have a default wallet?

It works just like Bitcoin not HD.
They shouldn't have to change their processes, normally.

bokobza: @sondreb do you want me to take this one over?

sondreb: As soon as #3017 is merged, I can finish this. It is somewhat dependent on that fix and the ability to set unlock period.

sondreb: I will improve the description of the "defaultwallet" argument. I don't think it should throw exception if it already exists, because I'd like to be able to have a single docker-compose.yml that I can run and re-run without needing any re-configuration and re-deployment. But the description should say "Loads the default wallet on startup. If not exists, will be created automatically."

sondreb: Can I do the following: Modify the GetAccount in RPC controller to retrieve the "default" wallet if it exists, otherwise use the first in the array, as before? It is going to be very weird if someone for some reason has two wallets (which one normally shouldn't do if one want to use RPC, at least not at this stage), and on startup it unlocks default, but when you perform "sendtoaddress", it loads a different wallet...

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
sondreb: Did a refactoring and improved the unlocking, only supports a single wallet but that is fine, as RPC only supports a single wallet for now.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
sondreb: Please check the latest code, should be fairly clear.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -37,13 +37,33 @@ public class WalletRPCController : FeatureController
...
             this.fullNode = fullNode;
             this.broadcasterManager = broadcasterManager;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.walletSettings = walletSettings;
+
+            if (walletSettings.DefaultWallet)
+            {
+                try
+                {
+                    this.UnlockWallet(walletSettings.DefaultWalletPassword, maxDurationInSeconds);
```
mikedennis: What do you think about putting timeout in config instead of hardcoded? Maybe if it's not specified not unlock wallet at all? I'm just thinking there may be value for something in having a default wallet created that is not unlocked for some usecase.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -360,16 +380,27 @@ public async Task<uint256> SendManyAsync(string fromAccount, string addressesJso
...
-
-            HdAccount account = this.walletManager.GetAccounts(walletName).FirstOrDefault();
+            if (this.walletSettings.DefaultWallet)
+            {
+                HdAccount account = this.walletManager.GetAccounts(WalletManager.DefaultWalletName).Single();
+                return new WalletAccountReference(WalletManager.DefaultWalletName, account.Name);
+            }
+            else
+            {
+                //TODO: Support multi wallet like core by mapping passed RPC credentials to a wallet/account
```
mikedennis: I'm not sure what this comment was originally getting at. I think bitcoin core only has a single wallet? If we want to abandon the idea of multiple wallets through RPC completely (I can't see how RPC supports multiple wallets on core), then you may be able to simplify by just first checking if "Default" is there and if not then grabbing the first and then you don't have to have that dependency to WalletSettings.

Accounts is a whole other mess, as seems like bitcoin core has changed the way that works in more recent versions.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
sondreb: That does not sound like a useful feature ;-) If there comes a scenario where default wallet that is locked is useful, I think it should be considered to be implemented then and not now. This covers a use case that is important, while a default wallet that is locked is not something I've heard about or encountered before. If someone has such as scenario, it is likely they can call the RPC methods to get this behavior, or the REST API, don't you think?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
sondreb: This is an existing comment, I have not modified it. I agree it is a bit messy, and I seriously dislike the existing implementation of grabbing the first and best wallet (which it does if no default available).

bokobza: I'm happy with getting the default one first. Grabbing the first is a bit random. How is the first even chosen? is it by alphabetical order or something else? The user doesn't know.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
bokobza: I think the wallet shouldn't be unlocked by default.
Having a default wallet doesn't mean it should be ready to stake or whatever straight away. 
Even in StratisX, the wallet isn't unlocked for staking straight away and we have to call `walletpassphrase`.
Also, I think my first point is valid: if the RPC stuff is loaded before the wallet, and a user doesn't have a default wallet and starts the node with -defaultwallet, it will fail in `WalletRPCController`'s ctor because there is no wallet to unlock yet. As it happens, the wallet stuff is loaded first, so it works, but it's really coincidental.

src/Stratis.Bitcoin.Features.Wallet/WalletSettings.cs
```diff
@@ -66,6 +80,9 @@ public static void BuildDefaultConfigurationFile(StringBuilder builder, Network
...
```
bokobza: It shouldn't unlock it.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
sondreb: "I think the wallet shouldn't be unlocked by default."

Could have yet another parameter that auto-unlocks the wallet, but I don't see the benefit and value of doing this, as there is already RPC method to call and unlock wallet after startup.

There are a few arguments to why not having manual unlock:

- Less secure, because the consumer of RPC needs to have the wallet password stored and sent over an completely insecure network connection. With the default password stored on the machine, it i less accessible from the outside.

- If the user can already do calls to unlock the wallet, the point of a default wallet disappears and they can do multiple calls to first create a wallet (it not exists), and then unlock it in another call.

With a default wallet, it becomes possible to perform payments on a docker instance running the full node, without the need for additional scripts / boot procedures.

Bitcoin Core works with an unencrypted default wallet that allows sending of BTC without any passwords. Since Stratis only does encrypted wallets, this default wallet implementation will "simulate" an unecrypted wallet with a default password.

Regarding the startup logic, could that be improved by having feature dependencies set to ensure that the wallet is loaded before RPC is being made available for calls and unlocking?

bokobza: I think the default wallet is something that we need for integration with NBitcoin who requires a default wallet. Otherwise I would have advised against, I don't particularly like the idea.

Do you know if it needs the wallet to be unlocked?
If we can unlock it with another call, that's my preferred option because it stays aligned with stratisX.
@sondreb @mikedennis

mikedennis: Yes for the nbitcoin tests I need to be able to unlock it through config. I can't change the tests to make an additional rpc call

bokobza: Ok cool. In this case, let's add an `unlockdefaultwallet` argument that only works with defaultwallet and defaultpassword.
@sondreb @mikedennis Can one of you do it soon? otherwise I can do it today.

sondreb: I have implemented the `unlockdefaultwallet`, renamed the `defaultwalletpassword` to `defaultpassword`. Changed the `defaultwallet` to be name, instead of a boolean flag.

Final issue which I'm unsure how to achieve: Making sure that the WalletRPCController constructor is always called after WalletFeature.InitializeAsync. At least during development and debugging, it was never an issue, but that doesn't mean it might be a problem if someone uses or initializes the controller differently in the future.

bokobza: ok that's great, I'll have a more thorough look tomorrow but it looks good at first glance 💪

bokobza:
@sondreb I did 2-3 cosmetic changes. Let's see what the build yields.
I still have 1-2 tests I'd like to do and then we're good to go.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -294,6 +321,45 @@ public Wallet LoadWallet(string password, string name)
             return wallet;
         }
 
+        /// <inheritdoc />
+        public void UnlockWallet(string password, string name, int timeout)
```
mikedennis: I think we want [NoTrace] on any methods with passwords to ensure they don't get written to log file.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -23,6 +26,9 @@ namespace Stratis.Bitcoin.Features.Wallet
     /// </summary>
     public class WalletManager : IWalletManager
     {
+        // <summary>As per RPC method definition this should be the max allowable expiry duration.</summary>
+        private const int maxDurationInSeconds = 1073741824;
```
mikedennis: Maybe clearer name would be maxUnlockDurationInSec or maxSecretCacheDurationInSec

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: indeed

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: 👍

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
bokobza: In fact, I'm wondering. The logging of passwords will only happen when the node is run with `loglevel = trace` and not in the normal run. Tracing is usually used by devs during testing. Do we really want to remove all the logging from the concerned methods?

bokobza: I'll leave this task for another day.

PR #2890, opened 2018-12-04
[RPC] add sendmany

mikedennis:
Part of #2758 

Note: subtractfeefrom is currently not supported because it is not yet supported in WalletTransactionHandler

https://bitcoincore.org/en/doc/0.16.1/rpc/wallet/sendmany/

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
@@ -275,18 +275,18 @@ namespace Stratis.Bitcoin.Features.Wallet
             }
 
             // Outputs addresses are keyvalue pairs of address, amount. Translate to Receipient list.
-            var recipients = new List<Recipient>();
+            IEnumerable<Recipient> recipients = null;
             if (!string.IsNullOrEmpty(addressesJson))
             {
-                JsonConvert.DeserializeObject<List<KeyValuePair<string, decimal>>>(addressesJson).ForEach(address => recipients.Add(new Recipient
+                recipients = JsonConvert.DeserializeObject<Dictionary<string, decimal>>(addressesJson).Select(address => new Recipient
                 {
                     ScriptPubKey = BitcoinAddress.Create(address.Key, this.fullNode.Network).ScriptPubKey,
                     Amount = Money.Coins(address.Value),
-                    SubtractFeeFromAmount = subtractFeeFromAddresses == null ? false : subtractFeeFromAddresses.Any(subAddress => subAddress.ToString().Equals(address))
-                }));
+                    SubtractFeeFromAmount = subtractFeeFromAddresses == null ? false : subtractFeeFromAddresses.Contains(BitcoinAddress.Create(address.Key, this.fullNode.Network))
+                });
             }
 
-            if (!recipients.Any())
+            if (recipients == null || !recipients.Any())
             {
                 throw new RPCServerException(RPCErrorCode.RPC_INVALID_PARAMETER, "No valid output addresses specified.");
             }
@@ -296,7 +296,7 @@ namespace Stratis.Bitcoin.Features.Wallet
                 AccountReference = accountReference,                
                 MinConfirmations = minConf,
                 Shuffle = true, // We shuffle transaction outputs by default as it's better for anonymity.                
-                Recipients = recipients                
+                Recipients = recipients.ToList()                
             };
 
             // Set fee type for transaction build context.
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
@@ -261,6 +261,73 @@ namespace Stratis.Bitcoin.Features.Wallet
             return unspentCoins.ToArray();
         }
 
+        [ActionName("sendmany")]
+        [ActionDescription("Creates and broadcasts a transaction which sends outputs to multiple addresses.")]
+        public async Task<uint256> SendManyAsync(string fromAccount, string addressesJson, int minConf = 1, string comment = null, string subtractFeeFromJson = null, bool isReplaceable = false, int? confTarget = null, string estimateMode = "UNSET")
+        {
+            var accountReference = this.GetAccount();
+           
+            // Optional list of addresses to subtract fees from.
+            IEnumerable<BitcoinAddress> subtractFeeFromAddresses = null;            
+            if (!string.IsNullOrEmpty(subtractFeeFromJson))
+            {
+                subtractFeeFromAddresses = JsonConvert.DeserializeObject<List<string>>(subtractFeeFromJson).Select(i => BitcoinAddress.Create(i, this.fullNode.Network));
+            }
+
+            // Outputs addresses are keyvalue pairs of address, amount. Translate to Receipient list.
+            var recipients = new List<Recipient>();
+            if (!string.IsNullOrEmpty(addressesJson))
+            {
+                JsonConvert.DeserializeObject<List<KeyValuePair<string, decimal>>>(addressesJson).ForEach(address => recipients.Add(new Recipient
+                {
+                    ScriptPubKey = BitcoinAddress.Create(address.Key, this.fullNode.Network).ScriptPubKey,
+                    Amount = Money.Coins(address.Value),
+                    SubtractFeeFromAmount = subtractFeeFromAddresses == null ? false : subtractFeeFromAddresses.Any(subAddress => subAddress.ToString().Equals(address))
+                }));
+            }
+
+            if (!recipients.Any())
+            {
+                throw new RPCServerException(RPCErrorCode.RPC_INVALID_PARAMETER, "No valid output addresses specified.");
+            }
+          
+            var context = new TransactionBuildContext(this.fullNode.Network)
+            {
+                AccountReference = accountReference,                
+                MinConfirmations = minConf,
+                Shuffle = true, // We shuffle transaction outputs by default as it's better for anonymity.                
+                Recipients = recipients                
+            };
+
+            // Set fee type for transaction build context.
+            context.FeeType = FeeType.Medium;
+            if (estimateMode.Equals("ECONOMICAL", StringComparison.InvariantCultureIgnoreCase))
+                context.FeeType = FeeType.Low;
+            else if (estimateMode.Equals("CONSERVATIVE", StringComparison.InvariantCultureIgnoreCase))
+                context.FeeType = FeeType.High;
+
+            try
+            { 
+                // Log warnings for currently unsupported parameters.
+                if (!string.IsNullOrEmpty(comment))
+                    this.logger.LogWarning("'comment' parameter is currently unsupported. Ignored.");
+                if (isReplaceable)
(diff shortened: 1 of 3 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+        [ActionDescription("Creates and broadcasts a transaction which sends outputs to multiple addresses.")]
+        public async Task<uint256> SendManyAsync(string fromAccount, string addressesJson, int minConf = 1, string comment = null, string subtractFeeFromJson = null, bool isReplaceable = false, int? confTarget = null, string estimateMode = "UNSET")
+        {
+            var accountReference = this.GetAccount();
+           
+            // Optional list of addresses to subtract fees from.
+            IEnumerable<BitcoinAddress> subtractFeeFromAddresses = null;            
+            if (!string.IsNullOrEmpty(subtractFeeFromJson))
+            {
+                subtractFeeFromAddresses = JsonConvert.DeserializeObject<List<string>>(subtractFeeFromJson).Select(i => BitcoinAddress.Create(i, this.fullNode.Network));
```
fassadlr: Perhaps wrap this in a try catch if the Json is malformed and return null?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
             return unspentCoins.ToArray();
         }
 
+        [ActionName("sendmany")]
+        [ActionDescription("Creates and broadcasts a transaction which sends outputs to multiple addresses.")]
+        public async Task<uint256> SendManyAsync(string fromAccount, string addressesJson, int minConf = 1, string comment = null, string subtractFeeFromJson = null, bool isReplaceable = false, int? confTarget = null, string estimateMode = "UNSET")
+        {
+            var accountReference = this.GetAccount();
```
fassadlr: You can declare this variable closer to where it is used.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
fassadlr: Also, can someone submit an empty list?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+            if (!string.IsNullOrEmpty(subtractFeeFromJson))
+            {
+                subtractFeeFromAddresses = JsonConvert.DeserializeObject<List<string>>(subtractFeeFromJson).Select(i => BitcoinAddress.Create(i, this.fullNode.Network));
+            }
+
+            // Outputs addresses are keyvalue pairs of address, amount. Translate to Receipient list.
+            IEnumerable<Recipient> recipients = null;
+            if (!string.IsNullOrEmpty(addressesJson))
+            {
+                recipients = JsonConvert.DeserializeObject<Dictionary<string, decimal>>(addressesJson).Select(address => new Recipient
```
fassadlr: Same here perhaps protect the deserialization with a try catch.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+            if (estimateMode.Equals("ECONOMICAL", StringComparison.InvariantCultureIgnoreCase))
+                context.FeeType = FeeType.Low;
+            else if (estimateMode.Equals("CONSERVATIVE", StringComparison.InvariantCultureIgnoreCase))
+                context.FeeType = FeeType.High;
+
+            try
+            { 
+                // Log warnings for currently unsupported parameters.
+                if (!string.IsNullOrEmpty(comment))
+                    this.logger.LogWarning("'comment' parameter is currently unsupported. Ignored.");
```
fassadlr: Change to:

this.logger.LogWarning("'{0}' parameter is currently unsupported. Ignored.", nameof(comment));

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+                if (isReplaceable)
+                    this.logger.LogWarning("'replaceable' parameter is currently unsupported. Ignored.");
```
fassadlr: Use `nameof` here as well.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -261,6 +261,73 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+                if (confTarget != null)
+                    this.logger.LogWarning("'conf_target' parameter is currently unsupported. Ignored.");
```
fassadlr: Use `nameof` here as well.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -181,5 +183,65 @@ public void TestRpcSendToAddressCantSpendWhenLocked()
                 Assert.Throws<RPCException>(() => rpcClient.SendToAddress(aliceAddress, Money.Coins(1.0m)));
             }
         }
+
+        [Fact]
+        public void TestRpcSendManyWithValidDataIsSuccessful()
```
fassadlr: Perhaps add some tests for malformed json or where the send fail?

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: replaceable is the name of the rpc parameter in rpc docs. The parameter name in this method doesn't matter as the order the parameters are passed is how it determines which one is which.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
mikedennis: Ditto here, rpc parameter name is not the same as method parameter name

mikedennis:
@fassadlr No can 'o worms thankfully! Added some more error handling and a test as per your suggestions.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
```diff
@@ -181,5 +184,91 @@ public void TestRpcSendToAddressCantSpendWhenLocked()
...
+                txid.Should().Be(tx.GetHash());                
+
+                // Check the output is the right amount.
+                var aliceCoins = tx.Outputs.AsCoins().First(c => c.ScriptPubKey == aliceAddress.ScriptPubKey);
+                aliceCoins.Amount.Should().Be(Money.Coins(coinsForAlice));
+                
+                var bobCoins = tx.Outputs.AsCoins().First(c => c.ScriptPubKey == bobAddress.ScriptPubKey);
+                bobCoins.Amount.Should().Be(Money.Coins(coinsForBob));
+
+                // TODO: Enable test below once WalletTransactionHandler supports recepients with SubtractFeeFromAmount=true
```
dangershony: @mikedennis can you create an issue to add support to this (unless there is one already)

fassadlr:
Hi Mike,

I've implemented some extra checks as per the core source code:

https://github.com/bitcoin/bitcoin/blob/86e0a33f5c382513d5179e3fdf158baf952d7e2f/src/wallet/rpcwallet.cpp#L785

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -265,72 +265,100 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+                    ScriptPubKey = recipientAddress,
+                    Amount = amount,
+                    SubtractFeeFromAmount = subtractFeeFromAddresses == null ? false : subtractFeeFromAddresses.Contains(BitcoinAddress.Create(address.Key, this.fullNode.Network))
+                };
+
+                recipients.Add(recipient);
             }
 
             var accountReference = this.GetAccount();
+            var spendable = this.walletManager.GetSpendableTransactionsInAccount(accountReference, minConf);
```
dangershony: Is this check not made inside the create trx?

mikedennis: Double checked, negative amount will throw an ArgumentOutOfRange exception in wallet. Duplicate recipient is an interesting one, I don't think the wallet cares about that as it's valid to send to the same address more than once (a bit dumb but valid). So I guess we can do this check in the RPC if core does.

fassadlr: > Double checked, negative amount will throw an ArgumentOutOfRange exception in wallet. 

OK sweet, I will remove this check then 👍

fassadlr:
@mikedennis let me if you are happy and we can push this once build passes 👍

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
```diff
@@ -253,21 +253,126 @@ public UnspentCoinModel[] ListUnspent(int minConfirmations = 1, int maxConfirmat
...
+            var recipients = new List<Recipient>();
+            foreach (var address in addresses)
+            {
+                // Check for duplicate recipients
+                var recipientAddress = BitcoinAddress.Create(address.Key, this.fullNode.Network).ScriptPubKey;
+                if (recipients.Any(r => r.ScriptPubKey == recipientAddress))
+                    throw new RPCServerException(RPCErrorCode.RPC_INVALID_PARAMETER, string.Format("Invalid parameter, duplicated address: {0}.", recipientAddress));
+
+                var amount = Money.Coins(address.Value);
+                if (amount <= 0)
```
mikedennis: This is checked when the transaction is built and ArgumentOutOfRange exception is thrown by wallet. Could potentially replace this check with a catch block for this exception.

src/Stratis.Bitcoin.IntegrationTests/RPC/RPCTestsMutable.cs
mikedennis: Done. Added #3023 issue

mikedennis:
@fassadlr Looks good, just one suggestion that you may be potentially able to replace the one check with just an exception handler.

src/Stratis.Bitcoin.Features.Wallet/WalletRPCController.cs
fassadlr: Ok yeah I have now also removed this check 👍

Issue #2842, opened 2018-11-27
SBFN PH: StratisX nodes are not syncing from SBFN nodes


maciejzaleski:
This test is based on the latest network topology: https://github.com/maciejzaleski/InternalTestnet/blob/master/Documentation/FullNode/InternalTestnet-NetworkDesign.draw.io.svg

**Setup**
 - Started network with NodeJFX fully synced
 - Waited for all nodes to sync; this completed successfully for all standard SBFN nodes and gateway nodes but failed for StratisX. StratisX will fully sync if it is connected to another StratisX node (NodeJGX). 
 - The problem seems to be present ONLY if StratisX is syncing blocks post IBD
 - The IBD block download (SBFN -> StratisX) seems to be running flawlessly

**Logs**:  https://stratisplatform-my.sharepoint.com/:u:/p/maciej_zaleski/EcMzTK1Y_dVCtwTpkxGH96EB0wn0YJJWj_bK-huvz8Khfg?e=de7g7L

**Network View**
![image](https://user-images.githubusercontent.com/6422181/49113233-e53b1d80-f28c-11e8-96f2-3100e9ee6808.png)

noescape00:
Looks like I got lucky that time. Resynced QT 4 times from C# node with full logging enabled but the bug didn't appear. 

@maciejzaleski is it easily reproducible on InternalTestnet?

noescape00:
I was able to reproduce a bug several times but it never lead to QT node being stuck for longer than 3-5 minutes. 

When QT syncs from C# node it can occasionally drop the connection `Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
 `

after which it would reconnect and it would take some time to start syncing again. 

I've also checked logs provided by @maciejzaleski and didn't find there a disconnection which was a precursor to the situation I found somewhat problematic.

However there might be a problem in the setup that might have caused the bug- bidirectional connections. 

HA:
```
Peer:[::ffff:192.168.98.101]:37021,      connected:inbound,         height:1241,          agent:StratisBitcoin:1.2.5
Peer:[::ffff:192.168.98.101]:16178,      connected:outbound,        height:1241,          agent:StratisBitcoin:1.2.5
Peer:[::ffff:192.168.98.171]:43076,      connected:inbound,         height:0,             agent:/Stratis:2.0.0.5/
```

HBX:

```
[...]

PR #2893, opened 2018-12-04
CMB logs and bugfix

noescape00:
Fix: #2842

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
@@ -105,18 +105,6 @@ namespace Stratis.Bitcoin.Features.Consensus.Behaviors
         }
 
         /// <inheritdoc />
-        public override async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync()
-        {
-            if (this.consensusManager.Tip.Height == this.checkpoints.GetLastCheckpointHeight())
-            {
-                this.logger.LogDebug("Sync requested because last checkpointed block was connected.");
-                await this.ResyncAsync().ConfigureAwait(false);
-            }
-
-            return await base.ConsensusTipChangedAsync().ConfigureAwait(false);
-        }
-
-        /// <inheritdoc />
         protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersPayload, out ChainedHeader lastHeader)
         {
             // If getHeadersPayload isn't a GetProvenHeadersPayload, return base implementation result
--- a/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
@@ -143,6 +143,7 @@ namespace Stratis.Bitcoin.Features.Consensus.Behaviors
             var provenHeadersPayload = new ProvenHeadersPayload();
 
             ChainedHeader header = this.GetLastHeaderToSend(fork, getHeadersPayload.HashStop);
+            this.logger.LogDebug("Last header that will be sent in headers payload is '{0}'.", header);
 
             for (int heightIndex = header.Height; heightIndex > fork.Height; heightIndex--)
             {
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -104,6 +104,18 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
             }
         }
 
+        /// <inheritdoc />
+        public override async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync()
```
dangershony: I don't really think this is needed, we anyway trigger a resync at the end of header presented

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -143,6 +143,7 @@ protected override Payload ConstructHeadersPayload(GetHeadersPayload getHeadersP
             var provenHeadersPayload = new ProvenHeadersPayload();
 
             ChainedHeader header = this.GetLastHeaderToSend(fork, getHeadersPayload.HashStop);
+            this.logger.LogDebug("Last header that will be sent in headers payload is '{0}'.", header);
```
dangershony: Do we want to keep all those debug logs?

noescape00: Yeah. 
I'm trying to separate logs to trace and debug because trace logs are too verbose but we have nothing on debug level.

dangershony: We probably need to also do that in many other places int eh code
We have no formal way to decide when we use debug logs

src/Stratis.Bitcoin/Consensus/ConsensusExceptions.cs
```diff
@@ -28,6 +28,14 @@ public ConnectHeaderException() : base()
         }
     }
 
+    /// <summary>Can validate PH yet.</summary>
```
fassadlr: This should be "can't"?

src/Stratis.Bitcoin/Consensus/ConsensusExceptions.cs
```diff
@@ -28,6 +28,14 @@ public ConnectHeaderException() : base()
...
+    public class InsufficientHeaderInformation : ConsensusException
```
fassadlr: Lets rename this to `InsufficientHeaderInformationException` just so that we stick to convention ;)

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -461,10 +473,11 @@ protected async Task<ConnectNewHeadersResult> PresentHeadersLockedAsync(List<Blo
             return result;
         }
 
-        /// <summary>Resyncs the peer whenever state is changed.</summary>
+        /// <summary>Sync when handshake is finished.</summary>
         private async Task OnStateChangedAsync(INetworkPeer peer, NetworkPeerState oldState)
         {
-            await this.ResyncAsync().ConfigureAwait(false);
+            if (peer.State == NetworkPeerState.HandShaked)
```
dangershony: State is already checked inside ResyncAsync so not a big deal here

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -99,7 +99,7 @@ public async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync()
                 {
                     result = await this.PresentHeadersLockedAsync(this.cachedHeaders, false).ConfigureAwait(false);
 
-                    if (result == null)
+                    if ((result == null) || (result.Consumed == null))
```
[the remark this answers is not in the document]
noescape00: fixed

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
noescape00: Yeah, added another check because this always bothers me. It's weird to resync on every state change.

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -99,8 +99,11 @@ public async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync()
...
                     {
+                        if (result == null)
+                            this.cachedHeaders.Clear();
```
dangershony: Should we not trigger a sync as well? set syncRequired = true

PR #2894, opened 2018-12-04
[P2P] Move IP range filtering to PeerConnectorDiscovery and Update ConnectionManager disconnect

### LINKED ISSUE #2892: If the node connection fail nodes will not be able to reconnect
Nodes cannot establish a connection even though they are in different IP address groups

It seems to fail in in `Stratis.Bitcoin.P2P.PeerConnector.PeerIsPartOfExistingGroup` method
It looks like the peers are added to the `this.connectedPeers` but never removed. So that if the first connection is unsuccessful the peer will remain in the group and new connection will fail because there is already a connection for that group.

I have added more logging to the NBitcoin `GetGroup` extension method to help investigate.
[IpGroupingLogging.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2643032/IpGroupingLogging.zip)

**Setup**
- **NodeAD**: 100.67.254.1
- **NodeJA**: 100.64.9.1

**Logs**
[Runtime_Logs.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2643028/Runtime_Logs.zip)

**Network status**
status

**Network packet exchange between NodeAD and NodeJA**
The first two handshakes occurred organically by NodeJA connecting to NodeAD. The remaining handshakes have been triggered by using Swagger API call `addnode`
image

**Code version**
```
   - NodeJA: 88b4e38e (03 Dec @ 21:44); Patches: Kevin,IpGroupingLogging 
   
*** Build date: 03/12/2018 22:42:01.07 

*** Recent commits 
*    88b4e38e - (HEAD -> master, origin/master, origin/HEAD) testCHT (#2857) (Mon, 3 Dec 2018 21:44:59 +0300) <noescape0>
*    1c827a86 - Fix bug when PH would be serialized instead of BlockHeader (#2887) (Mon, 3 Dec 2018 20:45:26 +0300) <noescape0>
*    3e27a734 - Disconnect after handshake if already have a peer on the same group (#2869) (Mon, 3 Dec 2018 20:41:45 +0300) <noescape0>
*    7ad3f4b9 - Bump to Version 1.2.6-beta (#2885) (Mon, 3 Dec 2018 09:45:30 +0000) <Jeremy Bokobza>
*    7639e1ae - [SC] Display history with empty block numbers first (#2798) (Mon, 3 Dec 2018 16:17:46 +1100) <Rowan de Haas>
*    58d3d1dd - Fix consecutive header bug (#2865) (Sat, 1 Dec 2018 13:54:20 +0000) <Francois de la Rouviere>
*    700c871a - Correctly store and load rewind data index,  (#2875) (Fri, 30 Nov 2018 23:46:54 +0000) <Dan Gershony>
*    c63e0c36 - Fix for "PHBS can serve reorged headers" (#2876) (Fri, 30 Nov 2018 18:13:46 +0300) <noescape0>
*    4d05de89 - Update SeedData (#2872) (Fri, 30 Nov 2018 10:04:03 +0000) <StratisIain>
*    08dc1140 - (origin/sc/v0.12.0-beta) nameof variables fix (#2868) (Thu, 29 Nov 2018 16:48:10 +0000) <Fazz>  
******************
```

fassadlr:
The issue was that we disconnect the peer in `ConnectionManager` too late. We disconnect it AFTER its been added to the `connectedPeers` collection. So it connects, we call `OnPeerAdded` and we call `ShouldDisconnect` and it disconnects the JUST added peer.

Also, we dont check for localhost addresses so even those get disconnected.

Fixes #2892


src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -123,7 +120,6 @@ public IReadOnlyNetworkPeerCollection ConnectedPeers
             this.Parameters.UserAgent = $"{this.ConnectionSettings.Agent}:{versionProvider.GetVersion()} ({(int)this.NodeSettings.ProtocolVersion})";
 
             this.Parameters.Version = this.NodeSettings.ProtocolVersion;
-            this.connectedPeersQueue = new AsyncQueue<INetworkPeer>(this.OnPeerAdded);
```
dangershony: If you delete this you introduce back the race condition, the async queue processes single threaded all attempts to check the filtering.

fassadlr: My point still stands that we can't check the range filtering after we have already added the peer to the `connectedPeers` collection. So currently we will always disconnect straight after we've added as the peer we are checking is already in the collection.

dangershony:
I am fine with this change @noescape00

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
noescape00: >So currently we will always disconnect straight after we've added as the peer we are checking is already in the collection.

No, because we ignore current peer during checking the groups: 

```
if (peer == connectedPeer)
    continue;
```

noescape00:
Removing AsyncQueue reintroduces a race condition

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
fassadlr: @noescape00 yes I wrongly assumed there :)

fassadlr:
> Removing AsyncQueue reintroduces a race condition

@dangershony and I decided to go with a lock 👍

fassadlr:
@noescape00 can you explain the race condition still after using the lock?

dangershony:
Ah the problem is exactly this 
```
if (peer == connectedPeer)
    continue;
```

In case we are already connected this will never work? why do skip this check?

fassadlr:
> 
> 
> Ah the problem is exactly this
> 
> ```
> if (peer == connectedPeer)
>     continue;
> ```
> 
> In case we are already connected this will never work? why do skip this check?

It is not skipped??

dangershony:
It is skipped and if we are the same peer we return false (don't disconnect)

PR #2901, opened 2018-12-05
syncing speedup

noescape00:


noescape00:
Don't merge yet. 
Need to test it after 2893 goes in

MithrilMan:
re-evaluating stuck headers every 100 block doesn't sound right to me, can't we have a timer instead, if we don't want to re-evaluate every block?

noescape00:
Not sure if timer is better- different nodes sync with different speed so the timer might have interval that is too short or too long depending on node.

src/Stratis.Bitcoin/Consensus/ConsensusManagerBehavior.cs
```diff
@@ -95,7 +95,7 @@ public async Task<ConnectNewHeadersResult> ConsensusTipChangedAsync()
 
             using (await this.asyncLock.LockAsync().ConfigureAwait(false))
             {
-                if (this.cachedHeaders.Count != 0)
+                if (this.cachedHeaders.Count != 0 && this.CanConsumeCache())
```
dangershony: Should we not disable this if `count == 1`?

noescape00: No. I think this case is handled before header even makes it to the cashe

noescape00:
>Don't merge yet.
>Need to test it after 2893 goes in

Tested. It works

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -104,6 +104,29 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
 
+        /// <inheritdoc />
+        protected override bool CanConsumeCache()
+        {
+            int height = this.consensusManager.Tip.Height;
+
+            // Try cashe consumption every N blocks advanced by consensus.
+            // N should be between 0 and max reorg. 20% of max reorg is a good random value.
+            // Higher the N the better performance boost we can get.
+            uint tryCacheEveryBlocksCount = this.network.Consensus.MaxReorgLength / 5;
```
dangershony: this calculation can go inside the if bellow

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
```diff
@@ -104,6 +104,29 @@ protected override async Task OnMessageReceivedAsync(INetworkPeer peer, Incoming
...
+
+            if (height == this.lastCheckpointHeight)
+                return true;
+
+            if (height > this.lastCheckpointHeight)
+            {
+                // After last checkpoint.
+                if ((height % tryCacheEveryBlocksCount == 0) || (height >= this.BestReceivedTip.Height))
```
dangershony: Is there a possibility BestReceivedTip can be null? (probably not as first run headers will always be found)

dangershony: How can this be possible?
`height >= this.BestReceivedTip.Height` perhaps equal to is possible  (so I guess for sanity also check greater then)

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
noescape00: Fixed

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
noescape00: >Is there a possibility BestReceivedTip can be null? (probably not as first run headers will always be found)

Good spot. Yeah, I think it's possible. 
Fixed

noescape00: >How can this be possible?
height >= this.BestReceivedTip.Height perhaps equal to is possible (so I guess for sanity also check greater then)

It's possible when peer is behind our consensus.

noescape00:
@dangershony fixes applied

src/Stratis.Bitcoin.Features.Consensus/Behaviors/ProvenHeadersConsensusManagerBehavior.cs
dangershony: ah yeah sure in case the peer is in a fork (within the max reorg) and trying to connect a longer chain it will be under tip indeed very well.

Issue #2903, opened 2018-12-05
unused payloads


MithrilMan:
In our codebase we have unused payloads.
I think we should remove them before releasing the fullnode.

some unused payloads I've found are:

> UTxOutputPayload
> FilterAddPayload
> FilterLoadPayload

noescape00:
Agreed. Please create a PR

MithrilMan:
I'm working on a PR and I've find other unused payloads:
BlockTxnPayload
CmpctBlockPayload
GetBlockTxnPayload
MerkleBlockPayload

SendCmpctPayload is not used but it's wired to this code
https://github.com/stratisproject/StratisBitcoinFullNode/blob/a6a8d2b743f903d3aafd0f65614ac63b780c10ca/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreBehavior.cs#L288-L292

I left it there but probably could be removed also (or implemented, but then we should implement all cmpct payloads)

PR #2909, opened 2018-12-06
unused payloads

MithrilMan:
removed unused payloads
see #2903


noescape00:
Please merge if you agree with keeping notfoundpayload. Otherwise remove it and lets discuss.

dangershony:
Looks good to me as well, but I suggest a network sync just in case (testnet is enough)

noescape00:
>but I suggest a network sync just in case 

Imho not needed- none of those payloads were referenced anywhere in our code. 
I've built the solution and found no build errors.

bokobza:
good stuff.

PR #2919, opened 2018-12-06
[Mempool] Adjustments to mempool tests

zeptin:
Some tests were marked as requiring #2474 before they could be completed.

However, transaction rejection for transactions newer than the future drift time is already separately tested in the integration tests. It is not entirely clear what was meant by 'expired transaction times' in some of the unimplemented test names, so these have been removed.

Positive and negative integration tests for non-relative locktimes have also been added.

### DIFF
--- a/src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
+++ b/src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
@@ -24,42 +24,8 @@ namespace Stratis.Bitcoin.Features.MemoryPool.Tests
         {
         }
 
-        [Fact(Skip = "Awaiting fix for issue #2474")]
-        public async void CheckFinalTransaction_WithStandardLockTimeAndValidTxTime_ReturnsTrueAsync()
-        {
-            string dataDir = GetTestDirectoryPath(this);
-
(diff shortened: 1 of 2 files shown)


src/Stratis.Bitcoin.Features.MemoryPool.Tests/MempoolValidatorTest.cs
```diff
@@ -80,55 +46,14 @@ public async void CheckFinalTransaction_WithNoLockTimeAndValidTxTime_ReturnsTrue
...
-            tx.Sign(network, minerSecret, false);
-
-            var state = new MempoolValidationState(false);
-
-            bool isSuccess = await validator.AcceptToMemoryPool(state, tx);
-            Assert.True(isSuccess, "Transaction with no nLockTime and a valid transaction time should have been accepted.");
-        }
-
-        [Fact(Skip = "Awaiting fix for issue #2474")]
-        public async void CheckFinalTransaction_WithStandardLockTimeAndExpiredTxTime_FailsAsync()
```
dangershony: Why delete it though?

zeptin: I'm not sure what an 'expired' transaction time is meant to be; if there is a necessary test case here that's missing then I can implement it

PR #2923, opened 2018-12-07
[SC] Remove wallet requirement for local execution

### LINKED ISSUE #2908: [SC] Local method execution shouldn't require a wallet to be set up. If some parameters aren't given or available, just use defaults.

rowandh:
Closes #2908.

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
@@ -1,0 +1,42 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Text;
+using Stratis.Bitcoin.Features.SmartContracts.ReflectionExecutor.Consensus.Rules;
+using Stratis.Bitcoin.Features.Wallet.Validations;
+
+namespace Stratis.Bitcoin.Features.SmartContracts.Models
+{
+    public class LocalCallContractRequest
+    {
+        [Required(ErrorMessage = "A destination address is required.")]
+        [IsBitcoinAddress]
+        public string ContractAddress { get; set; }
+
+        [Required(ErrorMessage = "A method name is required.")]
+        public string MethodName { get; set; }
+
+        [Required(ErrorMessage = "An amount is required.")]
+        public string Amount { get; set; }
+
+        [Range(SmartContractFormatRule.GasPriceMinimum, SmartContractFormatRule.GasPriceMaximum)]
+        public ulong GasPrice { get; set; }
+
+        [Range(SmartContractFormatRule.GasLimitCallMinimum, SmartContractFormatRule.GasLimitMaximum)]
+        public ulong GasLimit { get; set; }
+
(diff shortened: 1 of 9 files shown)


src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+        [Required(ErrorMessage = "An amount is required.")]
+        public string Amount { get; set; }
+
+        [Range(SmartContractFormatRule.GasPriceMinimum, SmartContractFormatRule.GasPriceMaximum)]
+        public ulong GasPrice { get; set; }
+
+        [Range(SmartContractFormatRule.GasLimitCallMinimum, SmartContractFormatRule.GasLimitMaximum)]
+        public ulong GasLimit { get; set; }
+
+        [Required(ErrorMessage = "Sender is required.")]
```
codingupastorm: Does this need to be Required? Could we inject a uint160(0) or so instead?

src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        [Required(ErrorMessage = "A destination address is required.")]
+        [IsBitcoinAddress]
+        public string ContractAddress { get; set; }
+
+        [Required(ErrorMessage = "A method name is required.")]
+        public string MethodName { get; set; }
+
+        [Required(ErrorMessage = "An amount is required.")]
+        public string Amount { get; set; }
```
codingupastorm: Also think this doesn't need to be required. Most common use of `local-call` will be to call something like `TotalSupply`. What do you think?

src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string Sender { get; set; }
+
+        public string[] Parameters { get; set; }
+
+        public override string ToString()
+        {
+            var builder = new StringBuilder();
+
+            builder.Append(string.Format("{0}:{1},{2}:{3},", nameof(this.GasPrice), this.GasPrice, nameof(this.GasLimit), this.GasLimit));
+            builder.Append(string.Format("{0}:{1},{2}:{3},", nameof(this.Sender), this.Sender, nameof(this.Parameters), this.Parameters));
```
codingupastorm: Extra comma on end of format string. Breaking change.

src/Stratis.Bitcoin.Features.SmartContracts/Models/LocalCallContractRequest.cs
rowandh: I think it's necessary. Consider times you might want to do a calculation with the amount transferred and return that as a result.

rowandh: Removed, though GitHub is not showing it as outdated here.

PR #2924, opened 2018-12-07
fixed FeeRate rounding issue 

MithrilMan:
fixed FeeRate rounding issue

### DIFF
--- a/src/NBitcoin/FeeRate.cs
+++ b/src/NBitcoin/FeeRate.cs
@@ -41,7 +41,7 @@ namespace NBitcoin
             if(feePaid.Satoshi < 0)
                 throw new ArgumentOutOfRangeException("feePaid");
             if(size > 0)
-                this._FeePerK = feePaid / size * 1000;
+                this._FeePerK = (long)(feePaid.Satoshi / (float)size * 1000);
             else
                 this._FeePerK = 0;
         }
--- a/src/NBitcoin/FeeRate.cs
+++ b/src/NBitcoin/FeeRate.cs
@@ -41,7 +41,7 @@ namespace NBitcoin
             if(feePaid.Satoshi < 0)
                 throw new ArgumentOutOfRangeException("feePaid");
             if(size > 0)
-                this._FeePerK = (long)(feePaid.Satoshi / (float)size * 1000);
+                this._FeePerK = (long)(feePaid.Satoshi / (decimal)size * 1000);
             else
                 this._FeePerK = 0;
         }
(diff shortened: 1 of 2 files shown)


src/NBitcoin/FeeRate.cs
```diff
@@ -41,7 +41,7 @@ public FeeRate(Money feePaid, int size)
             if(feePaid.Satoshi < 0)
                 throw new ArgumentOutOfRangeException("feePaid");
             if(size > 0)
-                this._FeePerK = feePaid / size * 1000;
+                this._FeePerK = (long)(feePaid.Satoshi / (float)size * 1000);
```
noescape00: Maybe

```
this._FeePerK = new Money(feePaid.Satoshi * (1000.0m / size), MoneyUnit.Satoshi);
```

to avoid using float

MithrilMan: I had to change to decimal because float was wrong (still rounding issue)
btw it uses Money implicit operator but I can use Money constructor too if needed

MithrilMan: btw yeah I prefer Money constructor that accepts decimal, good point

MithrilMan: on the other side, money decimal constructor does conversions, while actually I don't need the conversion

src/NBitcoin/FeeRate.cs
```diff
@@ -41,7 +41,7 @@ public FeeRate(Money feePaid, int size)
             if(feePaid.Satoshi < 0)
                 throw new ArgumentOutOfRangeException("feePaid");
             if(size > 0)
-                this._FeePerK = feePaid / size * 1000;
+                this._FeePerK = (long)(feePaid.Satoshi / (decimal)size * 1000);
```
noescape00:
```
this._FeePerK = (long)(feePaid.Satoshi / ((decimal)size * 1000));
```

Brackets missing. Otherwise it doesn't fix the problem. 

Ps
Please add a test for it.

dangershony: yes a test is a good idea

MithrilMan: why you say it doesn't fix?

MithrilMan: agree on test

MithrilMan: well after testing it, it's working as intended.
Actually it will brake if you max out (long.MaxValue) left value because if size < 1000 it will overflow no matter where you put brackets

noescape00: Ah, yeah, right. 
Stupid mistake from me.

PR #2925, opened 2018-12-08
Wallet API refactoring and added new API method

noescape00:
There were several request models that had account and wallet name duplicated. Inherited them from a new model- WalletAccountNameModel. 

Added API method: SplitCoins
It's useful for splitting a huge coin when starting staking to maximize returns. Also useful for tests that involve solo staking or pos network bootstrapping.


noescape00:
@dangershony I also have plans to add wallet controller with dev tools where we will have tx flooder for example. Having this on API will enable us to test flooding easily on testnet and internal test tool (there is a scenario for tx flooding but I don't think it's implemented, so having this might be of use).

dangershony:
May be a good idea to take some of the existing pow tests and convert it to pos

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -530,7 +530,7 @@ public IActionResult GetHistory([FromQuery] WalletHistoryRequest request)
         /// <returns></returns>
         [Route("balance")]
         [HttpGet]
-        public IActionResult GetBalance([FromQuery] WalletBalanceRequest request)
+        public IActionResult GetBalance([FromQuery] WalletAccountNameModel request)
```
dangershony: All wallet tests are passing? changing the type does not break anything?

src/Stratis.Bitcoin.Features.Wallet/Models/RequestModels.cs
```diff
@@ -24,6 +24,15 @@ public override string ToString()
         }
     }
 
+    public class WalletAccountNameModel : RequestModel
```
bokobza: NACK, wallet name and account name are not both required all the time.
These properties are not inheriting from one another on purpose.

noescape00:
@bokobza good point. Fixed.

src/Stratis.Bitcoin.Features.Wallet/Controllers/WalletController.cs
```diff
@@ -1172,6 +1172,51 @@ public IActionResult SyncFromDate([FromBody] WalletSyncFromDateRequest request)
...
         }
 
+        /// <summary>Creates requested amount of UTXOs each of equal value.</summary>
+        [HttpPost]
+        [Route("splitcoins")]
+        public IActionResult SplitCoins([FromBody] SplitCoinsRequest request)
+        {
+            Guard.NotNull(request, nameof(request));
+
+            // checks the request is valid
```
bokobza: Capital 'C' and full stop at the end please 😉

Issue #2430, opened 2018-10-09
StakeValidator does not have any unit and/or integration tests


fenix2222:
Add test coverage for `StakeValidator` as this is a very important part of the consensus code and it must have test coverage.

Neurosploit:
I'll have a look to see what I can do @fenix2222.

Neurosploit:
https://github.com/stratisproject/StratisBitcoinFullNode/pull/2929

PR #2929, opened 2018-12-09
#2430 StakeValidator tests

Neurosploit:
Unit tests for the stakevalidator.
Extracted the minconfirmations into a method and added some guards.


dangershony:
AWESOME.... and so many tests.Obviously this will only go in after the release (as I see hcanges to the validator itself)

Neurosploit:
@dangershony fine with me ;)

PR #2930, opened 2018-12-09
Adding BlockStoreCacheTip property and more debug logs

dangershony:
Attempting to fix the delay in X nodes that can happen sometimes on IBD

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
@@ -80,6 +80,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         private readonly CancellationTokenSource cancellation;
 
+        /// <inheritdoc cref="IBlockStoreQueue.BlockStoreCacheTip"/>
+        public ChainedHeader BlockStoreCacheTip { get; private set; }
+
         public BlockStoreQueue(
             ConcurrentChain chain,
             IChainState chainState,
@@ -162,7 +165,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 throw new BlockStoreException("Block store initialized after consensus!");
             }
 
-            this.chainState.BlockStoreCacheTip = initializationTip;
+            this.BlockStoreCacheTip = initializationTip;
 
             // Start dequeuing.
             this.currentBatchSizeBytes = 0;
@@ -308,7 +311,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     this.logger.LogTrace("Block '{0}' was re-added to pending.", chainedHeaderBlock.ChainedHeader);
                 }
 
-                this.chainState.BlockStoreCacheTip = chainedHeaderBlock.ChainedHeader;
+                this.BlockStoreCacheTip = chainedHeaderBlock.ChainedHeader;
             }
 
             this.blocksQueue.Enqueue(chainedHeaderBlock);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
@@ -80,7 +80,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         private readonly CancellationTokenSource cancellation;
 
-        /// <inheritdoc cref="IBlockStoreQueue.BlockStoreCacheTip"/>
+        /// <inheritdoc/>
         public ChainedHeader BlockStoreCacheTip { get; private set; }
 
         public BlockStoreQueue(
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
@@ -162,6 +162,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 throw new BlockStoreException("Block store initialized after consensus!");
             }
 
+            this.chainState.BlockStoreCacheTip = initializationTip;
+
             // Start dequeuing.
             this.currentBatchSizeBytes = 0;
             this.dequeueLoopTask = this.DequeueBlocksContinuouslyAsync();
@@ -305,6 +307,8 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     this.pendingBlocksCache.Add(chainedHeaderBlock.ChainedHeader.HashBlock, chainedHeaderBlock);
                     this.logger.LogTrace("Block '{0}' was re-added to pending.", chainedHeaderBlock.ChainedHeader);
                 }
+
+                this.chainState.BlockStoreCacheTip = chainedHeaderBlock.ChainedHeader;
             }
 
             this.blocksQueue.Enqueue(chainedHeaderBlock);
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
@@ -26,7 +26,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
                 int distanceFromConsensusTip = this.chainState.ConsensusTip.Height - this.chainState.BlockStoreTip.Height;
 
-                // Once store is 5 blocks form the consensus tip then flush on every block.
+                // Once store is less then 5 blocks form the consensus tip then flush on every block.
                 if (distanceFromConsensusTip < 5)
                     return true;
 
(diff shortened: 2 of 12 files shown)


src/Stratis.Bitcoin/Base/ChainState.cs
```diff
@@ -14,6 +14,9 @@ public interface IChainState
         /// <summary>The highest stored block in the repository or <c>null</c> if block store feature is not enabled.</summary>
         ChainedHeader BlockStoreTip { get; set; }
 
+        /// <summary>The highest stored block in the block store cache or <c>null</c> if block store feature is not enabled.</summary>
+        ChainedHeader BlockStoreCacheTip { get; set; }
```
noescape00: I don't think it should go to chain state since we will get rid of chainstate.

Also since this tip is used only in BSBehavior which is in the same namespace we can already add this to BSQ itself instead of adding it to chainstate.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
```diff
@@ -14,6 +14,16 @@ public BlockStoreQueueFlushCondition(IChainState chainState)
         }
 
         /// <inheritdoc/>
-        public bool ShouldFlush { get { return this.chainState.IsAtBestChainTip; } }
+        public bool ShouldFlush
+        {
+            get
+            {
+                // Once store is 5 blocks form the consensus tip then flush on every block.
+                if (this.chainState.ConsensusTip.Height - this.chainState.BlockStoreTip.Height < 5)
```
noescape00: That will flush every block and disable batching. 

So let's say we've started the node and initialized ConsensusTip and BlockStoreTip at height 1000. 
Now CT advances to 1001, BS is still at 1000. 
(1001 - 1000) < 5 == true

I think it worked well like it used to be before.

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
```diff
@@ -232,8 +232,8 @@ private void AddComponentStats(StringBuilder builder)
             {
                 var chainHeadersBehavior = peer.Behavior<ConsensusManagerBehavior>();
 
-                string peerHeights = $"(r/s):{(chainHeadersBehavior.BestReceivedTip != null ? chainHeadersBehavior.BestReceivedTip.Height.ToString() : peer.PeerVersion?.StartHeight.ToString() ?? "-")}";
-                peerHeights += $"/{(chainHeadersBehavior.BestSentHeader != null ? chainHeadersBehavior.BestSentHeader.Height.ToString() : peer.PeerVersion?.StartHeight.ToString() ?? "-")}";
+                string peerHeights = $"(r/s):{(chainHeadersBehavior.BestReceivedTip != null ? chainHeadersBehavior.BestReceivedTip.Height.ToString() : peer.PeerVersion?.StartHeight + "*" ?? "-")}";
```
fassadlr: What does "*" do here?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
fassadlr: @noescape00 batching will still occur when CT is more than 5 blocks from BST, so to me this change is correct 👍

fassadlr:
@dangershony this change looks good, just please fix the 3 unit tests ;)

src/Stratis.Bitcoin/Base/ChainState.cs
noescape00: >Are we positive that chains state will go away? ConcurrentChain will go for sure yes but why chain state?

Because it's a global state that breaks the architecture.
Every component can easily change any value inside on chainState even though it's not supposed to. 

It was added to as workaround for the problem that some projects can't reference each other but still need to share a state. This problem will soon be solved. After it's solved I see no justification for chain state. 

>This was the smallest change possible in terms of number of lines to change, if I inject store I'd need to change many tests and still need to add the caching property to store.

Yes, implementing it in a way that improves the architecture instead of breaking it will take more time and lines of code, but it's worth it in a long run.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
noescape00: >Batching will basically happen through out IBD, won't it?

Not if internet is too slow or SSD is too fast so consensus can't advance faster than batch can be saved to hard drive. 
In those edge cases we will not have that race condition we rely on and it will result in having a lot of IO calls during IBD. 

>So we now have a situation where batching will happen during IBD 

Most of the times but not always. We rely on race condition for this to happen.

dangershony: Yeah I see the logic in Ivan's claim we will persist to disk in the beginning, so actually we need to also add the IBD check here, I do remember we used to check IBD for some reason it was removed, is the a problem we also check IBD?

src/Stratis.Bitcoin/Connection/ConnectionManager.cs
dangershony: Its an indicator the value is from the version and is not real.
Perhaps we just not show a number instead of the value from version payload

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
noescape00: >for some reason it was removed

Because we go out of IBD a few hundreds blocks before being synced I think

Adding IBD as a condition will help a lot, indeed. 

But still, why do we want this over saving a batch when CT is close to the best peer's tip?

dangershony: We end up with a lot more IO if not.
It was observed in stratis that CT syncs very fast (when blocks are empty) then every block that is pushed to the BSQ async loops is persisted on each dequeue of a block (because CT is now within 5 blocks from best tip) in stratis it can mean more then 2000 blocks.
If anything we should not check CT agasint best peer tip but block BSQ tip to get more correct behaviour.
See my last PR that will address the IBD issue.

noescape00: >It was observed in stratis that CT syncs very fast (when blocks are empty) then every block that is pushed to the BSQ async loops is persisted on each dequeue of a block 

I think this can be fixed just by 
```
public bool ShouldFlush { get { return this.chainState.IsAtBestChainTip && !isIBD; } }
```

dangershony: `isIBD` looks at CT time so it will always be true if `IsAtBestChainTip =true` assuming best tip advances faster then CT.

This will not fix the issue.

noescape00: >isIBD looks at CT time so it will always be true

```
if (this.chainState.ConsensusTip.Header.BlockTime.ToUnixTimeSeconds() < (this.dateTimeProvider.GetTime() - this.consensusSettings.MaxTipAge))
                return true;
```

It won't be true for old blocks, only for those that were created in the last few hours

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueue.cs
```diff
@@ -80,6 +80,9 @@ public class BlockStoreQueue : IBlockStoreQueue
 
         private readonly CancellationTokenSource cancellation;
 
+        /// <inheritdoc cref="IBlockStoreQueue.BlockStoreCacheTip"/>
```
noescape00: /// <inheritdoc />

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
```diff
@@ -7,13 +7,31 @@ namespace Stratis.Bitcoin.Features.BlockStore
...
+        {
+            get
+            {
+                // If the node is in IBD we don't flush on each block.
+                if (this.blockDownloadState.IsInitialBlockDownload())
+                    return false;
+
+                int distanceFromConsensusTip = this.chainState.ConsensusTip.Height - this.chainState.BlockStoreTip.Height;
+
+                // Once store is 5 blocks form the consensus tip then flush on every block.
```
noescape00: Once store is less than 5 blocks form the consensus tip

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreQueueFlushCondition.cs
```diff
@@ -7,13 +7,31 @@ namespace Stratis.Bitcoin.Features.BlockStore
...
+                if (distanceFromConsensusTip < 5)
```
noescape00: I'd still use `IsAtBestChainTip` over this

dangershony: But then we get the effect of flushing many blocks once tip is synced and tore is far behind.

PR #2934, opened 2018-12-10
added logs for test package method

noescape00:


dangershony:
Good but please also add logs for `PosPowBlockDefinition.TestPackage`

noescape00:
>Good but please also add logs for PosPowBlockDefinition.TestPackage

Good point. 

Btw is the logic there correct? 
I always thought that tx time can't be > block header time. But there we compare it against adjusted time + future drift.
Is that correct?

noescape00:
this is from pos coinstake rule

```
            // Check transactions.
            foreach (Transaction transaction in block.Transactions)
            {
                // Check transaction timestamp.
                if (block.Header.Time < transaction.Time)
                {
                    this.Logger.LogTrace("Block contains transaction with timestamp {0}, which is greater than block's timestamp {1}.", transaction.Time, block.Header.Time);
                    this.Logger.LogTrace("(-)[TX_TIME_MISMATCH]");
                    ConsensusErrors.BlockTimeBeforeTrx.Throw();
                }
            }
```

looks like block definition can select a tx for mining and if we create such a block it will be invalid

PR #2940, opened 2018-12-11
Fix deadlock when peer disconnects inside a setstate

dangershony:
Solving the deadlock issue using by queueing send messages inside the `OnStateChangedAsync` in to a new thread, this will avoid having the same thread deadlock on the state change lock handle.


dangershony:
Fixed by #2491

PR #2941, opened 2018-12-11
Fixing deadlock by sending messages using a queue.

dangershony:


noescape00:
Please add TODO comments to remove it in the future. 
Otherwise looks like it fixes the problem.

dangershony:
I would actually not hurry to remove this, as changing it means we need to make bigger refactor changes on the way we call behaviors (this may come) however until ten this is an ok fix, and when we refactor the behaviours we will make a lot of changes in that area a todo wont matter much.

I will how ever make change to the method name to remove the async from it

PR #2948, opened 2018-12-12
Check StakeModifierV2 property is not null on a PH

dangershony:
Low priority but this was reported as an issue on one of the tests in simulation network.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
+++ b/src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
@@ -241,6 +241,12 @@ namespace Stratis.Bitcoin.Features.Consensus.Rules.ProvenHeaderRules
 
             if (previousProvenHeader != null)
             {
+                if (previousProvenHeader.StakeModifierV2 == null)
+                {
+                    this.Logger.LogTrace("(-)[MODIF_IS_NULL]");
+                    ConsensusErrors.InvalidPreviousProvenHeaderStakeModifier.Throw();
+                }
+
                 //Stake modifier acquired from prev PH.
                 this.Logger.LogTrace("(-)[PREV_PH]");
                 return previousProvenHeader.StakeModifierV2;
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -284,12 +290,6 @@ private void CheckStakeKernelHash(PosRuleContext context, UnspentOutputs staking
 
             uint256 previousStakeModifier = this.GetPreviousStakeModifier(chainedHeader);
 
-            if (previousStakeModifier == null)
```
fassadlr: OK so this just moved to `GetPreviousStakeModifier` method?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
```diff
@@ -241,6 +241,12 @@ private uint256 GetPreviousStakeModifier(ChainedHeader chainedHeader)
 
             if (previousProvenHeader != null)
             {
+                if (previousProvenHeader.StakeModifierV2 == null)
```
MithrilMan: I'm wondering if this check, or better the if at L242 should be done after the if below `if (chainedHeader.Previous.Height == 0)` L255

MithrilMan: probably even after `if (chainedHeader.Previous.Height == this.LastCheckpointHeight)` 
basically if we have a valid StakeModifierV2 (from genesis or checkpoint) we can even live without having a PH as a previous header, thoughts?

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
MithrilMan: the null check yes, even if looking at code we may still have a null if 
`this.LastCheckpoint.StakeModifierV2` is null (I don't expect a checkpoint having a null StakeModifierV2 tho, but maybe better add a check anyway in GetPreviousStakeModifier

dangershony: naa checkpoint must have a modifier specified (well otherwise it will throw anyway)

src/Stratis.Bitcoin.Features.Consensus/Rules/ProvenHeaderRules/ProvenHeaderCoinstakeRule.cs
dangershony: Why? in the other cases stake modifier is provided by checkpoint or store

PR #2952, opened 2018-12-13
Sidechain Allowed Scripts

codingupastorm:


src/Stratis.Bitcoin.Features.SmartContracts/Rules/AllowedScriptTypeRule.cs
```diff
@@ -60,10 +62,19 @@ private void CheckOutput(TxOut output)
...
             // For cross-chain transfers
             if (PayToMultiSigTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
                 return;
 
+            // For cross-chain transfers
+            IEnumerable<Op> ops = output.ScriptPubKey.ToOps();
+            if (ops.Any() && ops.First().Code == OpcodeType.OP_RETURN)
+                return;
+
             new ConsensusError("disallowed-output-script", "Only P2PKH and smart contract scripts are allowed.").Throw();
```
rowandh: This message should probably be updated

src/Stratis.Bitcoin.Features.SmartContracts/Rules/AllowedScriptTypeRule.cs
```diff
@@ -60,10 +62,19 @@ private void CheckOutput(TxOut output)
...
+            // For cross-chain transfers
+            if (PayToScriptHashTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
+                return;
+
             // For cross-chain transfers
             if (PayToMultiSigTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
                 return;
 
+            // For cross-chain transfers
+            IEnumerable<Op> ops = output.ScriptPubKey.ToOps();
```
dangershony: I think there is a template for opreturn `TxNullDataTemplate`

src/Stratis.Bitcoin.Features.SmartContracts/Rules/AllowedScriptTypeRule.cs
codingupastorm: Good point

src/Stratis.Bitcoin.Features.SmartContracts/Rules/AllowedScriptTypeRule.cs
dangershony: Ah the problem here is that MAX_OP_RETURN_RELAY is different for Stratis so we may need to fix it a bit.

codingupastorm: Hmm, have put this in. In terms of just checking the script template type, I believe this will work.

src/Stratis.Bitcoin.Features.SmartContracts/Rules/AllowedScriptTypeRule.cs
```diff
@@ -60,11 +62,19 @@ private void CheckOutput(TxOut output)
...
+            if (PayToScriptHashTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
+                return;
+
             // For cross-chain transfers
             if (PayToMultiSigTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
                 return;
 
-            new ConsensusError("disallowed-output-script", "Only P2PKH and smart contract scripts are allowed.").Throw();
+            // For cross-chain transfers
+            if (TxNullDataTemplate.Instance.CheckScriptPubKey(output.ScriptPubKey))
```
dangershony: Is the size here not a problem?

I guess not, this are rules for the POA network right?
So effectively we allow opreturn of size of 83 bytes *similar to BTC).

PR #2966, opened 2018-12-16
[Light Wallet] Fix Breeze node and introduce pruning

fassadlr:
In this PR the node will now prune its database on startup and shut down.

An async loop will run that periodically deletes blocks from the block store. It will keep blocks up until 1000 blocks from the store tip.

On shutdown the database will be "compacted" so that all the deleted references are removed from the file and the database file re-saved.

On startup the node will check to see if its needs to be pruned (i.e. if the pruned tip is more than 1000 blocks from the store tip) if it is needed the database will be pruned and shrunk.

Some tests are still failing which I am looking at.

I have tested this extensively and the breeze wallet syncs fully.

The size of the DB after compactin is about 290kb.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -40,6 +40,12 @@ namespace Stratis.Bitcoin.Features.BlockStore
         Task DeleteAsync(HashHeightPair newTip, List<uint256> hashes);
 
         /// <summary>
+        /// Delete blocks and their transactions from disk.
+        /// </summary>
+        /// <param name="hashes">List of block hashes to be deleted.</param>
+        Task DeleteBlocksAsync(List<uint256> hashes);
+
+        /// <summary>
         /// Determine if a block already exists
         /// </summary>
         /// <param name="hash">The hash.</param>
@@ -513,6 +519,27 @@ namespace Stratis.Bitcoin.Features.BlockStore
                     List<Block> blocks = this.GetBlocksFromHashes(transaction, hashes);
                     this.OnDeleteBlocks(transaction, blocks.Where(b => b != null).ToList());
                     this.SaveTipHashAndHeight(transaction, newTip);
+                    transaction.Commit();
+                }
+            });
+
+            return task;
+        }
+
+        /// <inheritdoc />
+        public Task DeleteBlocksAsync(List<uint256> hashes)
+        {
+            Guard.NotNull(hashes, nameof(hashes));
+
+            Task task = Task.Run(() =>
+            {
+                using (DBreeze.Transactions.Transaction transaction = this.DBreeze.GetTransaction())
+                {
+                    transaction.SynchronizeTables(BlockTableName, CommonTableName, TransactionTableName);
+                    transaction.ValuesLazyLoadingIsOn = false;
+
+                    List<Block> blocks = this.GetBlocksFromHashes(transaction, hashes);
+                    this.OnDeleteBlocks(transaction, blocks.Where(b => b != null).ToList());
                     transaction.Commit();
                 }
             });
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -22,14 +22,14 @@ namespace Stratis.Bitcoin.Features.BlockStore
         DBreezeEngine DBreeze { get; }
 
         /// <summary>
-        /// Delete blocks and their transactions.
+        /// Deletes blocks and indexes for transactions that belong to deleted blocks.
         /// <para>
         /// It should be noted that this does not delete the entries from disk (only the references are removed) and
         /// as such the file size remains the same.
         /// </para>
         /// </summary>
-        /// <param name="hashes">List of block hashes to be deleted.</param>
-        /// <returns>The awaited task.</returns>
+        /// <remarks>TODO: This will need to be revisited once DBreeze has been fixed or replaced with a solution that works.</remarks>
+        /// <param name="hashes">List of block hashes to be deleted.</param>        
         Task DeleteBlocksAsync(List<uint256> hashes);
 
         /// <summary>
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -91,7 +91,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
         private readonly ILogger logger;
 
-        protected readonly Network network;
+        private readonly Network network;
 
         private static readonly byte[] RepositoryTipKey = new byte[0];
 
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -18,6 +18,20 @@ namespace Stratis.Bitcoin.Features.BlockStore
     /// </summary>
     public interface IBlockRepository : IBlockStore
     {
+        /// <summary> The dbreeze database engine.</summary>
+        DBreezeEngine DBreeze { get; }
+
+        /// <summary>
+        /// Delete blocks and their transactions.
+        /// <para>
+        /// It should be noted that this does not delete the entries from disk (only the references are removed) and
+        /// as such the file size remains the same.
+        /// </para>
+        /// </summary>
+        /// <param name="hashes">List of block hashes to be deleted.</param>
+        /// <returns>The awaited task.</returns>
+        Task DeleteBlocksAsync(List<uint256> hashes);
+
         /// <summary>
         /// Persist the next block hash and insert new blocks into the database.
         /// </summary>
@@ -40,16 +54,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
         Task DeleteAsync(HashHeightPair newTip, List<uint256> hashes);
 
         /// <summary>
-        /// Delete blocks and their transactions.
-        /// <para>
-        /// It should be noted that this does not delete the entries from disk (only the references are removed) and
-        /// as such the file size remains the same.
-        /// </para>
-        /// </summary>
-        /// <param name="hashes">List of block hashes to be deleted.</param>
-        Task DeleteBlocksAsync(List<uint256> hashes);
-
-        /// <summary>
         /// Determine if a block already exists
         /// </summary>
         /// <param name="hash">The hash.</param>
@@ -68,62 +72,30 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <param name="txIndex">Whether to index transactions.</param>
         Task SetTxIndexAsync(bool txIndex);
 
-        /// <summary> 
-        /// The lowest block height that the repository has.
-        /// <para>
-        /// This also indicated where the node has been pruned up to.
-        /// </para>
-        /// </summary>
-        HashHeightPair PrunedTip { get; }
-
-        /// <summary>
-        /// Sets the pruned tip.
-        /// <para> 
-        /// It will be saved once the block database has been compacted on node initialization or shutdown.
-        /// </para>
-        /// </summary>
-        /// <param name="tip">The tip to set.</param>
-        void UpdatePrunedTip(ChainedHeader tip);
-
         /// <summary>Hash and height of the repository's tip.</summary>
         HashHeightPair TipHashAndHeight { get; }
 
         /// <summary> Indicates that the node should store all transaction data in the database.</summary>
         bool TxIndex { get; }
-
-        /// <summary>
-        /// Compacts the block and transaction database by resaving the database file without
-        /// all the deleted references.
-        /// </summary>
-        /// <param name="consensusTip">The current tip of consensus.</param>
-        /// <param name="nodeInitializing">Indicates whether or not this method is called from node startup or not.</param>
-        Task PruneDatabase(ChainedHeader consensusTip, bool nodeInitializing);
     }
 
     public class BlockRepository : IBlockRepository
     {
-        private const string BlockTableName = "Block";
+        internal const string BlockTableName = "Block";
 
-        private const string CommonTableName = "Common";
+        internal const string CommonTableName = "Common";
 
-        private const string TransactionTableName = "Transaction";
+        internal const string TransactionTableName = "Transaction";
 
-        private readonly DBreezeEngine DBreeze;
+        public DBreezeEngine DBreeze { get; }
 
         private readonly ILogger logger;
 
-        private readonly Network network;
-
-        private readonly StoreSettings storeSettings;
-
-        private static readonly byte[] PrunedTipKey = new byte[2];
+        protected readonly Network network;
 
         private static readonly byte[] RepositoryTipKey = new byte[0];
 
         private static readonly byte[] TxIndexKey = new byte[1];
-
-        /// <inheritdoc />
-        public HashHeightPair PrunedTip { get; private set; }
 
         /// <inheritdoc />
         public HashHeightPair TipHashAndHeight { get; private set; }
@@ -131,23 +103,21 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <inheritdoc />
         public bool TxIndex { get; private set; }
 
-        public BlockRepository(Network network, DataFolder dataFolder, ILoggerFactory loggerFactory, StoreSettings storeSettings)
-            : this(network, dataFolder.BlockPath, loggerFactory, storeSettings)
+        public BlockRepository(Network network, DataFolder dataFolder, ILoggerFactory loggerFactory)
+            : this(network, dataFolder.BlockPath, loggerFactory)
         {
         }
 
-        public BlockRepository(Network network, string folder, ILoggerFactory loggerFactory, StoreSettings storeSettings)
+        public BlockRepository(Network network, string folder, ILoggerFactory loggerFactory)
         {
             Guard.NotNull(network, nameof(network));
             Guard.NotEmpty(folder, nameof(folder));
-            Guard.NotNull(storeSettings, nameof(storeSettings));
 
             Directory.CreateDirectory(folder);
             this.DBreeze = new DBreezeEngine(folder);
 
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.network = network;
-            this.storeSettings = storeSettings;
         }
 
         /// <inheritdoc />
@@ -160,16 +130,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 using (DBreeze.Transactions.Transaction transaction = this.DBreeze.GetTransaction())
                 {
                     bool doCommit = false;
-
-                    if (this.storeSettings.Prune)
-                    {
-                        if (this.LoadPrunedTip(transaction) == null)
-                        {
-                            this.PrunedTip = new HashHeightPair(genesis.GetHash(), 0);
-                            transaction.Insert(CommonTableName, PrunedTipKey, this.PrunedTip);
-                            doCommit = true;
-                        }
-                    }
 
                     if (this.LoadTipHashAndHeight(transaction) == null)
                     {
@@ -188,127 +148,6 @@ namespace Stratis.Bitcoin.Features.BlockStore
             });
 
             return task;
-        }
-
-        /// <inheritdoc />
-        public async Task PruneDatabase(ChainedHeader consensusTip, bool nodeInitializing)
-        {
-            this.logger.LogInformation($"Pruning started.");
-
-            if (nodeInitializing)
-            {
-                if (IsDatabasePruned())
-                    return;
-
-                await this.PrepareDatabaseForPruningAsync(consensusTip);
-            }
-
-            this.CompactDataBase();
-
-            this.logger.LogInformation($"Pruning complete.");
-
-            return;
-        }
-
-        private bool IsDatabasePruned()
-        {
-            if (this.TipHashAndHeight.Height <= this.PrunedTip.Height + this.storeSettings.PruneBlockMargin)
-            {
-                this.logger.LogDebug("(-):true");
-                return true;
-            }
-            else
-            {
-                this.logger.LogDebug("(-):false");
-                return false;
-            }
-        }
-
-        /// <summary>
-        /// Compacts the block and transaction database by recreating the tables without the deleted references.
-        /// </summary>
-        /// <param name="consensusTip">The last fully validated block of the node.</param>
-        private async Task PrepareDatabaseForPruningAsync(ChainedHeader consensusTip)
-        {
-            int upperHeight = this.TipHashAndHeight.Height - this.storeSettings.PruneBlockMargin;
-
-            var toDelete = new List<ChainedHeader>();
-
-            ChainedHeader startFromHeader = consensusTip.GetAncestor(upperHeight);
-            ChainedHeader endAtHeader = consensusTip.GetAncestor(this.PrunedTip.Height);
-
-            this.logger.LogInformation($"Pruning blocks from height {upperHeight} to {endAtHeader.Height}.");
-
-            while (startFromHeader.Previous != null && startFromHeader != endAtHeader)
-            {
-                toDelete.Add(startFromHeader);
-                startFromHeader = startFromHeader.Previous;
-            }
-
-            await this.DeleteBlocksAsync(toDelete.Select(cb => cb.HashBlock).ToList()).ConfigureAwait(false);
-
-            this.UpdatePrunedTip(consensusTip.GetAncestor(upperHeight));
-        }
-
-        private HashHeightPair LoadPrunedTip(DBreeze.Transactions.Transaction dbreezeTransaction)
-        {
-            if (this.PrunedTip == null)
-            {
-                dbreezeTransaction.ValuesLazyLoadingIsOn = false;
-
-                Row<byte[], HashHeightPair> row = dbreezeTransaction.Select<byte[], HashHeightPair>(CommonTableName, PrunedTipKey);
-                if (row.Exists)
-                    this.PrunedTip = row.Value;
-
-                dbreezeTransaction.ValuesLazyLoadingIsOn = true;
-            }
-
-            return this.PrunedTip;
-        }
-
-        /// <summary>
-        /// Compacts the block and transaction database by recreating the tables without the deleted references.
-        /// </summary>
-        private void CompactDataBase()
-        {
-            Task task = Task.Run(() =>
-            {
-                using (DBreeze.Transactions.Transaction dbreezeTransaction = this.DBreeze.GetTransaction())
-                {
-                    dbreezeTransaction.SynchronizeTables(BlockTableName, TransactionTableName);
-
-                    var tempBlocks = dbreezeTransaction.SelectDictionary<byte[], Block>(BlockTableName);
-
-                    if (tempBlocks.Count != 0)
-                    {
-                        this.logger.LogInformation($"{tempBlocks.Count} blocks will be copied to the pruned table.");
-
-                        dbreezeTransaction.RemoveAllKeys(BlockTableName, true);
-                        dbreezeTransaction.InsertDictionary(BlockTableName, tempBlocks, false);
-
-                        var tempTransactions = dbreezeTransaction.SelectDictionary<byte[], Block>(TransactionTableName);
-                        if (tempTransactions.Count != 0)
-                        {
-                            this.logger.LogInformation($"{tempTransactions.Count} transactions will be copied to the pruned table.");
-                            dbreezeTransaction.RemoveAllKeys(TransactionTableName, true);
-                            dbreezeTransaction.InsertDictionary(TransactionTableName, tempTransactions, false);
-                        }
-
-                        // Save the hash and height of where the node was pruned up to.
-                        dbreezeTransaction.Insert(CommonTableName, PrunedTipKey, this.PrunedTip);
-                    }
-
-                    dbreezeTransaction.Commit();
-                }
-
-                return Task.CompletedTask;
-            });
-        }
-
-        /// <inheritdoc />
-        public void UpdatePrunedTip(ChainedHeader tip)
-        {
-            this.PrunedTip = new HashHeightPair(tip);
         }
 
         /// <inheritdoc />
@@ -638,7 +477,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
                 dbreezeTransaction.RemoveKey<byte[]>(BlockTableName, block.GetHash().ToBytes());
         }
 
-        private List<Block> GetBlocksFromHashes(DBreeze.Transactions.Transaction dbreezeTransaction, List<uint256> hashes)
+        public List<Block> GetBlocksFromHashes(DBreeze.Transactions.Transaction dbreezeTransaction, List<uint256> hashes)
         {
             var results = new Dictionary<uint256, Block>();
 
@@ -701,10 +540,11 @@ namespace Stratis.Bitcoin.Features.BlockStore
             {
                 using (DBreeze.Transactions.Transaction transaction = this.DBreeze.GetTransaction())
                 {
-                    transaction.SynchronizeTables(BlockTableName, CommonTableName, TransactionTableName);
+                    transaction.SynchronizeTables(BlockRepository.BlockTableName, BlockRepository.CommonTableName, BlockRepository.TransactionTableName);
                     transaction.ValuesLazyLoadingIsOn = false;
 
                     List<Block> blocks = this.GetBlocksFromHashes(transaction, hashes);
+
                     this.OnDeleteBlocks(transaction, blocks.Where(b => b != null).ToList());
 
                     transaction.Commit();
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -40,7 +40,11 @@ namespace Stratis.Bitcoin.Features.BlockStore
         Task DeleteAsync(HashHeightPair newTip, List<uint256> hashes);
 
         /// <summary>
-        /// Delete blocks and their transactions from disk.
+        /// Delete blocks and their transactions.
+        /// <para>
+        /// It should be noted that this does not delete the entries from disk (only the references are removed) and
+        /// as such the file size remains the same.
+        /// </para>
         /// </summary>
         /// <param name="hashes">List of block hashes to be deleted.</param>
         Task DeleteBlocksAsync(List<uint256> hashes);
@@ -67,6 +71,15 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <summary>Hash and height indicating where the node's block store has been pruned up to.</summary>
         HashHeightPair PrunedTip { get; }
 
+        /// <summary>
+        /// Sets the pruned tip.
+        /// <para> 
+        /// It will be saved once the block database has been compacted on node initialization or shutdown.
+        /// </para>
+        /// </summary>
+        /// <param name="tip">The tip to set.</param>
+        void UpdatePrunedTip(ChainedHeader tip);
+
         /// <summary>Hash and height of the repository's tip.</summary>
         HashHeightPair TipHashAndHeight { get; }
 
@@ -77,7 +90,9 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// Compacts the block and transaction database by resaving the database file without
         /// all the deleted references.
         /// </summary>
-        Task PruneDatabase(ChainedHeader consensusTip);
+        /// <param name="consensusTip">The current tip of consensus.</param>
+        /// <param name="nodeInitializing">Indicates whether or not this method is called from node startup or not.</param>
+        Task PruneDatabase(ChainedHeader consensusTip, bool nodeInitializing);
     }
 
     public class BlockRepository : IBlockRepository
@@ -167,16 +182,72 @@ namespace Stratis.Bitcoin.Features.BlockStore
         }
 
         /// <inheritdoc />
-        public async Task PruneDatabase(ChainedHeader consensusTip)
+        public async Task PruneDatabase(ChainedHeader consensusTip, bool nodeInitializing)
         {
-            if (!this.storeSettings.Prune)
-                return;
+            if (nodeInitializing)
+            {
+                if (IsDatabasePruned())
+                    return;
 
-            if (IsDatabasePruned())
-                return;
+                await this.PrepareDatabaseForPruningAsync(consensusTip);
+            }
 
-            await PrepareDatabaseForPruningAsync(consensusTip);
+            this.CompactDataBase();
 
+            return;
+        }
+
+        private bool IsDatabasePruned()
+        {
+            return this.TipHashAndHeight.Height <= this.PrunedTip.Height + MaxBlocksToKeep;
+        }
+
+        /// <summary>
+        /// Compacts the block and transaction database by recreating the tables without the deleted references.
+        /// </summary>
+        private async Task PrepareDatabaseForPruningAsync(ChainedHeader consensusTip)
+        {
+            var upperHeight = this.TipHashAndHeight.Height - MaxBlocksToKeep;
+
+            var toDelete = new List<ChainedHeader>();
+
+            var startFromHeader = consensusTip.GetAncestor(upperHeight);
+            var endAtHeader = consensusTip.GetAncestor(this.PrunedTip.Height);
+
+            this.logger.LogInformation($"Pruning triggered, delete from {upperHeight} to {endAtHeader.Height}.");
+
+            while (startFromHeader.Previous != null && startFromHeader != endAtHeader)
+            {
+                toDelete.Add(startFromHeader);
+                startFromHeader = startFromHeader.Previous;
+            }
+
+            await this.DeleteBlocksAsync(toDelete.Select(cb => cb.HashBlock).ToList()).ConfigureAwait(false);
+
+            this.UpdatePrunedTip(consensusTip.GetAncestor(upperHeight));
+        }
+
+        private HashHeightPair LoadPrunedTip(DBreeze.Transactions.Transaction dbreezeTransaction)
+        {
+            if (this.PrunedTip == null)
+            {
+                dbreezeTransaction.ValuesLazyLoadingIsOn = false;
+
+                Row<byte[], HashHeightPair> row = dbreezeTransaction.Select<byte[], HashHeightPair>(CommonTableName, PrunedTipKey);
+                if (row.Exists)
+                    this.PrunedTip = row.Value;
+
+                dbreezeTransaction.ValuesLazyLoadingIsOn = true;
+            }
+
+            return this.PrunedTip;
+        }
+
+        /// <summary>
+        /// Compacts the block and transaction database by recreating the tables without the deleted references.
+        /// </summary>
+        private void CompactDataBase()
+        {
             Task task = Task.Run(() =>
             {
                 using (DBreeze.Transactions.Transaction dbreezeTransaction = this.DBreeze.GetTransaction())
@@ -203,49 +274,12 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
                 return Task.CompletedTask;
             });
-
-            return;
         }
 
-        private bool IsDatabasePruned()
+        /// <inheritdoc />
+        public void UpdatePrunedTip(ChainedHeader tip)
         {
-            return this.TipHashAndHeight.Height <= this.PrunedTip.Height + MaxBlocksToKeep;
-        }
-
-        private async Task PrepareDatabaseForPruningAsync(ChainedHeader consensusTip)
-        {
-            var upperHeight = this.TipHashAndHeight.Height - MaxBlocksToKeep;
-
-            var chainedHeadersToDelete = new List<ChainedHeader>();
-
-            var startFrom = consensusTip.GetAncestor(upperHeight);
-            var endAt = consensusTip.GetAncestor(this.PrunedTip.Height);
-
-            while (startFrom.Previous != null && startFrom != endAt)
-            {
-                chainedHeadersToDelete.Add(startFrom);
-                startFrom = startFrom.Previous;
-            }
-
-            await this.DeleteBlocksAsync(chainedHeadersToDelete.Select(c => c.HashBlock).ToList()).ConfigureAwait(false);
-
-            this.PrunedTip = new HashHeightPair(consensusTip.GetAncestor(upperHeight));
-        }
-
-        private HashHeightPair LoadPrunedTip(DBreeze.Transactions.Transaction dbreezeTransaction)
-        {
-            if (this.PrunedTip == null)
-            {
-                dbreezeTransaction.ValuesLazyLoadingIsOn = false;
-
-                Row<byte[], HashHeightPair> row = dbreezeTransaction.Select<byte[], HashHeightPair>(CommonTableName, PrunedTipKey);
-                if (row.Exists)
-                    this.PrunedTip = row.Value;
-
-                dbreezeTransaction.ValuesLazyLoadingIsOn = true;
-            }
-
-            return this.PrunedTip;
+            this.PrunedTip = new HashHeightPair(tip);
         }
 
         /// <inheritdoc />
@@ -643,6 +677,7 @@ namespace Stratis.Bitcoin.Features.BlockStore
 
                     List<Block> blocks = this.GetBlocksFromHashes(transaction, hashes);
                     this.OnDeleteBlocks(transaction, blocks.Where(b => b != null).ToList());
+
                     transaction.Commit();
                 }
             });
--- a/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
@@ -68,7 +68,12 @@ namespace Stratis.Bitcoin.Features.BlockStore
         /// <param name="txIndex">Whether to index transactions.</param>
         Task SetTxIndexAsync(bool txIndex);
 
-        /// <summary>Hash and height indicating where the node's block store has been pruned up to.</summary>
+        /// <summary> 
+        /// The lowest block height that the repository has.
+        /// <para>
+        /// This also indicated where the node has been pruned up to.
+        /// </para>
+        /// </summary>
         HashHeightPair PrunedTip { get; }
 
         /// <summary>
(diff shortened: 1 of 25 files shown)


zeptin:
So I've only taken a very brief look at this, but note that a pruned node should have amended service bits to reflect the reduced capabilities, I believe it's `NODE_NETWORK_LIMITED` instead of `NODE_NETWORK` if memory serves.

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletSyncManager.cs
```diff
@@ -22,27 +22,23 @@ public interface IWalletSyncManager
         /// <summary>
         /// Processes a new block.
         /// </summary>
-        /// <param name="block"></param>
         void ProcessBlock(Block block);
 
         /// <summary>
         /// Processes a new transaction which is in a pending state (not included in a block).
         /// </summary>
-        /// <param name="transaction"></param>
```
bokobza: Why did you remove those?

src/Stratis.BreezeD/Program.cs
```diff
@@ -21,42 +21,38 @@ public static async Task Main(string[] args)
...
 
-                // Start Full Node - this will also start the API.
-                if (node != null)
-                    await node.RunAsync();
+                await node.RunAsync();
             }
             catch (Exception ex)
             {
-                Console.WriteLine("There was a problem initializing the node. Details: '{0}'", ex.ToString());
+                Console.WriteLine($"There was a problem initializing the node: '{ex.ToString()}'");
```
bokobza: Why this change? It's exactly the same as StratisD.

src/Stratis.BreezeD/Program.cs
```diff
@@ -21,42 +21,38 @@ public static async Task Main(string[] args)
...
                     .UseBlockNotification()
                     .UseTransactionNotification()
                     .UseApi();
 
                 IFullNode node = fullNodeBuilder.Build();
 
-                // Start Full Node - this will also start the API.
-                if (node != null)
-                    await node.RunAsync();
+                await node.RunAsync();
```
bokobza: Why this change? It's exactly the same as StratisD.

src/Stratis.BreezeD/Program.cs
```diff
@@ -21,42 +21,38 @@ public static async Task Main(string[] args)
...
                 else
                 {
-                    nodeSettings = new NodeSettings(networksSelector:Networks.Bitcoin, agent: agent, args: args);
+                    nodeSettings = new NodeSettings(networksSelector: Networks.Bitcoin, agent: "Breeze", args: args);
                 }
 
                 IFullNodeBuilder fullNodeBuilder = new FullNodeBuilder()
                     .UseNodeSettings(nodeSettings)
+                    .UseBlockStore()
+                    .UsePosConsensus()
```
bokobza: not necessarily PosConsensus, as this can also be used for bitcoin.
Maybe move the node construction inside the `if(stratis)` and `else`.

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -210,7 +217,57 @@ public void ProcessBlock(Block block)
 
                     // The wallet is falling behind we need to catch up.
                     this.logger.LogWarning("New tip '{0}' is too far in advance, put the puller back.", newTip);
-                    this.blockNotification.SyncFrom(this.walletTip.HashBlock);
+
+                    CancellationToken token = this.nodeLifetime.ApplicationStopping;
+
+                    ChainedHeader next = this.walletTip;
+                    while (next != newTip)
```
bokobza: Is this taken from the WalletSyncManager?

src/Stratis.Bitcoin.Features.LightWallet/Blocks/ILightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using NBitcoin;
+
+namespace Stratis.Bitcoin.Features.LightWallet.Blocks
+{
+    public interface ILightWalletBlockStoreService : IDisposable
+    {
+        ChainedHeader PrunedUpToHeaderTip { get; }
```
bokobza: Please add some details about what these do.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -58,55 +68,81 @@ public interface IBlockRepository : IBlockStore
...
         private const string BlockTableName = "Block";
 
-        private const string TransactionTableName = "Transaction";
-
         private const string CommonTableName = "Common";
 
-        /// <summary>Instance logger.</summary>
-        private readonly ILogger logger;
+        /// <summary> The amount of blocks to keep from the block store's tip.</summary>
+        private const int PruneBlockMargin = 1000;
```
noescape00: Probably should be configurable

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -139,6 +185,127 @@ public virtual Task InitializeAsync()
             return task;
         }
 
+        /// <inheritdoc />
+        public async Task PruneDatabase(ChainedHeader consensusTip, bool nodeInitializing)
```
noescape00: Instead of calling this on shutdown and initialization I'd also call it from a background thread that would cleanup old blocks every N seconds. Otherwise for pruning to kick in you need to restart the node to reduce how much disk space you are consuming

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -139,6 +185,127 @@ public virtual Task InitializeAsync()
...
+            }
+        }
+
+        /// <summary>
+        /// Compacts the block and transaction database by recreating the tables without the deleted references.
+        /// </summary>
+        /// <param name="consensusTip">The last fully validated block of the node.</param>
+        private async Task PrepareDatabaseForPruningAsync(ChainedHeader consensusTip)
+        {
+            var upperHeight = this.TipHashAndHeight.Height - PruneBlockMargin;
```
noescape00: Unvar here and everywhere where var policy is violated

bokobza:
> So I've only taken a very brief look at this, but note that a pruned node should have amended service bits to reflect the reduced capabilities, I believe it's `NODE_NETWORK_LIMITED` instead of `NODE_NETWORK` if memory serves.

Indeed, for Bitcoin, and we don't have this in StratisX.

noescape00:
Just to mention: pruning has very little sense before we fix problem with DBrezze - it doesn't delete actual data, just removes keys but the data is still there and is consuming disk space.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
bokobza: on bitcoin, it's 288 blocks.

noescape00: By default, yes. And this value came from amount of time to produce 288 blocks. For stratis it takes 10 times less time to produce 288 blocks so it will make sense to have value of 2880. 
Also we need to remember that we can't prune more than MaxReorg blocks.

src/Stratis.BreezeD/Program.cs
fassadlr: OK yes I have fixed this 👍

src/Stratis.BreezeD/Program.cs
fassadlr: Ah I just thought its an unnecessary check as if the `node` was null here we actually might want to let startup break so that we are aware of it?

src/Stratis.BreezeD/Program.cs
fassadlr: Just a bit of cleanup ;)

src/Stratis.Bitcoin.Features.Wallet/Interfaces/IWalletSyncManager.cs
fassadlr: They were not completed correctly but I will put them back and complete the `param`.

src/Stratis.Bitcoin.Features.LightWallet/Blocks/ILightWalletBlockStoreService.cs
fassadlr: OK I've added lots of comments on the interface etc. 👍

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
fassadlr: Correct :)

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: @noescape00 yes @dangershony and I discussed this but wouldn't this mean that we need to introduce a global lock? 

We are worried to make this PR a big one so said that it might be OK to let the node do the shrinking on startup and shutdown only, otherwise we need to makes do the node which might have a muhc deeper impact?

noescape00: I don't think global lock will be needed. Why? 

>We are worried to make this PR a big one so said that it might be OK to let the node do the shrinking on startup and shutdown only, otherwise we need to makes do the node which might have a much deeper impact?

If we do it like that there is little sense to have pruning. I think we need to make pruning work similar to what BTC does.

fassadlr:
> Just to mention: pruning has very little sense before we fix problem with DBrezze - it doesn't delete actual data, just removes keys but the data is still there and is consuming disk space.

@noescape00 This has been "fixed" with a workaround, please see the `CompactDatabase` method in block repository ;)

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -58,55 +68,81 @@ public interface IBlockRepository : IBlockStore
         /// <param name="txIndex">Whether to index transactions.</param>
         Task SetTxIndexAsync(bool txIndex);
 
+        /// <summary>Hash and height indicating where the node's block store has been pruned up to.</summary>
+        HashHeightPair PrunedTip { get; }
```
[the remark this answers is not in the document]
fassadlr: @noescape00 are you suggesting that I move ALL of the  prune related properties and methods to `PrunedBlockRepository`? In other words create a new class that inherits from `BlockRepository`?

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -139,6 +185,127 @@ public virtual Task InitializeAsync()
...
+
+            while (startFromHeader.Previous != null && startFromHeader != endAtHeader)
+            {
+                toDelete.Add(startFromHeader);
+                startFromHeader = startFromHeader.Previous;
+            }
+
+            await this.DeleteBlocksAsync(toDelete.Select(cb => cb.HashBlock).ToList()).ConfigureAwait(false);
+
+            this.UpdatePrunedTip(consensusTip.GetAncestor(upperHeight));
```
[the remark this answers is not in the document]
fassadlr: This is not `consensusTip` but the `ChainedHeader` that equates to the upper height of where the node were pruned up to.

Or the lowest highest of what the node has after pruning.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: OK I will change to 2880 👍

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: OK so we have to remember that there are 2 concepts here:

**Pruning:**

This deletes the blocks from the store but the file size on disk remains the same.

**Compacting:**

This will recreate the database file on disk without the deleted reference, thus decreasing the file size.

We can't compact the block database during the node's life time as we physically deleting the file on disk and recreating it.. UNLESS we take a lock and wait for it to finish.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
noescape00: Also please make it configurable

fassadlr: Put it in StoreSettings?

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
noescape00: Yeah. I'd vote against compacting since it's much more time consuming and increases startup or shutdown greatly. 

It's easier for us to change db that actually deletes data then go with smth other than proper pruning.

fassadlr: @noescape00 I would it if we can do this, cause then we don't have to worry about startup and shutdown. But a change like this might mean retesting with the testers, which is time we don't have? What do you think.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
noescape00: Yeah, sounds good

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: > Yeah. I'd vote against compacting since it's much more time consuming and increases startup or shutdown greatly.

Just FYI, if we keep on pruning during the node's lifetime and compact during startup and shutdown it happens almost instantly.

src/Stratis.BreezeD/Program.cs
dangershony: FYI you dont need the `ToString()`

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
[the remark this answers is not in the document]
dangershony: >I think it will be better to inherit from BlockRepository and do IPrunedBlockRepository

I am not sure I agree, because pruning is a property also of the node itself (not just dbreeze)
If we use inheritance it means in addition to stratisd will have prunedStratisD (or we will have to check the pruned flag and replace the implementation on startup

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
dangershony: Some info form the release notes (might be out of date though)
https://github.com/bitcoin/bitcoin/blob/v0.11.0rc1/doc/release-notes.md#block-file-pruning

The pruned value was always planned to go in the settings, but we need to set a minimum of pruned blocks, it just seemed to me that max reorg is too much of a close call, but besides that a pruned node will not signal Network Services so will never be asked to serve blocks, bitcoin by minimum prunes anything after 2 days of blocks, and the value it seems is based on mb size not.

So either we add a minimum to the network class or we use a multiple of max reorg (which actually might be disabled)

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
dangershony: unfortunately compacting is a limitation of debreeze, if we had to do it in a background task it indeed means a global lock (because this will actually copy table around) so an ok solution is shrink onstartup (and fassa also wanted to do it on shutdown)

src/Stratis.Bitcoin.Features.BlockStore/StoreSettings.cs
```diff
@@ -56,10 +60,11 @@ public static void PrintHelp(Network network)
...
             var builder = new StringBuilder();
 
-            builder.AppendLine($"-txindex=<0 or 1>         Enable to maintain a full transaction index.");
-            builder.AppendLine($"-reindex=<0 or 1>         Rebuild chain state and block index from block data files on disk.");
-            builder.AppendLine($"-prune=<0 or 1>           Enable pruning to reduce storage requirements by enabling deleting of old blocks.");
-            builder.AppendLine($"-maxCacheBlocksCount=<number> The maximum amount of blocks the cache can contain. Default is {DefaultMaxCacheBlocksCount}.");
+            builder.AppendLine($"-txindex=<0 or 1>              Enable to maintain a full transaction index.");
+            builder.AppendLine($"-reindex=<0 or 1>              Rebuild chain state and block index from block data files on disk.");
+            builder.AppendLine($"-prune=<0 or 1>                Enable pruning to reduce storage requirements by enabling deleting of old blocks.");
+            builder.AppendLine($"-pruneblockmargin=<2880 +>     The amount of blocks to keep on disk.");
```
dangershony: I suggest instead of pruneblockmargin  we just use the prune value and instead of bool it will be the number of blocks to prune or 0/-1 disabled

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
```diff
@@ -0,0 +1,179 @@
...
+                        var tempTransactions = dbreezeTransaction.SelectDictionary<byte[], Block>(BlockRepository.TransactionTableName);
+                        if (tempTransactions.Count != 0)
+                        {
+                            this.logger.LogInformation($"{tempTransactions.Count} transactions will be copied to the pruned table.");
+                            dbreezeTransaction.RemoveAllKeys(BlockRepository.TransactionTableName, true);
+                            dbreezeTransaction.InsertDictionary(BlockRepository.TransactionTableName, tempTransactions, false);
+                        }
+
+                        // Save the hash and height of where the node was pruned up to.
+                        dbreezeTransaction.Insert(BlockRepository.CommonTableName, prunedTipKey, this.PrunedTip);
```
dangershony: This is compacting why do we need to save the pruned tip?

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
```diff
@@ -0,0 +1,179 @@
...
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Compacts the block and transaction database by recreating the tables without the deleted references.
+        /// </summary>
+        /// <param name="consensusTip">The last fully validated block of the node.</param>
+        /// <returns>The awaited task.</returns>
+        private async Task PrepareDatabaseForPruningAsync(ChainedHeader consensusTip)
```
dangershony: The name consensusTip is not correct here right? it should be storeTip

dangershony: Why the name `PrepareDatabaseForPruningAsync` this is actually the pruning code no?

noescape00: it should be repo tip. 
in case store is > prune amount ahead of the the repo (which is almost always the case in IBD) and then we kill the node, start it and have a reorg- then the node is dead and can't recover.

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,105 @@
...
+                return;
+
+            if (this.blockRepository.TipHashAndHeight.Height == (this.PrunedUpToHeaderTip?.Height ?? 0))
+                return;
+
+            if (this.blockRepository.TipHashAndHeight.Height < (this.PrunedUpToHeaderTip?.Height ?? 0 + this.storeSettings.PruneBlockMargin))
+                return;
+
+            var heightToPruneFrom = this.blockRepository.TipHashAndHeight.Height - this.storeSettings.PruneBlockMargin;
+            ChainedHeader startFrom = this.consensusManager.Tip.GetAncestor(heightToPruneFrom);
```
dangershony: this should be store tip not  consensusManager.Tip

src/Stratis.BreezeD/Program.cs
fassadlr: 👍

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (this.PrunedUpToHeaderTip != null && startFrom == this.PrunedUpToHeaderTip)
+                return;
+
+            this.logger.LogInformation($"Pruning triggered, delete from {heightToPruneFrom} to {this.PrunedUpToHeaderTip?.Height ?? 0}.");
```
dangershony: Reminder In trace we normally dont use string interpolation

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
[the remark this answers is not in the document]
fassadlr: Yeah IMO its better to have this logic separate from the BlockRepository so that we can test and configure the node easier.

src/Stratis.Bitcoin.Features.BlockStore/StoreSettings.cs
fassadlr: OK yeah let me fix this 👍

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -210,7 +217,57 @@ public void ProcessBlock(Block block)
...
+                    {
+                        // While the wallet is catching up the entire node will wait.
+                        // If a wallet is recovered to a date in the past. Consensus will stop until the wallet is up to date.
+
+                        // TODO: This code should be replaced with a different approach
+                        // Similar to BlockStore the wallet should be standalone and not depend on consensus.
+                        // The block should be put in a queue and pushed to the wallet in an async way.
+                        // If the wallet is behind it will just read blocks from store (or download in case of a pruned node).
+
+                        token.ThrowIfCancellationRequested();
```
dangershony: There is a pending PR that changes this maybe use that code?

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -210,7 +217,57 @@ public void ProcessBlock(Block block)
...
+
+                        next = newTip.GetAncestor(next.Height + 1);
+                        Block nextblock = null;
+                        int index = 0;
+                        while (true)
+                        {
+                            token.ThrowIfCancellationRequested();
+
+                            nextblock = this.blockStore.GetBlockAsync(next.HashBlock).GetAwaiter().GetResult();
```
dangershony: Here you could actually just use ConsensusManager.GetBlockData instead, this is the intended way to get blocks

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
fassadlr: The Pruning Service uses this to determine if the node has been pruned or not and from what height its need to prune.

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
fassadlr: > 
> 
> The name consensusTip is not correct here right? it should be storeTip

Yes it should be the repository's tip.... Fixed!

fassadlr: > 
> 
> Why the name `PrepareDatabaseForPruningAsync` this is actually the pruning code no?

Yes this should actually be called `PrepareDatabaseForCompactingAsync`... Fixed!

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed 👍

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
fassadlr: Is that code tested? I dont wanna put code in this PR that hasnt been merged and tested?

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
fassadlr: OK Fixed :)

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
dangershony: >it should be repo tip.
`StoreTip` and `RepoTip` are the same thing I believe, the cache has its own tip `StoreCacheTip`

src/Stratis.Bitcoin.Features.BlockStore/StoreSettings.cs
dangershony: Bitcoin does it that way, just for consistency check how do they disable pruning

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
dangershony: Or just ignore this comment for now

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -82,15 +86,25 @@ private void AddInlineStats(StringBuilder log)
...
             }
         }
 
         public override Task InitializeAsync()
         {
+            if (this.storeSettings.Prune)
+            {
+                this.logger.LogInformation("Pruning BlockStore...");
+                this.prunedBlockRepository.InitializeAsync(this.network).GetAwaiter().GetResult();
+                this.prunedBlockRepository.PruneDatabase(this.consensusManager.Tip, true).GetAwaiter().GetResult();
```
dangershony: why not put PruneDatabase in Initialize

dangershony:
Note also we may need to consider changing this 
` this.connectionManager.Parameters.Services = (this.storeSettings.Prune ? NetworkPeerServices.Nothing : NetworkPeerServices.Network);`

As suggested above

src/Stratis.Bitcoin.Features.BlockStore/StoreSettings.cs
fassadlr: Bitcoin regards -prune=0 as disabled. I've made the changes.

fassadlr:
> 
> 
> Note also we may need to consider changing this
> ` this.connectionManager.Parameters.Services = (this.storeSettings.Prune ? NetworkPeerServices.Nothing : NetworkPeerServices.Network);`
> 
> As suggested above

Should I change to NODE_NETWORK_LIMITED ?

dangershony:
One more comment, we need to protect the fullnode against pruning, in case a user synced fullnode then moved to pruning, as you suggested the last pruned block.
This means we need to be able to check PrunedTip form the block store of the feature 
If pruning is disabled but PrunedTip  is set the node cannot startup

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
dangershony: Its been tested on sidechains

src/Stratis.Bitcoin.Features.LightWallet/LightWalletSyncManager.cs
```diff
@@ -234,13 +235,13 @@ public void ProcessBlock(Block block)
...
                         next = newTip.GetAncestor(next.Height + 1);
-                        Block nextblock = null;
+                        ChainedHeaderBlock nextblock = null;
                         int index = 0;
                         while (true)
                         {
                             token.ThrowIfCancellationRequested();
 
-                            nextblock = this.blockStore.GetBlockAsync(next.HashBlock).GetAwaiter().GetResult();
+                            nextblock = this.consensusManager.GetBlockDataAsync(next.HashBlock).GetAwaiter().GetResult();
```
dangershony: I think you need to also check `nextblock.Block == null`

fassadlr: OK yeah fixed.

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Stratis.Bitcoin.Features.BlockStore
+{
+    /// <summary>
+    /// Prunes the block store database by deleting blocks lower than a certain height.
+    /// </summary>
+    public interface IPrunedBlockRepository
+    {
+        /// <summary>
+        /// INitializes the pruned block repository.
```
bokobza:
```suggestion
        /// Initializes the pruned block repository.
```

src/Stratis.BreezeD/Properties/launchSettings.json
```diff
@@ -14,7 +14,7 @@
     },
     "Stratis Test": {
       "commandName": "Project",
-      "commandLineArgs": "stratis -testnet -debug=all -loglevel=debug"
+      "commandLineArgs": "stratis -testnet -prune=2880"
```
bokobza: Maybe create a new one here.

fassadlr: I changed this so that we always start up breeze here with pruning on ;)

How do you mean "create a new one"?

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
```diff
@@ -189,8 +189,8 @@ public override void ValidateDependencies(IFullNodeServiceProvider services)
             if (this.minerSettings.Mine || this.minerSettings.Stake)
             {
                 services.Features.EnsureFeature<BlockStoreFeature>();
-                var blockSettings = services.ServiceProvider.GetService<StoreSettings>();
-                if (blockSettings.Prune)
+                var storeSettings = services.ServiceProvider.GetService<StoreSettings>();
+                if (storeSettings.Prune != 0)
                     throw new ConfigurationException("BlockStore prune mode is incompatible with mining and staking.");
```
noescape00: But staking or mining should be enabled in pruned mode. 
I see no reason why it can't be enabled

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -123,7 +123,7 @@ public override Task InitializeAsync()
             }
 
             // Signal to peers that this node can serve blocks.
-            this.connectionManager.Parameters.Services = (this.storeSettings.Prune != 0 ? NetworkPeerServices.Nothing : NetworkPeerServices.Network);
+            this.connectionManager.Parameters.Services = (this.storeSettings.PruningEnabled ? NetworkPeerServices.Nothing : NetworkPeerServices.Network);
```
noescape00: Please add a TODO here to add NetworkLimited which is what BTC uses for pruned nodes.

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Compacts the block and transaction database by resaving the database file without
+        /// all the deleted references.
+        /// </summary>
+        /// <param name="consensusTip">The current tip of consensus.</param>
+        /// <param name="network">The network the node is running on.</param>
+        /// <param name="nodeInitializing">Indicates whether or not this method is called from node startup or not.</param>
+        /// <returns>The awaited task.</returns>
+        Task PruneDatabase(ChainedHeader consensusTip, Network network, bool nodeInitializing);
```
noescape00: This is not pruning, it's a cleanup. 

Please change it to remove the blocks during the runtime instead of removing them on startup and shutdown. 

Reasons why cleanup is not good enough: 

1) It makes startup and shutdown a lot longer
2) Node still can store whole blockchain if you sync it in IBD. It doesn't work for devices with small amount of HDD. Imagine you want to sync BTC network and you have 20 gigs on your hard drive. Your only way would be to stop and start the node every 10-15 gb or smth.

noescape00:
@fassadlr please implement pruning over db cleanup.

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
fassadlr: We can't enable staking or mining here as this was not how Breeze was before.

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
fassadlr: This is pruning? We are deleting blocks from the database whilst keeping a certain amount of blocks on disk?

It is already pruning during run time., see the `LightWalletBlockStoreService` which runs a period task to prune the block store.

Compacting only releases unused disk space on start up and shut down.

The only way start up will take long is if you enable pruning on a full synced node. We can change this so that pruning is only allowed on a empty data folder node?

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.Miner/MiningFeature.cs
noescape00: Yeah. But prune mode is not for breeze only. FN will have pruning option too. 

Breeze is basically a FN with pruning enabled.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -18,6 +18,20 @@ namespace Stratis.Bitcoin.Features.BlockStore
     /// </summary>
     public interface IBlockRepository : IBlockStore
     {
+        /// <summary> The dbreeze database engine.</summary>
+        DBreezeEngine DBreeze { get; }
+
+        /// <summary>
+        /// Delete blocks and their transactions.
```
noescape00: deletes blocks and indexes for transactions that belong to deleted blocks

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -18,6 +18,20 @@ namespace Stratis.Bitcoin.Features.BlockStore
...
+        /// <para>
+        /// It should be noted that this does not delete the entries from disk (only the references are removed) and
+        /// as such the file size remains the same.
```
noescape00: Add a TODO to fix this since it's a dbreeze problem, not repository's

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -18,6 +18,20 @@ namespace Stratis.Bitcoin.Features.BlockStore
...
+        /// </para>
+        /// </summary>
+        /// <param name="hashes">List of block hashes to be deleted.</param>
+        /// <returns>The awaited task.</returns>
```
noescape00: no need to have a comment like that

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
```diff
@@ -530,6 +540,29 @@ public Task DeleteAsync(HashHeightPair newTip, List<uint256> hashes)
...
+            Task task = Task.Run(() =>
+            {
+                using (DBreeze.Transactions.Transaction transaction = this.DBreeze.GetTransaction())
+                {
+                    transaction.SynchronizeTables(BlockRepository.BlockTableName, BlockRepository.CommonTableName, BlockRepository.TransactionTableName);
+                    transaction.ValuesLazyLoadingIsOn = false;
+
+                    List<Block> blocks = this.GetBlocksFromHashes(transaction, hashes);
+
+                    this.OnDeleteBlocks(transaction, blocks.Where(b => b != null).ToList());
```
noescape00: Rename `OnDeleteBlocks` to `DeleteBlocksInternal` or smth. Same with txes. 
When methods start's with `On` it's a callback from an event. In this case it's not a callback

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -141,6 +169,7 @@ public static IFullNodeBuilder UseBlockStore(this IFullNodeBuilder fullNodeBuild
                     {
                         services.AddSingleton<IBlockStoreQueue, BlockStoreQueue>().AddSingleton<IBlockStore>(provider => provider.GetService<IBlockStoreQueue>());
                         services.AddSingleton<IBlockRepository, BlockRepository>();
+                        services.AddSingleton<IPrunedBlockRepository, PrunedBlockRepository>();
```
noescape00: Shouldn't it be the same instance as block repo? 

So instead of block repo you have pruned block repo, but you have 2 interfaces. So when you ask for block repo or pruned block repo you always receive instance of pruned block repo

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -82,15 +86,32 @@ private void AddInlineStats(StringBuilder log)
 
             if (highestBlock != null)
             {
-                string logString = $"BlockStore.Height: ".PadRight(LoggingConfiguration.ColumnLength + 1) + highestBlock.Height.ToString().PadRight(8) +
-                             $" BlockStore.Hash: ".PadRight(LoggingConfiguration.ColumnLength - 1) + highestBlock.HashBlock;
-
-                log.AppendLine(logString);
+                if (!this.storeSettings.PruningEnabled)
```
noescape00: why logging is disabled if in pruned node? 
even when we prune we have BSQ tip

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
```diff
@@ -82,15 +86,32 @@ private void AddInlineStats(StringBuilder log)
...
+                    builder.Append("BlockStore.Height: ".PadRight(LoggingConfiguration.ColumnLength + 1) + highestBlock.Height.ToString().PadRight(8));
+                    builder.Append(" BlockStore.Hash: ".PadRight(LoggingConfiguration.ColumnLength - 1) + highestBlock.HashBlock);
+                    log.AppendLine(builder.ToString());
+                }
             }
         }
 
         public override Task InitializeAsync()
         {
+            this.prunedBlockRepository.InitializeAsync().GetAwaiter().GetResult();
```
noescape00: this is an async method. please await

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
```diff
@@ -91,6 +85,12 @@ protected override void OnNextCore(ChainedHeaderBlock blockPair)
                 return;
             }
 
+            if (this.storeSettings.PruningEnabled)
```
noescape00: we should still announce blocks to peers when we are in pruned mode

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Prunes and compacts the block store database by deleting blocks lower than a certain height and recreating the database file on disk.
+    /// </summary>
+    public interface IPrunedBlockRepository
+    {
+        /// <summary>
+        /// Initializes the pruned block repository.
+        /// </summary>
+        /// <returns>The awaited task.</returns>
```
noescape00: no need for this comment

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// The method first prunes by deleting blocks from the block store that are below the <see cref="StoreSettings.AmountOfBlocksToKeep"/> from the store tip.
+        /// </para>
+        /// <para>
+        /// Once this is done the database is compacted by resaving the database file without the deleted references, reducing the file size on disk.
+        /// </para>
+        /// </summary>
+        /// <param name="blockStoreTip">The current tip of the store.</param>
+        /// <param name="network">The network the node is running on.</param>
+        /// <param name="nodeInitializing">Indicates whether or not this method is called from node startup or not.</param>
+        /// <returns>The awaited task.</returns>
```
noescape00: please remove this type of comment from everywhere in this PR

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
```diff
@@ -0,0 +1,50 @@
...
+        Task PruneAndCompactDatabase(ChainedHeader blockStoreTip, Network network, bool nodeInitializing);
+
+        /// <summary> 
+        /// The lowest block height that the repository has.
```
noescape00: hash and height.

src/Stratis.Bitcoin.Features.LightWallet/Blocks/ILightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// This service starts an async loop task that periodically deletes from the blockstore.
+    /// <para>
+    /// If the height of the node's block store is more than <see cref="LightWalletBlockStoreService.MaxBlocksToKeep"/>, the node will 
+    /// be pruned, leaving a margin of <see cref="LightWalletBlockStoreService.MaxBlocksToKeep"/> in the block database.
+    /// </para>
+    /// <para>
+    /// For example if the block store's height is 5000, the node will be pruned up to height 4000, meaning that 1000 blocks will be kept on disk.
+    /// </para>
+    /// </summary>
+    public interface ILightWalletBlockStoreService : IDisposable
```
noescape00: please move this interface to the file with implementation. Remove light wallet from the name. It's not only for light wallet, it's for FN as well

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
```diff
@@ -0,0 +1,178 @@
...
+                    transaction.Commit();
+                }
+            }
+
+            if (nodeInitializing)
+            {
+                if (IsDatabasePruned())
+                    return;
+
+                await this.PrepareDatabaseForCompactingAsync(blockRepositoryTip);
```
noescape00: configureawait(false)

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <inheritdoc/>
+        public void Start()
+        {
+            this.PrunedUpToHeaderTip = this.chainState.BlockStoreTip.GetAncestor(this.prunedBlockRepository.PrunedTip.Height);
+
+            this.asyncLoop = this.asyncLoopFactory.Run($"{this.GetType().Name}.{nameof(this.PruneBlocksAsync)}", async token =>
+            {
+                await PruneBlocksAsync();
```
noescape00: this.
and configureawait

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,114 @@
...
+            this.blockRepository = blockRepository;
+            this.prunedBlockRepository = prunedBlockRepository;
+            this.chainState = chainState;
+            this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
+            this.nodeLifetime = nodeLifetime;
+            this.storeSettings = storeSettings;
+        }
+
+        /// <inheritdoc/>
+        public void Start()
```
noescape00: we use Initialize()

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Delete blocks continuously from the back of the store.
+        /// </summary>
+        /// <returns>The awaited task.</returns>
+        private async Task PruneBlocksAsync()
+        {
+            if (this.blockRepository.TipHashAndHeight.Height < this.storeSettings.AmountOfBlocksToKeep)
+                return;
```
noescape00: for every return there should be a trace log like `(-)[SOME_MESSAGE]`

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (this.blockRepository.TipHashAndHeight.Height == (this.PrunedUpToHeaderTip?.Height ?? 0))
+                return;
+
+            if (this.blockRepository.TipHashAndHeight.Height < (this.PrunedUpToHeaderTip?.Height ?? 0 + this.storeSettings.AmountOfBlocksToKeep))
+                return;
+
+            var heightToPruneFrom = this.blockRepository.TipHashAndHeight.Height - this.storeSettings.AmountOfBlocksToKeep;
```
noescape00: unvar

noescape00:
Didn't finish reviewing the PR, please check whole PR for codestyle violations.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: Removed all of them.

src/Stratis.Bitcoin.Features.BlockStore/BlockRepository.cs
fassadlr: This is changes that is not related to the PR. We can do later.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
fassadlr: Block Repository stands on its own. Block repo is injected into PrunedBlockRepo.

This is too big a change to make now and can be improved later 👍

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
fassadlr: With the light wallet we never had a block store height but I suppose we can always show seeing as pruning is a FN function. I will fix.

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreFeature.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.BlockStore/BlockStoreSignaled.cs
fassadlr: Are you sure? As this line was actually right at the top of the method. I had to move to blocks can be downloaded (but still not announced).

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
fassadlr: Done

src/Stratis.Bitcoin.Features.BlockStore/IPrunedBlockRepository.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/Blocks/ILightWalletBlockStoreService.cs
fassadlr: I would like to keep interfaces in separate files. I will rename.

src/Stratis.Bitcoin.Features.BlockStore/PrunedBlockRepository.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed

src/Stratis.Bitcoin.Features.LightWallet/Blocks/LightWalletBlockStoreService.cs
fassadlr: Fixed.

PR #2975, opened 2018-12-18
Add maxblkmem configuration option

mikedennis:
Lets you configure the unused block cache size used during IBD. I think will be useful for low memory devices as the default is 200MB.


mikedennis:
Yeah I think the big guy is something to do with headers and hashing but this one was low hanging fruit.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -27,11 +28,11 @@ public class ConsensusManager : IConsensusManager
         /// Maximum memory in bytes that can be taken by the blocks that were downloaded but
         /// not yet validated or included to the consensus chain.
         /// </summary>
-        private long maxUnconsumedBlocksDataBytes = 200 * 1024 * 1024;
+        public long MaxUnconsumedBlocksDataBytes { get; private set; }
```
fassadlr: @mikedennis this can still have an access modifier of private or protected  as it is not being used outside of the component.

noescape00:
Looks good to me

src/Stratis.Bitcoin/Configuration/Settings/ConsensusSettings.cs
```diff
@@ -40,6 +46,7 @@ public ConsensusSettings(NodeSettings nodeSettings)
             this.UseCheckpoints = config.GetOrDefault<bool>("checkpoints", true, this.logger);
             this.BlockAssumedValid = config.GetOrDefault<uint256>("assumevalid", nodeSettings.Network.Consensus.DefaultAssumeValid, this.logger);
             this.MaxTipAge = config.GetOrDefault("maxtipage", nodeSettings.Network.MaxTipAge, this.logger);
+            this.MaxBlockMemoryInMB = config.GetOrDefault("maxblkmem", 200, this.logger);
```
noescape00: I'd probably add min value and throw here if it's too low. 

Min value should be smth like: `max block size * amount of blocks we expect to be generated in 1h`

fassadlr: @noescape00 good idea! How about 50-60 blocks then? We dont actually need to add min property just check that `maxblkmem` isn't less than 60MB (set as a `const`)?

noescape00: We can't hardcode the amount of blocks since it will differ between the chains, better to calculate it based on the target spacing

dangershony: Lets just put the min value to 200 easy and will avoid complications, for now I am happy for this as it is.

dangershony: @mikedennis submit a second PR that throws if size is less then 200.

dangershony: It should limit the amount of blocks you download at any given moment.

PR #2979, opened 2018-12-19
Fix visibility of Address.Create

rowandh:
Address.Create is internal to Stratis.SmartContracts. In the SBFN solution, Stratis.SmartContracts is a friend assembly of all other smart contract projects, but this does not carry over to the NuGet packages.

### DIFF
--- a/src/Stratis.SmartContracts.CLR/AddressExtensions.cs
+++ b/src/Stratis.SmartContracts.CLR/AddressExtensions.cs
@@ -18,12 +18,12 @@ namespace Stratis.SmartContracts.CLR
 
         public static Address ToAddress(this uint160 address)
         {
-            return Address.Create(address.ToBytes());
+            return CreateAddress(address.ToBytes());
         }
 
         public static Address ToAddress(this string address, Network network)
         {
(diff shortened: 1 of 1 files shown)


src/Stratis.SmartContracts.CLR/AddressExtensions.cs
```diff
@@ -18,32 +18,43 @@ public static uint160 ToUint160(this string base58Address, Network network)
...
-            return Address.Create(result);
+            return CreateAddress(result);
         }
 
         public static string ToBase58Address(this uint160 address, Network network)
         {
             return new BitcoinPubKeyAddress(new KeyId(address), network).ToString();
         }
+
+        private static Address CreateAddress(byte[] bytes)
```
bokobza: It looks like the Address.Create is not used anywhere after this change (apart from tests) Can we remove it?

rowandh: We can, but I didn't want to make any changes to the Stratis.SmartContracts NuGet package.

bokobza: ok

PR #2980, opened 2018-12-19
[Breeze] Check if PruneEnabled in LightWalletFeature

fassadlr:


src/Stratis.Bitcoin.Features.LightWallet/LightWalletFeature.cs
```diff
@@ -164,7 +165,9 @@ public override void Dispose()
             this.asyncLoop.Dispose();
             this.walletSyncManager.Stop();
             this.walletManager.Stop();
-            this.lightWalletBlockStoreService.Dispose();
+
+            if (this.storeSettings.PruningEnabled)
```
bokobza: maybe do `this.lightWalletBlockStoreService?.Dispose();` without checking for PruningEnabled, just in case.

PR #2987, opened 2018-12-20
Add a proxy wallet rpc controller for cold staking wallet.

### LINKED ISSUE #2986: Staking wallet causes wallet RPC to no longer function
Thanks to @sondreb for pointing this out.

I did a quick test and this is indeed the case. If you build the fullnode with .UseWallet() wallet RPC's work fine. If you have .UseStakingWallet() the wallet RPC's no longer work.

I believe the root cause is related to this code in WebHostExtensions.cs (in RPC Feature):
```
public static IWebHostBuilder ForFullNode(this IWebHostBuilder hostBuilder, FullNode fullNode)
        {
            hostBuilder.ConfigureServices(s =>
            {
                IMvcCoreBuilder mvcBuilder = s.AddMvcCore(o =>
                {
                    o.ModelBinderProviders.Insert(0, new DestinationModelBinder());
                    o.ModelBinderProviders.Insert(0, new MoneyModelBinder());
                });

                // Include all feature assemblies for action discovery otherwise RPC actions will not execute
                // https://stackoverflow.com/questions/37725934/asp-net-core-mvc-controllers-in-separate-assembly
                foreach (Assembly assembly in fullNode.Services.Features.OfType<FullNodeFeature>().Select(x => x.GetType().GetTypeInfo().Assembly).Distinct())
                    mvcBuilder.AddApplicationPart(assembly);
            });

            return hostBuilder;
        }
```

Basically this is sucking in all the controllers from the various added features. But WalletRPCController is in WalletFeature NOT ColdStakingFeature. So it's not included in the RPCHost.

Quick and dirty fix is to add an instance to ColdStakingFeature but maybe the logic of above needs to get rethought to look through a feature's services and then find those controllers instead of just the assembly for the feature.

mikedennis:
Fix #2986

### DIFF
--- a/src/Stratis.Bitcoin.Features.ColdStaking/Controllers/ColdStakingWalletRPCController.cs
+++ b/src/Stratis.Bitcoin.Features.ColdStaking/Controllers/ColdStakingWalletRPCController.cs
@@ -1,0 +1,16 @@
+﻿using Microsoft.Extensions.Logging;
+using Stratis.Bitcoin.Features.Wallet;
+using Stratis.Bitcoin.Features.Wallet.Interfaces;
+
+namespace Stratis.Bitcoin.Features.ColdStaking.Controllers
+{
+    /// <summary> All functionality is in WalletRPCController, just inherit the functionality in this feature.</summary>
+    public class ColdStakingWalletRPCController : WalletRPCController
+    {
+          
+        public ColdStakingWalletRPCController(IWalletManager walletManager, IWalletTransactionHandler walletTransactionHandler, IFullNode fullNode, IBroadcasterManager broadcasterManager, ILoggerFactory loggerFactory) : 
+            base(walletManager, walletTransactionHandler, fullNode, broadcasterManager, loggerFactory)
+        {
+        }
+    }
+}


src/Stratis.Bitcoin.Features.ColdStaking/Controllers/ColdStakingWalletRPCController.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.Extensions.Logging;
+using Stratis.Bitcoin.Features.Wallet;
+using Stratis.Bitcoin.Features.Wallet.Interfaces;
+
+namespace Stratis.Bitcoin.Features.ColdStaking.Controllers
+{
+    /// <summary> All functionality is in WalletRPCController, just inherit the functionality in this feature.</summary>
+    public class ColdStakingWalletRPCController : WalletRPCController
```
dangershony: Why do we need this? is the low level controller not registered by default?

mikedennis: Check out #2986, it's because the RPC controller discovery is done at the feature level by assembly and the wallet RPC controller resides in another feature's assembly.

dangershony: ah right ok yes and we never actually register the wallet controller.

fassadlr:
@mikedennis can you confirm that this works as expected?

PR #2988, opened 2018-12-20
[SC] Changes to Address and version bumps

rowandh:


src/Stratis.SmartContracts/Address.cs
```diff
@@ -25,25 +25,22 @@ public Address(Address other)
...
-        internal Address(uint pn0, uint pn1, uint pn2, uint pn3, uint pn4)
+        /// <summary>
+        /// Creates a new 160-bit wide Address.
+        /// </summary>
+        /// <param name="pn0">The first 32 bits of the address.</param>
+        /// <param name="pn1">The second 32 bits of the address.</param>
+        /// <param name="pn2">The third 32 bits of the address.</param>
+        /// <param name="pn3">The fourth 32 bits of the address.</param>
+        /// <param name="pn4">The last 32 bits of the address.</param>
+        public Address(uint pn0, uint pn1, uint pn2, uint pn3, uint pn4)
```
codingupastorm: In normal circumstances I think we would likely agree to try and avoid this, but given the current circumstances, I think it's harmless 👍

PR #2991, opened 2018-12-21
Useful validation message

codingupastorm:


src/Stratis.SmartContracts.CLR/Exceptions/SmartContractValidationException.cs
```diff
@@ -10,6 +12,13 @@ namespace Stratis.SmartContracts.CLR.Exceptions
     /// <remarks>TODO: We can possibly merge this with <see cref="SmartContractValidationResult"/>.</remarks>
     public sealed class SmartContractValidationException : SmartContractException
```
rowandh: We could get rid of this in a future PR

src/Stratis.SmartContracts.CLR/Exceptions/SmartContractValidationException.cs
```diff
@@ -10,6 +12,13 @@ namespace Stratis.SmartContracts.CLR.Exceptions
...
     {
+        public override string Message
+        {
+            get
+            {
+                return base.Message + "\n" + string.Join('\n', this.Errors.Select(x=> x.Message));
```
bokobza: Better use Environment.NewLine to let dotnet decide how to handle that.

PR #2999, opened 2018-12-22
Injectable StandardScripts

codingupastorm:
Sets a base for future work allowing STRAT / BTC / sidechains to have different standards

Possible value for ColdStaking too, rather than use of static registry


zeptin:
@codingupastorm `sending_from_one_address_to_fifty_addresses` is failing, but it's unclear if it's because of this PR. Looking at the test logs locally the transaction goes into the mempool as an orphan (inputs not found yet), but then it never gets put into the wallet on the destination node.

codingupastorm:
@zeptin it's failing in master for me locally too atm, maybe a recent PR messed with it?

dangershony:
We are probably going to need this PR pushed earlier then later to fix the OP_RETURN issue @quantumagi  FYI

PR #3013, opened 2018-12-27
[Pruning] Create unit tests for PruneBlockRepository and PruneBlockStoreService

fassadlr:
Also, I moved the Prune classes to the block store feature. Next up is to move it to Stratis.Bitcoin.

### DIFF
--- a/src/Stratis.Bitcoin.Features.BlockStore.Tests/PruneBlockStoreServiceTests.cs
+++ b/src/Stratis.Bitcoin.Features.BlockStore.Tests/PruneBlockStoreServiceTests.cs
@@ -1,0 +1,219 @@
+﻿using System.Threading.Tasks;
+using Moq;
+using NBitcoin;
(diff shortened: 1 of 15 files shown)


src/Stratis.Bitcoin.Features.BlockStore.Tests/PruneBlockStoreServiceTests.cs
```diff
@@ -0,0 +1,219 @@
+﻿using System.Threading.Tasks;
```
dangershony: Should this whole class go in light wallet tests?

fassadlr: Hi @dangershony yes, `PruneBlockRepository` and `PruneBlockStoreService` will now be in the blockstore feature, so the tests should be here as well.

They are both block store features/components and is not specific to the light wallet per se 👍

PR #3017, opened 2018-12-30
[RPC] - Refresh wallet unlock when transaction is sent

### LINKED ISSUE #3015: Could not build the transaction. Details: Script error on input 0 (InvalidStackOperation)
I received the following error when perform a send using the "sendtoaddress" RPC method:

```
"Could not build the transaction. Details: Script error on input 0 (InvalidStackOperation)"
```

I was able to reproduce the issue for a while:

I had a single wallet with a single UTXO, and I was attempting to send about half of what existed in the UTXO. I did a lot of debugging, and found that the problem spawns here:

image

I did the operation many times, and I restarted debugging in Visual Studio a couple of times. The third time, the payment was successful. I had done no changes.

This was on TEST net I reproduced it, and it has coinbase maturity of 10. At the time I was testing, the UTXO had 40+ in maturity, and 50 on the last test that failed.

A few confirmations later, and a node restart, it worked.

I doubt this is a maturity issue, as I have easily been able to send many other transactions using the "sendtoaddress" RPC call, without waiting more than a single confirmation.

mikedennis:
Fixes #3015 

Previously would reset the unlock timeout back to the hardcoded 5 minutes after creating the transaction instead of using the duration specified in the original unlocking request.

Also changed the error handling for signing the transaction in the wallet transaction builder context. Make use of "Sign" in the context to specify when to sign the transaction instead of relying on passing in a null password for wallet. Otherwise can get strange errors that are hard to diagnose such as the one specified in the issue above.

### DIFF
--- a/src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
+++ b/src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
@@ -38,6 +38,8 @@ namespace Stratis.Bitcoin.Features.Wallet
 
         private readonly MemoryCache privateKeyCache;
 
+        private readonly Dictionary<string, TimeSpan> privateKeyCacheTimeout;
+
         protected readonly StandardTransactionPolicy TransactionPolicy;
 
         private readonly IWalletManager walletManager;
@@ -56,6 +58,7 @@ namespace Stratis.Bitcoin.Features.Wallet
             this.walletFeePolicy = walletFeePolicy;
             this.logger = loggerFactory.CreateLogger(this.GetType().FullName);
             this.privateKeyCache = new MemoryCache(new MemoryCacheOptions() { ExpirationScanFrequency = new TimeSpan(0, 1, 0) });
+            this.privateKeyCacheTimeout = new Dictionary<string, TimeSpan>();
             this.TransactionPolicy = transactionPolicy;
         }
 
@@ -198,11 +201,13 @@ namespace Stratis.Bitcoin.Features.Wallet
             if (this.privateKeyCache.TryGetValue(cacheKey, out SecureString secretValue))
             {
                 this.privateKeyCache.Set(cacheKey, secretValue, duration);
+                this.privateKeyCacheTimeout.AddOrReplace(cacheKey, duration);
             }
             else
             {
                 Key privateKey = Key.Parse(wallet.EncryptedSeed, walletPassword, wallet.Network);
                 this.privateKeyCache.Set(cacheKey, privateKey.ToString(wallet.Network).ToSecureString(), duration);
+                this.privateKeyCacheTimeout.AddOrReplace(cacheKey, duration);
             }
         }
 
@@ -214,6 +219,7 @@ namespace Stratis.Bitcoin.Features.Wallet
             Wallet wallet = this.walletManager.GetWalletByName(walletAccount.WalletName);
             string cacheKey = wallet.EncryptedSeed;
             this.privateKeyCache.Remove(cacheKey);
+            this.privateKeyCacheTimeout.Remove(cacheKey);
         }
 
         /// <summary>
@@ -257,12 +263,24 @@ namespace Stratis.Bitcoin.Features.Wallet
             if (this.privateKeyCache.TryGetValue(cacheKey, out SecureString secretValue))
             {
                 privateKey = wallet.Network.CreateBitcoinSecret(secretValue.FromSecureString()).PrivateKey;
-                this.privateKeyCache.Set(cacheKey, secretValue, new TimeSpan(0, 5, 0));
+                TimeSpan duration;
+                if (!this.privateKeyCacheTimeout.TryGetValue(cacheKey, out duration))
+                {
+                    // if for some reason we can't find the previous duration then use 5 minutes.
+                    duration = new TimeSpan(0, 5, 0);
+                }
+
+                this.privateKeyCache.Set(cacheKey, secretValue, duration);
+                this.privateKeyCacheTimeout.AddOrReplace(cacheKey, duration);
             }
             else
             {
                 privateKey = Key.Parse(wallet.EncryptedSeed, context.WalletPassword, wallet.Network);
-                this.privateKeyCache.Set(cacheKey, privateKey.ToString(wallet.Network).ToSecureString(), new TimeSpan(0, 5, 0));
+
+                // Wallet unlocked with password gets a 5 minute duration.
+                TimeSpan duration = new TimeSpan(0, 5, 0); 
+                this.privateKeyCache.Set(cacheKey, privateKey.ToString(wallet.Network).ToSecureString(), duration);
+                this.privateKeyCacheTimeout.AddOrReplace(cacheKey, duration);
             }
 
             var seedExtKey = new ExtKey(privateKey, wallet.ChainCode);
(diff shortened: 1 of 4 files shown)


src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -257,15 +263,24 @@ protected void AddSecrets(TransactionBuildContext context)
...
+                    // if for some reason we can't find the previous duration then use 5 minutes.
+                    duration = new TimeSpan(0, 5, 0);
+                }
+
+                this.privateKeyCache.Set(cacheKey, secretValue, duration);
+                this.privateKeyCacheTimeout.AddOrReplace(cacheKey, duration);
             }
             else
             {
-                if (string.IsNullOrEmpty(context.WalletPassword))
```
fassadlr: So are we saying that `context.WalletPassword` can never be null?

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -257,15 +263,24 @@ protected void AddSecrets(TransactionBuildContext context)
             if (this.privateKeyCache.TryGetValue(cacheKey, out SecureString secretValue))
             {
                 privateKey = wallet.Network.CreateBitcoinSecret(secretValue.FromSecureString()).PrivateKey;
-                this.privateKeyCache.Set(cacheKey, secretValue, new TimeSpan(0, 5, 0));
+                TimeSpan duration;
+                if (!this.privateKeyCacheTimeout.TryGetValue(cacheKey, out duration))
```
bokobza: @mikedennis do you know why isn't `AddSecrets()` using `CacheSecret()`?

mikedennis: It's already got cachekey so would be a bit repetitive code.  Could probably move the code for updating cache and timeout to a common method and call from both though?

bokobza: What I'm saying is that the method AddSecrets() has code that does what CacheSecret () does but not using it.

this is inside AddSecrets():
https://github.com/stratisproject/StratisBitcoinFullNode/blob/d7d152346200ff65d973626f3b3d17be996e0042/src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs#L245-L262

this is CacheSecrets():
https://github.com/stratisproject/StratisBitcoinFullNode/blob/d7d152346200ff65d973626f3b3d17be996e0042/src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs#L182-L200

If you change CacheSecrets to return the private key you can use this method inside AddSecrets rather than having the same code duplicated.

bokobza:
The problem here is that RPC and API behave differently.
Every time we build a transaction via the API, the cache duration gets reset back to 5 minutes. For RPC, we set it in `walletpassphrase` and then when time is up, we have to call it again, no matter how many times we call `sendtoaddress`.

The best idea I could come up with is to add a boolean in the `TransactionBuildContext` object specifying whether we want to keep the secret in cache (maybe `bool CacheSecret`).
This way:
- it's more configurable (even an API user can decide not to want to cache anything)
- the RPC `sendtoaddress` can build a transaction without changing the current duration set by `walletpassphrase`.

Thoughts? @mikedennis @sondreb

sondreb:
Sounds good! What I had to add was a solution to auto-unlock on startup and keep it unlocked for the maximum length (many years). With that setup I had to populate an array in WalletManager, which WalletTransactionHandler parsed in constructor. I don't see any need for multiple wallet un-lock on startup, when the usecase is basically to support a single default wallet for RPC-compatibility.

So any solution works, I just think we must fix this issue so that wallets don't get auto-locked when they have been explicitly unlocked for X amount of time.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
mikedennis: Yes that would work. If I changed CacheSecret() to return the private key then we could be used from AddSecrets().  I'm debating whether we might want to change the name of the method then to GetPrivateKey() or something like that?

mikedennis:
I double checked the bitcoin code, and you are indeed correct. It seems like it's a absolute unlock, not a relative unlock from the last transaction like the API. I like your idea @bokobza but I wonder whether it would cause confusing behaviour if using a combination of walletUI/RPC/API. For example say I unlocked the wallet with RPC and then did a transaction with the wallet UI then my unlock would be reset back to 5 minutes.

So I guess the question becomes whether it's more important to have consistent behaviour regardless of interface or whether it is better to have RPC behave the same as bitcoin.

I'm think I'm weighing towards Jeremy's solution and having RPC behave exactly the same as bitcoin. Do we still want the logic here for resetting the cache to the same duration as it was originally cached or back to the hardcoded 5 minutes?

sondreb:
Oh, I did not catch that it was different. I think they should work the same, absolute timeouts and only that. And same across APIs.

bokobza:
1. I think the RPC one should be like bitcoin (or at least StratisX) so that it'll be easier for people to use or upgrade to.

2. Absolute timeouts is my favourite option too. 

3. If someone uses RPC to unlock the wallet for a year and then uses the API, the duration will get overwritten to 5 minutes indeed, unless they use the API with CacheSecret = false. 
In practice, my view is that it's unlikely someone would use both to create transactions.

So I think if we add CacheSecret (set to true by default), we're good.

mikedennis:
Sounds good. I will refactor to add CacheSecret property to the tx build context and update the RPC endpoints.

mikedennis:
Ran as many tests as I could in the time I had. Smart contract integration tests are failing but not sure due to my change or something else. Unfortunately ran out of time to investigate further.

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
bokobza: I think you can leave the name as is.
CacheSecret caches the secret and returns the cached item. I think the caching is the real function here.
A bit like when you create an object via a POST api call and you return the created object because it may be useful. Same here.

bokobza:
This is now ready for a final review.
@sondreb @mikedennis @dangershony

src/Stratis.Bitcoin.Features.Wallet/WalletTransactionHandler.cs
```diff
@@ -179,7 +179,7 @@ public Money EstimateFee(TransactionBuildContext context)
 
         /// <inheritdoc />
         [NoTrace]
-        public void CacheSecret(WalletAccountReference walletAccount, string walletPassword, TimeSpan duration)
+        public SecureString CacheSecret(WalletAccountReference walletAccount, string walletPassword, TimeSpan duration)
```
mikedennis: It looks as though no code currently consumes the return value from this method, so I'll write a quick unit test that tests this just to make sure it is good to go.

bokobza:
Looks good.
Thanks @mikedennis !

bokobza:
Breaking change: Now in order to cache the secret, `CacheSecret = true` needs to be used when creating a transaction context.

PR #3022, opened 2019-01-03
Fix memory leak in memcache

dangershony:


src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -65,6 +65,16 @@ public MemoryCache(IEqualityComparer<TKey> comparer = null)
         /// <returns><c>true</c> if cache contains the item, <c>false</c> otherwise.</returns>
         protected abstract bool IsCacheFull(CacheItem item);
 
+        /// <summary>And item was added to the cache.</summary>
```
fassadlr: "An item was added..."

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
```diff
@@ -65,6 +65,16 @@ public MemoryCache(IEqualityComparer<TKey> comparer = null)
...
+        protected virtual void ItemAddedLocked(CacheItem item)
+        {
+        }
+
+        /// <summary>And item was removed from the cache.</summary>
```
fassadlr: "An item was added..."

src/Stratis.Bitcoin/Utilities/MemoryCache.cs
dangershony: ah well spotted

dangershony:
Add also better log of cache size
image

PR #3026, opened 2019-01-03
Removed saving the wallet every time we receive a trx affecting the user's wallets.

bokobza:
This will lead to an improvement in performance.

**When is the wallet saved?**
The wallets are saved every 5 minutes + every time the user creates an account, addresses or the user removes all the transactions.
In the worst case, like a node crash, the wallet file is 4:59 minutes old. On restart, the wallet will just get updated with the 5 minutes old blocks.


dangershony:
@zeptin possibly yes, though that block should be accepted by the network and then we can just download it if we need to.

PR #3027, opened 2019-01-04
Fix broken test

mikedennis:


mikedennis:
Block hashes for both blocks inserted were the same so that's why there was only one block and why the test failed previously.

PR #3031, opened 2019-01-04
[ProvenHeader] Add extra trace logging

fassadlr:


src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderRepository.cs
```diff
@@ -147,7 +147,7 @@ public Task PutAsync(SortedDictionary<int, ProvenBlockHeader> headers, HashHeigh
 
             if ((this.provenBlockHeaderTip != null) && (newTip.Hash == this.provenBlockHeaderTip.GetHash()))
             {
-                this.logger.LogTrace("(-)[BLOCKHASH_MISMATCH]");
+                this.logger.LogTrace("(-)[BLOCKHASH_MISMATCH]{0}:{1},{2}:{3})", nameof(newTip), newTip, nameof(this.provenBlockHeaderTip), this.provenBlockHeaderTip);
```
noescape00: Please separate this into 2 logs. One with the tips and another as an exit logs. 
Also add `''` for the tips

src/Stratis.Bitcoin.Features.Consensus/ProvenBlockHeaders/ProvenBlockHeaderStore.cs
```diff
@@ -188,6 +188,8 @@ public async Task<ProvenBlockHeader> GetAsync(int blockHeight)
         /// <inheritdoc />
         public void AddToPendingBatch(ProvenBlockHeader provenBlockHeader, HashHeightPair newTip)
         {
+            this.logger.LogTrace("({0}:{1},{2}:{3})", nameof(provenBlockHeader), provenBlockHeader, nameof(newTip), newTip);
```
noescape00: `''` are missing. 

I think it's better to enable trace logs in fody config instead of adding manual entry\exit logs

PR #3033, opened 2019-01-04
Remove too much info logs on the mempool

dangershony:


bokobza:
Why do we put these in debug? They shouldn't happen often, should they?

dangershony:
>Why do we put these in debug? They shouldn't happen often, should they?  

Yes they do, on every trx that is accepted you get a number of errors that its already in the mempool.
And besides I think info should be hsppy path mostly (and not failures)

Issue #3012, opened 2018-12-27
Error when storing a proven header


dangershony:
```
[2018-12-25 02:35:34.7851 78] FATAL: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0+<<StartAsync>b__0>d.MoveNext FlushChain threw an unhandled exception
[2018-12-25 02:35:34.7851 78] ERROR: Stratis.Bitcoin.Utilities.AsyncLoop+<>c__DisplayClass16_0+<<StartAsync>b__0>d.MoveNext FlushChain threw an unhandled exception: Stratis.Bitcoin.Features.Consensus.ProvenBlockHeaders.ProvenBlockHeaderException: Invalid new tip hash, tip hash has not changed.
   at Stratis.Bitcoin.Features.Consensus.ProvenBlockHeaders.ProvenBlockHeaderRepository.PutAsync(SortedDictionary`2 headers, HashHeightPair newTip) in C:\Users\mainnet\Desktop\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Consensus\ProvenBlockHeaders\ProvenBlockHeaderRepository.cs:line 155
   at Stratis.Bitcoin.Features.Consensus.ProvenBlockHeaders.ProvenBlockHeaderStore.SaveAsync() in C:\Users\mainnet\Desktop\StratisBitcoinFullNode\src\Stratis.Bitcoin.Features.Consensus\ProvenBlockHeaders\ProvenBlockHeaderStore.cs:line 238
   at Stratis.Bitcoin.Base.BaseFeature.<StartChainAsync>b__27_0(CancellationToken token) in C:\Users\mainnet\Desktop\StratisBitcoinFullNode\src\Stratis.Bitcoin\Base\BaseFeature.cs:line 246
```

dangershony:
[node-2018-12-26.14.zip](https://github.com/stratisproject/StratisBitcoinFullNode/files/2712006/node-2018-12-26.14.zip)

maciejzaleski:
Resolved

PR #3042, opened 2019-01-07
[ProvenHeaders] Remove same hash check

fassadlr:
This check is redundant as when we reorg we just overwrite the values in the proven header repo. I.e. it is possible to call flush chain and the `provenBlockHeaderTip` be the same as the new tip we are trying to store.

Relates to #3012


dangershony:
This looks right to me because PH store is suppose to overwrite whats in disk so it should not have any logic as to what's being persisted other then that its sequential (which is done in the calling layer)

PR #3044, opened 2019-01-07
Signal a trx that is broadcast form the wallet

dangershony:


dangershony:
@bokobza please review this.

bokobza:
If the peers don't want it, do we still want to signal it?
Does this fix something or is it for another component (i.e who's gonna consume this)?

dangershony:
Yeah @zeptin is right, now that we are spamming the network I found that if I try to build many trx mempool quickly rejects trx because outputs are already spent, this will just signal a trx to the wallet as if it came form the mempool.

PR #3045, opened 2019-01-07
[Consensus] Re-enqueue failed block download

fassadlr:


dangershony:
The problem here is that when a block failed to download we never retriggered it the download so the node got stuck.

dangershony:
>CHT handles cases when block download failed.  

I could not find anywhere CHT handles failed download


>In this case we should disconnect headers associated with failed blocks which CHT already does  

You may be correct but if download failed because of network issues and not protocol violation, then a redownload should be ok, especially that the download may come from a nother peer this time.

noescape00:
>You may be correct but if download failed because of network issues and not protocol violation, then a redownload should be ok, especially that the download may come from a nother peer this time.

In case there was a fail that didn't lead to peer being disconnected BP will move failed download to reassign queue and reassign it. Only if reassigning failed (there are no peers claiming the block) BP will fail the job and propagate the failure to the consumer (CM)

dangershony:
Ah I see @noescape00 claims that BP is already doing a redownload so this is not the correct logic.
BP will stop trying to download a block only when no peers claiming it are available.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1011,25 +1010,41 @@ private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
                     throw new InvalidOperationException("Unsolicited block");
                 }
 
-                if (block != null)
+                chainedHeader = this.chainedHeaderTree.GetChainedHeader(blockHash);
```
noescape00: I think `BlockDownloaded` is not the right place for this code. 

`BlockDownloaded` is a low level callback from BP. Actual consumer might be another component, not CHT. 

For example we are a pruned node and blocks 0-100k are pruned. Wallet is syncing from 50k. Blocks that are requested are not marked as requested by CHT because they are already part of consensus. 

So this code should be only for blocks that were requested by CHT itself.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1011,25 +1010,41 @@ private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
...
+                    {
+                        this.callbacksByBlocksRequestedHash.Remove(blockHash);
+                    }
+
+                    this.logger.LogTrace("(-)[CHAINED_HEADER_NOT_FOUND]");
+                    return;
+                }
 
-                    if (chainedHeader == null)
+                if (block == null)
```
noescape00: code starting from this line doesn't have to be in peer lock

when it's outside of the peer lock there is no need for `bool reassignDownload`, reassigning can be done right away

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1011,25 +1010,41 @@ private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
...
-                else
-                {
-                    this.logger.LogDebug("Block '{0}' failed to be delivered.", blockHash);
-                }
+            }
+
+            if (reassignDownload)
+            {
+                this.logger.LogWarning("Downloading block for '{0}' failed, it will be enqueued again.", chainedHeader);
+                this.DownloadBlocks(new[] { chainedHeader }, this.ProcessDownloadedBlock);
```
noescape00: here you are adding a callback. then on 1081 what was added will be removed. Original race condition is preserved. 
returning after the reassigning should fix the problem

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Ok but this code was there before this change, the only reason we get the chained header earlier is so that we can construct ChainedHeaderBlock later when we pass it to ProcessDownloadedBlock.

fassadlr: If you look at how it was done before we only get ChainedHeader if Block is not null... So if we do it like that we can't construct ChainedHeaderBlock.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: @dangershony suggested that it needs to be in a lock as `chainedHeader.BlockDataAvailability` can change whilst we check the block availability state?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Ah yes OK, we should just return. here. Will fix.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: I don't agree @noescape00, this is just to fix the race condition so it should be in the lower level code.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: If I am not mistaken any access to the ChainedHeader statuses should happen inside CHT and under a lock (in this case we do it outside CHT) we could move this check in to CHT (similar to other places where we call CHT only to change a status)

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1054,6 +1070,10 @@ private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
                     return;
                 }
             }
+            else
+            {
+                this.logger.LogDebug("Block '{0}' failed to be delivered.", blockHash);
```
dangershony: I would actually suggest we throw an exception here, this can not happen anymore right?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: >I don't agree @noescape00, this is just to fix the race condition so it should be in the lower level code.

So let's say we are in this scenario with pruned node. CT is at 200k, earlist block we have is 100k, we need some block below 100k. Wallet asks for it. It's `BlockDataAvailabilityState` is still `HeaderOnly`, not `BlockRequired` as in case when CHT asked for it. 

So the BP fails for some reason and we go into `BlockDownloaded` with block being `null`. On 1029 we won't go into that if and `reassignDownload` will be false. Now we will go to 1075 (and here you are suggesting to throw in another review message, but it's a valid scenario according to current logic).

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: It can. See my msg above.

noescape00:
please merge latest master in to get CM callback fix

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: >@dangershony suggested that it needs to be in a lock as chainedHeader.BlockDataAvailability can change whilst we check the block availability state?

Fair point, it's possible. 

Then please ignore my comment.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1011,25 +1010,42 @@ private void BlockDownloaded(uint256 blockHash, Block block, int peerId)
...
-                {
-                    this.logger.LogDebug("Block '{0}' failed to be delivered.", blockHash);
-                }
+            }
+
+            if (reassignDownload)
+            {
+                this.logger.LogWarning("Downloading block for '{0}' failed, it will be enqueued again.", chainedHeader);
+                this.DownloadBlocks(new[] { chainedHeader }, this.ProcessDownloadedBlock);
+                return;
```
noescape00: please add exit log

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: hmmm so basically a failed download is a valid scenario even if we have a peer that can provide this block?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
dangershony: Ok I think we are not dicussing anymore the fact tha the code for 
```
chainedHeader = this.chainedHeaderTree.GetChainedHeader(blockHash); 
```
Is outside the if `block != null` condition right?

Logically speaking this change is only to bypass the potential race condition, if the header is in the state of `BlockRequired` we have to either succeed or remove the header this is the logic I understand to be correct, but you claim that may not be the case, in that case if block download failed we have to set the state of the header back to HeaderOnly no?

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
fassadlr: Done

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
noescape00: Yes, but only due to the race condition.
Peers disconnected, a new peer connected that can provide that block, now fail callback executed.

PR #3049, opened 2019-01-08
GetOrDownloadBlocksAsync bugfix

noescape00:
Fixed bug when provided callback wouldn't be registered


noescape00:
@dangershony no, just found while drawing a map of interactions in CM when we were discussing race condition to verify it

PR #3050, opened 2019-01-08
Remove some logging about logging

bokobza:
I was looking at some full logs and I noticed that we log what we're about to log.
Look here:

image

The blue bit is the log, the red bit is the meta-log.


bokobza:
Thanks @fassadlr and @noescape00 .

PR #3051, opened 2019-01-08
Make cht internal

noescape00:


dangershony:
Looks god to me I will let @MithrilMan review as well

noescape00:
This makes integration tests fail because DI can't call CM's constructor. 

```
System.InvalidOperationException : A suitable constructor for type 'Stratis.Bitcoin.Consensus.ConsensusManager' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
```

Alternative approach is to create CHT inside of CM instead of letting DI do it but this is something that we decided not to do long time ago in favor of tests. 

Therefore closing the PR

MithrilMan:
probably for this scenario the best thing to do is having an internal method to mock IChainedHeaderTree and set externalvisibleto to test project, and exclude IChainedHeaderTree from ConsensusManager  constructor, but create instead an instance of ChainedHeaderTree in the ConsensusManager constructor body

MithrilMan:
I think you could even fix the problem, changing how we register ConsensusManager into the serviceCollection 
instead of
`services.AddSingleton<IConsensusManager, ConsensusManager>();`
you could use a Factory method that instantiate ConsensusManager
it's an overload of AddSingleton
like we did for BlockStoreQueue
`services.AddSingleton<IBlockStoreQueue, BlockStoreQueue>().AddSingleton<IBlockStore>(provider => provider.GetService<IBlockStoreQueue>());`

if you want you can reopen this PR and make changes or I can change it later

noescape00:
Reopening per request from @MithrilMan, he will work on PR

src/Stratis.Bitcoin/Base/BaseFeature.cs
```diff
@@ -384,7 +386,26 @@ public static IFullNodeBuilder UseBaseFeature(this IFullNodeBuilder fullNodeBuil
                     services.AddSingleton<ISelfEndpointTracker, SelfEndpointTracker>();
 
                     // Consensus
-                    services.AddSingleton<IConsensusManager, ConsensusManager>();
+                    services.AddSingleton<IConsensusManager>(provider => new ConsensusManager(
```
noescape00: please add a comment why CM is created differently comparing to other components

noescape00:
@dangershony I'm fine with @MithrilMan 's changes. 
Please approve if you are too.

dangershony:
@fassadlr @bokobza opinion? I dont mind this change.

fassadlr:
Looks good to me @dangershony !

PR #3052, opened 2019-01-08
Remove outpoint lookup when UTXO is spent

bokobza:
Saved from oblivion (Nov' 18).

Remove an outpoint from the lookup table when a UTXO is spent.
Recovered a 90MB wallet containing about 60,000 received outputs and 55,000 spent outputs items, all recovered without problem.


src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1340,6 +1348,20 @@ private void AddInputKeysLookupLocked(TransactionData transactionData)
...
+        /// Remove from the list of unspent outputs kept in memory.
+        /// </summary>
+        private void RemoveInputKeysLookupLock(TransactionData transactionData)
+        {
+            Guard.NotNull(transactionData, nameof(transactionData));
+            Guard.NotNull(transactionData.SpendingDetails, nameof(transactionData.SpendingDetails));
+
+            lock (this.lockObject)
+            {
+                this.outpointLookup.Remove(new OutPoint(transactionData.Id, transactionData.Index));
```
fassadlr: Is it a good idea to new up a new `OutPoint` in order to remove it?

fassadlr: Is there perhaps a better way to do this?

bokobza: That's the best way.

The `outpointLookup` object is a `Dictionary<OutPoint, TransactionData>` so you can only remove an item by specifying the key, here an `OutPoint`.
The OutPoint class overrides the `==` operator so comparison is serving our purpose here.
https://github.com/stratisproject/StratisBitcoinFullNode/blob/13b272602e63c33fee51ed0f5dc4dc29646a24b0/src/NBitcoin/Transaction.cs#L129-L140

fassadlr: Yes the implicit operator will work but if you want to remove the item without a new instance of `Outpoint` you can do this:

```                
var toRemove = this.outpointLookup.FirstOrDefault(op => op.Key.Hash == transactionData.Id && op.Key.N == transactionData.Index);
this.outpointLookup.Remove(toRemove.Key);
```

bokobza: true, but isn't `var toRemove` a new instance of `OutPoint`?

fassadlr: In this case `toRemove` will be the actual intance in the dictionary that we want to remove.

src/Stratis.Bitcoin.Features.Wallet/WalletManager.cs
```diff
@@ -1297,7 +1303,9 @@ public void LoadKeysLookupLock()
                         if (address.Pubkey != null)
                             this.scriptToAddressLookup[address.Pubkey] = address;
 
-                        foreach (TransactionData transaction in address.Transactions)
+                        // Get the UTXOs that are unspent or spent but not confirmed.
+                        // We only exclude from the list the confirmed spent UTXOs.
+                        foreach (TransactionData transaction in address.Transactions.Where(t => t.SpendingDetails?.BlockHeight == null))
```
dangershony: We call this again after a reorg right?

PR #3053, opened 2019-01-08
Versioning from VSTS - WORKING

monsieurleberre:
Just comment please do not merge yet... I need to test the use of VersionPrefix on VSTS first ;)
One thing to bear in mind is that the assembly.info is generated at build time in .Net Core


bokobza:
where do you keep the rest of the version number? By that I mean the `0.0.4-beta`?

monsieurleberre:
The rest is going to be stored in build variables :) I am still wondering if everything should not be going there actually.
Then we can have a manual build to release non beta versions for instance...
image

monsieurleberre:
Actually the above solution is only convenient if you have a single build version, I will put this back in the <VersionPrefix> and only append the build number from VSTS :)

src/FodyNlogAdapter/FodyNlogAdapter.csproj
```diff
@@ -3,9 +3,20 @@
...
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
   </PropertyGroup>
+  
+  <PropertyGroup Condition=" '$(Version)' == '' ">
+    <VersionPrefix Condition=" '$(VersionPrefix)' == '' ">3.0.4</VersionPrefix>
+    <Version Condition=" '$(VersionSuffix)' != '' ">$(VersionPrefix).$(VersionSuffix)</Version>
+    <Version Condition=" '$(Version)' == '' ">$(VersionPrefix)</Version>
+  </PropertyGroup>
+   
+  <PropertyGroup Condition=" '$(Version)' == '' ">
```
bokobza: Is this doubled-up on purpose?

src/FodyNlogAdapter/FodyNlogAdapter.csproj
```diff
@@ -3,9 +3,20 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageId>Stratis.FodyNlogAdapter</PackageId>
-    <Version>3.0.0.4-beta</Version>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
   </PropertyGroup>
+  
+  <PropertyGroup Condition=" '$(Version)' == '' ">
+    <VersionPrefix Condition=" '$(VersionPrefix)' == '' ">3.0.4</VersionPrefix>
```
bokobza: should it not be 3.0.0, rather than 3.0.4?

src/FodyNlogAdapter/FodyNlogAdapter.csproj
monsieurleberre: no that is my bad sorry

src/FodyNlogAdapter/FodyNlogAdapter.csproj
monsieurleberre: you want to use the build number as the last number so that you can be sure you don't end up with 2 packages having the same number, so on this occasion I have to push that number _4_ one level up

noescape00:
This PR is open for 23 days. 
@monsieurleberre @bokobza should we close it or someone will pick it up?

bokobza:
This PR is waiting to come in after the next release.

fassadlr:
Closing as we aren't versioning via VSTS. Should the need arise for this it can be reopened/investigated.

PR #3059, opened 2019-01-09
Reduce memory size by removing callbacks

dangershony:
Reduce memory size by not using the same callback for consensus download.

On testnet this takes away 40mb of ram during IBD (mainnet will be double the size)


noescape00:
Alternative solution would be adding an assumption that if callback is null then it's CM's callback, otherwise it's a custom one. 
That way we will be able to also avoid allocation for `DownloadedCallbacks` class creation. 

This solution will do but I think we can optimize it even more.

src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -1219,7 +1249,7 @@ private void ProcessDownloadQueueLocked()
                     request = blockPullerRequest;
                 }
 
-                this.blockPuller.RequestBlocksDownload(request.BlocksToDownload);
+                //this.blockPuller.RequestBlocksDownload(request.BlocksToDownload);
```
dangershony: ah ops I need to delete this.

dangershony:
>Alternative solution would be adding an assumption that if callback is null then it's CM's callback, otherwise it's a custom one.

yes I stared that way but the problem is we add a call back if its already there if we get an additional call to GetBlock form outside of CN

noescape00:
>yes I stared that way but the problem is we add a call back if its already there if we get an additional call to GetBlock form outside of CN

hash : null -- call CM callback
hash : list {callback} -- call only callback
hash : list {callback, null} -- call callback and CM callback

though it's also kinda hacky

dangershony:
Ah you mean adding a null callback to the list itself, hmm yeah seems over complicated and the additional allocated class is a very small additional foot print.

noescape00:
@dangershony up to you actually. Probably code without hacks is better than hacky but more optimal in terms of resources code.

PR #3060, opened 2019-01-09
Proven header size was not calculated correctly when inserting to store

dangershony:


src/Stratis.Bitcoin.Features.BlockStore/ProvenHeadersBlockStoreSignaled.cs
```diff
@@ -97,6 +97,9 @@ private void CreateAndStoreProvenHeader(int blockHeight, ChainedHeaderBlock chai
 
             ProvenBlockHeader newProvenHeader = ((PosConsensusFactory)this.network.Consensus.ConsensusFactory).CreateProvenBlockHeader(block);
 
+            // Calculate the PH size.
+            newProvenHeader.ToBytes(this.network.Consensus.ConsensusFactory);
```
MithrilMan: I think this should be placed in
```
public ProvenBlockHeader CreateProvenBlockHeader(PosBlock block)
{
   return new ProvenBlockHeader(block);
}
```

or in the ProvenBlockHeader(Block) constructor itself

fassadlr: +1 what @MithrilMan said.. then we always have the size of the header.

PR #3061, opened 2019-01-09
Added coinview test

noescape00:


src/Stratis.Bitcoin.Features.Consensus.Tests/CoinViews/CoinviewTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            this.rewindDataIndexCache.InitializeAsync(this.concurrentChain.Height, this.cachedCoinView);
+
+            this.random = new Random();
+
+            ChainedHeader newTip = ChainedHeadersHelper.CreateConsecutiveHeaders(1000, this.concurrentChain.Tip, true, null, this.network).Last();
+            this.concurrentChain.SetTip(newTip);
+        }
+
+        [Fact]
+        public async Task TestRewindAsync()
```
dangershony: This tests only works on the cached data right? (I did not notice a call to flush while you aggregate)

dangershony: We should probably add to the test a mixture of flush and none flush cases, should be fairly easy no?

I suspect the bip30 bug happened after a double reorg meaning we cleared the cache on the first reorg.

PR #3065, opened 2019-01-10
[SC] Add statics tests

### LINKED ISSUE #2037: [SC] Remove static support
Remove static support:
* They don't make sense in a contract context at the moment. The contract's Type should never need to define methods that are separate from the contract instance.
* Static functionality can be emulated by defining a method that does not use 'this'.
* If there's a future use case we can introduce support.

Criteria:
* Static fields and methods on a top-level contract should cause validation to fail.
* Static fields and methods in nested types should cause validation to fail.
* Static constructors should cause validation to fail.

@codingupastorm

rowandh:
Add some tests to demonstrate functionality. See: #2037.

### DIFF
--- a/src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
+++ b/src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
@@ -1065,7 +1065,7 @@ public class Test : SmartContract
     public string SumStr() 
     {
         var summation = """";
-		var strings = new [] { ""1"",""2"",""3"",""4"",""5"",""6"",""7"",""8"",""9"",""10""};
+        var strings = new [] { ""1"",""2"",""3"",""4"",""5"",""6"",""7"",""8"",""9"",""10""};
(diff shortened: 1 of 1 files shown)


src/Stratis.Bitcoin.Features.SmartContracts.Tests/DeterminismValidationTests.cs
```diff
@@ -1037,5 +1037,157 @@ public class Test2 {
...
+            ContractCompilationResult compilationResult = ContractCompiler.Compile(adjustedSource);
+            Assert.True(compilationResult.Success);
+
+            byte[] assemblyBytes = compilationResult.Compilation;
+            IContractModuleDefinition decomp = ContractDecompiler.GetModuleDefinition(assemblyBytes).Value;
+
+            var result = this.validator.Validate(decomp.ModuleDefinition);
+
+            Assert.False(result.IsValid);
+            Assert.NotEmpty(result.Errors);
```
codingupastorm: Would be nice to improve the error verification across all of these in the future, along with const strings for error messages.

PR #3069, opened 2019-01-10
[SC] [Experimental] Shared node builder for tests and daemon

codingupastorm:


src/Stratis.SmartContracts.Networks/NetworkNodes.cs
```diff
@@ -0,0 +1,51 @@
...
+using Stratis.Bitcoin.Features.SmartContracts.PoW;
+using Stratis.Bitcoin.Features.SmartContracts.Wallet;
+
+namespace Stratis.SmartContracts.Networks
+{
+    public static class NetworkNodes
+    {
+        public static IFullNodeBuilder GetPoWSmartContractNodeBuilder(NodeSettings settings)
+        {
+            return new FullNodeBuilder()
```
bokobza: The problem with this is that the Networks project will then need to have a reference to all these other projects.
Someone that would want to run a very lean node, or even that just wants to get the Network types, will have to download all these other dependencies.

codingupastorm: Yeah agreed, I think it shouldn't go in .Networks. 

The important part though is that we have a consistent build ordering method _somewhere_ to be consumed by both tests and daemons.

Any ideas where you would put it @bokobza ?

bokobza:
Not sure there is a good place for that, really.
We should just use the program.cs ones as the reference and then copy from there. 

Another idea is to have another project that only contains nodeBuilds definitions that will be used by both programs.cs and integration tests, but that's yet another project...

dangershony:
While I get the change I don't see the point.
I don't mind if you want to build the SC node in a different place though.

PR #3072, opened 2019-01-10
[API] Start the API earliest and make consensus information available to node status

### LINKED ISSUE #3021: [Wallet] Start API on rewinding
Currently when the node starts rewinding on startup, the API is not started.
Core UI can't detect the API so it shows the unable to connect message, advising to restart Core and try again. Doing this will also stop the rewind process.

Is there anyway to start the API and give me a rewinding status? Preferably with an "x blocks left" tied to it so I can display it in the UI?

We can generalise this issue to start the API before anything really, this way you can include the given exception in the status call if anything goes wrong, allowing me to display it in the UI.

fassadlr:
I'm not sure if this is the correct way to do this but can you guys please comment whether it is OK to add consensus status messages to the API in this way?

### DIFF
--- a/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
+++ b/src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -168,6 +167,9 @@ namespace Stratis.Bitcoin.Consensus
         }
 
         /// <inheritdoc />
+        public string StatusMessage { get; set; }
+
+        /// <inheritdoc />
         /// <remarks>
         /// If <see cref="blockStore"/> is not <c>null</c> (block store is available) then all block headers in
         /// <see cref="chainedHeaderTree"/> will be marked as their block data is available.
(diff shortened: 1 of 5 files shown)


src/Stratis.Bitcoin/Consensus/ConsensusManager.cs
```diff
@@ -190,9 +193,13 @@ public async Task InitializeAsync(ChainedHeader chainTip)
                 pendingTip = chainTip.FindAncestorOrSelf(consensusTipHash);
 
                 if ((pendingTip != null) && (this.chainState.BlockStoreTip.Height >= pendingTip.Height))
+                {
+                    this.StatusMessage = null;
                     break;
+                }
 
                 this.logger.LogInformation("Consensus at height {0} is ahead of the block store at height {1}, rewinding consensus.", pendingTip, this.chainState.BlockStoreTip);
+                this.StatusMessage = string.Join("Rewinding consensus from height {0} to {1}.", pendingTip.Height, this.chainState.BlockStoreTip.Height);
```
bokobza: I'd rather have a method that returns the state of the ConsensusManager, and the nodeController call this method and decide what to do with it.
The state doesn't have to be an enum but can be an object containing the tip, the rewinding or not status etc..

fassadlr: @bokobza see my comments here #3021 .. can you comment on my last comment there?

fassadlr:
Closing as will be submitting another PR for this.

